PR #112, opened 2015-04-20
Syncing pom with maven-jsf-plugin current version

hatemalimam:
In order to make sure that the project would build for new comers
(empty .m2 repo) a necessary update to maven-jsf-plugin version in pom.xml is
recommended. And since the current version of maven-jsf-plugin is 1.3.3-SNAPSHOT the pom file is update accordingly.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -274,7 +274,7 @@
             <plugin>
                 <groupId>org.primefaces</groupId>
                 <artifactId>maven-jsf-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


cagataycivici:
We rather not use a snapshot version and this case is mentioned at our wiki;

https://github.com/primefaces/primefaces/wiki/Building-From-Source

hatemalimam:
Oh I haven't noticed the wiki. :/

PR #116, opened 2015-04-20
Added README

hatemalimam:
An initial commit for README.md file

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,95 @@
+# PrimeFaces
+
+**Ultimate JSF Framework**
+
+![PrimeFaces icon](http://blog.primefaces.org/wp-content/uploads/2011/08/bloglogo.png)
+
+
+### Overview
+***
+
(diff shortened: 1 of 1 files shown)


cagataycivici:
@tandraschko @mertsincan What do you guys think? Doesn't it duplicate information at pf.org so we have to maintain two separate introductions. Should we use something like this or just add a link to primefaces.org?

tandraschko:
+0 as the it's really duplicate content
It would be better to bring some new content like "how to create good issues":
- tested on newest versions
- provide a xhtml and bean snippet e.g. via gist
- ...
- ...

hatemalimam:
Well I don't agree 100% that this is +0, that would be a clear underestimating of github's indexing and search results. Considering the fact not every user knows PrimeFaces, and github is a good source to let the people know. The current content might not provide the utimate desired goal. But still having this file is a plus from my point of view. And a lot of projects would have duplicate info like https://github.com/twbs/bootstrap/blob/master/README.md . Also the contribution part is still blury. An idea would be that if someone submits a pull-request, they should create a new branch and ask to merge with a specific one rather than or to the master. That could be a helpful part of the README. I hope I'm helping here.

cagataycivici:
Applied and edited with some changes, thank you.

PR #122, opened 2015-04-21
Display inplace as blocked

MilanRS:
Option to set inplace width in pixels or percents like in p:column.
It is great option if you have inputtext in p:inplace and input text is 1 letter length and you want to have editable area to be 50px width or 75% of parent element.

Example:
no width set: http://i.imgur.com/NyH4KpT.jpg
width set to 90%: http://i.imgur.com/Btp2qr3.jpg
http://i.imgur.com/MHNk7XS.jpg

Minor changes in code.
Proposed on code.google.com Jul 11, 2013 and later updated with needed code.


tandraschko:
can be done by css

MilanRS:
Are you sure? If you make it inline-block, after editing it will have `display: inline` again.

tandraschko:
Yup, via styleClass.

Issue #127, opened 2015-04-21
Schedule: Selectable feature


cagataycivici:
Fullcalendar "Allows a user to highlight multiple days or timeslots by clicking and dragging".

http://arshaw.com/fullcalendar/docs/selection/selectable/
http://arshaw.com/js/fullcalendar-1.5.4/demos/selectable.html

It's an incredibly good feature and it would be great if the primefaces schedule component could expose this feature.

mdeedub1

tryptichon:
Hi.

I took a peek into this. It seems, that a new event is needed to fulfill this request, because you have to transmit a start and an end date. Event "dateSelect" uses an java object "SelectEvent", but the only parameter you can set there is one object - selecting a range would need two.

First of all, we'd need a new ajax event "rangeSelect".

Then something like a class SelectRangeEvent has to be implemented.

``` java
class SelectRangeEvent extends AjaxBehaviorEvent {
  private Object start;
  private Object end;
```

then somewhere in Schedule.java:

``` java
SelectRangeEvent selectRangeEvent = 
  new SelectRangeEvent(this, behaviourEvent.getBehaviour(), startDate, endDate);
```

OR 

you still use SelectEvent, but the transmitted Object has the start and end date as properties inside. This way we'd need something like a class SelectRange with both properties for SelectEvent.

``` java
class SelectRange {
 private Object start;
 private Object end; ... 
```

then somewhere in Schedule.java:

``` java
SelectEvent selectRangeEvent = 
  new SelectEvent(this, behaviourEvent.getBehaviour(), new SelectRange(startDate, endDate);
```

What would be the best way to tackle this?

cagataycivici:
I'd go with ScheduleRangeSelectEvent or DateRangeSelectEvent to be more generic.

tryptichon:
Argh. I went with SelectRangeEvent now. But since this Event might be used with generic ranges (i.E. numbers) by other widgets, I hope it will be accepted nevertheless. ;-)
See https://github.com/primefaces/primefaces/pull/159

PR #136, opened 2015-04-21
Correct way to calculate the time with DST across the whole schedule. Di...

tryptichon:
Correct way to calculate the time with DST across the whole schedule. Disregard the offset in "ScheduleRenderer.java" - it's no use if you select a date across a DST border. Calculate the right offset inside queueEvent from UTC instead, because now you have the date from the schedule to correctly calculate the offset.


tryptichon:
Careful: I needed this for https://github.com/primefaces/primefaces/issues/127, so the according pull request https://github.com/primefaces/primefaces/pull/159 supersedes this here.

cagataycivici:
Added with a minor change, thanks.

PR #137, opened 2015-04-21
Added attribute "showWeekNumbers" to schedule.

tryptichon:
Activate the attribute weekNumbers in fullCalendar.io to display the current week of the year in the schedule.


cagataycivici:
Why do we also need timeFormat here?

tryptichon:
We don't. The patch removes the line with timeFormat and adds the same line again without the ';' at the end because the attr() for "weekNumbers" follows.

cagataycivici:
Missed that sorry, I'll review, thank you.

cagataycivici:
Could you please create an issue for this as well?

PR #138, opened 2015-04-21
Added attribute "viewFormat" to schedule. 

tryptichon:
Configuration as in http://fullcalendar.io/docs/views/View-Specific-Options/
i.E. &lt;p:schedule viewFormat="{ \"week\":{ \"columnFormat\":\"ddd, DD.MM\" }}" ...&gt;
(Remember: Object attribute names must be enclosed in double quotes like "week" or jQuery.parseJSON won't be able to parse it.)

### DIFF
--- a/src/main/java/org/primefaces/component/schedule/ScheduleRenderer.java
+++ b/src/main/java/org/primefaces/component/schedule/ScheduleRenderer.java
@@ -193,6 +193,11 @@ public class ScheduleRenderer extends CoreRenderer {
             wb.attr("columnFormat", columnFormat, null);
         }
             
+        String viewFormat = schedule.getViewFormat();
+        if(viewFormat != null) {
+            wb.attr("viewFormat", viewFormat, null);
+        }
+
         encodeClientBehaviors(context, schedule);
 
         wb.finish();	
--- a/src/main/resources-maven-jsf/ui/schedule.xml
+++ b/src/main/resources-maven-jsf/ui/schedule.xml
@@ -211,6 +211,13 @@
             <defaultValue>false</defaultValue>
             <description>Displays description of events on a tooltip, default value is false.</description>
         </attribute>
+        <attribute>
+           <name>viewFormat</name>
+           <required>false</required>
+           <type>java.lang.String</type>
+           <description>Complete view format configuration as JSON Object, see http://fullcalendar.io/docs/views/View-Specific-Options/ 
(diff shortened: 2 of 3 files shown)


cagataycivici:
But this introduces new viewFormat attribute, for backward compatibility, we should use columnFormat.

tryptichon:
The problem is, that the JS source for the FullCalendar on which this schedule.js is based upon, jumped from Version 1.6.x to Version 2.2.5. The methods to decode the old columnFormat-string are gone, and as far as I can see, these were hand written parsers. Are we sure we want to backport that?

I might be able to parse the old string using regular expressions though.

Like this (error handling aside):

`// i.E. this.cfg.columnFormat = "month:'ddd',week:'ddd, MM/dd',day:'dddd, MM/dd/yyyy'";`
`var monthFormat = columnFormat.match(/['"]?month['"]?[\s]*:[\s]*['"](.*?)['"]/i)[1];`
`var weekFormat = columnFormat.match(/['"]?week['"]?[\s]*:[\s]*['"](.*?)['"]/i)[1];`
`var dayFormat = columnFormat.match(/['"]?day['"]?[\s]*:[\s]*['"](.*?)['"]/i)[1];`
`this.cfg["views"] = { month: { columnFormat: monthFormat}, week: { columnFormat: weekFormat}, day: {columnFormat: dayFormat }};`

That should emulate the old behaviour, although it might not be 100% identical.

cagataycivici:
I believe we did a pretty good job at keeping Schedule backward compatible except this one. We had logs in a couple of areas for other removed options.

My suggestion would be use columnFormat and parse for backward compatibility and give a warning that there are new *Format attributes like monthFormat, weekFormat and dayFormat which are separate attributes.

So if you could change your request to something like this, we can accept it, otherwise I plan to implement this approach which is similar to what we done before in schedule.

tryptichon:
Okay. So I better kick my proposed attribute viewFormat completely and just use columnFormat for compatibility, then introduce monthFormat, weekFormat and dayFormat?

cagataycivici:
Yes, columnFormat for the backward compatibility and new dayFormat, weekFormat and monthFormat attributes as the new way. Both should be supported, if columnFormat is defined, we should convert it somehow and give a log to the user to prefer new *Format attributes instead.

tryptichon:
Alright, will do. :)

tryptichon:
https://github.com/primefaces/primefaces/pull/160 ;-)

PR #141, opened 2015-04-22
Update to README.md

srbala:
Typo in mobile namespace and link to showcase source project.


srbala:
Since the patch is correction to README.md file, I am not sure a forum discussion is required. If necessary I will create one.

cagataycivici:
Thank you.

PR #143, opened 2015-04-22
fixed problem with minified version

ktor:
it's now minified to:
    function(a){if(a.PrimeFaces){b.debug("PrimeFaces already loaded, ignoring duplicate execution.");return}

which caused "Cannot read property 'debug' of undefined" JS error when loaded two times on a page - common case in portal environment where portlets use primefaces, see following report:
http://stackoverflow.com/questions/29521708/ajax-file-upload-error-in-jsf-primefaces-cannot-read-property-debug-of-undef

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/core/core.js
+++ b/src/main/resources/META-INF/resources/primefaces/core/core.js
@@ -1,7 +1,7 @@
 (function(window) {
     
     if(window.PrimeFaces) {
-        PrimeFaces.debug("PrimeFaces already loaded, ignoring duplicate execution.");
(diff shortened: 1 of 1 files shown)


cagataycivici:
PrimeFaces resides in window namespace so it should not make any difference.

@tandraschko @mertsincan What do you guys think?

tandraschko:
I think it may be a "optimization" by the minifier plugin.
+1 for applying it, it doesn't harm and makes sense as the if statement before also check "window.PrimeFaces"

cagataycivici:
Thank you, merging it.

PR #159, opened 2015-04-22
Expose "selectable" feature to select ranges in the schedule.

tryptichon:
See Issue https://github.com/primefaces/primefaces/issues/127


cagataycivici:
Is there any relation with UTCMillisToTimezoneDate and this new event? I don't understand why do we need to change these just to add a new event.

tryptichon:
Since this here needed the calculation of milliseconds to local date multiple times, I transferred the process into an own method and I used the calculation I suggested in https://github.com/primefaces/primefaces/pull/136.

If that's not wanted, I can put the old calculation into UTCMillisToTimezoneDate.

Nevertheless https://github.com/primefaces/primefaces/pull/136 and this pull here collide, so only one of them should be merged while the other one should be discarded.

HTH

tandraschko:
It currently conflicts - please provide a new PR!

PR #161, opened 2015-04-22
Corrected .gitignore

ova2:
I've corrected .gitignore for all availabe IDEs and cases.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,8 @@
-/target/
-.settings/
+.*
+target
+*.?ar
+*.iml
+*.ipr
+*.iws
+*.NavData
+*.navData
--- a/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
+++ b/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
@@ -135,14 +135,13 @@ PrimeFaces.ajax = {
 
         delays: {},
 
-        requests: new Array(),
+        requests: [],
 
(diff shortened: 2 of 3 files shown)


tandraschko:
Oleg, please don't mix different stuff.
Your pull requests also don't seems correct:
-                var sourceId = null,
-                $this = this,
-                var $this = this,

ova2:
Well, multiple Pull Requests are merged automatically to one. Sorry. I don't have control for that

Thomas, why it is not correct? I just followed advices by IntelliJ and rewritten:

``` javascript
var sourceId = null,
$this = this,
sourceId = ...
```

to the

``` javascript
var $this = this,
sourceId = ...
```

It is correct. In the old code piece, `sourceId` was declared twice!

P.S. In the fact, the PrimeFaces JS code needs to be optimized. I think nobody looked really into the JS code and tried to optimize it. I would check it first with `JSLint` and then declare `"use strict"` directive as many JS libs do that. I'm sure you can do then some basic optimizations.

tandraschko:
Ahh, i see. Yep looks fine.
@cagataycivici Would you like to review first?

srbala:
When all changes are in same branch (in this case [master](https://github.com/ova2/primefaces/commits/master)), github tends to pull all-in-one together. Fixes needs to be in isolated branch for separate pull requests.

cagataycivici:
Please create separate pull requests with references to issues.

Issue #133, opened 2015-04-21
Add URL option to ScheduleEvent


cagataycivici:
Expose the url option at the fullcalendar to our java API

http://arshaw.com/fullcalendar/docs/event_data/Event_Object/

tryptichon:
Hi.

I tried that in my app, but the results are not what I expected. Without an URL you click an event in the schedule to edit it. When the URL is set though,clicking such an event will always call the url behind it. So how do you distinguish between "I want to go to that URL" and "I want to edit that event" by clicking on it, in other words: How do you edit an event that has an URL set? 
Making the title of the event a link to that URL might be a solution. What do you think?

EDIT: The whole box is an "a"-Tag, so setting another "a"-Tag inside won't work :(

Regards.

tryptichon:
After some investigation in http://fullcalendar.io/docs/mouse/eventClick/ I found, that returning "false" in this.cfg.eventClick prevents the URL to be opened. So we can have either the call to the URL and the event happening simultaneously, or only the event itself. :(

Maybe "false" should always be returned, so the application handles the event by itself and then decides, what to do with the URL. Would that suffice?

cagataycivici:
If URL is defined on event maybe we should not bind the event click?

tryptichon:
The documentation and comments in the source code indicate, that the way to prevent the URL to be called, is having "eventClick" return false. There seems to be no way to prevent the event itself to be bound to the click (see line 3385 in schedule.js)

``` javascript
click: function (seg, ev) {
    return view.trigger('eventClick', this, seg.event, ev); // can return `false` to cancel
},
```

The only clean way to do this is (imho) inside of this.cfg.eventClick itself:

``` javascript
        this.cfg.eventClick = function(calEvent, jsEvent, view) {
            if($this.cfg.behaviors && ! calEvent.url) {
[...]
            }
        };
```

It should be written down somewhere, that setting the URL-property overrides the event. It confused the hell out of me the first time I encountered it. ;)

Change commited to pull request. :)

P.S.: Another caveat: The URL will always be called in the same window. That might be confusing as well. Maybe we can introduce an attribute "urlTarget" to the schedule component to control the target of the URL and use the this.cfg.eventClick to call that url as proposed in http://fullcalendar.io/docs/mouse/eventClick/?

PR #164, opened 2015-04-23
Additional field "URL" for Schedule Events.

tryptichon:
For Issue https://github.com/primefaces/primefaces/issues/133. The Url is set in the event, although clicking on it in the schedule will only trigger the event, never open the Url directly. That's up for the Java-Application that catches the event to handle.


kukel:
Can someone of the core team either fix this or close it? Having an interesting pull request open for over half a year without any remark/comment is not appreciated

kukel:
ping... someone?

tandraschko:
If you would fix the conflicts, i would apply it.

tandraschko:
Nevermind, i merged it and will commit it now. Would you please give it a test-try @kukel?

Issue #167, opened 2015-04-24
orderlist: unselect event not fired often enough


hoax:
I found some strange behaviour with the new (since v5.2) events on orderList. The main problem here is that the unselect event is not fired often enough. Imagine a simple list with 3 elements "a", "b" and "c":
- click on "a" -> select "a" event fired -> "a" is selected
- click on "b" -> select "b" event fired -> "b" is selected
- control key + click on "c" -> select "c" event fired -> "b" and "c" are selected

No unselect event is fired at the time "b" is clicked. So now my backing bean can't know whether all the items are currently selected or "b" and "c" or maybe just "c".

Same problem appears to be with the reorder event:
Moving a selected item via drag and drop fires a reorder event but no unselect event although the selection is explicitly removed by orderlist.js afterwards.
Moving a selected item via the action buttons fires a reorder event likewise, but here the selection is kept.

This behaviour can be seen on the showcase for orderlist too.

---

I am using the orderlist to let the user create and sort a list of strings. With additional buttons the user can add new entries and delete the currently selected. The described behaviour prevents me to provide a correct delete functionality as I can never know which elements are currently selected.

cagataycivici:
I've taken another approach, please see #185.

Instead of sending concurrent itemUnselectEvent requests to the server, SelectEvent now has metaKey,ctrlKey boolean properties. So when you listen to itemSelect event, if metaKey is on then previous selections are not unselected so kept. 

If it is false, then it means there is only one selection. I also plan to use this generic way in other components that have similar key bindings.

hoax:
In my opinion #185 is just a workaround. If there is an unselect-event I expect it to fire when an unselect happens, regardless if this information can also be retrieved some other way. Otherwise the unselect-event should be renamed to somewhat else as the name does not reflect its actual behaviour.

cagataycivici:
With your code changes assume user selects 100 items with cmd key and selects the 101th one without metakey, then server will get 100 ajax requests for unselect. This is far from idea, with #185, there is only one request and you can know what is selected and unselected.

hoax:
You are right, that's one disadvantage of my current implementation but nothing that can't get fixed. How about processing your event from #185 on the backend and fire unselect-events from there (I guess that's possible though I'm pretty new to JSF).

cagataycivici:
Thank you but we'll not take any further action regarding these.

PR #168, opened 2015-04-24
improve orderlist's events to represent what really happend on the ui

hoax:
Fixes issue #167.

I fixed the events to work as (I think) they are supposed to be:
Every unselect in the UI results in a corresponding unselect event.
According to my former example (see issue #167) that is:
- click on "a" -> select "a" event fired -> "a" is selected
- click on "b" -> unselect "a" event fired -> select "b" event fired -> "b" is selected
- control key + click on "c" -> select event fired -> "b" and "c" are selected

Also after drag and drop the correct unselect event is fired.

(in onDragDrop fireReorderEvent has to be fired before fireItemUnselectEvent, otherwise the object associated with the event on server side is not the dragged (and selected) item but instead the item that was previously displayed at the drop position).


cagataycivici:
See my comment at #167 please.

Issue #51, opened 2015-04-18
Pagination of lazy data with an unknown total row count


cagataycivici:
Support paginator even when total row count is not specified.

cagataycivici:
Review for 5.2.4.

cagataycivici:
@kukel Is there PR about this? Can't find it.

cagataycivici:
I couldn't understand anything from the patch. I need to do a discussion, anyone up for this? @kukel, @erik-wramner ?

erik-wramner:
Sure. The two patches are related, but I'll go through mine.

Normally the row count is known when the dataTable is first rendered. However, in my use case it is not. The problem is that counting the rows is so expensive, so we don't want to do it. It is much faster to just show the first page. We can navigate to the next page without problems, but we don't know the actual row count until we reach the last page - probably never. For live scrolling the dataTable initializes scrollLimit when it is first renedered and in our case it will be set to pageSize + 1, so the table knows that there is more to read. When the user scrolls down the dataTable will call loadLiveRows in order to get the next (and as it thinks final) page from the server. The server reads the next page and unless it is the last one it knows that there are at least (2 \* pageSize + 1) rows and modified the row count for the LazyDataModel accordingly. However, we need to notify the client-side code as this is an AJAX call. The patch in DataTableTemplate fixes that, totalRecords will be sent to the client side in the response.

The patch in oncomplete modifies the method signature so that the args (containing totalRecords) are available and sets the scrollLimit to the new row count value.

Finally I have had problems with live scroll keeping to load rows even though all have been read, so the patch in loadLiveRows adds an extra check that prevents that from happening.

Does that help? If not just ask.

kukel:
@cagataycivici: There is: https://github.com/primefaces/primefaces/pull/224

What the patch does is largely the same as what @erik-wrammer described. After each page event  check if the 'new' rowCount is larger than the current rowcount. If so, it still enables the 'next' page button. That is all. What effectively happens is to take the new rowcount that is send from the server into account.

The client is responsible then, in the load method of the lazydatamodel, to update the rowcount each time if the end is not reached. How to use this is to retrieve 'page-size' + 1 rows in a lazy datamodel and only retrieve 'page-size' records, but set the new rowCount at least one higher than the new total rowcount (previous rowcount + page size).  For an example check http://forum.primefaces.org/viewtopic.php?f=3&t=31126

kukel:
[...]

PR #224, opened 2015-05-04
Fix for Pagination of lazy data with an unknown total row count (#51)

kukel:
Tested this and it works. Is a showcase example needed? Or some other kind of documentation?


kukel:
Can someone of the core team either fix this or close it? Having an interesting pull request open for over half a year without any remark/comment downright is not appreciated

cagataycivici:
Merged manually.

Issue #38, opened 2015-04-17
Tooltips for RowEditor icons


cagataycivici:
Add *title attributes to rowEditor component icons.

PR #225, opened 2015-05-05
Added title attributes for edit,save and cancel icons

sudheerj:
Tested on showcase example and it works.


cagataycivici:
Is there an issue ticket here?

sudheerj:
Yes,the issue is available here https://github.com/primefaces/primefaces/issues/38. Only the first commit is related to this pull request and other commits not belonging to this pull request.

cagataycivici:
There are duplicate calls to get*Title which are unncessary also there is a unrelated commit to schedule.

Issue #229, opened 2015-05-05
Selectable Datatable without LazyDataModel


indracrg:
In the Issue [#7173](https://code.google.com/p/primefaces/issues/detail?id=7173&can=1&q=label%3DTargetVersion-5.2%20status%3AFixed&colspec=ID%20Stars%20Type%20Status%20Priority%20TargetVersion%20Owner%20Summary&num=500) the behavior of getRowData changed and now the only model that applies is LazyDataModel, if I keep implementing SelectableDataModel (as in 5.1) the code goes to the else part and this.getRowKey is always null because I don't have in my html the rowKey parameter. For me having a Data Model and the rowKey parameter is redundant. 

Maybe this is in different direction but also it seems that I can't extend from ArrayDataModel anymore.

PR #230, opened 2015-05-05
Fix getRowData method for SelectableDataModel use

indracrg:
With 5.2 we noticed that the new implementation of getRowData excluded the possibility of the use of the SelectableDataModel without LazyDataModel, forcing to add rowKey parameter in the html (#229). For those who has the same problem here you have our fix.

### DIFF
--- a/src/main/java-templates/org/primefaces/component/datatable/DataTableTemplate.java
+++ b/src/main/java-templates/org/primefaces/component/datatable/DataTableTemplate.java
@@ -710,30 +710,35 @@ import org.primefaces.util.SharedStringBuilder;
     }
 
     public Object getRowData(String rowKey) {
-        if (getDataModel() != null && getDataModel() instanceof LazyDataModel) {
-            DataModel model = getDataModel();
-            if(!(model instanceof SelectableDataModel)) {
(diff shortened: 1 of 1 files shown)


tandraschko:
Your PR breaks the selection without datamodel but i fixed it.

indracrg:
tandraschko Thank you for the fix! We hope to see soon a version with it.

cagataycivici:
Is your fix in 5.1.18 and 5.2.3 @tandraschko

tandraschko:
only 5.2.4. I changed it in 5.2 trunk some months ago, so this doesn't affect 5.1.x.

cagataycivici:
What was the original change? Why it needs to be fixed?

tandraschko:
can't remember the issue number but the rowKey attribute, if available, was still used when you have a LazyDataModel. So i changed to logic that always LazyDataModel is used but i forgot the "normal" SelectableDataModel case.

matsandreassen-triona:
This one? https://code.google.com/p/primefaces/issues/detail?id=7173

tandraschko:
Yup but without any description. Discussed it with Cagatay ~1 year ago per mail.

PR #233, opened 2015-05-06
Add itemSelect event support for SelectOneMenu in community version

sudheerj:


cagataycivici:
Not valid pull request as it contains other commits, I think you need to create a branch for each one.

PR #245, opened 2015-05-07
Add internationalization support for weekNumberTitle property of schedule component

sudheerj:
Add internationalization support for weekNumberTitle property of schedule


cagataycivici:
Should better be retrieved from Locales at js side.

Issue #129, opened 2015-04-21
Global contextMenu overlays component specific ones


cagataycivici:
Different components with different specific contextMenus (attribute "for") fire context menus properly, to say, each component its own contextMenu. But if you define one generic contextMenu (no attribute "for") in the form, it will overlay all the specific ones. It can be an event bubbling problem.

http://forum.primefaces.org/viewtopic.php?f=3&t=27672

PR #247, opened 2015-05-07
Fix global contextMenu overlays component specific ones

sudheerj:
The issue was created here #129. Also we removed some unnecessary calls of preventDefault() functions because it is already existed in show() function.


sudheerj:
The issue was created here https://github.com/primefaces/primefaces/issues/129.We removed some unnecessary calls of preventDefault() functions because it  is already existed in show() function.

cagataycivici:
Looks good, thanks.

Issue #128, opened 2015-04-21
OverlayPanel contextMenu event should block browser menu


cagataycivici:
``` xml
<p:overlayPanel showEvent=contextmenu, ...>
```

Browser context menu is also shown on right click, which should be blocked.

PR #248, opened 2015-05-07
Fix for OverlayPanel contextMenu event should block browser menu

sudheerj:
The actual issue was created here https://github.com/primefaces/primefaces/issues/128


kukel:
Please refer to the original issue number with just  a 'hash' and the number #128

Issue #104, opened 2015-04-19
Remove disabledSelection from Column API


cagataycivici:
It was deprecated since we moved it into datatable so time to remove.

cagataycivici:
Thank you @sudheerj

PR #249, opened 2015-05-07
Remove disabledSelection from Column API

sudheerj:
The issue was created here https://github.com/primefaces/primefaces/issues/104


sudheerj:
The pull request is as above 1ec300f and the issue is #104

Issue #238, opened 2015-05-06
ActiveIndex for ImageSwitch


cagataycivici:
Add activeIndex option to set the initial image to be displayed.

PR #252, opened 2015-05-08
Add activeIndex for ImageSwitch

sudheerj:
The issue was created here https://github.com/primefaces/primefaces/issues/238


sudheerj:
The code indentation is proper in my IDE but it got slightly changed in primefaces master repo :)

cagataycivici:
I think type should be int.

Issue #126, opened 2015-04-21
boxWidth and boxHeight for ImageCropper


cagataycivici:
The ImageCropper component lacks an option to set the displayed image size. In the current state I don't see a way to limit the maximum size when using images with higher resolution without any preprocessing.
The underlying JavaScript library JCrop supports this feature by using the the boxWidth and boxHeight parameters.
Simply adding these attributes to the renderer code seems to be enough to enable this feature in the ImageCropper component.

cagataycivici:
Reverted as values should be 0 as integer by default.

neXus1987:
in Primefaces 7.0 the ImageCropper boxWidth and boxHeight won´t work as aspected.  The image is smaller, but the selection area is too big. As i googled, there is something with trueSize for jCrop, maybe this will help!?

PR #262, opened 2015-05-10
Add boxWidth and boxHeight for ImageCropper

sudheerj:
The issue was created here #126


cagataycivici:
Reverted as values should be 0 as integer by default.

sudheerj:
You are right.The default values of JCrop boxWidth,boxHeight are 0 but not null.

PR #284, opened 2015-05-14
ColumnGroup footer issue with DynamicColumns

sudheerj:


cagataycivici:
Fixed #299 using the suggestion, thank you.

PR #289, opened 2015-05-15
Fix #287 live scroll with unknown row count

### LINKED ISSUE #287: Support lazy live scroll in dataTable with unknown row count
Issue #51 adds support for pagination with unknown row count. I would like to see the same support for the lazy-loaded live scroll in the data table. I have a working solution (will soon be available as a fork) that I would like to submit.

Support for an unknown row count is useful in many situations where counting the records would be expensive and where users typically browse through only a few pages, not all. This is even more natural with the live scroll feature than with standard paging, so it seems like a sensible thing to add.

erik-wramner:


kukel:
Can someone of the core team either fix this or close it? Having an interesting pull request open for over half a year without any remark/comment is not appreciated

cagataycivici:
Merged manually

PR #310, opened 2015-05-19
When entering an invalid value, Multiple Autocomplete with required='true' doesn't add the 'ui-state-error' class.

mertsincan:


cagataycivici:
Looks good, please create an issue and merge. Also for Elite branches.

Issue #152, opened 2015-04-22
ShowEffect option for DialogFramework


cagataycivici:
Add support for showEffect customization.

cagataycivici:
Part in DialogNavigationHandler could have type checking instead of implicitly checking for a specific attribute name like positionType.

PR #323, opened 2015-05-21
Add showEffect and hideEffect for Dialog framework

sudheerj:
The issue was created here #152


cagataycivici:
Part in DialogNavigationHandler could have type checking instead of implicitly checking for a specific attribute name like positionType.

Issue #329, opened 2015-05-22
AutoComplete spams query request in IE8


VsevolodGolovanov:
IE 8.0.6001.18702.

Reproducing.

```
<h:form>
    <p:autoComplete completeMethod="#{bean.completeText}" />
</h:form>
<p:log />
```

(Complete method returns something, whatever).
Type any text in the autocomplete. One query happens.
Now mousehover above the autocomplete field. Notice a bunch of query requests happening in p:log. That's problem 1.
Try selecting an item from a suggestion list. See that another query request happens, shows suggestion list again, messes up autocomplete value. That's problem 2.

Reproducible on the current git HEAD.

cagataycivici:
@mertsincan , please review.

mertsincan:
I tested this issue on IE 8.0.6001.18702, but I didn't see this bug and attached my test video to dropbox;
https://www.dropbox.com/s/efcbf1jtd6qkhwg/autocompleteTest.flv?dl=0

VsevolodGolovanov:
Reviewed a bit more.
AutoComplete relies on browser detection in its scripts.
The issue is reproducible with an "intranet" site and a `X-UA-Compatible`/`IE=edge` meta tag or header, when IE is working in a "Browser Mode: IE8 Compat View", "Document Mode: IE8 Standards".
`IE=edge` fixes the document mode, but does nothing to the browser mode. So the browser detection doesn't work: `PrimeFaces.isIE(8)` returns false. It's possible to improve browser detection, e.g. by checking `navigator.userAgent` for Trident version.

If the initially suggested fix is considered invalid, then I'm still curious about the purpose of `$this.suppressInput` - it's set to false on the first keydown, never to be set to true again. Optimization? It does break the "copy-paste with a mouse" case.

cagataycivici:
Compat view is not supported by PF. We test against actual IE versions, without downgrades via compat.

PR #330, opened 2015-05-22
AutoComplete spams query request in IE8 #329

VsevolodGolovanov:
Fixed suppression, removed unnecessary preventDefault.


VsevolodGolovanov:
Fix tested in IE8, IE9, Chrome 43, Firefox 35, Opera 12.

cagataycivici:
As the issue can't be replicated, marking as won't fix. @mertsincan any comments?

PR #347, opened 2015-05-26
Multiple autocomplete is deleting on submit all entries with the same label if one of it was previously deleted

mertsincan:
Steps to reproduce the Error on Multiple AutoComplete (Showcase):
1. Type 'a' and complete to 'Afterdark'
2. add another entry by typing 'a' and completing to 'Afterdark'
3. delete one entry
4. add another entry by typing 'b' and completing to 'Blitzer'
5. push the submit Button
Now instead of having two entries ('Afterdark' and 'Blitzer') the autocomplete just shows the entry 'Blitzer'.


cagataycivici:
@mertsincan Please proceed with fix. Patch looks good.

Issue #210, opened 2015-04-30
PositionType absolute option for Dialog Framework


claudineij:
Like issue #6 it will be great if we have the same functionality for Dialog Framework, for example:

``` java
options.put("positionType", "absolute")
```

arunvc1:
It would be nice to have top in options
eg top: '5px'

http://forum.primefaces.org/viewtopic.php?f=3&t=40364&p=127615&hilit=dialog+framework+top#p127615

tandraschko:
Duplicate of #495

PR #354, opened 2015-05-28
Add PositionType support for DialogFramework

sudheerj:
The issue was available at #210


cagataycivici:
Part in DialogNavigationHandler could have type checking instead of implicitly checking for a specific attribute name like positionType.

Issue #41, opened 2015-04-17
ColumnToggler compatibility with Scrollable DataTable


cagataycivici:
Add support for columnToggler in a scrollable datatable.

andrespontt:
I think this should be the expected behavior (using jquery datatables):

http://plnkr.co/edit/xQzMy1TVd8zmpOYZA2tf?p=preview

cagataycivici:
Added support for scrollable datatable, note that frozen columns + column toggler not yet supported.

PR #384, opened 2015-05-31
scrollable support for columnToggler

sudheerj:
The issue is available at #41


cagataycivici:
Will add after refactoring.

Issue #297, opened 2015-05-18
Modal dialog z-index problem when minimized


sudheerj:
Modal dialog cannot be maximized after once it is minimized.

PR #385, opened 2015-05-31
Fix for Modal dialog z-index problem when minimized

sudheerj:
The issue is available here #297


cagataycivici:
Moved it to the Dock function, thanks.

PR #400, opened 2015-06-04
Dropdown support for autocomplete in multiple mode

sudheerj:
The issue is available at #9


cagataycivici:
I'll check thanks with little refactor.

PR #416, opened 2015-06-10
Sticky resize based on window resize

sudheerj:


cagataycivici:
@sudheerj Is there an issue about this? Can you please create one if not.

PR #531, opened 2015-07-15
DataTable var evaluated as null for rowSelection ajax event listeners

sudheerj:


cagataycivici:
Although I like the patch, there are too many changes and will consider applying it if we get too many complaints about this.

Issue #48, opened 2015-04-18
DataExporter choice of the CSV Separator


cagataycivici:
We'd like to choose the CSV separator for the DataExporter component, for instance use semicolon or tab in place of comma.

Why ?
Because in no Anglo-Saxon countries (at least in France), Microsoft Excel uses semicolon as the default CSV file separator. As a result, a CSV document exported with the DataExporter component doesn't open properly (the user has to reopen the document in choosing the coma separator to have the columns rightly displayed).

We suggest to add an attribute "CsvSeparator" to the DataExporter component that will allow to choose the Csv separator character.

Quix0r:
When will this be included? I may need to change the "column" separator.

gulbinas:
+1

Kill3r444:
Could we please have an update on this topic ? This feature would be very useful to a lot of users. For now I  include a custom CSVExporter in my project, but this is a lot of work to check for change at each Primefaces update. Anyway, thanks for your work, I love your framework !

tandraschko:
You can sponsor the change via PrimeFaces PRO or do it by yourself and provide a PR!

cnsgithub:
I'm interested in providing a PR...

Kill3r444:
@cnsgithub : not a native english speaker, so not sure what you mean : do you want to provide a PR yourself, or do you want me to provide one ?

cnsgithub:
I'll provide a PR, stay tuned.

cnsgithub:
PR: https://github.com/primefaces/primefaces/pull/3822

Please see PR comment for details.

Kill3r444:
Thanks, but I'm a newbie here  : what do I have to do now ? may I test your code in any way ?

cnsgithub:
You may fork or clone this repository and build your own primefaces snapshot based on the master branch: https://github.com/primefaces/primefaces/wiki/Building-From-Source

Otherwise, you can wait until a member of this repository merges the PR and releases a new minor (available to the public) or bugfix release (PrimeFaces Pro/Elite).

Kill3r444:
I reviewed your changes and have a question : why do you use a StringBuilder object while I have a Writer object in my Primefaces source (6.2.5) ?

cnsgithub:
For questions regarding code changes, please comment in the PR at the appropriate lines.

Kill3r444:
This isn't a change you made, it is a difference between my PF version (downloaded from PF website) and the one on the repository

[...]

PR #541, opened 2015-07-19
#48 Allow to select the value separator for the CSVExporter

MonsieurV:


cagataycivici:
We need a way to extend the tag handler somehow because every exporter might have different options e.g. csv, xml, pdf font... So need a generic way to add these kind of options, maybe like p:exporter options="#{bean.options}" where options provide a class that implements ExporterOptions where CSVExporterOptions provide separator.

PR #559, opened 2015-07-24
groupId and artifactId itext change

jrd:
This PR was done by forking tag 4.0 (sha1 c3f2f46). This change can also be reported to other version as well.
itext dep changed its groupId/artifactId even for older versions so this change is required in order to keep PDF export working.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
     <groupId>org.primefaces</groupId>
     <artifactId>primefaces</artifactId>
     <packaging>jar</packaging>
-    <version>4.0</version>
+    <version>4.1</version>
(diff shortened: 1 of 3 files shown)


cagataycivici:
We can't refer to those versions due license issues also it is a good idea to add a pull request to a "tag".

calou:
We got the same issue on our project using Primefaces 4.0.  The dependency to iText has been renamed and the former packages of iText have been replaced on the maven central repository. Projects that uses Primefaces 4.0 won't build which, in our case, is an issue. I would hate to have to fork PF4 repository just to change the dependency and imports of iText, it doesn't feel right.

PR #587, opened 2015-07-30
Datatable celledit IE issue

sudheerj:


cagataycivici:
What is the issue here? Is there a ticket? @sudheerj

tandraschko:
@sudheerj Would you please provide a issue + description?

sudheerj:
The commit looks like a wrong one.I didn't remember the exact issue since it was created long time back.I will check it once and get back.

Issue #651, opened 2015-08-24
validateLongRange with a minimum of 0 does not work with client side validation


weissreto:
If I use `<f:valdidateLongRange minimum="0" maximum="10">` with client side validation then values from Long.MIN_VALUE .. 10 are accepted instead of 0 .. 10! 

If I configure `minimum="1"` or `minimum="-1"` then the validation correctly accepts only values from 1 .. 10 respective -1 .. 10

tandraschko:
Could you please try to change your pull requests just to use Integer instead of int? This wouldn't require the additional boolean var.

weissreto:
Hi Thomas

Do you mean something like this:

```
public class LongRangeValidator extends javax.faces.validator.LongRangeValidator implements ClientValidator {

    private Map<String,Object> metadata;
    private Long minimum;
    private Long maximum;

    public Map<String, Object> getMetadata() {
        metadata = new HashMap<String, Object>();

         if(minimum!=null)
            metadata.put(HTML.VALIDATION_METADATA.MIN_VALUE, minimum);

        if(maximum!=null)
            metadata.put(HTML.VALIDATION_METADATA.MAX_VALUE, maximum);

        return metadata;
    }

    public String getValidatorId() {
        return LongRangeValidator.VALIDATOR_ID;
    }

    public void setMaximum(long maximum) {
        super.setMaximum(maximum);
        this.maximum=maximum;
    }

    public void setMinimum(long minimum) {
        super.setMinimum(minimum);
         this.minimum=minimum;
    }
}
```

tandraschko:
yep, exactly :)
if this is working fine, i have no problems to accept the pull request!

cagataycivici:
Why can't we check against Double.MIN_VALUE and Double.MAX_VALUE to decide if values should go in metadata? Not sure about duplicating field and methods already available in parent class.

cagataycivici:
Ok, I see now. isMaximumSet() and isMinimumSet() are private. Wish they weren't. I actully prefer the booleans, less confusing for me since we need to check if these values are checked. Using Long, duplicates the field at parent class.

PR #652, opened 2015-08-24
Fix Issue #651

weissreto:
This changeset fixes issue #651. 

The old implementations only set the meta data if the value is not 0. So a value of 0 was not propagated to and considered by the client side validation. Moreover the value returned by getMinimum() is depending on the implementation of javax.faces.validator.LongRangeValidator. Some implmenations return a 0. Others returns Long.MIN_VALUE. 

The validators now set boolean flags if a minimum or maximum value are set. This flags are considered when building the meta data. The meta data is only set if also the minimum or maximum were set.


tandraschko:
I would use Integer instead of int, so you don't need the additional flag. WDYT @cagataycivici?

cagataycivici:
Thanks for the feedback.

PR #692, opened 2015-09-14
Update core.dialog.js

henriquemem:
add header option


cagataycivici:
Not clear what the change does.

PR #714, opened 2015-09-22
Client side bean validation message interpolator

LordEagle84:
Client side validation message must respect the message interpolator of
the ValidationFactory in charge. Otherwise the validation message shows the
template of the message instead of the interpolated value.

### DIFF
--- a/src/main/java/org/primefaces/validate/bean/BeanValidationMetadataMapper.java
+++ b/src/main/java/org/primefaces/validate/bean/BeanValidationMetadataMapper.java
@@ -24,9 +24,11 @@ import java.util.Map;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+
 import javax.el.PropertyNotFoundException;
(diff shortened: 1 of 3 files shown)


tandraschko:
Its currently by design that the messages should be managed on the client.
@cagataycivici If we allow the message resolving on the server, we should make it activateable and deactivate it per default because it affects the performance.
Everytime a new MessageInterpolatingConstraint will be instantiatet and the MessageInterpolator will be called.

tandraschko:
e.g. manipulate the resolved constraint metadata instead wrapping it is a better way as we don't need a new object instance every time.

cagataycivici:
What is the problem with current implementation and what does the pr solve?

tandraschko:
The best usecase: there are some applications which maintains translations in the database, which can be changed during runtime by admins.
I will take care of it for the future. I already created a issue about it.

tandraschko:
done via issue #778

Issue #500, opened 2015-07-06
SelectCheckboxMenu with SelectItemGroups


jrodalo:
It would be nice if SelectCheckboxMenu supports SelectItemGroup like SelectOneMenu does.

With the following code:

``` xml
<p:selectCheckboxMenu value="#{checkboxView.selectedCars}" label="Cars">
    <f:selectItems value="#{checkboxView.cars}" />
</p:selectCheckboxMenu>
```

and

``` java
//cars
SelectItemGroup g1 = new SelectItemGroup("German Cars");
g1.setSelectItems(new SelectItem[] {
  new SelectItem("BMW", "BMW"),
  new SelectItem("Mercedes", "Mercedes"),
  new SelectItem("Volkswagen", "Volkswagen"
});

SelectItemGroup g2 = new SelectItemGroup("American Cars");
g2.setSelectItems(new SelectItem[] {
  new SelectItem("Chrysler", "Chrysler"),
  new SelectItem("GM", "GM"),
  new SelectItem("Ford", "Ford")
});

cars = new ArrayList<SelectItem>();
cars.add(g1);
cars.add(g2);
```

I expect to see something like this:

![primefaces_2](https://cloud.githubusercontent.com/assets/1197514/8520074/273a92d2-23cf-11e5-97a9-5db8deb80573.png)

But Primefaces 5.2.x its rendering this:

![primefaces_1](https://cloud.githubusercontent.com/assets/1197514/8520086/3ee784bc-23cf-11e5-8d80-e1b91626a59c.png)

cagataycivici:
Sounds good to me :)

OskarsPakers:
As a workaround I have used standard `h:selectManyCheckbox` component and to applied prime-like styling using primeUI.

```
$("[name='myForm:myCheckboxId']").puicheckbox();
```

darwiesh:
is it soon to see the issue marked as closed ??

OskarsPakers:
I added a pull request for this bugfix, but still no update.
https://github.com/primefaces/primefaces/pull/804

tandraschko:
Mergen - Thanks for the PR!

drj123:
if this issue is resolved. where I can find the sample code

schmolleg:
Hi all, 
ich still have the same issue. I use primefaces 6.0 but the example code from above doesn't work. Does anyone have sample code. It's very urgent and i don't get the grouping working.

OskarsPakers:
[...]

PR #804, opened 2015-10-19
SelectItemGroup element support in SelectManyCheckbox

OskarsPakers:
In case of SelectItemGroup elements appearing in the list, pageDirection layout supports grouped items. Similar to feature described in #500 but for SelectManyCheckbox.


OskarsPakers:
Support for SelectItemGroup was discussed in the forum quite some time ago here:
http://forum.primefaces.org/viewtopic.php?f=3&t=24672&sid=17f9dd24523eebcfcfb36d19c307a41e

tandraschko:
I just did a very very short review of your changes.
Could it be that you miss a closing tr in the SelectManyCheckboxRenderer after line 264?

Would be great if you could also provide a PR for the showcase to demonstrate the feature. So we can also validate the changes better.

OskarsPakers:
There was indeed missed closing tr. I fixed it.
Created pull request for primefaces/showcase: https://github.com/primefaces/showcase/pull/12

tandraschko:
your PR now has more changes like styling/blanks etc.
Would be great if you could remove that, otherwise we can't see the real changes.

OskarsPakers:
Fixed all spaces/indents. Now those which are left are because of the new code or additional if clause.

spauny:
Any update on this? I tested and it seems to work properly.

tandraschko:
#500

tandraschko:
Merged! Thanks for the PR!

PR #820, opened 2015-10-27
Fix input focus on inputfields in treetable

nliechti:
Fix issue https://code.google.com/p/primefaces/issues/detail?id=7952


cagataycivici:
What is your PF version? I think this was fixed in Feb 10.

https://github.com/primefaces/primefaces/commit/0579ab29ca8e558722746fc0e1239de0cd9992b2

tandraschko:
works fine for me

PR #908, opened 2015-11-24
Check if Cell Editor is rendered before applying style class

firsthour:
Check if the p:cellEditor is rendered before applying the cell editor style class to a datatable cell. Right now it only checks if the cell editor is null, so if a p:cellEditor is rendered=false, you can still click and highlight the cell but it does nothing.


tandraschko:
Would you please create a ticket for it? I will apply it then ;)

firsthour:
Created this: https://github.com/primefaces/primefaces/issues/943

firsthour:
Thank you! I guess from now on I'll create an issue right away!

Issue #53, opened 2015-04-18
Advanced DataTable filtering fails with Dynamic Columns


cagataycivici:
Using the new filter mechanism in PF5 with Dynamic Columns does not filter.
When debugging (using a lazyDataModel) the filter map is always empty. 

``` xml
<p:dataTable value="#{abc01BackingBean.testList}" var="vo" widgetVar="testTable2">

            <p:columns value="#{abc01BackingBean.columnsList}" var="column"
                       filterBy="#{column.property}">
                <f:facet name="header">
                    #{column.header}
                </f:facet>
                <f:facet name="filter">
                    <h:inputText onchange="PF('testTable2').filter()"/>
                </f:facet>
                <h:outputText value="#{vo[column.property]}"/>
            </p:columns>
        </p:dataTable>
```

the86freak:
is there a known workaround for that issue?

mafaul:
See http://stackoverflow.com/questions/24181130/filter-a-lazy-datatable-with-dynamic-columns/27876091#27876091

daniel-armino:
Cagatay, one question about this error, do you know how version this ticket will be ok? I'm developing an application with datatable colums and dynamic filter on Primefaces 5.2 and this error remains.

mafaul:
Having this issue resolved would really be awesome! Keep up the good work.

kkoehler:
Hi,

we encountered the same problem. Why not integrating the proposed solution from stackoverflow?
I created a pull request for that:
https://github.com/kkoehler/primefaces/pull/1

Kristian

kkoehler:
https://github.com/primefaces/primefaces/pull/930

PR #930, opened 2015-12-02
Solution from stackoverflow for issue #53 - FilterFeature for Datatable

kkoehler:
Hi,

this is just the proposed solution from:

http://stackoverflow.com/questions/24181130/filter-a-lazy-datatable-with-dynamic-columns/27876091#27876091

https://github.com/primefaces/primefaces/issues/53

Thanks,
Kristian


cagataycivici:
This is a workaround only, it doesn't solve the actual problem of getting the local value from the filter facet. Also restricts that id of filter facet component must be filter. Solution would be finding out why;

``` java
filterValue = ((ValueHolder) filterFacet).getLocalValue();
```

return null all the time.

cagataycivici:
Please see #53

PR #1000, opened 2016-01-07
Correct square display of empty item + refresh filtered data on global select/unselect 

schlebe:
There is 2 corrections in this PullRequest.
1. Correct square display of empty item in ListBox display by a SelectManyCheckBox

The height of empty item(line) in ListBox is not correctly displayed !
The square of empty item is overriden by the square of next item.
For this correction, I fill the empty line with a &nb; character 
I have seen this solution in your code for another correction :-)
1. refresh filtered data on global select/unselect

when clicking on checkbox displayed in header of ListBox linked to a SelectManyCheckBox, the filtered data in DataTable are not refreshed.
But when I click on an item to select or unselect it, that work.
I think that the problem is that the filtering event is not propagated to dataTable.
To correct it, I have only changed 2 lines in adding a second parameter with TRUE value.
That's working on my PC.


schlebe:
I have some problem with GIT.
I have commited 2 sources for a new PullRequest but my first change are always there.
I will comment here my request. If this will not work, I will create a new PullRequest later.

My request

add a new attributes PROTECTED to p:selectBooleanCheckbox component.

This attribute prohibit input in CheckBox.

It is like disable=TRUE without changing the display's attributes.

I have only implemented this change on p:selectBooleanCheckbox but this new functionnality can be implemented on all components.

PR #1018, opened 2016-01-13
Animate flag should be used for content

CSchulz:
Content visibility changes should use the animate flag, too. Otherwise there are disturbing delays between the blocking div layer and the information shown to the user.


kukel:
What is the issue this patch is related to?

CSchulz:
@kukel Thanks for the hint, it seems like my git settings screwed up.

Issue #130, opened 2015-04-21
Readonly option for SelectBooleanCheckbox


cagataycivici:
Please add readonly option for SelectBooleanCheckbox component so it can be used e.g. in tables for showing boolean values graphically. Thank you very much.

mauromol:
Another use case for readonly vs disabled state for checkboxes (although in `selectManyCheckbox`, instead of `selectBooleanCheckBox`) is the following.
I have a list of elemens in a `selectManyCheckbox`. I want to prevent the user from changing the state of just some of these check boxes (ie.: some must be clickable, some must not).
However, if I use `itemDisabled="true"` for the values I want to make read-only, on submit PF will consider those values as "unchecked" and change my model by removing any corresponding elements from the underlying collection. This is due to the way HTML handles the "disabled" state for input fields (i.e.: they are not submitted - and for a single checkbox this is interpreted as "set to false").

mauromol:
Maybe the `selectManyCheckbox` case is worth opening a dedicated bug (I can't think of a proper use case for the current behaviour), what do you think?

cagataycivici:
https://stackoverflow.com/questions/155291/can-html-checkboxes-be-set-to-readonly

mauromol:
It's not clear to me whether this has been implemented or not (and in which version?).

melloware:
@mauromol I don't believe it was fixed because there were no commits associated with it.  I think what @cagataycivici is pointing out is that there is NO such thing as a ReadOnly checkbox.   That Stack Overflow article states as much and just points out a bunch of hacky workarounds to try and make it look readonly.

mauromol:
@cagataycivici can you confirm what @melloware says?

PR #1040, opened 2016-01-18
adding PROTECTEDx attribut to selectBooleanCheckbox

schlebe:
DISABLE attribute on selectBooleanCheckBox locks the widget but also changes the display of widget.

I need a solution more readable for user that locks(protects) the widget but doesn't change the display attributes.

To do that, I lock only the input field linked to this widget.

Attention: the real attribute is "protectedx" with x at end.
I cannot define an attribute "protected" because this name is defined by ENUM and protected is a key for java language.

I think that this problem can be resolved (not by me) in correcting maven-jsf application so that it work with "protected" name like it already work with "for" name in CheckBox.xml file.

### DIFF
--- a/src/main/java/org/primefaces/component/selectbooleancheckbox/SelectBooleanCheckboxRenderer.java
+++ b/src/main/java/org/primefaces/component/selectbooleancheckbox/SelectBooleanCheckboxRenderer.java
@@ -132,6 +132,19 @@ public class SelectBooleanCheckboxRenderer extends InputRenderer {
         styleClass = !checkbox.isValid() ? styleClass + " ui-state-error" : styleClass;
         styleClass = disabled ? styleClass + " ui-state-disabled" : styleClass;
 
+        //2016-01-14:BS: new attribute PROTECTED
+        
+        //<REMOVE>
+        //styleClass = disabled ? styleClass + " ui-state-disabled" : styleClass;
+        //</REMOVE>
+        
+        //<ADD>
+        if (disabled && !checkbox.isProtectedx())
+            {
+            styleClass += " ui-state-disabled";
+            }
+        //</ADD>
+
         String iconClass = checked ? HTML.CHECKBOX_CHECKED_ICON_CLASS : HTML.CHECKBOX_UNCHECKED_ICON_CLASS;
 
         writer.startElement("div", null);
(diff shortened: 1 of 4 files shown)


kukel:
'readonly' is a better name for this property, so it is in line with the `h:selectBooleanCheckbox`. But this should either be done to **all** components, or none.

schlebe:
Ok, readonly is a better name.
But I cannot to this for all component. 
Actually, I need it only for selectBooleanCheckBox

but you help me, because now, I know that for my issue, I can use h:selectBooleanCheckBox :-)

10 minutes later ...
I have changed my program to use h:selectBooleanCheckBox at place of p:selectBooleanCheckBox.

The problem is that the CheckBox is now ReadOnly, but I can change visually the CheckBox !!!

For your information, I work in HTML 5 under Chrome 47.0.2526.111m and the tag generated by JSF to represent h:selectBooleanCheckBox is

<input type="checkbox" name="j_idt179:0:j_idt207:0:PerformanceTable:21:j_idt243" checked="checked" readonly="readonly">

So, I will add a new functionality in PrimeFaces that already exists in "same" JSF component but the JSF implementation doesn't work. 

QUESTION

Because JSF doesn't work correctly, can this functionality be implemented in PrimeFaces ?

REMARK

I think that RESPONSIVE functionality has not been implemented for all components !
Why READONLY must be implemented for all components. I think this can be implemented slowly one component at a time.

What's next ?

kukel:
please edit your comment and format the 'xml'... it is invisible

kukel:
I mean at least for all real input components. (checkbox, radio, textinput etc...)

schlebe:
there is no XML in my comment so I don't understand how I can format it ?

Ok, I understand that this new attribute must "normally" be implemented for all input component.
The problem is that I have written a correction only for selectBooleanCheckbox and actually, I don't need more and I hope that my correction is pushed in next PrimeFaces release.

Personnaly, I find a best solution to implement my ChangeRequest only for selectBooleanCheckbox and wait that somebody else (or perhaps me) implements ReadOnly attribute on all other input components.

kukel:
you have html tags in your comment I think. Between: _"For your information, I work in HTML 5 under Chrome 47.0.2526.111m and the tag generated by JSF to represent h:selectBooleanCheckBox is"_  and _"So, I will add a new functionality..."_

tandraschko:
Please create a forum post which describes your problem (maybe also with pictures), then create a issue with a link to the post.
Afterwards we can discuss about the name and if we will do it for all components.
The current PR can't will not be applied but thanks for your effort till now.

schlebe:
This functionnality has already been implemented by **cagataycivici**.
Thanks to him
[https://github.com/primefaces/primefaces/issues/130]

schlebe:
Sorry, this functionnality that I have implemented in this PullRequest has been noted as "New Feature" by cagataycivici but is not actually developped by him !
I hope only that he will read my post to take some time to read my PullRequest and implement it in new Release of PrimeFaces !

PR #1043, opened 2016-01-19
Row hover for radio and checkbox datatable selection

sudheerj:


tandraschko:
will apply it by myself as it conflicts

PR #1083, opened 2016-01-29
Delete item with backspace in AutoComplete multiple mode

sudheerj:


tandraschko:
Already fixed in trunk.

PR #1087, opened 2016-01-31
OneButton - allow deselection

sudheerj:


kukel:
Then please do the same for radio...

tandraschko:
The PR has a conflict, could you please check that?

sudheerj:
Ronald,I will create separate PR for radio.

Yes Thomas I will check that conflicts.

tandraschko:
merged it

PR #1108, opened 2016-02-04
Change in DataTable Lazy

fbrissi:
Paginate in DataTable Lazy with keys RIGHT and LEFT.


kukel:
does this block scrolling with the keys in a scrollable datatable?

fbrissi:
I do not know if I understand What you said. In my case, most users prefer to be able to navigate pages than the scroll, but I did some test to see how it would look with the scroll, and while the focus is on the line is paging, to switch the focus to the scroll was the scroll now and use the keys.

If that's not what you meant, please rephrase your question so I can understand better.

Sorry but I'm poor with English.

kukel:
Yes, that is kind of what I meant... if left-right scrolling is possible, it should do that first with the keys in my opinion (backwards compatibility) and only if the right boundary is reached and the key is still pressed, after a delay it should paginate. Same for left key... Users should **not** need to focus the scrollbar first if scrolling is possible.

fbrissi:
Great, I liked the suggestion, I will make this change.
Thank you for watching.

fbrissi:
I'll close this pull request and open again when finished kukel suggestion.
Thank you for your suggestion kukel.

PR #1121, opened 2016-02-09
fix #1058

### LINKED ISSUE #1058: Disable effect queuing
Hi,

In my application there are several input controls connected to a button via `p:effect for="btn"`. If an input value changes, the button will pulsate. However, if one changes multiple input values one after the other, the effects will be queued and executed consecutively, resulting in very long pulsation.

Here's an old but still valid explanation that also shows how to overcome this issue: http://www.learningjquery.com/2009/01/quick-tip-prevent-animation-queue-buildup

The solution might be to call the jQuery `stop( [clearQueue ] [, jumpToEnd ] )` function, see https://api.jquery.com/stop/.

As a workaround I've overridden PrimeFaces' `EffectRenderer` by modifying its `EffectBuilder` as follows:

```
public EffectBuilder(String type, String id) {
    buffer = new StringBuffer();
    buffer.append("$(PrimeFaces.escapeClientId('");
    buffer.append(id);
    buffer.append("'))");
    buffer.append(".stop(true, true)"); // line added
    buffer.append(".effect('");
    buffer.append(type);
    buffer.append("',{");
}
```

I would suggest to add an extra parameter to `p:effect` to better control behavior.

Kind regards

cnsgithub:


tandraschko:
Please modify your PR that you don't change identation or line breaks of existing code.

cnsgithub:
Done, please verify.

tandraschko:
Merged! Thanks for the contribution!

PR #1187, opened 2016-03-01
FIx cross-platfom keyboard events

deathbeam:
This fixes cross-platform keyboard event for example for inputText with mask, where for example arrow keys do not works on Firefox and works on Chrome (because FIrefox uses charCode and not keyCode). So I used JQuery's event,which, what is cross-platform. Tested it on enterprise application on what we are working. Actually, this thing was causing many bugs in our application, so I would really like to see this officially fixed.


ghost:
+1

kukel:
Since there is no 'voting' like in stackoverflow, a +1 from me to

xKopax:
+1

dinhani:
+1

ivolucas:
+1

tandraschko:
should be fixed by #1334, please verify.

PR #1191, opened 2016-03-01
New Exel exporter that uses the Apache POI streaming api. 

ivolucas:
New exel file exporter that uses the Apache POI streaming api. 
Requires v3.8 or upper of Apache POI
#1192


tandraschko:
Please create a issue as we don't accept PR's without a issue.
Please also state the benefits and why it's required.

tandraschko:
Applied your PR - thanks for the contribution!

As we don't maintain @author, i removed it.
Probably we need to find a way in the future to list the contributors :)

PR #1200, opened 2016-03-03
Show tooltip with error message instead of title using p:message

### LINKED ISSUE #1198: Message: tooltip doesn't work correctly if title is used
Consider following code, if you moving your cursor over the input it shows a tooltip with "An input".

```
<p:tooltip />
<p:inputText id="input" value="..." title="An input" />
<p:message for="input" display="tooltip" />
```

Add a validation to the input field and try it with wrong values. I expect to see the error message in the tooltip, but it keeps the title value. Removing the title attribute it works like expected.

input

CSchulz:
Set data to specify that there is a tooltip with an error message and use it instead of the title attribute

Fixes #1198 

There is another possible solution https://github.com/CSchulz/primefaces/tree/write-error-to-title-attribute


tandraschko:
Thanks for the PR but will close it for. I commented your issue and after you answered, i can still check if we can solve this with this PR.

PR #1211, opened 2016-03-07
Fix doRedirect for IE and onbeforeunload alerts

msgeissler:
IE behaves very strange when a redirect is triggered and an onbeforeunload function is defined to show an alert (e.g. window.onbeforeunload = function () {return 'You will lose your changes...'};).
If the user declines the redirect and stays on the current page, IE's window.location will throw an error that will kill the JS execution for that page.

Secondly, for whatever reason, IE will try to unload the current page two times when a user declines the redirect and window.location is used (two alerts are shown from window.onbeforeunload). After that it will throw the error mentioned above. Using window.location.assign() will avoid the multiple attempts at unloading the page.

Tested in IE 11 and Chrome 48.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
+++ b/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
@@ -702,7 +702,11 @@ PrimeFaces.ajax = {
     ResponseProcessor: {
 
         doRedirect : function(node) {
-            window.location = node.getAttribute('url');
+            try {
+        		window.location.assign(node.getAttribute('url'));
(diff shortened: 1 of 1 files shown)


tandraschko:
Please create a issue for it, too.

why do you use assign instead =?
And also please use PrimeFaces.warn instead console.warn.

msgeissler:
I changed the code and added an issue with a description and a short instruction to reproduce this error, as it cannot be reproduced using jsfiddle and others.

I hope the description is understandable.

tandraschko:
Merged, thanks for the PR!

PR #1227, opened 2016-03-12
fixed #690

### LINKED ISSUE #690: mobile SelectBooleanCheckbox fires onchange event twice per toggle.
The mobile renderer uses a label for the itemLabel, and the "for" attribute is set to the input id, so clicking the label automatically triggers the input to toggle states.  The javascript is also triggering a change on the element, which results in ajax requests being submitting twice.  It also produces unexpected results if the ajax request rerenders a parent of the checkbox.

goking:
An checkbox input element will fire 'change' event by itself, so it not need to trigger 'change' event in the  toggle method. fixed #690


tandraschko:
Looks fine, thanks for the PR!

Issue #1220, opened 2016-03-09
Tree always visits every node with wrong context


bodabela:
You may consider this patch to provide the proper node type implementation.

I had a problem in case of the rendered attribute relayed on an attribute that was present only in the data bean for the specific node type:

```
<p:treeNode id="model-tree-#{node.id}" type="component">
    <h:outputText value="#{node.name}" />
    <h:outputText value=" - ##{node.id}" rendered="#{node.id!=null}" />
    <h:inputHidden value="#{node.id}" rendered="#{node.id!=null}" />
</p:treeNode>
<p:treeNode type="componentAttribute">
    <h:outputText value="#{node.name} : #{node.value}" />
</p:treeNode>

```

[0001-nodeType-support-for-treeNode.zip](https://github.com/primefaces/primefaces/files/166010/0001-nodeType-support-for-treeNode.zip)

Cheers

bodabela:
Of corse the problematic attribute is the "id" in this case. The componentAttribute data bean has no "id" attribute.

cagataycivici:
Send a pull request please instead of a zip.

tandraschko:
I can't see how your PR is related to fix this issue.
Please provide more descirption and a complete example (xhtml + bean).

bodabela:
It is not visiting the  child tree unless it has the matching nodeType.
If you visit a certain node content that has an EL content that is not matching the actual bean instance (e.g. has no attribute) you find the EL expression fail.

PR #1245, opened 2016-03-17
#1220 treeNode's type attribute

bodabela:
I consider during the visit if the type is the required one of the treeNode. If it is missing it might cause a problem in case of different bean attributes of the data object.


mkomko:
Can you please merge this pull request (after testing) so that it is available in the next Elite release? The underlying bug is making our life difficult. Thank you!

tandraschko:
See my comment in the issue.

PR #1252, opened 2016-03-18
Add CellEdit and RowEdit support to TreeTable

mertsincan:


mertsincan:
Added for 6.0 and 5.3.10.

PR #1264, opened 2016-03-25
Add onTabClose callback

ghetolay:
Adds a onTabClose callback to be called as soon as a tab as finished to close (after animation). For client side use and most likely ui related stuff.

Using this personally, thought it won't hurt to add it directly into Primefaces since it has really low impact.

### DIFF
--- a/src/main/java/org/primefaces/component/accordionpanel/AccordionPanelRenderer.java
+++ b/src/main/java/org/primefaces/component/accordionpanel/AccordionPanelRenderer.java
@@ -120,7 +120,8 @@ public class AccordionPanelRenderer extends CoreRenderer {
         
         wb.attr("multiple", multiple, false)
         .callback("onTabChange", "function(panel)", acco.getOnTabChange())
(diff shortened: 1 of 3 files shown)


tandraschko:
The patch looks very incomplete as the "onTabClose" is not defined in the component class.
Also please create a issue - always for each PR.

ghetolay:
Sorry did it some time ago and had to extract it from my fork also since this is working for me for some time and there was no compilation errors I didn't double check it. 
Now I've merged the rest of it and checked it against a showcase fork. Work for both single and multiple selection.

Also on the js part I think it would be good, in order to avoid code duplication, to at least create a `_renderClose()` function called by `show()` and `hide()`. But since it may be better way to do it (call `hide()` on the previous tab selected during `show()`) and it's not directly related to the onTabClose feature I didn't do it.

ghetolay:
Related issue created [here](https://github.com/primefaces/primefaces/issues/1271).

PR #1285, opened 2016-03-31
Enable autocomplete highlighting if custom content is used

joergmattes:
Autocomplete highlighting doesn't work if custom content (<column ...>)
is used. With this patch a developer can choose to add highlighting to a
certain column by adding the ui-autocomplete-item styleClass.


tandraschko:
Thanks for the PR but creating a PR without a issue and detailed description + example hard to verify for us and therefore invalid.
If the issue is still important for you, please create a issue and will check that. Please also mention this PR in the issue, maybe i can just merge it ;)

Issue #1303, opened 2016-04-07
function findRootWindow inside iframe does infinite loop


alberto-marin-garcia:
Testing primefaces 5.3.9 / 5.3.10 in JBoss EAP 6.4.5

Bug opening dialogs inside iframes, using this context using frames and iframes:

```
<html>
<head></head>
<frameset rows="*" >
<frame id="mainFrame" >
<iframe id="primefacesFrame" width="100%" height="100%">
<!-- iframe with primefaces content-->
</iframe>
<frame >
<noframes>
</noframes>
</frameset>
</html>
```

inside iframe with id="primefacesFrame" runs a primefaces app , but when opening modal dialogs the function findRootWindow begins an infinite loop :

function PrimeFaces.dialog.DialogHandler.findRootWindow
var a = window;
while (a.frameElement)
a = window.parent;
}
return a

in first iteration the object a.frameElement points to primefacesFrame frame,
at second iteration a.frameElement points points to frame with id "mainFrame",
at 3rd iteration a.frameElement points to frame with id "mainFrame",
at 4th iteration a.frameElement points to frame with id "mainFrame", ...and so on

tandraschko:
Would be great if you could provide a pull request.

mertsincan:
Can you please attach a sample code for us to replicate?

mertsincan:
Next time, please use https://help.github.com/articles/basic-writing-and-formatting-syntax/ for your codes. I could see your test code now. I'll review and get back to you.

mertsincan:
I didn't see a bug on console of browser. Could you please attach sample codes for us?

alberto-marin-garcia:
Thank you mertsincan.

My HTML example:

```
<html lang="fr">
 <head>
  <title></title>

 </head>
 <frameset id="frameset" rows="*" border="0" onLoad="window.top.menu.addGrowlInfoMessage('',msgHello);">
  <frame  id="menu"  name="menu" src="/menu.do" scrolling="no" noresize>

    <!--begin  menu.do -->
        <div id="capaPrincipal">
            <iframe name="principal" id="principal" src="/fondo.do" scrolling="no" width="100%" frameborder="0" marginheight="0" marginwidth="0" style="border-top: medium solid #FDC153;" >
            </iframe>
        </div>

    <div id="capaPrincipalEE6" style="position:absolute;top:0;left:0;width:100%;z-index:4;background:white;visibility:hidden">
[...]

PR #1327, opened 2016-04-20
findRootWindow ifinite loop

Julienng:
this should be `w` instead of `window` into the while to avoid infinite
loop. #1303


cagataycivici:
@mertsincan Please review.

@Julienng Thank you.

Julienng:
You're welcome

tandraschko:
Merged, looks fine for me.

PR #1331, opened 2016-04-23
p:captcha callback support

### LINKED ISSUE #1241: p:captcha callback support
reCaptcha v2 provides two callback-apis.

First:

> callback
> Optional. The name of your callback function to be executed when the user submits a successful CAPTCHA response. The user's response, g-recaptcha-response, will be the input for your callback function.

Second:

> expired-callback
> The name of your callback function to be executed when the recaptcha response expires and the user needs to solve a new CAPTCHA.

Both are not supported with p:captcha.

gpoluch:
Fixes #1241 callback and callback-expired from reCaptcha v2 were not supported by p:captcha.
With this patch captcha provide two attributes for javascript callbacks:
- callback (executed when the user submits a successful captcha response)
- expired (executed when the recaptcha response expires and the user needs to solve a new captcha).

### DIFF
--- a/src/main/java/org/primefaces/component/captcha/CaptchaRenderer.java
+++ b/src/main/java/org/primefaces/component/captcha/CaptchaRenderer.java
@@ -76,9 +76,7 @@ public class CaptchaRenderer extends CoreRenderer {
         wb.attr("sitekey", publicKey)
             .attr("theme", captcha.getTheme(), "light")
             .attr("language", captcha.getLanguage(), "en")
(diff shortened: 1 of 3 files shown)


tandraschko:
I would apply your PR if it wouldn't destroy our formatting and i could see only the exact changes.

gpoluch:
Formatting corrected at 218fb2e.

tandraschko:
Merged - thanks for the PR!

PR #1419, opened 2016-05-15
Empty parameter value rendering

aleskurlovich:
Hello. I found problem with empty f:param value rendering. For example, next code will initiate HTTP request with additional form data.

Code:
`<p:commandLink...>
   <f:param name="param" value="#{null}"/>
  </p:commandLink>`

Request data:
`param: null`

But I guess "param" value should be empty at least. I updated AjaxRequestBuilder class for solving it.

Referenced issue: #1424


tandraschko:
Please also provide a issue for tracking.

PR #1461, opened 2016-05-30
Implement containment property into DiagramModel

### LINKED ISSUE #1460: Diagram: Implement model attribute for draggable containment
For the possibility to implement scrollable diagrams I need the configuration for containment.

Look [here](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/diagram/diagram.js#L40)

NoFr1ends:
Fix #1460


tandraschko:
Thanks for the PR!

Issue #1297, opened 2016-04-05
Avoid label "Choose" when we don't define the attribute for p:fileUpload


JaviRpo:
Don't put default label "Choose" when we don't define the attribute for p:fileUpload

If i put the attribute without value, the draw button doesn't have height

![image](https://cloud.githubusercontent.com/assets/8562436/14293746/7e73332e-fb33-11e5-8cf2-a4f5123ddc45.png)

In the imagen there are a button (left) with a fileUpload (right)

tandraschko:
Hmm, what do you mean?

JaviRpo:
p:fileUpload is similar to p:commandButton ( visually ), the difference is in functionality.

In this case, i want to put a p:fileUpload only with a icon without label and I hope that the visual appearance of the 2 components are the same.

So, when i try to put the p:fileUpload component without the label attribute the component draws with a default label "Choose"
![image](https://cloud.githubusercontent.com/assets/8562436/14302217/33e87720-fb64-11e5-9145-391c3f562bba.png)
![image](https://cloud.githubusercontent.com/assets/8562436/14302220/38ee8d90-fb64-11e5-8bd8-c88902a03771.png)
(this is a comparation between a p:commandButton (left) and p:fileUpload (right)

and when i put the attribute label with empty value, the component p:fileUpload draws smaller than p:commandButton
![image](https://cloud.githubusercontent.com/assets/8562436/14302261/99c48584-fb64-11e5-842f-cb5041938ee1.png)
![image](https://cloud.githubusercontent.com/assets/8562436/14302246/7afd3844-fb64-11e5-8fa5-2dd184d212cf.png)

it the same when i use the component in advanced mode for the 3 buttons
![image](https://cloud.githubusercontent.com/assets/8562436/14307340/d2b76c38-fb93-11e5-9f9b-f79b790fe459.png)

sudheerj:
We should have default labels for fileupload component.So if label,uploadLabel,cancelLabel are empty then we should add icon only setting similar to button components.

Rapster:
PR is here #1471

subzro:
There is still a small bug in the choose button.
No matter if the label is null or set by custom value, the upload component will always look like this:
![pf-upload-label-bug](https://cloud.githubusercontent.com/assets/20240792/16518573/e0c1678c-3f83-11e6-870f-696b12fd4c00.png)

[...]

PR #1471, opened 2016-05-31
Fix #1459 #1473 #1297

### LINKED ISSUE #1459: icon attribute for p:menuButton
it would be good if the component p:menuButton had the icon attribute, so we can change the default icon and create it without label, only with icon

Rapster:
And some formatting (replace tab by 4 spaces)


tandraschko:
Please provide different PR for formatting and logic changes. Your real changes are almost invisible ;)

Rapster:
My formatting is fine actually (it respects the coding style set by Primefaces anyway), it's the previous one which was a mess though (a mix of tab and spaces...). My advise would be to use a little trick I know to ignore whitespaces so you can see real changes: add **?w=1** to the url where you want review changed files (in this case https://github.com/primefaces/primefaces/pull/1471/files?w=1)

Rapster:
I don't understand why each commit I do are in this PR, as I want to create a PR for each commit :/

michael-las:
@Rapster: PRs work on branches. That way you can discuss the PR, make changes and commit those changes to the same branch/PR. If you create a branch for each fix, then you'll be able to create a PR for each branch/fix.

PR #1475, opened 2016-05-31
Prefixed totalRecords by clientId

### LINKED ISSUE #1463: paginator issue when updating lazy data table on filter event
In PrimeFaces 5.1/5.3:

``` html
  <h:form id="form">
    <p:dataTable var="item" value="#{dtFilterView.cars}"
                 filteredValue="#{dtFilterView.filteredCars}"
                 paginator="true" rows="5" id="otherTable">

        <p:ajax event="filter" update=":form:table" />

        <p:column headerText="Id">
            <h:outputText value="#{item.id}" />
        </p:column> 
        <p:column headerText="Brand" filterBy="#{item.brand}" >
            <h:outputText value="#{item.brand}" />
        </p:column>
    </p:dataTable>

    <p:dataTable var="car" value="#{dtLazyView.lazyModel}" 
                 lazy="true" rows="5" id="table" paginator="true"
                 paginatorPosition="bottom">
        <p:column filterBy="#{car.id}" headerText="Id" filterMatchMode="contains">
        </p:column> 
        <p:column filterBy="#{car.brand}" headerText="Brand" filterMatchMode="contains">
        </p:column>
    </p:dataTable>
  </h:form>
```
# Steps to reproduce
1. Open page 
2. Filter 'otherTable' by brand
3. Paginator of 'otherTable' is identical to the paginator of 'table'

In this example it seems that the paginator just didn't update. If one uses different models, one can proof that statement 3 is correct. [UPDATE: One can proof it also by first setting a filter on 'table' and afterwards filter 'otherTable']

I guess, this has to do with the callback parameter 'totalRecords'.

michael-las:
This should fix #1463

### DIFF
--- a/src/main/java-templates/org/primefaces/component/datagrid/DataGridTemplate.java
+++ b/src/main/java-templates/org/primefaces/component/datagrid/DataGridTemplate.java
@@ -63,9 +63,11 @@ import javax.faces.event.BehaviorEvent;
             //Update paginator for callback
             if(this.isPaginator()) {
                 RequestContext requestContext = RequestContext.getCurrentInstance();
+                FacesContext context = getFacesContext();
+                String clientId = this.getClientId(context);
 
                 if(requestContext != null) {
-                    requestContext.addCallbackParam("totalRecords", lazyModel.getRowCount());
+                    requestContext.addCallbackParam(clientId + "_totalRecords", lazyModel.getRowCount());
                 }
             }
         }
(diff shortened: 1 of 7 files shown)


tandraschko:
Thanks for the PR but applied another fix. Nevertheless, prefixing our internal callbackParams probably makes sense in the future.

michael-las:
Thanks for fixing. I'm trying to understand what's happening: the updated component including the paginator will be rendered as part of the partial DOM while the paginator of the source component will be updated per callback? That is, my PR fixes the symptom but not the problem, right?

tandraschko:
Right
We have actually 3 cases when the DataTable renders something
1) non-ajax
    Renders everything - complete DOM and script
2) updated via ajax
    same as 1)
3) internal ajax call and partial update of the datatable component (e.g. filter)
    renders only the rows and the datatable widgets replaces the rows manually
    additionally infos like totalRecors are passed to update the paginator - this needs only be done for the current datatable which triggered the request, not for every datatable on the view

michael-las:
Thank you. Can I delete the branch?

tandraschko:
yup

Issue #1381, opened 2016-05-03
Datatable reset selection state after click outside of Browser


frkmed:
After selected a row in the datatable and when i click outside of the browser or change the tab of browser the selected row reseted to the first row : you can see this if you navigate to http://www.primefaces.org/showcase/ui/data/datatable/selection.xhtml  and go to secend example "Single with Row Click" and select the third row than click outside of Browser and return you will see that the row was reseted. thank's.

tandraschko:
Works fine for me in FF.

frkmed:
not working for me in FF

tandraschko:
@cagataycivici @mertsincan  does it work for you?

frkmed:
my problem is like @markusg80  with issue #1230

kukel:
Works for me in FireFox (45.0.2)

markusg80:
It does not work also in Firefox 46.0.1. When u follow the description of @frkmed above inside the Showcase, after you click again inside the "Single with Row" example you will notice that the first row changes the color. If you take this example with more rows, scrolling enabled, selection event and you scroll down and select a row at the end and click outside browser and again in the table then the scrollposition jumps to top and first row change color. Go and modify the showcase example like like in my example in issue #1230. If you not try it you'll not find the bug. Good Luck

cagataycivici:
Works for me on FF 46.0.1.

frkmed:
I have tested this issue in many browser and also in  FF 46.0.1. but the probleme persist please can you give me a solution.

markusg80:
Problem persist for me too. Should i create a example project for you?

mtbadi39:
Works for me without problems on : 
FF 46.0.1
Chrome 50.0.2661.102

markusg80:
Steps to reproduce better:

Take the Single with Row Click example and modify it to load more cars (for example 100) and enable scrolling. Scroll down so long the first visible rows disappears and then select one random row. After this click on Taskbar and then click back to datatable. then the table scrolls to top and first row get hover color. And now another row is selected because selection event is fired after scroll to top is complete. I can add a screencast

Issue #1230, opened 2016-03-14
Datatable reset Scroll state after click outside of Browser


markusg80:
Steps to reproduce the issue:

Create scrollable datatable with selection enabled and add ajax event for rowSeclect. Scroll down to position outside of initial viewport (for example in the middle). Select a row. Take a click outside of the browser and again click inside the table.

Result:
The datatable change scrollposition to top and first row get hover color. the selected row keep selected but is outside of viewport.

Same effect is also on the Showcase

kukel:
what showcase example? I don't see an example with this combination

markusg80:
Ok i correct my statement. You can modify the select events showcase example with adding scrollable="true" scrollHeight="150" and then look to create enough cars (for example 100).

```
<p:dataTable id="singleDT" var="car" value="#{dtSelectionView.cars2}" selectionMode="single"  selection="#{dtSelectionView.selectedCar}" rowKey="#{car.id}"
  scrollable="true" scrollHeight="150">
    <f:facet name="header">
        Single with Row Click
    </f:facet>
    <p:column headerText="Id">
        <h:outputText value="#{car.id}" />
    </p:column>
    <p:column headerText="Year">
        <h:outputText value="#{car.year}" />
    </p:column>
    <p:column headerText="Brand">
        <h:outputText value="#{car.brand}" />
    </p:column>
    <p:column headerText="Color">
        <h:outputText value="#{car.color}" />
    </p:column>
    <f:facet name="footer">
        <p:commandButton process="singleDT" update=":form:carDetail" icon="ui-icon-search" value="View" oncomplete="PF('carDialog').show()" />
    </f:facet>
</p:dataTable>
```

frkmed:
Hello markusg80, any solution for that issue

markusg80:
At the moment i doesn't see a solution because until now i not do debuging inside primefaces. Maybe when i find time i can try to find the reason of the issue

michael-las:
provided PR #1512

PR #1512, opened 2016-06-09
scrollable table now  also scrolls to selected row when focused.

michael-las:
Fix for #1230 and #1381.

Now, when the table body gets the focus, it will scroll to the selected row. If no row is selected, it will scroll to the first selectable row (which is the current behavior).

Test, please.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
+++ b/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
@@ -441,7 +441,11 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
         this.getFocusableTbody().on('focus', function(e) {
             //ignore mouse click on row
             if(!$this.mousedownOnRow) {
-                $this.focusedRow = $this.tbody.children('tr.ui-widget-content.ui-datatable-selectable').eq(0);
+                $this.focusedRow = $this.tbody.children('tr.ui-widget-content.ui-datatable-selectable.ui-state-highlight').eq(0);
+                if ($this.focusedRow.length == 0) {
+                    $this.focusedRow = $this.tbody.children('tr.ui-widget-content.ui-datatable-selectable').eq(0);
+                }
+
                 $this.highlightFocusedRow();
                 
                 if($this.cfg.scrollable) {


markusg80:
For me it sound plausible but how can i test it? I try to implement the new database.js in my primefaces.jar but it not works. I also cleared the cache of netbeans and glassfish

michael-las:
The best way would be to check out the branch using git and build primefaces from scratch. During the build process many js-files will be aggregated into a single components.js which will be served to the browser. 

So, if you don't want to build PF on your own, you could try to edit components.js, find the appropriate lines and replace it by the code committed.

markusg80:
In the jar i've not found a file components.js. So at he moment i cant test this. Sorry for that i not know PF internals

michael-las:
Sorry, this fix is for PF 6.0 where components.js is located under META-INF/resources/primefaces. BTW: I don't know the internals of PF, either. I'm just trying to help by fixing some, mostly simple, primefaces bugs.

markusg80:
But you've fixed a Problem for that Primefaces says it doesn't exist. And thank you for that! I hope this fix will be implementet in 5.3 branch too

michael-las:
This fix can't be implemented in 5.3 since the code is really different between the two versions.

The behavior in 6.0 can be explained, because whenever the table body gets the focus, the patched function will be executed. So, all you need to do is to switch to another window and switch back to the browser (or, if you have more components on the page, you can navigate to another component and back to the table using tab/shift-tab).

In 5.3 I don't see code related to this. Therefore, I just tried it [edit: I tried an unpatched version of 5.3 and was not able to reproduce the bug]: works for me. Try to create a minimal example that reproduces the error and create a new issue, if neccessary. Explain the steps to reproduce the bug in every detail, including the OS, its version, browser, its version and so on.

markusg80:
I will make a test projekt! But for the moment i dont find the time! So please wait its coming soon..,

markusg80:
But will this fix be implementet in 6.0? At the moment its not

michael-las:
No, the pull request is not merged up to now. If you don't want to build PF on your own, you'll need to patch META-INF/resources/primefaces/components.js in primefaces-6.0.jar (also see my answer 8 days ago).

markusg80:
For wich version this fix will be abailsble? 6.0? 5.3?

PR #1520, opened 2016-06-13
Fix #1518

### LINKED ISSUE #1518: p:autocomplete can delete item when this is disabled
I have an autocomplete component with attribute multiple="true", so when the component is disabled="true", the inside items can be deleted with X icon

Rapster:
Eclipse removed some dummy spaces. Add **?w=1** to see real changes


tandraschko:
Could you please check your PR? it has more much changes

Rapster:
Arf I pushed something else on my branch since then, and the delta between this commit and my last one is quite big yes...

PR #1571, opened 2016-07-03
Fix #1297

### LINKED ISSUE #1297: Avoid label "Choose" when we don't define the attribute for p:fileUpload

Rapster:
Fix #1297 and better format, plus add some conventionnal brackets. Add **?w=1** to see real changes


tandraschko:
Please fix your PR to not change our formattings (new line / { })

Rapster:
Done

Rapster:
I think this one can be easily merged

tandraschko:
Will check it before 6.1 release ;)

tandraschko:
Seems to be fixed by mert. Will close it.

PR #1582, opened 2016-07-05
Fix #43

### LINKED ISSUE #43: RowExpansion problem with a Lazy Datatable
A datatable getting its data from a lazydatamodel. The table also contains a rowexpander. This is how it looks:

``` xml
<p:datatable var="productgroup" value="#{productbean.productgroups}">
    <p:column>
        <p:rowToggler/>
    </p:column>
    <p:column>
        <h:outputText value="#{productgroup.name}"/>
    </p:column>
    <p:rowExpansion>
        <p:datatable var="product" value="#{productgroup.products}">
            <p:column>
                <h:outputText value="#{product.name}"/>
            </p:column>
            <p:column>
                <h:outputText value="#{product.description}"/>
            </p:column>      
        </p:datatable>
    </p:rowExpansion>
</p:datatable>
```

The EL expression #{productbean.productgroups} returns a LazyDataModel.

Initially the datatable is displayed OK. However when I expand a row, JSF reports that it cannot evaluate the #{productgroup.products} expression (the expression on the second datatable).

brainlag:
Fix #43.

### DIFF
--- a/src/main/java/org/primefaces/component/datatable/feature/RowExpandFeature.java
+++ b/src/main/java/org/primefaces/component/datatable/feature/RowExpandFeature.java
@@ -32,7 +32,9 @@ public class RowExpandFeature implements DataTableFeature {
     public void encode(FacesContext context, DataTableRenderer renderer, DataTable table) throws IOException {
         Map<String,String> params = context.getExternalContext().getRequestParameterMap();
         int expandedRowIndex = Integer.parseInt(params.get(table.getClientId(context) + "_expandedRowIndex"));
-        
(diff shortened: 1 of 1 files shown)


tandraschko:
Couldn't it be that lazy data is already loaded when RowExpandFeature is encoded?
That are very sensible places and needs much debugging and reviews.

brainlag:
Yes that is possible. First I thought DataTable.loadLazyData() only loads if it is not loaded yet but after looking at it again it is clear that this is not the case. I will add another check to be sure the LazyDataModel is not loaded yet.

PR #1616, opened 2016-07-21
quill editor

husnutapan:


tandraschko:
Sorry but we already have 2 editors in the primefaces universe
1) cleditor in PF
2) ckeditor in PFE
i think we will add another but this probably needs discussions
@cagataycivici  WDYT?

cagataycivici:
Our current editor is old and we need something modern, I guess we can just replace cleditor instead?

tandraschko:
we should yes, but i don't think that ckeditor or quill will suite or license terms
ckeditor would be the better choice

tandraschko:
quill is also only compatible with IE11+
i think the best choice is really CKEditor but we can't include it directly in PF as it's GPL/MPL/LPGL which is not compatible with apache2.
Our resources jar for ckeditor is currently also not released under apache2.

tandraschko:
Will close it for now, we will discuss it internally as quill can't be used (license doesn't match and we have customers with < IE11)

Issue #1654, opened 2016-08-01
Dynamic ui:include + UIData (with rowStatePreserved) rendering problem


slovi:
There is a situation where UIData with rowStatePreserved is not rendered. NullPointer exception is thrown in this case.

The example of such situation is:
- we have index.xhtml with ui:include with src attribute as expression

```
        <h:form>
            <h:panelGrid columns="1" bodyrows="3">
                <p:commandLink action="#{mainFrameBean.setMenu('iframe1.xhtml')}" process="@this" update=":mainframe">Frame 1</p:commandLink>
                <p:commandLink action="#{mainFrameBean.setMenu('iframe2.xhtml')}" process="@this" update=":mainframe">Frame 2</p:commandLink>
                <p:commandLink action="#{mainFrameBean.setMenu('iframe3.xhtml')}" process="@this" update=":mainframe">Frame 3</p:commandLink>
            </h:panelGrid>
        </h:form>

        <p:outputPanel id="mainframe">
            <ui:include src="/frames/#{mainFrameBean.menuValue}" />
        </p:outputPanel>

```
- we have datagrid coded in iframe2.xhtml

```
<p:dataGrid var="car" value="#{dataGridView.cars}" columns="3" layout="grid"
        rows="12" paginator="true" id="cars" rowStatePreserved="true"
        paginatorTemplate="{CurrentPageReport}  {FirstPageLink} {PreviousPageLink} {PageLinks} {NextPageLink} {LastPageLink} {RowsPerPageDropdown}"
        rowsPerPageTemplate="6,12,16">

        <f:facet name="header">
            Cars for Sale
        </f:facet>

        <p:panel header="#{car.id}" style="text-align:center">
            <h:outputText value="#{car.year}" />
        </p:panel>

    </p:dataGrid>
```

Then if you click "Frame 2" commandlink you get nullpointer returned in the partial response XML.

The problem is UIData incorrectly handles initial descendant state during the render response phase.

The problem is also in the reference JSF 2.2.0 implementation. Myfaces 2.2.10 is OK. I've created pull request for primefaces UIData.

Jiri

kukel:
PF version affected? Did you try latest Mojarra? (2.2.12 I think)

slovi:
Affected versions of primefaces are 5.2, 5.3, 6.0, 6.1 (SNAPSHOT) ...

The incorrect method is org.primefaces.component.api.UIData.saveDescendantInitialComponentStates

[...]

PR #1655, opened 2016-08-01
#1654 - initialdescendant component state is incorrectly handled during

slovi:
Initialdescendant component state is incorrectly handled during render response phase - it is neccesary to check if the initialState is marked. Otherwise we get delta state in case we need full state.


slovi:
This PR can be used (as a result of the discussion in #1654)

PR #1672, opened 2016-08-05
populateFitlerParameterMap doesn't detect empty filter's values when …

schlebe:
…filterValue is an array

populateFilterParameterMap() method prepare filter of SelectCheckBoxMenu EVEN IF value's filter are empty !
The problem happens because filterValue.toString().trim().equals(Constants.EMPTY_STRING)) doesn't work correctly when filterValue is an array !

This bug is discussed on Primefaces forum at http://forum.primefaces.org/viewtopic.php?f=3&t=46468


tandraschko:
Thanks for providing a PR but i didn't liked the comments in your PR, so i didn't apply it directly.
Fixed it by:  #1698
Please give it a try

PR #1673, opened 2016-08-05
FilterBy attribute of <p:columns> cause a crash when valueExpression …

schlebe:
…is a method

The program return an error when it executes the substring on expressionString because indexOf("]") return -1.

This worked correctly on Primefaces 5.3 !

This bug is discussed on Primefaces forum at http://forum.primefaces.org/viewtopic.php?f=3&t=46468

SORRY IMPORTANT: in code source I have forgotten to remove followinf line just before int iOpen =

expressionString = expressionString.substring(expressionString.indexOf("[") + 1, expressionString.indexOf("]"));  
             expressionString = "#{" + expressionString + "}";

and I don't find how to remove it now !

### DIFF
--- a/src/main/java-templates/org/primefaces/component/datatable/DataTableTemplate.java
+++ b/src/main/java-templates/org/primefaces/component/datatable/DataTableTemplate.java
@@ -625,8 +625,23 @@ import javax.faces.event.BehaviorEvent;
     public String resolveDynamicField(ValueExpression expression) {
         if(expression != null) {
             String expressionString = expression.getExpressionString();
+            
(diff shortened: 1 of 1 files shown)


tandraschko:
please check tutorials how to use PR's correclty then.
I will apply correct PR's without your "test if".

schlebe:
these 2 lines has been removed in another PullRequest on another branch.
Sorry, I'm not used to work on GITHUB.

tandraschko:
I won't apply it for as i don't understand it.
Would you please create a example based on the showcase to replicate it and create a issue?
As you mentioned in the forum post, this method shoudln't be called if you have static columns (and i don't see a dynamic column in your example)

PR #1706, opened 2016-08-18
Fix #1633

### LINKED ISSUE #1633: OutputLabel does not support composite component
In a very simple way, here is a composite:

```
<ui:component 
>
    <cc:interface>
        <cc:attribute name="value" />
        <cc:attribute name="required" />
        <cc:editableValueHolder name="calendar" />
    </cc:interface>

    <cc:implementation>
        <p:calendar id="calendar" value="#{cc.attrs.value}" required="#{cc.attrs.required}" showOn="button"/>  
    </cc:implementation>
</ui:component>
```

Here how I use it:

```
<p:outputLabel for="rat:calendar" value="Hello" />
<my:calendar id="rat" required="true"/>
```

A mandatory icon should be displayed but it does not. In `OutputLabelRenderer` everything looks fine, except in line 118, `input.isRequired()` returns false while I explicitly set true.

Rapster:
Here is a first approach, what do you think? If you're with okay it, I'd like to refactor a bit as the whole `encodeEnd` method is quite hard to read...

### DIFF
--- a/src/main/java/org/primefaces/component/outputlabel/OutputLabelRenderer.java
+++ b/src/main/java/org/primefaces/component/outputlabel/OutputLabelRenderer.java
@@ -19,8 +19,10 @@ import java.io.IOException;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+
 import javax.el.PropertyNotFoundException;
 import javax.el.ValueExpression;
+import javax.faces.component.ContextCallback;
 import javax.faces.component.UIComponent;
 import javax.faces.component.UIInput;
 import javax.faces.context.FacesContext;
@@ -47,6 +49,7 @@ public class OutputLabelRenderer extends CoreRenderer {
         OutputLabel label = (OutputLabel) component;
         String clientId = label.getClientId(context);
         String value = ComponentUtils.getValueToRender(context, label);
+        UIComponent targetParent = null;
         UIComponent target = null;
         String targetClientId = null;
         UIInput input = null;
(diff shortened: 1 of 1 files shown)


Rapster:
targetParent should be the parent of the target input, what I did does not cover entirely this case:

```
Composite_A.xhtml -> targetParent
 `-- Composite_B.xhtml
 |   `--Composite_C.xhtml
     |   `-- h:inputText -> target
```

In that case, targetParent should be `Composite_C.xhtml` and not `Composite_A.xhtml`. I need to retrieve the UINamingContainer of the `target`. Is there an existing which can do this, or we need to implement our own which retrieve the last composite through `cc:editableValueHolder`?

Rapster:
I just tried, calling `getParent()` on target and looks good enough :) What I did should be moved to an util method so I can reuse it in DynaFormRenderer of PE

Rapster:
As long `deepestEditableValueHolderTarget` is used, we should provide an effecient way to return the `editableValueHolder`, targetClientId, required, labelledBy and valid value, so we don't have to execute same test everywhere this method is called

tandraschko:
Shoudln't it be enough to reuse the "just lookup" logic from SEF?

```
            IdContextCallback callback = new IdContextCallback();
            context.getViewRoot().invokeOnComponent(context, "idFromFindComponent", callback);
            component = callback.getComponent();
```

?
As far as i understand Leo, the resolved component should have the correct context attached and #isRequired should automatically work.

Rapster:
It is not enough actually, I just tried your suggestion but not working... Using `invokeOnComponent` will temporarely set the component composite in the context _(through pushComponentToEL(context, this);)_, and after `invokeContextCallback` is called, composite is detached from it _(through popComponentFromEL(context))_. Attributes like `required` should be evaluated in `invokeContextCallback` call.

Rapster:
I change the way how to solve this problem in my project. I will renew the PR in a month, when I'll be back from holidays :)

PR #1747, opened 2016-09-07
Fix 1687

andrei-ivanov:
Now the message looks like this: `getRowKey(T object) must be implemented by package.SiteAccessRequestsPage$1 when basic rowKey algorithm is
 not used [component=siteAccessRequestsTable,view=/site-access/SiteAccessRequests.xhtml].`

Hmm, the jquery upgrade revert ended up here, even though I didn't commit it to the local branch, but I think the push included it.
Can you somehow ignore it? Or give me a hint on how to fix it? My git knowledge is pretty basic.

### DIFF
--- a/src/main/java/org/primefaces/model/LazyDataModel.java
+++ b/src/main/java/org/primefaces/model/LazyDataModel.java
@@ -19,6 +19,8 @@ import java.io.Serializable;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


tandraschko:
2 things:
1) please don't pull the jquery revert
2) please do a null check on the current component (not sure if it would fail e.g. when you use a exporter)

PR #1773, opened 2016-09-20
Fix #1772 SearchExpressionFacade JS Enhancement

### LINKED ISSUE #1772: PR: SearchExpressionFacade JS Enhancement
I am submitting a Pull Request for this enhancement.
## 1) Environment
- PrimeFaces version: 6.04
- Does it work on the newest PrimeFaces version? No
- Application server + version: Mojarra 2.28
- Affected browsers: All
## 2) Enhancement

While working on a PrimeFaces Extension issue https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/289 I had the need to get the resolved components from SearchExpression Facade not as Jquery Selectors but as their String CSS selectors to be passed to another JQuery component as selector text.

Example: "@(div.mystyle :input) @(.ui-inputtext)" as "div.mystyle :input, .ui-inputtext"

Currently the JS has resolveComponentsAsSelector() which returns Jquery Selector Objects $() only.  I have created a new method resolveComponentsAsCssSelectors() that is then re-used by resolveComponentsAsSelector() but can be used standalone as well to get an array of String elements representing the Jquery Selectors.

melloware:
I wasn't sure if CssSelector is the right name for the method but wanted to distinguish it from Jquery Selectors since this method returns strings and not Jquery Objects.

This simply breaks apart the evaluating of the CSS selector strings from the resolving of components.  It doesn't change anything about this class it simply breaks the step into two methods so they can be re-used by anyone easily.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/core/core.expressions.js
+++ b/src/main/resources/META-INF/resources/primefaces/core/core.expressions.js
@@ -2,13 +2,13 @@ PrimeFaces.expressions = {};
 
 PrimeFaces.expressions.SearchExpressionFacade = {
 
-   /**
-     * Converts expressions into an array of Jquery Selector objects. 
-     * e.g. @(.ui-inputtext) to '$(.ui-inputtext)'
+	/**
+	 * Converts expressions into an array of Jquery Selector objects. 
+	 * e.g. @(.ui-inputtext) to '$(.ui-inputtext)'
 	 *
 	 * @param expressions the expressions to evaluate
 	 * @return an array of Jquery Selector objects
-     */
+	 */
 	resolveComponentsAsSelector: function(expressions) {
 		var elements = $();
 		var cssSelectors = PrimeFaces.expressions.SearchExpressionFacade.resolveComponentsAsCssSelectors(expressions);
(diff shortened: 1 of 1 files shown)


tandraschko:
Why do you need this?

melloware:
So in the PF Extensions fix for Tooltip in this ticket..

https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/289

The issue was the global tooltip code was not working properly because it was doing this...

`$('body').off('.tooltip').on(this.cfg.show.event + '.tooltip', selector, function(event) {`

the value for "selector" is expecting a comma separated list of selectors like ".ui-inputfield, [title]" etc.  But the code was previously getting a list of JQuery elements $() using...

 `PrimeFaces.expressions.SearchExpressionFacade.resolveComponentsAsSelector(this.cfg.forTarget);`

But that returns all resolved Jquery elements not the original CSS selectors.    You can see what I did here:

https://github.com/primefaces-extensions/core/blob/master/src/main/resources/META-INF/resources/primefaces-extensions/tooltip/1-tooltip.js#L31

Maybe there is a better way to do what I need?

tandraschko:
To be honest, i never used the selector attribute on 'on'.
Wouldn't this be the same?
`selector.off('.tooltip').on(this.cfg.show.event + '.tooltip', function(event)`

selector would be the resolved elements by resolveComponentsAsSelector

melloware:
OK I will give that a try and report back.

melloware:
That does not seem to work.

tandraschko:
maybe you can analyze the code a little bit, attaching events the document may not be the best way but i also don't know whats the actual purpose of this line.

melloware:
I am closing this pull request you are right I see what is going on and this pull request is not actually correct anyway.  Sorry!

PR #1778, opened 2016-09-21
Don't hide autocomplete panel if itemtip clicked

firsthour:
If the itemtip is the target of the mousedown event, then don't hide the entire autocomplete result panel. This allows the user to click an anchor link in the itemtip or copy text from it if necessary.


tandraschko:
Thanks for the PR but fixed it a little bit nicer.

firsthour:
No problem, just appreciate you taking the time to look at it!

Issue #1790, opened 2016-09-22
ajax update not work for selectonemenu if inside <ui:component> and rendered via facet


markusg80:
## 1) Environment
- PrimeFaces version: 6.0.4, 5.3.19
- Does it work on the newest PrimeFaces version? No
- Application server + version: Glassfish 4.1.1
- Affected browsers: Firefox, Chrome
## 2) Expected behavior

A Button inside the ui:component opens dialog (wich is declared inside the >ui:component>) with a form inside and content rendered from a facet located in parent xhtml. SelectOneMenu should have actual value selected after update.
## 3) Actual behavior

form is updated and also all values on the Backing Bean are good. But nothing inside the SelectOneMenu is selected. -> the ajax update doesn't. work
## 4) Steps to reproduce

In Sample Project inside table choose row with value A to D and click edit Button
## 5) Sample XHTML

See sample project
## 6) Sample bean

See sample project
## 7) Workaround

Remove the form inside the ui:component dialog and place it in the edit named facet from parent xhtml
[ccselectbug.zip](https://github.com/primefaces/primefaces/files/488268/ccselectbug.zip)
## 8) PR to solve the Problem
#1791

tandraschko:
It doesn't work with MyFaces either.
I think it would be the best if you would create a issue on MyFaces side and let them check if it should work. If it will get fixed in MyFaces, we can easily check the changes.

tandraschko:
Background: Not everything that works in Mojarra must work by definition in the JSF specs. Leonardo from MyFaces will likely check that spec compatibilty and fix it if possible.

PR #1791, opened 2016-09-22
solve ajax update not work for selectonemenu if inside <ui:component> and rendered via facet 

markusg80:
Solve issue #1790. Please review if this generate side effects


tandraschko:
We do it the same as myfaces (the jsf impl) does.
Does it work with h:selectOneMenu correctly for you?

tandraschko:
Same for Mojarra. Will close the PR as we are aligned with the JSF specs.

markusg80:
With h:selectonemenu it works!

markusg80:
Please reopen

tandraschko:
If it works with h:selectOneMenu, than the fix in this PR is not correct. The empty value handling, which is changed by your PR, is the same as in the JSF impl.

Issue #1799, opened 2016-09-30
IE11: Could not complete the operation due to error 800a025e


beevelop:
> This issue has been logged with the help of Sentry and can be reproduced reliably in our environment.
## 1) Environment
- PrimeFaces version: 5.3
- Affected browsers: IE11 (11.0.9600.*)
## 2) Expected behavior

Executing `PrimeFaces.clearSelection()` should not trigger an exception.
## 3) Actual behavior

Running `PrimeFaces.clearSelection()` triggers the following exception for rowClicks on a datatable in IE11:

```
Error: Could not complete the operation due to error 800a025e.
  at b.clearSelection(/javax.faces.resource/primefaces.js:1:4094)
  at onRowClick(/javax.faces.resource/primefaces.js:18:20202)
  at Anonymous function(/javax.faces.resource/primefaces.js:18:5632)
  at bH.event.dispatch(/javax.faces.resource/jquery/jquery.js:25:40977)
  at cd.handle(/javax.faces.resource/jquery/jquery.js:25:37104)
  at d(/javax.faces.resource/js/raven.min.js:2:3984)
```
## 4) Steps to reproduce

Click on a row in a datatable or manually execute `PrimeFaces.clearSelection()` with the browser's developer console.

tandraschko:
Merged, thanks!

PR #1800, opened 2016-09-30
Core: Prevent IE exceptions for clearSelection

beevelop:
This PR relates to #1799

---

``` js
window.getSelection().rangeCount > 0
```

is required in order to prevent `IndexSizeError`-exceptions in IE

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/core/core.js
+++ b/src/main/resources/META-INF/resources/primefaces/core/core.js
@@ -253,7 +253,7 @@
             if(window.getSelection) {
                 if(window.getSelection().empty) {
                     window.getSelection().empty();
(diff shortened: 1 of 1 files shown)


tandraschko:
if "window.getSelection().rangeCount > 0" is enough, why you added the second check?

beevelop:
@tandraschko Sorry, I was not quite clear about that:

``` js
window.getSelection().rangeCount > 0
// prevents IndexSizeError-exceptions in IE for the following check (see right below)
// getRangeAt(0) can trigger an IndexSizeError if there is no selection

window.getSelection().getRangeAt(0).getClientRects().length > 0
// prevents the actual exception "Error: Could not complete the operation due to error 800a025e."
```

PR #1815, opened 2016-10-07
Fix #1633

### LINKED ISSUE #1633: OutputLabel does not support composite component

Rapster:
Better fix for #1633. Once merged, I'll commit the fix for https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/389


tandraschko:
To be honest, i currently don't like the fix very much.
Maybe it's a better way to always use ContextCallback or something but i don't like to rebuild the state of component in a extra pojo. I would be great to work on the component directly.

Rapster:
What do you mean by rebuild? The state is only created once, and then updated. Sure, it'd be great to work directly on the component but as mentionned by MyFaces member, there is no other way to do so. Feel free to provide ideas for a better fix, I'm using this one so far on our project and works great.

src/main/java/org/primefaces/util/EditableValueHolderState.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public boolean isValid() {
+        return valid;
+    }
+
+    public boolean isRequired() {
+        return required;
+    }
+
+    public boolean isEditableValueHolder(UIComponent component) {
```
Rapster: This should be moved into ComponentUtils or something

tandraschko:
Just to remember, i applied a PR in this class: #1960

tandraschko:
I currently working on a new search API for JSF 2.3, so i will check how we can easier solve this problem.

Rapster:
Interesting! How do you intend to do that? We'll need a solution for JSF 2.2 though. It'd be cool to have this PR in next PF release so I don't have to maintain a custom PF version. Moreover, I need this PR in PE as wekk, so I'm maintaining two custom versions

tandraschko:
Yep, sure.
Will come back later, after JSF 2.3 API and before 6.1 release :P

Rapster:
Final release of JSF 2.3 is planned for January?

tandraschko:
nope, but finalising the specs.

Rapster:
The new API should cover all those components which have a "for" attribute. Like outputLabel, tooltip, to complete

tandraschko:
the JSF API won't cover it, it was already to late and it's quite complex. If we do it in PF, it's a just hack till (hopefully) the next JSF version.

tandraschko:
I currently working on this and it works fine in MyFaces, however not in Mojarra.

It seems that the context is not available, also if invokeOnComponent is used. Therefore it's fails to resolve the "cc.*" EL's and it fails to get the correct "required" value.

I will push my changes. But it would be great if you could have a look why it doesn't work on Mojarra.

tandraschko:
I also pushed some tests to the showcase, just need to remove the comment in outputLabel.xhtml.

tandraschko:
Thanks for the PR but i solved it a little bit different. Will close this now. Lets discuss in the ticket.

Rapster:
Thanks @tandraschko I will try your work on my project as soon as I get back from vacation. I'm not sure I understood the problem. How come it worked with my PR, what you did looks the same as my PR? Another thing, it'd be great to fix same issue in PE in DynaForm ;)

tandraschko:
Enjoy your vacation and lets talk when you are back ;) We still have some weeks till a final release.

PR #1850, opened 2016-10-20
jquery plugins better description

erickdeoliveiraleal:
jquery plugins better description


tandraschko:
Thanks! Applied some changes but no all.

erickdeoliveiraleal:
I stil think that the description is wrong.  jQueryUI position is not present, and Draggable etc, is in umbrella of UI Interactions.

tandraschko:
better? ;)

erickdeoliveiraleal:
👍

Issue #1915, opened 2016-11-22
Menu item outcome link doesn't evaluate navigation rule parameter value


AndrewShwaika:
[OutcomeTargetRenderer.patch.txt](https://github.com/primefaces/primefaces/files/606266/OutcomeTargetRenderer.patch.txt)
## 1) Environment
- PrimeFaces version: 6.0.8
- Does it work on the newest PrimeFaces version? no
- Application server + version: Glassfish 4.1
- Affected browsers: all

## 2) Expected behavior

If navigation case in faces-config.xml has view-param value expression it should be evaluated

## 3) Actual behavior

Outputs as written

Class com.sun.faces.renderkit.html_basic.OutcomeTargetRenderer from org.glassfish/javax.faces-2.2.13 has method addNavigationParams with such evaluation, lines 218-230.

Supposed patch is attached

Rapster:
Please provide a PR instead

PR #1916, opened 2016-11-22
#1915 Evaluate navigation rule parameter value

AndrewShwaika:


tandraschko:
Why does it only take of the first value?

AndrewShwaika:
As original from Mojarra javax.faces 2.2.13

tandraschko:
ok - how is it done in MyFaces?

AndrewShwaika:
Yes. MyFaces evaluates them all.

tandraschko:
So it would be great if you could provide a PR based on MyFaces.

PR #1917, opened 2016-11-22
#1915 Evaluate navigation rule parameter value. All values as MyFaces

AndrewShwaika:
Iterate all parameter values.

Sorry for removing white spaces on empty lines


tandraschko:
Looks good but it would be great if you could make the util methods protected and non-static

Rapster:
Use Java Convention: do not add an open bracket on a new line but following method signature

Rapster:
Don't worry about the white spaces, I'd say it's a good thing ;)

Issue #1947, opened 2016-12-02
Show delay for OverlayPanel


androidominus:
Hi

I would to ask if in the next releases of PrimeFaces you can add a show delay for OverlayPanel.

This is just to avoid a nasty user experience in case of several overlaypanels targeting close components (like buttons in table rows). 
If you pass over the targets fast the overlay panels will appear about in the same time.

To work around this I just overrided the show function of the widget adding a delay.

I think it is very simple to you to add a property delay as well as tooltip.

Thank you

Giovanni

Rapster:
It looks like you already figured how to do it. Feel free to provide a PR

tandraschko:
@androidominus The PR looks fine, however it would be great to provide a sample, so that i can retest your changes.

androidominus:
@tandraschko Thomas what kind of sample do you need? Should be a minimal war containing just the usage of the new attribute or what? Thank you

tandraschko:
A small xhtml with the new attribute + attached to a button or something is enough :)

androidominus:
Should be enough.

`<p:graphicImage id="img" name="/demo/images/images.png" style="cursor:pointer" title="Show Delay" />
<p:overlayPanel  for="img" showEffect="blind" hideEffect="explode" showEvent="mouseover" hideEvent="mousedown"
                   showCloseIcon="true" showDelay="1000">
	<p:graphicImage name="/demo/images/nature/nature2.jpg" width="300" />
</p:overlayPanel>`

PR #1948, opened 2016-12-04
Added showDelay to OverlayPanel.

androidominus:
I did just few modification to the original code to consider the new parameter showDelay.
This parameter is optional and if setted show the overlay panel after the delay in milliseconds.


Rapster:
See #1947

tandraschko:
can't we use jquery's delay method here?

androidominus:
I don't think so...
This method doesn't allow to cancel the delay and the clearTimeout function is necessary to prevent the show event in case of a rapid transition on target.

PR #1952, opened 2016-12-05
Fix Issue # 1944(Image Cropper throws java.awt.image.RasterFormatException: negative or zero width, when cropper is deselected)

FastainCS09:
p:imageCropper throws exception when the image is deselected.
For this i have added a check in function in 
**@Class ImageCropperRenderer** 
@Method  **getConvertedValue**

add these two lines ,where i check the width and height are not equals to zero.

  if((w==0)&&(h==0)){
            return null;
        }


tandraschko:
please apply our formatting (you miss empty spaces)

Rapster:
Use same condition as in Raster constructor, e.g:
```java
if (width <= 0 || height <= 0) {
    throw new RasterFormatException("negative or zero " + ((width <= 0) ? "width" : "height"));
}
```

you'll return null instead ;)

FastainCS09:
Thanks ,I have formatted the code as suggested above.The last commit is with formatted changes. :+1:

PR #1962, opened 2016-12-10
jQuery.fn.size() is deprecated; use the .length property as of Jquery 1.8+

melloware:
As reported by the JQuery 1.x migrate tool size() should be .length as of 1.8+.  This will help ease the eventual transition to JQ 3.0.

```
JQMIGRATE: Migrate is installed with logging active, version 1.4.1
jquery-migrate-1.4.1.js:45 JQMIGRATE: jQuery.fn.size() is deprecated; use the .length property
migrateWarn @ jquery-migrate-1.4.1.js:45
```


tandraschko:
But there is currently no need to apply this?
I assume that its currently very hard to upgrade JQ 3.0 as we use not only a few external plugins.
Also i don't think that we can upgrade to JQ 3.0 in 2017.
There are still IE8 users....

melloware:
Ahh understood.  No need for this but this was the Jquery 1.X migration tool saying these were deprecated and since we are on 1.11.3 I thought it was OK to fix these now as a "why not".

PR #1974, opened 2016-12-19
Fix#1889 (p:calendar timeControlType="select" does not work with national time pattern)

FastainCS09:
The **if (hasTime()) Condition** in CalenderRenderer.java is buggy,it returns true only when the pattern is not null or contains (":") ,so that's why no pattern runs except that contains separator **(:)**.

Here is the code for checking time.
**public boolean hasTime() { 
     String pattern = getPattern(); 
     return (pattern != null && pattern.indexOf(":") != -1); 
}**
The work around is to comment the check and it will works for all the patterns alongwith any separator.
It has no consequence,even the user doesn't mention time with the pattern.

### DIFF
--- a/src/main/java-templates/org/primefaces/component/calendar/CalendarTemplate.java
+++ b/src/main/java-templates/org/primefaces/component/calendar/CalendarTemplate.java
@@ -77,7 +77,7 @@ import javax.faces.event.PhaseId;
     public boolean hasTime() {
         String pattern = getPattern();
 
-        return (pattern != null && pattern.indexOf(":") != -1);
+        return (pattern != null && (pattern.contains("HH") || pattern.contains("mm") || pattern.contains("ss")));
     }
 
     @Override
--- a/src/main/java/org/primefaces/component/calendar/CalendarRenderer.java
+++ b/src/main/java/org/primefaces/component/calendar/CalendarRenderer.java
@@ -200,7 +200,7 @@ public class CalendarRenderer extends InputRenderer {
             wb.attr("showOtherMonths", true).attr("selectOtherMonths", calendar.isSelectOtherMonths());
         }
 
-        if (calendar.hasTime()) {
+       
             String timeControlType = calendar.getTimeControlType();
             
             wb.attr("timeOnly", calendar.isTimeOnly())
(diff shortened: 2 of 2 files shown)


tandraschko:
So there is no reason that we had the if statement around the code block?

FastainCS09:
Yeah,I checked it,even though , you donot mention time in the pattern,it will have no issue .

tandraschko:
| you donot mention time in the pattern,

not sure if in understand. It's possible to set time in the pattern.

FastainCS09:
Just a continuation of the above comment, if u look at the funciton **hasTime()**
The last statement states : 
`return (pattern != null && pattern.indexOf(":") != -1);`

so if the pattern is "**MM:DD:YYYY**" ,,it still returns true,though the time is not given. So in any case,either  we comment the if block or check c only that the pattern is not null ,or other work aorund is to check,whether it has Pattern HH,MM,SS and then only returns true.

tandraschko:
Yeah but it's possible to set e.g. pattern="MM/dd/yyyy HH:mm”.

FastainCS09:
yeah sure,i agree with you that we can add the time ,but if the pattern is e.g norwegian **MM.dd.yyyy HH.mm** then in that case ,it returns false.That was the problem,for which i propose that solution.

tandraschko:
Yeah but your fix looks just like an dirty workaround. We should fix the hasTime() method.

FastainCS09:
If you want to retain the if code block ,then one solution  should be like this ,it is just a sample solution.

`return (pattern != null && pattern.contains("HH")  || pattern.contains("MM) || pattern.contains("SS");`

FastainCS09:
if u want to improve the above code ,then feel free to give me your suggestion and i will then create a another updated pull request.

tandraschko:
yeah, something like that looks better.
but "MM" is month, we need to check "mm". "ss" are also lower case.

FastainCS09:
Ok No issue,I will take care of lower case as well.So i will now create an updated Pull Request.Thanks for taking out time and giving me your feedback.

tandraschko:
Thanks!

FastainCS09:
You welcome

PR #2028, opened 2017-01-20
Fixed error counting the number of options in a select when there are…

didiez:
#… items inside groups (SelectItemGroup). 
This issue can cause problems rendering selects with groups with a large number of options inside (not setting a fixed height by default).

Grouping example in the showcase http://www.primefaces.org/showcase/ui/input/oneMenu.xhtml

For example: a `<p:selectOneMenu>` with 4 groups of 5 items each, ends up rendering a huge panel with an `<ul>` with 24 `<li>` without scroll (`height: auto`) because for the renderer there are only 4 items


tandraschko:
Would you please create a issue with a example, so that i can replicate and the error and test your PR?

PR #2053, opened 2017-02-02
add ui-widget styleClass to SelectOneMenu when editable is true. Fix #2043 

### LINKED ISSUE #2043: SelectOneMenu: minor style issue with editable="true"
## 1) Environment
- PrimeFaces version: 6.0.14

## 2) Expected behavior
The Fontsize of a selectOneMenu should always be the same no matter wether it is editable or not

## 3) Actual behavior
If a selectoneMenu is editable the InputField used is having a smaller fontsize than the rendered label if editable is false. 

## 4) Steps to reproduce
Just look at it in the Primefaces Showcase: http://www.primefaces.org/showcase/ui/input/oneMenu.xhtml

The Editable InputField has a fontsize ofg 13.3333px, while the non editable has  14.4px (Firefox)

## FIX:
To fix this adding ui-widget to the editable input field as style class is enough. Since the regular Input Fields also have "ui-widget" in their style classes, I think it was just forgotten here for the input field.
``` 
Current way:
<input class="ui-selectonemenu-label ui-inputfield ui-corner-all" type="text">

Correct way:?!
<input class="ui-widget ui-selectonemenu-label ui-inputfield ui-corner-all" type="text">
```

sizer:
Fix #2043
add ui-widget styleClass to SelectOneMenu when editable is true.

### DIFF
--- a/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
+++ b/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
@@ -170,7 +170,7 @@ public class SelectOneMenuRenderer extends SelectOneRenderer {
             writer.startElement("input", null);
             writer.writeAttribute("type", "text", null);
             writer.writeAttribute("name", menu.getClientId(context) + "_editableInput", null);
-            writer.writeAttribute("class", SelectOneMenu.LABEL_CLASS, null);
+            writer.writeAttribute("class", SelectOneMenu.LABEL_CLASS + " ui-widget", null);
             
             if(menu.getTabindex() != null) {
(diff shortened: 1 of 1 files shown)


tandraschko:
both generated markups has the ui-widget in the parent, so the fix in this PR is not "cool".
The difference is that the editable is a "input" and the non-editable is a "label".
Would be great if you could look for a better way to fix it. On the other hand, would be correct if the editable selectOneMenu has a different front size as a inputText?
Should all of them share the same font-size?!

gako:
You are right the nested ui-widgets are not really "nice" ;-)
Another way to fix this would be by using css rules to adjust the size of the nested editable inputfield to be the same size.  But that would have to been done in all the themes themself, which makes it hard to fix this issue in a single point.

Issue #2045, opened 2017-02-01
update hotkey / enable hotkey in password fields


martin654:
# 1) Environment
- PrimeFaces version: 6.0.14
- Does it work on the newest PrimeFaces version? no
- Application server + version: any
- Affected browsers: any

## 2) Expected behavior
p:hotkey behavior with p:password is the same as p:inputText

## 3) Actual behavior
<p:hotkey/> filters binded keys inside <p:password/> field

## 4) Steps to reproduce
1) open page
2) write some text into password field
3) try to remove text using backspace key - it does not work
4) try the same in username field - it works

## 5) Sample XHTML
...
```
<p:hotkey bind="backspace" ignoreAutoUpdate="true" global="false" handler="" />

<h:form>
  <!-- In p:inputText I can press backspace to remove inserted text ->
  <p:inputText id="username" />
  <!-- In p:password backspace does not work ->
  <p:password id="password" />
</h:form>
```
..
 
## 6) Sample bean
not needed

martin654:
Could this change be enough?

```
 src/main/resources/META-INF/resources/primefaces/hotkey/hotkey.js | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/src/main/resources/META-INF/resources/primefaces/hotkey/hotkey.js b/src/main/resources/META-INF/resources/primefaces/hotkey/hotkey.js
index fbd71c7..732637c 100644
--- a/src/main/resources/META-INF/resources/primefaces/hotkey/hotkey.js
+++ b/src/main/resources/META-INF/resources/primefaces/hotkey/hotkey.js
@@ -44,7 +44,7 @@
 		handleObj.handler = function( event ) {
 			// Don't fire in text-accepting inputs that we didn't directly bind to
 			if ( this !== event.target && (/textarea|select/i.test( event.target.nodeName ) ||
-				 event.target.type === "text") ) {
+				 event.target.type === "text" || event.target.type === "password" ) ) {
 				return;
 			}
 			
```

martin654:
The same problem is with p:textEditor

PR #2075, opened 2017-02-10
#2045 hotkey filters keys in password field

ludekz:


tandraschko:
Please provide a PR here: https://github.com/jeresig/jquery.hotkeys
If it's fixed, please provide a PR with the updates script here.
It's better to fix the plugin itself than hack everything in PF.

tandraschko:
Also we still have a very old version, maybe its already fixed. So please check ;)

Issue #2038, opened 2017-01-26
Improve SelectCheckboxMenu performance


mertsincan:
Reported By PRO User;
```
We are using <p:selectCheckboxMenu ../> component in our application.
We have noticed that when the option list in drop down increases by more than 2000 entries, selecting and un-selecting All options using toggle checkbox takes more time.(approximately 3-5 secs)
...
```

PR #2090, opened 2017-02-15
Fix #2089

### LINKED ISSUE #2089: selectManyMenu: poor performance when selecting all elements
In my view I have a single checkbox that triggers selectAll/unselectAll on a p:selectManyMenu. This is extremely slow for a large item set (~ 2000 elements).

Environment:
- PrimeFaces version: 6.0
Especially slow in Microsoft Edge

Mobe91:
This improves performance of selectAll/unselectAll on selectManyMenu by partially cloning and replacing the DOM.
Might also be relevant for #2038


tandraschko:
i removed forms.js in favor of small javascripts for each widget.
So please provide a fixed PR again.
NOTE: it would be great if you could first provide a issue for adding selectAll/unselectAll methos for those widgets and then, in a second step, provide a PR to improve the performance.
Thanks!

Issue #1940, opened 2016-12-01
p:inputNumber doesn't work in conjunction with SeleniumIDE


movalz:
## 1) Environment
- PrimeFaces version: 6.0.9
- Does it work on the newest PrimeFaces version? No
- Application server + version: Tomcat
- Affected browsers: FireFox with Selenium IDE 2.9.1

## 2) Expected behavior
When recording user actions on a page with SeleniumIDE Recorder and entering a value into a p:inputNumber field this typing should be registered and added to the Recorder's command list similar to 'type   myFieldId   myEnteredValue'. When replaying the recorded actions this command 'type ...' should put the reccorded value into the number field again.

## 3) Actual behavior
There are 2 issues:
1. the recording of the entered value doesn't function. This is an issue of used JS library autoNumeric which I have found there and have to be fixed there (see https://github.com/BobKnothe/autoNumeric/issues/199).

2. when I add the command 'type   myFieldId   myEnteredValue' by hand and replay it with SeleniumIDE it doesn't function: the value appears in the field but isn't copied to internal hidden input element of p:inputNumber ==> the value is lost when refreshing the page.
This is a PrimeFaces issue. Reason is a missing call of $this.copyValueToHiddenInput() in case of 'onchange' for the visible input field. I have temporarily fixed it for me by adding an onchange handler which does it explicitly:
```
        this.inputExternal.on('change', function(e) {
            if (window['_Selenium_IDE_Recorder']) {
            	$this.copyValueToHiddenInput();
            }});		
```
I'm not sure about potential side effects of this, so I do the copy action in case of running with SeleniumIDE Recorder only.

## 4) Steps to reproduce
See description above. 

## 5) Sample XHTML
arbitrary usage of p:inputNumber

## 6) Sample bean
arbitrary usage of p:inputNumber which stores it's value in a backing bean

tandraschko:
The second issue should be fixed already - please give it a try.
As we can't fix the first issue, i will close it for now.

PR #2129, opened 2017-03-01
Fix problem always empty submittedValue in p:inputNumber

7robertodantas:
Currently the encodeOutput method is creating an input with suffix id "_input", but the decode method it is expecting / capturing the value of an input with suffix id "_hinput". That is causing empty string in submittedValue, no matter value the inputNumber is filled on.

```java

protected void encodeOutput(FacesContext context, InputNumber inputNumber, String clientId) throws IOException {
        String inputId = clientId + "_input";
     // [...]
}

```

```java
   @Override
       // [...]
        String inputId = inputNumber.getClientId(context) + "_hinput";
    }
```

### DIFF
--- a/src/main/java/org/primefaces/component/inputnumber/InputNumberRenderer.java
+++ b/src/main/java/org/primefaces/component/inputnumber/InputNumberRenderer.java
@@ -64,7 +64,7 @@ public class InputNumberRenderer extends InputRenderer {
 
         decodeBehaviors(context, inputNumber);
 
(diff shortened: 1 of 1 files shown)


7robertodantas:
Probably this problem is also related to issue #1940

tandraschko:
Your PR would destroy the component. 
If you would check the generated markup, there is a input (_input) and a hidden input (_hinput). The cleaned value will be copied by JS from _input to _hinput.

Issue #2157, opened 2017-03-09
SelectCheckboxMenu multiple mode caused duplicities in label


vonnai:
## 1) Environment
PrimeFaces version: 6.0.16
Does it work on the newest released PrimeFaces version? NO
Does it work on the newest sources in GitHub? NO
Application server + version: Tomcat 7.70
Affected browsers: ALL

## 2) Expected behavior
The select checkbox menu has not the same value rendered twice in label. 

## 3) Actual behavior
If component is rendered with default values and I check all from overlay panel, all default values are in label twice.

## 4) Steps to reproduce
Open your showcase on selectcheckboxmenu page. In CheckboxView.java add to postconstructor selectedCities2 = new String[]{"Miami"}; and redeploy. The selectcheckbox menu will be rendered with one item label Miami. Then open selectCheckboxMenu overlay panel and click to checkbox for ALL items and close it. Selectcheckboxmenu label has two Miami labels. I think it's caused by attribute data-item-value which is missing on li tag after component has been rendered.

PR #2181, opened 2017-03-15
fix #2172, #2157 select checkbox menu bugs

### LINKED ISSUE #2172: SelectCheckboxMenu multiple mode different itemValue and itemLabel caused errors
## 1) Environment
- PrimeFaces version: 6.0.16
- Does it work on the newest released PrimeFaces version? NO
- Application server + version: Tomcat 7.70
- Affected browsers: ALL

## 2) Expected behavior
If I use inside <f:selectItem itemLabel="Test" itemValue="test"/> I expect, that I don't have to has item label and item value equals.

## 3) Actual behavior
The component doesn't work if itemLabel and itemValue are not equal. I expect, that data-item-value attribute on tag ul li.ui-selectcheckboxmenu-token has value from attribute itemValue and not from itemLabel.

## 4) Steps to reproduce

try this and check something, it throws jquery exception Syntax error, unrecognized expression: [data-item-value=Test] - it used description as value.

`<p:selectCheckboxMenu multiple="true" label="Components">
                                <f:selectItem itemValue="Test" itemDescription="test"/>
                                <f:selectItem itemValue="Test2" itemDescription="test2"/>
                                <f:selectItem itemValue="Test3" itemDescription="test3"/>
</p:selectCheckboxMenu>`
or
`
<p:selectCheckboxMenu multiple="true" label="Components">
                                <f:selectItem itemValue="Test"/>
                                <f:selectItem itemValue="Test2"/>
</p:selectCheckboxMenu>
`

PS:
The similary problems with attribute data-item-value I found in [#2157](https://github.com/primefaces/primefaces/issues/2157)

vonnai:
Hi, I fixed issues #2172, and #2157. There was problems with attribute data-item-value. Attributes itemLabel and itemValue had to be equal. I added support for itemEscaped attribute. It's possible to have items in multiple label not escaped. It seems, everything in multiple mode works fine.

### DIFF
--- a/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
+++ b/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
@@ -193,11 +193,31 @@ public class SelectCheckboxMenuRenderer extends SelectManyRenderer {
             for(int i = 0; i < length; i++) {
                 Object value = Array.get(values, i);
                 writer.startElement("li", null);
+                
                 writer.writeAttribute("class", SelectCheckboxMenu.TOKEN_DISPLAY_CLASS, null);
-
+                writer.writeAttribute("data-item-value", value, null);
+                
                 writer.startElement("span", null);
                 writer.writeAttribute("class", SelectCheckboxMenu.TOKEN_LABEL_CLASS, null);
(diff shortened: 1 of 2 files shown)


mertsincan:
Thanks a lot for your PR ;)

vonnai:
you'r welcome, thanks for fast merge and review :-)

PR #2189, opened 2017-03-19
Fix #1495 - Remove repeated items in p:autocomplete result search

### LINKED ISSUE #1495: Add unique attribute to p:autocomplete
When an item is already selected in p:autocomplete, it would be very good if the result query search remove/omit these values

image

the idea is not to have repeated values

image

Rapster:


tandraschko:
As its a new feature, i will merge it after the release.

tandraschko:
Would you please not change the newline on "else"? We actually do a newline before "else".

Rapster:
Don't know what happened...

PR #2190, opened 2017-03-19
Fix #2185 - Ribbon onTabChange

### LINKED ISSUE #2185: Ribbon onTabChange
The documentation say that ribbon delivered from tabview, and any event that implemented in tabview should use. I try to implement onTabChange event to hold actual page index in server side, but the javascript doesn't invoked. I think there missing event handler, so I looked around the code. 

primefaces/ribbon/src/main/java/org/primefaces/component/ribbon/RibbonRenderer.java
primefaces/ribbon/src/main/resources-maven-jsf/ui/ribbon.xml

There is no onTabChange event implemented. Please change the source for handle tab change events (attached)
[ribbon.zip](https://github.com/primefaces/primefaces/files/847651/ribbon.zip)

Rapster:
I did not change attributes since some are not supported yet (RTL for example)

### DIFF
--- a/src/main/java/org/primefaces/component/ribbon/RibbonRenderer.java
+++ b/src/main/java/org/primefaces/component/ribbon/RibbonRenderer.java
@@ -22,10 +22,18 @@ import javax.faces.context.FacesContext;
 import javax.faces.context.ResponseWriter;
 import org.primefaces.component.tabview.Tab;
 import org.primefaces.renderkit.CoreRenderer;
+import org.primefaces.util.ComponentUtils;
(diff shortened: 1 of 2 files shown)


tandraschko:
ribbon now extends from TabView, it's a big change. Why is it required?

Rapster:
To avoid a lot of copy paste :)

tandraschko:
For what exactly?

Rapster:
Support of client behaviors. The ribbon javascript class already extends tabview actually.

Issue #2062, opened 2017-02-07
Dialog: position wrong if scrolled before shown


djmj:
If you have a dialog with `position="50, 50"` and open the dialog everything is fine and style is `top: 50px;`.

If you reload the page, scroll down the page by 100 pixel and then open the dialog, the dialog is positioned wrong. Its header is not visible, because its style is `top: -50px;`.

This is wrong. It must remain `top: 50px;` instead. Dialog is unaware of scrolling.

The same applies to vertical scrolling.

nivaldin:
Almost same problem here... If I change the number of rows shown in a DataTable paginated, then try to show a dialog, it calculates its top position so big that it's not show on the page.

djmj:
Any change this fix will make it into 6.1, since it seems like a very small change?

ghost:
I have similar issue. When I use a collapsible panel in combination with dialogs, it gives strange results. When I have multiple panels in the page which are all expanded, my dialogs are not visible anymore. When I have just a few, depending on the height of the panel, you can see the dialog is coming partially on the screen. This with default behaviour (no attributes configured). Also configuring attributes doesn't help. Sometimes using positionType absolute helps rendering on the screen, but alignment is not done. The only solution right now is to use position=top, which is the only way to force the dialog to be always on screen. Please fix.

tandraschko:
Feel free to provide an PR - thats the easiest way to get it fixed.

tandraschko:
Did you try positionType="absolute"?

tandraschko:
It's not fixed by my commit - so please feel free to investigate.

melloware:
Here is a test case showing the issue:  
[pf-2062.zip](https://github.com/primefaces/primefaces/files/2720865/pf-2062.zip)

The problem is with this code:
https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/dialog/dialog.js#L353-L362

GedMarc:
@melloware 

```
 collision: “none” 
```
L:352
The recalc is sending it off on collision detection.

melloware:
@GedMarc I tested it locally with my test case above and collision:none has no effect on this.  The bug still remains.

GedMarc:
@melloware  Let me see if I can spot the difference between what I've got and the demo site :)

melloware:
If you donwload this:  https://github.com/primefaces/primefaces/files/2720865/pf-2062.zip

[...]

Issue #1862, opened 2016-10-26
Dialog with fitViewport = true, is not resized on windows resize


vbors:
## 1) Environment
- PrimeFaces version: 6.0
- Affected browsers: Chrome, Firefox (tested)
## 2) Expected behavior

If Dialog has attribute fitViewport  = true,
For After resize windows the dialog must be resized to fit height of new windows. 
...
## 3) Actual behavior
1. Now when resizing windows it  only adjusts the position
   `bindResizeListener: function () {
       var a = this;
       $(window).on(this.resizeNS, function () {
           a.initPosition()
       })
   }`
   ..

melloware:
I think this is related to #899 and fixed by #2201

vonnai:
The point 2. I don't think, that dialog should change size after update some component inside. Dialog should have fixed size to viewport and if a component overflows, scrollbar is the best practice.

vbors:
Yes you are right, I expressed wrong. 
The problem is that now if you have a component with lazy loading, after loading component the dialog exceed the viewport.

vbors:
I fixed this in my project with an css 'max-height' property.

vbors:
this is my code from function fitViewport of my extended Dialog:

```
                var maxContentHeight = pageContentHeight - titleHeight - innerPadding;
                if (this.jq.innerHeight() > pageContentHeight) {
                    this.content.height(maxContentHeight);
                    if (this.cfg.width !== "auto") {
                        this.jq.width(parseInt(this.cfg.width, 10) + 15);//add scroll width
                    }
                }

                //for resizing on partial update
                var cssMaxHeight = this.content.css('max-height');
                if (!cssMaxHeight || cssMaxHeight === 'none') {
                    this.content.css('max-height', maxContentHeight)
                }
```

melloware:
Can your PR accomodate both fixes?

vonnai:
@vbors Did you try my fix? I think, it's not possible to exceed dialog viewport after lazy load.

melloware:
My apologies I did not try the fix I was just asking if there was a clever way to fix all scenarios with your PR.  Like even adding a new property on dialog if someone does need this behavior.

vonnai:
@melloware be cool, I think my PR fixes the point 1 of this issue.

vbors:
@vonnai I have not tried your fix, When I get time, I'll do a test and let you know :)

vbors:
I read your commit changes and there i have a question:
[...]

PR #2201, opened 2017-03-24
fix #899: enhance dialog fitViewport funcionality

### LINKED ISSUE #899: fitViewport in dialog does consider header/footer height
The facelets page below should show a dialog with a content, that is to large to fit on the window. It is expected, that the dialog content is sized, so that header an footer fit the window.

Currently after clicking on "Show Dialog" the dialog is correctly opened, but the footer is invisible (outside the window to the south). When the dialog is resized, the dialog snaps to the right size.

The test was done with:

primefaces 5.3
payara 4.1.153

Client browsers Firefox 42, Chromium 45 and IE 11

``` xml
<?xml version='1.0' encoding='UTF-8' ?>
      xmlns:c="http://xmlns.jcp.org/jsp/jstl/core">
    <h:head>
        <title>Demo fitViewport</title>
    </h:head>
    <h:body>
        <h:form>
            <p:commandButton  type="button" onclick="PF( 'Test' ).show()" value="Show Dialog"/>

            <p:dialog id="Test" widgetVar="Test" header="Test" fitViewport="true" modal="true">
                <div style="height: 2000px">
                    Demo content
                </div>
                <f:facet name="footer">
                    <p:commandButton type="button" onclick="PF( 'Test' ).hide()" value="Hide Dialog"/>
                </f:facet>
            </p:dialog>
        </h:form>
    </h:body>
</html>
```

vonnai:
Hi I fixed issue #899. If fitViewPort is true, dialog computes height by dialog margin, titlebar, content and footer (if exists). For correct compute dialog height I had to create his clone with different css styles. I added support for responsivity. If window resize is fired, fitVieport function is called.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/dialog/dialog.js
+++ b/src/main/resources/META-INF/resources/primefaces/dialog/dialog.js
@@ -108,17 +108,10 @@ PrimeFaces.widget.Dialog = PrimeFaces.widget.BaseWidget.extend({
 
     fitViewport: function() {
         var winHeight = $(window).height(),
-        contentPadding = this.content.innerHeight() - this.content.height();
-
-        if(this.jq.innerHeight() > winHeight) {
-            this.content.height(winHeight - this.titlebar.innerHeight() - contentPadding);
-        }
-        
-        var winHeight = $(window).height(),
         dialogMargin = this.jq.outerHeight(true) - this.jq.outerHeight(),
-        titlebarHeight, contentPadding, footerHeight;
+        titlebarHeight, contentPadding, footerHeight; 
 
-        if (!this.isVisible()) {
+        if (!this.isVisible()) { 
             var dialogClone = this.jq.clone()
                 .attr("id", false)
                 .css({visibility:"hidden", display:"block", position:"absolute"});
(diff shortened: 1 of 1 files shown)


melloware:
I think this is also related to #1862 can you verify?

vonnai:
You are right

melloware (approved):

tandraschko:
Whats wrong with the current calculation of the size?

melloware:
@vonnai please feel free to correct me if I am wrong but I think the biggest things this PR does is..

1) For lazy loaded dialogs it correctly computes the height.
2) Its responsive, If content is bigger than viewport, height is correctly computed and there is scrollbar on the right side of dialog. 
3) Entire dialog is always visible and for example buttons in footer part will be always visible too.
4) If the window size changes this is responsive so it will re-calculate the dialog.

ghost:
Could also fix https://github.com/primefaces/primefaces/issues/2062 but not sure.

In which new PF version will this be included?

tandraschko:
Probably in 6.2

brunoaandrade:
@tandraschko, in version 6.2 i still have the same problem.

Issue #2217, opened 2017-03-30
When a datatable filter returns no records, the number of records in the paginator is not updated


quidam:
## 1) Environment
- PrimeFaces version: 6.1.RC2
- Does it work on the newest released PrimeFaces version? Works on 6.0.
- Does it work on the newest sources in GitHub? no.
- Application server + version: Wildfly 9.0.2 and 10.0
- Affected browsers: Firefox 52

## 2) Expected behavior

When a filter is applied to a datatable, the total number of records in the paginator should reflect the number of filtered records.

## 3) Actual behavior

When the filter returns no records, the paginator is not updated.

## 4) Steps to reproduce

Should appear in any datatable with filter and paginator, using PrimeFaces 6.1RC2 or later.

## 5) Sample XHTML

n.a.

## 6) Sample bean

n.a.

## 7) Proposed fix

According to my analysis, this issue is caused by the code
```javascript
if(args && args.totalRecords) {
```
in line 1487 of datatable.js. According to my understanding, this should check for the existence of the argument totalRecords in the arguments in the HTTP response. But in the case of no records, the value '0' is interpreted as 'false' and the paginator is not updated. Therefor I propose to change the line to
```javascript
if(args && typeof args.totalRecords !== 'undefined') {
```
I have applied the fix locally and it fixes the issue. So I have created a [Pull Request](https://github.com/primefaces/primefaces/pull/2219) for this fix.
There are three more occurrences of this problematic check within the file, which imho should be corrected, too.

melloware:
@tandraschko I think you can close this issue as a duplicate of this one.  https://github.com/primefaces/primefaces/issues/2178

PR #2219, opened 2017-03-30
Fix for issue #2217 (When a datatable filter returns no records, the …

quidam:
…number of records in the paginator is not updated)


tandraschko:
Could you also post the sample response?
I don't understand the fix exactly as 

`if (args.totalRecords)`

should be the same as

`if(typeof args.totalRecords !== 'undefined') {`

quidam:
There is a difference when args.totalRecords has a value of 0. As totalRecords is set, 
```javascript
args.totalRecords
```
should evaluate to true, but evaluates to false, so the paginator is not updated.
```javascript
typeof args.totalRecords !== 'undefined'
```
evaluates to true, whenever totalRecords is set, so the paginator always is updated, when the total number of records is available.

tandraschko:
i always wonder about javascript... :D

PR #2224, opened 2017-03-31
Call event change when selecting with the space key

fbrissi:
The onchange event is not firing when an item is selected with the space key.

I made the change so that when selecting with the space key the same behavior of change with the navigation key is done.


tandraschko:
Please create an issue with an example first.

fbrissi:
Sorry.
Is created now.

[Call event change when selecting with the space key](https://github.com/primefaces/primefaces/issues/2225)

Issue #2247, opened 2017-04-06
Wrong behavior in SelectOneRadio with non selection item


buett:
## 1) Environment
- PrimeFaces version: 6.0, 6.1.RC2, 6.1-SNAPSHOT
- Does it work on the newest released PrimeFaces version? Version? no, 6.0
- Does it work on the newest sources in GitHub? (Build by source ->  https://github.com/primefaces/primefaces/wiki/Building-From-Source): no
- Application server + version: tomcat 8.5.x
- Affected browsers: Chrome, Firefox

## 2) Expected behavior

I have a Status Enum A,B in an p:selectOneRadio. Additional I add a f:selectItem "None" for no selection. The non selection item should be selected if the value of the component is null.

## 3) Actual behavior

The non selection item is not selected if the value of the component is null.

## 4) Steps to reproduce

1. Create an enum and bind a value of the enum type to a p:selectOneRadio
2. Add f:selectItem with label "none" and no itemValue
3. Add f:selectItems bind to the values of the enum
4. set the value to null in the model
5. render page

## 5) Sample XHTML

```xhtml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
        "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml"
      xmlns:h="http://java.sun.com/jsf/html"
      xmlns:f="http://java.sun.com/jsf/core"
      xmlns:p="http://primefaces.org/ui" xmlns:ffc="http://at.finanzfenster.web.components">

<h:head>

</h:head>
<h:body>
    <h:form>
        <p:selectOneRadio value="#{sampleBean.value}">
            <f:selectItem itemLabel="None"/>
            <f:selectItems value="#{sampleBean.values}"/>
        </p:selectOneRadio>
    </h:form>
</h:body>
</html>
```

## 6) Sample bean

```java
public class SampleBean {

    public enum Status {
        A, B
    }

    private Status value;

    public Status getValue() {
        return value;
    }

    public void setValue(Status value) {
        this.value = value;
    }

    public Status[] getValues() {
        return Status.values();
    }
}
```

buett:
I could provide a patch or could trigger a pull request.

tandraschko:
a PR would be great

buett:
PR #2252

[...]

PR #2252, opened 2017-04-07
check selection in SelectOneRadioRender with null or empty values

buett:
see primefaces issue #2247


tandraschko:
@melloware Would you like to check this? You fixed something similar for SelectOneMenu

melloware:
Yes, let me investigate.

melloware (approved):
Nice work!

melloware:
@tandraschko I have looked at this change and the other one I did was a much simpler fix.  @buett fix is not only a good fix it cleans up the code greatly as it had 3 sections of duplicate code.  I give this fix the thumbs up!

PR #2257, opened 2017-04-10
Fix #1763 Timeline showCurrentTime default.

### LINKED ISSUE #1763: Timeline Component attribute showCurrentTime has no effect when value is fale
## 1) Environment
- PrimeFaces version: 6.0.4
- Does it work with the curent PrimeFaces version: No
- Payara 4.1.1.163
- Affected browsers:
  All, Not dependent on browser version
## 2) Expected behavior

That setting showCurrentDate="false" results in the current date line not being rendered.
## 3) Actual behavior

Setting showCurrentDate="false" has no effect, the current date line is still rendered.
## 4) Steps to reproduce

Take the timeline showcase example and set showCurrentDate="false", the current time bar is still rendered.
## 5) Sample XHTML

See above
## 6) Sample bean

See above

melloware:


tandraschko:
Thanks! but applied another fix

PR #2272, opened 2017-04-14
added initial .travis.yml

krichter722:
Allows usage of the [Travis CI service](https://travis-ci.org) which provides free - as in gratis - testing for FLOSS projects and integrates well with github.com. It can be used as additional CI service in order to spot issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: java
+# specifying other dists doesn't make sense because `trusty` doesn't have `oraclejdk8` (fails due to `Sorry, but JDK '[oraclejdk8]' is not known.`) which is the only supported JDK currently. Testing on Mac OSX doesn't make too much sense since it will run the same `maven` based build routine.
+
+jdk:
+- oraclejdk8
+- oraclejdk7
+- openjdk7
+# `openjdk8` not available
+
+install:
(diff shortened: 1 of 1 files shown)


tandraschko:
Pull Requests will not be applied without a issue and description.

krichter722:
Description updated.

melloware (approved):

melloware:
I like the idea of adding Travis CI builds to PF Github.   @krichter722 just create a an Issue and then update this PR to reference that ticket number.

Rapster:
JDK 5 must be set and not otherwise. Thanks @krichter722

tandraschko:
Java5 and 6 are EOL, not sure if we need this. WIll merge it.

Rapster:
EOL: couldn't agree more, but PF is still keeping a compatibility to it though. I quote from https://github.com/primefaces/primefaces/wiki/Contributing-to-Primefaces:

> All your code should compile and run in Java 5.0.

jxmai:
Travis is not activated yet just by adding the `.travis.yml`

**1. need to go to https://travis-ci.org/**

image

**2. Sign in with github**

**3. Add Repo**

image

**4. Flip the switch on**

image

@tandraschko  Only Project maintainer can switch on CI, so no pull request can help. Since we already had the `travis.yml` in place. Why not turn CI on :D ?

In the end, we should expect to see the build status next to either a commit or a pull request.

PR #2302, opened 2017-04-24
Fix #2301 Tree filter with more than 10 nodes

### LINKED ISSUE #2301: Tree filter with more than 10 nodes
From this forum thread:  https://forum.primefaces.org/viewtopic.php?f=3&t=50532

## 1) Environment
- PrimeFaces version: 6.1
- Does it work on the newest released PrimeFaces version? Version? NO
- Application server + version: JBOSS EAP 6.4
- Affected browsers: ALL

## 2) Expected behavior

Tree Filtering works for more than 10 nodes

## 3) Actual behavior

The check:

`if(rk.startsWith(rowKey) || rowKey.startsWith(rk))`

is wrong in my case there is a node that was matched with id: 0_101 and this check match also node with other ids e.g.: 0_1

I understand the logic of this check (matching parents of matched nodes) but it doesn't work in this case and in general I don't thing its a very good idea to put to many assumptions and logic into the id structure. The id should be unique and 'dumb' so it could be overridden easily by applications.

## 4) Steps to reproduce

Create a tree with many nodes and try to filter.

Suggested fix from @tefron:
```java
        if (filter && filteredRowKeys.size() > 0) {
            for (String rk : filteredRowKeys) {
                String rowKeyExt = rowKey + "_";
                String rkExt = rk + "_";
                if (rk.startsWith(rowKeyExt) || rowKey.startsWith(rkExt) || rk.equals(rowKey)) {
                    match = true;
                    if (!node.isLeaf() && !rowKey.startsWith(rk)) {
                        node.setExpanded(true);
                    }
                    break;
                }
            }
```

A PR will follow.

melloware:
Fix #2301 Tree filter with more than 10 nodes

### DIFF
--- a/src/main/java/org/primefaces/component/tree/TreeRenderer.java
+++ b/src/main/java/org/primefaces/component/tree/TreeRenderer.java
@@ -564,10 +564,12 @@ public class TreeRenderer extends CoreRenderer {
         List<String> filteredRowKeys = tree.getFilteredRowKeys();
         boolean match = false;
         if(filter && filteredRowKeys.size() > 0) {
-            for(String rk : filteredRowKeys) {
-                if(rk.startsWith(rowKey) || rowKey.startsWith(rk)) {
+            for(String filteredRowKey : filteredRowKeys) {
+                final String rowKeyExt = rowKey + "_";
+                final String filteredRowKeyExt = filteredRowKey + "_";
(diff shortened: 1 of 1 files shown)


tandraschko:
Thanks but next time without final please! ;)

melloware:
Sorry about that.  Will do for next time!

PR #2310, opened 2017-04-26
Fix #2244 Document TransferEvent

### LINKED ISSUE #2244: Document TransferEvent
Please document `org.primefaces.event.TransferEvent` (see https://www.primefaces.org/docs/api/6.0/ for the empty Javadoc).

melloware:
Fix #2244 Document TransferEvent


tandraschko:
Still no +1 from my side
will think about it but not merge it for now
If we document one event, we should document all and IMO if the variable names are good enough, we need to ducumentation. And the class usage (for which component and event) is actually documented in the userguide

melloware:
I have no problem with that.  We can close this PR and close ticket #2244 as WontFix.

PR #2353, opened 2017-05-10
Fix #2352 TriStateCheckbox not respecting valid=false.

### LINKED ISSUE #2352: TriStateCheckboxRenderer: missing styleClass "ui-state-error" for failed validation
## 1) Environment
- PrimeFaces version:
6.1

## 2) Expected behavior
TriStateCheckbox.setValid(false) results in class "ui-state-error" on rendered component

## 3) Actual behavior
valid flag is ignored

Comparing SelectBooleanCheckboxRenderer and TriStateCheckboxRenderer the latter is missing the line concerning the validation styleClass. There might be a reason for that, but I can only consider it a bug.

melloware:
Fix #2352 TriStateCheckbox not respecting valid=false. This makes it just like SelectBooleanCheckbox.


4a6f42:
LGTM, thanks!

PR #1072, opened 2016-01-27
ShowAll records option in RowsPerPageTemplate

sudheerj:
ShowAll records option in RowsPerPageTemplate

PR #2393, opened 2017-05-30
Fix #2153 - showAllOption in paginator

### LINKED ISSUE #2153: Re-Add "*" and "All Rows" paginator options
#1072 from sudheerj/ShowAllOption … was overriden by changes from #191 

Please bring back the option to show all rows from the table

lagar84:
Any non-number input, like 'All rows', '*', etc. will render Integer.MAX_VALUE - In practice, is equivalent to all rows.

### DIFF
--- a/src/main/java/org/primefaces/component/paginator/RowsPerPageDropdownRenderer.java
+++ b/src/main/java/org/primefaces/component/paginator/RowsPerPageDropdownRenderer.java
@@ -71,7 +71,13 @@ public class RowsPerPageDropdownRenderer implements PaginatorElementRenderer {
             writer.writeAttribute("autocomplete", "off", null);
 
             for( String option : options){
-                int rows = Integer.parseInt(option);
+                int rows = Integer.MAX_VALUE;
+                try{
+                    rows = Integer.parseInt(option);
+                }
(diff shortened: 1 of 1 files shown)


melloware (approved):

tandraschko:
Not sure - can't we just skip the current option if it isn't a number?

lagar84:
If we skip the option, it will be not possible to show an "All rows" option to the user. Given the context of this selectOneMenu, sounds reasonable to render all rows if it is not a number. 
This is almost the same as https://github.com/primefaces/primefaces/pull/1072

tandraschko:
I see
woudln't it better to use "uidata.getRowCount()" then?

lagar84:
I thought of using uidata.getRowCount() like #1072, but Integer.MAX_VALUE sounded like a magic number for all rows - it could be useful to test for this particular number in LazyDataModel to retrieve all rows.
I really dont have a preference for Integer.MAX_VALUE or getRowCount, any of it will be good.

Edit: If a row is added to the table, using getRowCount could miss one row? I am not sure, but I was thinking..)

tandraschko:
Thanks for the PR - did it with a IF instead Try/Catch

Issue #2397, opened 2017-05-31
p:outputLabel failed to find component in composite


AndrewShwaika:
# Failed to find EditableValueHolder in composite

CompositeUtils.invokeOnEditableValueHolder() method doesn't use value of 'targets' attribute of
editableValueHolder tag.

It should use AttachedObjectTarget.getTargets() method.

## 1) Environment
- PrimeFaces version: 6.1.1
- Application server + version: Glassfish 4.1.1, Payara 4.1.1-171
- Affected browsers: All

## 2) Expected behavior

outputLabel should find editable value holder in composite component

## 3) Actual behavior

FacesException: Cannot find editableValueHolder with name: "entry"  in composite component with id: "compositeId"

## 4) Steps to reproduce

Page with <p:outputLabel> with 'for' attribute pointed to composite

## 5) Sample XHTML
	  <p:outputLabel for="compositeEntry" value="#{msg.Label}"/>
	  <lib:entityEntry id="compositeEntry" ....

Composite:

	  <cc:interface>
	...
	    <cc:editableValueHolder name="entry" targets="ac"/>
	...
	  </cc:interface>
	  <cc:implementation>
	...
	      <p:autoComplete id="ac" ....
	...
	  <cc:implementation>

tandraschko:
Does the following case still work:
https://github.com/primefaces/primefaces/issues/1633
? We need to support both cases, the user must NOT define a targets atribute.

AndrewShwaika:
It should work.
Implementation of AttachedObjectTarget.getTargets() method in mojarra and myfaces handles both cases.

tandraschko:
@Rapster Could you please check that, too?

I'm not sure if we should support it as "name" targets a single component and "targets" targets multiple components.
We can only support a single component.

AndrewShwaika:
Mojarra 2.2.9
com.sun.faces.facelets.tag.composite.AttachedObjectTargetImpl.getTargets()
MyFaces 2.2.11
org.apache.myfaces.view.facelets.tag.composite.AttachedObjectTargetImpl.getTargets()

Both implementations first look at 'targets' and if it empty use 'name' for component lookup

Rapster:
Here is description of `name` attribute:

> The value of this attribute maps back to the "for" attribute on an attachable object nested within a composite component. If the "targets" attribute is not specified, this value also represents the component ID of the target component within the that the <composite:implementation> ActionListener should be mapped to.

[...]

PR #2398, opened 2017-05-31
#2397 Use AttachedObjectTarget.target.getTargets()

AndrewShwaika:
Use AttachedObjectTarget.target.getTargets()


tandraschko:
Thanks for the PR but will close it for now like mentioned in the ticket. sorry.

PR #2408, opened 2017-06-02
PF-2407: remove autocomplete attribute from hidden inputs because of …

OBreidenbach:
…w3c validator


melloware (approved):
Changes look good to me as long as their was no PF reason for having these autocomplete attributes on there in the first place.

tandraschko:
@mertsincan @cagataycivici do you remember why we added this? AFAIK there was a reason but can't remember.

mertsincan:
Hi @tandraschko,

We need to keep autocomplete on hidden elements as it causes issues on firefox. Exp;
on Basic AccordionPanel demo;
- select the second header(Godfather Part II)
- refresh page
- please see the value of hidden input on DOM;

Screenshot without autocomplete="off";
without autocomplete_off

You can see the same behavior in other components(without autocomplete="off")

tandraschko:
Yep, ok. See we should just close it as won't fix?

Issue #2383, opened 2017-05-23
p:textEditor does not support p:ajax


brainlag:
## 1) Environment
- PrimeFaces version: 
  6.1.0
- Does it work on the newest released PrimeFaces version? Version?  
  no
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) 
  no
- Application server + version: all
- Affected browsers: all

## 2) Expected behavior
p:ajax inside p:textEditor works like p:inputTextarea

## 3) Actual behavior
page does not even render because javax.faces.component.behavior.ClientBehaviorHolder is not implemented

## 4) Steps to reproduce
Just add a p:ajax tag inside a p:textEditor tag

Since p:textEditor uses contenteditable instead of a input field there where some shortcuts made by not implementing p:ajax support. But nevertheless, people want to replace textareas with it and expect it to behave like a texarea (p:inputTextArea).

I tried to resolve this problem myself and send a PR but I don't know how to hook up the javascripts events with p:ajax. Some guidance would be required.

tandraschko:
Just compare with p:editor. I think there is something like #encodeClientBehaviors available in the renderers.

brainlag:
Nope. p:editor doesn't support p:ajax either.

tandraschko:
then compare with inputText or something ;)

Rapster:
@brainlag and feel free to provide a PR ;)

PR #2429, opened 2017-06-11
PF-2383 Add p:ajax support to p:textEditor

brainlag:
This hooks up all basic events for p:ajax inside a p:textEditor. See #2383.

### DIFF
--- a/src/main/java-templates/org/primefaces/component/texteditor/TextEditorTemplate.java
+++ b/src/main/java-templates/org/primefaces/component/texteditor/TextEditorTemplate.java
@@ -1,0 +1,17 @@
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import javax.faces.event.AjaxBehaviorEvent;
+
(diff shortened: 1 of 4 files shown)


melloware (approved):

tandraschko:
I'm not sure if that much JS is really required.
If you check InputText e.g. than there is no additional JS required for the event handling.
The reason for that is, that the events (e.g. oninput) is directly rendered on the input element instead of passing it to script as behavior script.
Isn't that possible with qill, too?

brainlag:
I didn't know how this should work and this worked for me. But after looking up your suggestion I think I could get ride of the javascript for the common events. But blur, focus, change and select will need javascript.

tandraschko:
why? i think these are common JS events - only select will be a custom event?
I would just try to add onblur/onfocus/onchange on the quill input and check if it works automatically

brainlag:
Quill does not use a input element, it uses a div with contenteditable which doesn't have native focus events.

tandraschko:
Ahh i see, yep, than it's ok.

tandraschko:
I think we can leave your PR as it it - Quill seems a little bit special, so it's ok IMO.
Will apply it later :)

mauromol:
I'd suggest you add this to the documentation. Thank you!

Issue #1290, opened 2016-04-01
dialog framework dialogs without height not centered by default


agoetze:
As the title says. You can easily verify this problem by looking at the current show case:
http://www.primefaces.org/showcase/ui/df/basic.xhtml

Button "View": no height set -> dialog not centered (vertically)
Button "Options": height set to 640 -> dialog centered (vertically)

Tested with Firefox 45 under Ubuntu and with Internet Explorer 11 under Windows 7.

Expected default behaviour: Dialog centered (horizontally and vertically)

sergi-mm:
We also experiment the same issue using Primefaces 5.3.14 (latest version of 5.3 at the moment of writting).

jaymesanches:
Is it not working in primefaces 6.0 yet?

smckeetx:
Using PF 6.1 the dialog is centered horizontally but not vertically. This is a problem.

Maikel-Nait:
BTW, I fixed it myself by overriding the original PrimeFaces.dialog.DialogHandler.

The original code in PF 6.1 has these two lines of code BEFORE a block that starts with the comment //adjust height

`dialogFrame.data('initialized', true);`
`rootWindow.PF(dialogWidgetVar).show();`
`//adjust height`

By just swapping these two lines AFTER the code that adjusts the height, the dialogs can be centered both horizontally and vertically as expected:


`//adjust height`
`var frameHeight = null;`
`if(cfg.options.contentHeight)`
`         frameHeight = cfg.options.contentHeight;`
`else`
`         frameHeight = $frame.get(0).contentWindow.document.body.scrollHeight + `
`                                  (PrimeFaces.env.browser.webkit ? 5 : 25);`
`$frame.css('height', frameHeight);`

Moved lines to the end of this code block:

`dialogFrame.data('initialized', true);`
`rootWindow.PF(dialogWidgetVar).show();`

So far it works fine in Chrome, Firefox and IE11 .... maybe this could be an easy fix for this problem ??

Rapster:
You can write a PR so @tandraschko or @mertsincan  can have a look and merge into PF master branch

Maikel-Nait:
Done. Created PR here:  [https://github.com/primefaces/primefaces/pull/2463](https://github.com/primefaces/primefaces/pull/2463)

PR #2463, opened 2017-06-22
Fix #1393 and #1290 - dialog framework : dialogs without height not centered by default

### LINKED ISSUE #1393: dialog framework does not work well in 6.0RC3
We updated to 6.0 and there is a problem in DialogFramework.
The dialog'position seems default on the center of screen,even when the dialog is lagrer than the base page that we have to move it.
ws000001
But it works good in 5.3.
ws000002

Maikel-Nait:
In PF 5.x , Dialog Framework was able to open centered dialog windows both horizontally and vertically, without the need to specify width or height attributes. In PF 6.x this stopped working. The solution seems to be as simple as moving a pair of javascript lines in [**core.dialog.js**], from their new location in PF 6.x, to their original location in PF 5.x. Tested with IE 11, Chrome and Firefox, and apparently working...

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/core/core.dialog.js
+++ b/src/main/resources/META-INF/resources/primefaces/core/core.dialog.js
@@ -111,10 +111,6 @@ if (!PrimeFaces.dialog) {
                     dialogFrame.attr('title', title.text());
                 }
 
-                dialogFrame.data('initialized', true);
-
-                rootWindow.PF(dialogWidgetVar).show();
-
                 //adjust height
                 var frameHeight = null;
                 if(cfg.options.contentHeight)
@@ -123,6 +119,10 @@ if (!PrimeFaces.dialog) {
                     frameHeight = $frame.get(0).contentWindow.document.body.scrollHeight + (PrimeFaces.env.browser.webkit ? 5 : 25);
 
                 $frame.css('height', frameHeight);
+                
+                // fix #1290 - dialogs are not centered vertically
+                dialogFrame.data('initialized', true);
+                rootWindow.PF(dialogWidgetVar).show();
             })
             .attr('src', frameURL);
         },


tandraschko:
Not sure if this is correct. 
If you check the current file in line 123, it seems that the frameHeight is calculated.
I think it won't work if the dialog is not visible. Therefor the dialog is openend and after that the frameheight is calculated.

Maikel-Nait:
This fix is actually for opening a dialog window with external context, as explained in the showcase here: [http://primefaces.org/showcase/ui/df/basic.xhtml](http://primefaces.org/showcase/ui/df/basic.xhtml) 

This change should not affect using the dialog overlay component, which is already working fine (and it's hidden by default)

Am I right ?

tandraschko:
No idea how DF works in detail but the current flow is this:
1) opening the dialog
2) calculate the frame height (line 123)

If you switch those 2 statements, i fear that the frameheight can not be calculated for every case correctly

As i don't have time currently for testing every case with every browser, i can't just apply it.

Maikel-Nait:
No problem. I will have more tests on each browser, both for external dialogs and dialog overlay component, hidden or not. If I don't find any problem I will let you know.

tandraschko:
any news?
@melloware @Rapster If anyone of you has time, could you check this? Currently -1 for this PR because i'm sure it will break something, like stated above.

melloware:
I don't use the DialogFramework so I have the same concerns that it might break something.  Maybe we add this PR after 6.2 goes live and that will give users a long time in 6.3-SNAPSHOT to test it out.

Maikel-Nait:
Hi all. I haven’t got time to do further testing with this change. It would be definitely much better if you could look into it... at the end, the original problem is reproducible with the showcase of the dialog framework, where it is clearly visible that the dialog is not centered.

jxmai:
I have briefly taken a look at this ticket, and I can assume the changes in https://github.com/primefaces/primefaces/commit/4bfd41751ad5ed7a4c33891a2ced9f278fe8a945 brought some side effect to both https://github.com/primefaces/primefaces/issues/1393 and https://github.com/primefaces/primefaces/issues/1290 if you look at the timeline and the evidence provided by the reporter.

> The solution seems to be as simple as moving a pair of javascript lines in [core.dialog.js], from their new location in PF 6.x, to their original location in PF 5.x. Tested with IE 11, Chrome and Firefox, and apparently working...


Also, the height improvement of https://github.com/primefaces/primefaces/commit/4bfd41751ad5ed7a4c33891a2ced9f278fe8a945 is unclear since there is not much detail attached :disappointed: .

Commit was made on ` 28 Jan 2016`

#1290  was reported on  `1 Apr 2016`

#1393 was reported on ` 9 May 2016`

jxmai:
Well, I support @Maikel-Nait  and I vote for changing the code back to where it was, since the only improvement I could see in https://github.com/primefaces/primefaces/commit/4bfd41751ad5ed7a4c33891a2ced9f278fe8a945 was the frameheight calculation logic:

https://github.com/primefaces/primefaces/commit/4bfd41751ad5ed7a4c33891a2ced9f278fe8a945#diff-766244ee4b37dc9686170a06033a2ea3R117

---
> No idea how DF works in detail but the current flow is this:

> opening the dialog
> calculate the frame height (line 123)


That's the problem. The current workflow looks really weird after https://github.com/primefaces/primefaces/commit/4bfd41751ad5ed7a4c33891a2ced9f278fe8a945. Normally, we should calculate the frame height before showing the dialog, not the other way around.

jxmai (approved):

tandraschko:
How does frame height calculation works if the dialog is not visible yet?

jxmai:
> How does frame height calculation works if the dialog is not visible yet?

That's a good question. I may need to look into it further. I guess it was in DOM somewhere already.

jxmai:
I understand what's going on, and please merge this PR. 

> How does frame height calculation works if the dialog is not visible yet?

The reason is the dialog has already been initialized long before it becomes visible.

image

Basically, the following code is responsible for initialzing the dialog content

https://github.com/primefaces/primefaces/blob/fc4c6e655e033502c4157c5303f311b2a64d9051/src/main/resources/META-INF/resources/primefaces/core/core.dialog.js#L62-L99

while the code we are reviewing is below the dialog initialization

https://github.com/primefaces/primefaces/blob/fc4c6e655e033502c4157c5303f311b2a64d9051/src/main/resources/META-INF/resources/primefaces/core/core.dialog.js#L114-L125

---

Currently if we resize the content before setting the height. Here is what is happening


image


apply height by   `$frame.css('height', frameHeight);`

image


Notice that the height increases based on the css `height` while the header stays at the same position, so this problem will be even worse if user has a large dialog like the one in #1393 

We should flip the order of dialog showing and height calculation just like this PR and everything should work correctly and nothing will be broken.

This PR will fix: #1393 and #1290 as @Maikel-Nait described.

tandraschko:
Thanks, i appreciate your analysis!

Maikel-Nait:
It feels really good to have contributed to your fantastic library ...   :-)
Keep up the good work !

Issue #2465, opened 2017-06-22
inputNumber: clearing value doesn't function with browser Edge


movalz:
## 1) Environment
- PrimeFaces version: 6.1.2
- Does it work on the newest released PrimeFaces version? Version? Yes (6.1.2)
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source): Yes
- Application server + version: Tomcat 7
- Affected browsers: Edge

## 2) Expected behavior
When user clears input of an p:inputNumber field (e.g. by marking it and pressing "delete" key) and triggers a request which updates the inputNumber field the field should stay empty.

## 3) Actual behavior
at the end of the request which updates the inputNumber field the field shows the last value before user has cleared the field.

## 4) Steps to reproduce
any scenario as I described in 2) 

## 5) Sample XHTML
<p:inputNumber id="age" value="#{myBean.age}" onchange="triggerAjaxRequest(component.clientId);" />

## 6) Reason
I have debugged and found following reason:
In function InputNumber.setValueToHiddenInput() following code is contained:
```
        if (value !== "") {
            this.hiddenInput.attr('value', value);
        } else {
            this.hiddenInput.removeAttr('value');
        }

```
Browser Edge doesn't works as expected when calling "this.hiddenInput.removeAttr('value')": it does nothing. The issue would be fixed if always
`this.hiddenInput.attr('value', value);`
would be called (or at least in case of Edge browser).

tandraschko:
Please provide a PR!

melloware:
@mertsincan please include this in 6.1.3 thank you!

PR #2466, opened 2017-06-22
#2465: inputNumber: clearing value doesn't function with browser Edge

movalz:
Instead of changing condition to "if (value !== "" || isEdge) {..." it may be ok to always use "this.hiddenInput.attr('value', value);"?


tandraschko:
Please change to always use "this.hiddenInput.attr('value', value);"
I think this is fine in every browser, i just didn't change that code part when i refactored it.

PR #2468, opened 2017-06-22
fix JSF 2.0 incompatible use of enum

sorontur:
The SKIP_ITERATION Enum is introduced in JSF 2.1
the fix was also applied on the UIData Component some years before. See
org.primefaces.component.api.UIData.shouldVisitRows(FacesContext,
VisitContext)


tandraschko:
Please create an issue first - i will merge it then.

sorontur:
Sure there you are: #2469

PR #2485, opened 2017-06-29
#bug2484 Fix duplicates removal

AndrewShwaika:


tandraschko:
Does the "[id='" + this.id + "']" selector works without escaping the ":" from the jsf generated ids?
For the ID selector, we need escaping (see PrimeFaces.escapeClientId or similar)

AndrewShwaika:
Yes, it works well without escaping the ":".
Syntax as ID selector is not suitable for attribute matching.

AndrewShwaika:
Actually it is not that easy.
ID selector for simple identifiers (without ':') uses getElementById() that returns only one element.
ID selector for JSF generated ids with ":", fails regexp matching as ID selector and fallback to querySelectorAll() that can return more than one element.
So I need to figure out how to handle both cases.

tandraschko:
So our current code works fine if there is a ":" inside the id?

tandraschko:
(we should also check the dialog code, we do something similar there)

AndrewShwaika:
I would like to propose mixed solution:
If this.jq.length > 1 then it was jsf generated identifier and should be handled by current code.
But we should handle simple identifier as proposed.

tandraschko:
Could you please add comments why we do this?

AndrewShwaika:
In case:
`<h:form prependId="false"...>
<p:menu overlay="true"..>`
Overlay panel will have simple id without ':'. It will cause duplicates of these panels at the bottom of the body after ajax updates.

tandraschko:
So there is still a bug? If it should be the answer to my last answer: it would be great if you could a small "code"-comment :)

Issue #2487, opened 2017-06-30
TreeTable Filter with Paginator breaks due to JSON parse failure


gako:
## 1) Environment
- PrimeFaces version: 6.1.2
- Application server + version: Glassfish 4.1.1 (Payara 4.1.164)
- Affected browsers: Firefox, Chrome, ...

## 2) Expected behavior

I filter the tree table with a value and the json response is validated correctly
...

## 3) Actual behavior

The return partial response after filtering cannot be parsed by the browser due to json problem

## 4) Steps to reproduce

Create a TreeTable that has both Feature Filtering and Paginator enabled at once.

FIX

In the TreeTableRendered class line ~917 selection callbackparameter is added even if it is *null*

Wrong:
```
RequestContext requestContext = RequestContext.getCurrentInstance();
requestContext.addCallbackParam("selection", tt.getSelectedRowKeysAsString());
```

To fix this a check wether getSelectedRowKeysAsString !=null should be added like in other places (datatable, ...)

Right:
```
RequestContext requestContext = RequestContext.getCurrentInstance();
if (tt.getSelectedRowKeysAsString() !=null) {
requestContext.addCallbackParam("selection", tt.getSelectedRowKeysAsString());
}
```

The Problem with the null value is that the PrimePartialResponseWriter creates a empty json object for a null value but still adds a "," So for the two request parameters of the treetable (totalRecords=0,selection=null) this results e.g. in

{"totalRecords":0**,**}

but it should be {"totalRecords":0} and (selection=null) should NOT be part of callbackparameters.

tandraschko:
A PR would be great!

gako:
PR: https://github.com/primefaces/primefaces/pull/2488

PS.: hope i did everything right, it's my first one :-)

PR #2488, opened 2017-06-30
TreeTableRenderer fix for Sort&Paginator

gako:
PR for issue: https://github.com/primefaces/primefaces/issues/2487

TreeTable Renderer fix for Callbackparam in sort


tandraschko:
Thanks for the PR - will do a small refactoring for performance as getSelectedRowKeysAsString is called 2 times currently.

Issue #2492, opened 2017-07-03
DataTable: Column groupRow based on contents of another column


normanheinrich:
## 1) Environment
- PrimeFaces version: 6.1
- Does it work on the newest released PrimeFaces version? No, its a new feature
- Application server + version: Tomcat 8
- Affected browsers: All

## 2) Expected behavior
If i set groupRow = "true" on a column i'd like to specify another column, 
so that the resulting rowspan is limited to the rowspan computed based on the contents of the referenced column.
I suggest a new attribute 'groupRowBase', that is set to the id of the column that should be used 
as reference to limit the rowspan.

If you don't set the attribute, the columns own content is used to compute grouping (default)

Please suggest a better attribute name if you don't like it...PR will follow

Sample: groupRowBase is set to the first column on columns 2, 6 and 7
![image](https://user-images.githubusercontent.com/29800605/27781786-de0ed7a0-5fcf-11e7-9812-39e764fbb67b.png)

## 3) Actual behavior
Grouping of the column is computed based on its own content.
![image](https://user-images.githubusercontent.com/29800605/27781768-cf6a3a82-5fcf-11e7-8e17-c2e6c321ed41.png)

## 5) Sample XHTML
```
<p:dataTable id="data">
	<p:column id="base" headerText="Accountnumber" groupRow="true">
		<h:outputText value="#{bean.accountnumber}" />
	</p:column>
	<p:column headerText="Accounttype" groupRow="true" groupRowBase="base">
		<h:outputText	value="#{bean.accounttype}" />
	</p:column>
	<p:column headerText="Schuldner">
		<h:outputText value="#{pc.getSchuldnerName(bean)}" />
	</p:column>
	<p:column headerText="Hauptforderung" groupRow="true" groupRowBase="base">
		<h:outputText value="#{bean.forderung.summen.hauptforderung}" />
	</p:column>
</p:dataTable>
```

## 6) Sample bean

..

melloware:
Original Poster never returned with a PR and no one else in 3 years has asked for this kind of feature.

xtsoler:
In your example:
<p:column headerText="Accounttype" groupRow="true" groupRowBase="base">
		<h:outputText	value="#{bean.accounttype}" />
</p:column>

just add after the output text a div, with style transparency 0 and the value of the other column like this:
<p:column headerText="Accounttype" groupRow="true" groupRowBase="base">
[...]

PR #2493, opened 2017-07-03
#2492 Column groupRow based on contents of another column

normanheinrich:
If i set groupRow = "true" on a column i'd like to specify another column,
so that grouping is computed based on the content of that column, not the grouped column itself.
I suggest a new attribute 'groupRowBase', that is set to the id of the column that should be used
to compute grouping


normanheinrich:
It turned out i need a more sophisticated way to compute the grouping (rowspan). I still need to use the contents of the actual column because the contents could differ, but rowspan should be limited to the rowspan of the referenced column.
Will close this and create another PR.

Issue #1995, opened 2017-01-02
Timeline events displayed on wrong group after zoom


PlassJ:
## 1) Environment
- PrimeFaces version: 6.0.13
- Does it work on the newest PrimeFaces version? No
- Application server + version: WildFly 10.1.0
- Affected browsers: all

## 2) Expected behavior
TimelineEvents should stay in there group after zoom.

## 3) Actual behavior
TimelineEvents move to an other group after zoom.

## 4) Steps to reproduce
Showcase -> timeline -> custom 
Point mouse cursor right bevor an event (User 2, red) zoom in. (img 1)
Zoom in until events "left" the viewport. (img 2)
After some zooming the event (from User 3) moves "up". (img 3)

## 5) Sample XHTML
See showcase.

## 6) Sample bean
See showcase.

![timeline1](https://cloud.githubusercontent.com/assets/13958943/21589469/5da03864-d0f2-11e6-9817-1096ad388876.png)
![timeline2](https://cloud.githubusercontent.com/assets/13958943/21589468/5da02856-d0f2-11e6-8a10-634ef73e3884.png)
![timeline3](https://cloud.githubusercontent.com/assets/13958943/21589467/5d9fe490-d0f2-11e6-8023-981458649ddb.png)

PlassJ:
added possible fix  #2006

mertsincan:
Thanks for your PR.

PR #2505, opened 2017-07-06
Timeline events displayed on wrong group after zoom (timeAxis not on top)

PlassJ:
Unfortunately only fixed the case with timeAxis on top (#1995 ).
Now fixing the other case xD


tandraschko:
JFYI:
if you don't provide a PR directly in the timeline project, we will probably override your fix in future versions when we update the script.
So please also do it in their side.

PlassJ:
Done. 
We'll see what happens.
https://github.com/almende/chap-links-library/pull/368

tandraschko:
Ahh, i see
its not in development anymore - maybe we should check if we upgrade to the sucessor.

PR #2508, opened 2017-07-06
Fix #2506

### LINKED ISSUE #2506: Galleria misses components.css resource dependency
## 1) Environment
- PrimeFaces version: **6.1**

## 2) Actual behavior

`primefaces/src/main/resources-maven-jsf/ui/galleria.xml` does not contain a resource dependency to `components.css`. This results in the `components.css` not being included when no other components are used on the same view that bring in this dependency. A missing components.css results in wrong rendering of the galleria since the `ui-helper-hidden` class is not available.

Mobe91:
This adds a resource dependency for components.css to galleria.

### DIFF
--- a/src/main/resources-maven-jsf/ui/galleria.xml
+++ b/src/main/resources-maven-jsf/ui/galleria.xml
@@ -118,6 +118,9 @@
 	</attributes>
 	<resources>
         <resource>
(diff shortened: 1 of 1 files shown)


tandraschko:
Thanks!

PR #2523, opened 2017-07-11
Fix #2522

### LINKED ISSUE #2522: Avoid unnecessarily computing fileUpload's allowTypes twice
Method `encodeScript()` on class `org.primefaces.component.fileupload.FileUploadRenderer` shouldn't call call `fileUpload.getAllowTypes()` twice. It could use a temporary variable.

Around line 100:

```
if (fileUpload.getAllowTypes() != null) {
       wb.append(",allowTypes:").append(fileUpload.getAllowTypes());
}
```
In my application, the database is queried to determine the "_allowed type_", what makes this double call a little cumbersome.
## 1) Environment
- PrimeFaces version: 6.0
- Application server + version: Any
- Affected browsers: Any
## 2) Expected behavior
The value of the expression for the "'allowTypes''" attribute of the '<p:fileUpload>' component 
should be computed only once per rendering.
...
## 3) Actual behavior
The value of the expression for the "'allowTypes''" attribute of the '<p:fileUpload>' component is computed twice.
..

## 4) Steps to reproduce
Render a page with a  '<p:fileUpload>' component.
.. 
## 5) Sample XHTML
```
    <p:fileUpload value="#{fileUploadView.file}" allowTypes="#{fileUploadView.allowTypes}"/>
</h:form>
```
## 6) Sample bean
```
package org.primefaces.showcase.view.file;
 
import org.primefaces.event.FileUploadEvent;
import org.primefaces.model.UploadedFile;
 
@ManagedBean
public class FileUploadView {
     
    private UploadedFile file;
 
    public UploadedFile getFile() {
        return file;
    }
 
    public void setFile(UploadedFile file) {
        this.file = file;
    }
    public String getAllowTypes(){
         String allowTypes = "(txt|pdf)"//actually, do heavy computation / IO
         return allowTypes;
    }
}
```
..

brunoabdon:
Pull request with a simple fix for enhancement proposed on #2522.

### DIFF
--- a/src/main/java/org/primefaces/component/fileupload/FileUploadRenderer.java
+++ b/src/main/java/org/primefaces/component/fileupload/FileUploadRenderer.java
@@ -100,7 +100,7 @@ public class FileUploadRenderer extends CoreRenderer {
                 .callback("onerror", "function()", fileUpload.getOnerror())
                 .callback("oncomplete", "function(args)", fileUpload.getOncomplete());
 
-            final String allowTypes = fileUpload.getAllowTypes();
+            String allowTypes = fileUpload.getAllowTypes();
 
             if (allowTypes != null) {
                 wb.append(",allowTypes:").append(allowTypes);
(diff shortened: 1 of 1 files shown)


tandraschko:
If you remove the "final", i will merge it ;) Thanks!

PR #2540, opened 2017-07-14
Fix #1036 SelectOneMenu: new customContent attribute for overriding d…

### LINKED ISSUE #1036: Improve SelectOneMenuRenderer for String objects (List<String> etc.)
Please make the `SelectOneMenuRenderer` work for a collection of `String` objects to make it work for custom content.

Please see: http://stackoverflow.com/questions/34656127/primefaces-selectonemenu-with-fontawesome-icons-custom-fselectitems/34813602?noredirect=1#comment57372040_34813602 
http://stackoverflow.com/questions/32735770/pselectonemenu-doesnt-render-custom-content-via-pcolumn-on-liststring

This can be achieved by introducing another attribute `customContent="true|false"` on the <p:selectOneMenu> and by getting rid of the `instanceOf` comparison in the `encodePanel(...)` method in the `SelectOneMenuRenderer` class. If `customContent` is set to true, the `encodeOptionsAsTable(...)` method will call `renderChildren(...)` for every `<p:column>` no matter if the `SelectItem` is an `instanceOf` `String` or not.

See also: https://github.com/primefaces/primefaces/blob/5_3/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java#L208
https://github.com/primefaces/primefaces/blob/5_3/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java#L249
https://github.com/primefaces/primefaces/blob/5_3/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java#L272

melloware:
- Added new property called customContent which defaults to FALSE.
- Switch if statement to if customContent=true attempt the custom content rendering no matter what, else do the default rendering behavior.

### DIFF
--- a/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
+++ b/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
@@ -268,7 +268,7 @@ public class SelectOneMenuRenderer extends SelectOneRenderer {
     
     protected void encodePanelContent(FacesContext context, SelectOneMenu menu, List<SelectItem> selectItems) throws IOException {
         ResponseWriter writer = context.getResponseWriter();
-        boolean customContent = menu.getVar() != null;
+        boolean customContent = menu.isCustomContent();
         
         if(customContent) {
             writer.startElement("table", menu);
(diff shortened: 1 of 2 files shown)


tandraschko:
the PR looks fine but could you please resolve the conflicts?

melloware:
Yep sorry this was also done before the checkstyle stuff. I will fix.

melloware:
Submitted a different PR so closing this one.

PR #2542, opened 2017-07-16
Fix #2536 - don't skip defaultCommand on enter key press if target is…

### LINKED ISSUE #2536: DefaultCommand not triggered if scope equals event.target
Tested with PrimeFaces 6.0 and 6.1.

Expected: 
Scope is set to inputText, Target is set to a commandLink. When I hit enter inside the inputText the commandLink should be executed.

Actual:
Nothing happens. I have to set the scope to the parent widget of the inputText (in my case a p:panel).

The root cause: 
Line 24 of forms.defaultcommand.js as it uses $.find to test if the event.target is a child of the scope. It never tests if event.target is the scope itself.

```($this.scope && $this.scope.find(e.target).length == 0)``` should be changed to something like ```($this.scope && $this.scope.find(e.target).length == 0 && this.scope.get(0) != e.target)```

jxmai:
… the same as scope

based on the idea from #2536 in the bug description.

>The root cause:
Line 24 of forms.defaultcommand.js as it uses $.find to test if the event.target is a child of the scope. It never tests if event.target is the scope itself.

> ($this.scope && $this.scope.find(e.target).length == 0) should be changed to something like ($this.scope && $this.scope.find(e.target).length == 0 && this.scope.get(0) != e.target)

replicated and tested locally:

image


tandraschko:
Thanks for the PR! Looks fine.

PR #2566, opened 2017-07-23
Ignore special keys on dataTable filters

kdaemonv:
Using key combinations to switch tasks (Alt+Tab) or to go back to filter's input from another input element (Shift+Tab) triggers dataTable filtering - that's an extra ajax request. And there are lots of other special keys which can trigger filtering when not needed (Enter and F* keys intentionally left unignored as some could use it to force filtering).


tandraschko:
Please create a issue first!

kdaemonv:
Created an issue #2587

PR #2651, opened 2017-08-21
Fix #2650 - Datatable: duplicate cliend ids in custom facet paginator

### LINKED ISSUE #2650: DataTable: duplicate cliend ids in custom facet paginator
## 1) Environment
- PrimeFaces version: 6.2-SNAPSHOT
## 2) Expected behavior
As an example, menu button is used in facet, and as a result, menuButton should display menu items

## 3) Actual behavior
MenuButton is displayed, but does no show up menu items

## 4) Steps to reproduce

1. Using 5/ in showcase in http://localhost:8080/showcase/ui/data/datatable/paginator.xhtml
2. try to click on menu button
3. menu items does not show up

## 5) Sample XHTML
```xml
            <p:dataTable var="car" value="#{dtPaginatorView.cars}" rows="10" 
                         paginator="true"
                         paginatorTemplate="{CurrentPageReport} {FirstPageLink} {PreviousPageLink} {PageLinks} {NextPageLink} {LastPageLink} {RowsPerPageDropdown} {Custom}"

                 <f:facet name="{Custom}">
                    <p:menuButton value="Options">
                        <p:menuitem value="Save" actionListener="#{menuView.save}" icon="ui-icon-disk"/>
                        <p:menuitem value="Update" actionListener="#{menuView.update}" icon="ui-icon-arrowrefresh-1-w"/>
                        <p:menuitem value="Delete" actionListener="#{menuView.delete}" ajax="false" icon="ui-icon-close"/>
                        <p:separator />
                        <p:menuitem value="Homepage" url="http://www.primefaces.org" icon="ui-icon-extlink"/>
                    </p:menuButton>
                </f:facet>

                <p:column headerText="Id">
                </p:column>

                <p:column headerText="Year">
                </p:column>

                <p:column headerText="Brand">
                </p:column>

                <p:column headerText="Color">
                </p:column>
            </p:dataTable>
```

## 6) Analysis
clientIds generated of both menubutton are the same... So I guess, the jQuery widget can't work properly...

Rapster:
I realized it is necessary to add the facet name in the id otherwise, it might still have id conlicts

### DIFF
--- a/src/main/java/org/primefaces/component/api/Pageable.java
+++ b/src/main/java/org/primefaces/component/api/Pageable.java
@@ -20,6 +20,8 @@ import javax.faces.context.FacesContext;
 
 public interface Pageable {
 
+    public String getId();
+
     public String getClientId(FacesContext context);
 
     public String getPaginatorPosition();
--- a/src/main/java/org/primefaces/renderkit/DataRenderer.java
+++ b/src/main/java/org/primefaces/renderkit/DataRenderer.java
@@ -96,7 +96,8 @@ public class DataRenderer extends CoreRenderer {
             else {
                 UIComponent elementFacet = pageable.getFacet(element);
                 if (elementFacet != null) {
-                    elementFacet.setId(pageable.getId() + "_" + position);
+                    String facetName = element.replaceAll("[\\W+]", "_").toLowerCase();
+                    elementFacet.setId(pageable.getId() + facetName + position);
                     elementFacet.encodeAll(context);
                 }
                 else {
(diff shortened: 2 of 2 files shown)


Rapster:
i'm wondering if I shouldn't use encodeFacet instead...

tandraschko:
I really don't like the fix and i'm not sure if it would have side effects.
The problem probably is, that we should really avoid to render a facet multiple times.

Rapster:
The only thing I don't like about this is that I'm using a regex, while there should probably be an util for that kind of thing... I guess then the support of facet should be done in header and not pagination. That would be a big breaking change...

But I think it still deserves some attention to make it work in the current state, I mean we're so close to get something working...

Rapster:
As this PR does not cover composite components, I prefer to close it. As a workaround, we only use top paginator (instead of both)

PR #2665, opened 2017-08-28
Fix #2664 - PanelGrid: default columns value

### LINKED ISSUE #2664: PanelGrid: default columns value
Default value should be set to one

Rapster:


tandraschko:
Not sure if this is correct. If you check the renderer, we have there a logic if "column =< 0"

Rapster:
Yes it can happen if the columns is being set dynamically, but by default I think 1 should be the default value

tandraschko:
I think we should deeper analyse the JSF impl and align it with it.

PR #2676, opened 2017-08-30
Use converter on attribute itemLabel when creating SelectItem

xqvier:
When using dropList the EL used in the itemLabel attribute of selectItems might be an Object where the String.valueOf(object) is not relevant.
This fix makes use of any javax.faces.convert.Converter declared for the object referenced so it can call the appropriate getAsString method to display it correctly.

I've been using it in one project and it's working well but i'm open to discuss of this implementation.

### DIFF
--- a/src/main/java/org/primefaces/renderkit/InputRenderer.java
+++ b/src/main/java/org/primefaces/renderkit/InputRenderer.java
@@ -25,6 +25,7 @@ import java.util.RandomAccess;
 import javax.el.ELException;
 import javax.el.ExpressionFactory;
 import javax.el.ValueExpression;
+import javax.faces.application.Application;
 import javax.faces.component.*;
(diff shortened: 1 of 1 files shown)


Rapster:
Looks okay, but you can refactor your code so it gets twice more easy to read ;-)

jxmai:
also PF dislike `final` keyword a lot, and it needs an issue linked to a pull request for documentation purposes

melloware:
Also make sure you do a "mvn clean package" to  make sure it passes all the new CheckStyle style requirements or else it will get rejected.

tandraschko:
Does it work the same in the JSF impl with h:* components?

tandraschko:
Will close it for now as no response. Please create a issue first, when can discuss it there.

PR #2689, opened 2017-09-04
Fix #2512 Update database.js

### LINKED ISSUE #2512: DataTable: validate cell editing values
## 1) Environment
- PrimeFaces version: 5.3
- Does it work on the newest released PrimeFaces version: no
- Application server + version: Wilfly 10 Final
- Affected browsers: Chrome & Firefox

## 2) Expected behavior
When adding a new row dynamically in a p:datatable, jsf validators such as "required", should be executed when the cell editor loses focus.

## 3) Actual behavior
The validation only kicks in, when you actually write something in the input text, the deletes it.

## 4) Steps to reproduce
add new row, select the cell editor, when you leave the input text, it should be validated.
in order for the validation to work, you have to write something in the input text, delete it and leave the input text.

## 5) Sample XHTML
```
<f:view>
    <p:growl autoUpdate="true"/>
    <h:form>
        <p:commandButton value="add row" update="table" action="#{myBean.addRow}"/>

        <p:dataTable value="#{myBean.users}" var="var" id="table" editMode="cell" editable="true">
            <p:column headerText="Quantity">
                <p:cellEditor>
                    <f:facet name="output">
                        <h:outputText value="#{var.name}"/>
                    </f:facet>
                    <f:facet name="input">
                        <p:inputText value="#{var.name}" required="true">
                            <f:validateRequired />
                        </p:inputText>
                    </f:facet>
                </p:cellEditor>
            </p:column>
        </p:dataTable>

        <p:commandButton value="submit" action="#{myBean.submit}"/>
    </h:form>
</f:view>
```

## 6) Sample bean

```
@Named
@RequestScoped

    private List<User> users = new ArrayList<>();


    public List<User> getUsers() {
        return users;
    }

    public void setUsers(List<User> users) {
        this.users = users;
    }

    public void addRow(){
        users.add(new User());
    }

    public void submit() {
        System.out.println(users);
    }
}

```

evilkid:
Add a valid flag to the cell, if the cell has changed, or the valid flag is set to false, then re submit the cell edit request.
set the valid flag to false when the server returns validation errors, true otherwise.

Possible fix for the [issue #2512](https://github.com/primefaces/primefaces/issues/2512)

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
+++ b/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
@@ -2412,6 +2412,7 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
     saveCell: function(cell) {
         var inputs = cell.find('div.ui-cell-editor-input :input:enabled'),
         changed = false,
+        valid = cell.data('valid'),
         $this = this;
 
         if(cell.data('multi-edit')) {
@@ -2427,7 +2428,7 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
             changed = (inputs.eq(0).val() != cell.data('old-value'));
         }
 
-        if(changed)
+        if(changed || !valid)
             $this.doCellEditRequest(cell);
         else
             $this.viewMode(cell);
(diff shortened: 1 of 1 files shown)


tandraschko:
@melloware @Rapster could you also check this? The PR itself looks fine, however i have no time to reproduce the error and check for other ways to fix it.

melloware:
I will see if I can test it out to see.

melloware (approved):

melloware:
I have tested that this fix does what is intended.   I only tested this scenario though and did not do an exhaustive test of other scenarios.

tandraschko:
Thanks for the PR!

PR #2693, opened 2017-09-06
Fix #2662 - Panel: add ui-panel-actions for facet "actions"

### LINKED ISSUE #2662: Panel: add ui-panel-actions for facet "actions"
Basically, our panel will contain menubutton in facet actions, instead of writing something like this:
```xml
<p:panel>
    <f:facet name="actions">
        <div class="float: right">
            <p:menuButton />
        </div>
    </f:facet>
</p:panel>
```

I could simply use:
```css
.ui-panel-actions {
    float: right;
}
```
Facet would generated in div (with inline-block style)

I'll try to provide a PR soon

Rapster:


Rapster:
I think encodeFacet should be placed in `CoreRenderer`, what do you think @tandraschko?

tandraschko:
What do you mean with the last question?

Rapster:
There is 6 implementaions of `encodeFacet` in primefaces, I think we could only have one implemented in `CoreRenderer` so every renderer could use it

tandraschko:
Do you mean this:

`        
UIComponent actionsFacet = panel.getFacet("actions");
if (actionsFacet != null) {
     actionsFacet.encodeAll(context);
}
`
?

Rapster:
yes

tandraschko:
Yep, why not but it wouldn't work for this case as clean as for a normal case :P

Rapster:
Why wouldn't it work?

PR #2704, opened 2017-09-08
Fix #2703: Added ComponentUtils.resolveLocale from Extensions

### LINKED ISSUE #2703: ComponentUtils.resolveLocale() from PF Extensions
@tandraschko I wanted to add this method to PF ComponentUtils to get it out of PrimeFaces Extensions.

Basically it uses the existing locale util method and if not found defaults to the JSF default Locale.  It is used by...

- PFE TimePicker
- PFE DocumentViewer
- PFE Calculator

```java
/**
     * Gets a {@link Locale} instance by the value of the component attribute "locale" which can be String or {@link Locale} or null. It can be used in any web
     * projects as a helper method.
     *
     * @param locale given locale
     * @return resolved Locale
     */
    public static Locale resolveLocale(Object locale) {
        if (locale instanceof String) {
            locale = ComponentUtils.toLocale((String) locale);
        }

        if (locale == null) {
            locale = FacesContext.getCurrentInstance().getViewRoot().getLocale();
        }

        return (Locale) locale;
    }
```

melloware:


tandraschko:
As discussed it in the ticket, will close this for now.

PR #2743, opened 2017-09-21
Fix #2720: Hidden elements should have the `hidden` HTML attribute

### LINKED ISSUE #2720: SelectBooleanCheckbox: Hidden elements should have the aria-hidden attribute
## 1) Environment
- PrimeFaces version: 6.1
- Application server + version: WebSphere 8.5.5.12
- Affected browsers: Internet Explorer/ Chrome/Firefox

## 2) Expected behavior

```
<p:selectBooleanCheckbox
  value="#{applicationBean.application.firstNation}"
  itemLabel="First Nation" />
```

Should render 

```
<div id="applicationForm:j_id1447386971_3_3b2db25b" class="ui-chkbox ui-widget"<div class="ui-helper-hidden-accessible">
  <input id="applicationForm:j_id1447386971_3_3b2db25b_input" name="applicationForm:j_id1447386971_3_3b2db25b_input" autocomplete="off" aria-checked="false" type="checkbox" hidden>
...
```

## 3) Actual behavior

```
<div id="applicationForm:j_id1447386971_3_3b2db25b" class="ui-chkbox ui-widget"<div class="ui-helper-hidden-accessible">
  <input id="applicationForm:j_id1447386971_3_3b2db25b_input" name="applicationForm:j_id1447386971_3_3b2db25b_input" autocomplete="off" aria-checked="false" type="checkbox" >
...
```

Missing `hidden` element causing accessibility tool to flag it.

## 4) Steps to reproduce

```
<p:selectBooleanCheckbox
  value="#{applicationBean.application.firstNation}"
  itemLabel="First Nation" />
```

## 5) Sample XHTML

```
<p:selectBooleanCheckbox
  value="#{applicationBean.application.firstNation}"
  itemLabel="First Nation" />
```

## 6) Sample bean

```
    Boolean firstNation;
    public Boolean getFirstNation() {
        return firstNation;
    }
    public void setFirstNation(final Boolean a) {
       firstNation = a;
    }
```

melloware:
Fix #2720: Hidden elements should have the `hidden` HTML attribute

I found everywhere "ui-helper-hidden-accessible" was hiding an input and added the hidden HTML attribute.

### DIFF
--- a/src/main/java/org/primefaces/component/checkbox/CheckboxRenderer.java
+++ b/src/main/java/org/primefaces/component/checkbox/CheckboxRenderer.java
@@ -81,7 +81,7 @@ public class CheckboxRenderer extends InputRenderer {
         writer.writeAttribute("type", "checkbox", null);
         writer.writeAttribute("class", "ui-chkbox-clone", null);
         writer.writeAttribute("data-itemindex", checkbox.getItemIndex(), null);
-        writer.writeAttribute("hidden", "true", null);
+        writer.writeAttribute("hidden", "hidden", null);
 
         if (tabindex != null) writer.writeAttribute("tabindex", tabindex, null);
         if (disabled) writer.writeAttribute("disabled", "disabled", null);
--- a/src/main/java/org/primefaces/component/checkbox/CheckboxRenderer.java
+++ b/src/main/java/org/primefaces/component/checkbox/CheckboxRenderer.java
@@ -23,6 +23,7 @@ import org.primefaces.component.radiobutton.RadioButtonRenderer;
 import org.primefaces.component.selectmanycheckbox.SelectManyCheckbox;
 import org.primefaces.expression.SearchExpressionFacade;
 import org.primefaces.renderkit.InputRenderer;
+import org.primefaces.util.Constants;
 import org.primefaces.util.HTML;
 import org.primefaces.util.SharedStringBuilder;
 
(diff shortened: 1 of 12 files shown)


tandraschko:
Is it correct to use a empty value? 

> In XHTML, attribute minimization is forbidden, and the hidden attribute must be defined as <element hidden="hidden">.

tandraschko:
maybe also hidden=true is more correct as it's a boolean attribute. not sure here @melloware.

melloware:
@tandraschko here is what I tried...

`writer.writeAttribute("hidden", null, null);` did not write anything out. It's like the response writer knew to ignore it.

`writer.writeAttribute("hidden", Constants.EMPTY_STRING, null);` writes out <input id="" hidden> just like the HTML5 docs had it.

`writer.writeAttribute("hidden", "true", null);` writes it out <input id="" hidden="true">

The HTML5 doc has it as a boolean so I suppose there is no harm in making it "true"?

https://www.w3schools.com/Tags/att_global_hidden.asp

tandraschko:
Thats a good question. Not sure if "true" or "" is more vaild. Would be great if you could check the docus and examples on the web.

melloware:
hidden="hidden" appears to be the most correct for XHTML.  So I have updated this PR.

**Differences Between HTML and XHTML**
In XHTML, attribute minimization is forbidden, and the hidden attribute must be defined as `<element hidden="hidden">`.

tandraschko:
I didn't check each component for now but i hope you only applied it to the right elements :D

PR #2762, opened 2017-09-29
Fix #942 - not to remove rowTogglerContent when a row is collpased …

### LINKED ISSUE #942: PrimeFaces 5.2 - DataTable RowExpansion <p:rowExpansion> with <p:selectBooleanCheckbox> along with pagination has weird behavior
My requirement is to show checkbox inside a rowexapansion of a datatable

Problem:
All the boolean attributes for the entities in the list for which the datatable is referring become false when I use <p:datatable> with <p:selectBooleanCheckbox> along with pagination

Steps to reproduce:
1. Try to use the prime faces rowExpansion from PrimeFaces showcase
2. Place selectBooleanCheckbox which shows the boolean value
3. Access the pagination of the datatable
4. See that for all the entries in the list the boolean attribute become false, also I tried to print the datatable in the managed bean, the data in the list also got modified, it is bit weird that with row expansion the data in list is being modified
6. When I test the same feature with rowExpansion and without selectBooleanCheckbox everything works normally

For more clarity I have attached the screenshots (data_after_row_exapantion.png, data_modifed_auto_after_row_exapantion.png)

data_after_row_exapantion
data_modifed_auto_after_row_exapantion

Am I missing anything or it is a defect in PrimeFaces?

jxmai:
…to avoid incorrect form mapping

Fix #942

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
+++ b/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
@@ -2044,11 +2044,8 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
     loadExpandedRowContent: function(row) {
         // To check whether or not any hidden expansion content exists to avoid reloading multiple duplicate nodes in DOM
         var expansionContent = row.next('.ui-expanded-row-content');
-        var reloadContentThroughAjax = true;
         if(expansionContent.length > 0) {
-            reloadContentThroughAjax = false;
-            row.addClass('ui-expanded-row');
-            expansionContent.show();
+            expansionContent.remove();
         }
         
         var $this = this,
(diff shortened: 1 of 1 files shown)


tandraschko:
@jxmai please add a comment why we only do a hide instead remove, otherwise it could happen that someone changes it again

jxmai:
@tandraschko I added some comments and I enhanced the `loadExpandedRowContent` function.


Although only hiding the content can solve #942 , the problem was if we only hide the data, the DOM will keep expanding everytime when we call `loadExpandedRowContent`.

The solution will be only to load the content when there is no content, and to show the content without calling AJAX afterward. I tested in an application with data and there should be no breaking changes.

tandraschko:
Hmm... Is it correct to not load it again? Currently the data will be received every time. With this change, old data might be displayed.
But it would be great if you could seperate the tickets + PRs. We should not mix mulitple issues and PRs together.

jxmai:
> Is it correct to not load it again? 

This is a valid concern. Although I tested and worked OK, the safest change might be to hide the content, but to reload everytime. 

> But it would be great if you could seperate the tickets + PRs. We should not mix mulitple issues and PRs together.

There is already a ticket attached.  I need to emphasis that we must use hide instead of remove to solve #942, but we also need to clean up the DOM so that the duplicate nodes won't be accumulated.

I will modify this PR and will ping you when it is ready to review.

jxmai:
@tandraschko I modfiied the PR.  The content will reload everytime on expand as before, but just to clean up any hidden content before loading a new one.

Please review.

PR #2773, opened 2017-10-03
Fix #2317 - set node correctly even if clicking  on the icon instead …

### LINKED ISSUE #2317: contextMenu error on toggler icon of tree
## 1) Environment
- PrimeFaces version: 6.1
- Does it work on the newest released PrimeFaces version? Version? There is no newer version yet.
- Does it work on the newest sources in GitHub? (Error may reproduced whatever)
- Application server + version: Tomcat 8. (In showcase too)
- Affected browsers: All

## 2) Expected behavior

ContextMenu should invoke for selected node

## 3) Actual behavior

Node remains unselected and ContextMenu invokes for indefinite node and delete operation shows error

## 4) Steps to reproduce

right click on ui-tree-toggler of node to invoke contextMenu and choose DELETE

## 5) Sample XHTML

from showcase (https://www.primefaces.org/showcase/ui/data/tree/contextMenu.xhtml)

## 6) Sample bean

from showcase (https://www.primefaces.org/showcase/ui/data/tree/contextMenu.xhtml)

jxmai:
…of clicking on the node content

Fix #2317

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/tree/tree.js
+++ b/src/main/resources/META-INF/resources/primefaces/tree/tree.js
@@ -30,7 +30,7 @@ PrimeFaces.widget.BaseTree = PrimeFaces.widget.BaseWidget.extend({
     },
 
     bindContextMenu : function(menuWidget, targetWidget, targetId, cfg) {
-        var nodeContentSelector = targetId + ' .ui-tree-selectable',
+        var nodeContentSelector = targetId + ' .ui-tree-selectable' + ' > span:not(.ui-tree-toggler)',
         nodeEvent = cfg.nodeType ? cfg.event + '.treenode.' + cfg.nodeType : cfg.event + '.treenode',
         containerEvent = cfg.event + '.tree';
 
--- a/src/main/resources/META-INF/resources/primefaces/tree/tree.js
+++ b/src/main/resources/META-INF/resources/primefaces/tree/tree.js
@@ -314,22 +314,24 @@ PrimeFaces.widget.BaseTree = PrimeFaces.widget.BaseWidget.extend({
     nodeRightClick: function(event, nodeContent) {
         PrimeFaces.clearSelection();
 
-        var node = nodeContent.parent(),
-        selectable = nodeContent.hasClass('ui-tree-selectable');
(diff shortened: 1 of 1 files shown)


melloware (commented):
So from this it looks like you just removed the "if" and everything is still working as expected?

melloware:
thumbs up then!

jxmai:
Yes, it is working as expected

image

**right clicking on the icon**

image

image

melloware (approved):

tandraschko:
Hmm not sure about this PR.
The code now also selects the node, if you click on the toggler?

melloware:
@tandraschko I think the intent was to make this work more like Windows does where if you right click an unselected tree node it selects the node and presents the context menu.  Rather than 2 steps of left click on the tree node you want then right click to bring up the context menu. @jxmai correct me if I am wrong?

tandraschko:
Nope. If you use the windows 7 explorer and click on the toggler via right click, the node is NOT selected.

jxmai:
There was some misunderstanding going on, so let's revisit #2317 

**The problem in summary**:  a null pointer will be thrown if user  right clicks on the icon and delete. Delete will never work even if user can prevent the null pointer from throwing since the node wasn't set properly

---

I presented two ideas:

> The issue should be because selected node is not set properly by right clicking the icon. Either we should prevent the context menu from popping up if user right clicks on the icon directly or we should set the node correctly rather than null behind the scene.

---

and @melloware  supported the second one

> I vote for selecting the node on right click.


but @tandraschko  supported the first one:

> +1 to fix our code and ignore right clicks on the toggler

tandraschko:
The right click on the node-icon should select the node, which already works fine for me.
The right click on the toggler-arrow-icon should not select a node, but it should also not even open the context menu.

jxmai:
OK then. Here is the trick which is for solution one. I will modify this PR shortly

```js
var nodeContentSelector = targetId + ' .ui-tree-selectable' + ' > span:not(.ui-tree-toggler)',
```

jxmai:
Tested and worked as expected

@tandraschko @melloware Please review.

melloware:
Looks good to me.

PR #2819, opened 2017-10-15
Fix #2600 - split button's menu width should fit its content

### LINKED ISSUE #2600: Splitbutton: panel width should fit his content
## 1) Environment
- PrimeFaces version: 6.1
- Affected browsers: Chrome

## 2) Expected behavior

image

FYI, autocomplete handle this case

## 3) Actual behavior

image

## 4) Steps to reproduce
To reproduce the expected behavior, I removed `.ui-menu` width set at 12.5em and` .ui-menu .ui-menuitem-link `set at 92% _(I don't understand this one by the way, I test the menus without it nd works nicely)_

## 5) Sample XHTML

```xml
 <h:form>       
            <p:growl id="messages" />

            <p:splitButton value="Long Long Long Long Long Long Long save" actionListener="#{buttonView.save}" update="messages" icon="ui-icon-disk">
                <p:menuitem value="Long long long long long long text" actionListener="#{buttonView.update}" update="messages" icon="ui-icon-arrowrefresh-1-w" />
            </p:splitButton>   
        </h:form>
```

jxmai:
Fix #2600

### DIFF
--- a/src/main/java-templates/org/primefaces/component/splitbutton/SplitButtonTemplate.java
+++ b/src/main/java-templates/org/primefaces/component/splitbutton/SplitButtonTemplate.java
@@ -6,6 +6,7 @@ import org.primefaces.util.HTML;
     public static final String MENU_ICON_BUTTON_CLASS = "ui-splitbutton-menubutton  ui-button ui-widget ui-state-default ui-corner-right ui-button-icon-only";
     public final static String BUTTON_TEXT_ONLY_BUTTON_CLASS = "ui-button ui-widget ui-state-default ui-corner-left ui-button-text-only";
     public final static String BUTTON_ICON_ONLY_BUTTON_CLASS = "ui-button ui-widget ui-state-default ui-corner-left ui-button-icon-only";
+    public static final String SPLITBUTTON_CONTAINER_CLASS = "ui-menu ui-splitbuttonmenu ui-menu-dynamic ui-widget ui-widget-content ui-corner-all ui-helper-clearfix ui-shadow";
 
     public String resolveStyleClass() {
         String icon = getIcon();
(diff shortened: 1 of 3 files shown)


tandraschko:
-1 for this fix
1) we should not render styles from the renderer, we have style classes for this
2) why is it actually necessary? We should check if another style or the markup prevents more width.

jxmai:
The width is defined by 

```css
.ui-menu {
    width: 12.5em;
    padding:0.3em;
    position:relative;
    outline: 0 none;
}
```
> why is it actually necessary? We should check if another style or the markup prevents more width.

It's better not to touch this css since `.ui-menu` is using across the entire codebase. It's 100% guarantee to break something if we try to mess with it :see_no_evil: 


---

So, we have four options:

1. ~~override the width in renderer~~ (not an option anymore)

> we should not render styles from the renderer, we have style classes for this 

2. setting the css in `alignPanel` in `form.splitbutton.js`

```js
alignPanel: function() {
        this.menu.css({left:'', top:'','z-index': ++PrimeFaces.zindex});

        if(this.menu.parent().is(this.jq)) {
            this.menu.css({
                left: 0,
                top: this.jq.innerHeight()
//  let's say width: max-content
            });
        }
        else {
            this.menu.position({
                my: 'left top'
                ,at: 'left bottom'
                ,of: this.button
            });
        }
    }
```

3. remove `.ui-menu` class (not recommend since it might also break something )

4. not to fix this issue in the framework and let user to override the css by themselves.


What do you think ? :thinking:

tandraschko:
2) is also a not go

we only have 2 solutions:
1) correctly apply with css, which only handles the splitbutton, not ui-menu in gernall
2) let the user decide

jxmai:
> correctly apply with css, which only handles the splitbutton, not ui-menu in gernall

I chose to go with this way, and I have corrected the PR.

Issue #1161, opened 2016-02-17
JS error in Calendar with "Current Date" button


mertsincan:
Reported by Pro User;
We have showButtonPanel="true" , clicking on current day button does nothing. We can see Exception in console. (It happens when timeOnly="#{ELBinding}" gets 'false')

PR #2827, opened 2017-10-17
Fix #2811: Regression issue firing too many events after TimePicker 1…

### LINKED ISSUE #2811: Spurious valueChange events in a Calendar with a datetime pattern
## 1) Environment

## 2) Expected behavior

## 3) Actual behavior

## 4) Steps to reproduce

## 5) Sample XHTML

```
<!DOCTYPE html>
<f:view locale="en">
	<h:head>
	</h:head>

	<h:body>
		<h:form>
			</p:calendar>

			</p:calendar>

			</p:calendar>

			<script>
				function vc(wVar) {
				}
			</script>
		</h:form>
	</h:body>
</f:view>
</html>
```

## 6) Sample bean

Unnecessary.

melloware:
Fix #2811: Regression issue firing too many events after TimePicker 1.6.3 upgrade.

Original ticket was https://code.google.com/archive/p/primefaces/issues/5702

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.timepicker.js
+++ b/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.timepicker.js
@@ -1579,7 +1579,6 @@
 		}
 
 		/* PrimeFaces Customization
-	    var tzoffset = $.timepicker.timezoneOffsetNumber(tp_inst.timezone);
 		var now = new Date();
 		now.setMinutes(now.getMinutes() + now.getTimezoneOffset() + parseInt(tzoffset, 10));
 		this._setTime(inst, now);
@@ -1588,7 +1587,7 @@
 		*/
 	      
         // PrimeFaces custom code to handle Today button
-        selectLocalTimezone(tp_inst); 
+		var tzoffset = $.timepicker.timezoneOffsetNumber(tp_inst.timezone);
         var now = new Date();
         this._setTime(inst, now);
         $('.ui-datepicker-today', inst.dpDiv).click();
--- a/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.timepicker.js
+++ b/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.timepicker.js
@@ -1578,18 +1578,16 @@
 		  return;
 		}
 
-		/* PrimeFaces Customization
+		var tzoffset = $.timepicker.timezoneOffsetNumber(tp_inst.timezone);
 		var now = new Date();
 		now.setMinutes(now.getMinutes() + now.getTimezoneOffset() + parseInt(tzoffset, 10));
 		this._setTime(inst, now);
 		this._setDate(inst, now);
+		
+		/* PrimeFaces Customization
 		tp_inst._onSelectHandler();
 		*/
-	      
         // PrimeFaces custom code to handle Today button
-		var tzoffset = $.timepicker.timezoneOffsetNumber(tp_inst.timezone);
-        var now = new Date();
-        this._setTime(inst, now);
         $('.ui-datepicker-today', inst.dpDiv).click();
 	};
 
--- a/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.timepicker.js
+++ b/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.timepicker.js
@@ -988,7 +988,9 @@
 				this.$input.val(formattedDateTime);
 			}
 
+			/* PrimeFaces Customization
 			// this.$input.trigger("change"); https://github.com/primefaces/primefaces/issues/2811
+			*/
 		},
 
 		_onFocus: function () {
--- a/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.timepicker.js
+++ b/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.timepicker.js
@@ -2259,12 +2259,10 @@
 	 * @return {void}
 	 */
 	$.timepicker.log = function () {
-	    /* PrimeFaces Customization
 		// Older IE (9, maybe 10) throw error on accessing `window.console.log.apply`, so check first.
 		if (window.console && window.console.log && window.console.log.apply) {
 			window.console.log.apply(window.console, Array.prototype.slice.call(arguments));
 		}
-		*/
 	};
 
 	/*
--- a/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.timepicker.js
+++ b/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.timepicker.js
@@ -984,7 +984,7 @@
 				this.$input.val(formattedDateTime);
 			}
 
-			this.$input.trigger("change");
+			// this.$input.trigger("change"); https://github.com/primefaces/primefaces/issues/2811
 		},
 
 		_onFocus: function () {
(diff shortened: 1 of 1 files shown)


jxmai:
I doubt there are more than one hack without comments since last time I checked showing me that the one line comment cannot fix the issue, and my doubt is real. I will explain in details later.

jxmai:
On the right is the file in Primefaces, on the left is the original file (Both are v 1.6.1)

image


---
image

---

image

---

image


---

When there is one hack , there are always more than one :tophat: 


---

There are a couple more notes in the 1.6.1 version in Primefaces

https://github.com/primefaces/primefaces/blob/1926c7e974cc0b3f692345823be8d4491a394f51/src/main/resources/META-INF/resources/primefaces/jquery/jquery-plugins.js#L11547

which is pointing to https://github.com/trentrichardson/jQuery-Timepicker-Addon/issues/848


and https://github.com/primefaces/primefaces/blob/1926c7e974cc0b3f692345823be8d4491a394f51/src/main/resources/META-INF/resources/primefaces/jquery/jquery-plugins.js#L12231

which is pointing to 

https://github.com/trentrichardson/jQuery-Timepicker-Addon/issues/281


while the latest code also forgot to include the fix back...

For example

https://github.com/primefaces/primefaces/blob/5a49909adfae441f9580c40fc834bf8ff8c1e01e/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.timepicker.js#L1575

melloware:
Let me analyze further and get back to you. If we do more changes I plan on commenting every single one.

melloware:
OK have a look at my latest commit.

tandraschko:
@mertsincan Could you also review this? You did some changes on that components the last days

src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.timepicker.js
```diff
@@ -1572,12 +1578,20 @@
 		  return;
 		}
 
-		var tzoffset = $.timepicker.timezoneOffsetNumber(tp_inst.timezone);
+		/* PrimeFaces Customization
+	    var tzoffset = $.timepicker.timezoneOffsetNumber(tp_inst.timezone);
```
jxmai: This customization is originated from https://github.com/trentrichardson/jQuery-Timepicker-Addon/issues/281 if you look at https://github.com/primefaces/primefaces/blob/1926c7e974cc0b3f692345823be8d4491a394f51/src/main/resources/META-INF/resources/primefaces/jquery/jquery-plugins.js#L12231, but it seems to be fixed in `1.6.2` and `1.6.3` natively if you look at this commit: https://github.com/trentrichardson/jQuery-Timepicker-Addon/commit/60b93e495387488a890c7e5aa2e834deb3fdb17a

melloware:
OK check my latest commit.  I removed the 1 line and put back their tz line.

melloware:
Sorry one more.  Basically the only line I change in that whole method now is this...

```javascript
	/* PrimeFaces Customization
	tp_inst._onSelectHandler();
	*/
        // PrimeFaces custom code to handle Today button
        $('.ui-datepicker-today', inst.dpDiv).click();
```

src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.timepicker.js
```diff
@@ -1577,7 +1583,12 @@
 		now.setMinutes(now.getMinutes() + now.getTimezoneOffset() + parseInt(tzoffset, 10));
 		this._setTime(inst, now);
 		this._setDate(inst, now);
+		
+		/* PrimeFaces Customization
 		tp_inst._onSelectHandler();
```
jxmai, line 1586: Any reason to comment this line out ? :thinking:

melloware:
Well it looks like PF the Today button updates the calendar and if you look at your comparison above it was commented out in the old 1.6.1.  Not sure why?

jxmai:
> Well it looks like PF the Today button updates the calendar and if you look at your comparison above it was commented out in the old 1.6.1. Not sure why?

https://github.com/primefaces/primefaces/commit/5ade2f744b9505ae1c20373c87b88657a9b57c58 in   https://github.com/primefaces/primefaces/issues/1161

melloware:
got it.  I think then I should do the exact same thing this commit was. Basically only setting the date and ignoring the time portion entirely?

jxmai:
I followed the instruction

> We have showButtonPanel="true" , clicking on current day button does nothing. We can see Exception in console. (It happens when timeOnly="#{ELBinding}" gets 'false')

```xhtml
<!DOCTYPE html>

<h:head>
</h:head>
<h:body>

	<h:form>
		<p:calendar showButtonPanel="true" timeOnly="false"></p:calendar>
	</h:form>

</h:body>
</html>
```

and the today button is working fine. Maybe the fix is no longer needed... ?

image

No exception in console of course.

melloware:
nice!  So I think we are good with just the one line change to disable their handle and call our datepicker click.

```javascript
/* PrimeFaces Customization
	tp_inst._onSelectHandler();
	*/
        // PrimeFaces custom code to handle Today button
        $('.ui-datepicker-today', inst.dpDiv).click();
```

jxmai:
Looking good so far :+1: . There is one more thing, the intention of the following comment block is unclear. I know it is a part of  https://github.com/primefaces/primefaces/commit/5ade2f744b9505ae1c20373c87b88657a9b57c58, but is it really needed especially when https://github.com/trentrichardson/jQuery-Timepicker-Addon/issues/281 has been addressed ?

```js
/* PrimeFaces Customization
  				onChangeMonthYear: function (year, month, dp_inst) {
  					// Update the time as well : this prevents the time from disappearing from the $input field.
  					// tp_inst._updateDateTime(dp_inst);
  					if ($.isFunction(tp_inst._defaults.evnts.onChangeMonthYear)) {
  						tp_inst._defaults.evnts.onChangeMonthYear.call($input[0], year, month, dp_inst, tp_inst);
  					}
  				},
```

melloware:
I agree i will uncommment.  I can't see how it affected anything. I just checked in that change.

src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.timepicker.js
```diff
@@ -2250,10 +2259,12 @@
 	 * @return {void}
 	 */
 	$.timepicker.log = function () {
+	    /* PrimeFaces Customization
```
jxmai: This comment block might not be needed since it is part of https://github.com/primefaces/primefaces/commit/5ade2f744b9505ae1c20373c87b88657a9b57c58. I guess what it is trying to do is to not to log errors to console  Since there is no error anyway, we might bring it back.

melloware: Reverted.  Agreed the console logging should be harmless and should actually show if there is an error

jxmai:
LGTM

jxmai (approved):

tandraschko:
I wonder if we should do the changes in jquery.ui.pfextensions? we already have there some overwrites for datepicker
And, changes in external sources, will someday be overwritten and isn't nice.

tandraschko:
AND changes should be mentioned in extensions.txt

melloware:
@tandraschko I didn't realize that but I like that solution much better.  Let me cancel this PR and see if I can rework a new PR to use the extensions way.

jxmai:
Extension override is the ideal way to go :+1: . I didn't know about it already exists in PF  ...

melloware:
I couldn't figure out how to make it work.  I just opened another PR with the correct fix and I verified that all the other old changes we talked about are not needed in the new code.

jxmai:
will take a look at the other PR later.

jxmai:
For tracking purposes: This PR is superseded by #2839

PR #2834, opened 2017-10-18
Fix #2831: Auto adjust the width to fit content of a Dialog Framework

### LINKED ISSUE #2831: Dialog Framework: auto width ignored
It's not possible to auto adjust the width to fit content of a Dialog Framework.

When I put the value of parameter "contentWidth" as "auto" it doesn't change and it is always 640, because this is the default value of property 'contentWidth'. In fact, inspecting the page at runtime it puts the iframe width value as "auto" literally, instead of translating it to the corresponding calculated value.

## 1) Environment
- PrimeFaces version: 6.1.7
- It does not work on the newest sources in GitHub
- Application server + version: Tomcat 8.0.27
- Affected browsers: All

## 2) Expected behavior
 Auto adjust the width to fit content of a Dialog Framework by setting "auto" value to "contentWidth" parameter.

## 3) Actual behavior
The "auto" value to "contentWidth" parameter is ignored.

## 4) Steps to reproduce
options.put("contentWidth", "auto");  

## 5) Sample XHTML
`<h:form>
    <p:commandButton value="Options" icon="ui-icon-extlink" actionListener="#{dfView.viewCarsCustomized}" />
</h:form>`

## 6) Sample bean
```
package org.primefaces.showcase.view.df;
 
import java.util.HashMap;
import java.util.Map;
import org.primefaces.context.RequestContext;
import org.primefaces.event.SelectEvent;
import org.primefaces.showcase.domain.Car;
 
@ManagedBean(name = "dfView")
public class DFView {
     
     public void viewCarsCustomized() {
        Map<String,Object> options = new HashMap<String, Object>();
        options.put("modal", true);
        options.put("contentWidth", "auto");        
         
        RequestContext.getCurrentInstance().openDialog("viewCars", options, null);
    }
     
    public void chooseCar() {
        RequestContext.getCurrentInstance().openDialog("selectCar");
    }
     
    public void onCarChosen(SelectEvent event) {
        Car car = (Car) event.getObject();
        FacesMessage message = new FacesMessage(FacesMessage.SEVERITY_INFO, "Car Selected", "Id:" + car.getId());
         
        FacesContext.getCurrentInstance().addMessage(null, message);
    }
     
    public void showMessage() {
        FacesMessage message = new FacesMessage(FacesMessage.SEVERITY_INFO, "What we do in life", "Echoes in eternity.");
         
        RequestContext.getCurrentInstance().showMessageInDialog(message);
    }
}
```

culjar:
Fix #2831  issue

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/core/core.dialog.js
+++ b/src/main/resources/META-INF/resources/primefaces/core/core.dialog.js
@@ -31,7 +31,7 @@ if (!PrimeFaces.dialog) {
                 titlebar.append('<a class="ui-dialog-titlebar-icon ui-dialog-titlebar-maximize ui-corner-all" href="#" role="button"><span class="ui-icon ui-icon-extlink"></span></a>');
             }
 
-            dialogDOM.append('<div class="ui-dialog-content ui-widget-content ui-df-content" style="height: auto;">' +
+            dialogDOM.append('<div class="ui-dialog-content ui-widget-content ui-df-content" style="height: auto; width: auto">' +
                     '<iframe style="border:0 none" frameborder="0"/>' + 
                     '</div>');
 
@@ -39,10 +39,7 @@ if (!PrimeFaces.dialog) {
 
             var dialogFrame = dialogDOM.find('iframe'),
             symbol = cfg.url.indexOf('?') === -1 ? '?' : '&',
-            frameURL = cfg.url.indexOf('pfdlgcid') === -1 ? cfg.url + symbol + 'pfdlgcid=' + cfg.pfdlgcid: cfg.url,
-            frameWidth = cfg.options.contentWidth||640;
-
-            dialogFrame.width(frameWidth);
+            frameURL = cfg.url.indexOf('pfdlgcid') === -1 ? cfg.url + symbol + 'pfdlgcid=' + cfg.pfdlgcid: cfg.url;
 
             dialogFrame.on('load', function() {
                 var $frame = $(this),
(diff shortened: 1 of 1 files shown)


tandraschko:
Could someone re-test it?

jxmai:
Indeed,  the width calculation is 

https://github.com/primefaces/primefaces/blob/ee5cc1285d731e078e2f95b51e02e9a4f4fc2aff/src/main/resources/META-INF/resources/primefaces/core/core.dialog.js#L43

Previously, if no width is defined or even `auto`, the `cfg.options.contentWidth` will be `undefined` and frameWidth will always be 640px

image

---

However, with this PR:

image

Here is the main point:

**The width auto is correctly set, which is nice in theory, but I don't see any visual differences as the end result**

image



@culjar Please explain the purpose of this PR and the use case you need to address since the end result seems the same before and after the PR.

jxmai:
I even tried to double the amount of columns and to see the `width:auto` can do anything better than 640px, but the end result is not desirable.

image

Since you need to set the width anyway to make the table above  become usable, I don't see any improvement can be made by setting `width:auto` 🤔 .

This PR currently is +0 for me. Any other opinion is welcome 😃

jxmai (commented):
Review Summary: 

This PR is OK, but I don't see the improvement currently.  +0 for me.

If the end result indeed no differences at all after reconfirming, then -1 for me since there are more lines of code with slightly worse performance than the previous version.

At the end of the day, the end result is all that matters.

culjar:
To me, works perfectly.

Next I put several screenshots so that you see the good functioning.

Dialog framework with datatable:
view

Options for open dialog:
codebean

Result in console of Chrome (You can see that it perfectly calculates the width of the iframe):
content
iframe

@jxmai I think the problem in your case is that you are not putting a width size to the datatable and therefore is unable to calculate it, try to set a fixed width or create a dialog framework with other components that are not datatable.

This PR works perfectly and is not +0.

jxmai:
@culjar Could you provide a full runnable sample for us to verify again since your code in #2831 is based on part of the showcase but not complete

- You can paste the whole XHTML + bean code in here or to fork https://github.com/primefaces/primefaces-test 

>  think the problem in your case is that you are not putting a width size to the datatable and therefore is unable to calculate it, try to set a fixed width or create a dialog framework with other components that are not datatable.

Not exactly. Please see below

image


image

culjar:
I downloaded primefaces-test and I made the example of Dialog Framework that is in the Showcase and does not work the automatic adjustment of the width of the dialog with the solution that I proposed.

At first I tested it in a project of mine and it seemed to work, but I have realized that it is not like that, it also does it wrong.

You're right @jxmai . Sorry to have put an invalid solution.

I hope someone can solve it as soon as possible.

tandraschko:
So we can close that PR?

jxmai:
Yes. This PR cannot address the issue. I might take a look later, but this PR can be closed.

PR #2839, opened 2017-10-19
Fix #2811: Calendar change events fired when date did not change.

### LINKED ISSUE #2811: Spurious valueChange events in a Calendar with a datetime pattern

melloware:
OK I was able to review all the old changes and none of them need to be applied anymore as they were fixed in TimePicker 1.6.3.

However, I was able to narrow down this bug and only fire the Ajax event if the date actually did change.

@tandraschko @jxmai I tried everything to move this code into jquery.pfextensions.js but it just would not pick up the methods.  Maybe I was doing it wrong but I was doing this...

`$.timepicker._updateDateTime = function (dp_inst) {`

Just like the$datepicker overrides but I must be missing something.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/jquery/extensions.txt
+++ b/src/main/resources/META-INF/resources/primefaces/jquery/extensions.txt
@@ -19,7 +19,9 @@ jquery.autosize.js:
 - Added fix to autosize from https://github.com/jackmoore/autosize/issues/225
 
 jquery.ui.timepicker.js:
-- Added fix to timepicker from https://github.com/trentrichardson/jQuery-Timepicker-Addon/issues/848
+- Timepicker Fix https://github.com/trentrichardson/jQuery-Timepicker-Addon/issues/848
+- Timepicker Fix https://github.com/primefaces/primefaces/issues/2811
+- 
 
 jquery.maskedinput.js:
 - line 116: Added arrow keys fix to mask (https://github.com/primefaces/primefaces/issues/1481)
--- a/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.js
+++ b/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.js
@@ -1,4 +1,6 @@
 /* Primefaces Extensions */
+
+/** DatePicker */
 (function () {
     var original_gotoToday = $.datepicker._gotoToday;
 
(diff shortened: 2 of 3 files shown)


jxmai:
@melloware I have reviewed your code, the reason of 

> I tried everything to move this code into jquery.pfextensions.js but it just would not pick up the methods. Maybe I was doing it wrong but I was doing this...

is the function you were trying to override is at `instance` level, so it cannot be affected by your newly added function.

The easiest way is to alter the original js file a bit, just add a single line into the `_netInst` function

```js
tp_inst._updateDateTime = $.timepicker._updateDateTime;
```

and I don't think we need to use JS IIFE at all if we go this route.

jxmai:
Here is the fix which should work

https://github.com/jxmai/primefaces/commit/bed7b088af4bc13ad402d0e22d54e4dd67261440


image

melloware:
@jxmai Great work and thanks for the assist!   I just updated my PR.

@mertsincan @tandraschko I also moved Mert's fix for #1421 to this same pattern where we extend the timpicker code in the pfextensions.ui.js

tandraschko:
Whats the actual bug here @melloware?
Can you provide a xhtml to reproduce?

melloware:
Yep I can make a test case. But the jist of it is say you have Date "2018-10-31" selected in the Calendar.  And you open the calendar and don't make any changes or pick the same date "2018-10-31" its firing the Ajax event "dateChange" when the date has not really changed.

mertsincan:
Hi @melloware,

"dateChange" event?? I didn't see it on Calendar.java.
```
private static final Collection<String> EVENT_NAMES = LangUtils.unmodifiableList("blur", "change", "valueChange", "click", "dblclick",
            "focus", "keydown", "keypress", "keyup", "mousedown", "mousemove", "mouseout", "mouseover", "mouseup", "select", "dateSelect", "viewChange",
            "close");
    private static final Collection<String> UNOBSTRUSIVE_EVENT_NAMES = LangUtils.unmodifiableList("dateSelect", "viewChange", "close");
```

melloware:
Sorry I meant  "change" event.  It should not fire because the date has not changed.

mertsincan:
Also, I think you can add your fixes into *$.extend(Object.getPrototypeOf($.timepicker), {* in pfextensions.ui.js

melloware:
Yeah I wasn't quite sure how to extend move the stuff out of Jquery TimePicker and into Extensions properly.  @jxmai helped me follow a pattern I have seen in DatePicker but I am definitely no expert on how to properly move that stuff out correctly.

mertsincan:
No problem ;) Please see this line; https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.js#L241
You can add your methods into it.

But, I couldn't replicate this issue on 6.2.x. Could you please attach a sample code for us to replicate?

melloware:
Attached is runnable sample: 
[pf-2811.zip](https://github.com/primefaces/primefaces/files/2534906/pf-2811.zip)

jxmai:
> Whats the actual bug here

JFYI: the original bug report was raised by https://github.com/primefaces/primefaces/issues/2811

> An unneccesary/duplicate valueChange event is triggered on each calendar input blur.

Basically what it says ajax could be fired by clicking in and clicking out although the event is `change`

mertsincan:
Thanks, @melloware ;) Good catch! Could you please send a new PR after adding your fixes into $.extend(Object.getPrototypeOf($.timepicker), { in pfextensions.ui.js ?

Also, I think the problem is caused by the following code;
```
overrides = { // line 188
        beforeShow: function (input, dp_inst) {
                if ($.isFunction(tp_inst._defaults.evnts.beforeShow)) {
                        return tp_inst._defaults.evnts.beforeShow.call($input[0], input, dp_inst, tp_inst);
                }
        },
        onChangeMonthYear: function (year, month, dp_inst) {
                // Update the time as well : this prevents the time from disappearing from the $input field.
                // tp_inst._updateDateTime(dp_inst);
                if ($.isFunction(tp_inst._defaults.evnts.onChangeMonthYear)) {
                        tp_inst._defaults.evnts.onChangeMonthYear.call($input[0], year, month, dp_inst, tp_inst);
                }
        },
        onClose: function (dateText, dp_inst) {
/********* Maybe,  '&& dateText !=$input.val()' can be added into the if statement ********/
                if (tp_inst.timeDefined === true && $input.val() !== '') { 
                        tp_inst._updateDateTime(dp_inst);
                }
                if ($.isFunction(tp_inst._defaults.evnts.onClose)) {
                        tp_inst._defaults.evnts.onClose.call($input[0], dateText, dp_inst, tp_inst);
                }
        }
};
```
For now, we can override _updateDateTime method. I think this method is more likely to change in the future :D

melloware:
Closing this PR as I submitted a cleaner one.

PR #2849, opened 2017-10-20
Fix #2833: TreeTable/Datatable improve filter look in narrow columns

### LINKED ISSUE #2833: DataTable: improve filter look in narrow columns
## 1) Environment
- PrimeFaces version: SHOWCASE, all versions are affected
- Affected browsers: all

## 2) Expected behavior
Filter input doesn't overflow and it's always centered, even on narrow columns.

## 3) Actual behavior
filter overflow

## 4) Steps to reproduce
Go to https://www.primefaces.org/showcase/ui/data/datatable/filter.xhtml and narrow the page

## 5) Sample XHTML
Showcase

## 6) Sample bean
Not needed

## 7) Possible fix
I've tried to add `max-width: calc(100% - 10px)` to the `ui-datatable .ui-column-filter` CSS rule and it seems to work:
filter with max width

melloware:
Fix #2833: TreeTable/Datatable improve filter look in narrow columns

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datatable/datatable.css
+++ b/src/main/resources/META-INF/resources/primefaces/datatable/datatable.css
@@ -59,8 +59,8 @@
 
 .ui-datatable .ui-column-filter {
     display: block;
-    width:100px;
-    max-width: calc(100% - 10px);
+    box-sizing: border-box;
+    width: 100%;
     margin:auto;
 }
 
--- a/src/main/resources/META-INF/resources/primefaces/datatable/datatable.css
+++ b/src/main/resources/META-INF/resources/primefaces/datatable/datatable.css
@@ -60,6 +60,7 @@
 .ui-datatable .ui-column-filter {
     display: block;
     width:100px;
+    max-width: calc(100% - 10px);
     margin:auto;
 }
 
--- a/src/main/resources/META-INF/resources/primefaces/treetable/treetable.css
+++ b/src/main/resources/META-INF/resources/primefaces/treetable/treetable.css
@@ -1,4 +1,4 @@
-fi.ui-treetable table {
+.ui-treetable table {
 	border-collapse:collapse;
     width: 100%;
     table-layout: fixed;
@@ -165,9 +165,9 @@ fi.ui-treetable table {
 /** Filter **/
 .ui-treetable .ui-column-filter {
     display: block;
-    margin:auto;
     box-sizing: border-box;
     width: 100%;
+    margin:auto;
 }
 
 /** Reflow **/
--- a/src/main/resources/META-INF/resources/primefaces/treetable/treetable.css
+++ b/src/main/resources/META-INF/resources/primefaces/treetable/treetable.css
@@ -166,6 +166,7 @@
 .ui-treetable .ui-column-filter {
     display: block;
     width: 100px;
+    max-width: calc(100% - 10px);
     margin: auto;
 }
 
--- a/src/main/resources/META-INF/resources/primefaces/treetable/treetable.css
+++ b/src/main/resources/META-INF/resources/primefaces/treetable/treetable.css
@@ -165,9 +165,9 @@ fi.ui-treetable table {
 /** Filter **/
 .ui-treetable .ui-column-filter {
     display: block;
+    margin:auto;
     box-sizing: border-box;
     width: 100%;
-    margin:auto;
 }
 
 /** Reflow **/
--- a/src/main/resources/META-INF/resources/primefaces/treetable/treetable.css
+++ b/src/main/resources/META-INF/resources/primefaces/treetable/treetable.css
@@ -1,4 +1,4 @@
-.ui-treetable table {
+fi.ui-treetable table {
 	border-collapse:collapse;
     width: 100%;
     table-layout: fixed;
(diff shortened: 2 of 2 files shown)


src/main/resources/META-INF/resources/primefaces/datatable/datatable.css
```diff
@@ -60,6 +60,7 @@
 .ui-datatable .ui-column-filter {
     display: block;
     width:100px;
```
jxmai: This can cause problem if we have an extra large column

for example:

image

jxmai: I expect it to almost fill the column by default. What do you think ?

jxmai:
@melloware I think a better fix will be set the width to be 90% - 95% by default, so we don't need to set the max-width at all. It can address both narrow columns and large columns by one shot.

the hardcoded width: 100px  is not a good rule IMO.

NicolaIsotta:
@jxmai unfortunately setting the width with only the percentage causes an imperfect alignment. The reason is the default `box-sizing`.
Changing to `border-box`, at least for this specific element, would allow us to safely use `width: 100%`, as the left and right `padding` is inherited by the `ui-inputfield` class.

melloware:
so yeah i always thought the 100px was just there to make sure it fit and let you override it.  @NicolaIsotta fix is to make sure if it goes smaller than 100px that it adjusts.

Any other thoughts?

jxmai:
OK. I see. I am OK with the changes then.

jxmai (approved):

jxmai:
@NicolaIsotta I just got a chance to revisit this ticket. If I set the filter width to 90%, the imperfect alignment did not happen at least during my testing.

image

image

image

NicolaIsotta:
It's (barely) visible only on very narrow columns:
align
`width: calc(100% - 10px)` is pixel perfect instead.

jxmai:
OK, I see. What about setting width to     `width: -webkit-fill-available;` then?  Since it's better to have a common solution for whatever resolution is possible for ultimate responsiness. TBH,  either `width:100px` , `max-width:calc(100% - 10px)` or even `width:90%` as default does not look right for having magic numbers if there is a better solution out there. Notice that we may still need to consider other browsers such as firefox which may need `width: -moz-available; `.  We can define width multiple times in css and let browser to choose whatever it recognizes so this should not be a problem.

For more info, you can check out the following stackoverflow thread:

https://stackoverflow.com/questions/26275982/putting-moz-available-and-webkit-fill-available-in-one-width

---

image


image

melloware:
OK so after reading all of this are there are more recommended changes to this PR other than the one I have made?

jxmai:
Consider the following css. What do you think ? I am overriding the css this way, but when this PR is applied, the newly introduced `max-width` attribute might affect the client's application and might create trouble for others that are overriding the css currently.

```css
.ui-treetable .ui-column-filter {
	width: 100%;
	width: -webkit-fill-available;
	width: -moz-available;
}
```

---

You can always override the width just like before (`width:100px`), but there is no extra max-width field to be worried about since that field on the parent css will propagate to all the children. Its fine for custom override but might not be OK as the default css in a framework.

jxmai:
In my opinion,  we need to improve the width css but avoid introducing new attributes as the default for ease of overriding.

If we set the max-width attribute on the parent like the current PR. Every user that is currently overrding the css will get `max-width` for no reason.

On the other hand, by only improving the width. User like @NicolaIsotta  can override the css back to the original by 

```css
width:100px;
max-width: calc(100% - 10px);
```

---
In conclusion

- The width:100px is almost always required overriding

- The width: `width: 100%;
	width: -webkit-fill-available;
	width: -moz-available;`  is usually not required any overriding (they can always override the width of course)


- The width: `100px + max-width` might need overriding for the large columns use case  (user cannot easily get rid of the max-width)

melloware:
I agree with you.  I think I like your solution of the following the best and makes it the "least" intrusive change.

```css
.ui-treetable .ui-column-filter {
	width: 100%;
	width: -webkit-fill-available;
	width: -moz-available;
}
```

NicolaIsotta:
TBH I'd go with:
```css
.ui-datatable .ui-column-filter {
	box-sizing: border-box;
	width: 100%;
}
```
which works with IE11 too. But that's just my thought.

jxmai:
Maybe we can include `border-box` too, but we need to see whether or not there is any size effect if we introduce another attribute. Actually the most ideal solution will be waiting for IE and edge to support intrisic sizing. https://developer.microsoft.com/en-us/microsoft-edge/platform/status/cssintrinsicsizing/ 

All the prefix css such as `-webkit... and -moz` are experimental and quite bleeding edge

image

melloware:
I just updated the PR.  I tested in Chrome, Firefox, IE11, and Edge and all of them the filter looks great and adjusts upon resizing the window.

NicolaIsotta (approved):

src/main/resources/META-INF/resources/primefaces/treetable/treetable.css
```diff
@@ -1,4 +1,4 @@
-.ui-treetable table {
+fi.ui-treetable table {
```
jxmai: why there is `fi` ? :thinking:

melloware: Thjat is really strange.  That is not in my local file and I can't see it in my commits.

melloware: Even weird I see it in the commit but it is not in my local file.

melloware:
OK I recommitted the files and it looks like it is now fixed.

PR #2855, opened 2017-10-22
Fix #2845 - encode header as caption

### LINKED ISSUE #2845: Datatable: Use caption HTML tag
Facet `Header` is generated using a div tag while caption HTML tag exists https://www.w3schools.com/tags/tag_caption.asp That part is quite easy to rewrite, but when it comes with a paginator it's a bit more tricky since paginator is not part of the `table` itself. Is paginator should be encoded in caption tag?

Doing this makes datatable even more compliant to HTML spec and would probably help to achieve #2723

jxmai:
For #2845 

In showcase:

image

image

only header is encoded in caption. The rest still encoded in div.

### DIFF
--- a/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
+++ b/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
@@ -340,8 +340,6 @@ public class DataTableRenderer extends DataRenderer {
             encodeSortableHeaderOnReflow(context, table);
         }
 
-        encodeFacet(context, table, table.getHeader(), DataTable.HEADER_CLASS);
-
         if (hasPaginator && !paginatorPosition.equalsIgnoreCase("bottom")) {
             encodePaginatorMarkup(context, table, "top");
         }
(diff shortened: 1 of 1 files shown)


jxmai:
HTML5 standard (tabular data):

https://www.w3.org/TR/html5/tabular-data.html#the-caption-element

> The caption element represents the title of the table that is its parent, if it has a parent and that is a table element.

Rapster:
Mmmmh using caption can't be done without having any thoughts about how to handle pagination... Applying this PR will break the original rendering no?

jxmai:
@rapster I guess your main purpose is to put  pagination in caption, and I have thought about that part. To be honest, I don't know if it is the correct thing to do as I assume have already read my comment in https://github.com/primefaces/primefaces/issues/2845#issuecomment-338495848. If there is little benefit of only fixing the header, we can close this PR or to see if you have any good suggestion about it.

> Applying this PR will break the original rendering no?

No, so far so good. None of the tables in the showcase is broken. I put several screenshots here but you can always fetch my fork as another upstream to check that out

image


image


image

Rapster:
I'm not 100% convinced either that putting pagination in caption is good either, but HTML table standard does not mention pagination, and as you mentionned other librairies use it differently. But I think, PF tables are the best :-p In the showcase, they don't show header + pagination, your code should break this use case

jxmai:
> but HTML table standard does not mention pagination,

That's a valid point. I think pagination in tabular is one of the grey areas that is missing spec since everyone implements it differently 😩 

---

>  In the showcase, they don't show header + pagination, your code should break this use case

I see what you mean. I guess the header will go below pagination in this case.

jxmai:
Let's close this PR first

PR #2859, opened 2017-10-23
Fix #2364 - overlaypanel fit viewport before showing

### LINKED ISSUE #2364: Dynamic overlaypanel display outside of viewport
## 1) Environment
- Current Showcase: PrimeFaces-6.1-SNAPSHOT on Mojarra-2.2.8.

## 2) Expected behavior

Overlaypanel should always try to display itself in visible viewport. If above the element is enough space it shall use this space instead below the element.

## 3) Actual behavior

Dynamic overlaypanel does not displays itself in visible viewport. 

## 4) Steps to reproduce

Decrease viewport size in showcase and click at dynamic button

overlaypanel-dynamic-outside

jxmai:
#2364

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/overlaypanel/overlaypanel.css
+++ b/src/main/resources/META-INF/resources/primefaces/overlaypanel/overlaypanel.css
@@ -2,6 +2,7 @@
     padding: 0;
     margin: 0;
     position: absolute;
+    overflow: auto;
 }
 
 .ui-overlaypanel-content {
--- a/src/main/resources/META-INF/resources/primefaces/overlaypanel/overlaypanel.js
+++ b/src/main/resources/META-INF/resources/primefaces/overlaypanel/overlaypanel.js
@@ -192,7 +192,7 @@ PrimeFaces.widget.OverlayPanel = PrimeFaces.widget.BaseWidget.extend({
         this.targetZindex = this.targetElement.zIndex();
 
         this.align(target);
-        this._fitViewPort();
+        this.fitViewPort();
 
         //replace visibility hidden with display none for effect support, toggle marker class
         this.jq.removeClass('ui-overlay-hidden').addClass('ui-overlay-visible').css({
@@ -215,7 +215,7 @@ PrimeFaces.widget.OverlayPanel = PrimeFaces.widget.BaseWidget.extend({
         }
     },
 
-    _fitViewPort: function() {
+    fitViewPort: function() {
         var originalHeight = this.jq.data("overlayOriginalHeight");
         var originalWidth = this.jq.data("overlayOriginalWidth");
         var windowHeight = $(window).height();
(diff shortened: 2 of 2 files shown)


melloware (approved):

tandraschko:
1) methods shoudl not begin begin with "_"
2) why do we need to store the initial width? IMO it's ok if it's resizes if the content changes

jxmai:
> methods shoudl not begin begin with "_"

Done

> why do we need to store the initial width? IMO it's ok if it's resizes if the content changes

The internal design of overlay panel is only hiding the content on invisible state. If we don't store the supposed initial dimension, the width and height  cannot be recovered when there is enough room to show the overlay.

tandraschko:
Can't follow you. Could you explain in detail?

jxmai:
Say for example. If I change `fitViewPort` to be 

```js
fitViewPort: function() {
        var windowHeight = $(window).height();
        var windowWidth = $(window).width();

        if (windowHeight < this.jq.height()) {
            this.jq.height(windowHeight);
        } 
        if (windowWidth < this.jq.width()) {
            this.jq.width(windowWidth);
        } 
    },
```

If we have a narrow screen:

image


then show the overylay panel again on a larger screen. As you can see the previous resize function has affected the dimension

image

tandraschko:
Yep of course. BUT if you have the smaller screen first, the overlaypanel would never resize, also if the window size increases.

jxmai:
so what do you think ? :thinking:  just leave it as the following function and not to store initial value for  resizing afterward ?

```js
fitViewPort: function() {
        var windowHeight = $(window).height();
        var windowWidth = $(window).width();

        if (windowHeight < this.jq.height()) {
            this.jq.height(windowHeight);
        } 
        if (windowWidth < this.jq.width()) {
            this.jq.width(windowWidth);
        } 
    },
```

jxmai:
IMHO, the overlay panel should detect the room available then show the content. Leaving narrow content when there is more room available is not user friendly. Keeping in mind that user could resize windows really often when docking the browser to a corner of a screen.

tandraschko:
+1 for 

```js
fitViewPort: function() {
        var windowHeight = $(window).height();
        var windowWidth = $(window).width();

        if (windowHeight < this.jq.height()) {
            this.jq.height(windowHeight);
        } 
        if (windowWidth < this.jq.width()) {
            this.jq.width(windowWidth);
        } 
    },
```

however, it doesn't work for 100% as the margin is calculated.

jxmai:
I slightly improved the PR, and please review. It's better to use `max-width` instead of width IMO. Also the width needs to be subtracted by the offset so that it won't go out of bound.

image

jxmai:
Option 2 css only:  #2900

tandraschko:
Tried it but didn't work for 100% as the scrollbar hides a part of the overlay.
Also the height calculation didn't work for 100%, not sure. I applied a very minimalistic version of the patch now, which only fixes the width.

tandraschko:
But note: this is only a small part of #2364 and doesn't fix the issue at all.

jxmai:
Alright, the width was more problematic in the first place, so the problem should be alleviated at least.

PR #2870, opened 2017-10-26
Fix #2862: SelectOneMenu editable try and match item in list.

### LINKED ISSUE #2862: SelectOneMenu: label instead value is used when editable=true
## 1) Environment
- PrimeFaces version: 6.1
- Does it work on the newest released PrimeFaces version? Version? No try
- Does it work on the newest sources in GitHub? (Build by source -https://github.com/primefaces/primefaces/wiki/Building-From-Source) No try
- Application server + version: WebLogic 12cR2
- Affected browsers: Safari 11

## 2) Expected behaviour
selectOneMenu will put selected item's itemValue from selectItems into selectOneMenu value field by ajax or submit
...

## 3) Actual behavior
selectOneMenu put selected item's itemLabel from selectItems into selectOneMenu value field by ajax or submit
..

## 4) Steps to reproduce
- coding backing bean Map<String,String> hm and initialise it with ID/Name pair.
- coding xhtml by adding hm as selectItems with itemLabel/itemValue pair
- run application to render UI.
- select one item from selectOneMenu dropdown list and press process it to submit
- check backing bean binding value
.. 

## 5) Sample XHTML
<p:selectOneMenu id="selectCustomerId" value="#{paymentControlBean.selectedCustomerId}" effect="fold" editable="true" filter="true" filterMatchMode="startsWith">
   <p:ajax update="paymentListToolbar"/>
   <f:selectItems value="#{hm.entrySet()}" var="entry" itemValue="#{entry.key}" itemLabel="#{entry.value}"/>
</p:selectOneMenu>
..

## 6) Sample bean
Map<String,String> hm = new HashMap<String, String>();
hm.put("Cust001","John Dow");
hm.put("Cust002","Mark Polo");
hm.put("Cust003","Noel Max");
..

melloware:
Fix #2862: SelectOneMenu editable try and match item in list.

- Use the user typed in value as a default
- Search the SelectItems list and if it matches a Label "ignoring case" it uses that items value and not its label.  

Question:  Should I ignore case?  So I guess you could have two labels "mello" and "Mello" and they could have different meanings?  I just prefer to search case insensitively and not annoy a user with having to type the text in EXACTLY.

### DIFF
--- a/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
+++ b/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
@@ -48,7 +48,19 @@ public class SelectOneMenuRenderer extends SelectOneRenderer {
         if (menu.isEditable()) {
             Map<String, String> params = context.getExternalContext().getRequestParameterMap();
 
-            menu.setSubmittedValue(params.get(menu.getClientId(context) + "_editableInput"));
+            // default to user entered input
+            String editorInput = params.get(menu.getClientId(context) + "_editableInput");
+            menu.setSubmittedValue(editorInput);
+
+            // #2862 check if it matches a label and if so use the value
+            List<SelectItem> selectItems = getSelectItems(context, menu);
+            for (int i = 0; i < selectItems.size(); i++) {
+                SelectItem item = selectItems.get(i);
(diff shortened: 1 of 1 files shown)


cidylong:
Thank you for you contribution. If it was tested, please put it back to production environment.

PR #2873, opened 2017-10-26
Fix #2102: Signature add PF styles.

### LINKED ISSUE #2102: Add PrimeFaces css to p:signature
PrimeFaces version: 6.0.14 - see showcase

Instead of using the keith-wood simple css class _kbw-signature_ signature looks much better using PrimeFaces css classes _ui-inputfield ui-widget ui-state-default ui-corner-all_.

On readonly one could use _ui-widget ui-widget-content ui-corner-all_.

The _kbw-signature_ class is the single css rule in a separate css file. Unnecessary if using PrimeFaces classes.

You can see a (responsive )demo here: 

https://github.com/djmj/signature
http://htmlpreview.github.io/?https://github.com/djmj/signature/blob/master/test/index.html

melloware:
Fix #2102: Signature add PF styles.

### DIFF
--- a/src/main/java-templates/org/primefaces/component/signature/SignatureTemplate.java
+++ b/src/main/java-templates/org/primefaces/component/signature/SignatureTemplate.java
@@ -2,6 +2,9 @@ import javax.el.ValueExpression;
 import javax.faces.context.FacesContext;
 import org.primefaces.component.api.UITree;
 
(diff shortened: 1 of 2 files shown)


tandraschko:
Would you add screenshots please?

melloware:
Before:

image

After (rounded corners):

image

tandraschko:
Thanks

PR #2876, opened 2017-10-27
Fix #2331: SelectOneMenu added keyup/keydown for Editable.

### LINKED ISSUE #2331: SelectOneMenu: onkeyup and onkeydown not effective
## 1) Environment
- PrimeFaces version:
6.0, 6.1

- Does it work on the newest released PrimeFaces version?
no

no

Apache Tomcat 8.5.14

## 2) Expected behavior
When selectOneMenu component has focus, configured onkeydown/onkeyup events should trigger when pressing down/releasing a key.
I'd expect these events to be triggered at least for editable selectOneMenus when typing some text. (This way, I'd be able to use it as dataTable filter component allowing custom input and filtering while typing.)
...

## 3) Actual behavior
Events ain't triggered.
..

## 4) Steps to reproduce
Add a selectOneMenu component to your view with javascript filled onkeyup and onkeydown attributes.
In browser, focus that component and press (the one and only) any key. Your expected javascripts do not get executed. In example below, you will only see alert message 'change' popup when changing by mous or on blur (for editable component after you typed something).

## 5) Sample XHTML
Only event fired properly is onchange event:
```
<p:selectOneMenu editable="false" onkeydown="alert('down')"
	onkeyup="alert('up')" onchange="alert('change')">
	<f:selectItem itemLabel="1" itemValue="1"/>
	<f:selectItem itemLabel="2" itemValue="2"/>
</p:selectOneMenu>

<p:selectOneMenu editable="true" onkeydown="alert('down')"
	onkeyup="alert('up')" onchange="alert('change')">
	<f:selectItem itemLabel="a" itemValue="a"/>
	<f:selectItem itemLabel="b" itemValue="b"/>
</p:selectOneMenu>
```

This issue has already been discussed in some forum threads:
https://forum.primefaces.org/viewtopic.php?f=3&t=42733
https://forum.primefaces.org/viewtopic.php?f=3&t=41793
https://forum.primefaces.org/viewtopic.php?f=3&t=35037

melloware:
Fix #2331: SelectOneMenu added keyup/keydown for Editable.

Tested and verified using showcase code and verify the console printed out the keyup and keydown events.

```xml
<p:selectOneMenu id="city" value="#{selectOneMenuView.city}" effect="fold" editable="true" onkeyup="console.log('keyup');" onkeydown="console.log('keydown');">
```

### DIFF
--- a/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
+++ b/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
@@ -225,6 +225,15 @@ public class SelectOneMenuRenderer extends SelectOneRenderer {
                 writer.writeAttribute("placeholder", menu.getPlaceholder(), null);
             }
 
+            if (menu.getOnkeydown() != null) {
+                writer.writeAttribute("onkeydown", menu.getOnkeydown(), null);
+                this.renderDomEvent(context, menu, "onkeydown", "keydown", "keydown", null);
+            }
+            if (menu.getOnkeyup() != null) {
+                writer.writeAttribute("onkeyup", menu.getOnkeyup(), null);
+                this.renderDomEvent(context, menu, "onkeyup", "keyup", "keyup", null);
+            }
+
             writer.endElement("input");
         }
         else {


tandraschko:
isn't it now rendered twice on 2 different element?

melloware:
From all of my testing keyup and keydown were never actually implemented on the select one but now when you have an editable="true" type Select One the keydown and keyup are captured as you type.  I don't know if the original intent of the SelectOne was to have keyup keydown on the actual SelectOne but that is definitely not being rendered.

PR #2877, opened 2017-10-27
Fix #753: SelectOneMenu alignment issues for Editable.

### LINKED ISSUE #753: SelectOneMenu: editable="true" causes misalignment
Hi, the following happens when you have multiple selectOneMenu next to each other, but one with editable property. (Code copied from the showcase).

Version: 5.3 - RC1

```
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:h="http://xmlns.jcp.org/jsf/html" xmlns:f="http://xmlns.jcp.org/jsf/core"
    xmlns:p="http://primefaces.org/ui" xmlns:pm="http://primefaces.org/mobile" xmlns:ui="http://xmlns.jcp.org/jsf/facelets">

    <h:head>
        <title>TestPage</title>
    </h:head>
    <h:body>

     <div style="border: 1px solid red; width: 100%">

        <p:selectOneMenu id="console" value="#{selectOneMenuView.console}">
            <f:selectItem itemLabel="Select One" itemValue="" />
            <f:selectItem itemLabel="Xbox One" itemValue="Xbox One" />
            <f:selectItem itemLabel="PS4" itemValue="PS4" />
            <f:selectItem itemLabel="Wii U" itemValue="Wii U" />
        </p:selectOneMenu>

        <p:selectOneMenu id="car" value="#{selectOneMenuView.car}">
            <f:selectItem itemLabel="Select One" itemValue="" />
            <f:selectItems value="#{selectOneMenuView.cars}" />
        </p:selectOneMenu>

        <p:selectOneMenu id="city" value="#{selectOneMenuView.city}" editable="true">
            <f:selectItem itemLabel="Select One" itemValue="" />
            <f:selectItems value="#{selectOneMenuView.cities}" />
        </p:selectOneMenu>


     </div>         
    </h:body>
</html>

```

Issue

melloware:
Verified the veritical-align:top fixes it with no side effects.

image

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/forms/forms.css
+++ b/src/main/resources/META-INF/resources/primefaces/forms/forms.css
@@ -224,6 +224,7 @@
     zoom: 1;
     cursor: pointer;
     padding-right: 2em;
+    vertical-align: top;
 }
 
 .ui-selectonemenu .ui-selectonemenu-trigger {


tandraschko:
Not sure if we should "hack" it this way. Why is one aligned bottom and the other aligned top? there must be any reason for this.

melloware:
Agreed maybe @sudheerj can chime in?

jxmai:
This is because `inline-block`is not doing the alignment job properly. It does not choose the baseline properly in this case when the children contains `input` (it is hard to replicate but it could happen). The workaround is OK though since it could be a common solution  and not `hack`. 

Here is the similar issue: https://stackoverflow.com/questions/19366401/my-inline-block-elements-are-not-lining-up-properly

> This is a common issue involving inline-block elements. In this case, the default value of the vertical-align property is baseline. If you change the value to top, it will behave as expected.

tandraschko:
All right, thanks!

PR #2891, opened 2017-10-31
Fix #89: DataTable filter IE/Edge "x" button.

### LINKED ISSUE #89: DataTable filter input clear icon on IE10 doesn't trigger filtering
With internet explorer 10 in each input field a "X" is drawn at the end of the field to clear the content. This is also drawn in the input fields for the column filters of a datatable. If I type something in the filter input field, the datatable contents will be filter after a delay. If I clear the content with backspace, delete or similar, all content will be displayed like before. If I clear the fields content with the "X", the filtered content remains.

melloware:
Fix #89: DataTable filter IE/Edge "x" button.


melloware:
Used this stack overflow post to find the solution

https://stackoverflow.com/questions/14498396/event-fired-when-clearing-text-input-on-ie10-with-clear-icon

PR #2898, opened 2017-11-03
Fix #873 - not to deselect all when selectManyMenu is under checkbox …

### LINKED ISSUE #873: selectManyMenu only selects one when clicking on the label
To reproduce visit http://www.primefaces.org/showcase/ui/input/manyMenu.xhtml

In the advanced menu click on the select item label, ie. Afterdark, Afternoon etc.

Expected Result:
All clicked items should remain selected.

Actual Result:
Only the last clicked item is selected

jxmai:
…mode

Fix #873 

image

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/forms/forms.selectmanymenu.js
+++ b/src/main/resources/META-INF/resources/primefaces/forms/forms.selectmanymenu.js
@@ -27,7 +27,7 @@ PrimeFaces.widget.SelectManyMenu = PrimeFaces.widget.SelectListbox.extend({
                 unchanged = (!metaKey && selectedItems.length === 1 && selectedItems.index() === item.index());
 
                 if(!e.shiftKey) {
-                    if(!metaKey) {
+                    if(!metaKey && !$this.cfg.showCheckbox) {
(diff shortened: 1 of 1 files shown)


jxmai:
There is a bug. Hold on. To reopen when it is ready.

jxmai:
False alarm. I thought the shift key behavior was changed, so I did the side by side comparison of the old code versus the new one. The result was OK (the same as before). 

The key of this PR is not to unselect all even if meta key is not pressed when checkbox is set

Before this pull:

```js
 if(!metaKey) {
    $this.unselectAll();
}
```

so if ctrl key is not pressed, the unselect all will be triggered which leads to the inconsistent  behaviors.

After this pull:

```js
if(!metaKey && !$this.cfg.showCheckbox) {
  $this.unselectAll();
}
```

We should take checkbox into consideration as jasonex7 described in the issue

> I think setting showCheckbox="true" should by default keep the selected items instead of just select the last one and clearing the previous selected ones.

jasonex7:
Thanks @jxmai I think that is the correct behavior from the user's perspective.

jxmai:
You're welcome :smile:

PR #2899, opened 2017-11-03
Fix #2130 - dialog max height and max width should fit viewport

### LINKED ISSUE #2130: Dialog locked to the top when minimize screen.
## 1) Environment
- PrimeFaces version:5.3
Its reproduce on the showcase (The latest version).
- Affected browsers:  Chrome 

## 2) Expected behavior
When open Dialog and minimize the screen and drag the Dialog to top of the Screen, The Dialog is not locked to the top and it is possible to close the Dialog.

## 3) Actual behavior
When open Dialog and minimize the screen and drag the Dialog to top of the Screen The Dialog is locked to the top and it is not possible to close the Dialog.

## 4) Steps to reproduce
dialog1
dialog2
dialog3

Open Dialog , minimize the screen ,Drag the Dialog to the Top try To close the Dialog or Drag it again.

jxmai:
Fix #2130 

The dialog can never go out of bound :hand: 

image

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/dialog/dialog.css
+++ b/src/main/resources/META-INF/resources/primefaces/dialog/dialog.css
@@ -1,4 +1,4 @@
-.ui-dialog { position: fixed; padding: 0; overflow: hidden; display:none; }
+.ui-dialog { position: fixed; padding: 0; overflow: hidden; display:none; max-height: 100vh; max-width: 100vw; }
 .ui-dialog.ui-dialog-absolute { position: absolute; }
 .ui-dialog.ui-overlay-hidden {display:block;}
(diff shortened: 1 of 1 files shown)


tandraschko:
We should really think about that one. Dialogs MIGHT be bigger as the viewport if it has much content.

jxmai:
Actually,  If dialog is bigger than viewport, this bug is **more visible** as we don't need to resize browser space at all to replicate the same bug.

image

after dragged:

image

with my PR, the initial screen looks exactly the same (everything including the scrollbar)

image

and the bug will be solved.

This PR is not `feature`  but `defect` IMHO.

jxmai:
> Dialogs MIGHT be bigger as the viewport if it has much content.

viewport css won't chop off anything if there is any overflowed content. There is no need to worry about that.

PR #2900, opened 2017-11-03
Fix #2364 - overlay panel should fit viewport using purely css

### LINKED ISSUE #2364: Dynamic overlaypanel display outside of viewport

jxmai:
Fix #2364 

This pull is an alternative option aside from #2859. Personally I prefer this approach since:

1. less js better performance

2. the overlay dialog is responsive now which is a big plus.

3. less code, less bug

There is no need to use overflow unlike #2859, and #2859  is not responsive

Notice that the reason we cannot use `max-width: 100vw` but to use `calc(100vw - (100vw - 100%));` is because we need to remove the width of the scrollbar if it exists.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/overlaypanel/overlaypanel.css
+++ b/src/main/resources/META-INF/resources/primefaces/overlaypanel/overlaypanel.css
@@ -2,6 +2,8 @@
     padding: 0;
     margin: 0;
     position: absolute;
+    max-width: calc(100vw - (100vw - 100%));
+    max-height: 100vh;
 }
 
 .ui-overlaypanel-content {


tandraschko:
Does this work on IE9?

jxmai:
Yes.

IE 9 has partial support  but it does not matter in our case since we are not using `vmin` or `vmax`


http://caniuse.com/#feat=viewport-units (click showall to see the full picture)

image



> Partial support in IE9 refers to supporting "vm" instead of "vmin".

> Partial support refers to not supporting the "vmax" unit.

jxmai:
And `calc` will work on IE 9+

tandraschko:
it works almost fine but it doesn't calculate the margin e.g. to the left side. No idea if you can do that with plain css. So the scrollbar hides a part of the overlaypanel currently if you minimize.

tandraschko:
Will close it for now but it would be great if we could improve it to work with CSS only.

PR #2906, opened 2017-11-07
Fix #2905

### LINKED ISSUE #2905: TextEditor: inside TabView don't load hidden input field
## 1) Environment
- PrimeFaces version: 6.1
- Does it work on the newest sources in GitHub? Didn't try yet.
- Application server + version: Tomcat 8.0
- Affected browsers: Chrome and IE11 (didn't test any other)

## 2) Expected behavior

If a texteditor is placed inside a tabview's tab, it should load the hidden input field eagerly, even if don't open the tab where the texteditor is placed.

## 3) Actual behavior

If a texteditor is placed inside a tabview's tab, the hidden input field is not loaded unless I click the tab. Then, if I submit the form without clicking the tab, it will set a blank value.

This is the first tab that is displayed:
image

Before I click the tab "Descrição", which contains a texteditor, the hidden input field is blank:
image

If I click the tab containing the texteditor, it's value is loaded:
image
image

## 4) Steps to reproduce

1. Create a form with a text editor input;
2. Place it inside the second tab of a tabview;
3. Set the value of the texteditor in the bean init method;
4. Access the page;
5. Submit the form and the value will be overriden;

## 5) Sample XHTML

```
<h:form id="form">
<p:tabView id="tabView" >
<p:tab id="fisrt" title="First" >
...
</p:tab>
<p:tab id="second" title="Second" >
<p:textEditor
				id="textEditor"
				widgetVar="textEditorWidget"
				value="#{sampleBean.textEditorValue}" />
</p:tab>
</p:tabView>
</h:form>
```

## 6) Sample bean

...

ps.: I can update the issue to provide more content to reproduce if needed.

jxmai:
fix #2905 

Sample: https://github.com/primefaces/primefaces/issues/2905#issuecomment-342602930

After this patch:

```
<p>test</p>
<p>test2</p>
test3
```

### DIFF
--- a/src/main/java/org/primefaces/component/texteditor/TextEditorRenderer.java
+++ b/src/main/java/org/primefaces/component/texteditor/TextEditorRenderer.java
@@ -88,7 +88,7 @@ public class TextEditorRenderer extends CoreRenderer {
         writer.writeAttribute("name", inputId, null);
         // #2905
         if (valueToRender != null) {
-            writer.writeAttribute("value", "<p>" + valueToRender + "</p>", null);
+            writer.writeAttribute("value", valueToRender, null);
         }
         writer.endElement("input");
 
--- a/src/main/java/org/primefaces/component/texteditor/TextEditorRenderer.java
+++ b/src/main/java/org/primefaces/component/texteditor/TextEditorRenderer.java
@@ -86,6 +86,10 @@ public class TextEditorRenderer extends CoreRenderer {
         writer.startElement("input", null);
         writer.writeAttribute("type", "hidden", null);
         writer.writeAttribute("name", inputId, null);
+        // #2905
+        if (valueToRender != null) {
+            writer.writeAttribute("value", "<p>" + valueToRender + "</p>", null);
+        }
         writer.endElement("input");
 
         writer.endElement("div");


jxmai:
we should consider the proposal in https://github.com/primefaces/primefaces/pull/2908 for https://github.com/primefaces/primefaces/issues/2907 before this pull. Since it is wrong to have an extra p tag in https://github.com/primefaces/primefaces/pull/2906/files#diff-378536e9c4e2f411833c2eabc3e43158R90 in order to be in consistent with the `input` generated by the texteditor.

We should not alter the original value by putting the extra p tag around. It was definitely a mistake in Primefaces not in quill editor.

image

Notice that we even have a comma at the end of a variable delaration... :sweat: 

https://github.com/primefaces/primefaces/blob/1441ba142f660d26310d0283d520120fd60de5d2/src/main/resources/META-INF/resources/primefaces/texteditor/texteditor.js#L141

After #2907 is addressed, I will modify this PR to remove those extra p tag.

jxmai:
> After #2907 is addressed, I will modify this PR to remove those extra p tag.

I changed by mind :smile:  The mistake in #2907  is too obvious, so I corrected the PR regardless.

Issue #2874, opened 2017-10-27
DataTables scroll bars bug when using scrollable="true" and selectionMode="multiple"


mertsincan:
Reported By PRO User;
> Hi,
We have DataTable and we use scrollable="true" & selectionMode="multiple" (both are required).
The problem occurs while we move the mouse across DataTables rows (hover) the scroll bars are appear & disappear.

Adam-1:
@mertsincan 
Please note this issue doesn't show in newer chrome versions, maybe they fixed it for 62+ versions.
(Seems the width was also computing the box shadow)

mertsincan:
Thanks @Adam-1 ;) I'll check it.

mertsincan:
I'll revert my commit. Thanks again!

PR #2923, opened 2017-11-10
#2874 - css optimization

jxmai:
css optimation for #2874 (Aristo theme)

Analysis in https://github.com/primefaces/primefaces/commit/0f77cf7c1fbef89c33b746d221a2b53fd603da85#commitcomment-25240616

Tested by custom showcase (Aristo theme) Hover and unhover won't trigger flickering.

image

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces-aristo/theme.css
+++ b/src/main/resources/META-INF/resources/primefaces-aristo/theme.css
@@ -1238,9 +1238,9 @@ a {
 }
 
 .ui-datatable tr.ui-datatable-selectable.ui-state-hover {
+    box-shadow: none;
     -moz-box-shadow: none;
     -webkit-box-shadow: none;
-     box-shadow: none;
 }
 
 /** Liferay Support **/


tandraschko:
Not sure about this one. Normally box-shadow should be after the browser specific ones...

jxmai:
Alright, I see your point that the goal is to use `box-shadow` rather than browser specific rules if possible.

---

JFYI: 

The first two rules are rarely used but they can be helpful for the initial versions of Firefox or Chrome

https://caniuse.com/#search=box-shadow

snapshot1


https://hacks.mozilla.org/2009/06/moz-box-shadow/

> To achieve the biggest possible coverage, it is advisable to define all three, the -moz, -webkit, and standard CSS3 syntax in parallel. Applicable browsers will then pick and adhere to the ones they support. For example:


```css
-moz-box-shadow: 1px 1px 10px #00f; 
 -webkit-box-shadow: 1px 1px 10px #00f;  
box-shadow: 1px 1px 10px #00f;
```






```

PR #2928, opened 2017-11-13
Fix #2927 - MenuButton: Missing ui-widget class

### LINKED ISSUE #2927: MenuButton: Missing ui-widget class

Rapster:


tandraschko:
I assume that the styling will be the same after that change?

Rapster:
Yes it will be the same, if you're wondering, you can still try it on the showcase modifying the CSS online

PR #2936, opened 2017-11-14
Fix #2901: Datatable scrollable styles update.

### LINKED ISSUE #2901: Theme: DataTable scrollable - header above scrollbar wrong styling
Open showcase https://www.primefaces.org/showcase/ui/data/datatable/scroll.xhtml and for example use theme blitzer. You will notice a red header above the scrollbar from the theme.

At header _ui-widget-header ui-datatable-scrollable-header_ just add _ui-state-default_. Same for footer.

scrollable

melloware:


tandraschko:
Not sure if i like the fix, a header is actually stateless. Maybe it should be fixex by the theme or maybe something else is wrong.

melloware:
Yeah its weird it seems to be in all themes.  They all exhibit this behavior it just looks extra weird on the blitzer theme.  Most other themes the styling looks OK because the colors match OK.

I never noticed this before because the themes I use it looks OK on to have it colored matching the header.

tandraschko:
not sure if we should use this state class for this purpose. We have to check other components or the old jquery-ui themeguide.

melloware:
Agreed.

djmj:
@melloware  It is just random that the styles match in other themes. We have some custom themes where it looks like blitzer.

@tandraschko  I do not understand what you mean _"header is stateless"_? Datatable header uses class _ui-state-default_ as seen in DOM of showcase. So the scrollable section should use it too.

melloware:
@djmj I think what @tandraschko means that state is for the headers of columns like "active" or "default" etc.  He is saying that the scroll gutter has no state.

PR #2945, opened 2017-11-15
Fix #2944

### LINKED ISSUE #2944: p:fileDownload not working with p:commandLink and f:param under some circumstance
## 1) Environment
- PrimeFaces version:
6.1
No. 6.1
No. 6.2-SNAPSHOT
WildFly 10.1.0.Final
- Affected browsers:
tested with IE11 and FF58
## 2) Expected behavior
Using p:fileDownload in combination with p:commandLink (ajax="false") and f:param should work.
## 3) Actual behavior
Does not work under some circumstances. At least if one adds a f:param to the p:commandLink that has a value containing a JavaScript character escape sequence e.g. \u. As can be seen with test project it works with standard JSF h:commandLink because backslashes in the rendered JS-code are escaped.
## 4) Steps to reproduce
[primefaces-test](https://github.com/irieill/primefaces-test)
## 5) Sample XHTML
[primefaces-test](https://github.com/irieill/primefaces-test)
## 6) Sample bean
[primefaces-test](https://github.com/irieill/primefaces-test)

## Fix
To fix this one must escape the f:param [value](https://github.com/primefaces/primefaces/blob/9f108135f303756c93b93a855423be2f3ec13de2/src/main/java/org/primefaces/renderkit/CoreRenderer.java#L463), maybe like it is done in [mojarra](https://github.com/javaserverfaces/mojarra/blob/c212abad39a683e6b80e221e9839b97f664d3b03/impl/src/main/java/com/sun/faces/renderkit/RenderKitUtils.java#L1543).

irieill:


tandraschko:
Couldn't we just reuse some of our escape utils? Maybe someone can check it? @melloware @jxmai @rapster ?

melloware:
Yes I think this can be fixed another way.  Let me submit a PR.

melloware:
@irieill please review my PR.  https://github.com/primefaces/primefaces/pull/2947

PR #2976, opened 2017-11-20
Fix #63 - Support pre and post processors in CSV and XML exporters

### LINKED ISSUE #63: DataExporter Pre/PostProcessor on CSV and XML documents
Hello,

We noticed that the Pre/PostProcessor option of the DataExporter component is only implemented for PDF and XLS documents.
See http://forum.primefaces.org/viewtopic.php?f=3&t=2422

We consider that this limitation is damageable, because Pre or Post processing on CSV or XML documents would be useful in many cases, even by simply getting the outputstream object. For example, it would have help us to resolve this issue : https://code.google.com/p/primefaces/issues/detail?id=4714 (admittedly in a tricky way)

So we suggest to add the DataExporter Pre/Post processor functionality for CSV and XML documents, even if the only proper option is to pass the stream.

Thanks.

Yoan.Tournade

firsthour:
This pull request addresses issue #63 by giving CSVExporter and XMLExporter the ability to handle pre and post processors similar to Excel and PDF.

The CSV and XML exporters passed around a PrintWriter object which I replaced with a StringBuilder, and the writer.write() calls are replaced with a builder.append(). There is a single write() at the end of the export method before its flush() and close().


melloware:
I like this PR but it seems the checks are failing.  Are you sure you are passing the checkstyle requirements by doing a "mvn clean package" locally?

melloware:
Looks like 4 checkstyle violations.

```
[ERROR] src/main/java/org/primefaces/component/export/CSVExporter.java:[45,11] (whitespace) WhitespaceAround: 'if' is not followed by whitespace.
[ERROR] src/main/java/org/primefaces/component/export/CSVExporter.java:[65,11] (whitespace) WhitespaceAround: 'if' is not followed by whitespace.
[ERROR] src/main/java/org/primefaces/component/export/XMLExporter.java:[43,11] (whitespace) WhitespaceAround: 'if' is not followed by whitespace.
[ERROR] src/main/java/org/primefaces/component/export/XMLExporter.java:[64,11] (whitespace) WhitespaceAround: 'if' is not followed by whitespace.
```

firsthour:
Didn't realize that was a thing, I'm on it.

melloware (approved):

tandraschko:
Why do you introduced the StringBuilder? Writing directly to the stream is a better approach actually.

firsthour:
It's been a few months but I believe the reason I did this is because I wanted to give the postProcessor a StringBuilder giving it the opportunity to modify the currently built document, instead of handing it a stream that it can not modify anymore.

tandraschko:
How or what do you modify the whole content in a string buidler?

firsthour:
StringBuilder provides a whole bunch of methods for inserting and replacement text.

If you're wondering about a use-case, we were required to escape Excel formulas in Excel and CSV outputs, basically turn =A1 into '=A1. This was already possible with Excel but this change enables this for CSV too.

Looking at org.primefaces.component.export.ExcelExporter, it uses the typical POI Workbook object which allows modification of any cell, basically StringBuilder gives CSV and XML a similar post-processorable object to work with before writing to the stream.

tandraschko:
Understood and merged! thanks.

Jonnathan89:
Good afternoon, friend sorry to be able to use the post processor with csv files, it is necessary to modify the general files of the primefaces.

firsthour:
No, you just need to set the postProcessor attribute of the p:dataExporter element pointing to a method with this signature:

`public void postProcessCsv(Object document)`

You can then cast the document to a StringBuilder and do whatever you want with it.

For example we cast the document to StringBuilder and then pass it to a post processor that replaces "=" with "'=" to disable formulas:

```
		if(document != null) {
			String content = document.toString();
			String newContent = StringUtils.replace(content, "\"=", "\"'=");
			document.replace(0, document.length(), newContent);
		}
```

Jonnathan89:
<p:dataExporter type="csv" target="tblConsultaDinamica" fileName="#{consultaDinamica.consulta.nombre}" postProcessor="#{consultaDinamica.modificarLetraExport}"/>


public void modificarLetraExport(Object document) {
        System.out.println(" ----------> 1 ");
        HSSFWorkbook wb = (HSSFWorkbook) document;
        HSSFSheet sheet = wb.getSheetAt(0);      
 
        for (Row row : sheet) {
            System.out.println(" ----------> 2 ");
            for (Cell cell : row) {
                System.out.println(" ----------> 3 ");
                String celda = cell.getStringCellValue() ;
                cell.setCellValue(celda.replaceAll("\"", ""));                
                System.out.println(" ----------> 4   " + celda);
            }
        }
    }

This is what I am doing, but executing the button does not start executing the method.

my version of primefaces is 5.2

melloware:
If you look at the ticket: https://github.com/primefaces/primefaces/issues/63

You will see this feature was not implemented until PF 7.0.  So this will do you no good on PF 5.2. I suggest upgrading.

PR #2985, opened 2017-11-22
Fix #2949: Omega theme should not BOLD button font.

### LINKED ISSUE #2949: Button: icon should have normal font-weight
## 1) Environment
- PrimeFaces version: 6.1 with Omega theme
- Application server + version: All
- Affected browsers: All

## 2) Expected behavior
Icon inside buttons/commandbuttons should have a normal font-weight.

## 3) Actual behavior
Icon has font-weight: bold. This worsens the look of some icons, especially on small buttons.

## 4) Steps to reproduce
See provided xhtml. The rendered result is:
cattura

## 5) Sample XHTML
```xhtml
<!DOCTYPE html>
    <h:head id="head1">
		<title>Test</title>

		<style type="text/css">
			.normalWeight {
				font-weight: normal !important;
			}
		</style>
    </h:head>
    <h:body id="body1">

		<p:button icon="fa fa-trash-o"/>

		<p:button icon="fa fa-trash-o normalWeight"/>

	</h:body>
</html>
```

## 6) Sample bean
Not needed.

melloware:


onursenture:
This change affects both text and icon. We should only make the icon font-weight: normal. Can you make this change and send PR again?

melloware:
Hmm how do i separate the icon from the text.  Both are reference .span and the icon and the text are both in spans?

NicolaIsotta:
@melloware 
You can exclude the icon by changing the current selector to
`.ui-button, .ui-button span:not(.fa), button.ui-button.ui-state-default span:not(.fa), .ui-button.ui-state-default span:not(.fa)`
or
`.ui-button, .ui-button span:not(.ui-icon), button.ui-button.ui-state-default span:not(.ui-icon), .ui-button.ui-state-default span:not(.ui-icon)`

melloware:
@NicolaIsotta I think we only need to handle the FA scenario and the ui-icon are sprites right and won't be affected by the font size?

melloware:
oh nevermind i see the style is ui-icon to capture even if its FontAwesome.  I like the second one its more explicit.

melloware:
Commit pushed.

melloware:
Closing this PR it was handled by this commit: https://github.com/primefaces/primefaces/commit/875b8a7243b28c43105ad3b08abdf95773e27632#diff-25d902c24283ab8cfbac54dfa101ad31

PR #2991, opened 2017-11-23
Fix #2988: More reliable JSF version detection

### LINKED ISSUE #2988: FileUpload doesn't work on JSF 2.3
FileUpload should work but it does not and I still don't know why but I tried many configuration variants and found out this stack trace when I set this in web.xml and use the `@FacesConfig(version = Version.JSF_2_3)` annotation. (But commons variant does still nothing.)
```
  <context-param>
    <param-name>javax.faces.ENABLE_CDI_RESOLVER_CHAIN</param-name>
  </context-param>
  <context-param>
    <param-name>primefaces.UPLOADER</param-name>
    <param-value>native</param-value>
  </context-param>
```
```
[2017-11-23T00:31:54.932+0100] [] [INFO] [] [javax.enterprise.resource.webcontainer.jsf.context] [tid: _ThreadID=28 _ThreadName=http-thread-pool::http-listener(10)] [timeMillis: 1511393514932] [levelValue: 800] [[
  javax.faces.FacesException: native uploader requires at least a JSF 2.2 runtime
javax.faces.FacesException: native uploader requires at least a JSF 2.2 runtime
        at org.primefaces.component.fileupload.FileUploadRenderer.decode(FileUploadRenderer.java:59)
        at javax.faces.component.UIComponentBase.decode(UIComponentBase.java:533)
        at javax.faces.component.UIInput.decode(UIInput.java:822)
        at javax.faces.component.UIComponentBase.processDecodes(UIComponentBase.java:910)
        at javax.faces.component.UIInput.processDecodes(UIInput.java:716)
        at com.sun.faces.context.PartialViewContextImpl$PhaseAwareVisitCallback.visit(PartialViewContextImpl.java:629)
        at com.sun.faces.component.visit.PartialVisitContext.invokeVisitCallback(PartialVisitContext.java:183)
        at javax.faces.component.UIComponent.visitTree(UIComponent.java:1479)
        at javax.faces.component.UIComponent.visitTree(UIComponent.java:1491)
        at javax.faces.component.UIComponent.visitTree(UIComponent.java:1491)
        at javax.faces.component.UIForm.visitTree(UIForm.java:379)
        at javax.faces.component.UIComponent.visitTree(UIComponent.java:1491)
        at javax.faces.component.UIComponent.visitTree(UIComponent.java:1491)
        at com.sun.faces.context.PartialViewContextImpl.processComponents(PartialViewContextImpl.java:423)
        at com.sun.faces.context.PartialViewContextImpl.processPartial(PartialViewContextImpl.java:284)
        at org.primefaces.context.PrimePartialViewContext.processPartial(PrimePartialViewContext.java:57)
        at javax.faces.component.UIViewRoot.processDecodes(UIViewRoot.java:1031)
        at com.sun.faces.lifecycle.ApplyRequestValuesPhase.execute(ApplyRequestValuesPhase.java:79)
        at com.sun.faces.lifecycle.Phase.doPhase(Phase.java:100)
        at com.sun.faces.lifecycle.LifecycleImpl.execute(LifecycleImpl.java:201)
        at javax.faces.webapp.FacesServlet.executeLifecyle(FacesServlet.java:731)
        at javax.faces.webapp.FacesServlet.service(FacesServlet.java:475)
...
```

## 1) Environment
- PrimeFaces version: 6.1
- Does it work on the newest sources in GitHub? No, tested against commit ca1d82f18c9fd78ed1435fde3f967b0c77ee3df6
- Application server + version: Payara 5...-SNAPSHOT, micro
- Affected browsers: all

## 2) Expected behavior
No stacktrace and working FileUpload component in both variants (native/commons).

## 3) Actual behavior
Request is sent but causes only a stacktrace for native variant and nothing for commons variant.

## 4) Steps to reproduce
web.xml (note: result is same with or without the filter-mapping):
```
<!--   <filter> -->
<!--     <filter-name>PrimeFaces FileUpload Filter</filter-name> -->
<!--     <filter-class>org.primefaces.webapp.filter.FileUploadFilter</filter-class> -->
<!--   </filter> -->
<!--   <filter-mapping> -->
<!--     <filter-name>PrimeFaces FileUpload Filter</filter-name> -->
<!--     <servlet-name>Faces Servlet</servlet-name> -->
<!--     <dispatcher>FORWARD</dispatcher> -->
<!--     <dispatcher>REQUEST</dispatcher> -->
<!--   </filter-mapping> -->
  <context-param>
    <param-name>primefaces.UPLOADER</param-name>
    <param-value>native</param-value>
  </context-param>
```

```
@FacesConfig(version = Version.JSF_2_3)
@Named
public class JsfConfiguration {
}
```

## 5) Sample XHTML
```
        <h:panelGrid columns="2">
          <h:outputText value="Input file: " />
          <p:fileUpload fileUploadListener="#{xImport.uploadFile}" mode="advanced" dragDropSupport="true"
           update="messages" sizeLimit="10000000" sequential="true" />
        </h:panelGrid>
      </h:form>
```
..

## 6) Sample bean
```
@Named
@ViewScoped // @RequestScoped - no difference
public class XImport implements Serializable {

  private static final long serialVersionUID = -6731613465171745866L;

  private static final Logger LOG = LogManager.getLogger(XImport.class);

  @Inject

  public void uploadFile(final FileUploadEvent event) {
    final UploadedFile file = event.getFile();
    LOG.debug("File size: {}", file.getSize());
    final FacesMessage message = new FacesMessage("Succesful", file.getFileName() + " is uploaded.");
    this.facesContext.addMessage(null, message);
  }
```

Final question: sources still must support Java 6.0?

dmatej:
Implemented new class which is used to detect and report anywhere it is needed.
Added also support for future JSF versions (optimistic).

### DIFF
--- a/src/main/java/org/primefaces/config/JsfVersionDetector.java
+++ b/src/main/java/org/primefaces/config/JsfVersionDetector.java
@@ -1,0 +1,173 @@
+/**
+ * Copyright 2009-2017 PrimeTek.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.primefaces.config;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import javax.faces.context.FacesContext;
+
+/**
+ * Detects the version of the JSF implementation.
+ */
+public class JsfVersionDetector {
+
+    private static final Pattern PATTERN = Pattern.compile("([0-9]+\\.[0-9]+).*");
+    private static final String VERSION_2_1 = "2.1";
+    private static final String VERSION_2_2 = "2.2";
+    private static final String VERSION_2_3 = "2.3";
+
+    private String declaredVersion;
+    private String heuristicVersion;
+    private boolean jsf23Supported;
+    private boolean jsf22Supported;
+    private boolean jsf21Supported;
+
+
+    /**
+     * Constructor - initializes all fields.
+     */
+    public JsfVersionDetector() {
+        this.declaredVersion = FacesContext.class.getPackage().getImplementationVersion();
+        this.heuristicVersion = detectHeuristicVersion();
+        final String version = getMajorAndMinorVersion(this.declaredVersion, this.heuristicVersion);
+        if (version == null) {
+            return;
+        }
+        this.jsf21Supported = VERSION_2_1.compareTo(version) <= 0;
+        this.jsf22Supported = VERSION_2_2.compareTo(version) <= 0;
+        this.jsf23Supported = VERSION_2_3.compareTo(version) <= 0;
+    }
+
+
+    private static String getMajorAndMinorVersion(final String declared, final String heuristic) {
+        final String version = declared == null ? heuristic : declared;
+        if (version == null) {
+            return null;
+        }
+        final Matcher matcher = PATTERN.matcher(version);
+        if (matcher.matches()) {
+            return matcher.group(1);
+        }
+        return null;
+    }
+
+
+    private String detectHeuristicVersion() {
+        if (detectJSF23()) {
+            return VERSION_2_3;
+        }
+        if (detectJSF22()) {
+            return VERSION_2_2;
+        }
+        if (detectJSF21()) {
+            return VERSION_2_1;
+        }
+        // not detected -> other
+        return null;
+    }
+
+
+    private boolean detectJSF23() {
+        return checkJsfVersion("javax.faces.component.UIImportConstants");
+    }
+
+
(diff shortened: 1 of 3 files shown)


melloware (approved):

tandraschko:
My thouhts:
- Its good to only use classForNane instead package versions
- clasa.forName is actually not good, we need to use the context class loader probably buf this is currently also wrong
- the filter should just reuse PrimeConfig
- we dont need a need class for detecting the version. One jsf version=one classForName statements which can stay im PrimeConfig

I can do the classloader part in 2,5 weeks but would be good if you can do the other points

dmatej:
I don't know why it was implemented in this way originally - I only created new class and because the original implementation was somehow weird, I aggregated all informations in one class.
Using classloader will be also problematic because new JSF allows something like compatibility regimes. 
Then the correct solution is to load version from faces-config.xml (now not needed any more) OR from @FacesConfig annotation, not from Package.getImplementationVersion method.

I'm not sure but I think JSF should directly provide some solution for that - but even if it would, it would not work for older JSF implementations still supported by Primefaces.

Then you should maintain some separate class for that because it is not a simple problem.

About the filter - beware of lifecycle. I had the same thought but I wasn't sure when the context and configuration are initialized. Maybe that is the reason why someone implemented it by copy and paste.
Btw ... hate the lack of javadocs.

dmatej:
Closed? So Primefaces are finally unusable with Payara 5? Or someone implemented another solution not meant here? Weird project ...

tandraschko:
I just closed the PR ;) Will update the issue in 10 mins...

PR #3007, opened 2017-11-27
Fix #3005: Inplace use emptyLabel if label="" or null.

### LINKED ISSUE #3005: Inplace: emptyLabel useless with configured label
# IMPORTANT 

## 1) Environment
- PrimeFaces version: 6.1.8
- Application server + version: Wildfly 11.0.0 (also with 10.1.0)
- Affected browsers: All

## 2) Expected behavior
emptyLabel will be used, if value of the underlying input field is empty (no matter if label is configured or not).

## 3) Actual behavior
If label is configured, emptyLabel will be ignored (see InplaceRenderer.java):
`if(label != null) { return label; }`

See also the bugreport with attached patch from 2012:
https://code.google.com/archive/p/primefaces/issues/4317

## 4) Steps to reproduce
Just use an p:inplace and configure label as well as emptyLabel.

## 5) Sample XHTML
`<p:inplace
   widgetVar="aliasWidget"
   label="#{myBean.alias}"
   emptyLabel="#{myBean.value}"
   editor="true">
      <p:inputText
         id="alias"
         label="MY_ALIAS"
         value="#{myBean.alias}">
      </p:inputText>
</p:inplace>`

## 6) Sample bean
`import javax.enterprise.context.SessionScoped;
import javax.inject.Named;
import java.io.Serializable;

@Named
@SessionScoped
    private String alias;

    public String getAlias() {
        return alias;
    }
    public void setAlias(String alias) {
        this.alias = alias;
    }
}
`

melloware:
Removed dead variable of the ResponseWriter
Changed "" to Constants.EMPTY_STRING
Implemented check for label="" to show emptyLabel value.

### DIFF
--- a/src/main/java/org/primefaces/component/inplace/InplaceRenderer.java
+++ b/src/main/java/org/primefaces/component/inplace/InplaceRenderer.java
@@ -23,6 +23,7 @@ import javax.faces.context.ResponseWriter;
 
 import org.primefaces.renderkit.CoreRenderer;
 import org.primefaces.util.ComponentUtils;
+import org.primefaces.util.Constants;
 import org.primefaces.util.HTML;
 import org.primefaces.util.WidgetBuilder;
 
@@ -112,7 +113,7 @@ public class InplaceRenderer extends CoreRenderer {
         String label = inplace.getLabel();
         String emptyLabel = inplace.getEmptyLabel();
 
-        if (label != null) {
+        if(label != null && !isValueBlank(label)) {
             return label;
         }
         else {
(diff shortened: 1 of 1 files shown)


src/main/java/org/primefaces/component/inplace/InplaceRenderer.java
```diff
@@ -112,7 +113,7 @@ protected String getLabelToRender(FacesContext context, Inplace inplace) {
         String label = inplace.getLabel();
         String emptyLabel = inplace.getEmptyLabel();
 
-        if (label != null) {
+        if (label != null && !isValueBlank(label)) {
```
Rapster: Write `!isValueBlank(label)` should be enough

melloware: Thanks!  Fixed..

PR #3011, opened 2017-11-27
Fixed NPE and other improvements

Quix0r:
What I have fixed so far:
- fixed NPE in InFilterConstraint when value is NULL
- added missing `@Override` annotation in all filter constraint classes
- Code such as `for (Iterator<? extends Object> it = collection.iterator(); it.hasNext();) {` can be replaced with a simple `for()` iteration

The NPE fixes an issue when a filtered value is NULL, which may happen when you have this JSF code:

````
<f:facet name="filter">
	<p:selectCheckboxMenu
		filter="true"
		filterMatchMode="contains"
		label="#{msg.LABEL_BRANCH_OFFICES}"
		onchange="PF('employeeList').filter()"
		updateLabel="true"
		title="#{msg.FILTER_BY_MULTIPLE_EMPLOYEES_TITLE}"
		>
		<f:converter converterId="BranchOfficeConverter" />
		<f:selectItem itemValue="#{null}" itemLabel="#{msg.NONE_SELECTED}" />
		<f:selectItems value="#{branchOfficeController.allBranchOffices()}" var="branchOffice" itemValue="#{branchOffice}" itemLabel="#{beanHelper.renderBranchOffice(branchOffice)}" />
	</p:selectCheckboxMenu>
</f:facet>
````

This has been stripped out from a real-world application (no example/demo stuff). So pardon me for custom stuff.

I also propose to move on from 1.6 to 1.7 as it (as you may know) has many improvements that shortens your code.

**Prior 1.6**:
````
if (someVar != null && someVar.equals(otherVar))
````

**With 1.7+**:
````
import java.util.Objects;

if (Objects.equals(someVar, otherVar))
````

As you can see, this is much smaller code. Sure there is lot's of more new stuff but I'm to lazy to dig that all up. 

Also I would recommend using "this." everywhere you access a class' field to avoid confusion and help the compiler.

Next is 'final' keyword which you should use in parameters and local variables where possible to avoid assigning parameters with new values accidentally and to help the compiler with optimizing the code.

I could do that both with 1.6 but surely not the `Objects` thing (only with 1.7+).

`@Override` also aids the compiler in finding accidentally changed method signatures.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -215,6 +215,7 @@
                     <source>1.6</source>
                     <target>1.6</target>
                     <encoding>${project.build.sourceEncoding}</encoding>
+                    <showDeprecation>true</showDeprecation>
                 </configuration>
             </plugin>
 
--- a/src/main/java/org/primefaces/model/filter/ContainsFilterConstraint.java
+++ b/src/main/java/org/primefaces/model/filter/ContainsFilterConstraint.java
@@ -20,6 +20,7 @@ import org.primefaces.util.Constants;
 
 public class ContainsFilterConstraint implements FilterConstraint {
 
+    @Override
     public boolean applies(Object value, Object filter, Locale locale) {
         String filterText = (filter == null) ? null : filter.toString().trim().toLowerCase(locale);
 
--- a/src/main/java/org/primefaces/model/filter/EndsWithFilterConstraint.java
+++ b/src/main/java/org/primefaces/model/filter/EndsWithFilterConstraint.java
@@ -20,6 +20,7 @@ import org.primefaces.util.Constants;
 
 public class EndsWithFilterConstraint implements FilterConstraint {
 
+    @Override
     public boolean applies(Object value, Object filter, Locale locale) {
         String filterText = (filter == null) ? null : filter.toString().trim().toLowerCase(locale);
 
--- a/src/main/java/org/primefaces/model/filter/EqualsFilterConstraint.java
+++ b/src/main/java/org/primefaces/model/filter/EqualsFilterConstraint.java
@@ -19,6 +19,7 @@ import java.util.Locale;
 
 public class EqualsFilterConstraint implements FilterConstraint {
 
+    @Override
     public boolean applies(Object value, Object filter, Locale locale) {
         if (filter == null) {
             return true;
--- a/src/main/java/org/primefaces/model/filter/ExactFilterConstraint.java
+++ b/src/main/java/org/primefaces/model/filter/ExactFilterConstraint.java
@@ -20,6 +20,7 @@ import org.primefaces.util.Constants;
 
 public class ExactFilterConstraint implements FilterConstraint {
 
+    @Override
     public boolean applies(Object value, Object filter, Locale locale) {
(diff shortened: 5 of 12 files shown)

### CODE REVIEW

**Comment by Quix0r** [comment]
This change will also alert you on using deprecated APIs as they are highly discouraged to be used (anymore).

**Comment by jxmai** [comment]
@Quix0r Could you create an issue and attach this PR. (This is the PF way to document stuff)

---

Also I think you're trying to mix two issues into one by briefly taking a look

1. fix NPE

2. comply with JRE 1.7 standard

**Comment by Quix0r** [comment]
Oh, sure thing. 👍

**Comment by tandraschko** [comment]
Jfyi: we won't upgrade to 1.7 soon and i also think the benefit is quite small. Upgrading to 1.8 makes really sense but this wont happen the next 1-2 years.

I hope for everyone out there that the compiler can optimize the code also without final keywords. We dont use it in all open source projects im working on.

**Comment by jxmai** [comment]
another thing: This first commit contains build failure. @Quix0r  Could you rebase your branch to combine two commits into one before we discuss to accept the changes or not ?

**Comment by jxmai** [comment]
@Quix0r I think you can reopen a new PR and only add null check :)

**Comment by Quix0r** [comment]
Oh, so no more than one commit per PR?

**Comment by jxmai** [comment]
> no more than one commit per PR?

not really. just to make sure not to include any commit that cannot build. 

LGTM now :+1:

**Comment by Quix0r** [comment]
I have finished a repeatable example app (with fake persistence, of course):
https://cloud.mxchange.org/index.php/s/149vSQYmDeENiuY/download?path=%2FPrimefaces&files=NpeInFilterConstraint.zip

File: pom.xml [hunk 1]
```diff
@@ -215,6 +215,7 @@
                     <source>1.6</source>
                     <target>1.6</target>
                     <encoding>${project.build.sourceEncoding}</encoding>
+                    <showDeprecation>true</showDeprecation>
```
Quix0r, line 218: A little off-topic but still okay, I guess.

melloware: I like it.

melloware (approved):

melloware:
@Quix0r thanks for the PR.  We always appreciate when someone has a bug and shows up with the fix!

jxmai (approved):

Quix0r:
Please don't forget to `cherry-pick` it into 6.1 branch as this version is also affected.

Quix0r:
Is the mentioned `cherry-pick` done?

jxmai:
@Quix0r We don't have the impact on elite unless PRO user request.

Quix0r:
??? 6.1 is no longer elite download: http://search.maven.org/remotecontent?filepath=org/primefaces/primefaces/6.1/primefaces-6.1.jar

jxmai:
@Quix0r Hot fixes can only go to elite versions (Let's say 6.1.x). Once a release is done, it cannot be modified ( 6.1 version should contain the same codebase for any user). That's what versioning is for.

---

Your patch will be in 6.2 even if no `cherry-pick` to elite 6.1.x.

Quix0r:
Okay, I was not aware of how you versioned. Thank you for explaining it.

PR #3024, opened 2017-11-29
Fix #726: Aristo theme allow large FontAwesome icons.

### LINKED ISSUE #726: Font aweseome - larger icons cutted of
Larger icons are cutted off and often the margin to the aligned text in button or menuitem is very small (see icon bed in image below).

Go to showcase and replace icons with larger icons: 

http://www.primefaces.org/showcase/ui/misc/fa.xhtml

If i remove _overflow: hidden_ from class _ui-icon_ it works

See image:

http://fs5.directupload.net/images/150926/jxdxrx8c.png

melloware:
Removed ui-icon style that was preventing large FA icons and seems to have no impact on regular icons as well.  Rule determined to be unnecessary.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces-aristo/theme.css
+++ b/src/main/resources/META-INF/resources/primefaces-aristo/theme.css
@@ -12,12 +12,6 @@ a {
     outline: none;
 }
 
-.ui-icon {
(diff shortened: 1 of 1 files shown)


mertsincan:
Thanks a lot for the PR ;)

PR #3031, opened 2017-11-30
Fix #2583, Fix #2584 - disable sortable on transfer start until save state is fi…

### LINKED ISSUE #2583: picklist: using transfer button and then fast double clicking on the element causes the item to disppear
picklist-fast-clicking-error

## 1) Environment
- PrimeFaces version: 6.1 (current showcase)
- Affected browsers: firefox + chrome, but probably all

## 2) Expected behavior

Clicking fast on the transfer button and then on the same element in the source list does not mess up the lists.

## 3) Actual behavior

Clicking fast in the described way removes the element from both lists.

## 4) Steps to reproduce

- go to https://www.primefaces.org/showcase/ui/data/pickList.xhtml
- scroll down to 'PickList on Steroids' example
- select "Afterdark' element (works the same with all elements but it is easy to reproduce with this one because it's close to the 'select' button
- click the -> button
- very fast, right after clicking the transfer button in the previous step, do the double click on the same element (Afterdark in this case). 

Doesn't always happen but if you're fast enough you will see the described problem.
It maybe easier to reproduce with changed `effectSpeed`.

### LINKED ISSUE #2584: picklist: using transfer button and then dragging the element messes up the lists
picklist-item-drag-error

## 1) Environment
- PrimeFaces version: 6.1 (current showcase)
- Affected browsers: firefox + chrome, but probably all

## 2) Expected behavior

Clicking on the transfer button and then transferring the same element does not mess up the lists.

## 3) Actual behavior

Clicking in the described way creates an empty space in the lists. Also, after this situation happens, the lists cannot be reordered anymore, both by dragging and using the buttons.

## 4) Steps to reproduce

- go to https://www.primefaces.org/showcase/ui/data/pickList.xhtml
- scroll down to 'PickList on Steroids' example
- select an item on the source list
- click the -> button
- very fast, right after clicking the transfer button in the previous step, drag the same element into the target list

Doesn't always happen but if you're fast enough you will see the described problem.
It maybe easier to reproduce with changed `effectSpeed`.

jxmai:
…nshed

#2583, #2584

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/picklist/picklist.js
+++ b/src/main/resources/META-INF/resources/primefaces/picklist/picklist.js
@@ -651,9 +651,10 @@ PrimeFaces.widget.PickList = PrimeFaces.widget.BaseWidget.extend({
     },
 
     transfer: function(items, from, to, type) {
-        var $this = this,
-        itemsCount = items.length,
-        transferCount = 0;
+        $(this.jqId + ' ul').sortable('disable');
+        var $this = this;
+        var itemsCount = items.length;
+        var transferCount = 0;
 
         if(this.isAnimated()) {
             items.hide(this.cfg.effect, {}, this.cfg.effectSpeed, function() {
@@ -720,6 +721,7 @@ PrimeFaces.widget.PickList = PrimeFaces.widget.BaseWidget.extend({
                 transferBehavior.call(this, ext);
             }
         }
+        $(this.jqId + ' ul').sortable('enable');
     },
 
     getListName: function(element){


src/main/resources/META-INF/resources/primefaces/picklist/picklist.js
```diff
@@ -651,9 +651,10 @@ PrimeFaces.widget.PickList = PrimeFaces.widget.BaseWidget.extend({
     },
 
     transfer: function(items, from, to, type) {
-        var $this = this,
-        itemsCount = items.length,
-        transferCount = 0;
+        $(this.jqId + ' ul').sortable('disable');
+        var $this = this;
```
melloware, line 655: Should you store $(this.jqId + ' ul') in a variable since you use it twice in this method rather than calculate it 2 times for the enable and disable?

melloware (changes requested):

src/main/resources/META-INF/resources/primefaces/picklist/picklist.js
jxmai: Well, two `$(this.jqId + ' ul')` are not in the same function. enable is located in the fireTransferEvent since  it is a callback, but we have to start the disable right at the beginning of the transfer function.

melloware: Ahh yep I didn't see the code fold I thought it was the same method!

melloware (approved):

src/main/resources/META-INF/resources/primefaces/picklist/picklist.js
jxmai: Here is the sortable definition

https://github.com/primefaces/primefaces/blob/cf20b82e386a5d483dd17476a62f04832d3f0768/src/main/resources/META-INF/resources/primefaces/picklist/picklist.js#L52-L54

Issue #2582, opened 2017-07-27
p:focus component doesn't work for selectOneButton


mertsincan:
Reported By PRO User;
> a p:focus with for-Attribute can't focus p:selectOneButton.
p:selectManyButton is working.

PR #3037, opened 2017-12-02
Fix #3029 - strengthen the focus function in core.js

### LINKED ISSUE #3029: Focus: (for, context) with selectOneRadio results in a non-selected item being focused
## 1) Environment
PrimeFaces 6.2-SNAPSHOT

## 2) Expected behavior

p:focus should set the focus on a selected radio item if there is one, not the first.

## 3) Actual behavior

p:focus always sets the focus on the first radio item.

## 4) Steps to reproduce

Use p:focus with `for` or `context` attributes with p:selectOneRadio.

## 5) Sample XHTML

	<p:focus for="console" />
	<p:selectOneRadio id="console" value="PS4">
		<f:selectItem itemLabel="Xbox One" itemValue="Xbox One" />
		<f:selectItem itemLabel="PS4" itemValue="PS4" />
		<f:selectItem itemLabel="Wii U" itemValue="Wii U" />
	</p:selectOneRadio>

This results in:
radio_focus

And if you press space:
radio_focus_space

This was partially fixed in #1182 only when p:focus is used without attributes.

jxmai:
#3029

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/core/core.js
+++ b/src/main/resources/META-INF/resources/primefaces/core/core.js
@@ -399,32 +399,36 @@
                     }
                     else {
                         var firstElement = jq.find(selector).eq(0);
-                        PrimeFaces.focusElement(firstElement);
+                        PrimeFaces.focusFirstElement(firstElement);
                     }
                 }
                 else if(context) {
                     var firstElement = $(PrimeFaces.escapeClientId(context)).find(selector).eq(0);
-                    PrimeFaces.focusElement(firstElement);
+                    PrimeFaces.focusFirstElement(firstElement);
                 }
                 else {
                     var elements = $(selector),
                     firstElement = elements.eq(0);
-                    if(firstElement.is(':radio')) {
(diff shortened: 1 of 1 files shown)


melloware (approved):

tandraschko:
Like it! but:
1) do we still need both funcitons?
2) are the wording of both functions really correct?

jxmai:
> are the wording of both functions really correct?

This will trace back to https://github.com/primefaces/primefaces/commit/288af59d58d7cde118e86b4acac3e5247316d6e3. We also need to take care of #2582 

> do we still need both funcitons?

Probably not, but we need  to refactor them into a logical equivalent function. I will update this PR shortly.

jxmai:
Please review

melloware:
@jxmai this looks good.  I wonder if that was the intent of https://github.com/primefaces/primefaces/issues/3045 by not selecting radio/checkbox in the dialog the real intent was don't focus "hidden" controls but that fix was quick and dirty to just never focus any radio or checkbox. We may be able to reuse this function to fix that dialog issue.

jxmai:
Replied : https://github.com/primefaces/primefaces/issues/3045#issuecomment-352478801

Issue #1283, opened 2016-03-31
Empty datatable does not display context menu


mertsincan:
Reported by PRO User;
Context menu does not display when Data table don't have record.
Our data table has context menu Refresh which loads the data table with new records.
Context menu is displayed if data table has records but if data table don't have record then context menu is not displayed.

mrukhaia:
I have the same issue with treeTable -- when it is empty, right click does not show context menu.

PR #3038, opened 2017-12-02
Fix #2171 - introducing 'disableContextMenuIfEmpty' flag - d…

### LINKED ISSUE #2171: DataTable: add option to enable/disable contextMenu for a DataTable without records
## 1) Environment
- PrimeFaces version: 6.0.16
- Does it work on the newest released PrimeFaces version? -> No

## 2) Expected behavior

When empty dataTable is rightClicked, I expect normal browser context menu will show up.

## 3) Actual behavior

If dataTable is empty and is rightclicked, p:contextMenu with "for" attribute shows up and you can click on every menuItem which is rendered and not disabled.

## 4) Steps to reproduce

Create empty dataTable, create Context menu for that table -> run -> right click on datatable

## 5) Sample XHTML

(dynamically rendered)

## 6) Sample bean

...

## PS.: 
I found this piece of code in components.js:

...DataTable...
... bindContextMenu: function(e, f, b, a) {
......
   if (j.hasClass("ui-datatable-empty-message")) {
       e.show(i)
   }
....

When "e.show(i)" is commented, it seems to be fixed

jxmai:
…efault is true

#2171

Example:

```html

<h:head>
</h:head>
<h:body>
	<p:dataTable id="table" disableContextMenuIfEmpty="true">
		<f:facet name="header">Table</f:facet>
	</p:dataTable>
	<p:contextMenu for="table"></p:contextMenu>
</h:body>
</html>
```


melloware:
maybe a better name would be "alwaysShowContextMenu"

or contextMenu="always" contextMenu="selection" (default)

jxmai:
> maybe a better name would be "alwaysShowContextMenu"

`alwaysShowContextMenu` might confuse user if they don't  attach any context menu at all thinking that attribute alone can pop up a context menu.

> or contextMenu="always" contextMenu="selection" (default)

first, contextMenu="always" would be the default since we cannot break the pro user's request

second, `contextMenu="selection"` might not be clear enough to describe  what the function is IMHO. The main issue here is that user wants to have a way to disable context menu when there is no record in a table.

What do you think 🤔 ?

melloware:
Hmm you are right that is confusing.   I was just trying to think of something less verbose.

tandraschko:
The suggested name is a -1 for me but lets discuss this next year after 6.2 ;)

jxmai:
Alright, I don't have a preference for now either. Let's discuss how to name this flag later :smiley:

jxmai:
A list in mind:

`disableContextMenuOnEmpty`

`disableMenuWhenEmpty`

`disableContextMenuIfEmpty`

`disableContextMenuIfNoRecord`

...

tandraschko:
I would like to merge it but we need to find a good name.

jxmai:
`disableContextMenuIfEmpty` seems to be intuitive and not too verbose. If you agree, I will modify the PR.

cnsgithub:
I would vote for `disableContextMenuIfEmpty`.

Or `contextMenuAlwaysVisible `like `paginatorAlwaysVisible `in `p:dataTable`/`p:dataList`/`p:dataGrid`/`p:treeTable`?

I don't know exactly the issue or previous feature requests, but couldn't we save this attribute at all telling users just to use something like `<p:contextMenu for="cars" rendered="#{not empty dtContextMenuView.cars}">`?

tandraschko:
+1 for disableContextMenuIfEmpty

jxmai:
Thanks for the feedback :+1: PR updated

---

@cnsgithub 

> Or contextMenuAlwaysVisiblelike paginatorAlwaysVisiblein p:dataTable/p:dataList/p:dataGrid/p:treeTable?

No, that is not a good name since context menu cannot be enabled just by the flag. `AlwaysVisible` seems misleading. 


> I don't know exactly the issue or previous feature requests, but couldn't we save this attribute at all telling users just to use something like <p:contextMenu for="cars" rendered="#{not empty dtContextMenuView.cars}">?


First of all, there was a feature request https://github.com/primefaces/primefaces/issues/1283 that has been treated as a bug for another group of users https://github.com/primefaces/primefaces/issues/2171, and the changes were all done in datatable, so it makes sense to make changes to datatable.js in order to give user an option.

Second, `<p:contextMenu for="cars" rendered="#{not empty dtContextMenuView.cars}">?` is clunky to use and the EL does not take **filterValues** into account.

Adding a new flag should be reasonable in this case.

Issue #3039, opened 2017-12-02
Dialog state (maximized/minimized) buggy across AJAX updates


Adam-1:
When a dialog is closed while in minimized/maximized state and then updated (via AJAX) its state become corrupted somehow.

If it was closed in maximized state, when it's shown again it's in normal size but it "thinks" it's still in maximized state so 2 clicks on the maximize button are required in order to maximize it.

If, on the other hand, it was closed while in minimized state, when it's shown again it appears behind the overlay (in case of overlay dialog) which makes the page unusable.

## 1) Environment
Tested on
Tomcat 7
Primfaces 6.1
Chrome

## 2) Expected behavior

The dialog should show in the former state, or at least be consistent (i.e. not show in regular state but thinking it's maximized and not blocking the page)

## 3) Actual behavior

- If the dialog was maximized before closing, two clicks on the maximize button are needed in order to maximize it

- If the dialog was minimized before closing and is modal, it's shown behind the overlay, making the page unusable

## 4) Steps to reproduce

1. Open a dialog
2. Maximize/minimize it
3. Close it
4. Fire any ajax that will cause the dialog to be updated
5. Open the dialog again

## 5) Sample XHTML

https://github.com/Adam-1/primefaces-test/blob/master/src/main/webapp/dialogTest.xhtml

## 6) Sample bean

N/A

Adam-1:
Added PR #3040 
(just my 2 cents, there may be better ways to handle it)

jxmai:
First of all, I could confirm this bug.  

about the PR: @Adam-1 

The changes are all over the place, could you explain the intention in details (your thought) that can help us see through the suggestion.

Adam-1:
@jxmai 
Well, since the problem arise when the dialog DOM is updated but the **client** state isn't, I thought I could restore the maximize/minimize state when the dialog is updated.

Most of the code is just method extraction of the actual maximizing/minimizing without all the checkings and event triggering.

When I detect in the "refresh" method that there was a state (and was probably overridden by the DOM update) I'm using the specific function (max/min) to restore the state of the dialog as before the update.

The minimize bug needed some workaround because it's transferring the dialog to the dock, and DOM updates could even cause duplicate dialogs in the page (one from the update in the original place, the other in the dock)

melloware:
@Adam-1 can you review this fix it definitely looks related to yours but not sure if it does as much as yours does.  https://github.com/primefaces/primefaces/issues/3050

[...]

PR #3040, opened 2017-12-02
Fix suggestion for #3039

Adam-1:
separating the actual code of maximize/minimize from the toggling code and using it to restore the dialog state when being refreshed.


Adam-1:
#3039

jxmai:
@Adam-1 Looks like #3039 (maximize and minimize) has been resolved by Mert :) , so you might close this PR.

tandraschko:
Fixed or outdated - will close this one for now.

Issue #3056, opened 2017-12-08
Upgrade to Font Awesome 5.0.2


ravihariharan:
Incorporate  Font Awesome 5.0.1 (fontawesome-free-5.0.1)

https://fontawesome.com/?utm_source=font_awesome_homepage&utm_medium=display&utm_campaign=fa5_released&utm_content=banner

melloware:
wow 896 icons.  This will most likely be done in 6.3 as PF is trying to button up the 6.2 release now.

ravihariharan:
I got some icons to work. but due to change in 5 some icons will have to changed in code. 

![image](https://user-images.githubusercontent.com/15189434/33787186-1eea679e-dc3a-11e7-857b-be3825b4afee.png)




![image](https://user-images.githubusercontent.com/15189434/33787304-a4011f04-dc3a-11e7-90ef-a6a8d252a506.png)


I can do a PR since this is not a Primefaces issue and will have to done to support "Font Awesome 5" in future.

ravihariharan:
https://github.com/ravihariharan/showcase/blob/master/src/main/webapp/ui/misc/fa.xhtml

ravihariharan:
![image](https://user-images.githubusercontent.com/15189434/33801648-8dd6e51a-dd2f-11e7-8877-8a54847ffd50.png)

Not sure of the impact of removing margin-top:-0.25em from .ui-icon  and hence leaving it as is for now.

kukel:
PF should not include their own version of fontawesome anymore but 'reference' the webjars version.... 
https://stackoverflow.com/questions/18891768/how-to-use-font-awesome-from-webjars-org-with-jsf#28652968

mertsincan:
Thanks a lot for the PR;) 

> Not sure of the impact of removing margin-top:-0.25em from .ui-icon and hence leaving it as is for now.

This issue is fixed.

mertsincan:
Reverted these changes for now.

tandraschko:
Why?

tandraschko:
@mertsincan Also the showcase is completely aligned to 5.x

melloware:
Agree with @tandraschko  the showcase is totally updated and this brings in over 500 new icons.  This is a huge addition to PF 6.2.

jxmai:
Not sure if it is related: Maybe some premium themes will be broken by font-awesome 5 ? At least a lot of built in icons in Atlantis theme disappeared  last time I checked.


**Less file**:

[...]

PR #3059, opened 2017-12-10
Upgrade to Font Awesome 5.0.1 #3056

ravihariharan:
Upgrade to Font Awesome 5.0.1 #3056


melloware (approved):

mertsincan:
Thanks a lot for the PR.

jxmai:
@mertsincan  @melloware  Sorry for late to the party. Well, I don't quite agree upgrading font awesome 5 in this release since it could break the UI everywhere on existed applications last time I checked. Keeping in mind that Primefaces is a business oriented UI framework that stability could be the top priority at least from my point of view.

Please note that Font awesome  5 is a really different from font awesome 4. A lot of icons have been removed, some icon definition were changed. 

Even if you check this page: http://localhost:8080/showcase/ui/misc/fa.xhtml. **At least 10%** of the icons are broken or need to be migrated which is a real pain.

image

tandraschko:
@jxmai Yep, thats not good BUT we would have the same problem for the next release. Sometimes users need to migrate something. 100% backward compatibility is just not possible.

melloware:
I agree with @tandraschko with major release like 6.2 is the time to update FA since if users are updating to Jquery 3.X etc this is already a major release users will have to review their app for changes.  Plus the new FA icons are awesome and they have doubled the amount of icons available.

melloware:
On that note I have to look at that showcase page i wonder if I can write something clever that parses the fa.css and generates that page dynamically so we don't have to update it everytime.

jxmai:
Ok, I see. I just wanted to notify that the font awesome 5 migration effort could be a bit high and to make sure we are on the same pace. Please not to forget updating the migration guide.

jxmai:
> if I can write something clever that parses the fa.css and generates that page dynamically so we don't have to update it everytime.

I like the idea

melloware:
@jxmai Mission accomplished!  https://github.com/primefaces/showcase/pull/57   The page is completely dynamic now.

PR #3082, opened 2017-12-15
Fix #3080 - add an option to keep previous messages alive on a new me…

### LINKED ISSUE #3080: Growl: keep messages on the screen when new message arrives
Currently `growl.js` will remove all the messages from the client when new message arrives (see [source code](https://github.com/primefaces/primefaces/blob/f6a0e54f4d11470bad319c6d21fd27f411015706/src/main/resources/META-INF/resources/primefaces/growl/growl.js#L30)):
```js
  //clear previous messages
  this.removeAll();
``` 

This causes unpleasant side-effects, like loosing the previous messages when the next message arrives right after the previous (this is visible on [showcase](https://www.primefaces.org/showcase/ui/message/growl.xhtml), just quickly post several messages). This gets even worse, when `autoupdate` is used and message posting is followed by immediate AJAX request from another component - messages are completely lost. 

Removing this line from JS improves UX a lot - messages are stacked on the client, and then disappear after timeout with other messages smoothly animating to the top.

jxmai:
…ssage is shown

#3080 

can be verified by http://localhost:8080/showcase/ui/message/growl.xhtml

```xhtml
<h:form>           
            <p:growl id="growl" showDetail="true" sticky="true" keepAlive="true" />  
      
            <p:panel header="Growl">  
                <h:panelGrid columns="2" cellpadding="5">  
                    <p:outputLabel for="msg" value="Message:" />   
                    <p:inputText id="msg" value="#{growlView.message}" required="true" />  
                </h:panelGrid>  
  
                <p:commandButton value="Save" actionListener="#{growlView.saveMessage}" update="growl" />  
            </p:panel>     
        </h:form> 
```

### DIFF
--- a/src/main/java/org/primefaces/component/growl/GrowlRenderer.java
+++ b/src/main/java/org/primefaces/component/growl/GrowlRenderer.java
@@ -59,7 +59,7 @@ public class GrowlRenderer extends UINotificationRenderer {
             .attr("sticky", growl.isSticky())
             .attr("life", growl.getLife())
             .attr("escape", growl.isEscape())
-            .attr("clearPreviousMsgsOnShow", growl.isClearPreviousMsgsOnShow());
+            .attr("keepMessages", growl.isKeepMessages());
 
         writer.write(",msgs:");
         encodeMessages(context, growl);
--- a/src/main/java/org/primefaces/component/growl/GrowlRenderer.java
+++ b/src/main/java/org/primefaces/component/growl/GrowlRenderer.java
@@ -54,12 +54,12 @@ public class GrowlRenderer extends UINotificationRenderer {
 
         writer.endElement("span");
 
-        
         WidgetBuilder wb = getWidgetBuilder(context);
         wb.initWithDomReady("Growl", growl.resolveWidgetVar(), clientId)
             .attr("sticky", growl.isSticky())
             .attr("life", growl.getLife())
-            .attr("escape", growl.isEscape());
+            .attr("escape", growl.isEscape())
+            .attr("clearPreviousMsgsOnShow", growl.isClearPreviousMsgsOnShow());
 
         writer.write(",msgs:");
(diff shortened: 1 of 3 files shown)


Rapster:
Rename into `keepMessages`?

jxmai:
Thanks. I like your suggestion :+1: I pushed another commit and updated the description of this PR.

Rapster (approved):

tandraschko:
I'm not sure but i thought we had a attribute "keepAlive" in the past :D
Whats better? keepAlive or keepMessages?

melloware:
keepAlive is better since I think it conveys what is happening.

melloware:
Actually KeepMessages may be ok here and keepAlive=True may be used for that other ticket where if the mouse is moved over the growl it won't dissappear?

Rapster:
+1 keepAlive

Regarding keeping the messages when mouse has moved over, i'd pick a name mentionning "hover" keyword

jxmai:
renamed the new attribute to `keepAlive` as suggested

tandraschko:
I assume that sticky is a little bit different? Didn't used growl since years :D

jxmai:
@tandraschko In summary, `Sticky` is different from `keepAlive`, and  both are needed.

`sticky`'s documentation:

> Specifies if the message should stay instead of hidden automatically. Default is false.

when sticky is false, a message will disappear automatically after some time

---

`KeepAlive` is to address a different scenario:

https://github.com/primefaces/primefaces/issues/3080#issue-282399134

> Currently growl.js will remove all the messages from the client when new message arrives 

> This causes unpleasant side-effects, like loosing the previous messages when the next message arrives right after the previous (this is visible on showcase, just quickly post several messages). This gets even worse, when autoupdate is used and message posting is followed by immediate AJAX request from another component - messages are completely lost.

PR #3093, opened 2017-12-19
Fix 3092 - Messages: Use isValueBlank to evaluate "for"

Rapster:


melloware (approved):

tandraschko:
It don't think it's the same:
 if (_for != null) {

if (isValueBlank(_for)) {

Rapster:
My bad...

PR #3094, opened 2017-12-19
isValueBlank: better implementation

Rapster:
Even better would to reuse the impl from StringUtils of Apache (but not sure if it is allowed due to license)
```java
    /**
     * <p>Checks if a CharSequence is whitespace, empty ("") or null.</p>
     *
     * <pre>
     * StringUtils.isBlank(null)      = true
     * StringUtils.isBlank("")        = true
     * StringUtils.isBlank(" ")       = true
     * StringUtils.isBlank("bob")     = false
     * StringUtils.isBlank("  bob  ") = false
     * </pre>
     *
     * @param cs  the CharSequence to check, may be null
     * @return {@code true} if the CharSequence is null, empty or whitespace
     * @since 2.0
     * @since 3.0 Changed signature from isBlank(String) to isBlank(CharSequence)
     */
    public static boolean isBlank(final CharSequence cs) {
        int strLen;
        if (cs == null || (strLen = cs.length()) == 0) {
            return true;
        }
        for (int i = 0; i < strLen; i++) {
            if (Character.isWhitespace(cs.charAt(i)) == false) {
                return false;
            }
        }
        return true;
    }
```

### DIFF
--- a/src/main/java/org/primefaces/util/ComponentUtils.java
+++ b/src/main/java/org/primefaces/util/ComponentUtils.java
@@ -189,7 +189,7 @@ public class ComponentUtils {
     
 
     public static boolean isValueBlank(String value) {
-        return value == null || Constants.EMPTY_STRING.equals(value.trim());
+        return value == null || value.trim().isEmpty();
     }
 
     public static boolean isRTL(FacesContext context, RTLAware component) {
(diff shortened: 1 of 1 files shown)


melloware:
+1

melloware (approved):

tandraschko:
I would actually just do the following:
replace
return value.trim().equals("");
with
return value.trim().isEmpty();

it should be available on java6

jxmai:
The main benefit of `"".equal(str)` is that we don't need to do null check, but null check is necessary in this case as we need to call `trim`. 

`isEmpty()` introduced in Java 1.6 is faster and cleaner.

Rapster:
Fair enough @tandraschko I'll change it

jxmai (approved):

PR #3114, opened 2017-12-27
#3112 - wording

jxmai:


Rapster (commented):
Don't understand the comment, get method is available no matter what the implementation is ... Is this truly the reason? Or is it for perf reasons to avoid the creation of one iterator?

jxmai:
> Or is it for perf reasons to avoid the creation of one iterator?

I guess the whole point is to use the plain index loop to boost performance as the ticket title says `[perf] use index loop for nav parameters`

> Don't understand the comment, get method is available no matter what the implementation is

I was trying to fix the wording problem but I did not think one step deeper. I agree that  List interfaces contains `get` method  anyway (https://docs.oracle.com/javase/6/docs/api/java/util/List.html) so the comment is not necessary.

tandraschko:
Index loop is only faster for RandomAcess, so it would be a bad change for a LinkedList ;)
On arraylists the change is quite good as you also safe a iterator instance.

tandraschko:
Removing the comment is not good, someone in the future will change it to a foreach ;)

jxmai:
> Index loop is only faster for RandomAcess, so it would be a bad change for a LinkedList ;)

Ah, I see your point now. will think of another comment later :+1:

tandraschko:
i don't think we need a special comment here, i just want to make sure that it's a good improvement as the JSF impls uses ArrayLists
I've done this improvement on many places ;)

jxmai:
Rebased to the first version.

PR #3133, opened 2018-01-05
Fixes #2749 - p:button does not use nested f:param when using href attribute

### LINKED ISSUE #2749: p:button does not use nested f:param when using href attribute
PrimeFaces version: 6.1.5
With `<p:button>` nested `<f:param>` elements are not passed as parameters to the url defined in `href ` attribute.

```
<p:button href="http://www.google.com">
    <f:param name="q" value="primefaces"/>
</p:button>
```
When clicking on the button, I would expect to go to http://www.google.com?q=primefaces
Instead, I go to http://www.google.com

Parameters are correctly taken into account when using the `outcome ` attribute instead of `href`.
I know I can directly put my parameters in the url in `href  `attribute (`href="http://www.google.com?q=primefaces"`), but passing them as nested `<f:param>` elements has several advantages:
- consistency (same behaviour as with `outcome` attribute)
- clarity (otherwise the `href ` can be very long and unreadable)
- possibility to use the disable attribute of `<f:param>`

jasonex7:
This commit Fixes #2749 allowing  `<p:button>` and `<p:link>` take `<f:param>` in consideration when using `href` attribute.


tandraschko:
getResourceURL shouldn't be used for links, only for resources like js/css etc.

tandraschko:
As we already used it, i will fix this: https://github.com/primefaces/primefaces/issues/3134
Your PR should be aligned with my fix then.

tandraschko:
I'm also not sure if we can support it easily. AFAICS all encodeXXXUrl methods in ExternalContext appends e.g. the windowId. I don't think it's good to apply windowId on external href's.

tandraschko:
Hhmmm, i have to talk about that with the JSF spec guys first.

jasonex7:
> getResourceURL shouldn't be used for links, only for resources like js/css etc.
> As we already used it, i will fix this: #3134
Your PR should be aligned with my fix then.

Yeah @tandraschko, that's right, the code already used `getResourceURL` for links other than js/css, etc, so I followed that. Do you think I should use another new method in the case of "href" attribute? because in this case I overloaded `getResourceURL`, but I could create another method with other name (e.g. `getHrefURL`) instead of overloading. 

> I'm also not sure if we can support it easily. AFAICS all encodeXXXUrl methods in ExternalContext appends e.g. the windowId. I don't think it's good to apply windowId on external href's.

The javadoc of the method I'm using (`encodeBookmarkableURL`) says the following (note bold part please):

> javax.​faces.​context.​ExternalContext
> public String encodeBookmarkableURL(String baseUrl, Map<String,List<String>> parameters)
> 
> The purpose of this method is to generate a query string from the collection of Parameter objects provided by the parameters argument and append that query string to the baseUrl. This method must be able to encode the parameters to a baseUrl that may or may not have existing query parameters. The parameter values should be encoded appropriately for the environment so that the resulting URL **can be used as the target of a link (e.g., in an href attribute)** in a JSF response. It's possible for an ExternalContext implementation to override this method in any way that would make the URL bookmarkable in that environment.

So I don't think it would be a problem use that method for this purpose, but I'd like to know what JSF guys have to say about it.

tandraschko:
The problem is that also encodeBookmarkableURL appends e.g. the ClientWindowID - i don't think that we should append the clientWindowId to a external href.

tandraschko:
fixed that now by removing encoding on href. It's not required as it's a external link. NOTE: we must not use e.g. #encodexxxUrl for a external link, so you have to align your PR.

jasonex7:
> The problem is that also encodeBookmarkableURL appends e.g. the ClientWindowID - i don't think that we should append the clientWindowId to a external href.

I've investigate more about ClientWindowId, you have right, if the user has activated that, the method `encodeBookmarkableURL` will append it to the url.

> fixed that now by removing encoding on href. It's not required as it's a external link. NOTE: we must not use e.g. #encodexxxUrl for a external link, so you have to align your PR.

Thanks for solving #3134. I've have a question for align my PR, it's about webapp context in the href attribute. In my personal case I want to use href of `<p:button>`for reference a servlet of the same webapp and pass some `<f:param>`, so for example if the URL of the servlet is: /myservlet/path, webapp context is `/app1` and use the <p:button> like this:

```
<p:button href="/myservlet/path">
    <f:param name="param1" value="#{someBean.param1}" />
    <f:param name="param2" value="#{someBean.param2}" />
</p:button>
```
Is ok that I expect the href will render `/app1/myservlet/path?param1=value1&param2=value2`? Or should it be without the app context: `/myservlet/path?param1=value1&param2=value2`?
The case of absolute URLs (e.g. http://www.google.com) is different and I'm understand that. I'd appreciate your comments @tandraschko. Thanks.

tandraschko:
I would say that href is for "complete" links and outcome is used for app-internal links, where f:param is appended. I'm also not sure if we should modify the href here.

jasonex7:
The problem with `outcome` is that it doesn't support servlet paths, AFAIK it only support JSF navigation cases. Do you think in that case I should append manually context path in href? Fox example:

```
<p:button href="#{facesContext.externalContext.contextName}/myservlet/path">
    <f:param name="param1" value="#{someBean.param1}" />
    <f:param name="param2" value="#{someBean.param2}" />
</p:button>
```

tandraschko:
I see. The problem is that we must build our own logic here to append params as it seems that we can't reuse a JSF-API.

Yep, contextPath must be appended manually. href is really basic, not like outcome.

jasonex7:
> Yep, contextPath must be appended manually. href is really basic, not like outcome.

OK, then my PR will use a method to assemble params into href attribute without modifying href and FacesContext unaware.

tandraschko:
1) the builds fails
2) please don't change the formattings

jasonex7:
> 2. please don't change the formattings

Sorry, I was trying to merge from master (I'm still newbie on github 😅 )

tandraschko:
No problem. You can just check your changes (https://github.com/primefaces/primefaces/pull/3133/files) and you see that many lines are changed in this PR just for formatting.

melloware:
@jasonex7 I have done the same thing in the past.  Sometimes I make such a mess I just cancel a PR and submit a new one with the changes after I updated from Master!  I know there are better ways to do it but sometimes I feel like its easier for me after i have made a mess of a branch. :)

jasonex7:
@tandraschko I've corrected the formatting issues that I could see. Do you still see some I missed?

@melloware Thanks for the suggestion. I will consider that as last resort.

tandraschko:
I don't think that your new method should use #getParams - it will append the flowId. Maybe you can add a param or just splitup the method in two methods
1) one method to get all params from the child f:params
2) modify the current getParams to reuse 1)

1) should be used by your new method then

jasonex7:
Now that you mention it, I was unnecessarily overloading `getParams` just for obtain `outcomeTarget.getParams()`.

src/main/java/org/primefaces/util/ComponentUtils.java
```diff
@@ -294,6 +294,83 @@ else if (value.contains(ResourceHandler.RESOURCE_IDENTIFIER)) {
             return context.getExternalContext().encodeResourceURL(url);
         }
     }
+    
+    /**
+     * Generate an <code>href</code> URL considering a base URL and some other params.
+     * This method consider existing query string and fragment in the base URL.
+     * @param baseUrl
+     * @param params
+     * @return
```
melloware: Since you have the javadoc here can you fill out the comment values for the params and return value?

melloware (changes requested):

src/main/java/org/primefaces/util/ComponentUtils.java
jasonex7: Sure. I've completed the javadoc for the method.

melloware (approved):

tandraschko:
It would be great if you could add (2-3) unittests

jasonex7:
I've now added a unit test in ComponentUtilsTest, please check.

rmpestano:
Hi guys,

> Yep, contextPath must be appended manually. href is really basic, not like outcome.

This has changed since 6.1? I'm asking because we had menus declared without context path like below:

```
   <ul class="sidebar-menu">
        <li>
            <p:link href="/index.xhtml">
                <i class="fa fa-home"></i>
                <span>Home</span>
            </p:link>
        </li>
        <li>
            <a href="/docs/index.html" target="_blank">
                <i class="fa fa-book"></i>
                <span>Documentation</span>
            </a>
        </li>
        <li>
            <p:link href="/pages/analytics/analytics.xhtml">
                <i class="fa fa-pie-chart"></i>
                <span>Analytics</span>
            </p:link>
        </li>
        <li class="header">
            General
        </li>
        <li>
            <p:link href="/pages/exception/exception.xhtml">
                <i class="fa fa-stack-overflow"></i>
                <span>Exceptions</span>
            </p:link>
        </li>
        <li>
...
```

And it just works in **6.1** (and also in 6.0), the contextPath is appended automatically but now in 6.2 it is not anymore.

Can you confirm that it has changed?

jasonex7:
Because of the fix #3134 PrimeFaces no longer append context path in href attribute because it doesn't use encodeResourceURL method of ExternalContext anymore

tandraschko:
outcome should be use to navigate between JSF views. href is something more basic, also to use absolute/external links, without appending any clientwindow or flowId.
Also the contextPath can be simply prepended with #{request.contextPath}

rmpestano:
This will break a lot of applications and slow down migration path, can't we deprecate and show a warning in this release and then remove and change the behaviour in a future release?

tandraschko:
There is no correct fix for this, sorry (based on the old solution).
Let me explain:
In 6.1, getResourceUrl was used. This appends the contextPath and maybe the sessionId.
The problem is that getResourceUrl can't be used anymore as this will trigger a resource push on JSF 2.3.
The next problem is that this implementation was already not correct, if used for JSF views, as getResourceUrl doesn't append the WindowId and FlowId, which is actually "required" for a href to JSF view.

Let me try something.

tandraschko:
See: https://github.com/primefaces/primefaces/issues/3154

tandraschko:
Fixed it in trunk. Please test it!
@jasonex7 You may also modify your PR.

jasonex7:
Thanks @tandraschko.  I think #3154 it's welcomed because though `href` is basic attribute, `<p:button>` and `<p:link>` are not basic, so one may expect those tags take contextPath into consideration when using internal links.

I've just committed changes to this PR for aligning with #3154.

tandraschko:
@jasonex7 
i need to refactor it, because we also need to remove the prependContextPath thing.
Any idea why we use own code here instead of ExternalContext#encodeRedirectURL ?

jasonex7:
> Any idea why we use own code here instead of ExternalContext#encodeRedirectURL ?

@tandraschko As we discussed earlier, it was mentioned that methods encodexxxxURL appended windowId, flowId, etc. Also, I'm not sure if encodeRedirectURL handle query string param keys encoding as it was requested by #3409

tandraschko:
oh sorry, so it was my fail. Seems like encodeRedirectUrl really does nothing JSF special, as it's used to encode links for external applications, too.

PR #3145, opened 2018-01-10
Fix #3143 - InputText: focus on RTL still on left side

### LINKED ISSUE #3143: InputText: focus on RTL still on left side
## 1) Environment
- PrimeFaces version: 6.2-SNAPSHOT
- Affected browsers: Chrome

## 2) Expected behavior
I expect the focus to be on the right when `p:inputText` get the focus

## 3) Actual behavior
Focus still on the left

## 4) Steps to reproduce
Setting `primefaces.DIR` to rtl

## 5) Proposal
InputText should implement RTLAware and encode `dir` HTML attribute (see https://www.w3schools.com/cssref/pr_text_direction.asp)

Rapster:
I decided to put the method `encodeDirection` in `ComponentUtils` as it is a utility method. I could have write that in `CoreRenderer` too, but since most of the `CoreRenderer` component are not `RTLAware` I thought it gives more sense to put it in `ComponentUtils`. Let me know what do you think

### DIFF
--- a/src/main/java/org/primefaces/component/inputtext/InputTextRenderer.java
+++ b/src/main/java/org/primefaces/component/inputtext/InputTextRenderer.java
@@ -75,7 +75,7 @@ public class InputTextRenderer extends InputRenderer {
             writer.writeAttribute("value", valueToRender, null);
         }
 
-        ComponentUtils.encodeDirection(context, inputText);
+        renderRTLDirection(context, inputText);
         renderPassThruAttributes(context, inputText, HTML.INPUT_TEXT_ATTRS_WITHOUT_EVENTS);
         renderDomEvents(context, inputText, HTML.INPUT_TEXT_EVENTS);
 
--- a/src/main/java/org/primefaces/component/inputtext/InputTextRenderer.java
+++ b/src/main/java/org/primefaces/component/inputtext/InputTextRenderer.java
@@ -16,9 +16,11 @@
 package org.primefaces.component.inputtext;
 
 import java.io.IOException;
+
 import javax.faces.component.UIComponent;
 import javax.faces.context.FacesContext;
 import javax.faces.context.ResponseWriter;
+
 import org.primefaces.context.RequestContext;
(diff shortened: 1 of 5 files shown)


src/main/java/org/primefaces/component/inputtext/InputTextRenderer.java
```diff
@@ -73,6 +75,7 @@ protected void encodeMarkup(FacesContext context, InputText inputText) throws IO
             writer.writeAttribute("value", valueToRender, null);
         }
 
+        ComponentUtils.encodeDirection(context, inputText);
```
melloware: @Rapster I see what you did here and your comment about it but I would still vote for renderRtlDirection() method to be consistent with others.  I know that putting that in the base class feels weird but not if we have more and more components using it?  I will wait for @tandraschko thoughts.

tandraschko:
I would put that method in InputRenderer or something. ComponentUtils is a too big pool.

src/main/java/org/primefaces/component/inputtext/InputTextRenderer.java
Rapster: @melloware yes I will reuse same naming convention

Rapster:
Yeah that's what I thought first, but `dir` attribute can be eligible to more than just inputs. But the truth is, if we start using this new attribute in existing RTL components I'm afraid it will mess the rendering. So I guess, I'll move the method in InputRenderer taking account @melloware comment

tandraschko:
we can still move it to CoreRenderer instead of InputRenderer

Rapster:
Let's keep it in InputRenderer for now, CoreRenderer is already a big tool too

melloware (approved):

tandraschko:
Could you fix the conflicts?

tandraschko:
@Rapster ?

Rapster:
Yep sorry, I'll try to do that asap :wink:

tandraschko:
Applied it

Issue #3111, opened 2017-12-26
Droppable: Add greedy attribute


wattsec:
Hi,

First of all, thanks for your great job with primefaces. 

I've been using this library for some time and I've always missed the capability to set the greedy option for droppable element so I've decided to make that change. I thing this may be useful for scenarios like wysiwyg editors and other kind of wizards. I hope this change could help anyone else.

Even though it is a small change and quite easy to make it backward compatible, I'm finishing my tests before the PR.

Regards,
J.

wattsec:
Hi,
I've made a PR with this improvement. I hope it may be useful.

Regards,
J.

PR #3147, opened 2018-01-10
Greedy

wattsec:
Adds greedy attribute to droppable component.

Default value is false, keeping backward compatibility with current implementation.


Rapster:
Please attach ticket number in your title (e.g `Fix #<ticket_number> - <ticket_title>`) that's how I use to submit PR

jxmai:
found the attached ticket: https://github.com/primefaces/primefaces/issues/3111. And yes I agree with @Rapster 's suggestion since there seems no link between this ticket and the pull based on the first impression.

Rapster:
The description should be clearer. From jQuery website (at least pick few words)

> By default, when an element is dropped on nested droppables, each droppable will receive the element. However, by setting this option to true, any parent droppables will not receive the element. The drop event will still bubble normally, but the event.target can be checked to see which droppable received the draggable element.

tandraschko:
Yep, @julioa please update the attribute description like rapster mentioned.

wattsec:
Hi, I've updated the attribute description. I hope is clear and short enough. J.

caiomerino92:
I need attribute greedy implementation, but I couldn't find it in gitHub.
I'm using primefaces 6.2, but there isn't greedy implementation yet.
How can I get access to the latest code which added greedy attribute to droppable component. Can anyone help me please?

Rapster:
Greedy will be available in 6.3, see #3111 

To get the lastest code, simply clone the repository and `mvn clean install`

caiomerino92:
Thanks a lot

Issue #3156, opened 2018-01-12
Autocomplete: Disable autoselect for items which are equal to current input


OBreidenbach:
With #2733 it was introduced that an item is selected if it is equal to the current input. We would like to disable this behavior for our use case.
We can provide a PR. We would add attribute to the autoComplete-Tag. Just let us know, how to name it.
Suggestion: disableAutoSelection

melloware:
Definitely suggest a PR.  And the name maybe it should just be autoSelection="true" with true being the default and you can turn it off with autoSelection="false".

OBreidenbach:
Added PR #3162
I would love to see it in the 6.2 release. For me, it would solve the problem with #3155

PR #3162, opened 2018-01-15
#3156; disable config of auto selection for autocomplete

OBreidenbach:


src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
```diff
@@ -809,9 +810,9 @@ PrimeFaces.widget.AutoComplete = PrimeFaces.widget.BaseWidget.extend({
 
             for(var i = 0; i < $this.currentItems.length; i++) {
                 var stripedItem = $this.currentItems[i];
```
melloware, line 812: I know you didn't make this variable but shouldn't "stripedItem" be "strippedItem" ?

melloware (approved):

src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
OBreidenbach: I think you are right, it should be "strippedItem".

Rapster: I did! My bad 😅

mertsincan:
Thanks for the PR ;)

PR #3183, opened 2018-01-19
Fix #3182 - resolveClientId can't be used as JSF function

### LINKED ISSUE #3182: resolveClientId can't be used as JSF function
## 1) Environment
- PrimeFaces version: 6.1.RC1
- Application server + version: Tomcat
- Affected browsers: 

## 2) Expected behavior
Should be able to use `#{p:resolveClientId('@(.ui-datatable)', view)}`

## 3) Actual behavior
As a result, an exception is thrown

## 4) Steps to reproduce
See 2/

## 5) Sample XHTML
See 2/

PR on his way

Rapster:


tandraschko:
Good catch, thanks!

PR #3188, opened 2018-01-19
Fix #3187: PrimeFaces OSGi headers require JSF 2.3, CDI unnecessarily

### LINKED ISSUE #3187: PrimeFaces OSGi headers require JSF 2.3, CDI unnecessarily
Currently the PrimeFaces OSGi `Import-Package` header requires all `javax.*` packages. All other packages are optionally imported. However, since PrimeFaces only optionally depends on a single JSF 2.3 package for component tree searching and on the CDI packages to support the deprecated PrimePush feature, those packages should be marked as optional (and the JSF packages should allow JSF 2.0.0+). This will allow PrimeFaces to be used in an OSGi environment with versions of JSF 2.0.0+ and without CDI present.

stiemannkj1:


melloware (approved):

tandraschko:
Thanks kyle!

Issue #2363, opened 2017-05-15
p:dataExporter PDF export does not show Euro (€) sign


rho66:
# IMPORTANT 
- If you open an issue, fill every item. Otherwise the issue might be closed as invalid.
- Before you open an issue, test it with the current/newest version.
- Try to find an explanation to your problem by yourself, by simply debugging. This will help us to solve your issue 10x faster
- Clone this repository https://github.com/primefaces/primefaces-test.git in order to reproduce your problem, you'll have better chance to receive an answer and a solution.
- Feel free to provide a PR (Primefaces is an open-source project, any fixes or improvements are welcome.)

## 1) Environment
- PrimeFaces version: 6.1
- Does it work on the newest released PrimeFaces version? Version? No
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) No
- Application server + version: WildFly-10.0.0.Final
- Affected browsers: Firefox, Chrome, IE

## 2) Expected behavior

PDF dataExporter should export Euro sign correctly with UTF-8 encoding

## 3) Actual behavior

PDF dataExporter does not show Euro sign in exported PDF file

## 4) Steps to reproduce

Create a p:dataTable with column headers containing an Euro sign and a PDF dataexporter with encoding type UTF-8

## 5) Sample XHTML

<p:dataTable id="data" paginatorTemplate="{Exporters}" value="#{dataExporterView.cars}">
  <f:facet name="{Exporters}">
    <p:commandButton title="PDF Export" ajax="false" icon="fa fa-file-pdf-o">
      <p:dataExporter type="pdf" target="data" fileName="export.pdf" encoding="UTF-8"/>
    </p:commandButton>
</p:dataTable>

## 6) Sample bean

see https://www.primefaces.org/showcase/ui/data/dataexporter/basic.xhtml

melloware:
@rho66 Have you debugged this at all to see where its not being encoded to UTF-8 in the code?

rho66:
[...]

PR #3203, opened 2018-01-23
Use OpenPdf 1.0.5 as dependency instead of iText 2.1.7.

ghost:


melloware:
Linking to:  https://github.com/primefaces/primefaces/issues/2363

tandraschko:
Not sure about this one.
We could change it but the user must still import it as we only set it to provided.

So we could change it but not necessary.
It would be better to mention this in the docu.

ghost:
Why not both? Update docs, and pom file.

tandraschko:
because the API is the same and everybody could fork it and provide "yet another bugfix fork of the old lowagie"

I think the only good solution for the future is to switch to other frameworks like pdfbox (does it support the same features?)

danteCarvalho:
For some reason, empty encoding="" works for me,i am using itext 2.1.7

tandraschko:
Will close it for now but you can create a issue, so we can check it in future again if its worth to do a hard switch to openpdf.
As i said, currently the user can use both as the API is the same.

Issue #3170, opened 2018-01-18
TextEditor: formattings are getting lost in PF6.2 RC1


RueKow:
## 1) Environment
- PrimeFaces version: 6.2
- Does it work on the newest released PrimeFaces version? Version? No
- Application server + version: Tomcat 8.5
- Affected browsers: all

## 2) Expected behavior

formattings should be persistent after submitting

## 3) Actual behavior

formattings are getting lost

## 4) Steps to reproduce

see forum [https://forum.primefaces.org/viewtopic.php?f=3&t=53836](https://forum.primefaces.org/viewtopic.php?f=3&t=53836)

## 5) Sample XHTML

```
<html xmlns="http://www.w3.org/1999/xhtml"
	xmlns:h="http://java.sun.com/jsf/html"
	xmlns:f="http://java.sun.com/jsf/core"
	xmlns:ui="http://java.sun.com/jsf/facelets"
	xmlns:p="http://primefaces.org/ui">

    <h:head>
    </h:head>

    <h:body>
		<h:form>
			<p:textEditor widgetVar="editor" value="#{testController.text}" height="300">
			</p:textEditor>
			<p:commandButton value="Submit"
				process="@form"
				update="@form" />
		</h:form>
    </h:body>

</html>
```
## 6) Sample bean

```
@ManagedBean
@SessionScoped
public class testController implements Serializable
{
	private String text;

	public String getText()
	{
		return text;
	}

	public void setText(String text)
	{
		this.text = text;
	}

}
```

tandraschko:
@melloware @jxmai could you check that?

melloware:
Found this on their GitHub...

https://github.com/quilljs/quill/issues/903

melloware:
Actually looking more in the PF Widget it does this code to determine HTML or not.  My guess is the format has changed and this is no longer properly determining whether its HTML or not.

```javascript
getEditorValue: function() {
        var html = $(this.editorContainer[0]).find('p').text();
        var value = (html == '<p><br></p>') ? '' : html;

        return value;
    },
```

jxmai:
I could replicate the bug and I will raise a patch shortly.

```js
var html = $(this.editorContainer[0]).find('p').text();
```

should be changed to 

```js
var html = $(this.editorContainer[0]).find('p')[0].innerHTML;
```

the result:

![image](https://user-images.githubusercontent.com/10467831/35106592-3cea0b36-fc34-11e7-929b-7fbae1231cbe.png)

[...]

Issue #2907, opened 2017-11-07
TextEditor: unnecessary p tag is generated to contaminate the original value


jxmai:
## 1) Environment
- PrimeFaces version: 6.2 trunk
- Does it work on the newest released PrimeFaces version? Version? No
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source)
- Application server + version: 
- Affected browsers: all

## 2) Expected behavior

The original value should not be contaminated by the texteditor.

## 3) Actual behavior

Extra p tag is always generated by the texteditor on form submit 

## 4) Steps to reproduce

Use the sample and click submit

![image](https://user-images.githubusercontent.com/10467831/32522397-c174b65c-c3dc-11e7-9e09-cab43dc6ca52.png)

In console
```
<p>test</p>
<p>test2</p>
test3
```

As you can see, the original editor does not put an extra p tag while the first two outputs generated by texteditor contain extra p tags

## 5) Sample XHTML

```xhtml
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml"
	xmlns:ui="http://java.sun.com/jsf/facelets"
	xmlns:f="http://java.sun.com/jsf/core"
	xmlns:p="http://primefaces.org/ui"
	xmlns:h="http://java.sun.com/jsf/html">

<h:head>
	<title>PrimeFaces Test</title>
</h:head>
<h:body>

	<h:form>

		<p:textEditor value="#{testView.test}"></p:textEditor>
		<p:textEditor value="#{testView.test2}"></p:textEditor>
		<p:editor value="#{testView.test3}"></p:editor>
		

		<p:commandButton value="Test"
			actionListener="#{testView.testCommandButton()}" process="@form"></p:commandButton>
	</h:form>

</h:body>
</html>
```

## 6) Sample bean

```java
package org.primefaces.test;

import java.io.Serializable;

import javax.faces.bean.ManagedBean;
import javax.faces.bean.ViewScoped;

@ManagedBean(name = "testView")
@ViewScoped
public class TestView implements Serializable {

    /**
     * 
     */
    private static final long serialVersionUID = 812243260866629419L;
    private String test = "test";
    private String test2 = "test2";
    private String test3 = "test3";

    public void testCommandButton() {
        System.out.println(test);
        System.out.println(test2);
        System.out.println(test3);
    }

    public String getTest() {
        return test;
    }

    public void setTest(String test) {
[...]

PR #3216, opened 2018-01-25
Strengthen #3170 & #2907 - address the regression affects headline wr…

jxmai:
…appers

**Address the regression mentioned by**: https://github.com/primefaces/primefaces/commit/98a598f50bf3eca887960c6a8928c17b0608db94#commitcomment-27101765

**investigation**: https://github.com/primefaces/primefaces/commit/98a598f50bf3eca887960c6a8928c17b0608db94#commitcomment-27102378

**Solution** https://github.com/primefaces/primefaces/commit/98a598f50bf3eca887960c6a8928c17b0608db94#commitcomment-27105660

**test code**

```xhtml
<!DOCTYPE html>
      xmlns:pe="http://primefaces.org/ui/extensions"

    <h:head>   
    </h:head>
    <h:body>
        
		 <h:form>
		 	<p:textEditor value="#{testView.text}"></p:textEditor>
		 	
		 	
		 	<p:commandButton value="Test" update="@form" actionListener="#{testView.printText()}"></p:commandButton>
		 </h:form>
    </h:body>
</html>
```

---

```java
package org.primefaces.test;

import java.io.Serializable;

@ViewScoped

    /**
     * 
     */
    private static final long serialVersionUID = -450911487369298388L;

    private String testString;

    private String text;

    @PostConstruct
    public void init() {
        testString = "Welcome to PrimeFaces!!!";
    }

    public void printText() {
        System.out.println(text);
    }

    public String getText() {
        return text;
    }

    public void setText(String text) {
        this.text = text;
    }

        return testString;
    }

        this.testString = testString;
    }
}

```

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/texteditor/texteditor.js
+++ b/src/main/resources/META-INF/resources/primefaces/texteditor/texteditor.js
@@ -138,8 +138,16 @@ PrimeFaces.widget.TextEditor = PrimeFaces.widget.DeferredWidget.extend({
     },
 
     getEditorValue: function() {
-        var html = $(this.editorContainer[0]).find('p')[0].innerHTML;
-        var value = (html == '<p><br></p>') ? '' : html;
+
+        // github issue: #2907
+        // textEditor generates <p></p> as a content wrapper automatically that will contaminate the original value on form submit & update.
+        // However, the content wrapper will be head element instead of p when Headline is in use.
+        // We only remove the content wrapper when it is a p element
+        var htmlWithPWrapper = $(this.editorContainer[0]).find('p')[0];
+        var html = htmlWithPWrapper ? htmlWithPWrapper.innerHTML : this.editorContainer.get(0).children[0].innerHTML;
+
+        // textEditor generates <p><br></p> when there is no content. After the outer p element is removed, we can safely treat <br> as an empty value
+        var value = (html == '<br>') ? '' : html;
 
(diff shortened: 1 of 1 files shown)


jxmai:
Quill's output is more complicated than I thought. There is another case if user mix p and h together. I will analyse it further and we might need to dump #2907 if necessary.

jxmai:
Yep, the outer p tag is necessary in serveral cases:

quill generates a wrapper for every single line by design that is difficult for us to mess with it. 

1. multiple lines input

```xhtml
"<p>e</p><p><br></p><p><br></p>"
```

2. mix p and h which is pretty much the same as point 1

```xhtml
<p>e</p><p><br></p><h1>f</h1>
```

Basically, every single line has a wrapper in quill. `p` is just one of them

so either we need to remove outer p tag for every line, or dump #2907.

I would say #2907 is an annoying bug that does not appear on any other editors because of the quill's design...

Let's see if removing p tag on every line is doable or not.

tandraschko:
I would actually dump 2907.

jxmai:
No, there is a bigger problem which can determine  #2907 as not fixable.

image

```
this.editorContainer.get(0).children[0].innerHTML
"<p>1</p><p>5</p>"
```

1 and 5 will mix together.

jxmai:
Yes, I will revert all the patches shortly. #2907  is not fixable by quill's design

PR #3221, opened 2018-01-26
[OWASP] fixes #3214 - XSS in p:textEditor

### LINKED ISSUE #3214: XSS in p:textEditor
## 1) Environment
- Application server + version: Wildfly 11
- Affected browsers: all

## 2) Expected behavior
There is a limited set of HTML tags quilljs is able to generate via toolbar. However, it is possible to manipulate the DOM on the client side and submit a tampered version to the server which in turn should sanitize the HTML before persisting or sendig it back to the client.

## 3) Actual behavior
It is possible to inject arbitrary HTML/JS code in DOM. The code could then be persisted to the database and reflected to clients trying to edit using p:textEditor or getting the text rendered through any component that uses `escape="false"`.

## 4) Steps to reproduce
You can test the reflective XSS on your showcase site.
1. Go to https://www.primefaces.org/showcase/ui/input/textEditor.xhtml
2. Open developer tools (F12) and enter in the console: `$('.ql-container').next('input:hidden').val("<script>alert('oops');</script>");`
3. Press the submit button. An alert will be shown after returning from server side.

## 5) Sample XHTML
n/a

## 6) Sample bean
n/a

## 7) Mitigation
This time it is a bit more difficult to fix the issue, because we cannot simply escape all HTML/JS. However, what we can do is to sanitize user HTML input - depending on the quilljs features turned on and thus are available via the toolbar. I.e., if images are not allowed, we should sanitize in a way, that all image tags are dropped and so on. And once again, OWASP is your best friend: https://github.com/OWASP/java-html-sanitizer Btw, I am not a member of OWASP... ;-) I have used it successfully as a workaround for this PrimeFaces issue.

cnsgithub:
This could be used as a basis for further development and needs some refinement.

Please note that a new dependency, owasp-java-html-sanitizer, has to be added. I am not sure if you are willing to accept this...

Please consider this pull request just as a proposal that shows a possible way to go.


tandraschko:
Please don't change the formattings.

Rapster (approved):

tandraschko:
If you can fix the conflicts and also shade it, i can merge it early next week.
After that, i'm away for over 3 weeks.

cnsgithub:
Apparently the component XML files disappeared. Are there any rules to follow when migrating attributes from `textEditor.xml` to `TextEditor.java`? Where did the meta information go?

E.g.
```xml
            <name>allowLinks</name>
            <required>false</required>
            <defaultValue>true</defaultValue>
            <description>Whether to allow links to be included.</description>

```
I will do the shading after #3525 has been merged, since this PR introduces maven-shade-plugin.

tandraschko:
have a look at: TextEditorBase
The descriptions goes directly into the primefaces-p.taglib.xml

cnsgithub:
@tandraschko Merged it. As I understand it correctly, `maven-jsf-plugin `is no longer needed? If so, please have a look at `pom.xml`, line 686: the javadocs still include `maven-jsf-plugin` folder.

tandraschko:
I will probably apply it next week and shade the depedency

tandraschko:
stupid questions but there are many possible tags now, which will be filtered now but worked in a previos version?

tandraschko:
Damn, we can't use it. It depends on Guava, which is FAR to big to be shaded.
If we can use another dependency, feel free to provide a new PR.

PR #3233, opened 2018-01-29
Fix #742 - Pixelperfect icons by using jquery-ui-font instead of bitmap icons in primefaces-aristo default theme

### LINKED ISSUE #742: SelectOneRadio and SelectBooleanCheckbox are not pixelperfect
Hi, ive noticed that these two components look odd compared to default html one. Is it possible to make it more crisp and 100% centered like the html components? 

Especially if the background gradient is removed, the dot doesn't look like it is centered

Image of Compnents

cnsgithub:
uses jquery-ui font (https://mkkeck.github.io/jquery-ui-iconfont/) for the default primefaces-aristo theme.. Pixelperfect icons even when scaling, and ui-icons bitmap files can be removed

image

Retains compatibility with all ui-icon usage of jquery ui
Issue: https://github.com/primefaces/primefaces/issues/742


tandraschko:
@onursenture @mertsincan Could you check that? But i would apply it after 6.2.

melloware (approved):

cnsgithub:
I noticed that some components like the calendar and the paginator also have some inline letter which causes the need for a workaround (which i have commited now in theme.css)

However i wasnt able to figure out where this happens.. and if this could be safely removed to avoid having the workaround

image

image


Workaround:

```css
/* Why? - because paginator items not only render the icon but also F, P, N, L inline */
.ui-paginator .ui-paginator-next .ui-icon,
.ui-paginator .ui-paginator-prev .ui-icon,
.ui-paginator .ui-paginator-first .ui-icon,
.ui-paginator .ui-paginator-last .ui-icon {
    color: transparent;
}

.ui-paginator a span:after{
    margin-left: -8px;
    color: #454545;
}
```

cnsgithub:
Can you add this to 6.3 now? If there are any problems let me know i've tested it with the showcase,and now it works fine. This would improve the visuals by a lot, and since probably 6.3 will take quite a while we can fix any issues arising in the meantime.

tandraschko:
thats up to @mertsincan / @onursenture

cagataycivici:
Thank you for the PR however font icons are a premium feature in PrimeFaces, all premium templates replace the old png icons with font-awesome icons already, we need to keep it as it is.

cnsgithub:
Hm.. Did I interpret this correctly? You are afraid of having less customers for the premium templates if proper scalable icons are introduced into the community version of PrimeFaces? And that the price for having those icons (to use in a commercial product) is minimum 290$? Really? Do the templates have not more than that to offer? I hardly believe that anyone would pay for just the icons. I think the key-features is the layout and SASS CSS. This is a very bad decision. Sooner or later, more complaints will arise, if you open a PrimeFaces website on a High-res screen (QHD/UHD) with reasonably scaled DPI (else the UI is too small) it already looks terrible.

The fact that FontAwesome is used for premium themes does not mean that including this is technically not possible, as the default theme *.css (aristo / omega) is not used anyway when a PrimeFaces theme is used. And it will also decrease the *.jar size due to less static images, and increase maintainability. Adding an icon to the sprites would be a lot more time-consuming.

However for those who still want this change and stumble across this PR/Issue i have created a simple drop-in *.css addition for PrimeFaces. Just add it to the website and you're good to go. No change to the JAR required. I will also use this for my PrimeFaces project instead of using this fix branch now, as the goal of my contributions is that all fixes/contributions are included into the regular builds, so that one day I can hopefully switch back to the regular *.jar (or ELITE builds i have purchased - which I am currently not even using due to some other unfixed issues)

See: https://github.com/cnsgithub/primefaces-openicons

PR #3235, opened 2018-01-29
Fix #3238 renderDynamicPassThruAttributes for p:captcha

### LINKED ISSUE #3238: p:captcha should support passtrough to add custom features or css
https://github.com/primefaces/primefaces/pull/3235

cnsgithub:
Add more flexibility to modify the captcha with custom code, so passthrough should be allowed. 

https://github.com/primefaces/primefaces/issues/3238


tandraschko:
Could you please create a ticket before creating a PR?
We only use tickets for tracking.

PR #3237, opened 2018-01-29
fixes #3234 - InputMask: lack of user input validation

### LINKED ISSUE #3234: InputMask: lack of user input validation
## 1) Environment
- Affected browsers: all

## 2) Expected behavior
It must be guaranteed, that the value of p:inputMask matches the pattern defined via `mask` attribute. Just doing client side validation is insufficient. The server also has to validate the submitted value.

## 3) Actual behavior
Validation of user input is performed on client side via JS only. So it's easily possible to bypass validation at all.

## 4) Steps to reproduce
In developer tools (F12) update the inputMask's value in the DOM to an invalid value:
`$(".ui-inputfield").val("DO NOT CALL ME");` Then submit the form by pressing the button.

## 5) Sample XHTML
```
	<h:head>
		<title>inputMask test</title>
	</h:head>
	<h:form>
	    <p:inputMask value="#{bean.phone}" mask="(999) 999-9999? x99999" />
	</h:form>
</html> 
``` 

## 6) Sample bean
```
package de.test.primefaces;

import java.io.Serializable;
import javax.inject.Named;

@Named("bean")
@SessionScoped
	private String phone;

	public String getPhone() {
		return phone;
	}

	public void setPhone(String phone) {
		this.phone = phone;
	}
	
}
```

## 7) Mitigation
Just repeat the input validation in InputMaskRenderer. Reject the value if it isn't conform with the specified mask. See https://www.owasp.org/index.php/Input_Validation_Cheat_Sheet#Client_Side_vs_Server_Side_Validation for further information and general guidelines.

cnsgithub:


melloware (approved):

tandraschko:
I think it's better to apply it in 6.3 - i'm not sure if some mask won't work.

cnsgithub:
Did you see the test file? Feel free to extend it with arbitrary masks and we'll see. I am voting for 6.2. ;-)

Issue #3267, opened 2018-02-01
Breadcrumb:  add keyboard support


ankitacroit09:
# IMPORTANT 
- !!! If you open an issue, fill every item. Otherwise the issue might be closed as invalid. !!!

- Before you open an issue, test it with the current/newest version.
- Try to find an explanation to your problem by yourself, by simply debugging. This will help us to solve your issue 10x faster
- Clone this repository https://github.com/primefaces/primefaces-test.git in order to reproduce your problem, you'll have better chance to receive an answer and a solution.
- Otherwise the example must be as small and simple as possible! It must be runnable without any other dependencies (like Spring,..., or project/company internal classes)!
- Feel free to provide a PR (Primefaces is an open-source project, any fixes or improvements are welcome.)

## 1) Environment
- PrimeFaces version: Latest
- Does it work on the newest released PrimeFaces version? Version?
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source)
- Application server + version:
- Affected browsers: All

## 2) Expected behavior
I want the keyboard accessibility should work when I am using the tab button.
There is no css changes for focus.

## 3) Actual behavior
When I am using tab button (Keyboard) not coming on the breadcrumb option.

## 4) Steps to reproduce

.. 

## 5) Sample XHTML

 <p:breadCrumb>
                   <p:menuitem value="Level" url="i.html"/>
                   <p:menuitem value="Level" url="html/ll.jsf"/>
                   <p:menuitem value="Level1" url="html/l1.jsf"/>
                   <p:menuitem value="Level2" url="html/l2.jsf"/>
                   <p:menuitem value="Level3" url="html/l3.jsf"/>
                   <p:menuitem value="Level4" url="html/l4.jsf"/>
                   <p:menuitem value="Current Page"/>
</p:breadCrumb>

## 6) Sample bean

..

ankitacroit09:
I have made the changes in primecode for this.
Can I contribute to the Prime code for this support.

The changes what I did inside the BaseMenuRenderer.java class method
```
@Override
protected void encodeMenuItem(FacesContext context, AbstractMenu menu, MenuItem menuitem) throws IOException {
	        ResponseWriter writer = context.getResponseWriter();
	        String title = menuitem.getTitle();
	        String style = menuitem.getStyle();
	        boolean disabled = menuitem.isDisabled();
	        String rel = menuitem.getRel();

	        writer.startElement("a", null);
[...]

PR #3273, opened 2018-02-02
#3267 Breadcrumb: add keyboard support 

ankitacroit09:
#3267


tandraschko:
FilesChanged: 1301. Something went wrong ;)

PR #3274, opened 2018-02-02
Update BaseMenuRenderer.java

ankitacroit09:
#3267 : Breadcrumb: add keyboard support


tandraschko:
We can't just change it. That must be reviewed by @mertsincan. I think it's currently by design to pass -1.

jxmai:
There seems something off. 0 might be the expected value 🤔 .

https://github.com/primefaces/primefaces/commit/dbf5295cf48662d89e5c292d80b14c133608f707

Default is 0 in xml

image

write -1 in the renderer ?

image

tandraschko:
I think thats a trick to make elements not visible for tabbing. So thats probably by design.

tandraschko:
Will close the PR for now, we need to check the issue.

PR #3275, opened 2018-02-03
Fix #2127: Picklist prevent client side manipulation.

### LINKED ISSUE #2127: Picklist: disabled items allow for client-side manipulation
This may cause security issues in cases where people rely on the disabled items assignment not being able to be manipulated on the client-side.

## 1) Environment
- PrimeFaces version: 6.0 and 6.0.15 and newest
- Application server + version: Apache Tomcat 8.5.11
- Affected browsers: All

## 2) Expected behavior
Disabled items cannot be transferred from source to target list or vice-versa.
Real behaviour: 

## 3) Actual behavior
Items can be transferred from source to target list or vice-versa by client-side manipulation, by removing the disabling CSS style class.

## 4) Steps to reproduce
I use a picklist with "itemDisabled" property.
The disabled entries can't be dragged or transferred from source list to target list and vice versa, as one would expect.

However, by simply removing the style class, e.g. via jQuery
`$('ul.ui-picklist-list > li.ui-state-disabled').removeClass('ui-state-disabled');`
I am able to swap items between the source and target lists.
These changes are then applied to the model (here: `personBean.personAssignment` DualListModel) upon submit, because a respective server-side check is missing.

## 5) Sample XHTML
```
<p:pickList id="personAssignment"
            value="#{personBean.personAssignment}"
            var="person"
            itemDisabled="true"
            itemLabel="#{person.toString()}"
            itemValue="#{person}"/>
```

## 6) Sample bean

```
public class PersonBean {
  private DualListModel<Object> personAssignment;

  public DualListModel<Object> getPersonAssignment() {
    if (personAssignment == null) {
        List<Object> source = Arrays.asList(new String[]{"One", "Two"});
        List<Object> target = Arrays.asList(new String[]{"A", "B"});
        personAssignment = new DualListModel<>(source, target);
    }
    return personAssignment;
  }

  public void setPersonSelection(DualListModel<Object> personAssignment) {
    this.personAssignment = personAssignment;
  }

}

```

melloware:
1. Cleaned up the example code
2. Chose to use javax.faces.component.UIInput.UPDATE as the error message which will be "An error occurred when processing your submitted information."
3. Refactored to make the validate method more readable now that it includes both REQUIRED validation and client side scrubbing validation.

### DIFF
--- a/src/main/java-templates/org/primefaces/component/picklist/PickListTemplate.java
+++ b/src/main/java-templates/org/primefaces/component/picklist/PickListTemplate.java
@@ -115,6 +115,10 @@ import javax.faces.event.BehaviorEvent;
      * @param oldEntries old/former entries of model source/target list
      */
     protected void checkDisabled(FacesContext facesContext, String label, List<?> newEntries, List<?> oldEntries) {
+        if (!isValid()) {
+            return;
+        }
+        
         String varName = getVar();
         String clientId = this.getClientId(facesContext);
         for (int i = 0; i < newEntries.size(); i++) {
--- a/src/main/java-templates/org/primefaces/component/picklist/PickListTemplate.java
+++ b/src/main/java-templates/org/primefaces/component/picklist/PickListTemplate.java
@@ -120,12 +120,14 @@ import javax.faces.event.BehaviorEvent;
             return;
         }
         
+        Map<String, Object> requestMap = facesContext.getExternalContext().getRequestMap();
         String varName = getVar();
         String clientId = this.getClientId(facesContext);
+        Object originalItem = requestMap.get(varName);
         for (int i = 0; i < newEntries.size(); i++) {
             Object item = newEntries.get(i);
             // Set the current item in request map to get its properties via stateHelper().eval() call
-            facesContext.getExternalContext().getRequestMap().put(varName, item);
+            requestMap.put(varName, item);
             boolean itemDisabled = isItemDisabled();
             // Check if disabled item has been moved from its former/original list
             if (itemDisabled && !oldEntries.contains(item)) {
@@ -135,6 +137,9 @@ import javax.faces.event.BehaviorEvent;
                 break;
             }
         }
+        
+        // put the original value back
+        requestMap.put(varName, originalItem);
     }
 
     @Override
--- a/src/main/java-templates/org/primefaces/component/picklist/PickListTemplate.java
+++ b/src/main/java-templates/org/primefaces/component/picklist/PickListTemplate.java
@@ -77,6 +77,7 @@ import javax.faces.event.BehaviorEvent;
         return EVENT_NAMES;
     }
     
+    @Override
     protected void validateValue(FacesContext facesContext, Object newValue) {
         super.validateValue(facesContext, newValue);
 
--- a/src/main/java-templates/org/primefaces/component/picklist/PickListTemplate.java
+++ b/src/main/java-templates/org/primefaces/component/picklist/PickListTemplate.java
@@ -76,34 +76,63 @@ import javax.faces.event.BehaviorEvent;
     public Collection<String> getEventNames() {
         return EVENT_NAMES;
     }
+    
+    protected void validateValue(FacesContext facesContext, Object newValue) {
+        super.validateValue(facesContext, newValue);
 
-	protected void validateValue(FacesContext facesContext, Object newValue) {
-		super.validateValue(facesContext, newValue);
-		
-		DualListModel model = (DualListModel) newValue;
-		if(isRequired() && model.getTarget().isEmpty()) {
-			String requiredMessage = getRequiredMessage();
-			FacesMessage message = null;
-			
-			if(requiredMessage != null) {
-				message = new FacesMessage(FacesMessage.SEVERITY_ERROR, requiredMessage, requiredMessage);
+        DualListModel<?> newModel = (DualListModel<?>) newValue;
+        DualListModel<?> oldModel = (DualListModel<?>) this.getValue();
+        
+        String clientId = this.getClientId(facesContext);
+        String label = this.getLabel();
+        if (label == null) {
+            label = clientId;
+        }
+        
+        if (isRequired() && newModel.getTarget().isEmpty()) {
+            String requiredMessage = getRequiredMessage();
+            FacesMessage message = null;
+
+            if (requiredMessage != null) {
+                message = new FacesMessage(FacesMessage.SEVERITY_ERROR, requiredMessage, requiredMessage);
             }
             else {
-                String label = this.getLabel();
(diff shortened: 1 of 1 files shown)


DaHoC:
:+1: Code looks good to me, happy that some of my code was helpful :smiley:

tandraschko:
Didn't checked the PR for all details but if you do:
facesContext.getExternalContext().getRequestMap().put(varName, item);
You also need to store the old item and restore it after you are finished.

src/main/java-templates/org/primefaces/component/picklist/PickListTemplate.java
```diff
@@ -76,34 +76,68 @@
...
+     * @param oldEntries old/former entries of model source/target list
+     */
+    protected void checkDisabled(FacesContext facesContext, String label, List<?> newEntries, List<?> oldEntries) {
+        if (!isValid()) {
+            return;
+        }
+        
+        String varName = getVar();
+        String clientId = this.getClientId(facesContext);
+        for (int i = 0; i < newEntries.size(); i++) {
```
DaHoC: Why not use the standard non-indexed iterator here, e.g. `for (final Object item : newEntries) {`

DaHoC:
Wouldn't it make sense to back this change by a simple unit test?

tandraschko:
1) index-loop are preferred for performance reasons if ArrayLists are used. If we don't know if it's a ArrayList, we can also use the for-each.
2) +1 for a unittest

DaHoC:
I couldn't find any sources that these indexed-based loops have better performance than the for-each loop - in fact, the contrary seems to be the case (besides better readability and being less error-prone): https://stackoverflow.com/a/256861/1877010

tandraschko:
performance > readability in frameworks

It performs better on ArrayLists/RandomAccess lists but of course not much better. The second point is that with a index-loop, no Iterator instance is beeing created.

Thats one of the points why MyFaces Core is so fast: https://issues.apache.org/jira/browse/MYFACES-3130

DaHoC:
I see, that was in Java 1.5 - With Java >= 1.7 that's no longer the case and no overhead and performance impact should be there - But let's not discuss such minor details, I'm fine with the indexed list, just wanted to know the cause :+1:

tandraschko:
Could you send me a post about that? I'm almost sure that there is still an Iterator instance. 
This is no big performance impact or something but 1 small enhancements is 1 big enhancement :D

tandraschko:
Yep, looks still the same. Check answer 4 e.g. https://stackoverflow.com/questions/256859/is-there-a-performance-difference-between-a-for-loop-and-a-for-each-loop/256861#256861

As i said, it's microoptimization which should not be used in application code. But in component framework or even the JSF impl, it can be a good enhancement.

DaHoC:
Yes, you are correct that there will be a single Iterator instance generated.
However, there have been performance improvements since Java 1.5:

> First, starting with Java 7, escape analysis was turned on by default. In this case, almost definately, any object allocated would be allocated on the stack, not the heap (practically free) and more than likely would completely live in the registers.

> Next, List.get() can be terribly slow, especially in the case of linked lists. You have to traverse the entire list to get to the "get" location. That means turning an O( n ) algorithm into an O( n2 ) algorithm.

> And finally, There is a REALLY good chance that the compiler/optimiser is already producing the optimal index lookup code. This is a somewhat easy optimisation that can be picked up by the JVM. On top of that, the JVM can further optimize the scenario of ArrayList iterator by removing the bounds checking making the iterator version even faster than the indexing version.

(source: https://www.reddit.com/r/java/comments/2uuzv1/top_10_easy_performance_optimisations_in_java/cobzlng/ as response to https://blog.jooq.org/2015/02/05/top-10-easy-performance-optimisations-in-java/)

Performance benchmarks with Java 7+ like https://github.com/chaschev/experiments-with-microbes/blob/master/src/main/java/stackoverflow/q2642004/ForEachLoopMicrobe.java indicate that for-each loops tend to be the fastest option

Interesting, this topic seems to be discussed controversially

melloware:
@tandraschko I just pushed a commit.  I followed other examples now and put the original value back in the RequestMap.  Good catch!

@DaHoC yes it is a controversial topic but I agree with Thomas that for a library like PF every optimization counts.  In my business logic code I wouldn't do this I would use a for...each as readability is more important.

src/main/java-templates/org/primefaces/component/picklist/PickListTemplate.java
```diff
@@ -76,34 +76,63 @@
...
+    protected void checkDisabled(FacesContext facesContext, String label, List<?> newEntries, List<?> oldEntries) {
+        String varName = getVar();
+        String clientId = this.getClientId(facesContext);
+        for (int i = 0; i < newEntries.size(); i++) {
+            Object item = newEntries.get(i);
+            // Set the current item in request map to get its properties via stateHelper().eval() call
+            facesContext.getExternalContext().getRequestMap().put(varName, item);
+            boolean itemDisabled = isItemDisabled();
+            // Check if disabled item has been moved from its former/original list
+            if (itemDisabled && !oldEntries.contains(item)) {
```
cnsgithub: @melloware What will happen if `oldEntries.contains(item)==false` but `newEntries.contains(item)==true`? I have not yet tested it so I am not sure if this is considered as well or may happen at all?

melloware:
@DaHoC @cnsgithub It looks like that case is not handled. I will update.

cnsgithub:
@melloware Do you still plan to work on this fix?

melloware:
@cnsgithub I was but haven't had the time yet.  Do you want to take what I have done above and modify it and submit a PR?  If so I can close this one once you submit yours if you have the time?

melloware:
Actually looking more at this I think we might be OK.  Right now it only cares about items in the newList and they were moved from the oldList.  So I don't think there is a scenario that isn't covered by this current PR.  Everything I have tried has worked as expected.

cnsgithub:
Ok, let's merge it. I'll try some cases after that.

tandraschko:
The PR itself looks good now. Would be great if someone could retest it before i merge.

tandraschko:
@cnsgithub hey security-chief, did you had time to check and test this? :D

cnsgithub:
Ha, not yet, sorry. But it's on my todo list.

tandraschko:
1) currently it conflicts, can you update it?
2) should we do it like in Select*, to just ignore the disabled items instead of adding a message?

melloware:
@tandraschko I will fix.  Question though since we are settting "setValid(false);" aren't we being a good citizen by displaying a message?

tandraschko:
I'm not sure. In Select* components, there were some cases were it broke the current funcionatlity, therefore i feel better to just ignore invalid submitted items.

PR #3278, opened 2018-02-04
Fix #1437 - Inplace: initial mode support

### LINKED ISSUE #1437: Inplace: default mode support
It should be possible to indicate in which "mode" inplace component should be displayed first. There would be two mode so far: input or output. AFAIK, "output" mode is displayed first, but what if user want it to display in "input" (editing) mode first? Inplace is more than just a fancy way to edit value, technically, it allows to associate a display and an editing mode in a very fancy way. In a dyna form, it would write:

```
<p:inplace mode="#{!formField.readonly ? 'input' : 'output'} disabled="#{formField.readonly}">
    <f:facet name="output">
        Yes or No
    </f:facet>
    <f:facet name="input">
        <h:selectBooleanCheckbox />
    </f:facet>
</p:inplace>
```

I can provide a fix if you're okay with that

Rapster:


melloware (approved):

tandraschko:
Maybe "initialMode" sounds better?

tandraschko:
I would probably choose:
initialFacet with value output or input to match the facet names.

tandraschko:
We could even think about value binding called "mode" - so you can store the sate in a bean.

Rapster:
I like "initialMode". I thought about keeping facet names but since it's not always based on facets and you can use inplace without having input, I thought it might be confusing for the users.  Though, if you could enlight me regarding your last point, I'm not sure I see you're saying

tandraschko:
Something like the activeIndex in the accordion.
If the clientSide switches the mode and p:ajax is registered, the mode will be written back to the "mode" expression.
If it's updated on the ajax, the mode will be the same again as in the bean.
With this option you could also set an initial mode.

Rapster:
That's something I can have a look later. For now, `initialMode` should be enough

tandraschko:
I don't like to introduce features, which we already know that it's not perfect, and we will replace it in the future.
It's actually a easy task - i can also do it for 6.3.

Rapster:
It's definitely up to you (since you're the PR decision maker), If you're sure to do that change I'd understand you'll integrate it for 6.3, but otherwise, would be nice to integrate it for 6.2 :-)

melloware (approved):

tandraschko:
Will close it for now and lets do it for 6.3.
It would however be great if you could implement it like i mentioned.

Thanks!

Rapster:
Okay, I'll try to do that. Could it be in 6.2 if I'm doing it soon enough? 6.3 will be released in a year or so and I'm planning to use it extensively

tandraschko:
If it's before RC2, yes. It's a small addition.

Rapster:
> If the clientSide switches the mode and p:ajax is registered, the mode will be written back to the "mode" expression.
If it's updated on the ajax, the mode will be the same again as in the bean.
With this option you could also set an initial mode.

Which `p:ajax` are we talking about? From what I understood, I have to introduce some kind of a state holder in inplace component

tandraschko:
1) yes
2) don't think so, have a look at activeIndex

Rapster:
I'm not sure what those answers are answering...

Rapster:
So yes, I have a to create a state holder, but what are you thinking when you say "don't think so"?

PR #3280, opened 2018-02-04
Fix #3279 - Inplace: wrong style buttons

### LINKED ISSUE #3279: Inplace: wrong style buttons
## 1) Environment
- PrimeFaces version: 6.2.RC2-SNAPSHOT
- Affected browsers:

## 2) Expected behavior
Buttons should be as height as input (see https://www.primefaces.org/showcase/ui/button/button.xhtml) 

## 3) Actual behavior

Buttons are too small
image
see https://www.primefaces.org/showcase/ui/input/inplace.xhtml

## 4) Steps to reproduce
```xml
<h:outputLabel for="ajax" value="Ajax:" />
    <p:inplace id="ajax" editor="true" defaultMode="content">
    <p:inputText value="#{inplaceView.text}" required="true" label="text" />
</p:inplace>
```

PR on his way

Rapster:
@tandraschko @jxmai @melloware @mertsincan  Let me know what you think about this new "style"
image

It looks like before that, it was done on purpose but this style always interrogate me... I'd like to keep it as the picture above shows as it is compliant with `p:button` showing only icon

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/inplace/inplace.css
+++ b/src/main/resources/META-INF/resources/primefaces/inplace/inplace.css
@@ -2,4 +2,3 @@
 .ui-inplace .ui-inplace-disabled {cursor: default;}
 .ui-inplace .ui-inplace-content {}
 .ui-inplace .ui-inplace-editor {margin-left:0.1em;}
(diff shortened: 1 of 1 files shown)


melloware (approved):

melloware:
Looks good to me.  Having the buttons the same size as the input box always seems like a good idea.

Issue #3285, opened 2018-02-05
DataTable: ColumnToggler hides when trying to scroll selection panel


koan00:
## 1) Environment

- PrimeFaces version: master (5099d89)
- Does it work on the newest released PrimeFaces version? Yes Version? 6.1
- Application server + version: Tomcat 7.0.84 / Mojarra 2.2.15
- Affected browsers: IE 11 and Chrome 64
- Unaffected browsers: FF 58

The toggleable column panel is hard coded to a maximum height of 200.  When the list of toggleable columns grows long, a vertical scroll bar is becomes part of the panel to handle the overflow. 

The panel also tries to detect a click outside itself, which should trigger a panel hide. When calculating the location and size of the panel, the jquery functions width() and height() are used. However these functions do not appear to include the scroll bar size.  When attempting to drag the scroll bar to scroll the panel contents, the event handler runs and determines the click is outside the panel and closes the panel.

## 2) Expected behavior
When clicking to drag the scroll bar, the panel remains open and scrolls.

## 3) Actual behavior
When clicking to drag the scroll bar, the panel closes.

I'll try to submit a PR shortly.

koan00:
PR: https://github.com/primefaces/primefaces/pull/3287

PR #3287, opened 2018-02-05
Fix for PF issue #3285. Use panel's offsetWidth and offsetHeight to m…

koan00:
Use panel's offsetWidth and offsetHeight to measure the Datatable column toggler panel.


melloware (approved):

tandraschko:
The PR itself is fine - however i'm sure we have similar code in other widgets. Would you please check that, too?

PR #3293, opened 2018-02-06
fixes #3286

### LINKED ISSUE #3286: imageCropper: uncontrolled resource consumption may lead to denial of service
## 1) Environment
- PrimeFaces version: primefaces-6.2.RC2-snapshot
- Affected browsers: all

## 2) Expected behavior
Number of bytes read from HTTP input stream should be limited, otherwise high resource consumption may lead to denial of service.

## 3) Actual behavior
An attacker could provide a specially crafted, infinite image stream that will be fed to `ImageIO ` without size limitation.

## 4) Steps to reproduce
Host or generate on the fly a very huge image on a HTTP server. Let `p:imageCropper image` point to it. Observe high CPU/heap consumption, e.g. via visualvm. This may for example exceed GC overhead limit and stop the JVM.

## 5) Sample XHTML
Use just the showcase and prepare it as mentioned above.

## 6) Sample bean
See 5)

## 7) Mitigation
Use a input stream that supports a limit. Limit it to whatever value could be useful in the use case of image cropping, e.g. 10 MB for heavy BMP should be sufficient? Or even better, provide an attribute with 10 MB as default and let the developer override this behavior if desired.

In commons-io there is a input stream that does exactly what would be needed here, `BoundedInputStream`: https://commons.apache.org/proper/commons-io/javadocs/api-2.6/org/apache/commons/io/input/BoundedInputStream.html

ImageCropperRenderer:
 ```java
               boolean isExternal = imagePath.startsWith("http");

                if (isExternal) {
                    URL url = new URL(imagePath);
                    URLConnection urlConnection = url.openConnection();
                    inputStream = urlConnection.getInputStream(); //limit this
                    contentType = urlConnection.getContentType();
                }
```

cnsgithub:
- introduced new attribute `p:imageCropper sizeLimit` which defaults to 10MB
- use `BoundedInputStream` (commons-io) to limit bytes to read

### DIFF
--- a/checkstyle-suppressions.xml
+++ b/checkstyle-suppressions.xml
@@ -6,4 +6,5 @@
 
 <suppressions>
     <suppress files="[\\/]generated-sources[\\/]" checks="[a-zA-Z0-9]*"/>
+    <suppress files="BoundedInputStream\.java"  checks="[a-zA-Z0-9]*"/>
 </suppressions>
--- a/pom.xml
+++ b/pom.xml
@@ -323,6 +323,9 @@
                     <includes>
                         <include>src/main/java/**</include>
                     </includes>
+                    <excludes>
+                        <exclude>**/BoundedInputStream.java</exclude>
+                    </excludes>
                 </configuration>
                 <executions>
                     <execution>
--- a/src/main/java/org/primefaces/component/imagecropper/ImageCropperRenderer.java
+++ b/src/main/java/org/primefaces/component/imagecropper/ImageCropperRenderer.java
@@ -36,10 +36,10 @@ import javax.faces.context.ResponseWriter;
 import javax.faces.convert.ConverterException;
 import javax.imageio.ImageIO;
 
-import org.apache.commons.io.input.BoundedInputStream;
 import org.primefaces.model.CroppedImage;
 import org.primefaces.renderkit.CoreRenderer;
 import org.primefaces.util.WidgetBuilder;
+import org.primefaces.util.BoundedInputStream;
 
 public class ImageCropperRenderer extends CoreRenderer {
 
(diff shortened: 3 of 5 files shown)


melloware:
@cnsgithub I think using BoundedInputStream is a nice fix for both issues.  A better way to get the PR applied might be to copy BoundedInputStream into the library as org.primefaces.util.BoundedInputStream.  The class is not that complicated vs bringing in the common-io dependency.

https://commons.apache.org/proper/commons-io/javadocs/api-2.5/src-html/org/apache/commons/io/input/BoundedInputStream.html

cnsgithub:
@melloware Well, I think there is already a transitive dependency to commons-io via commons-fileupload, see also PrimeFaces user guide, section 2.2 (dependencies).

tandraschko:
But Commons-fileupload is not a required depedency and we wont remove the native upload.

melloware:
Correct these are optional dependencies only if you use that component.  Your fix now enforces a required dependency.

cnsgithub:
I understand.

But in my opinion it's a bad idea to just copy&paste a file in a specific version. It's hard to maintain and stay updated...

melloware:
I totally understand.  I think its trying to balance PF's goal of being a standalone JAR without requiring people to include 20 transitive dependencies bloating their WAR with libraries that may be used for 1 or 2 lines of code.  At least that is what I have seen from Optimus Prime and the team as their goal over the years.

tandraschko:
As the BoundedInputStream is actually a "simple" wrapper and commons-io is just a transitive dependency, i would copy it.

cnsgithub:
I Just changed some things and hope that it fulfills the requirements:

- copied `BoundedInputStream ` from commons-io-2.6 (2.5 is already outdated)
- excluded the class from checkstyle, so the file can be later compared more easily to newer revisions
- excluded the class from PrimeFaces license checking since it brings its own Apache ASF 2.0 license (we are conform with it, aren't we?

melloware (approved):

melloware:
Yep the licenses are friendly for copying that class.   Nice work on the PR!

tandraschko:
Yep nice! Thanks.

PR #3299, opened 2018-02-07
fixes #3264 fixes #3296- selectManyCheckbox: disabled values are accepted

### LINKED ISSUE #3264: selectManyCheckbox: disabled values are accepted
## 1) Environment
- Affected browsers: all

## 2) Expected behavior
Values of disabled components may never be accepted by the server, albeit submitted by the client.

## 3) Actual behavior
At least when using custom layouting with `p:selectManyCheckbox` and `p:checkbox`, select items that are disabled via `f:selectItem itemDisabled="true"` and `p:checkbox disabled="true"` are in fact rendered correctly (grayed out), however may be bypassed by the client cause of missing checks on the server side. This may lead to broken access control if used for restricted areas/features.

Regarding the other SelectMany components no all-clear can be given at the moment, especially RadioButtonRenderer is probably susceptible also. I'll document them in this issue in any case.

## 4) Steps to reproduce
In developer tools (F12) update the disabled and checked states in the DOM:
`$('input[name$=custom]').prop('checked',true).prop('disabled',false);` Then submit the form by pressing the button. When installing a breakpoint in `Bean:setFeatures` you will see that _Elite_ was accepted as well.

## 5) Sample XHTML
```
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:h="http://java.sun.com/jsf/html" xmlns:p="http://primefaces.org/ui" xmlns:f="http://java.sun.com/jsf/core">  
	<h:head>
		<title>Testing disabled attribute</title>
	</h:head>
	<h:form>
	    <p:outputPanel id="customPanel" style="margin-bottom:20px">
	    	Features:
	        <p:selectManyCheckbox id="custom" value="#{bean.features}" layout="custom">
	            <f:selectItem itemLabel="Free" itemValue="Free" />
	            <f:selectItem itemLabel="Elite" itemValue="Elite" itemDisabled="true" />
	        </p:selectManyCheckbox>
	 
	        <div class="ui-grid ui-grid-responsive">
	            <div class="ui-grid-row">
	                <div class="ui-grid-col-4" style="text-align:center">
	                    <h:outputLabel for="opt1" value="Free" style="display:block;margin-bottom:8px"/>
	                    <p:checkbox id="opt1" for="custom" itemIndex="0" />
	                </div>
	                <div class="ui-grid-col-4" style="text-align:center">
	                    <h:outputLabel for="opt2" value="Elite" style="display:block;margin-bottom:8px"/>
	                    <p:checkbox id="opt2" for="custom" itemIndex="1" disabled="true" />
	                </div>
	            </div>
	        </div>
	    </p:outputPanel>
	</h:form>
</html>  
``` 

## 6) Sample bean
```
package de.test.primefaces;

import java.io.Serializable;
import javax.inject.Named;
import org.primefaces.PrimeFaces;

@Named("bean")
@SessionScoped

	private String[] features;

	public String[] getFeatures() {
		return features;
	}

	public void setFeatures(String[] features) {
		this.features = features;
	}
	
}

```

## 7) Mitigation
`Checkbox:isDisabled `must be considered in the appropriate renderer.

### LINKED ISSUE #3296: selectManyCheckbox: state of disabled, checked select items gets lost
## 1) Environment
- PrimeFaces version: primefaces-6.2.RC2-snapshot, primefaces-6.1.11
- Affected browsers: all

## 2) Expected behavior
Submission of form that contains preselected/checked but disabled select items must not change the selection, i.e. the disabled select items must stay checked.

## 3) Actual behavior
Preselected/checked select items are rendered correctly. When submitting the form, the browser omits the disabled options and PrimeFaces does not restore them before calling `UISelectMany.setSubmittedValue`.

## 4) Steps to reproduce
Set breakpoint in `Bean.setSelectedFeatures`. Open the page. 'Elite' is disabled and preselected. Select the 'Free' checkbox and submit the form. The setter will be called containing just 'Free', 'Elite' will be lost.

## 5) Sample XHTML
```xml
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:h="http://java.sun.com/jsf/html" xmlns:p="http://primefaces.org/ui" xmlns:f="http://java.sun.com/jsf/core">  
	<h:head>
		<title>Testing disabled attribute</title>
	</h:head>
	<h:form>
	    <p:selectManyCheckbox value="#{bean.selectedFeatures}">
	    	<f:selectItems value="#{bean.availableFeatures}" />
	    </p:selectManyCheckbox>
	    
	</h:form>
</html>  
```

## 6) Sample bean
```java
package de.test.primefaces;

import java.io.Serializable;
import java.util.ArrayList;
import java.util.Arrays;
import java.util.List;
import javax.faces.model.SelectItem;
import javax.inject.Named;
import org.primefaces.PrimeFaces;

@Named("bean")
@SessionScoped

	private List<String> selectedFeatures = new ArrayList<>(Arrays.asList("Elite")); // preselected
	
	private List<SelectItem> availableFeatures = Arrays.asList(
			new SelectItem("Free", "Free", "Free", false), 
			new SelectItem("Elite", "Elite", "Elite", true)); // preselected item is disabled

	public List<String> getSelectedFeatures() {
		return selectedFeatures;
	}

	public void setSelectedFeatures(List<String> selectedFeatures) {
		this.selectedFeatures = selectedFeatures;
	}

	public List<SelectItem> getAvailableFeatures() {
		return availableFeatures;
	}

	public void setAvailableFeatures(List<SelectItem> availableFeatures) {
		this.availableFeatures = availableFeatures;
	}
	
}
```

## 7) Additional information

- I came across this issue in a debugging session when trying to fix https://github.com/primefaces/primefaces/issues/3264.
- Switching `p:selectManyCheckbox` to `h:selectManyCheckbox` produces the same incorrect behavior. That means the issue exists for PrimeFaces and Mojarra (I am using version 2.3.3) as well and must be fixed in both of them. Please correct me if I am wrong. Interestingly, Mojarra has accepted but closed unfixed a similar issue: https://github.com/javaserverfaces/mojarra/issues/2559. If someone can confirm the issue we should also open an issue at Mojarra.

cnsgithub:
- tried to fix it in `SelectManyRenderer ` and `SelectOneRenderer ` since all `selectOne*` and `selectMany*` components inherit from them:
image
- not entirely sure if we need special handling in `CheckboxRenderer `(custom layouting only) too since it should already get handled by `SelectManyRenderer`, simple test says 'no'
- performance should be considered in `SelectRenderer.isDisabledSelectItemSubmitted`, not sure if there is a easier/cheaper way to do this


tandraschko:
Could you please check if the JSF impl completely ignores this case or how did they fix it?

src/main/java/org/primefaces/renderkit/SelectManyRenderer.java
```diff
@@ -39,7 +39,11 @@ public void decode(FacesContext context, UIComponent component) {
         Map<String, String[]> params = context.getExternalContext().getRequestParameterValuesMap();
 
         if (params.containsKey(submitParam)) {
-            selectMany.setSubmittedValue(params.get(submitParam));
+            String[] submittedValues = params.get(submitParam);
+            if (isDisabledSelectItemSubmitted(context, selectMany, (Object[]) getValues(selectMany), submittedValues)) {
+                return;
```
melloware: Hmm should it return or should it throw an exception because the client side has been manipulated?  if we don't throw an exception it will just look like silently nothing is happening?

cnsgithub:
@tandraschko Tested with `h:selectManyCheckbox`. The behavior is very much the same as with PrimeFaces. Apparently the reference implementation also has this security bug.

@melloware You're right, a meaningful exception should be preferred to silent ignoring. I just orientated on existing renderers that simply return. However, that's kind of different since there are more than just one items. I'll elaborate on it.

cnsgithub:
Just filed a bug at mojarra: https://github.com/javaserverfaces/mojarra/issues/4330

melloware (approved):

cnsgithub:
The last commit fixes #3296 also.

jxmai:
FYI: This branch has conflicts

cnsgithub:
Now conflicts have been resolved again.

melloware (approved):

cnsgithub:
Can we apply these changes? 

The same issue has been fixed in Mojarra as well (https://github.com/javaserverfaces/mojarra/issues/4330) and would kill two birds with one stone. :wink:

tandraschko:
One thing here: Wouldn't it be better to use ArrayUtils or something for a contains instead of creating new lists?
AFAIR we already have something

cnsgithub:
You're right, looping over an array is slightly more efficient than using `List.contains`. However, `ArrayUtils.contains ` does not support `Object[]`.

See https://www.programcreek.com/2014/04/check-if-array-contains-a-value-java/. Binary search would be the best performing method, however it assumes that arrays are sorted...

Should we add an overloaded method of `contains`?

tandraschko:
Yep, i would overload it. And maybe rename the current "contains" to containsIgnoreCase

cnsgithub:
Done. Please review again.

PR #3308, opened 2018-02-09
fixes #3282 slider input validation

### LINKED ISSUE #3282: slider: lack of user input validation
## 1) Environment
- PrimeFaces version: primefaces-6.2.RC2-snapshot
- Application server + version: any
- Affected browsers: any

## 2) Expected behavior
`minValue` and `maxValue` should be enforced at server side and value has to be rejected if outside the range.

## 3) Actual behavior
User input validation is neither done on server nor on client side.

## 4) Steps to reproduce
Just go to your showcase https://www.primefaces.org/showcase/ui/input/slider.xhtml and enter some values that are outside the range.

## 5) Sample XHTML
https://www.primefaces.org/showcase/ui/input/slider.xhtml

## 6) Sample bean
https://www.primefaces.org/showcase/ui/input/slider.xhtml

cnsgithub:
- Range checks are performed in process validations phase, therefore the slider now extends `UIInput`
- if values are out of range a validation error will be thrown and the submitted values will be rejected
- tested with your showcase xhtml, as well range as simple slider seem to work correctly

### DIFF
--- a/src/main/java-templates/org/primefaces/component/slider/SliderTemplate.java
+++ b/src/main/java-templates/org/primefaces/component/slider/SliderTemplate.java
@@ -76,18 +76,20 @@ import org.primefaces.util.MessageFactory;
                 String valueToStr = getSubmittedValue(inputTo).toString();
                 double valueFrom = Double.valueOf(valueFromStr);
                 double valueTo = Double.valueOf(valueToStr);
-                if (valueFrom < this.getMinValue() || valueFrom > this.getMaxValue()) {
-                    this.setValid(false);
-                    inputFrom.setValid(false);
-                }
-                if (valueTo > this.getMaxValue() || valueTo < this.getMinValue()) {
-                    this.setValid(false);
-                    inputTo.setValid(false);
-                }
                 if (valueTo < valueFrom) {
                     this.setValid(false);
                     inputFrom.setValid(false);
                     inputTo.setValid(false);
+                }
+                else {
+                    if (valueFrom < this.getMinValue() || valueFrom > this.getMaxValue()) {
+                        this.setValid(false);
+                        inputFrom.setValid(false);
+                    }
+                    if (valueTo > this.getMaxValue() || valueTo < this.getMinValue()) {
+                        this.setValid(false);
+                        inputTo.setValid(false);
+                    }
                 }
             }
             else {
--- a/src/main/java-templates/org/primefaces/component/slider/SliderTemplate.java
+++ b/src/main/java-templates/org/primefaces/component/slider/SliderTemplate.java
@@ -10,6 +10,9 @@ import java.util.Map;
 import java.util.HashMap;
 import javax.faces.event.PhaseId;
 import javax.faces.event.BehaviorEvent;
+import org.primefaces.expression.SearchExpressionFacade;
+import javax.faces.application.FacesMessage;
+import org.primefaces.util.MessageFactory;
 
 
     private final static String DEFAULT_EVENT = "slideEnd";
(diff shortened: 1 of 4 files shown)


src/main/java-templates/org/primefaces/component/slider/SliderTemplate.java
```diff
@@ -57,3 +60,60 @@ public void queueEvent(FacesEvent event) {
...
         }
     }
+
+    public final static String VALUE_OUT_OF_RANGE = "primefaces.slider.OUT_OF_RANGE";
+
+    @Override
+    public void validate(FacesContext context) {
+        super.validate(context);
+        
+        if(isValid()) {
```
melloware, line 70: I prefer to use `if (!isValid()) { return; }` to short circuit the whole method rather than having all of the code inside a giant IF statement.  Just my preference.

melloware (approved):
Nice work.

src/main/java-templates/org/primefaces/component/slider/SliderTemplate.java
```diff
@@ -57,3 +60,60 @@ public void queueEvent(FacesEvent event) {
...
+            String[] inputIds = this.getFor().split(",");
+            if (this.isRange()) {
+                UIInput inputFrom = (UIInput) SearchExpressionFacade.resolveComponent(context, this, inputIds[0].trim());
+                UIInput inputTo = (UIInput) SearchExpressionFacade.resolveComponent(context, this, inputIds[1].trim());
+                String valueFromStr = getSubmittedValue(inputFrom).toString();
+                String valueToStr = getSubmittedValue(inputTo).toString();
+                double valueFrom = Double.valueOf(valueFromStr);
+                double valueTo = Double.valueOf(valueToStr);
+                if (valueFrom < this.getMinValue() || valueFrom > this.getMaxValue()) {
+                    this.setValid(false);
```
jxmai: ~~Why not use `non-shortcircuit or` to force evaluating both statements ?~~

```
if(valueFrom < this.getMinValue() | valueFrom > this.getMaxValue())
  setInvaliad
```

jxmai: Nevermind, my bad :sweat_smile: . Your implementation is Good. You are evaluating the same field.

jxmai: Just a nitpick

if valueFrom and valueTo are both out of range, there is no need to excute `valueTo < valueFrom` since both fields are already invalid at that point. 

Those three if statements will execute in order regardless of the condition is valid or invalid that can waste some cycles.

cnsgithub: Let's say `valueFrom=2, valueTo=1, minValue=0, maxValue=2`. 

Then the first two conditions aren't met, however the third condition is since `valueTo < valueFrom`. So the third condition is absolutely required.

Your suggestion would be to additionally check these prerequisites in the third condition? Then you would save the setter but introduce getter cycles. Does that make sense?

jxmai: > Your suggestion would be to additionally check these prerequisites in the third condition? Then you would save the setter but introduce getter cycles. Does that make sense?

No, what I meant was to move the third condition to the first.
```
if (valueTo < valueFrom) {  
    // both are invalid then no need to check individual fields
} else {
 //  checkIndividual fields
  if (valueFrom) {

  }

  if (valueTo) {

  }
}
```

instead of 

```
if (valueFrom) {

}

if (valueTo) {

}

if (valueFrom > valueTo) {

}
```
---

Let's say `valueFrom = 4, valueTo = 0, minValue = 1, maxValue = 3`

 `if (valueFrom < this.getMinValue() || valueFrom > this.getMaxValue())` will be true

`if (valueTo > this.getMaxValue() || valueTo < this.getMinValue())` will be true

`if (valueTo < valueFrom)` will also be true (wasted)

---

Let's say valueFrom=2, valueTo=1, minValue=0, maxValue=2.

 `if (valueFrom < this.getMinValue() || valueFrom > this.getMaxValue())` will be false

`if (valueTo > this.getMaxValue() || valueTo < this.getMinValue())` will be false

`if (valueTo < valueFrom)` will also be true

the third case is necessary but it will also be faster if we check valueTo < valueFrom first.

cnsgithub:
@jxmai Good morning. Now I got it. Just saved some cycles.

tandraschko:
Looks good, thanks! Just swichtecd the isValid() if as melloware suggested, its easier to read if the if-block is to big.

Issue #3310, opened 2018-02-09
DataTable: row is selected when clicking on inputSwitch inside a column


Kameecoding:
## 1) Environment
- PrimeFaces version: 6.1.0
- Does it work on the newest released PrimeFaces version? No Version? 6.1.0
- Does it work on the newest sources in GitHub? No
- Application server + version: Application Server independent
- Affected browsers: browser independent

p:inputSwitch inside a p:dataTable with singleSelection mode

## 2) Expected behavior
Clicking any part of the switch should not changethe selection in the table only the switch itself.

## 3) Actual behavior
Clicking the label on the switch changes selection in the table, while clicking the handle part doesn't.
Demo gif here: https://i.imgur.com/MRbxpEr.gifv
## 4) Steps to reproduce

1. Select row in the table
2. Click on the label part of the input switch in another row

## 6) Sample bean
Sample here: https://github.com/Kameecoding/primefaces-test/commit/c156d9aed29780ccd52da2833535b04cbfa759cd

## 7) Cause
In `\src\main\resources\META-INF\resources\primefaces\datatable\datatable.js`
```
onRowClick: function(event, rowElement, silent) {
    //Check if rowclick triggered this event not a clickable element in row content
    if($(event.target).is(this.rowSelectorForRowClick)) {
```
evaluates to true when clicking the label, because the event target is the span
and the selector: `"td:not(.ui-column-unselectable),span:not(.ui-c)"` does include it.

Kameecoding:
@tandraschko is there a way to get this fix into 6.2.0?

tandraschko:
Maybe, actually it's already to late ;)

kukel:
Same happens for commandButtons iirc...

tandraschko:
So we probably need a better/more global fix. We should target that with 6.3. So please investigate a little bit also check if components have the same problems.

kukel:
Very recent: https://forum.primefaces.org/viewtopic.php?f=3&t=54087 and from loooong ago: https://stackoverflow.com/questions/19246362/jsf-disable-button-when-table-row-is-selected-and-row-selection-listener-is-call

Kameecoding:
well the Selector based on which the selection is triggered is only checking for a span with class `ui-c`

both using div and tagging these affected spans with `ui-c` class fixes this issue

A more general fix could be a check to see if the element is contained within a component element.

Kameecoding:
# Proposed solution:

[...]

PR #3311, opened 2018-02-09
Fix for issue #3310

Kameecoding:
Added` 'ui-c' `class to the span that contains the inputSwitch label, this fixes the issue with very minimal changes to the code


melloware (approved):

tandraschko:
does a ".not()" method works really in a selector?

Kameecoding:
meanwhile I found some issues with it, but can't cancel the PR

tandraschko:
ok, i cancel it
please come back with a better fix :)

PR #3322, opened 2018-02-12
fixes #3320 - min value of spinner broken

### LINKED ISSUE #3320: Spinner: does not work with 0, negatives or blank values
## 1) Environment
- PrimeFaces version: 6.2.RC2
- Application server + version: Payara 4.1.173
- Affected browsers: Google Chrome

## 2) Expected behavior
image

## 3) Actual behavior
image

image

## 4) Steps to reproduce
Change spinner value to 0, a negative number or try submitting it without a value.

## 5) Sample XHTML
                <h:form>
                    <p:spinner value="#{projectBean.asdf}" size="37" >
                        <p:ajax process="@this" update=":growl" /> 
                    </p:spinner>
                </h:form>

## 6) Sample bean
    private Integer asdf;

    public Integer getAsdf() {
        return asdf;
    }

    public void setAsdf(Integer asdf) {
        this.asdf = asdf;       
        JsfUtil.addSuccessMessage("Done", "Code changed to: " + asdf);
    }

cnsgithub:
- changed default `min ` value of spinner to `0.0`, since similar components use this as well, this is more consistent and solves some nasty bugs when dealing with `java.lang.Double.MIN_VALUE ` which is (contrary to my expectations) positive and would have been wrong before, if there were any range checks; negative numbers have to be allowed explicitly by specifying `min` attribute value
- another bug producing the message `must be a number consisting of one or more digits` happened when value was mapped to `Integer ` because the internal `double ` value was not converted back correctly

### DIFF
--- a/src/main/java/org/primefaces/component/spinner/SpinnerRenderer.java
+++ b/src/main/java/org/primefaces/component/spinner/SpinnerRenderer.java
@@ -53,7 +53,7 @@ public class SpinnerRenderer extends InputRenderer {
                 submittedValue = submittedValue.substring(0, (submittedValue.length() - suffix.length()));
             }
 
-            parsedSubmittedValue = Double.valueOf(submittedValue);
+            parsedSubmittedValue = Double.parseDouble(submittedValue);
             if (parsedSubmittedValue < spinner.getMin()) {
                 parsedSubmittedValue = spinner.getMin();
             }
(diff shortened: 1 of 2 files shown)


tandraschko:
I'm not sure if chaning the default value breaks other cases now?!

tandraschko:
It probably breakes something yes. See: .attr("min", spinner.getMin(), Double.MIN_VALUE) in the Renderer.
Maybe it's the best thing to revert for now. NOTE: we a really really short before 6.2 final and we should only apply hotfixes.

tandraschko:
Also valueOf is used but casted to a primitive.

tandraschko:
I will try it now. If something is broken again, i will revert it. It's to risky for the final realease.

cnsgithub:
Indeed something is broken again. Unfortunately, spinner.js cannot handle `-java.lang.Double.MIN_VALUE` correctly, i.e. when you have an initial value of `0` and then click at the down arrow, `0` is submitted instead of `-1`. It needs some deeper refactoring. Feel free to revert the changes.

tandraschko:
Yep, currently got the same result.
Let's do it for 6.3, so we have more time for testing.

PR #3327, opened 2018-02-12
Use offsetHeight and offsetWidth instead of jQuery .height() and .width() functions to determine panel dimensions

koan00:
As requested by @tandraschko in the comments of PR #3287 - investigate other places where offsetWidth and offsetHeight should be used instead of jQuery .height() and .width() functions.

According to jQuery 3.0 migration notes ( https://jquery.com/upgrade-guide/3.0/#breaking-change-width-height-css-quot-width-quot-and-css-quot-height-quot-can-return-non-integer-values ) the change to width() and height() functions may have introduced issues in calculations expecting the old pre 3.0 functionality. However it appears the alternate method to calculate dimensions changes more than just float vs integer return type.

I did not attempt to change all calls to height() and width() in the PF application, as that would have been well beyond my means to test. I did focus on the areas where the code was identical to the code fixed in PR #3287. In most cases, the variance was not as great, but there was still a meaningful variance in dimensions.


koan00:
Since this is taking a while for approval, moving this commit to another branch.

PR #3333, opened 2018-02-14
fixes #3330 - empty values have to be accepted

### LINKED ISSUE #3330: InputNumber: 6.2 RC2 exception when decode invoked on empty submitted value
## 1) Environment
- PrimeFaces version: 6.2 RC2
- Does it work on the newest released PrimeFaces version? Version? yes, 6.1.13
- Application server + version: any
- Affected browsers: any

## 2) Expected behavior
p:inputNumber processes empty submitted value as usual. 

## 3) Actual behavior
NumberFormatException is thrown.
```
java.lang.NumberFormatException: null
	at java.math.BigDecimal.<init>(BigDecimal.java:596)
	at java.math.BigDecimal.<init>(BigDecimal.java:383)
	at java.math.BigDecimal.<init>(BigDecimal.java:806)
	at org.primefaces.component.inputnumber.InputNumberRenderer.decode(InputNumberRenderer.java:72)
	at javax.faces.component.UIComponentBase.decode(UIComponentBase.java:832)
```

## 4) Steps to reproduce
Open showcase page for inputNumber, leave "Empty value (sign)" empty and click new "crash" button.

## 5) Sample XHTML
Add button to inputNumber.xhtml page:

`<p:commandButton process="@form" update="@form" value="crash" />`

Bug was introduced in issue https://github.com/primefaces/primefaces/issues/3239

cnsgithub:
- take care: when value is bound to primitive and `""` gets submitted, model could be updated with `0.0 ∉ [minValue, maxValue]`, workaround is to declare `required="true"` when using with primitives or not at all using primitives

### DIFF
--- a/src/main/java/org/primefaces/component/inputnumber/InputNumberRenderer.java
+++ b/src/main/java/org/primefaces/component/inputnumber/InputNumberRenderer.java
@@ -72,13 +72,15 @@ public class InputNumberRenderer extends InputRenderer {
         try {
             if (ComponentUtils.isValueBlank(submittedValue)) {
                 ValueExpression valueExpression = inputNumber.getValueExpression("value");
-                Class<?> type = valueExpression.getType(context.getELContext());
-                if (type != null && type.isPrimitive()) {
-                    // avoid coercion of null or empty string to 0 which may be out of [minValue, maxValue] range
-                    submittedValue = String.valueOf(new BigDecimal(inputNumber.getMinValue()).doubleValue());
-                }
-                else {
-                    submittedValue = "";
+                if (valueExpression != null) {
+	                Class<?> type = valueExpression.getType(context.getELContext());
+	                if (type != null && type.isPrimitive() && !ComponentUtils.isValueBlank(inputNumber.getMinValue())) {
+		                // avoid coercion of null or empty string to 0 which may be out of [minValue, maxValue] range
+		                submittedValue = String.valueOf(new BigDecimal(inputNumber.getMinValue()).doubleValue());
+	                }
+	                else {
+		                submittedValue = "";
+	                }
                 }
             }
             else {
--- a/src/main/java/org/primefaces/component/inputnumber/InputNumberRenderer.java
+++ b/src/main/java/org/primefaces/component/inputnumber/InputNumberRenderer.java
@@ -73,14 +73,14 @@ public class InputNumberRenderer extends InputRenderer {
             if (ComponentUtils.isValueBlank(submittedValue)) {
                 ValueExpression valueExpression = inputNumber.getValueExpression("value");
                 if (valueExpression != null) {
-	                Class<?> type = valueExpression.getType(context.getELContext());
-	                if (type != null && type.isPrimitive() && !ComponentUtils.isValueBlank(inputNumber.getMinValue())) {
-		                // avoid coercion of null or empty string to 0 which may be out of [minValue, maxValue] range
-		                submittedValue = String.valueOf(new BigDecimal(inputNumber.getMinValue()).doubleValue());
-	                }
-	                else {
-		                submittedValue = "";
-	                }
+                    Class<?> type = valueExpression.getType(context.getELContext());
+                    if (type != null && type.isPrimitive() && !ComponentUtils.isValueBlank(inputNumber.getMinValue())) {
+                        // avoid coercion of null or empty string to 0 which may be out of [minValue, maxValue] range
+                        submittedValue = String.valueOf(new BigDecimal(inputNumber.getMinValue()).doubleValue());
+                    }
+                    else {
+                        submittedValue = "";
+                    }
                 }
             }
             else {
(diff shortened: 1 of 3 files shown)


cnsgithub:
- think the above mentioned is owed to coercion of empty strings to `0` in the EL implementation (I am currently using `org.glassfish:javax.el:3.0.1-b08-jbossorg-1` shipped with WildFly 11)

cnsgithub:
- the behavior of the EL implementation seems to be conform with the specification https://github.com/javaee/el-spec/blob/master/spec/pdf/EL3.0.latest.pdf; in section 1.23.3 it reads `If A is null or "", return 0`.
- so how can we fix this issue in PrimeFaces?

tandraschko:
Whats the exact problem with coercion in this case?

cnsgithub:
It's just an edge case: Suppose you have primitive value binding (`double ` instead of `Double`) and `minValue=1`. Then the client submits an empty string which gets converted by JSF to `null`. This `null `value gets in turn coerced by EL to `0` and the model is updated accordingly resulting in effectively bypassing the range checks since `0 < minValue`.

I'm just about to improve the checks.

cnsgithub:
With the last commit `minValue` will be applied to primitives instead of `0`. This should be double-checked in other components as well as it may break security constraints.

Shall we propose to change the EL spec at https://github.com/javaee/el-spec/?

src/main/java/org/primefaces/component/inputnumber/InputNumberRenderer.java
```diff
@@ -67,25 +69,39 @@ public void decode(FacesContext context, UIComponent component) {
...
 
-        if (submittedValue != null) {
-
-            BigDecimal value = new BigDecimal(submittedValue);
-            if (!ComponentUtils.isValueBlank(inputNumber.getMinValue())) {
-                BigDecimal min = new BigDecimal(inputNumber.getMinValue());
-                if (value.compareTo(min) < 0) {
-                    submittedValue = String.valueOf(min.doubleValue());
+        try {
+            if (submittedValue == null || submittedValue.trim().isEmpty()) {
```
Rapster: Use `ComponentUtils.isValueBlank()` instead

tandraschko:
i think the "value" expression must not be available for every (weird) case, so a NPE will occur.

tandraschko:
As you see, fixing something small isn't always easy as there are several weird cases.
Next time we should only apply hotfixes after RC1.

cnsgithub:
@tandraschko Please show me the use case for `p:inputNumber` without `value ` attribute. You should also double-check your code for similar locations. 

However, as `UIComponent:getValueExpression` says `Return the ValueExpression used to calculate the value for the specified attribute or property name, if any` you're right, there might be cases in theory where `ValueExpression ` is missing. I'll consider it.

And in my opinion security-related bugs are always 'hot' and I would never release a version without fixing all of them.

src/main/java/org/primefaces/component/inputnumber/InputNumberRenderer.java
```diff
@@ -67,25 +69,41 @@ public void decode(FacesContext context, UIComponent component) {
...
+                    if (value.compareTo(max) > 0) {
+                        submittedValue = String.valueOf(max.doubleValue());
+                    }
                 }
             }
-            
-            inputNumber.setSubmittedValue(submittedValue);
         }
-
+        catch (NumberFormatException ex) {
```
Rapster: Just by curiosity, after all these input tests. Is `NumberFormatException` is still meant to happen?

cnsgithub: Yes, NumberFormatException may still happen if client illegally submits non-empty string (e.g. 'crash'). I just thought that wrapping this exception in a FacesException would be more compliant to the PrimeFaces rules, if any.

cnsgithub:
Just provided an unit test covering 100% of the decode method. Changing easymock to the latest version is okay? I didn't find any usages of easymock anyway. It would be nice if PrimeFaces could provide more tests for complex things in the future.

tandraschko:
Cool, keep up the good work!
Upgrading easymock is ok. Are you also fimilar with mockito? I only used mockito in the past.

cnsgithub:
No, I am not familar with mockito, but it shouldn't matter too much. 

If we can save some boilerplate code (e.g. `InputNumberTest.setup`) migrating to mockito would be welcome.

tandraschko:
Lets talk about the mocking framework for 6.3. I preffered mockito in the past, maybe easymock is now very similar.

melloware:
+1 for Mockito.

I have used many mocking libraries over the years and Mockito is superior in every way to every other mocking library I have used.  Mockito was originally based on ideas from EasyMock but took them all to the next level.  here is the author's post about how he was an EasyMock user before he wrote Mockito...

https://monkeyisland.pl/2008/01/14/mockito/

"EasyMock is a decent piece of software, credits go to kick-ass people who wrote it. Initial hacks on Mockito were done on top of the EasyMock code but since then I rewrote most of the codebase."

Issue #3277, opened 2018-02-04
SelectOneMenu: Issue with dynamic="true" and required validation.


koan00:
## 1) Environment
- PrimeFaces version: master (5099d890cca46db839045d1b3815a66de6902e3c)
- Does it work on the newest released PrimeFaces version? No, dynamic="true" is a new feature.
- Does it work on the newest sources in GitHub? No.
- Application server + version: Tomcat 7.0.82 / Mojarra 2.2.15
- Affected browsers: IE 11, FF 58, Chrome 63

## 2) Expected behavior
Upon selecting a value in the menu, the value remains set and 'required' validation passes.

## 3) Actual behavior
Upon selecting a value in the menu, the value gets unset and 'required' validation fails

## 4) Steps to reproduce
From top to bottom, select values for all 3 drop downs. Then submit the form. 

Slightly different behavior between the top 2 items. The first item also has a nested <p:ajax> while the second does not.  Finally, last one does not have the dynamic="true" flag set.

## 5) Sample XHTML

`<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml"
      xmlns:ui="http://java.sun.com/jsf/facelets"
      xmlns:f="http://java.sun.com/jsf/core"
      xmlns:p="http://primefaces.org/ui"
      xmlns:h="http://java.sun.com/jsf/html">

    <h:head>   
        <title>PrimeFaces Test</title>
		<f:facet name="first">
			<meta http-equiv="X-UA-Compatible" content="IE=edge"/>
			<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
		</f:facet>
	</h:head>
    <h:body>
			<p:log id="testLogger" />
		<h:form id="dummy"></h:form>
		<form>test regular form</form>
        <h1>#{testView.testString}</h1>
		<p:accordionPanel>
			<p:tab title="test2" rendered="false">
				dummy content
			</p:tab>
		<p:tab title="test">
			<h:form id="testForm">

				<p:panelGrid>
				<p:row>
					<p:column>List1:</p:column>
					<p:column>
						<p:selectOneMenu autoWidth="false" id="list1" effect="none" value="#{testView.itemValue1}" dynamic="true"
										filter="true" filterMatchMode="contains" required="true">
								<f:selectItem itemLabel="– Choose Value 1 –" itemValue="#{null}"/>
								<f:selectItem itemLabel="Item1" itemValue="1"/>
[...]

PR #3339, opened 2018-02-14
Fix for PF issue #3277

koan00:
Fix for PF issue #3277


tandraschko:
I'm not sure about this fix. The problem seems that the select value isn't even submitted to the server.
But will check your PR now.

tandraschko:
Thanks for the PR but fixed it a little bit nicer with jQuery.

PR #3365, opened 2018-02-23
Fix #2861 - Tree: filter on a property of node values

### LINKED ISSUE #2861: Tree: filter on a property of node values
## 1) Environment
- PrimeFaces version: 6.1.7
- Application server + version: Tomcat 8.5.20
- Affected browsers: Firefox, Chrome, Edge

## 2) Expected behavior
When building a `<p:tree>` of nodes whose data is a bean, not a `String`, I expect to be able to apply filtering, say *startsWith*, on a property of these beans, as documented in the user manual:
```xml
<p:tree value="#{treeBean.root}" var="node" filterBy="#{node.name}">
    <p:treeNode>
        <h:outputText value="#{node}"/>
    </p:treeNode>
</p:tree>
```

## 3) Actual behavior
Filtering only works as expected when node values are `String`. When using a bean property, it returns no result.

## 4) Steps to reproduce
See example.

## 5) Sample XHTML
```xml
<p:tree value="#{treeBean.root}" var="node" filterBy="#{node.name}">
    <p:treeNode>
        <h:outputText value="#{node.name}"/>
    </p:treeNode>
</p:tree>
```

## 6) Sample bean
```java
public class TreeBean {

    private TreeNode root;

    public TreeBean() {
        this.root = new DefaultTreeNode();
        this.root.getChildren().add(new DefaultTreeNode(new Fruit("Lemon")));
        this.root.getChildren().add(new DefaultTreeNode(new Fruit("Apple")));
        this.root.getChildren().add(new DefaultTreeNode(new Fruit("Cherry")));
        this.root.getChildren().add(new DefaultTreeNode(new Fruit("Banana")));
    }

    public TreeNode getRoot() {
        return root;
    }

    public class Fruit {

        private String name;

        public Fruit(final String name) {
            this.name = name;
        }

        public String getName() {
            return name;
        }

    }
}
```

## 7) Analysis
The expression passed in `filterBy` attribute seems to be ignored, or more precisely, it seems to be only used to trigger filtering or not, whatever the value of the expression is.
Then `org.primefaces.model.filter.StartsWithFilterConstraint.applies()` does an explicit `toString()` conversion of the node value before applying the filter (line 34):

```java
return value.toString().toLowerCase(locale).startsWith(filterText);
```
This is true for several implementations of `FilterConstraint`.

And indeed, if I add a custom `toString()` method to my `Fruit` class, it works as expected:
```java
        @Override
        public String toString() {
            return this.name;
        }
```

Rapster:


Rapster:
I'm not sure why it shows all these whitespace changes while I didn't do anything about it... _(Add ?w=1 at the url to see the real changes)_

tandraschko:
Please don't mix bugfixes and refactoring (like removing useless casts and else...) in one PR.

Rapster:
I know Thomas, the commit "Refactor" should have been part of the fix itself actually. Regarding the rest, they're only warning thrown by SONAR (easy and safe to do)

tandraschko:
I don't have a problem with that refactoring but with them, i don't like to merge them before 6.2.
If 6.3 is ok for you, everything with the PR is fine

Rapster:
Glad to hear that! I understand your point of view, but I don't see side effects of this PR. `filterBy` is a brand new feature for 6.2 and  it would be a pity to deliver it half working

tandraschko:
take e.g. the if statement about the icon. It's not the same as before.
Therfore, if you would like to to get a fix in, i need a PR without any cosmetics or refactorings. We will release a final, not a RC anymore.

Rapster:
For better readability, I change the statement. But it's definitely the same. I'll provide a PR only with a fix then

tandraschko:
AFAICS it's not the same:
if expanded and RTL -> a collapsed icon is now rendered

Rapster:
Ah indeed, good catch! :wink: I'll change the PR next week, with only the fix and make another PR for SONAR

Rapster:
PR updated :wink:

PR #3370, opened 2018-02-27
Fixes #3364 lazydatamodel iterator

### LINKED ISSUE #3364: Implement Iterator for LazyDataModel
## 1) Environment
- PrimeFaces version: 6.1
- Application server + version: jetty-8.1.9.v20130131
- Affected browsers: All browsers

## 2) Expected behavior

For each over dataModel ends loop.

## 3) Actual behavior

For each over dataModel causes endless loop.

## 4) Steps to reproduce

Create a lazy datatable with a datamodel, then try to do a loop over it.
In the following sample, just click in a row to trigger onRowSelect method.

## 5) Sample XHTML

Just use the showcase DataTable - Lazy:
http://localhost:8080/showcase/ui/data/datatable/lazy.xhtml

## 6) Sample bean

Using showcase project, in class LazyView, add the following code to onRowSelect method:

```java
for (Car car : this.lazyModel) {
	System.out.println(car.getId());
}
```

cnsgithub:


tandraschko:
Nice contribution - looks good! Will merge it after the final release.

melloware (approved):

melloware:
@tandraschko I didn't make the comment on the review but a general question.  I have not seen many other classes use java.util.* in imports I didn't know if the project had code guidelines or not to explicitly call out each import or allowed to use *?

tandraschko:
Currently we completely avoid * imports.
IMO a good rule is: if there are more then X imports from the same package, use *. I usually use 7 as X in my projects at $work.

tandraschko:
Is there a checkstyle rule for something? If yes, i think we should also set it 5 or 7. WDYT?

tandraschko:
will remove the * imports for now. Thanks

PR #3455, opened 2018-03-08
fixes #3453 - bind defaultCommand to keypress instead of keydown to prevent wrong value to be submitted when using with firefox autocompletion

### LINKED ISSUE #3453: defaultCommand: with browser autocomplete not working together in Firefox
## 1) Environment
- PrimeFaces version: 6.*

## 2) Expected behavior

defaultcommand-chrome

## 3) Actual behavior

defaultcommand-firefox

## 4) Steps to reproduce

2. Clear the field. Type n

## 5) Sample XHTML & bean

## 6) Related issues

cnsgithub:


melloware (approved):

tandraschko:
Could it break other things?

david0:
@cnsgithub: I noticed that with this change pressing `l` will now trigger the default command. I think the problem is that `e.which` seems to return the charCode (which is 108 for l) here.
After changing this from `e.which` to `e.keyCode` everything works like expected.

cnsgithub:
@david0 Not sure about this. Did you test other things as well? Then I would apply your proposed patch to this branch.

david0:
@cnsgithub I did some testing with other browsers and found out that changing `which` to `keyCode` solves it only for Firefox ([detailed results](https://gist.github.com/david0/36a6b3cf0d07dd1fe1296fcfbe8527ba#file-keypress_vs_keydown-md)). I tested with Firefox, Chrome and Internet Explorer, I never saw that numpad enter was reported with code 108. 
Hence I would suggest to just drop `e.which == keyCode.NUMPAD_ENTER`.

cnsgithub:
I cannot just drop it without knowing any side effects. These NUMPAD_ENTER issues are very weird to me. 

Here is another example where NUMPAD_ENTER seems to be involved: https://github.com/primefaces/primefaces/issues/3615 

Therefore, I'll close this pull request unmerged, sorry...

david0:
The real question is why NUMPAD_ENTER was introduced in the first place. Maybe @cagataycivici can help, since it looks like he has written that initially?

Dropping NUMPAD_ENTER here should also fix #3615. I would see that bug as another reason to drop numpad enter here. 

Maybe there are some rare edge cases with old browsers or a very special keyboard driver, but for me it seems like would fix at least two common use cases and maybe break one very special case which -once we know these case- could be handled more specifically.

jxmai:
Just my two cents. After looking at the comment from lagar84 https://github.com/primefaces/primefaces/issues/3615#issuecomment-384742776, we can realize that **`NUMPAD_ENTER` is not reliable and should be removed / refactored**

https://github.com/primefaces/primefaces/blob/608e00ddd7c4df3df66f1a87d5eb001f2414560b/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.js#L140-L143


I might be wrong but why differentiate num_pad_enter and enter in the first place since they are both sharing with the keycode 13 ?

For example,

https://www.cambiaresearch.com/articles/15/javascript-char-codes-key-codes

there is simply no `num_pad_enter`


I also found a gist somewhere, keycode 108 does not mean `num_pad_enter` at all (take it with a grain of salt...)

https://gist.github.com/lbj96347/2567917#file-jskeycode-txt-L74

```
keycode 108 = KP_Separator KP_Separator
```

Looks like keycode 108 is keypad separator, I cannot think of anything it can be related to keypad enter.


A totally different note on emacs manual, there is another name called `kp-enter`, which does not exist in browser

https://www.gnu.org/software/emacs/manual/html_node/emacs/Function-Keys.html

snapshot1


snapshot1

cnsgithub:
@tandraschko Let's reopen this PR and remove NUMPAD_ENTER as suggested by @jxmai and @david0 ?

jxmai:
I agree with @cnsgithub. The 108 keycode (`KP_Separator` aka `Keypad Separator`) should mean the `,` key as mentioned by https://github.com/primefaces/primefaces/issues/3615#issuecomment-384716172

melloware:
This is probably why Jquery originally removed the NUMPAD_ENTER constant between Jquery 1.10 and 1.11.

melloware:
Can you guys review my change I made this fix along with removing NUMPAD_ENTER all in one PR.

https://github.com/primefaces/primefaces/pull/3844

PR #3478, opened 2018-03-12
Fix #3457: Layout script output in incorrect place.

### LINKED ISSUE #3457: Layout: Missing Center-pane Element Error (PF 6.1 to PF 6.2)
## 1) Environment
- PrimeFaces version: 6.2
- Does it work on the newest sources in GitHub? (Build by source ->  https://github.com/primefaces/primefaces/wiki/Building-From-Source) No
- Application server + version: GlassFish Server 4.1.1
- Affected browsers: ALL

## Expected behavior and Actual behavior
For p:commandbuttons with ajax=false or h:commandlinks, a refresh for full page is to take over. When we only upgraded from Primefaces version 6.1 to version 6.2, now clicking on these buttons causes the following error:
**/UI Layout Initialization Error
The center-pane element does not exist.
The center-pane is a required element.
We need to mention that when the commandbutton is clicked, we have a growl type of message to be updated.**
                <p:messages severity="error" showDetail="true" escape="false" closable="true"  style="width:800px; padding-left: 280px"><p:autoUpdate /></p:messages>
                <p:growl id="growl" severity="info, warn" showDetail="true" globalOnly="true" escape="false" sticky="true"><p:autoUpdate /></p:growl> 

We are not sure of the reason of this error except that when we use version 6.1, the Error does not show and all functionalities are fine.

**Could that have to do with an updating settings of certain elements in Primefaces? Please give us any hint.**

.. 

Sample XHTML:
Example 1:
                        <p:commandButton value="Submit" ajax="false"
                                         style ="margin-left: 20px"
                                         onclick="delayShow()"
                                         oncomplete="PF('statusDialog').hide()"
                                         actionListener="#{yyyBean.handleFileUpload()}"/>  

Example 2:
                    <p:commandButton value="Update" action="#{xxxBean.updateSummary()}" 
                                     onclick="PF('statusDialog').show()" ajax="false"/>

Example 3:
                        <h:commandLink style="font-size: 13px; font-weight: bold" action="#{zzzBean.doAnalysis('option_1')}" 
                                       onclick="PF('statusDialog').show()"
                                       value="Some value!"/>

melloware:
Wow this was tricky.  On ajax="false" update the script was being output before the DOM was built and this the Center Pane was missing.  I have no idea how this wasn't broken before!

I simply moved encodeScript() method to the encodeEnd() instead of encodeBegin() and it fixes it.  Much thanks to @GedMarc!

### DIFF
--- a/src/main/java/org/primefaces/component/layout/LayoutRenderer.java
+++ b/src/main/java/org/primefaces/component/layout/LayoutRenderer.java
@@ -38,8 +38,6 @@ public class LayoutRenderer extends CoreRenderer {
         Layout layout = (Layout) component;
         String clientId = layout.getClientId(context);
 
-        encodeScript(context, layout);
-
         if (layout.isElementLayout()) {
             writer.startElement("div", layout);
             writer.writeAttribute("id", clientId, "id");
@@ -57,6 +55,8 @@ public class LayoutRenderer extends CoreRenderer {
         if (layout.isElementLayout()) {
             writer.endElement("div");
         }
+        
+        encodeScript(context, layout);
     }
 
     protected void encodeScript(FacesContext context, Layout layout) throws IOException {


tandraschko:
:D
it was MAYBE broken before because the scripts were executed with $(), also on ajax

PR #3483, opened 2018-03-13
Added missing descriptions

### LINKED ISSUE #3480: Missing descriptions in the taglib
Not really an issue, but almost all of the components in the file [standard-facelets-taglib.xml](https://github.com/primefaces/primefaces/blob/1a084be2a7ef33e9c4c38fa16b48e5e022d8500b/src/main/resources-maven-jsf/standard-facelets-taglib.xml) do not have the description attribute.

NicolaIsotta:
Fixes #3480 
I've taken descriptions from the manual.

### DIFF
--- a/src/main/resources-maven-jsf/standard-facelets-taglib.xml
+++ b/src/main/resources-maven-jsf/standard-facelets-taglib.xml
@@ -70,6 +70,7 @@
 
 <tag>
     <tag-name>autoUpdate</tag-name>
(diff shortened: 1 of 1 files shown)


tandraschko:
2 small notes:
1) AutoUpdate: i would add that it can be skipped with "ignoreAutoUpdate" on p:ajax or buttons etc.
2) `p:repeat:` its actually a hacky workaround for a old mojarra implementation detail / bug when used in combination with iteratring PF components. MyFaces should be fine.

NicolaIsotta:
PR updated.

PR #3503, opened 2018-03-20
Fix #2690 - Splitbutton: display as button in case no children rendered

### LINKED ISSUE #2690: Splitbutton: display as button in case no children rendered
If there is not child items defined in SplitButton, don't display arrow (e.g should look like a button then). It makes think there is something wrong...

Rapster:


Rapster:
`ComponentUtils.shouldRenderChildren()` is very usefull and could be added to the JSF functions. For example, I have this `MenuButton` containing a lot of items but all are not rendered, in such case I don't want to display it then. This method would help me to achieve that.

Issue #3520, opened 2018-03-23
XSS despite using ComponentUtils.escapeText


cnsgithub:
## 1) Environment
- PrimeFaces version: 6.3-snapshot
- Does it work on the newest released PrimeFaces version? Version? no
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) no
- Application server + version: n/a
- Affected browsers: n/a

## 2) Expected behavior
Developers can rely on `ComponentUtils.escapeText` to properly escape JS.

## 3) Actual behavior
Unfortunately, `ComponentUtils.escapeText `is just for escaping JSON values as already denoted here two years ago: https://github.com/primefaces/primefaces/issues/1194#issuecomment-192155762

Escaping for JSON and JS contexts is similar but obviously not totally identical. Whereas in JS single quotes may also be used for strings, JSON strings must use double quotes, see https://www.json.org/. 

Currently, the big problem is that `escapeText `is used and spread all over the PrimeFaces java sources. And that developers rely on this functioning as a fully secure JS escaper. However, that's wrong.

## 4) Steps to reproduce
This is just one example, there are probably various locations that are vulnerable, too.

Include and click a button in the showcase http://localhost:8080/showcase/ui/button/button.xhtml like shown in 5). 

There will be a console log despite escaping in `ButtonRenderer` (which I previously tried to fix in https://github.com/primefaces/primefaces/issues/3213 and now turned out to be insufficient).

Another susceptible location might be in `CoreRenderer.buildNonAjaxRequest`.

## 5) Sample XHTML
`        <p:button href="');console.log('XSS');window.open('" />`

## 6) Sample bean
n/a

## 7) Mitigation
- Short term: Identify all vulnerable locations, i.e. locations that construct JS using single quotes and make use of `ComponentUtils.escapeText`. In those locations replace single by double quotes. And/or include escaping of single quotes in `ComponentUtils.escapeText` (this might however break valid JSON).
- Long term: Replace all `escape*` methods with appropriate variants from https://github.com/OWASP/owasp-java-encoder. Sensitize developers to proper escaping. Once again: different contexts need different encodings.

melloware:
@cnsgithub I like where you are going with this but here is what I propose and maybe @tandraschko or @mertsincan have different ideas

[...]

PR #3525, opened 2018-03-26
[OWASP] enhance encoding (#3520)

cnsgithub:
Step 1: Copied encoders from owasp-java-encoder

Some notes:
- decided to leave the package structure of owasp-java-encoder untouched, so we can better differentiate between primefaces and third-party sources; also this should make it easier to maintain (btw: should we also move `BoundedInputStream `from commons.io to its original package `org.apache.commons.io`?) -> added `package-info.java` to document owasp-java-encoder's version
- added the newly added classes to `checkstyle-suppressions.xml` to keep them comparable to https://github.com/OWASP/owasp-java-encoder
- added the newly added classes to license-maven-plugin's exclusion list to keep the original BSD 3.0 license headers (required as per the licensing model)

Next step will be the introduction of `EscapeUtils `proposed by @melloware in https://github.com/primefaces/primefaces/issues/3520#issuecomment-375633571 that just delegates calls to `org.owasp.encoder.Encode`.


cnsgithub:
There are no more calls to `ComponentUtils.escapeText/escapeEcmaScriptText/escapeXml`. They are all `@Deprecated`. Normally, we could remove them also.

I did some rudimentary testing on the showcase. Primarily the ajax part was of interest for me since it was necessary to adjust `PrimePartialResponseWriterTest` accordingly, since it was a bit lenient in XML escaping. At first glance everything is working as expected. Developer console did not show any errors as well.

What I was not sure about is `XmlUtils.escapeTag `which does some special escaping by replacing illegal values with specific separators. To not break compatibility with data export consumers, I didn't touch it yet.

cnsgithub:
`p:button href` mentioned in https://github.com/primefaces/primefaces/issues/3520#issue-307963031 is no longer vulnerable to XSS.

cnsgithub:
Reviews are very welcome.

src/main/java/org/owasp/encoder/ASCIIBits.java
```diff
@@ -0,0 +1,124 @@
...
+// INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+// (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+// HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+// OF THE POSSIBILITY OF SUCH DAMAGE.
+
+package org.owasp.encoder;
+
```
melloware: @tandraschko I am wondering if we should rename the package to org.primefaces.encoder only to not conflict Classloader if someone does include the OWASP JAR itself.  The copyright and all text stays so people will know where this code came from, I just worry about Classloader clollisions. Your thoughts?

checkstyle-suppressions.xml
```diff
@@ -7,4 +7,5 @@
 <suppressions>
     <suppress files="[\\/]generated-sources[\\/]" checks="[a-zA-Z0-9]*"/>
     <suppress files="BoundedInputStream\.java"  checks="[a-zA-Z0-9]*"/>
+    <suppress files="org[\\/]owasp[\\/]encoder[\\/]" checks="[a-zA-Z0-9]*"/>
```
melloware: Change to org.primefaces.encoder?

tandraschko:
Could someone check the legal approach of copying the class into PF? With Apache2, it's no problem...

tandraschko:
It might hard to say: PF is Apache2 licensed but contains several classes form other sources with other licenses...

melloware:
OWASP is BSD3 licensed which is actually even more friendly than Apache 2.0.  We can basically do whatever we want with the code and it only implies these 3 restrictions:


    * Redistributions of source code must retain the above
      copyright notice, this list of conditions and the following
      disclaimer.
PASSED - we included it

    * Redistributions in binary form must reproduce the above
      copyright notice, this list of conditions and the following
      disclaimer in the documentation and/or other materials
      provided with the distribution.
I THINK WE JUST NEED TO INCLUDE IN OUR LICENSE WE ALSO INCLUDE BSD3.0 licensed software

    * Neither the name of the OWASP nor the names of its
      contributors may be used to endorse or promote products
      derived from this software without specific prior written
I THINK WE WILL PASS THIS ONCE WE REMOVE THE WORD "OWASP" FROM THE PACKAGE NAME.
      permission.

cnsgithub:
> I THINK WE WILL PASS THIS ONCE WE REMOVE THE WORD "OWASP" FROM THE PACKAGE NAME.

In my understanding we're fully conform to this clause because we do not promote PrimeFaces by saying 'powered by OWASP' or anything like that. 

Apache's legal FAQ (http://www.apache.org/legal/resolved.html#category-a) states:
> FOR THE PURPOSES OF BEING INCLUDED IN AN APACHE PRODUCT, WHICH LICENSES ARE CONSIDERED TO BE SIMILAR IN TERMS TO THE APACHE LICENSE 2.0?
> Works under the following licenses may be included within Apache products:
> BSD 3-clause

src/main/java/org/owasp/encoder/ASCIIBits.java
cnsgithub: Good point. But that may also happen with other dependencies not included in the primefaces JAR.

Just some thoughts: 
- Can JAR sealing help us here?
- Can we leverage a tool like [Apache Maven Shade](http://maven.apache.org/plugins/maven-shade-plugin/)?

melloware: I thought Maven Shade plugin just basically created an uberjar from your dependencies but it still just includes them normally so its no different from a classloader perspective if two jars both have org.owasp.encoder the classloader has to pick one.   I am not familiar with JAR sealing?

cnsgithub: I just rethought: JAR sealing would not help us here.

However, I'll commit to this PR branch an integration of the shade plugin, be prepared for the upcoming commit comment.

cnsgithub:
Regarding the maven shade plugin:
I already have a running version, albeit not yet commited due to a problem with missing license in owasp-java-encoder. I just filed an issue: https://github.com/OWASP/owasp-java-encoder/issues/12

This is how it basically works:
- no manual copying of owasp-java-encoder sources to PrimeFaces source needed
- instead add a regular compile time dependency to it in the pom (better maintainability)
- use maven-shade-plugin to include owasp-java-encoder classes as well as the license (currently missing) and relocate `org.owasp.encoder package `to shaded.`org.owasp.encoder` to avoid classloading problems like so:
```xml
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-shade-plugin</artifactId>
                <version>3.1.0</version>
                <executions>
                    <execution>
                        <phase>package</phase>
                        <goals>
                            <goal>shade</goal>
                        </goals>
                        <configuration>
                            <transformers>
                                <transformer implementation="org.apache.maven.plugins.shade.resource.IncludeResourceTransformer">
                                    <resource>META-INF/LICENSE_OWASP_ENCODER</resource>
                                    <file>META-INF/LICENSE</file>
                                </transformer>
                            </transformers>
                            <relocations>
                                <relocation>
                                    <pattern>org.owasp.encoder</pattern>
                                    <shadedPattern>shaded.org.owasp.encoder</shadedPattern>
                                </relocation>
                            </relocations>
                        </configuration>
                    </execution>
                </executions>
            </plugin>
```
- the JAR root will look like
image

melloware:
That is slick @cnsgithub That way we can just include the dependency and it will bring the code in and also protect the Classloaded since these will be called shaded.

cnsgithub:
Last commit introduced `maven-shade-plugin` as well as `license-maven-plugin` to include and relocate `owasp-java-encoder` files and to include third-party-licenses in primefaces JAR (also see https://github.com/OWASP/owasp-java-encoder/issues/12#issuecomment-378527696).

The JAR structure will look like this now:
```
|   THIRD-PARTY.txt
|   
+---META-INF
|   |   
|   \---resources
|       |   licenses.xml
|       |   
|       +---licenses
|       |       apache license v2 - license-2.0.html
|       |       bouncy castle licence - licence.html
|       |       bsd license - bsd-3-clause.txt.html
|       |       cddl + gplv2 with classpath exception - cddl+gpl.html
|       |       common development and distribution license (cddl) plus gpl - cddl+gpl.html
|       |       common development and distribution license (cddl) v1.0 - cddlv1.0.html
|       |       common development and distribution license (cddl) version 1.0 - cddl.html
|       |       mit license - mit-license.php.html
|       |       mozilla public license - mpl-1.1.html
|       |       similar to apache license but with the acknowledgment clause removed - license.txt
|       |       the apache software license, version 2.0 - license-2.0.txt
|       |       the bsd 3-clause license - bsd-3-clause.txt.html
|       |       
\---primefaces
    \---shaded
        \---org
            \---owasp
                \---encoder
                        ASCIIBits.class
                        CDATAEncoder.class
                        ChainedEncoder.class
                        CSSEncoder$Mode.class
                        CSSEncoder.class
                        Encode$Buffer.class
                        Encode.class
                        EncodedWriter.class
                        Encoder.class
                        Encoders.class
                        HTMLEncoder.class
                        JavaEncoder.class
                        JavaScriptEncoder$Mode.class
                        JavaScriptEncoder.class
                        Unicode.class
                        UnsupportedContextException.class
                        URIEncoder$Mode.class
                        URIEncoder.class
                        XMLCommentEncoder.class
                        XMLEncoder$Mode.class
                        XMLEncoder.class
```

E.g. `THIRD-PARTY.txt` contains this line:
`     (The BSD 3-Clause License) Java Encoder (org.owasp.encoder:encoder:1.2.1 - https://www.owasp.org/index.php/OWASP_Java_Encoder_Project/encoder)`

The original license texts are copied to `META-INF/resources/licenses`, `licenses.xml` contains a summary.

@tandraschko @melloware What do you think? Should we give this PR a try?

tandraschko:
Thats not only up to me, i need to discuss this before.
And currently i don't have much time, i will priorise the older PRs before.

cnsgithub:
@tandraschko It's two months later now... Are there any plans to merge it?

tandraschko:
No Time currenty, sry. Need also discuss thus First.

tandraschko:
@cnsgithub really would like to discuss this with the PrimeTeam, but even all of my private discussion-topics are not all done till today.
I promise i will take care of it before 6.3 but WDYT about doing it in 3 steps?
1) Introduce EscapeUtils, move all of our existing methods there
2) Correct all usages to use the right method there
3) switch to OWASP in the EscapeUtils

It would be much better to follow all the changes.

cnsgithub:
@tandraschko Thanks for the feedback. I think the problem of point 2 is, that there isn't an adequate encoding implementation for each context right now. So it makes no big difference if we just stay with the old method names or refactor existing code to use newly introduced ones. All calls to the OWASP encoders are already wrapped in `EscapeUtils`. I'm not interested in repeatedly refactoring code that has been proposed for three months now (it has conflicts again now). I also do not have enough time to do so. I think PrimeTek should be interested in merging things like that in own initiative.

tandraschko:
Triggered the discussion with PrimeTek now.

Just 2 things before i could merge after the discussion:
1) please don't change our imports, it's probably your IDE setting about grouping imports
2) fix the conflicts

cnsgithub:
Resolved the conflicts.

pom.xml
```diff
@@ -192,6 +192,14 @@
             <version>2.7.4</version>
             <scope>provided</scope>
         </dependency>
+
+        <!-- will be included and relocated by shade plugin -->
+        <dependency>
+            <groupId>org.owasp.encoder</groupId>
+            <artifactId>encoder</artifactId>
+            <version>1.2.1</version>
+            <scope>compile</scope>
```
Rapster: Using `compile` does not make that dependency transitive? Or shade will remove this dependency from the pom.ml

tandraschko:
Conflicts now - can you resolve them? (I'm still waiting for feedback from PrimeTek)

tandraschko:
Got the GO ;)
If you can fix the conflicts, i can merge it early next before holiday.

About the transitive dependency, it would should be fine if we set "<optional>true</optional>"

Rapster:
What is the policy about dependencies? IMO, it makes sense to use shade if we use a few classes over plenty. In this case, we use all classes (according to https://github.com/primefaces/primefaces/pull/3525#issuecomment-378541688), why not using it as a true dependency then? In general, I would use shading whenever copying a class from another repo is needed.

@cnsgithub you asked
> (btw: should we also move `BoundedInputStream` from commons.io to its original package org.apache.commons.io?)

Yes it makes sense, but if we use last version of commons-io (which is 2.5) it will copy a lot more than just `BoundedInputStream` because it has a "dependency" to `IOUtils` (see https://github.com/apache/commons-io/blob/master/src/main/java/org/apache/commons/io/input/BoundedInputStream.java) It'll copy a bunch of classes just because `BoundedInputStream` needs `IOUtils#EOF`. Although, it should be fine using `commons-io` 2.4

When using `maven-shade-plugin` we just have to make sure, shade dependencies are no longer transitive (removed from the final pom.xml)

Another thing, Thomas said

> please don't change our imports, it's probably your IDE setting about grouping imports

So far, imports are sorted in this order:

- java packages
- javax packages
- others (like org.primefaces)

tandraschko:
Cagatay only allowed me to shade it - because it's used by many core features and PF goal is 1jar = everything cool :D

-> we can use OWASP if we shade it, everything else needs to be untouched (like the BoundedInputStream or json).

melloware:
Personally I prefer transitive dependencies over shaded ones and here is why.  If for some security reason a user wants to override the version of OWASP or (JSON future) they can simply doing by updating their pom.xml and excluding the versions included with PF.  If however you wrap it up inside the PF JAR there is no way to override anymore without ripping the jar open and removing classes.

Rapster:
@melloware Yes, but it has also dowsides:

- for PF, what if the upgrading version of the dependency (overrided by the end user) is not compatible with PF (projects using guava suffered a lot about this), 

- or the project itself , what if PF is actually overriding the dependency of the project, this an create regressions.

melloware:
Agreed it definitely has its pro's and cons.

melloware:
OK that settles it we shade it :)

Rapster:
@cnsgithub you should use `minimizeJar` option, it will copy only what you need :wink:

Rapster:
> -> we can use OWASP if we shade it, everything else needs to be untouched (like the BoundedInputStream or json).

Why the approach should be different?

Rapster (changes requested):
What is been shaded does not need to be included in checkstyle or license plugin, or it does?

tandraschko:
AFAIK checkstyle checks the source and shading uses the compiled classes? so nn to exclude something actually?!

Rapster:
Yes no need, as an example, I had a try shading json (which does not follow our checkstyle) and had no issues with checkstyle plugin after excluding

cnsgithub:
@Rapster I removed the checkstyle and license exclusions as you suggested.

This PR was quite a time ago and it wasn't that easy to merge, so hopefully I did not miss something. Sorry, not much time currently...

tandraschko:
Looks like you readded getHrefURL getResourceURL isValueBlank to ComponentUtils - please check this changes

cnsgithub:
Removed re-added methods.

cnsgithub:
There are some failing tests in `NativeUploadedFileTest`. Those are however not related to this PR.

Rapster:
@mertsincan Can you have a look please? The problem is still there

Rapster:
I wish few of my feedbacks would be answered before merging...

tandraschko:
the codechanges are fine, just pom needs a final tuning now because of the license
i removed as the it broke the build on my machine

tandraschko:
@cnsgithub thanks for the changes!
Can you also migrate all code to EscapeUtils?

cnsgithub:
That should have been done already, except for new code, if any. We could see that very easily by just dropping the deprecated escape methods in `ComponentUtils`.

Rapster:
I left a feedback regarding shaded naming policy, because it's gonna happen again with other library. I thought of `org.primefaces.owasp.encoder` just like we have `org.primefaces.json`, but instead it's `primefaces.shaded.org.owasp.encoder` @melloware left a comment I didn't understand, so I'd like to understand it to make sure I don't miss anything.

cnsgithub:
I don't like the idea of mixing up the namespaces. However, feel free to use a prefix of your choice.

Rapster:
It's not a matter of personal choice or taste, the idea is to follow guidelines of what PrimeFaces have done so far. I consider json as a shaded dependency (even though it's not entirely true), so I would proceed in the very same way with owasp (this is why I proposed `org.primefaces.owasp.encoder`) . Unless we think `primefaces.shaded` is the good prefix to go with, I'm totally up for it as long we keep doing that with the next dependencies (e.g org.json) I just don't want to have some shaded stuff on one side and the other shaded stuff on the other side... 

But Thomas seems to not mind it, so I guess there is no need to continue this conversation. Next time, please, answer the feedback itself :wink:

cnsgithub:
I generally also prefer consensus over personal taste. But it has been a long time, almost half a year. And there was not so much feedback from you regarding this point at this time. And in fact I still have problems to find your review containing this shading namespace proposal. I think sometimes we should rather concentrate on the issues than discussing trivial things that can be changed without efforts in the future anyway. owasp was the first lib to be shaded, other libs may follow these rules or redefine them, not the other way around.

tandraschko:
i changed it to org.primefaces.shaded.owasp.encoder for now

we should use "org.primefaces." prefix everywere i think it makes sense to use a "shaded" package
we can move json for 7.0, too

melloware:
@Rapster my comment has been answered I think from reading all the changes I am good to go with all the fixes here!

Rapster:
@melloware I didn't find it, but I can figure out what you and cnsgithub meant :wink:

@tandraschko Cool! I'll create an issue so we don't forget about this for 7.0

melloware:
The only thing I see now in Maven after updating is this...

```
[INFO] --- license-maven-plugin:1.16:aggregate-download-licenses (aggregate-download-licenses) @ primefaces ---
[WARNING] Unable to retrieve license for dependency: org.owasp.esapi:esapi
[WARNING] http://creativecommons.org/licenses/by-sa/3.0/
[WARNING] Server returned HTTP response code: 403 for URL: http://creativecommons.org/licenses/by-sa/3.0/
```

PR #3531, opened 2018-03-26
use shields.io

cnsgithub:
maven-badges.herokuapp.com is often down.


melloware (approved):
THis is great I am fixing PFE too because its annoying Heroku is down all the time!

PR #3540, opened 2018-03-28
Fix #3487: Keyboard support unicode characters.

### LINKED ISSUE #3487: Keyboard: Unicode Character does not support at p:keyboard in attribute layoutTemplate
## Environment
	
* PrimeFaces version: 6.2
* Application server: Wildfly 10
* Affected browsers: ALL
	
## Expected behavior and Actual behavior
	
When we only upgraded from Primefaces version 6.1 to version 6.2, now clicking on these keyboard button (or) keyboard causes the following error:
	
'\\' 'u' '1' '0' '4' '1' ..... 'back' ..... '\\' 'u' '1' '0' '4' '4' ... 'clear'
each character set in one keyboard layout button
	
## Sample XHTML:

```xml
<p:keyboard id="myIdState"
	readonly="#{facesContext.currentPhaseId.ordinal eq 6}"
	style="width:40px;margin-left:-4px;" layout="custom"
	layoutTemplate="\u1041\u1042\u1043-back,\u1044\u1045\u1046-clear,\u1047\u1048\u1049-close,\u1040"
	value="#{UserManagementApp.idState}"
	maxlength="2">
```
but this code ok
	
```xml		
<p:keyboard id="myIdState"
	readonly="#{facesContext.currentPhaseId.ordinal eq 6}"
	style="width:40px;margin-left:-4px;" layout="custom"
	layoutTemplate="၁၂၃-back,၄၅၆-clear,၇၈၉-close,၀"
	value="#{UserManagementApp.idState}"
	maxlength="2">
```

When we use version 6.1, the Error does not show and all functionalities are fine.
	
**Could that have to do with updating settings of certain elements in Primefaces? Please give us any hint.**

melloware:
Found a trick to convert the code to unicode https://stackoverflow.com/questions/7885096/how-do-i-decode-a-string-with-escaped-unicode

This may have been a Jquery change that this somehow worked in PF 6.1 but not in 6.2.  Either way the test case now passes.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/keyboard/keyboard.js
+++ b/src/main/resources/META-INF/resources/primefaces/keyboard/keyboard.js
@@ -961,8 +961,8 @@ PrimeFaces.widget.KeyboardUtils = {
 	},
 
 	createLayoutFromTemplate : function(template) {
-	    // GitHub #3487: Unicode conversion
-	    template =  decodeURIComponent(JSON.parse('"' + template.replace(/\"/g, '\\"') + '"'));
+		// GitHub #3487: Unicode conversion
+		template =  decodeURIComponent(JSON.parse('"' + template.replace(/\"/g, '\\"') + '"'));
 		var lines = template.split(','),
 		template = new Array(lines.length);
 
--- a/src/main/resources/META-INF/resources/primefaces/keyboard/keyboard.js
+++ b/src/main/resources/META-INF/resources/primefaces/keyboard/keyboard.js
@@ -961,6 +961,8 @@ PrimeFaces.widget.KeyboardUtils = {
 	},
 
 	createLayoutFromTemplate : function(template) {
+	    // GitHub #3487: Unicode conversion
+	    template =  decodeURIComponent(JSON.parse('"' + template.replace(/\"/g, '\\"') + '"'));
 		var lines = template.split(','),
 		template = new Array(lines.length);
 


src/main/resources/META-INF/resources/primefaces/keyboard/keyboard.js
```diff
@@ -961,6 +961,8 @@ PrimeFaces.widget.KeyboardUtils = {
 	},
 
 	createLayoutFromTemplate : function(template) {
+	    // GitHub #3487: Unicode conversion
```
Quix0r: PrimeFaces uses tabs (tab character) for indenting code, please convert.

Quix0r (changes requested):
Please fix this.

src/main/resources/META-INF/resources/primefaces/keyboard/keyboard.js
Quix0r: Also maybe the comment about github ticket is maybe unwanted.

melloware: Comments get stripped when the library is built so they are harmless to be in there.  I actually find them useful when a piece of code looks strange to go back and look at the ticket.

melloware: Also I fixed the spaces to tabs

Quix0r: Nice, thank you. :-) No mixing of indenting styles ...

PR #3546, opened 2018-04-02
DataView

huseyinturkmenoglu:


Rapster:
Again, a description will be highly appreciated, otherwise your PR will never be  merged. To do so, create a ticket. Since it looks like a new component, I would also suggest to create a PR in the showcase

tandraschko:
I would also like to review it but its hard without example :)
I'm not sure if we really need both list- and griditem

tandraschko:
mert just commited it, will close it.

PR #3554, opened 2018-04-05
Fix #3553: FileUpload aria-disabled for buttons.

### LINKED ISSUE #3553: FileUpload: does not update the 'aria-disabled' attribute for the upload and cancel buttons
# IMPORTANT 

- Before you open an issue, test it with the current/newest version.
- Try to find an explanation to your problem by yourself, by simply debugging. This will help us to solve your issue 10x faster
- Clone this repository https://github.com/primefaces/primefaces-test.git in order to reproduce your problem, you'll have better chance to receive an answer and a solution.
- Otherwise the example must be as small and simple as possible! It must be runnable without any other dependencies (like Spring,..., or project/company internal classes)!
- Feel free to provide a PR (Primefaces is an open-source project, any fixes or improvements are welcome.)

## 1) Environment
- PrimeFaces version: 6.2
- Affected browsers: All

## 2) Expected behavior
The widget should update the 'aria-disabled' accordingly to the state of the button.
...

## 3) Actual behavior
The 'aria-disabled' is always true
..

## 4) Steps to reproduce
Chose a file to upload, the upload and cancel buttons will become active but the 'aria-disabled' attribute is not updated
.. 

## 5) Sample XHTML

..

## 6) Sample bean

..

melloware:
Fix #3553: FileUpload aria-disabled for buttons.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/fileupload/2-fileupload.js
+++ b/src/main/resources/META-INF/resources/primefaces/fileupload/2-fileupload.js
@@ -448,11 +448,11 @@ PrimeFaces.widget.FileUpload = PrimeFaces.widget.BaseWidget.extend({
     },
 
     disableButton: function(btn) {
-        btn.prop('disabled', true).addClass('ui-state-disabled').removeClass('ui-state-hover ui-state-active ui-state-focus');
+        btn.prop('disabled', true).attr('aria-disabled', true).addClass('ui-state-disabled').removeClass('ui-state-hover ui-state-active ui-state-focus');
     },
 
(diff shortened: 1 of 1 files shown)


sarvigalava:
I think you can use the PrimeFaces.skinButton function from core.js

melloware:
@sarvigalava It looks like above in bindEvents they already are skin buttton.

` PrimeFaces.skinButton(this.buttonBar.children('button'));`

However I don't think you want to call skin button again and rebind all the events etc.  I think my fix above is the safer fix.

PR #3560, opened 2018-04-06
Catch errors when resetting Google reCaptcha

msgeissler:
With the shutdown of reCaptcha API v1 the behaviour of the reset-function changed. If there is no p:captcha component on the page it will now throw an error and kill any subsequent javascript calls. This can be a problem if an ajax-request causes a forward or navigation to a new page where no captcha exists.

This change will catch the error and log it using PrimeFaces.info for a short information and PrimeFaces.debug if anyone is actually interested in the stack trace / error itself.

Issue for this PR will be created in a few minutes.

### DIFF
--- a/src/main/java-templates/org/primefaces/component/captcha/CaptchaTemplate.java
+++ b/src/main/java-templates/org/primefaces/component/captcha/CaptchaTemplate.java
@@ -56,7 +56,7 @@ import org.primefaces.json.JSONObject;
             } finally {
             	// the captcha token is valid for only one request, in case of an ajax request we have to get a new one
 	        if(context.getPartialViewContext().isAjaxRequest()) {
-	            PrimeFaces.current().executeScript("try { grecaptcha.reset() } catch (error) { PrimeFaces.info('Could not reset grecaptcha, that is OK if there is no reCaptcha on the page.'); PrimeFaces.debug(error); }");
+	            PrimeFaces.current().executeScript("if (document.getElementById("g-recaptcha-response")) { try { grecaptcha.reset() } catch (error) { PrimeFaces.error(error); } }");
 	        }
 	    }
 
--- a/src/main/java-templates/org/primefaces/component/captcha/CaptchaTemplate.java
+++ b/src/main/java-templates/org/primefaces/component/captcha/CaptchaTemplate.java
@@ -56,7 +56,7 @@ import org.primefaces.json.JSONObject;
             } finally {
             	// the captcha token is valid for only one request, in case of an ajax request we have to get a new one
 	        if(context.getPartialViewContext().isAjaxRequest()) {
-	            PrimeFaces.current().executeScript("grecaptcha.reset()");
+	            PrimeFaces.current().executeScript("try { grecaptcha.reset() } catch (error) { PrimeFaces.info('Could not reset grecaptcha, that is OK if there is no reCaptcha on the page.'); PrimeFaces.debug(error); }");
 	        }
 	    }
 
--- a/src/main/java-templates/org/primefaces/component/captcha/CaptchaTemplate.java
+++ b/src/main/java-templates/org/primefaces/component/captcha/CaptchaTemplate.java
@@ -56,7 +56,7 @@ import org.primefaces.json.JSONObject;
             } finally {
             	// the captcha token is valid for only one request, in case of an ajax request we have to get a new one
 	        if(context.getPartialViewContext().isAjaxRequest()) {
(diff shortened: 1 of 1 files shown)


msgeissler:
As a small update / clarification: I am not a 100% certain that the shutdown of the v1 API is the cause or if Google pushed an update to v2 close to the time of the shutdown. I do know that a page, that ran in production for 2+ years and with no changes to the captcha relevant code, started throwing the error described in the issue with the last weekend (which coincides with the v1 shutdown).

Another solution to the problem, if a try-catch is not the desired one, would be to only add the script if there is no navigation or forwarding happening or removing it in the case of a navigation. Even then it is not guaranteed that the captcha is still on the page / being rendered after the update, unless we can check for the presence of a captcha in a quick fashion and only run the code then.

tandraschko:
You say:

>  If there is no p:captcha component on the page it will now throw an error and kill any subsequent javascript calls.

However, the code in CaptchaTemplate.java is only executed if there is a p:captcha component.

msgeissler:
I may be wrong but:
Imagine a page with the captcha and a user presses submit. The server then has to validate the captcha, which means it will go through the `validateValue` and add the script to `PrimeFaces.current()`.
Now comes the part where I am not 100% sure what PrimeFaces / JSF does, but based on what I have seen in our production system it must be the following:
If there is now a redirect to another page, the script is still part of the response and will be executed.

The CaptchaTemplate has to run the validation before any action on the server is triggered, so it should / must happen before a redirect or any p:captcha may be hidden by the response. If you use the captcha to hide itself on success, the CaptchaTemplate also must be executed to validate the request even if it is not rendered after that anymore (so the client side js would also fail).

tandraschko:
Yep, redirect is a good case!

tandraschko:
In that case, i think "grecaptcha" is defined on the page but just the reset call fails?

tandraschko:
It would be great if could add a if-statement around it, something like: if(grecaptcha.isAvailableOnView){...}

Maybe you can check the docu of grecaptcha.

msgeissler:
> In that case, i think "grecaptcha" is defined on the page but just the reset call fails?

Yeah, that is exactly the behaviour.  It does not fail because "grecaptcha" is undefined but because of `Error: No reCAPTCHA clients exist.`, so the "greacaptcha" cannot find any component (which is true in that case).

I already had a look at the API and sadly there does not seem to be any method for that (https://developers.google.com/recaptcha/docs/display#js_api). Using the browser console I also could not find any additional information (with my limited knowledge of the API and without using any internals of grecaptcha). That is why I went with that solution for now.

Another idea was to search the page for a captcha element and trigger the reset only if one was found, but I wasn't sure if that would actually be better or worse for the performance (especially because that has to happen every time and the error would only occur in a few selected cases).

tandraschko:
Shouldn't this be enough?

`if (document.getElementById('g-recaptcha-response')) {....}`

msgeissler:
I will give that a try later. I guess getting elements by id should be quick enough.
If that works I will hopefully be able to update my PR today or tomorrow.

tandraschko:
Great!

msgeissler:
Your tip using `document.getElementById('g-recaptcha-response')` seems to work quite nicely. One or two more tests in our project and I will update the PR.

Just a quick question before that: do you think we should keep the try-catch? Looking back at the code I am already a bit amazed by the fact that it worked before without a problem. Personally, I am not sure whether or not to keep it. On the one hand we should not need the try-catch as we make sure that the element is there beforehand by checking the HTML that p:captcha should generate, on the other hand the javascript is completely out of PrimeFaces control as it is provided by Google and we never know when Google will change something again.

tandraschko:
Hmm, thats a good question. Is there a list of possible errors?

msgeissler:
The API does not specify any error code, at least not in the documentation. :/
The only one I know of is the `Error: No reCAPTCHA clients exist.` and that should be fixed with the element check. I guess leaving the try-catch in would just be a preventive action to safeguard against any .reset() errors breaking any following JS calls.
I can remove it if you want.

I am also not sure why the CI cannot build it at the moment, the JS works fine and nothing else changed. I will have another look at the build-error tomorrow.

msgeissler:
> I am also not sure why the CI cannot build it at the moment, the JS works fine and nothing else changed. I will have another look at the build-error tomorrow.

I am an idiot, I know why it is failing. Fixing it in a sec.

tandraschko:
Thanks for providing the PR!

PR #3575, opened 2018-04-10
Fix #3563 - DefaultMenuItem: use builder pattern

### LINKED ISSUE #3563: DefaultMenuItem: use builder pattern
## 1) Environment
- PrimeFaces version: 6.2

## 2) Expected behavior
Fluent API to create menu items. 
```java
DefaultMenuItem.builder()
            .value("Label")
            .icon("fa-box")
            .styleClass("ui-cart-icon")
            .outcome("recordCart")
            .ajax(false)
            .build();
```

`DefautMenuItem` has 40 attributes, builder would make creation much more easier than constructors.

## 3) Actual behavior
```java
BAMenuItem item = new DefaultMenuItem(cart.getLabel(), cart.getIcon(), null, value);
item.setOutcome(cart.getOutcome())
item.setStyleClass(cart.getStyleClass());
```

Rapster:


src/main/java/org/primefaces/model/menu/DefaultMenuItem.java
```diff
@@ -69,36 +69,26 @@
...
-    public DefaultMenuItem(Object value, String icon) {
-        this.value = value;
-        this.icon = icon;
-    }
-
-    public DefaultMenuItem(Object value, String icon, String url) {
-        this.value = value;
-        this.icon = icon;
-        this.url = url;
+    private DefaultMenuItem() {
```
cnsgithub, line 72: To not break compatibility I'd prefer to keep the public CTORs as well, maybe setting them `@Deprecated`. But that actually depends if this feature will be included in a patch or minor release?

tandraschko:
1) please don't remove the constructors
2) please make a new class in a builder package, there is no need to add this in the pojo itself.

Rapster:
1. I can deprecated the constructors, even though I feel it should no longer be present as explained in the issue

2. inner builder class is quite of a common pattern, and it's quite understandable because it's more consistent... and also, because it's not an "independant" class, as it is highly related to the pojo itself. That's the reasons why I "choose" to make it an inner class.

tandraschko:
1) We should leave it as it is, builder is just something additional and still valid to use. 
Also we can't just remove something without a "@depreaced release" before.

2) I understand that point but IMO inner classes should be avoided. In 99% cases i only use inner classes if the loc are far below 100 lines. SearchExpressionFacade#ResolveComponentCallback is a good small example IMO.
Also the builder will be inherited to a child and is even not extendable because of final.

Issue #3547, opened 2018-04-03
selectOneMenu : disabled on using placeholder


mertsincan:
Reported By PRO User;
> We are using p:selectOneMenu with placeholder option.
> We noticed that, ui-state-disabled is added when placeholder is present which makes it look disabled.
ex:
```
        <p:selectOneMenu id="console" value="" style="width:125px" placeholder="Select One">
            <f:selectItem itemLabel="" itemValue="" />
            <f:selectItem itemLabel="Xbox One" itemValue="Xbox One" />
            <f:selectItem itemLabel="PS4" itemValue="PS4" />
            <f:selectItem itemLabel="Wii U" itemValue="Wii U" />
        </p:selectOneMenu>
```

jajajaja:
Hello,

With this fix it is impossible from CSS to address component state without selection. For example if I want to display placeholder with different style. 

I suggest to add additional class like 'ui-state-no-selection' instead of removing line with 'ui-state-disabled'

PR #3583, opened 2018-04-13
Relates to #3547.

jajajaja:
Added new class for customization selectOneMenu placeholder by css.


tandraschko:
@mertsincan can you check this? This guy already asked a question in the original issue

tandraschko:
Could you please create a new issue + rename the styleclass to "ui-selectonemenu-label-placeholder"? I think it makes more sense as ui-state-* are very global

tandraschko:
Applied your change with the name "ui-selectonemenu-label-placeholder"

tandraschko:
See: https://github.com/primefaces/primefaces/issues/3994

PR #3589, opened 2018-04-16
Fixes 3582 input validation calendar

cnsgithub:


tandraschko:
the getObjectAsDate method, don't we already have a similar logic somewhere?

cnsgithub:
Unfortunately, I didn't find an appropriate method for this, just the other way around - `getValueAsString`.

Issue #3581, opened 2018-04-13
Tree: Can't scroll in MS Edge / Internet Explorer with max-height


AdriaoNeves:
## 1) Environment
- Only happens in Microsoft Edge and Internet Explorer 11. Firefox and Chrome are ok.
- Only happens if you click in the down arrow of the scroll, if you use the mouse scroll it's ok.
- It happens in the primefaces showcase  6.2, Tree component.
- PrimeFaces version: 6.2
- Does it work on the newest released PrimeFaces version? Bug occurs. Version? 6.2

- Application server + version: JBOSS 7.2
- Affected browsers:
IE11, Edge.

## 2) Expected behavior
Scroll all the way down.

## 3) Actual behavior
If you set a max-height to the tree container, when you try to scroll down the scroll it will never reach bottom, it will jump back after some clicks.
![tree-scroll-issue](https://user-images.githubusercontent.com/423796/38725253-a571b4ee-3efe-11e8-82ec-8a3a148cf172.gif)

## 4) Steps to reproduce
It happens in the primefaces showcase  6.2.
- Configure a tree with a max-height.
- Scroll down by clicking in the down arrow.
- The scroll will never reach bottom.
- If you use the mouse scroll it works ok.

## 5) Related
The following new feature when implemented may solve the issue:
https://github.com/primefaces/primefaces/issues/15

HuseyinTurkmenoglu:
Hi, 

I created sample code for your issue. But I don't see any problem according to your gif for Microsoft Edge and IE 11.

Code: 

```
<h:form>
            <h3 style="margin-top:0">Client</h3>
            <p:scrollPanel mode="native" style="width:250px; max-height: 170px">
                <p:tree value="#{treeBasicView.root}" var="node">
                    <p:treeNode>
                        <h:outputText value="#{node}" />
                    </p:treeNode>
                </p:tree>
            </p:scrollPanel>
            <h3>Ajax</h3>
            <p:tree value="#{treeBasicView.root}" var="node" dynamic="true">
                <p:treeNode>
                    <h:outputText value="#{node}" />
                </p:treeNode>
            </p:tree>
</h:form>
```

If Sample code is not correct according to your code, please you could create sample code?

huseyinturkmenoglu:
This problem is caused by the focus event. Chrome is not getting the focus, but IE and Microsoft Edge.

I'm working on this PR for [15](https://github.com/primefaces/primefaces/issues/15) .

[...]

PR #3607, opened 2018-04-25
Tree Scrollable

huseyinturkmenoglu:
This PR  is for https://github.com/primefaces/primefaces/issues/3581


tandraschko:
-1 for treeStyleClass
styleClass is actually used for the root element, which is actually the tree

PR #3608, opened 2018-04-25
TreeScrollable

huseyinturkmenoglu:
This PR for https://github.com/primefaces/primefaces/issues/3581


tandraschko:
As i already said in the other PR:
treeStyleClass is IMO the same as styleClass, not sure why you introduce a new attribute.

PR #3611, opened 2018-04-25
Crash when label value contains CSS selector special chars

xtron:
Issue https://github.com/primefaces/primefaces/issues/3612


cnsgithub (approved):

tandraschko:
Please create a issue with example first.

xtron:
created https://github.com/primefaces/primefaces/issues/3612

tandraschko:
@melloware can you check?=

melloware (approved):

melloware:
@tandraschko verified this is a good change and I do not see any side effects.

PR #3636, opened 2018-05-03
Fix proposal for Issue #3635

### LINKED ISSUE #3635: Chart: LineChart not working when beginning with null-values
## 1) Environment
- PrimeFaces version: 6.2
- Does it work on the newest released PrimeFaces version? Version? 6.2 is newest release
- Application server + version: Wildfly 10
- Affected browsers: Firefox, Chrome (latest)
- Example: See gist: https://gist.github.com/metaxmx/49a966c022f40bcdcfbe51f8596c94af

## 2) Expected behavior
- In LineCharts points with the y-axis value "null" are ignored by the graph

## 3) Actual behavior
- If the first point in a chart series has a value "null" the graph is not rendered at all (only the canvas).

## 4) Steps to reproduce
- Create PrimeFaces Chart of type "line"
- Create Chart Model with a Chart Series, containing some null-Values in the middle -> no problem, points are ignores
- Create Chart Model with a Chart Series, starting with a null-Value
- chart is not rendered

## 5) Sample XHTML

```
<p:chart type="line" model="#{chartController.chartModel}"
         id="chartTest"
         style="width:100%; height:210px;">
</p:chart>

<p:chart type="line" model="#{chartController.chartModelWithNull}"
         id="chartTest2"
         style="width:100%; height:210px;">
</p:chart>
```

## 6) Sample bean

```
package com.illucit.controller;

import org.primefaces.model.chart.ChartSeries;
import org.primefaces.model.chart.LineChartModel;

import javax.enterprise.inject.Model;
import java.util.LinkedHashMap;
import java.util.Map;

@Model
public class ChartController {

    private LineChartModel chartModel = new LineChartModel();

    private LineChartModel chartModelWithNull = new LineChartModel();

    public ChartController() {
        chartModel.setTitle("Test Chart");
        ChartSeries series = new ChartSeries();
        Map<Object, Number> data = new LinkedHashMap<Object, Number>();
        data.put(10, 10);
        data.put(20, 30);
        data.put(30, 10);
        data.put(40, null);
        data.put(50, 15);
        series.setData(data);
        chartModel.addSeries(series);

        chartModelWithNull.setTitle("Test Chart, beginning with Null value");
        ChartSeries seriesWithNull = new ChartSeries();
        Map<Object, Number> dataWithNull = new LinkedHashMap<Object, Number>();
        dataWithNull.put(10, null);
        dataWithNull.put(20, 30);
        dataWithNull.put(30, 30);
        dataWithNull.put(40, 40);
        dataWithNull.put(50, 60);
        seriesWithNull.setData(dataWithNull);
        chartModelWithNull.addSeries(seriesWithNull);
    }

    public LineChartModel getChartModel() {
        return this.chartModel;
    }

    public LineChartModel getChartModelWithNull() {
        return chartModelWithNull;
    }
}
```

## 7) Analysis
- null values are encoded as JavaScript String "null", instead of JavaScript value null as in older PrimeFaces versions
- Apparently jqPlot does not work well, when it encounters a String as a value where it expects a number

metaxmx:
This commit fixes #3635 
y-values in LineChart that are null are rendered as JS null, not as JS String "null"

### DIFF
--- a/src/main/java/org/primefaces/component/chart/renderer/LineRenderer.java
+++ b/src/main/java/org/primefaces/component/chart/renderer/LineRenderer.java
@@ -39,7 +39,7 @@ public class LineRenderer extends CartesianPlotRenderer {
             for (Iterator<Object> x = series.getData().keySet().iterator(); x.hasNext();) {
                 Object xValue = x.next();
                 Number yValue = series.getData().get(xValue);
-                String yValueAsString = escapeChartData(yValue);
+                String yValueAsString = (yValue == null) ? "null" : escapeChartData(yValue);
                 String xValueAsString = escapeChartData(xValue);
 
                 writer.write("[");


tandraschko:
does "null" fix that or should it better be an empty string?

metaxmx:
jqPlot seems to interpret an empty String as value 0, showing a wrong data point, as where a "null" value will omit the data point.

So, in my graph tests "null" has introduced the expected behaviour again (the same as it was in PrimeFaces 6.1 and before).

PR #3653, opened 2018-05-09
Fix #3625

### LINKED ISSUE #3625: [Documentation] AjaxStatus: never shows the success and error facets.
## 1) Environment
- PrimeFaces version: 6.2.0
- Does it work on the newest released PrimeFaces version? No.
- Application server + version: Wildfly 12.0
- Affected browsers: All

## 2) Expected behavior
When an ajax request fails ajaxStatus should show its error facet. Likewise in the case of success the success facet should be shown.

## 3) Actual behavior
The complete event always immediately follows an error or success event. This causes the success and error facets to be hidden immediately in favor of the complete facet.

## 4) Steps to reproduce
Just use the sample code below and watch the error facet never appearing.

## 5) Sample XHTML
```xml
<p:ajaxStatus>
	<f:facet name="error">
		<h:outputText value="Error" />
	</f:facet>
	<f:facet name="success">
		<h:outputText value="Success" />
	</f:facet>
	<f:facet name="default">
		<h:outputText value="Idle" />
	</f:facet>
	<f:facet name="start">
		<h:outputText value="Sending" />
	</f:facet>
	<f:facet name="complete">
		<h:outputText value="Done" />
	</f:facet>
</p:ajaxStatus>

<h:form>
    <p:poll listener="#{counterView.increment}"/>
</h:form>
```

## 6) Sample bean
```java
package org.primefaces.showcase.view.ajax;
 
import java.io.Serializable;
 
@ManagedBean
@ViewScoped
public class CounterView implements Serializable {
     
    public void increment() {
        throw new IllegalStateException("Boy, that escalated quickly.");
    }
}
```
## 7) Analysis
From my point of view the culprit is https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/ajaxstatus/ajaxstatus.js#L38 The complete event always marks the end of an ajax request regardless of its outcome. Because in the time between the success / error event and the complete event only the DOM tree is updated the error and success facets never have a chance to appear on screen at all. I see the following options to resolve this issue:

1. If no complete facet is defined and either the error or success facet is currently visible then don't hide them. Additionally you could issue a warning message, when the complete facet is defined together with a success or error facet to make the user aware they will never be shown.
1. Deprecate the success and error facets because they are useless in their current state.

bzurmaar:
Fix #3625 for the problem details.

f5a88de8768a9e6f1c5da6b276ac0ef3e68847b7 introduces the following change: When an ajax request completes the success and error facets are only hidden when a complete facet is present. Otherwise the success or error facet respectively will remain visible until a new ajax request is started.

Additionally to this change the documentation should be updated to reflect this imho .

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/ajaxstatus/ajaxstatus.js
+++ b/src/main/resources/META-INF/resources/primefaces/ajaxstatus/ajaxstatus.js
@@ -35,7 +35,13 @@ PrimeFaces.widget.AjaxStatus = PrimeFaces.widget.BaseWidget.extend({
             callback.apply(document, args);
         }
 
-        this.jq.children().hide().filter(this.jqId + '_' + event).show();
+        if (event !== 'complete' || this.jq.children().filter(this.toFacetId('complete')).length) {
+            this.jq.children().hide().filter(this.toFacetId(event)).show();
+        }
+    },
+
+    toFacetId: function(event) {
+        return this.jqId + '_' + event;
     },
 
     bindToStandard: function() {


tandraschko:
does this work for all cases?
I assumed a if statement like: 
`if (event==error|sucess && noCompleteFacetDefined) { return; }`

bzurmaar:
According to my tests it works like a charm.

* The `event !== 'complete'` ensures all events different from complete are handled the same way they used to be before the change.
* In the other case, i.e. the complete event, the switching of the facet only happens if `this.jq.children().filter(this.toFacetId('complete')).length` is truthy, i.e. a complete facet was defined. If no complete facet is defined the previously shown facet remains visible. This may either be the error or success facet or no facet at all if success and error weren't defined.

I don't get the logic behind your if statement. If the event is error or success it's too early to prevent the complete facet from being shown. You can only make this decision in the complete event.

tandraschko:
all right!

Issue #3477, opened 2018-03-12
SelectOneMenu: Editable displays value instead of label


JHookDev:
## 1) Environment
 - PrimeFaces version: 6.2
 - Does it work on the newest released PrimeFaces version? No Version? 6.2
 - Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source)
 - Application server + version: Tomcat 7.0.23
 - Affected browsers: Tested with Chrome and IE
 
 ## 2) Expected behavior
 When rendering the dropdown, if the value matches a value in the select items, it should display the matching label for that value.
 
 
 ## 3) Actual behavior
 When rendering the dropdown, the value is displayed even when the value matches an item in the select items.
 
 ## 4) Steps to reproduce
 Using the sample below XHTML and Bean below, select a value from the dropdown and press submit. Upon the page being rerendered you will see the value for the selected item rather than the label.
  
 
 ## 5) Sample XHTML
 ```
<f:view>
  <h:form id="selectOneMenuForm">
    <p:outputLabel for="editableDropdown" value="Editable: " />
    <p:selectOneMenu id="editableDropdown" value="#{selectOneMenu.selection}"
        effect="fold" editable="true">
        <f:selectItem itemLabel="Select One" itemValue="" />
        <f:selectItems value="#{selectOneMenu.options}" />
    </p:selectOneMenu>
    <p:commandButton id="submit" value="Submit" ajax="false"/>
  </h:form>
</f:view>
```
 
 
 ## 6) Sample bean
 ```
@ManagedBean
public class SelectOneMenu {
	
	private String selection;
    private Map<String,String> options = new HashMap<String, String>();

    @PostConstruct
	public void init() {
		options = new HashMap<String, String>();
		options.put("Label1", "Value1");
		options.put("Label2", "Value2");
	}

	public String getSelection() {
		return selection;
	}

	public void setSelection(String selection) {
		this.selection = selection;
	}

	public Map<String, String> getOptions() {
		return options;
	}

	public void setOptions(Map<String, String> options) {
		this.options = options;
	}

}
```

melloware:
This was fixed before so there must be a regression somewhere:  https://github.com/primefaces/primefaces/issues/2862

melloware:
[...]

PR #3685, opened 2018-05-18
Fix# 3477

JHookDev:
#3477 

Modify encodeLabel so that the label of the matching select item is presented as the input's value when the selectOneMenu is editable. If no match in select items, then the valueToRender will be displayed.


melloware (approved):

tandraschko:
I don't know the editable selectOneMenu exactly but does it make sense the render the label as "value" attr?

JHookDev:
@tandraschko I think it does, so that the user sees the label that corresponds to the SelectOneMenu selection, whether it's from a stored value or one they just selected. The SelectOneMenuRenderer also handles decoding editable versions, in that it searches the select items labels for a match to the input's value attribute. Setting the submitted value to the select item value if a match is found. If a match is not found, the submitted value remains the input's value attribute, i.e. the edited value from the user.

PR #3690, opened 2018-05-19
Remove checkstyle exclusions

krichter722:
The NullPointerException which required excludes in the checkstyle
Maven plugin configuration are fixed in version 3.0.0 so that they can
be removed. The necessary changes in the code are done.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -274,11 +274,10 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-checkstyle-plugin</artifactId>
-                <version>2.17</version>
+                <version>3.0.0</version>
                 <configuration>
                     <consoleOutput>true</consoleOutput>
                     <configLocation>./checkstyle.xml</configLocation>
-                    <excludes>**/UIData.java,**/Base64.java,**/UITabPanel.java,**/json/**,**/UITree.java,**/UIRepeat.java,**/AbstractBehaviorHandler.java</excludes>
                 </configuration>
                 <executions>
                     <execution>
--- a/src/main/java/org/primefaces/behavior/ajax/AjaxBehavior.java
+++ b/src/main/java/org/primefaces/behavior/ajax/AjaxBehavior.java
@@ -60,10 +60,14 @@ public class AjaxBehavior extends AbstractBehavior implements AjaxSource {
         form(String.class),
         skipChildren(Boolean.class);
 
-        public final Class<?> expectedType;
+        private final Class<?> expectedType;
 
         PropertyKeys(Class<?> expectedType) {
             this.expectedType = expectedType;
+        }
+
+        public Class<?> getExpectedType() {
+            return expectedType;
         }
     }
 
--- a/src/main/java/org/primefaces/behavior/ajax/AjaxBehaviorHandler.java
+++ b/src/main/java/org/primefaces/behavior/ajax/AjaxBehaviorHandler.java
@@ -80,25 +80,25 @@ public class AjaxBehaviorHandler extends AbstractBehaviorHandler<AjaxBehavior> {
         Application application = ctx.getFacesContext().getApplication();
         AjaxBehavior behavior = (AjaxBehavior) application.createBehavior(AjaxBehavior.BEHAVIOR_ID);
 
-        setBehaviorAttribute(ctx, behavior, this.process, AjaxBehavior.PropertyKeys.process.expectedType);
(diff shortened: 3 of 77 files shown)


tandraschko:
Thanks for your other PRs but i wont apply this PR.
Most of the changes lines are by design currently.

Also i wont touch e.g. UIData/TabPanel as most of it's code is a copy of Mojarra AFAIR. Same for the json packages.

krichter722:
> Most of the changes lines are by design currently.

I'm not sure I understand this point. The PR improves code quality which has few if anything to do with design. If you consider encapsulation of fields a design change, then it'd be less than a fraction of a percent of the PR and it can still be excluded, even though it won't have measurable impact on anything afaik except allowing to work with the fields more easy.

The changes in `UIData` and `TabPanel` don't contain any functional changes, the same goes for the JSON classes. Are the classes synced regularily by copying from somewhere or have they been copied once - either is not a good idea btw?

I don't want force you to your luck. I just want to make sure that the changes here are understand - and shed some light the code copying practice.

tandraschko:
the inline if's are per design currently, the primetek team uses it heavily.


if regularily means 1-x years, yes :D
In the future we could maybe just switch to javaee json apis but i dont think it will happen in the next years as we still have spring users e.g.

krichter722:
I accept your decision to not merge this. The work isn't lost somehow and it can still be done in the future. Just some last points which might enhance your point of view:

> the inline if's are per design currently, the primetek team uses it heavily.

If you're referring by "inline ifs" to if statements in one line: that just makes the code harder to read. It's just a style issue which you're disallowing in 99% of your classes by checkstyle for a reason.

> In the future we could maybe just switch to javaee json apis but i dont think it will happen in the next years as we still have spring users e.g.

Sounds reasonable. Whatever it is, just don't continue copying code :)

krichter722:
An example why this is very useful and you should merge this (and probably will after you spent a ton of time for issues like the following): https://github.com/primefaces/primefaces/pull/3708/commits/79edba79c5c01496050df4b607ff012e23d0d27a#diff-268a0e0efb864521be5657c7dca21bf4L189 (changes of code style enlarge PRs and diffs which prolongs reviews, bisects and other investigation of issues and just look unprofessional).

tandraschko:
I have no idea how's that related to the linked PR and our discussion about e.g. UIData excludes and the inline if-statements.

krichter722:
> I have no idea how's that related to the linked PR and our discussion about e.g. UIData excludes and the inline if-statements.

The marked line 189 of OutputLabelRenderer.java shows that the PR author adds an unnecessary non-functional change which is forbidden by Primefaces checkstyle setup (enforces newline between `}` and `catch`) and thus enlarges the PR + review and needs to be removed later on.

I assume that the change to 3.0.0 and/or removal of excludes makes the whole checkstyle thing work in the first place (and thus has bigger impact than just on the excluded files) because otherwise it's not explainable how this PR can [pass Travis](https://travis-ci.org/primefaces/primefaces/builds/382597785). interesting...

tandraschko:
Weird actually yes. But i have no problem to merge a pr which just upgrades the plugin

krichter722:
The thing is that now, I remember why I suggested the removal of excludes (I agree that it make few sense although not no sense at all to remove them[1]). I didn't share the following, sorry, and have some memory gaps: I noticed that most of the checkstyle checks simply don't work and you can see in the list of changed files that checkstyle correctly finds a ton of issues in files which are not included.

Back, when I created the PR, I assumed that `excludes` where only added because if maven-checkstyle-plugin 2.17 is executed without excludes it fails due to

```
Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:check (checkstyle) on project primefaces: Execution checkstyle of goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:check failed. NullPointerException -> [Help 1]
```

That is not the case as you explained and it was my bad assumption to believe otherwise.

The question is now how to get the checkstyle plugin to detect all the checkstyle violoations which have been fixed in the PR (after investigation, the changes to excluded files can be dropped, of course). 77 Files have been touched because of checkstyle issues, ~10 are excluded, so something is definitely not right, here.

I believe that the checkstyle plugin started working as it should, i.e. showing issues in other files, after I fixed all issues in the excludes, but I have no good idea how to verify that.

---

[1] I'd like to stress that syncing files like that seems not ideal and that I'd like to see the excludes being removed after they're made obsolete one day.

PR #3695, opened 2018-05-22
Make tree-related root rowkey a constant

krichter722:


tandraschko:
The root string in SEF has nothing to do with the tree, so please dont change this.

src/main/java/org/primefaces/expression/SearchExpressionConstants.java
```diff
@@ -33,7 +35,7 @@
     public static final String ROW_KEYWORD =                KEYWORD_PREFIX + "row";
     public static final String COLUMN_KEYWORD =             KEYWORD_PREFIX + "column";
     public static final String ID_KEYWORD =                 KEYWORD_PREFIX + "id";
-    public static final String ROOT_KEYWORD =               KEYWORD_PREFIX + "root";
+    public static final String ROOT_KEYWORD =               KEYWORD_PREFIX + ROOT_ROW_KEY;
```
melloware: You can remove this one as SEF keyword root may change independently of the Tree root rowkey.

melloware (changes requested):

src/main/java/org/primefaces/expression/SearchExpressionConstants.java
krichter722: @melloware My bad, removed and updated PR

PR #3696, opened 2018-05-22
Add initial .gitattributes

krichter722:
.gitattributes with '* text=auto' converts line endings automatically
on check-in and check-out which reduces merge work and discussion over
long term.

Unfortunately this requires a large set of files to be changed once to change the line feed to LF. Changes after adding `.gitattributes` are not required ever again.


tandraschko:
added defaults like suggested by github https://help.github.com/articles/dealing-with-line-endings/

Commit:
https://github.com/primefaces/primefaces/commit/e67a480234d8b9b98e4e2186560a2291ce5f37f9

will do the change later

Also i don't like to merge such big PR's, unable to see other changes if done.

tandraschko:
Please review and i will refresh the repo like mentioned in the link

PR #3701, opened 2018-05-22
Fix #3700

### LINKED ISSUE #3700: CommandLink: Firefox warning "unreachable code after return statement" for non-ajax CommandLink
## 1) Environment
- PrimeFaces version: 6.2.4
- Affected browsers: Firefox 60.0

## 2) Expected behavior

When using a <p:commandLink ajax="false" /> with a <p:fileDownload /> to provide a downloadable File, then a click on the link should provide the downloaded file, but stay on the current view.

THere should be no error or warning.

## 3) Actual behavior

When I click the Link in Firefox, the download works fine, but I also get an annoying Console-Warning "**unreachable code after return statement**" ([this one)](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Fehler/Stmt_after_return)). 

In Chrome this warning does not occur.

## 4) Steps to reproduce

1) Put a `<p:commandLink ajax="false">` on a view, with a nested `<p: downloadFile />`
2) Open the view in Firefox and click the link (with open JavaScript Console)
3) The download is executed, but also a warning "unreachable code after return statement" is printed to the console, making it harder to detect real errors and warnings in the application

## 5) Sample XHTML

View Code:
```
<p:commandLink ajax="false">
	<p:fileDownload value='#{downloadBean.downloadExcel([...])}' contentDisposition="attachment" />
	Download
</p:commandLink>
```

Generated Html Code:
```
<a onclick="PrimeFaces.addSubmitParam('theform',{'[removed]'}).submit('theform');return false;PrimeFaces.onPost();">Download</a>
```

## 6) Analysis

The problem seems to originat from CoreRenderer.java. The Component CommandLink calls the function `CoreRenderer.buildNonAjaxRequest()` if `ajax=false` which is required for file downloads to work.

```
[...]
        if (submit) {
            Object target = component.getAttributes().get("target");
            request.append(".submit('").append(formId).append("'");

            if (target != null) {
                request.append(",'").append(target).append("'");
            }

            request.append(");return false;");
        }

        if (!submit && !params.isEmpty()) {
            request.append(";");
        }

        request.append("PrimeFaces.onPost();");

        return request.toString();
```

As you can see, if submit=true (which is true in this case) after rendering "return false;" there is also rendered an obsolete "PrimeFaces.onPost();" which is never executed, but which results in this warning.

metaxmx:
#3700

### DIFF
--- a/src/main/java/org/primefaces/renderkit/CoreRenderer.java
+++ b/src/main/java/org/primefaces/renderkit/CoreRenderer.java
@@ -491,7 +491,9 @@ public abstract class CoreRenderer extends Renderer {
             request.append(";");
         }
 
-        request.append("PrimeFaces.onPost();");
+        if (!submit) {
+            request.append("PrimeFaces.onPost();");
+        }
 
         return request.toString();
     }


tandraschko:
I dont thinkt he fix is correct currently, but i also don't think the initial functionality of onPost works ;)
Will have to think about it.

tandraschko:
Thanks for the PR but fixed it with:
https://github.com/primefaces/primefaces/commit/65edccafe3f57d1d1ca2650e31236cd55bf98101

please give it a try

metaxmx:
Hi,
your fix still calls the "onPost()" when a non-ajax call with submit=true is executed.
The onPost() however stops all running ajax calls. When you are triggering a download by this via <p:fileDownload />, stopping all background ajax-responses ist probably not what you want because you stay on the page.

I assume, there needs to be an additional distinction between a regular non-ajax request (i.e. submitting a form via POST; then all running ajax-responses may be terminated) and a Download-Request (seems like a non-ajax request, but when the server answers with a download, the user stays on the view and still can consume all running ajax-responses).

alfonx:
tandraschko  wrote:
> Thanks for the PR but fixed it with:
> [65edcca](https://github.com/primefaces/primefaces/commit/65edccafe3f57d1d1ca2650e31236cd55bf98101)
> please give it a try

We tried and tried, but your fix does NOT work. Please reopen this Ticket and/or #3700 again.

tandraschko:
@metaxmx removed the onPost method for non-ajax post scripts

metaxmx:
Thank you very much!
The <p:fileDownload> call works now as expected.

Issue #3673, opened 2018-05-17
Schedule: add rendering attribute to org.primefaces.model.ScheduleEvent


OBreidenbach:
I want to use the feature 'background-events' from full calendar: 
https://fullcalendar.io/docs/background-events

For that I want to add an attribute to org.primefaces.model.ScheduleEvent. 
I will provide a PR but I am not sure what type the attribute should have. In full calender you can set it to
rendering: 'background'
or
rendering: 'inverse-background'
So making is just a string attribute and the user needs to know, which values to set?
Or make it boolean and just support 'background' and not 'inverse-background'
Or something else?
What do you think?

melloware:
Just my two cents...I like supporting all possibilities.  So I suggest making it a string and documenting it correctly with the options.  Also you should possibly throw an exception if the value is not one of the expected values letting someone know they configured the component incorrectly.

OBreidenbach:
PR https://github.com/primefaces/primefaces/pull/3703
I did create a enum for carrying the rendering value. No need of exceptions or documenting the correct values.
Tell me what you think,

PR #3703, opened 2018-05-23
#3673 add rendering attribute to the schedule event to use full calen…

OBreidenbach:
…dar feature
https://github.com/primefaces/primefaces/issues/3673


tandraschko:
I would create a seperate class file ScheduleRenderingMode instead of doing it via subclasses.

OBreidenbach:
Superseded by https://github.com/primefaces/primefaces/pull/3730

PR #3708, opened 2018-05-23
OutputLabelRenderer: Evaluate composite constraints

msgeissler:
The current OutputLabelRenderer will look at the constraints directly added to the value property and will miss any constraint added as part of a composite constraint.

This PR will change the behaviour and also look at the composite constraints to check whether bean validation (or the required marker) should be set.

(The issue for this PR will follow later.)


tandraschko:
Will close it for now as discussed in the issue

PR #3716, opened 2018-05-23
Fixed #3715: InputMask: incorrect handling of optional characters

### LINKED ISSUE #3715: InputMask: incorrect handling of optional characters
## 1) Environment
- PrimeFaces version: **6.3-SNAPSHOT**
- Does it work on the newest released PrimeFaces version? Version? **No, 6.3-SNAPSHOT**
- Application server + version: **Wildfly Full 12.0.0.Final**
- Affected browsers: **Chromium 66**

## 2) Expected behavior
An InputMask whose mask has X optional characters should accept any number of optional characters lower than X. Example: "9?9999" should consider the following valid: "9", "99", "999", "9999", etc

## 3) Actual behavior
An InputMask whose mask has X optional characters accepts: 0 optional characters or X optional characters. Example:  "9?9999" considers valid "9" and "99999".

The input value can be valid to the Javascript part of the component, and not to the Java implementation, in which case the value is set to null.

## 4) Steps to reproduce
1. Use an InputMask with the mask: "9?999"
2. Type "99" and submit the form
3. The front end will accept the value (it will not be cleared)
4. The value will be null to the backing bean

## 5) Sample XHTML
```
...
<h:form>
<p:inputMask value="#{bean.value}" mask="9?999" />
<p:commandLink actionListener="#{bean.submit()}"/>
</h:form>								
...
```

## 6) Sample bean
```
@ManagedBean
@ViewScoped
public class Bean {
   private String value;

   public void submit() {
         assert this.value != null;  //fails
   }
...
}

```

tiagojn:
See: #3715

### DIFF
--- a/src/main/java/org/primefaces/component/inputmask/InputMaskRenderer.java
+++ b/src/main/java/org/primefaces/component/inputmask/InputMaskRenderer.java
@@ -22,8 +22,8 @@ import java.util.regex.Pattern;
 import javax.faces.component.UIComponent;
 import javax.faces.context.FacesContext;
 import javax.faces.context.ResponseWriter;
-import org.primefaces.context.PrimeApplicationContext;
 
+import org.primefaces.context.PrimeApplicationContext;
 import org.primefaces.renderkit.InputRenderer;
 import org.primefaces.util.ComponentUtils;
 import org.primefaces.util.HTML;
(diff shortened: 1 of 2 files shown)


tiagojn:
The build fails, it looks like a test is incorrect:

the mask **"9?99"** should generate the regex: `"[0-9][0-9]?[0-9]?"` and not `"[0-9]([0-9][0-9])?"`

cnsgithub:
Yes, please feel free to adjust the test accordingly.

cnsgithub (approved):

tandraschko:
I would apply it but the tests must be valid.

tiagojn:
My implementation was wrong because it didn't handle literals and whitespaces correctly. I fixed it along with the expected results of the tests.

tiagojn:
third time's a charm, fixed identation issues (sorry this is my first PR)

PR #3717, opened 2018-05-23
Messages: Removed duplicate detail and summary from messages

mafaul:

image


mafaul:
https://stackoverflow.com/questions/13467881/duplicate-text-in-pmessages-summary-and-detail

tandraschko:
As already mentioned, Pull requests without detailed issues will not get accepted.
Also probably this PR will not get accepted as you can just leave the detail empty in your own code, other people may like exactly this behavior.

mafaul:
Apologies - I will create a detailed issue then. Is is specified like this:
<p:inputText required="true" requiredMessage="Value Required" />

tandraschko:

PR #3718, opened 2018-05-23
Growl: Update growl not to show detail if it matches the content of summary

mafaul:

image


tandraschko:

PR #3719, opened 2018-05-23
Message: Only show detail if it actually differs from summary

mafaul:

image


tandraschko:
Pull requests without detailed issues will not get accepted.

tandraschko:

PR #3722, opened 2018-05-24
Messages, Message, Growl: Do not show duplicate values for summary and detail

mafaul:
If summary and detail is visible, only display the summary content if the detail would have been exactly the same.

If only summary or only detail is visible, it should not be affected by this PR.

If the detail message differs from the summary, both should be displayed.


tandraschko:
see the ticket, will close it for now.
Please also don't create a PR until we discussed it.

Issue #3737, opened 2018-05-29
Schedule: fullcalendar fix date formatting ranges with postfix period characters


OBreidenbach:
## 1) Environment
- PrimeFaces version: 6.2
- Does it work on the newest released PrimeFaces version? Version? no, 6.2
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) no
- Application server + version: wildfly 10.1
- Affected browsers: firefox, ...

## 2) Expected behavior

Date ranges should render as followed: 4. – 10. Jun 2018 (german date format)

## 3) Actual behavior

Date ranges render as followed: 4 – 10. Jun 2018 (german date format)

I just ported the full calender fix, commit:
78616b211b8e27800040a9ee5570b1594ea0d78a
12/2/2017 in file date-formatting.ts

OBreidenbach:
PR https://github.com/primefaces/primefaces/pull/3738

OBreidenbach:
Obsolete because of https://github.com/primefaces/primefaces/issues/1817 upgrading FullCalendar to 3.9.0.

Issue #1817, opened 2016-10-07
Schedule: Update to new version of Fullcalendar (3.9.0)


luizphilipi:
In version [3.0.0](https://github.com/fullcalendar/fullcalendar/releases/tag/v3.0.0) Fullcalendar has a new feature, List View. This will improve user experience a lot.

erickdeoliveiraleal:
I think it will not be updated for now because FullCalendar uses jQuery 2 and PF uses 1.9, also FullCalendar 3 does not support IE 8, which is a requirement for now in PF.

erickdeoliveiraleal:
https://github.com/fullcalendar/fullcalendar/releases/tag/v3.0.0

tandraschko:
exactly! we can plan for the future.

luizphilipi:
Ok, thanks for the quick response.

luizphilipi:
Hi @tandraschko, now with jquery upgrade this issue turns viable, right?
FullCalendar needs jquery v2.0.0+, i just made some changes to schedule.js and everythings works right, now we use list view in our project.

tandraschko:
Yep! If you would provide a PR, i can review and test it.

melloware:
Duplicate of: https://github.com/primefaces/primefaces/issues/2626

mbeedub:
+1 for this

It would be a fantastic addition to PF and would pave the way potentially for the use of FullCalendar scheduler (even though that's a premium add-on)

andreichiritescu:
any idea when this feature will be added?

tandraschko:
Without milestone means no plan. But you can sponsor it or implement by yourself and provide a pull request.

OBreidenbach:
PR https://github.com/primefaces/primefaces/pull/3739

OBreidenbach:
I did add a new attribute for slotLabelFormat and deprecated the axisFormat attribute, because there were already some other deprecated attributes (for Instance: ignoreTimezone).
In my humble opinion all this deprecated attributes should be removed. They refer to FullCalendar attributes already removed. It is confusing if you look at the primefaces and the FullCalendar documentation.

tandraschko:
@OBreidenbach 
Just for you info:
- I also created a issue for removing the depracted attributes after the next versio (6.4 or 7.0)
- I also created a issue for the documentation
- I also adjusted the migration guide

It's always more the like code changes.

OBreidenbach:
@tandraschko 
Thank you for the info.

PR #3738, opened 2018-05-29
#3737 fix date formatting ranges with postfix period characters from …

OBreidenbach:
…fullcalendar

https://github.com/primefaces/primefaces/issues/3737


OBreidenbach:
Obsolete because of #1817 upgrading FullCalendar to 3.9.0.

PR #3739, opened 2018-05-30
#1817 Upgrade FullCalendar 2.9.1 -> 3.9.0

OBreidenbach:
https://github.com/primefaces/primefaces/issues/1817


tandraschko:
Is the yui Upgrade required? Had problems in the past with IT when enable compression with the release profile

OBreidenbach:
Without the yui Upgrade I got the errors in the attached file during the release build. So the upgrade solved those.
[exceptionUpgradeFullCalendar.txt](https://github.com/primefaces/primefaces/files/2074949/exceptionUpgradeFullCalendar.txt)

I also so tried to exclude the schedule.js in the pom.xml, that also solved the problem. But I thought that the yui upgrade is better solution.

tandraschko:
Yep, tried upgrading the plugin again but the "release" profile doesn't compress the resources anymore. If you can fix that, i can merge your PR.

OBreidenbach:
Compression is fixed by setting force=true in "release" profile for the yui plugin.
But then the exceptions during compressing the schedule.js were back.

I now use the minified file from fullcalendar and aggregate it during the build with the primefaces extensions. So I do not have to minify the fullcalendar.js.
It also has the advantage for future upgrades that you just has to replace the js file.

I guess the yui upgrade was not necessary. But it is working now, so you maybe want to keep it?

Maybe the yui plugin should be replaced by some other plugin, (Minify Maven Plugin?) which is able to handle the fullcalendar.js file, in the long run.

tandraschko:
Thanks, i like the changes!

PR #3773, opened 2018-06-12
Fixes 3772 xss tristatecheckbox

cnsgithub:
Now `ResponseWriter.startTag/writeAttribute` methods are used in conjunction with proper JSON escaping of `data-titlestates` attribute values. The latter made some problems when attribute values are wrapped in single quotes, therefore JSON array was changed to JSON object internally (more on this problem: https://stackoverflow.com/questions/6071618/how-to-pass-an-array-into-jquery-data-attribute).

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/tristatecheckbox/tristatecheckbox.js
+++ b/src/main/resources/META-INF/resources/primefaces/tristatecheckbox/tristatecheckbox.js
@@ -92,8 +92,9 @@ PrimeFaces.widget.TriStateCheckbox = PrimeFaces.widget.BaseWidget.extend({
             // change title to the new one
             var iconTitles = this.box.data('titlestates');
             if(iconTitles!=null) {
-                if(iconTitles.titles!=null && iconTitles.titles.length>0) {}
+                if(iconTitles.titles!=null && iconTitles.titles.length>0) {
                     this.box.attr('title', iconTitles.titles[newValue]);
+                }
             }
 
             // fire change event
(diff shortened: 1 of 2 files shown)


src/main/resources/META-INF/resources/primefaces/tristatecheckbox/tristatecheckbox.js
```diff
@@ -91,8 +91,10 @@ PrimeFaces.widget.TriStateCheckbox = PrimeFaces.widget.BaseWidget.extend({
 
             // change title to the new one
             var iconTitles = this.box.data('titlestates');
-            if(iconTitles!=null && iconTitles.length>0){
-                this.box.attr('title', iconTitles[newValue]);
+            if(iconTitles!=null) {
+                if(iconTitles.titles!=null && iconTitles.titles.length>0) {
```
melloware, line 95: Why is this iconTitles != null done twice in a row?

melloware (changes requested):

src/main/resources/META-INF/resources/primefaces/tristatecheckbox/tristatecheckbox.js
cnsgithub: Sorry? The second one tests for iconTitles.**titles**. This has been introduced due to the problem mentioned above.

We could however compact the code into a single line... Instead of nesting the two if statements...

melloware: Ahh I see.

cnsgithub: Would you like to have a single line?

melloware: No I think a single line makes it messier to read. I am OK with 2 lines just my eyes didn't see the ".title" in the 2nd line.

cnsgithub: Sorry, too late. Just committed...

tandraschko:
Thanks for the PR!
I changed the if to a one-liner and also did some cosmetics and probably a small bugfix in this JS. 
The codestyle was also quite uncommon for PF.

Issue #3794, opened 2018-06-18
TreeTable: column width does not work


ravihariharan:
# IMPORTANT 
- !!! If you open an issue, fill every item. Otherwise the issue might be closed as invalid. !!!

- Please use the naming convention for the title:
  ${component}: ${title}
  Example:
  SelectOneMenu: Converter not called
- Before you open an issue, test it with the current/newest version.
- Try to find an explanation to your problem by yourself, by simply debugging. This will help us to solve your issue 10x faster
- Clone this repository https://github.com/primefaces/primefaces-test.git in order to reproduce your problem, you'll have better chance to receive an answer and a solution.
- Otherwise the example must be as small and simple as possible! It must be runnable without any other dependencies (like Spring,..., or project/company internal classes)!
- Feel free to provide a PR (Primefaces is an open-source project, any fixes or improvements are welcome.)

## 1) Environment
- PrimeFaces version: 6.2
- Does it work on the newest released PrimeFaces version? Version? No
- Does it work on the newest sources in GitHub? (Build by source ->No  https://github.com/primefaces/primefaces/wiki/Building-From-Source)
- Application server + version:Any
- Affected browsers:Any

## 2) Expected behavior
TreeTable p:column width should generate  `style="width:**px"`
...

## 3) Actual behavior

`<th class="ui-state-default ui-resizable-column" id="form:j_idt116:j_idt118" role="columnheader"><span class="ui-column-title">Name</span></th>`
..

## 4) Steps to reproduce

.. 

## 5) Sample XHTML
/showcase/src/main/webapp/ui/data/treetable/basic.xhtml

Example from showcase
 <p:column headerText="Name" width="50">
                    <h:outputText value="#{document.name}" />
                </p:column>
..

## 6) Sample bean

..

tandraschko:
probably never implemented, so -> enhancement

melloware:
@ravihariharan safe to close now?

PR #3798, opened 2018-06-19
TreeTable column enhancements

ravihariharan:
TreeTable: column aria-label does not work #3797
TreeTable: column width does not work #3794
TreeTable : Paginator Template issue #3801


ravihariharan:
Closed this PR since the Author and Committer had Corporate email id

Issue #3801, opened 2018-06-19
TreeTable : Paginator Template issue


ravihariharan:
# IMPORTANT 
- !!! If you open an issue, fill every item. Otherwise the issue might be closed as invalid. !!!

- Please use the naming convention for the title:
  ${component}: ${title}
  Example:
  SelectOneMenu: Converter not called
- Before you open an issue, test it with the current/newest version.
- Try to find an explanation to your problem by yourself, by simply debugging. This will help us to solve your issue 10x faster
- Clone this repository https://github.com/primefaces/primefaces-test.git in order to reproduce your problem, you'll have better chance to receive an answer and a solution.
- Otherwise the example must be as small and simple as possible! It must be runnable without any other dependencies (like Spring,..., or project/company internal classes)!
- Feel free to provide a PR (Primefaces is an open-source project, any fixes or improvements are welcome.)

## 1) Environment
- PrimeFaces version: 6.2
- Does it work on the newest released PrimeFaces version? Version? No
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) NO
- Application server + version: All
- Affected browsers: All

## 2) Expected behavior
As per user guide 6.2 - Paginator Template : Page 147
Also {CurrentPageReport} has it’s own template defined with currentPageReportTemplate option.
You can use {currentPage},{totalPages},{totalRecords},{startRecord},{endRecord} keyword
within currentPageReportTemplate. **Default is {currentPage} of{totalPages}.**

## 3) Actual behavior

The currentPageReportTemplate default is not set.  The workaround is to explicitly define 
`currentPageReportTemplate="{currentPage} of {totalPages}"`

> SEVERE: null
> java.lang.NullPointerException
> 	at org.primefaces.component.paginator.CurrentPageReportRenderer.render(CurrentPageReportRenderer.java:36)
> 	at org.primefaces.renderkit.DataRenderer.encodePaginatorMarkup(DataRenderer.java:112)
> 	at org.primefaces.component.treetable.TreeTableRenderer.encodeRegularMarkup(TreeTableRenderer.java:412)
> 	at org.primefaces.component.treetable.TreeTableRenderer.encodeMarkup(TreeTableRenderer.java:314)
> 	at org.primefaces.component.treetable.TreeTableRenderer.encodeEnd(TreeTableRenderer.java:214)
[...]

PR #3804, opened 2018-06-20
TreeTable enhancements & defect fix

ravihariharan:
TreeTable: column aria-label does not work #3797
TreeTable: column width does not work #3794
TreeTable : Paginator Template issue #3801


tandraschko:
thanks but next time please 1 PR for 1 issue

PR #3810, opened 2018-06-21
TreeTable : Paginator Position not working

ravihariharan:
Fix for Paginator Position


tandraschko:
Why not change the IF-statements to:
if "top" -> render("top")

instead of

if !="botom" -> render("top")?

ravihariharan:
The code take care of "both" default case.

also kept it in sync with [DataTableRenderer](https://github.com/primefaces/primefaces/blob/f6a0e54f4d11470bad319c6d21fd27f411015706/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java)

PR #3816, opened 2018-06-26
Fix #3817 - Datatable: provide widget helper functions for navigation and editing

### LINKED ISSUE #3817: Datatable: provide widget helper functions for navigation and editing
Feature request:

Improve navigation and editing possibilities for the javascript datatable object.

- rowUp and rowDown in single selection mode
- editSelectedRow, saveSelectedRow and cancelSelectedRow when one row is selected
- nextPage and prevPage

users may choose to bind these functions to a hotkey.

eduarddrenth:
fixes #3817 
- functions for navigating rows in single selection mode
- functions for editing when one row is selected
- functions nextPage/prevPage


tandraschko:
Thanks for your work but we don't apply changes without a issue with a description before.
Also please add a documentation for each method.

eduarddrenth:
ok...datatable.js does not contain any documentation, where do you want me to put the documentation?

tandraschko:
Just a method documentation

eduarddrenth:
done

tandraschko:
Review:
1) $('a.ui-paginator-next:first').click(); - it doesn't use the elements of the current DataTable, is just uses the first on the page. Thats wrong, same for previous.
2) Please add a empty line after each method and please also add a empty line between the method description and the @return

eduarddrenth:
Thanks! I'll work on it

eduarddrenth:
O, I'm sorry just added AjaxSessionTimeoutListener which I want to donate to this PR by accident, should have been a separate PR.

tandraschko:
I'm not 100% happy with some naming like "rowDown".
@Rapster @melloware  could you assist please to make it "nicer" from the cosmetic perspective? Also such stuff like new lines between methods.

Rapster:
What about selectPreviousRow/selectNextRow? Also, it only work if selection is radio based, but what about multiple mode? I think these methods should handle that case too

melloware:
@Rapster those were the exact Métis names I was thinking of.  Good point on the radio selection.

eduarddrenth:
So, what would be intuitive?

for single mode:

selectPreviousRow:
- select previous row
- select last row if currently selected row is the first row

selectNextRow
- select next row
- select first row if currently selected row is the last row

now for multiple mode some questions remain:

- We do know which rows are selected, but do we know which row is current?
- check wether next/prev isn't selected already
- selections may be discontinuous
- should next/previous add to the already selected? Or should the existing selection be cleared. I think clearing existing selection should be left to developers. They may also want to do things like Shift-uparrow.

eduarddrenth:
I noticed that I dumped the branche I created this PR in.

Would the functionality in this PR be welcome? Our people are very content with them because they enable keyboard navigation.

If not I'll close this PR and the issue

melloware:
I would say close this PR and resubmit a new one please with the updated names.

eduarddrenth:
Happy to do that, but this PR got stalled at the time because of an unfinished discussion about behaviour of row functions in multiple selection mode.

Suggestion:

- selectPreviousRow/selectNextRow: do nothing in multiple selection mode
- editSelectedRow/cancelSelectedRow: switch/cancel row editing of all selected rows in multiple selection mode. Perhaps rename function to ...Rows.

What do you think?

melloware:
I like those names and now the new build system makes you properly JSDoc them else the build fails so we know they will have quality documentation for each method!

eduarddrenth:
So you like the names in my last suggestion, but do you and others agree with the multiple selection mode functionality I describe?

melloware:
As for multiple not sure.  I don't think we know which row is current.

tandraschko:
i'm not sure in general - selection is quite complex
as long as we don't collect all cases and really design a good API here, i feel like we shouldnt add it

eduarddrenth:
The functions that are functionally clear at the moment:

- editSelectedRows / cancelSelectedRows
- nextPage / previousPage

The implementation of the edit/cancel functions shouldn't trigger too many events in multiple selection mode. This probably implies a redesign of events.

So, perhaps better skip this PR for now, until more people ask for fuctions like this.

melloware:
Agreed.

PR #3822, opened 2018-06-27
closes #48 - CSV delimiter may be specified

### LINKED ISSUE #48: DataExporter choice of the CSV Separator

cnsgithub:
**Sample XHTML:**
 ```xml
<p:dataExporter type="csv" target="tbl" fileName="cars" options="#{dataExporterView.csvOptions}" />
```

**Sample bean:**
```java
    public ExporterOptions getCsvOptions() {
        // return new CSVOptions('"', ';', "\n");
        // is equivalent to 
        return CSVOptions.EXCEL_NORTHERN_EUROPE;
    }
```

**Some notes:**
- predefined options in the style of [super-csv](https://github.com/super-csv/super-csv
): STANDARD, EXCEL, EXCEL_NORTHERN_EUROPE, TABS
- besides delimiter character also allow specification of quote character and end of line symbols (Windows -> CRLF, Unix -> LF, Mac -> CR) to be most flexible
- `Exporter `and `ExporterOptions `are not very well designed, resulting in massive use of `UnsupportedOperationException`
- tested before and after with binary comparison -> no differences

### DIFF
--- a/src/main/java/org/primefaces/component/export/CSVExporter.java
+++ b/src/main/java/org/primefaces/component/export/CSVExporter.java
@@ -33,6 +33,20 @@ import org.primefaces.util.Constants;
 
 public class CSVExporter extends Exporter {
 
+    private CSVOptions csvOptions;
+    
+    public CSVExporter(ExporterOptions options) {
+        csvOptions = CSVOptions.EXCEL;
+        if (options != null) {
+            if (options instanceof CSVOptions) {
+                csvOptions = (CSVOptions) options;
+            }
+            else {
+                throw new IllegalArgumentException("Options must be an instance of CSVOptions.");
+            }
+        }
+    }
+    
     @Override
     public void export(FacesContext context, DataTable table, String filename, boolean pageOnly, boolean selectionOnly,
             String encodingType, MethodExpression preProcessor, MethodExpression postProcessor, ExporterOptions options,
@@ -96,7 +110,7 @@ public class CSVExporter extends Exporter {
 
             if (col.isRendered() && col.isExportable()) {
                 if (firstCellWritten) {
-                    builder.append(",");
+                    builder.append(csvOptions.getDelimiterChar());
                 }
 
                 UIComponent facet = col.getFacet(columnType.facet());
@@ -127,7 +141,7 @@ public class CSVExporter extends Exporter {
             }
         }
 
-        builder.append("\n");
+        builder.append(csvOptions.getEndOfLineSymbols());
     }
 
     @Override
@@ -142,7 +156,7 @@ public class CSVExporter extends Exporter {
 
             if (col.isRendered() && col.isExportable()) {
                 if (firstCellWritten) {
-                    builder.append(",");
+                    builder.append(csvOptions.getDelimiterChar());
                 }
 
                 try {
@@ -172,7 +186,7 @@ public class CSVExporter extends Exporter {
             addColumnValue(builder, col.getChildren(), col);
 
             if (iterator.hasNext()) {
-                builder.append(",");
+                builder.append(csvOptions.getDelimiterChar());
             }
         }
     }
@@ -184,20 +198,20 @@ public class CSVExporter extends Exporter {
     }
 
     protected void addColumnValue(StringBuilder builder, String value) throws IOException {
-        value = (value == null) ? "" : value.replaceAll("\"", "\"\"");
+        value = (value == null) ? "" : value.replace(csvOptions.getQuoteString(), csvOptions.getDoubleQuoteString());
 
-        builder.append("\"" + value + "\"");
+        builder.append(csvOptions.getQuoteChar()).append(value).append(csvOptions.getQuoteChar());
     }
 
     protected void addColumnValue(StringBuilder builder, List<UIComponent> components, UIColumn column) throws IOException {
         FacesContext context = FacesContext.getCurrentInstance();
 
-        builder.append("\"");
+        builder.append(csvOptions.getQuoteChar());
 
         if (column.getExportFunction() != null) {
             String value = exportColumnByFunction(context, column);
             //escape double quotes
-            value = value == null ? "" : value.replaceAll("\"", "\"\"");
+            value = value == null ? "" : value.replace(csvOptions.getQuoteString(), csvOptions.getDoubleQuoteString());
 
             builder.append(value);
         }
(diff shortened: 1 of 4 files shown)


src/main/java/org/primefaces/component/export/CSVExporter.java
```diff
@@ -184,20 +198,20 @@ protected void addColumnValue(StringBuilder builder, UIComponent component) thro
     }
 
     protected void addColumnValue(StringBuilder builder, String value) throws IOException {
-        value = (value == null) ? "" : value.replaceAll("\"", "\"\"");
+        value = (value == null) ? "" : value.replace(csvOptions.getQuoteString(), csvOptions.getDoubleQuoteString());
```
melloware, line 202: Just curious why you chose replace instead of the original replaceAll to find and escape all the quotes in the string?

cnsgithub: Hi @melloware,

Thanks for reviewing my commit.

I chose `replace` (which also replaces all occurrences) because of its obviously better performance. `replaceAll` uses regex matching. Also since quote character may be specified dynamically now, I would have to use `java.util.regex.Matcher.quoteReplacement` to avoid `PatternSyntaxException` otherwise.

__Benchmark__
```java
	public static void main(String[] args) {
		final String str = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789\"9876543210ZYXWVUTSRQPONMLKJIHGFEDCBA";
		final String quote = "\"", doubleQuote = "\"\"";
		final int repeats = 100_000_000;
		
		long start = System.currentTimeMillis();
		for (int i = 0; i < repeats; i++) {
			str.replaceAll(quote, doubleQuote);
		}
		System.out.println(String.format("replaceAll took %d ms", System.currentTimeMillis() - start));

		start = System.currentTimeMillis();
		for (int i = 0; i < repeats; i++) {
			str.replace(quote, doubleQuote);
		}
		System.out.println(String.format("replace took %d ms", System.currentTimeMillis() - start));
	}
```

```
replaceAll took 74971 ms
replace took 8814 ms
```

Cheers

melloware: fantastic answer.  I support this 100%.

jxmai: Interesting, I had the exact opposite result running the benchmark snippet above.

```java
replaceAll took 56330 ms
replace took 127743 ms
```
```
Linux 4.4.0-128-generic
openjdk version "1.8.0_171"
```


To be fair, you can paste the snippet to some online Java IDE such as https://www.compilejava.net/
(There is a 5 seconds time limit so you can just reduce the number of repeats :eyes: )
```java
public class Test {    
    public static void main(String[] args) {
		final String str = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789\"9876543210ZYXWVUTSRQPONMLKJIHGFEDCBA";
		final String quote = "\"", doubleQuote = "\"\"";
		final int repeats = 1_000_000;
		
		long start = System.currentTimeMillis();
		for (int i = 0; i < repeats; i++) {
			str.replaceAll(quote, doubleQuote);
		}
		System.out.println(String.format("replaceAll took %d ms", System.currentTimeMillis() - start));

		start = System.currentTimeMillis();
		for (int i = 0; i < repeats; i++) {
			str.replace(quote, doubleQuote);
		}
		System.out.println(String.format("replace took %d ms", System.currentTimeMillis() - start));
	} 
}
```

snapshot1

cnsgithub: @jxmai Probably there are differences in JVM/OS/CPU setup.

My setup: Windows 8.1, Intel Core i7-3770 @ 3.4GHz (4 cores, 8 logical processors)

**Oracle JDK 9.0.4, 100e6 repetitions**
```
replaceAll took 64815 ms
replace took 7111 ms
```

**OpenJDK 10.0.1, 100e6 repetitions**
```
replaceAll took 71586 ms
replace took 8083 ms
```

**OpenJDK 10.0.1, 1e6 repetitions, without JIT compiling (-Djava.compiler=NONE)**
```
replaceAll took 47183 ms
replace took 18953 ms
```

The JVM (neither Oracle JDK nor OpenJDK) does not seem to use more than one processors  on Windows:
image

cnsgithub: **OpenJDK 9, Debian, Intel Xeon X5660 @ 2,8GHz (4 cores), 100e6 repetitions**
```
replaceAll took 52079 ms
replace took 12339 ms
```

cnsgithub: Solution of the mysterious behavior: @jxmai was using JDK 8 whereas @cnsgithub was using JDK 9/10.

**JDK 8**
```java
    public String replace(CharSequence target, CharSequence replacement) {
        return Pattern.compile(target.toString(), Pattern.LITERAL).matcher(
                this).replaceAll(Matcher.quoteReplacement(replacement.toString()));
    }
```

**JDK 9/10**
```java
    public String replace(CharSequence target, CharSequence replacement) {
        String tgtStr = target.toString();
        String replStr = replacement.toString();
        int j = indexOf(tgtStr);
        if (j < 0) {
            return this;
        }
        int tgtLen = tgtStr.length();
        int tgtLen1 = Math.max(tgtLen, 1);
        int thisLen = length();

        int newLenHint = thisLen - tgtLen + replStr.length();
        if (newLenHint < 0) {
            throw new OutOfMemoryError();
        }
        StringBuilder sb = new StringBuilder(newLenHint);
        int i = 0;
        do {
            sb.append(this, i, j).append(replStr);
            i = j + tgtLen;
        } while (j < thisLen && (j = indexOf(tgtStr, j + tgtLen1)) > 0);
        return sb.append(this, i, thisLen).toString();
    }
```

**Conclusion**
JDK ≥ 9 users profit from using `String.replace`. For JDK < 9 it should be irrelevant if we use `String.replace` or `String.replaceAll`.

melloware: Yep I am using JDK8 also so that makes perfect sense.    Good to see they finally offered an optimized version of replace in JDK 9+!

tandraschko:
Thanks!

Issue #3615, opened 2018-04-25
DefaultCommand: Problem with keypad comma with ABNT2


kemperalves:
## 1) Environment

- PrimeFaces version: Running PrimeFaces-6.2.3 on Mojarra-2.3.2.
- Keyboard Model: Microsoft Wireless Keyboard 800
- Keyboard Layout: Brazilian ABNT 2 https://ibb.co/kBQMOc 

ps - Tested with others anbt 2 keyboard layouts with the comma in the same position and the problem occurs too

## 2) Expected behavior

User type a comma in the keypad and the comma need to be "typed" in the component.

## 3) Actual behavior

When typing data in a form and this form has a defaultCommand, if I press the comma key in the numeric keypad inside at any input component, the defaultCommand is actioned

## 4) Steps to reproduce

Go to this page on primefaces showcase, using an keyboard with the keypad, press the comma button and the form will be submitted...

https://www.primefaces.org/showcase/ui/misc/defaultCommand.xhtml

## 5) Sample XHTML

No need, its possible to simulate in the primefaces showcase

## 6) Sample bean

No need, its possible to simulate in the primefaces showcase

lagar84:
I am from Brazil too, and I can also reproduce the problem. But in ABNT2 layout, above numpad enter is a 'period', or 'dot', not a comma. Comma is on the left side, and does not trigger defaultCommand. The key above "keypad enter" is indeed triggering defaultCommand, and obviously should not. From the picture, is the key between the 'plus' (+) key and 'enter'.

(my keyboard is a generic mtek abnt2)

Edit: event.which report it as keycode 108

Workaround: put this in your template or xhtml file: 
```
<script type="text/javascript">
//<![CDATA[
  try{
	$.ui.keyCode.NUMPAD_ENTER = 13; //fix primefaces defaultCommand
  }catch(e){}
//]]>
</script>
```

cnsgithub:
Sorry, I am not able to reproduce this issue, even after switching to Brazil ABNT2 keyboard layout.

kemperalves:
This is the key in my keyboard thas are submitting the form with the defaultCommand

![whatsapp image 2018-04-26 at 14 00 46](https://user-images.githubusercontent.com/27026273/39320465-9572b58c-495a-11e8-87d1-bdc1b280c620.jpeg)

lagar84:
In file
[...]

Issue #2448, opened 2017-06-16
p:defaultCommand triggered by browser password managers due to missing keyCode.NUMPAD_ENTER declaration


mjiderhamn:
## 1) Environment
- PrimeFaces version: 5.0+
- Does it work on the newest released PrimeFaces version? No (determined from source code)
- Does it work on the newest sources in GitHub? No (5162cf1bf281cacebd665ae5922fc5a0b1469406)
- Application server + version: Resin (irrelevant)
- Affected browsers: Chrome 59, Safari 9

## 2) Expected behavior
`<p:defaultCommand />` is supposed to be triggered when the user presses Enter on their keyboard.

## 3) Actual behavior
As of Chrome 59 (59.0.3071.86) and Safari 9 (conditions for OS and SSL have not been thoroughly investigated), if the login credentials is stored in the browsers password manager, the `<p:defaultCommand />` is triggered as soon as the username has been filled - i.e. before the password has been filled. In a typical case this causes the login form to be posted resulting in a failed login attempt. In case the page with the failed login contains the username and password fields, this may - as in our case - cause the page to reload over and over and over, effectively hindering the user from login in until stored password is removed from browsers password manager.

## 4) Steps to reproduce
Create login page with `<p:defaultCommand />`
Using Chrome 59+, log in and accept suggestion to store password.
Log out and go to login page.

## 5) Sample XHTML
```xml
<p:inputText .../>
<p:password .../>
<p:defaultCommand .../>
```

## Cause
Upon investigation, it turns out that the cause of this is the fact that `$.ui.keyCode.NUMPAD_ENTER` is no longer defined. In 4.0 it was declared in [`jquery-plugins.js`](https://github.com/primefaces/primefaces/blob/4_0/src/main/resources/META-INF/resources/primefaces/jquery/jquery-plugins.js#L35).
`$.ui.keyCode.ENTER` is still declared in [`keyboard.js`](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/keyboard/keyboard.js#L87).

[...]

PR #3823, opened 2018-06-27
Fix #3453: bind defaultCommand to keypress and rely on e.which

### LINKED ISSUE #3453: defaultCommand: with browser autocomplete not working together in Firefox

### LINKED ISSUE #3454: Confirm: add precondition to dialog opening
Related to this topic https://forum.primefaces.org/viewtopic.php?f=3&t=40408

It would be really nice to have some precondition before confirmDialog might open. This could be a new attribute set on p:confirm. My use case is the following:

I have a datatable, and a button allowing to delete the selection. Before the dialog shows up to confirm the removal, a check should be applied to make sure items have been selected (in such case a message could be displayed and dialog not showing up)

david0:
This fixes #3453 by binding defaultCommand to keypress.
Numpad enter is never reported as `e.which == keyCode.NUMPAD_ENTER` ([details](https://gist.github.com/david0/36a6b3cf0d07dd1fe1296fcfbe8527ba)). Instead on **keypress** `l` is repored as `108` (but not on keydown). Because of this, its needed drop numpad too.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/forms/forms.defaultcommand.js
+++ b/src/main/resources/META-INF/resources/primefaces/forms/forms.defaultcommand.js
@@ -17,9 +17,9 @@ PrimeFaces.widget.DefaultCommand = PrimeFaces.widget.BaseWidget.extend({
         }
 
         //attach keypress listener to parent form
-        this.jqTarget.closest('form').off('keydown.' + this.id).on('keydown.' + this.id, function(e) {
+        this.jqTarget.closest('form').off('keypress.' + this.id).on('keypress.' + this.id, function(e) {
            var keyCode = $.ui.keyCode;
-           if(e.which == keyCode.ENTER || e.which == keyCode.NUMPAD_ENTER) {
+           if(e.which == keyCode.ENTER) {
                 //do not proceed if event target is not in this scope or target is a textarea,button or link
                 if (($this.scope && $this.scope[0] != e.target && $this.scope.find(e.target).length == 0)
                    || $(e.target).is('textarea,button,input[type="submit"],a')) {


Rapster:
Wrong issue number I think

jxmai:
:+1:  For removing `NUMPAD_ENTER`, but it should be related to https://github.com/primefaces/primefaces/issues/3453, https://github.com/primefaces/primefaces/pull/3455 and https://github.com/primefaces/primefaces/issues/3615.

melloware (approved):

cnsgithub (approved):

melloware:
Just searching through the source for *.js containing NUMPAD_ENTER found a LOT of locations.  Are we saying NUMPAD_ENTER is not valid anymore and it should be removed from all the places in the code?

jxmai:
Indeed, there are still 61 results from 30 js files by searching `NUMPAD_ENTER`. Since the core problem is on the `NUMPAD_ENTER` which is not really representing `NUMPAD_ENTER`, they should be removed as well.

melloware:
here is the original issue why NUMPAD_ENTER was re-added.  Can you guys please review and confirm its still OK to remove.

https://github.com/primefaces/primefaces/issues/2448

david0:
#2448 has [re-added NUMPAD_ENTER to the keyCode enumeration](https://github.com/primefaces/primefaces/commit/399c8d35c22d98eb65127e1be0b7c65419c95617) (not to the condition itself) just to avoid that this part of the condition is fullfilled when it should not be. So removing is still good for 2448.

> Now, when the browsers password manager of the above mentioned browsers fills the form, it seems to send a keydown event with e.which = undefined. And since $.ui.keyCode.NUMPAD_ENTER also is undefined, it means that any test checking for e.which == keyCode.NUMPAD_ENTER - such as the one for defaultCommand - results in true.

melloware:
Can you guys review my change I made this fix along with removing NUMPAD_ENTER all in one PR.

https://github.com/primefaces/primefaces/pull/3844

melloware:
@david0 You can close this PR as mine was successfully merged.  Thanks for all your help!

PR #3832, opened 2018-06-29
fixes #3825 - erroneous use of escapeText on calls to WidgetBuilder.attr

### LINKED ISSUE #3825: Erroneous use of escapeText on calls to WidgetBuilder.attr
## 1) Environment

## 2) Expected behavior
JS syntax errors must never occur, input must be redisplayed correctly.

## 3) Actual behavior
Double escaping (JS and JSON) may result in JS syntax errors or wrong display.

Since Dec, 28 2017 `WidgetBuilder.attr` escapes the value itself, which in my opinion is the correct approach. However, some of the older callers do the escaping themselves, e.g. in `PasswordRenderer`, lines 67 to 70:
```java
                    .attr("promptLabel", escapeText(password.getPromptLabel()))
                    .attr("weakLabel", escapeText(password.getWeakLabel()))
                    .attr("goodLabel", escapeText(password.getGoodLabel()))
                    .attr("strongLabel", escapeText(password.getStrongLabel()));
```

## 4) Steps to reproduce
Modify the [showcase ](http://localhost:8080/showcase/ui/input/password.xhtml) as shown below and enter just one character within the turkish feedback input. The forward slash of the `weakLabel `will be erroneously escaped and preceded with a backslash.

image

## 5) Sample XHTML
```xml
                <p:password id="turkishFeedback" value="#{passwordView.password3}" feedback="true" 
                            promptLabel="PROMPT" weakLabel="WEAK/"
                            goodLabel="GOOD" strongLabel="STRONG" />
```

## 6) Sample bean
Use the unmodified showcase bean.

## 7) Solution
Find all potential occurrences of `\.attr\(.*escapeText` regex and remove the `escapeText` calls. Ensure that the attributes are escaped correctly in the appropriate JS files.

cnsgithub:
double escaping


melloware (approved):

tandraschko:
I, personally, feel better to escape it on server as on client.
WDYT? Or doesn't it really matter?

cnsgithub:
On server side you cannot know how it will be used on client side, i.e. you don't know the context (JS, HTML or whatever). Anyway, it has to be escaped twice, one time on server side (for JS context) and one time on client side (mostly for HTML usage) cause of different contexts. The mistake was to double escape it on the server side and lack of escaping on client side.

tandraschko:
But what happens when you have a ' inside a Label? The js crashes normally

cnsgithub:
You mean the attribute name? I didn't change the escaping (i.e. no escaping at all) of the attribute name. I think this is an edge case because `WidgetBuilder`'s attribute names are given by the framework and would never include quotes? Attribute values are of importance however, since user-supplied input goes there. And those values are escaped by `ComponentUtils.escapeEcmaScriptText`, i.e. `'` will be replaced with `\'`.

tandraschko:
I was talking about attributes like "invalidFileSizeLabel" or similar.
But yep, we really do a escapeText, and then escapeEcmaScriptText.
What about the unicode escapes in escapeText? I didn't implement it, so no idea.

cnsgithub:
I didn't implement it either. I would merge https://github.com/primefaces/primefaces/pull/3525, then things will get clearer. :wink:

PR #3844, opened 2018-07-03
Fix #3453,#3455,#3615: Remove NUMPAD_ENTER.

### LINKED ISSUE #3453: defaultCommand: with browser autocomplete not working together in Firefox

melloware:
Fixes #3453,#3455,#3615, by removing NUMPAD_ENTER which was removed in Jquery 1.11 and restored for PF https://github.com/primefaces/primefaces/issues/2448 which it should not have been.

This PR rolls up all the changes in one.


tandraschko:
So keypad enter doesn't work anymore with this PR, when you e.g. would like to "select "a autocomplete item?

cnsgithub:
I think it should not stop working.

melloware:
So it is my understanding that NUMPAD_ENTER is not always 108 depedning on the browser and should not be relied upon.  @jxmai @david0 please chime in?

Here is a great chart @david0 put together showing the exact problem:  https://gist.github.com/david0/36a6b3cf0d07dd1fe1296fcfbe8527ba

david0:
Yes, I actually never saw numpad enter reported as 108. It was always reported as regular enter (13).

Maybe it was different with some very old browsers, but at last this change will fix the problems for all modern browsers.

jxmai:
Yes, 108 does not mean numpad enter at all. I also wrote similar post on https://github.com/primefaces/primefaces/pull/3455#issuecomment-386097198.

---

and yes, the keypad enter's keycode is 13, which is the same as Enter key.

PR #3853, opened 2018-07-05
Fix #3255: Keyboard update to plugin 2.1.0 JS version.

### LINKED ISSUE #3255: Keyboard: update plugin
http://keith-wood.name/keypad.html

Version should be updated from v.1.2.4 to v2.1.0

I've seen in the commit history that the component was broken after jquery update, and only the "broken" places were changed. Why not invest some more minutes and update the lib instead when this happens ;-)

melloware:
Updated to new plugin and tested the functionality and theming using the showcase and everything looks good.


tandraschko:
Would be great if you can also splitup our widget and the external js

melloware:
Will do.  I prefer it that way.

melloware:
OK separated the plugin from the widget and updated the pom.xml to aggregate them.

Issue #1529, opened 2016-06-16
FileUpload: auto upload bug


Karolis92:
If user selects many files to upload and during upload process presses `x` to remove file while it's not uploaded yet - any files after it are not uploaded and javascript errors occurre:

```
TypeError: g.row is null            fileupload.js.xhtml:1:24907
TypeError: a is null            fileupload.js.xhtml:1:29116
```

After this error upload doesn't work anymore.

I noticed, then auto upload is off - `x` buttons are disabled when upload begins, but not in auto upload mode.

`auto="false"`:
![auto-false](https://cloud.githubusercontent.com/assets/19680461/16116809/a1b290c8-33d7-11e6-948c-4e8a7957e451.JPG)
`auto="true"`:
![auto-true](https://cloud.githubusercontent.com/assets/19680461/16116810/a1ddd2d8-33d7-11e6-82fb-fe726eab9de1.JPG)

Nikoas83:
i have the same issue !!! Its really simple to reproduce, juste create fileUpload with multiple and auto options enabled, upload files and cancel one with X button : you can see in console js browser multiple errors and the real issue that is the upload continue in background.... Test on FF and CHROME last version and Primefaces 5.3

`<h:form enctype="multipart/form-data" >
                    <p:fileUpload id="fileUploaderId" fileUploadListener="#{actionManagerBean.uploadFile}"  mode="advanced" multiple="true" auto="true" previewWidth="40"
 /></h:form>`

Errors in console : 
TypeError: g.row is null                                      fileupl...s&v=5.3 (ligne 1, col. 24948)
TypeError: a.row is null                                      fileupl...s&v=5.3 (ligne 1, col. 29157)

kukel:
Tried PF 6 to?

Nikoas83:
Yes. You can test on showcase, in auto mode (this is just a little difficult cause file is limit to 100ko but after many attempts...) to reproduce bug.

Rapster:
Related to #844

Issue #845, opened 2015-11-03
FileUpload: allow to capture cancel event


german1410:
It would be nice to be able to capture when user cancel current file upload in order to take actions. For example if the page uses blockUI or spotlight while file is being uploaded there is no transparent way to close the blockUI or spotlight when user cancel the upload. 

User can always attach a listener to the cancel button using java script but I thing it would be nice to have an 'oncancel'  attribute in order to perform action when user cancel the upload.

Issue #844, opened 2015-11-03
FileUpload: java script error on cancel


german1410:
I'm seeing the following error on the browser debug console when I cancel a file upload:

Uncaught TypeError: Cannot read property 'fadeOut' of null

I'm using prime faces 5.2 with Mojarra 2.1.29-04
Browser: Chrome Version 46.0.2490.80 m
I'm able to reproduce this on prime faces showcase too

The bigger the file, the more errors I get.

german1410:
More details I was able to capture:
Uncaught TypeError: Cannot read property 'children' of null
ucfg.progress                  @ fileupload.js.jsf?ln=primefaces&v=5.2:1
y.Widget._trigger             @ jquery-plugins.js.jsf?ln=primefaces&v=5.2:27
b.widget._onProgress     @ fileupload.js.jsf?ln=primefaces&v=5.2:1
(anonymous function)     @ jquery-plugins.js.jsf?ln=primefaces&v=5.2:27
(anonymous function)     @ fileupload.js.jsf?ln=primefaces&v=5.2:1
bI.event.dispatch            @ jquery.js.jsf?ln=primefaces&v=5.2:25
ce.handle                        @ jquery.js.jsf?ln=primefaces&v=5.2:25

mertsincan:
I tested with Chrome 46.0.2490 but I couldn't replicate this issue(Exp; http://www.primefaces.org/showcase/ui/file/upload/single.xhtml) Can you please explain your case step by step? What should I do in fileupload demo to see this issue?

german1410:
This happens on the following page http://www.primefaces.org/showcase/ui/file/upload/auto.xhtml

sudheerj:
The upload process begins automatically very quickly in auto mode.When I tried to upload bigger size files then I can find different as follows.

![image](https://cloud.githubusercontent.com/assets/3127317/15091248/8ce39e94-1460-11e6-84df-8b96a8a7228a.png)

german1410:
1) Go to http://www.primefaces.org/showcase/ui/file/upload/auto.xhtml using chrome
2) Open the debug console
3) Upload a file, now limit is 100k there so it will be more difficult to reproduce.
4) Before the file is uploaded cancel the operation

![primefacesuploadcancelerror](https://cloud.githubusercontent.com/assets/4230387/15114422/4d8a35f2-15d0-11e6-86b8-c3bcb4db2444.png)

Rapster:
[...]

PR #3862, opened 2018-07-10
Add real file upload cancellation and oncancel callback

danielthegray:
- Fixes the upload that continues in the background (issues #1529, #844)
- Adds an oncancel callback function (#845)

### DIFF
--- a/src/main/java/org/primefaces/component/fileupload/FileUploadRenderer.java
+++ b/src/main/java/org/primefaces/component/fileupload/FileUploadRenderer.java
@@ -100,6 +100,7 @@ public class FileUploadRenderer extends CoreRenderer {
                     .nativeAttr("onAdd", fileUpload.getOnAdd())
                     .callback("onstart", "function()", fileUpload.getOnstart())
                     .callback("onerror", "function()", fileUpload.getOnerror())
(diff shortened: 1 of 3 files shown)


tandraschko:
Did you test it in all the common browsers?
Otherwise, the changes are looking good +1

Rapster:
Sounds awesome! Thanks a lot @thedanielgray

melloware (approved):

melloware:
This looks good to me too!

danielthegray:
I've tested it on IE11, Chrome (I'm on 67), Firefox (I tried it with the most recent Firefox Developer Edition as well), and Opera.

The main function call added is the `abort()` method of the `XMLHttpRequest`, which is supported widely across browsers (even since IE7: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/abort#Browser_compatibility) so I think it should behave stably across all browsers.

The `onCancel` callback I implemented here also works on all the mentioned browsers. (IE11, Chrome, Firefox and Opera)
**If it doesn't work on your browser, let me know and I will look into it!**

tandraschko:
Thanks!

PR #3870, opened 2018-07-13
Add optional package dependencies for dynamically loaded classes (such as classes loaded via Class.forName()) since their imports are not automatically added

stiemannkj1:


tandraschko:
@stiemannkj1 thanks kyle. Could you please also create a issue for it? We use it for tracking and the changelist.

stiemannkj1:
@tandraschko, done: https://github.com/primefaces/primefaces/issues/3874

Issue #3854, opened 2018-07-05
MOVE_SCRIPTS_TO_BOTTOM combines non javascript


mulkgun:
In my project I have Web GL script<script type="x-shader/x-vertex"> and <script type="x-shader/x-fragment"> MOVE_SCRIPTS_TO_BOTTOM combines them into javascript block. I think it should group them by their type.

cnsgithub:
Might be a good point, but please use the issue template, otherwise chance is low that someone will hear you.

tandraschko:
Not sure If this is required as they are just collected by order. So you can just group them by yourself.

cnsgithub:
```html
<script type="text/javascript">
   console.log('foo');
</script>
PAGE CONTENT
<script type="x-shader/x-vertex">
   //non-javascript code
</script>
```

Will be transformed to 

```html
PAGE CONTENT
<script>
   console.log('foo');
   //non-javascript code
</script>
```

or

```html
PAGE CONTENT
<script type="text/javascript">
   console.log('foo');
</script>
<script type="x-shader/x-vertex">
   //non-javascript code
</script>
```

?

cnsgithub:
Just tried it, the answer is that it results in

```html
PAGE CONTENT
<script type="text/javascript">
   console.log('foo');
   //non-javascript code
</script>
```
which is of course incorrect.

tandraschko:
should we even touch them?

cnsgithub:
I would consequently do so. 

But I am not sure how we can do this in `MoveScriptToBottomResponseWriter` since it's a bit complex. @tandraschko You wrote it?

tandraschko:
yup

cnsgithub:
Btw, there might be other attributes for the `script `tag that must be taken into account, e.g. `async`, `defer`, `nonce`...

And can we turn `"script".equals` into `"script".equalsIgnoreCase`?

tandraschko:
not sure if we should take care of every case, it's makes to code probably to complex.

cnsgithub:
We could just check for attributes besides `type` and if so exclude them from being combined?

tandraschko:
I think also merging them is no problem. But adding support for async, defer, nonce, etc. seems quite complex.
Is it always correct to render all scripts merged in following order:
1) javascript
2) x-shader/x-vertex
3) x-shader/x-fragment

i wonder if its sometimes required that a fragment must be placed before vertex and vice versa.
[...]

PR #3872, opened 2018-07-16
add: some unit tests for MoveScriptsToBottomResponseWriter (#3854)

cnsgithub:
@tandraschko Instead of ignoring failing tests I added some FIXME markers. I hope that's okay for you.

Test coverage of methods containing most complex logic, i.e. `startElement`, `endElement`, `writeAttribute`, `mergeAndMinimizeInlineScripts `is 100%.

Some issues/ideas (see FIXMEs also):
- tag handling should be case-insensitive (e.g. `script `vs. `SCRIPT`)
- `savedInlineTags `statistics is wrong, should start with 1 if two inline tags are merged into one
- if `PrimeFaces` is not used at all minimization will blow up the code by adding `var pf=window.PrimeFaces` :wink: 
- `writeText `should be used in place of `write `(possible HTML injection)
- consider different script types (as the issue says)
- pass through attributes that had been present before (consider merging them correctly)
- consider further optimization by dropping duplicate included scripts
- do not write redundant script tag for inline scripts if there are not any inline scripts at all

### DIFF
--- a/src/test/java/org/primefaces/application/resource/MoveScriptsToBottomResponseWriterTest.java
+++ b/src/test/java/org/primefaces/application/resource/MoveScriptsToBottomResponseWriterTest.java
@@ -1,0 +1,206 @@
+package org.primefaces.application.resource;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.InOrder;
+
+import javax.faces.context.ResponseWriter;
+import java.io.IOException;
+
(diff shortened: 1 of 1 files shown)


tandraschko:
nice, thanks!
will check it later

melloware:
Nice work @cnsgithub !

tandraschko:
@cnsgithub 
done
we should clarify if we should use writeText vs write

And i'm not sure about the case with asnyc and deffered
It will be very hard or impossible to support this with all cases

Issue #3519, opened 2018-03-23
Prevent client-side changing of the datatable rows parameter (also affects other pageable components)


kukel:
## 1) Environment
- PrimeFaces version: up to latest source
- Application server + version: N/A
- Affected browsers: N/A

## 2) Expected behavior
Client side the rows value of the datatable should not be prone to manipulation

## 3) Actual behavior
You can change it client-side and with large datasets run into problems

## 4) Steps to reproduce
Manipulate the rows per page value on the client via a browser developer tool

## 5) Sample XHTML
Can be done in the online showcase

## 6) Sample bean
Can be done in the online showcase

The really cool [OptimusFaces ](https://github.com/omnifaces/optimusfaces) by the renowned JSF/OmniFaces  BalusC [has a fix for this](https://github.com/omnifaces/optimusfaces/blob/develop/src/main/java/org/omnifaces/optimusfaces/renderer/ExtendedDataTableRenderer.java) by overriding the decode method and in it do

    String rowsParam = getRequestParameter(context, component.getClientId() + "_rows");

	if (rowsParam != null && !asList(((DataTable) component).getRowsPerPageTemplate().split("[,\\s]+")).contains(rowsParam)) {
		throw new IllegalArgumentException("Unsupported rows per page value: " + rowsParam); // Prevent hackers from being able to exceed this.
	}

Other pageable components like https://www.primefaces.org/showcase/ui/data/dataGrid.xhtml are affected to

cnsgithub:
The following components should be affected:
- `dataGrid`
- `treeTable`
- `dataTable`
- `dataList`
- `dataView`

cnsgithub:
PR: https://github.com/primefaces/primefaces/pull/3843

stiemannkj1:
@kukel, @cnsgithub I've sent a PR that changes this code a bit in order to validate the rows per page even when `rowsPerPageTemplate` is `null`. I tested the changes thoroughly, but you still may want to review my changes since this is a security related issue: https://github.com/primefaces/primefaces/pull/3885

PR #3880, opened 2018-07-17

stiemannkj1:


melloware (approved):

tandraschko:
What are the changes in UIData?
They look ok but if it's a bugfix, we need a ticket plz :P

stiemannkj1:
@tandraschko, with UIData, I ran into an issue with the fix for #3519. It seems like that issue should be reopened since my fix is basically improving the original fix for that ticket. But If you'd rather I open a new ticket, I will. Let me know.

tandraschko:
Thats OK, Just a seperate PR would be great

stiemannkj1:
@tandraschko, done:

- https://github.com/primefaces/primefaces/pull/3885
- https://github.com/primefaces/primefaces/pull/3886

Issue #3505, opened 2018-03-21
Evaluate content-security-policy for inline scripts


tandraschko:
https://developers.google.com/web/fundamentals/security/csp/#if_you_absolutely_must_use_it

cnsgithub:
What do you plan to use `Content-Security-Policy` for?

tandraschko:
as the header says, for the inline-scripts

cnsgithub:
Moving it to external JS would be the recommended way for a number of reasons. This is not possible in PF?

tandraschko:
no

cnsgithub:
Ok, I accept your taciturnity. :wink: 

I just thought discussing it more broadly would finally result in a better implementation...

mertsincan:
@tandraschko, I researched it for a PRO User. Unfortunately, it is technically not possible in PF. Inline events used in HTML element won't work. Please see;  https://stackoverflow.com/questions/31060722/cordova-refuse-to-execute-inline-event-handler-because-it-violates-the-followi

My test code;
This code won't work. It throws an exception on console related to CSP
```
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml"
      xmlns:ui="http://java.sun.com/jsf/facelets"
      xmlns:f="http://java.sun.com/jsf/core"
      xmlns:p="http://primefaces.org/ui"
      xmlns:h="http://java.sun.com/jsf/html">

    <h:head>   
        <title>PrimeFaces Test</title>
        <meta http-equiv="Content-Security-Policy" content="script-src 'nonce-123456789'"></meta>
        <script nonce="123456789" type="text/javascript">
            function clickHandler() {
                 alert('clickHandler');
          }
        </script>
    </h:head>
    <h:body>
        <button onclick="clickHandler()">Click Button!</button>
    </h:body>
</html>
```

Please try this code instead of the above code;
```
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml"
      xmlns:ui="http://java.sun.com/jsf/facelets"
      xmlns:f="http://java.sun.com/jsf/core"
      xmlns:p="http://primefaces.org/ui"
      xmlns:h="http://java.sun.com/jsf/html">

    <h:head>   
        <title>PrimeFaces Test</title>
        <meta http-equiv="Content-Security-Policy" content="script-src 'nonce-123456789'"></meta>
        <script nonce="123456789" src="https://code.jquery.com/jquery-3.2.1.js" />
[...]

PR #3884, opened 2018-07-18
add: experimental, basic support for Content-Security-Policy scripts (#3505)

cnsgithub:
Hey guys,

As a proof of concept I implemented basic support for `Content-Security-Policy `scripts. Do not worry, it's just experimental and of course disabled by default. I tested some rudimentary things in the showcase and that worked for me. Please note, that support for `partial-response` is not yet implemented.

I would like you to give it a try by just adding the following to the showcase's `web.xml` - feedback is very welcome:

```xml
    <context-param>
        <param-name>primefaces.csp.ENABLED</param-name>
    </context-param>
    <context-param>
        <param-name>primefaces.csp.SUPPORTED_DIRECTIVES</param-name>
        <param-value>script-src</param-value>
    </context-param>
```

When looking at the browser console you will notice something like that:

`Refused to load the script 'https://www.google-analytics.com/analytics.js' because it violates the following Content Security Policy directive: "script-src 'nonce-OWNiNjI0ZWEtOWM2My00MDI4LTk2NzYtYWM2ZTQ5MTk0OGEz' 'nonce-Y2Q2NTAxZGQtODYzNi00MGJiLTg5ZGItYmQ2ZDljNzFlZDJl' 'nonce-NTk5NzEzMjgtN2E3OS00MDlmLTg5ZTgtZTI3YTY3MjMyOGE0' 'nonce-ZjM5NzQ0MTEtODk1Mi00OWE2LTlmZTgtNjk0OWY4YWRlOGNh' 'nonce-YzFhNDZlNGMtYzI0NS00OGVlLWJmNTctYTZkMmJiZWZlMTIw' 'nonce-MGMzMDg0MzgtZWI1Yy00ZjJjLTg1OTgtODY1OTgwMGU0ZDQ0' 'nonce-MjEyMWViMzktZDY1Yy00ZmQ0LTg1NzMtMDkyNTFjMTg4NzM0' 'nonce-ZDRmMjA5ODEtNWU3Ni00YjgzLWEwNTctMmEwYmViZGRkYWY2' 'nonce-ZDI4NGY1NjYtY2RkOC00MDUyLWI4YTUtZTFhZDcxMzJkOGMw'".`

When looking at the jetty console you will notice that the CSP violation has been reported:

`	WARNING: CSP violation reported by browser: {"csp-report":{"referrer":"","status-code":200,"disposition":"enforce","script-sample":"","violated-directive":"script-src","original-policy":"script-src 'nonce-MWYyNjlhMzgtNjgyYy00MTlhLTg2NGQtOWVlNjJkM2NjNmM3' 'nonce-NmJhZGI2YzktZGYwZC00ZjZhLTlkMjctZTk1YTZjMTI4YWUy' 'nonce-MzMzOTY3ZmMtYzJjYS00ZGFiLThjY2UtZjUxNTk0ZWQ5NmYz' 'nonce-Y2I0YTE4NDktZTk5OS00ZWUzLTkxYTQtZWIzMTU2NTI5NTNj' 'nonce-OTZhMmFlZDEtZWU0NC00NmQyLThhZDgtYWU3NjUzNzQxNDcw' 'nonce-MTZmYmUzNDUtNmY3NC00OWE2LWEzMmQtZWI5MTA1ZjRhNDlh' 'nonce-ZmM5NWNmMjctNWI1ZC00MWI2LWEyMjYtNTExODg3Y2ZlZTZi' 'nonce-YjM1NTM4NzEtZWVhZC00NzdlLTk2YmMtZjcwNDc1YjI5NWMy' 'nonce-ZWMxNTA0MmYtMjdlNy00MTNiLTlmMWMtM2M2MTI0Mzc3ODg5'; report-uri /showcase/csp-report;","source-file":"http://localhost:8080/showcase/","column-number":34,"document-uri":"http://localhost:8080/showcase/","line-number":511,"effective-directive":"script-src","blocked-uri":"https://www.google-analytics.com/analytics.js"}}`

Then just go back to your `web.xml` and add these lines:

```xml
    <context-param>
        <param-name>primefaces.csp.ALLOWED_HOSTS</param-name>
        <param-value>https://www.google-analytics.com</param-value>
    </context-param>
```

Your browser will be happy again.

When inspecting the DOM you will notice the nonces that have been added to all `script` elements. Furthermore you will not find any `onclick` etc. attributes anymore. Those are all merged into the last `script` block:

```xml
<script nonce="">var pf=PrimeFaces;
pf.csp1=function(id,evt,js){document.getElementById(id).addEventListener(evt,js);}
pf.csp2=function(id,attr,jsUri){document.getElementById(id).setAttribute(attr,jsUri);}
pf.csp1("SubMenu-Ajax","click",function(){Showcase.openSubMenu(this);});
pf.csp1("SubMenu-Input","click",function(){Showcase.openSubMenu(this);});
pf.csp1("SubMenu-Button","click",function(){Showcase.openSubMenu(this);});
pf.csp1("SubMenu-Data","click",function(){Showcase.openSubMenu(this);});
pf.csp1("SubMenu-Panel","click",function(){Showcase.openSubMenu(this);});
pf.csp1("SubMenu-Overlay","click",function(){Showcase.openSubMenu(this);});
pf.csp1("SubMenu-Menu","click",function(){Showcase.openSubMenu(this);});
pf.csp1("SubMenu-Charts","click",function(){Showcase.openSubMenu(this);});
pf.csp1("SubMenu-Message","click",function(){Showcase.openSubMenu(this);});
pf.csp1("SubMenu-Multimedia","click",function(){Showcase.openSubMenu(this);});
pf.csp1("SubMenu-File","click",function(){Showcase.openSubMenu(this);});
pf.csp1("SubMenu-DragDrop","click",function(){Showcase.openSubMenu(this);});
pf.csp1("SubMenu-CSV","click",function(){Showcase.openSubMenu(this);});
pf.csp1("SubMenu-DF","click",function(){Showcase.openSubMenu(this);});
pf.csp1("SubMenu-Misc","click",function(){Showcase.openSubMenu(this);});
</script>
```

Please note that in contrast to Firefox, Chrome will not show the nonces. But they are in place, check this by inspecting the headers:

image

__Some TODOs__

- [x] support `partial-response`
- [ ] support `partial-response eval`
- [x] support hashes as well
- [x] provide JUnit tests
- [ ] find better solution than increasing the buffer size
- [ ] test, test, test... :wink: 
- [x] provide showcase
- [ ] get PrimeFaces team to integrate it... :tongue: 

And as I said, I would be glad to get some feedback...


src/main/java/org/primefaces/application/resource/csp/Constants.java
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.primefaces.application.resource.csp;
+
+/**
+ * Some constants for Content-Security-Policy headers and directives
+ */
+public enum Constants {
+
```
melloware: Sadly I don't prefer Constants as name, give the Enum a real name like CspHeader.java and CspDirective.java.  I find it makes the code more readable and prevents common name like Constants.java

src/main/java/org/primefaces/application/resource/csp/ContentSecurityPolicyFilter.java
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * This servlet filter is responsible for setting appropriate Content-Security-Policy (CSP) response headers that are either whitelisted
+ * via {@link Constants.ContextParams#CONTENT_SECURITY_POLICY_HOST_WHITELIST} or determined automatically.
+ * Currently the <code>script-src</code> directive is supported only. It is planned to support other directives in the future as well
+ * that can then be specified via {@link Constants.ContextParams#CONTENT_SECURITY_POLICY_SUPPORTED_DIRECTIVES}.
+ * Per default browsers are instructed to report any CSP violations to the {@link ContentSecurityPolicyReportServlet} endpoint if not
+ * provided via {@link Constants.ContextParams#CONTENT_SECURITY_POLICY_REPORT_URI}.
+ */
+@WebFilter("/*")
```
melloware: Does @WebFilter imply Servlet 3.0?  Would it work on Servlet 2.5 still?

src/main/java/org/primefaces/application/resource/csp/ContentSecurityPolicyReportServlet.java
```diff
@@ -0,0 +1,76 @@
...
+import java.io.InputStreamReader;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Built-in default endpoint browsers may call via POST if Content-Security-Policy (CSP) violations occur. We just do some logging.
+ * To provide your own endpoint just change {@link Constants.ContextParams#CONTENT_SECURITY_POLICY_REPORT_URI} accordingly.
+ * Data must be valid JSON as specified in <a href="https://w3c.github.io/webappsec-csp/2/#violation-reports">Content Security Policy Level 2 - Reporting</a>.
+ */
+@WebServlet(ContentSecurityPolicyReportServlet.URL)
```
melloware: Same is @WebServlet only Servlet 3.0 and higher?

src/main/java/org/primefaces/application/resource/csp/ContentSecurityPolicyReportServlet.java
```diff
@@ -0,0 +1,76 @@
...
+            if (LOG.isLoggable(Level.FINE)) {
+                LOG.fine("CSP violation reported by browser seems to be invalid JSON: " + jsonStr);
+            }
+        }
+    }
+
+    /**
+     * Read up to 10 KiB of POST data to mitigate DoS
+     * @param input
+     * @return JSON string reported by browsers
```
melloware: Missing javadoc on input (nitpick).

melloware (changes requested):
Really nice start on this.  Lot of thought and effort went into making this work.

src/main/java/org/primefaces/application/resource/csp/Constants.java
cnsgithub: @melloware Thanks for your review. 

Okay, I will rename it to `CspHeader` and abbreviate the other class names as well (`Csp` prefix instead of `ContentSecurityPolicy` which is ways too long, almost like Spring class names :wink: ). However there has already been a `Constants` class in PrimeFaces so I thought this might be ok and the package structure is meaningful enough.

src/main/java/org/primefaces/application/resource/csp/ContentSecurityPolicyFilter.java
cnsgithub: Both `@WebServlet` and `@WebFilter` annotations are defined in Servlet 3.0 specification, that's right. When looking at `pom.xml` PrimeFaces is already using Servlet 3.0 and JSF 2.3 so I am not sure if we have to support older versions at all?

src/main/java/org/primefaces/application/resource/csp/ContentSecurityPolicyReportServlet.java
cnsgithub: See above...

src/main/java/org/primefaces/application/resource/csp/ContentSecurityPolicyReportServlet.java
cnsgithub: OMG, refined that javadoc! :+1:

cnsgithub:
In the latest commit 41264ce you will find the first steps to support partial responses as well. Therefore I extended `PrimePartialResponseWriter` and delegate the non-partial-response-specific stuff to the former `CspScriptsResponseWriter` to avoid code duplication.

Basically, when looking at a partial response we have this structure (response when clicking at _This Surname_ in [Partial Process Showcase](http://localhost:8080/showcase/ui/ajax/process.xhtml)):

```xml
<?xml version='1.0' encoding='UTF-8'?>
<partial-response><changes><update id="j_idt114:grid"><![CDATA[<table id="j_idt114:grid" cellpadding="5" style="margin-bottom:10px">
<thead>
<tr><th colspan="2" scope="colgroup"><div id="j_idt114:msgs" class="ui-messages ui-widget" aria-live="polite" data-global="false" data-summary="data-summary" data-severity="all,error" data-redisplay="true"><div class="ui-messages-info ui-corner-all"><span class="ui-messages-info-icon"></span><ul><li role="alert" aria-atomic="true"><span class="ui-messages-info-summary">Welcome null sdf</span></li></ul></div></div></th></tr>
</thead>
<tbody>
<tr>
<td><label id="j_idt114:j_idt115" class="ui-outputlabel ui-widget" for="j_idt114:firstname">Firstname:</label></td>
<td><input id="j_idt114:firstname" name="j_idt114:firstname" type="text" class="ui-inputfield ui-inputtext ui-widget ui-state-default ui-corner-all" data-p-label="Firstname" /><script id="j_idt114:firstname_s" type="text/javascript" nonce="Y2RhNjc3YWQtZjY2OC00ZWI1LWIxNTItOTkxOWQwYTY3NDBk">PrimeFaces.cw("InputText","widget_j_idt114_firstname",{id:"j_idt114:firstname"});</script></td>
</tr>
<tr>
<td><label id="j_idt114:j_idt116" class="ui-outputlabel ui-widget" for="j_idt114:surname">Surname:<span class="ui-outputlabel-rfi">*</span></label></td>
<td><input id="j_idt114:surname" name="j_idt114:surname" type="text" value="sdf" aria-required="true" class="ui-inputfield ui-inputtext ui-widget ui-state-default ui-corner-all" data-p-label="Surname" data-p-rmsg="Surname is required." data-p-required="true" /><script id="j_idt114:surname_s" type="text/javascript" nonce="YWJmODhiOWYtZTJlZi00NDFiLWFkMGItYzE0YThiZmMwMzlm">PrimeFaces.cw("InputText","widget_j_idt114_surname",{id:"j_idt114:surname"});</script></td>
</tr>
</tbody>
</table>
]]></update><update id="j_id1:javax.faces.ViewState:0"><![CDATA[6102594184729907959:4455767693534189078]]></update></changes></partial-response>
```

Notice the `script `elements contained therein which have already been processed by `CspScriptsPartialResponseWriter` (i.e. nonces have been added). Javascript event or URI handlers should be moved to `script `blocks for each `update `or `insert` node.

One bad thing I encountered is the `eval `node which gets included in the partial response when calling `PrimeFaces.executeScript`. My idea was to catch everything between `startEval `and `endEval `and move it somewhere to a `script `block containing a valid `nonce`. However, that's not yet implemented, see TODOs in `CspScriptsPartialResponseWriter`.

One more thing to notice is the way PrimeFaces is updating the DOM elements in `core.ajax.js`, `updateElement` function: `$(PrimeFaces.escapeClientId(id)).replaceWith(content);`. I had to change this because of a [jquery issue](https://github.com/jquery/jquery/issues/3541) losing all attributes (our `nonce` attribute, too) which in turn causes CSP violations.

Another important thing: When experimenting with the ajax stuff in the showcase at a first glance it looks like everything's already working fine. However, when looking more carefully, you will notice that all XHRs are followed by full page refreshs which of course is a no-go. Maybe someone has an idea what could be causing this behavior? In network monitor you can see that somehow the XHR gets aborted (although the partial response had been sent back and our `CspFilter` had set the new nonces in the header before). After the XHR abort another POST request is fired refreshing the full page. Currently no idea, need to debug that. Any help would be appreciated.

image

cnsgithub:
> However, when looking more carefully, you will notice that all XHRs are followed by full page refreshs which of course is a no-go. Maybe someone has an idea what could be causing this behavior? 

Solved that with my last commit ac61ed1. Button's default action was performed after XHR, i.e. form submit... :cry: 

We should now prepare an integration test with all the features turned on: ajax/non-ajax, event handlers, javascript URI handlers, script blocks, `f:ajax`, `p:ajax`... Does anyone feel like doing that? :stuck_out_tongue_winking_eye: PR on https://github.com/primefaces/showcase would be quite nice.

Btw, a good site for testing for valid CSP headers is here: https://cspvalidator.org/#url=https://www.primefaces.org/showcase/

tandraschko:
basically i like the idea!
i will check it in detail if all cases are supported

cnsgithub:
Ok, I'm working hard on it... :wink:

cnsgithub:
Encountered some difficulties:

1. jquery's `DOMEval `function which is called implicitly by `replaceWith` or `html ` for example (used by 
 PrimeFaces) causes CSP violations because they are doing this without preserving the `nonce` attribute:
 `doc.head.appendChild( script ).parentNode.removeChild( script );`
 Adding `nonce` to the `preservedScriptAttributes` could solve this in the future:
 ```javascript
    var preservedScriptAttributes = {
        type: true,
        src: true,
        noModule: true,
        nonce: true
    };
 ```
   See https://github.com/jquery/jquery/pull/3766 for futher information.
   I tried patching PrimeFaces' jquery.js, however there seems to be another (browser-related?) problem, 
   see 2.

2.  It seems that browsers do not union-merge CSP directives received with individual responses. The _Content Security Policy Level 2_ specification is at least in my opinion not clear enough how user agents should behave, see https://www.w3.org/TR/CSP2/#enforcing-multiple-policies for further information about handling of multiple policies. Interesting reading: https://github.com/w3c/webappsec-csp/issues/215. This behavior currently results in CSP violations for scripts contained in partial responses. I don't know how to get around this problem right now. Precalculating nonces, putting them in a registry and set them in the header in the very first place could work, albeit of course is not ideal since we can not know how many nonces will be needed beforehand... Awaiting response from w3c/salvation guys here...

3. Somehow the generated SHA-256 hashes seem to be invalid. See https://github.com/shapesecurity/salvation/issues/202 for further information.

cnsgithub:
> 1. jquery's DOMEval function which is called implicitly by replaceWith or html for example (used by
> PrimeFaces) causes CSP violations

Patched this in PrimeFaces jquery.js (one-liner).

> 2. It seems that browsers do not union-merge CSP directives received with individual responses.

Solved this by generating just one nonce per session and reuse it for all scripts. From a security point of view this should be fine as well. For the same reason not generating any hashes anymore (kept the function however for future use).

> 3. Somehow the generated SHA-256 hashes seem to be invalid. 

This statement was wrong. It's rather a bug in validation of cspvalidator.org, salvation lib resp.

Furthermore, I removed support for javascript URI handlers since there is no general and straightforward solution. However, use of `javascript:` URIs should be very rare in PrimeFaces itself, except `href="javascript:void(0);"` in some *.js. However, those calls could be rewritten to `href="#"`.

tandraschko:
Short review:

- It would be great if we can avoid the the checkstyle exclude
- could you patch the jquery method via overwrite in another file? Otherwise your change will probably be lost on the next update
- i'm not sure about the HeaderWriterResponse and the ServletFilter
  Whats the meaning of it? Can't we plugin in the JSF lifecycle? Feels a bit to hacky.
- CspReportServlet - is this something optional, right? Does the browser report back to the server if CSP fails?

cnsgithub:
> It would be great if we can avoid the the checkstyle exclude

I think we could do that. However, I just thought because `OnCommittedResponseWrapper `is a copy from `spring-security-web` it would be better to not change it to make it easier to compare with newer versions in the future. There are already other candidates as well.

> could you patch the jquery method via overwrite in another file? Otherwise your change will probably be lost on the next update

That makes sense, absolutely. Okay, can you give me a hint how to do that? Btw, I hope that next jquery version 3.4.0 will include that fix.

> i'm not sure about the HeaderWriterResponse and the ServletFilter
> Whats the meaning of it? Can't we plugin in the JSF lifecycle? Feels a bit to hacky.

Do you mean a JSF `PhaseListener `(render response complete) or something like that? I tested that in the first place. Cannot remember right now why I discarded this idea...

> CspReportServlet - is this something optional, right? Does the browser report back to the server if CSP fails?

Yes, the `report-uri` directive and hence the servlet is optional, albeit very reasonable in my opinion. Please see @melloware's review comments and my answer also regarding `@WebServlet `annotation. I thought it would be nice to have the endpoint already built-in. Users may override the endpoint in `web.xml`.

tandraschko:
Yep, PhaseListener would be better e.g.


About overwriting something from jquery, please see: https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.js

melloware:
One of the things I want to make sure to test is that this works with OmniFaces CombinedResourceHandler as well.

melloware:
So far everything is looking good!   I tested with OmniFaces CombinedResource Handler and it properly added the nonce to it like so...

```xml
<script type="text/javascript" 
src="/showcase/javax.faces.resource/eNp1jeEKgzAQg59I-79vc5SbVurdeWkRYQ8_x4aIm78S8oXEPM_8oMSI09LYt_CRfsLTblhnpQ1Z8O5g1DUROCJ5thqERHerpbD_44U2bfWyntT5J5pNhaXiApAHodqcw-HOP-H7g3FPX6mWWNo.js.xhtml?ln=omnifaces.combined&amp;v=1534098660000" 
nonce="ZTcwZDExNTEtMmNlOS00ZjI2LWIxMTEtMDg1ODUyNzZiODE2">
```

tandraschko:
Can you work on it the next weeks @cnsgithub ?

cnsgithub:
Rebased with need to re-introduce `Base64`.

melloware:
Awesome work!

tandraschko:
Could you recreate the PR? 299 files changed is hard to read :D
Why do you need to reintroduce base64?

cnsgithub:
Damn, I was afraid of that. :cry: That's not fun.

> Why do you need to reintroduce base64?

We talked about that in gitter a few weeks ago. That's because script nonces are base64 values.

tandraschko:
If you would not always delete your messages in gitter, it would be easier to follow.

What about DatetypeConverter from java7?

cnsgithub:
> What about DatetypeConverter from java7?

Tested, throwing NPE for some reason. In addition, after reading its [javadoc](https://docs.oracle.com/javase/7/docs/api/javax/xml/bind/DatatypeConverter.html#parseBase64Binary(java.lang.String)), I'm not sure if `xsd:base64Binary` is necessarily a usual base64 string.

cnsgithub:
New pull request: https://github.com/primefaces/primefaces/pull/4262

PR #3885, opened 2018-07-18
fixes #3519 - rowsPerPageTemplate input validation (ensure that the rowsPerPageTemplate attribute is optional and validate against rows attribute as well)

### LINKED ISSUE #3519: Prevent client-side changing of the datatable rows parameter (also affects other pageable components)

stiemannkj1:


tandraschko:
@kukel  can you verify pls?

melloware (approved):

PR #3886, opened 2018-07-18

stiemannkj1:


tandraschko:
Thanks for your effort. I just wonder if there is no easier way for the OSGI stuff? quite much code :D

stiemannkj1:
@tandraschko, technically we could reduce the `PrimeExternalContext` code by using exception handling for logic and loading the classes on every request (basically how `PrimeExternalContext` used to work).

We could also use exception handling for logic in the `MessageFactory` to reduce the code by doing something like this:

```
try {
ResourceBundle resourceBundle = ResourceBundle.getBundle(..., threadContextClassLoader);
}
catch (ResourceNotFoundException e) {
resourceBundle = ResourceBundle.getBundle(..., osgiBundleClassLoader);
}
```

The problem with doing things that way is that we're relying on exception handling for logic multiple times in a single request and [exception handling is expensive and slow](https://stackoverflow.com/questions/299068/how-slow-are-java-exceptions).

The way I've currently designed it, only two or three exceptions should ever be thrown/handled for logic purposes. After that every request can rely on saved values.

tandraschko:
Couldn't we cache the info in PrimeEnvironment?
I would change the detection to try-catch and cache it for each application inside PrimeEnvironment.

stiemannkj1:
@tandraschko, that's a good idea for the `PrimeExternalContext` Liferay detection code, but I'm not sure how it would work for `MessageFactory`. Could you give a quick example of what you are looking for? Should I be caching the `ResourceBundle`s, the message `String`s, the OSGi detection, or all three?

tandraschko:
Ich would add the liferay and osgi detection there :)

stiemannkj1:
@tandraschko, check the changes here: https://github.com/primefaces/primefaces/pull/3908

PR #3890, opened 2018-07-20
Add generic parameter MenuElement to MenuGroup.getElements

krichter722:
The parameter is used in subclasses anyway and MenuElements are the
only kind of elements which make sense to place in the list of
elements of a menu group.


Rapster:
Please create an issue first, this is where you explain your problem, and gives us a better traceability

krichter722:
Well, happy tracing, then...

PR #3898, opened 2018-07-23
Fix #3894: Poll do not start polling if stop = true.

### LINKED ISSUE #3894: Poll: Fix "stop" attribute to not start polling if true
Looking in the poll.xml which generates the documentation for the component is this:

        <attribute>
            <name>stop</name>
            <required>false</required>
            <defaultValue>false</defaultValue>
            <description>Stops polling when true.</description>
        </attribute>

I cannot find anywhere this property is actually used and I have confirmed it does nothing setting this EL expression.  I think somewhere along the lines this component was modified to autoStart="" attribute but the stop was never removed.

melloware:


tandraschko:
Is it tested? This.config should be this.cfg?!

melloware:
Sorry I just checked in that correction.  I had multiple things I was trying out and this was the simplest solution but somehow checked in one of my invalid attempts.  

yes I tested it, basically if Stop is true the component is marked inactive to start even if autoStart="true".

tandraschko:
We should really think if we should merge it.
The documentation says "Stops polling when true."
I think it's ok to control 2 behaviors with 2 different flags.

melloware:
I am ok with it that too.  I just thought it was strange if stop was = true on startup that others might be misled too.  I am definitely OK rejecting this PR.

tandraschko:
Thanks but i would really close it for now as this are really 2 different behaviors

PR #3908, opened 2018-07-27

stiemannkj1:


stiemannkj1:
@tandraschko, I was able to remove all Liferay and Liferay detection code and add an extension point for Liferay Faces to provide functionality from our bridge. For the OSGi code, I also removed the detection and just made it work the same in both the OSGi and non-OSGi environments.

tandraschko:
i like the changes on the MessageFactory but not sure about the PrimeExternalContext.

What about if Liferay would wrap PrimeExternalContext? I think the should be easy to support.

PR #3909, opened 2018-07-27
Improve OSGi support and Liferay detection

stiemannkj1:


stiemannkj1:
@tandraschko, sorry to waste so much time on this. Since push removal is a breaking change, I assumed that it was okay to remove the unnecessary methods from `PrimeExternalContext` now.

tandraschko:
Makes sense! thanks!

PR #3910, opened 2018-07-29
Fix checkstyle warnings for non-encapsulated fields

krichter722:


Rapster (approved):

tandraschko:
I would remove the javadoc from the method or really provide a good javadoc

krichter722:
@tandraschko Agreed, changed.

PR #3915, opened 2018-07-30
Activated trailing whitespace checkstyle check

krichter722:
Activated and upgraded trailing whitespace checkstyle check to be
conform with checkstyle 8.x. Trailing whitespace increases the number of
merge commit and more importantly rebasing work and lowers the
motivation for a less merges or even a merge-free easier to read
history.

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -114,12 +114,12 @@
         <!-- See http://checkstyle.sf.net/config_misc.html -->
         <!-- module name="ArrayTypeStyle"/ -->
         <!-- module name="FinalParameters"/ -->
-        <!-- Line with Trailing Spaces (disabled as it's to noisy)
-        <module name="GenericIllegalRegexp">
-            <property name="format" value="\s+$"/>
-            <property name="message" value="Line has trailing spaces."/>
+        <!-- Line with Trailing Spaces -->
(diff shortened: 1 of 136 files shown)


tandraschko:
I know it's a >bit< annoing but i'm not sure if we can do this, i first have to talk with the PrimeTek team if they are ok with it as the build may file sometimes.

Also i don't think it's very annoying as you can still use ?w=1 on github.

krichter722:
I understand. If everybody show discipline and builds locally before pushing to Travis the build failure should be detected fast since `mvn validate` runs very early and `checkstyle` is pretty fast.

tandraschko:
Applied it manually, thanks

krichter722:
@tandraschko You can ping me next time, and I'll rebase the PR. Just a note from me: the build failures on Travis are not (root) caused by this checkstyle change (I know the controversy around it), because someone made a commit on master breaking the build with a plugin before. Tips: Maintainers should discuss discipline and only commit a few lines of changes directly into master.

krichter722:
@tandraschko I went through the changes and now understand that it makes a lot of sense that you applied the changes manually, sorry. Excellent checkstyle setup! My comment on breaking the build on master remains the same.

PR #3918, opened 2018-07-30
Document DefaultMenuItem.url, outcome and command

krichter722:
The user might be confused that a new page can be loaded with outcome
(which specifies a navigation case) and and url (which allows to
navigate to an arbitrary URL). The specification of url causes command
to be ignored which the user should be informed about in the Javadoc.

Inspired by https://stackoverflow.com/questions/51598575/whats-the-difference-between-pmenuitems-url-and-outcome-attributes.

### DIFF
--- a/src/main/java/org/primefaces/model/menu/DefaultMenuItem.java
+++ b/src/main/java/org/primefaces/model/menu/DefaultMenuItem.java
@@ -34,14 +34,33 @@ public class DefaultMenuItem implements MenuItem, UIOutcomeTarget, AjaxSource, S
     private String onclick;
     private String style;
     private String styleClass;
+    /**
+     * The URL to redirect to after the menu item has been clicked. Similar to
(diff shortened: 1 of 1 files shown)


tandraschko:
+1
it would be great if the docu would be a bit more detailed

the big difference is that url really means the native url and is not touched (no prepending of the contextPath, not appending the sessionId or windowId), just encoded. Also f:params might be appended as url params (not sure about this without looking at the code!)

Outcome really means the JSF outcome, which is resolved by the NavigationHandler

command is a MethodExpression as string

krichter722:
@tandraschko Thanks for you feedback. I incorporated your suggestion in an amended commit. I have too less understanding of Primefaces to research reliable or efficient whether `f:params` are appended to `url` (therefore I omitted it in the current draft). Please feel free to make further suggestions and add research results.

I speak of fully qualified URLs instead of native ones - assuming we mean the same - because the term is used more frequently and is well defined in the context of URLs.

PR #3927, opened 2018-08-03
Schedule: do not bind a view change listener on prev/next year buttons

cghislai:
When the prev/next year buttons are enabled on the schedule, they are serialized as the 'view' state of ajax requests: `scheduleId_view: prevYear`.
Under some circumstance I couldn't reproduce on the showcase, this caused the schedule to instantiate a view from a spec which didn't exit, causing a javacript error and preventing the schedule to be rendered : 
```
var spec = this.getViewSpec(viewType); // viewType === 'prevYear', spec === false
return new spec['class'](this, viewType, spec.options, spec.duration);
// Uncaught TypeError: spec.class is not a constructor 

```

This pull request prevents the viewChange listener to be bound on those buttons, like for the prev/next month buttons.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/schedule/1-schedule.js
+++ b/src/main/resources/META-INF/resources/primefaces/schedule/1-schedule.js
@@ -191,7 +191,8 @@ PrimeFaces.widget.Schedule = PrimeFaces.widget.DeferredWidget.extend({
     },
 
     bindViewChangeListener: function() {
-        var viewButtons = this.jqc.find('> .fc-toolbar button:not(.fc-prev-button,.fc-next-button,.fc-today-button)'),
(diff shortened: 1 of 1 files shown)


melloware:
Please create an Issue to match up with this PR.  All issue are tracked for release notes and for others to see the problem before a fix is made.

cghislai:
Fix #3928

melloware (approved):

melloware:
Thanks for creating the ticket.

tandraschko:
@OBreidenbach hey mister schedule :) Would you like to retest it?

OBreidenbach:
Funny title, mister schedule. Interesting how easy it was to earn it. ;-)
I can confirm that the bug is fixed and I did not see any regressions.

tandraschko:
:D :P
I really appreciate help from the other people of the community.

PR #3930, opened 2018-08-04
Provide BaseMenuModel.equals and hashCode

krichter722:
BaseMenuModel.equals and hashCode should be based on elements which is
the central class data and useful for subclasses.

### DIFF
--- a/src/main/java/org/primefaces/model/menu/BaseMenuModel.java
+++ b/src/main/java/org/primefaces/model/menu/BaseMenuModel.java
@@ -18,6 +18,7 @@ package org.primefaces.model.menu;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 /**
  * Base implementation for model of a programmatic menu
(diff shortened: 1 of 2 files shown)


melloware (approved):

melloware:
Please create an Issue ticket that this PR can be linked to.  PrimeTek doesn't accept PR's without an accompanying Issue ticket!

krichter722:
@melloware issue #3931. Added `equals` and `hashCode` to DefaultMenuItem as well.

krichter722:
I just realized that I changed this because I was working with the view rather than the model (JPA entities) which is never a good idea (`MenuModel` is more of a view component than a model in the sense of JSF's and Primefaces' MVC afaik). Therefore this change might not be a good idea.

tandraschko:
Also not sure about it.
I think it would make more sense to compare the id but the id is auto generated by the menu renderers?!

krichter722:
I thought more about it and it doesn't make sense, sorry for bothering you with this.

Issue #3746, opened 2018-05-31
core.js: PrimeFaces.getFacesResource() throws TypeError "null" when version parameter is passed to it in a portlet environment


stiemannkj1:
## 1) Environment
- > PrimeFaces version:

    `6.2`

- > Does it work on the newest released PrimeFaces version? Does it work on the newest sources in GitHub?

    The bug exists in `master` (`6.3-SNAPSHOT`) as well.

- > Application server + version:

    Tomcat (should be reproducible in any app server)

- > Affected browsers:

    Chrome, Firefox (should be reproducible in any browser)

## 2) Steps to reproduce

1. Copy the sample xhtml into a webapp project and navigate to the page.
2. Click through each of the alerts and note the test results.

## 3) Expected behavior

If the bug is fixed, all alerts will show a "test passed." message.

## 4) Actual behavior

If the bug still exists, an the final alert (for testing a portlet resource URL with a version) will appear with the following message:

> portlet resource URL (with version): TEST FAILED!

The following error will also appear in the browser logs:

```
Uncaught TypeError: Cannot read property '1' of null
    at Object.getFacesResource (core.js.xhtml?ln=primefaces&v=6.3-SNAPSHOT:357)
    at (index):40
```

## 5) Sample XHTML

```
<html xmlns="http://www.w3.org/1999/xhtml"
      xmlns:h="http://xmlns.jcp.org/jsf/html"
	  xmlns:f="http://xmlns.jcp.org/jsf/core"
	  xmlns:p="http://primefaces.org/ui">
	<h:head />
	<h:body>
		<h:outputScript>

			try {
				var testWebappURL = window.PrimeFaces.getFacesResource("test.css", "testLib");
				console.log(testWebappURL);
				alert('web app resource URL (without version): test passed.');
			}
			catch (error) {
				alert('web app resource URL (without version): TEST FAILED!');
				throw error;
			}

			try {
				var testWebappURL = window.PrimeFaces.getFacesResource("test.css", "testLib", "7.0");
				console.log(testWebappURL);
				alert('web app resource URL (with version): test passed.');
			}
			catch (error) {
				alert('web app resource URL (with version): TEST FAILED!');
				throw error;
			}

			// Replace the core.js resource url with a portlet core.js resource url.
[...]

PR #3939, opened 2018-08-06
Fix #3936: JS getFacesResource() function more generic to handle case…

### LINKED ISSUE #3936: PrimeFaces.getFacesResource() does not work with OmniFaces CombineResourceHandler
You can read about the issue on StackOverflow here:  https://stackoverflow.com/questions/51679748/determine-jsf-library-resource-using-javascript

The real issue is that [this JS method](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/core/core.js#L341) assumes core.js is on the page for determining the URL and script replacement.  However when using OmniFaces CombinedResourceHandler all the JS files are put into one big JS file at runtime so there is only `UUID.js=ln=omnifaces.combined` on the page and no PrimeFaces core.js so the logic in [this JS method](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/core/core.js#L341) breaks.

@tandraschko Should I move the changes that I just made in PrimeFaces Extensions over to PrimeFaces?  It makes these JS methods much more flexible and still works for either PF or OmniFaces and still works in Portlet environment.  If so I will submit a PR.

I fixed it in Extensions with this ticket: https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/601

melloware:
Refactored the JS functions to work when core.js is not in the DOM. Especially in the case of OnmiFaces CombinedResourceHandler being used.

https://stackoverflow.com/questions/51679748/determine-jsf-library-resource-using-javascript


tandraschko:
Does PF still use #getFacesResource?
Does #getResourceScriptURI work with portlets correctly ('/core.js' vs '=core.js')?

Also what about moving those resource utils to e.g. core.resources.js?

melloware:
So PF does not use these functions at all but they seem really useful.  So they probably are safe to be removed. It looks like they are only used by 3 PFE components DocumentViewer, CKEditor, and SpeedTest.

yes the regex handles the Portlet environment or normal environment.

I can move them to core.resources.js no problem.

So Should I just update this PR to delete the getFacesResource method since I already have it all working in PF Extensions?

tandraschko:
AFAIR getFacesResource was used by the dynamic resource loading and should not be used anymore.

I think we have 2 possible ways:
1) Let me check the premium layouts tomorrow. If it isn't used there anymore, i would just depreacte it.
2) Move the methods from extensions to PF but do a cleanup and remove unused stuff. I still think one method is one. The lookup of the base resource, should not require any additional methods. I think it should be easy to fix the current method in PF.

I prefer 2) and fix our method.
We must just change 2 places:
scriptURI.replace('core.js', name);
and
scriptURI = scriptURI.replace(namespace + 'ln=primefaces', namespace + 'ln=' + library);

the impl in extensions looks a bit over engineered. We must just get the first JSF resource url and replace the resourcename and library.

melloware:
Agreed I feel like deprecating it or even deleting it since no current PF component uses it would be the right move.   

The Extensions stuff (over-engineered) was all there before I just fixed it.  However I will confirm that CKEDITOR GETURL uses all of those methods including the isExtension and getExtension.  So in PFE all of those methods are used.

Let me know what you find out about Premium templates if they are used.

tandraschko:
I know, it was me :D

I think PF should fix it as it's a very generic method, we can be used in future again, not just by extensions

melloware:
OK so where did we end up?  Did you want me to just to write 1 getFacesResource method and move it to core.resources.js and remove those other methods only used by CKEDITOR?

tandraschko:
Note sure :) The other methods are very special for CKEditor, because the of the special CKEDITOR_GETURL method.
getFacesResource is used by other components, too. right?

WDYT?

melloware:
getFacesResource is used by CKEditor, Speedtest and DocumentViewer.  I feel like I just move all the changes above into core.resources.js.  The methods may look like overkill but they ended up being elegant and reusable.

melloware:
I refactored to core.resources.js.  I tested it in the browser locally by doing this in Console.

`PrimeFaces.resources.getFacesResource('test.js', 'melloware', 2.4);`

**Result:**
"http://localhost:8080/showcase/javax.faces.resource/test.js.xhtml?ln=melloware&v=2.4"

tandraschko:
1) the methods are now in both files? we should just leave getFacesResource there as @deprecated and call resources.getFacesResources() for now.
2) I don't like the method #getResourceScriptName. The method caches the result as static but the param is dynamic actually. Thats a big design issue.

melloware:
Good catch.  Requested changes made!

tandraschko:
Did you try all the methods with a-non-extension mapping?
Did you also try it on Liferay? Otherwise we have to ask if someone can test it.

melloware:
When you say Non Extension mapping what do you mean?  Meaning FacesServlet has no URL mappings in web.xml?

As for LifeRay yes we need someone to test it as I have no Liferay setup.  That being said I am fairly confident it will work as I am using a capture group to look for "/core.js" or "=core.js" like portlets do it.

```
// find script...normal is '/core.js' and portlets are '=core.js'
var scriptRegex = new RegExp('\/' + PrimeFaces.RESOURCE_IDENTIFIER + '(\/|=)(.*?)\.js');
```

tandraschko:
extension mapping means =
you map the FacesServlet to e.g. *.xhtml or *.jsf

non-extenion mapping (don't know the correct name) means=
you map the FacesServlet to a directory instead extensions. e.g. /faces/*

melloware:
Tested successfully with this using the showcase.

```xml
    <servlet>
        <servlet-name>Faces Servlet</servlet-name>
        <servlet-class>javax.faces.webapp.FacesServlet</servlet-class>
    </servlet>
    <servlet-mapping>
        <servlet-name>Faces Servlet</servlet-name>
        <url-pattern>/faces/*</url-pattern>
    </servlet-mapping>
```

**Then in Chrome Console type:**
```javascript
PrimeFaces.resources.getFacesResource('test.js', 'melloware', 2.4);
```

**Result:**
> "http://localhost:8080/showcase/faces/javax.faces.resource/test.js?ln=melloware&v=2.4"

tandraschko:
@stiemannkj1 Could you do us an favor and test such a JS call as from melloware above with this PR applied, if it's still correct on Liferay? :) Thx!

stiemannkj1:
@tandraschko, I'll check it tomorrow. Thanks for the heads up. I'll let you konw.

stiemannkj1:
@tandraschko, it looks like there's some Liferay detection that didn't get transferred:

```
diff --git a/src/main/resources/META-INF/resources/primefaces/core/core.resources.js b/src/main/resources/META-INF/resources/primefaces/core/core.resources.js
index 2a504b8..8be0ade 100644
--- a/src/main/resources/META-INF/resources/primefaces/core/core.resources.js
+++ b/src/main/resources/META-INF/resources/primefaces/core/core.resources.js
@@ -86,12 +86,12 @@ if (!PrimeFaces.resources) {
           /**
            * For a URI parses out the name of the script like primefaces-extensions.js
            * 
-           * @param the URI of the script
+           * @param scriptURI the URI of the script
            * @returns {string} The script name.
            */
           getResourceScriptName : function(scriptURI) {
              // find script...normal is '/core.js' and portlets are '=core.js'
-             var scriptRegex = new RegExp('\/' + PrimeFaces.RESOURCE_IDENTIFIER + '(\/|=)(.*?)\.js');
+             var scriptRegex = new RegExp('\/?' + PrimeFaces.RESOURCE_IDENTIFIER + '(\/|=)(.*?)\.js');
              return scriptRegex.exec(scriptURI)[2] + '.js';
           },
 
@@ -102,7 +102,14 @@ if (!PrimeFaces.resources) {
            */
           getResourceScriptURI : function() {
              if (!PrimeFaces.resources.SCRIPT_URI) {
-                PrimeFaces.resources.SCRIPT_URI = $('script[src*="/' + PrimeFaces.RESOURCE_IDENTIFIER + '/"]').first().attr('src');
+                PrimeFaces.resources.SCRIPT_URI =
+                        $('script[src*="/' + PrimeFaces.RESOURCE_IDENTIFIER + '/"]').first().attr('src');
+
+                // portlet
+                if (!PrimeFaces.resources.SCRIPT_URI) {
+                    PrimeFaces.resources.SCRIPT_URI =
+                            $('script[src*="' + PrimeFaces.RESOURCE_IDENTIFIER + '="]').first().attr('src');
+                }
              }
              return PrimeFaces.resources.SCRIPT_URI;
           }
```

I haven't tested the above patch, but it should work. I'll take closer look on Monday to make sure nothing is broken.

@melloware, I was trying to test with `pe:documentViewer` (which is where some issues with PrimeFaces extensions were discovered), but it looks like PrimeFaces Extensions now has its own version of this code. Is that new for `6.3-SNAPSHOT`? If so, it'll need to have the same Liferay detection as PrimeFaces in order to work in Liferay.

Related: https://github.com/primefaces/primefaces/issues/3746

melloware:
@stiemannkj1 To answer your questions..

1. Yes this code is in Extensions 6.2.8 because I needed it fixed now and not wait for 6.3 to be released.

2. My goal was to port the code to 6.3 core PF and get that code out of PF Extensions thus this PR.

3. The LifeRay detection wasn't left out. The code previously was looking for '/core.js' or in LifeRay '=core.js'.  I do all of that with 1 Regex that looks for either situation and parses it out.

```javascript
var scriptRegex = new RegExp('\/?' + PrimeFaces.RESOURCE_IDENTIFIER + '(\/|=)(.*?)\.js');
```

4. You can test it by either loading Extensions 6.2.8 in LifeRay and opening up a Javascript console in your LifeRay app and entering this JS command:

`PrimeFacesExt.getFacesResource('test.js', 'melloware', 2.4);`

That should properly create a valid URL for LifeRay.  Let me know if it does not.

stiemannkj1:
@melloware, @tandraschko, I've sent 2 PRs to ensure this code works in Portlets:

1. https://github.com/melloware/primefaces/pull/2 for PrimeFaces
2. https://github.com/primefaces-extensions/core/pull/52 for PrimeFaces Extensions Core

I tested the code thoroughly in both Pluto (the Portlet Spec reference implementation) and Liferay. However, I didn't test in a webapp/servlet environment.

melloware:
@stiemannkj1 Looks good I just merged your commits and tested locally.  everything is still working in the regular servlet environment.

tandraschko:
So everything is fine now? :D

melloware:
Yes all tested in both regular mode and portlet mode!

PR #3960, opened 2018-08-11
Improve p:accordionPanel's activeIndex attribute documentation

### LINKED ISSUE #3959: Documentation: p:accordionPanel's activeIndex documentation should mention that it requires a value expression
## 1) Environment
- Does it work on the newest released PrimeFaces version? Version? No. 6.3-SNAPSHOT
- Application server + version: ./.
- Affected browsers: ./.

## 2) Expected behavior

It should be clear that `p:accordionPanel`'s `activeIndex` attribute needs to be a value expression. A method expression causes `Illegal Syntax for Set Operation`.

## 3) Actual behavior

The fact that `p:accordionPanel`'s `activeIndex` attribute requires a value expression because it reflects the currently active index in it, causes confusion among users, see e.g. https://stackoverflow.com/questions/24370861/illegal-syntax-for-set-operation-how-to-tell-jsf-i-dont-want-a-setter/24416903.

## 4) Steps to reproduce

./.

## 5) Sample XHTML

./.

## 6) Sample bean

./.

krichter722:
Add a hint that it needs to be a value expression.

closes #3959

### DIFF
--- a/src/main/resources-maven-jsf/ui/accordionPanel.xml
+++ b/src/main/resources-maven-jsf/ui/accordionPanel.xml
@@ -38,7 +38,7 @@
             <required>false</required>
             <type>java.lang.String</type>
             <defaultValue>0</defaultValue>
-            <description>Index of the active tab or a comma separated string of indexes when multiple mode is on. Default is zero.</description>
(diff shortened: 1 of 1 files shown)


tandraschko:
I don't like that big description.
I would just add something like "As activeIndex is a bidrectional attribute, it doesn't support MethodExpressions".

krichter722:
Agreed. Changed.

PR #3964, opened 2018-08-14
fix #2089

### LINKED ISSUE #2089: selectManyMenu: poor performance when selecting all elements

Mobe91:
This improves performance of selectAll/unselectAll on selectManyMenu by partially cloning and replacing the DOM.
It is a follow-up PR on https://github.com/primefaces/primefaces/pull/2090.

@tandraschko Regarding your comment on the other PR:
> NOTE: it would be great if you could first provide a issue for adding selectAll/unselectAll methos for those widgets and then, in a second step, provide a PR to improve the performance.

These methods seem to have been added to selectlistbox widget already.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/forms/forms.selectlistbox.js
+++ b/src/main/resources/META-INF/resources/primefaces/forms/forms.selectlistbox.js
@@ -62,10 +62,46 @@ PrimeFaces.widget.SelectListbox = PrimeFaces.widget.BaseWidget.extend({
             this.setupFilterMatcher();
         }
     },
+    
+    selectAll: function() {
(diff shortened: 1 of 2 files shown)


tandraschko:
Hmmmm, i really not sure about the change.
Isn't there another way to optimize it, without the clone and a complete reinit of the widget?

tandraschko:
Fixed without cloning

PR #3970, opened 2018-08-16
Fix #3967 - Remove maven-jsf-plugin

### LINKED ISSUE #3967: Remove maven-jsf-plugin
See https://github.com/primefaces/maven-jsf-plugin/issues/4

- [X]  Generate abstract base component + component impl
- [X] Remove "Generated, Do Not Modify"
- [ ] ~~Consider maven-license-plugin to change file header if elite package~~ (Not now)
- [ ] Cleaning 
  - [ ] Remove  comments related to maven-jsf-plugin
  - [X] delete unnecessary PropertyKeys#ctor since SONAR will complain
  - [x] generates-sources faces-config.xml and taglib.xml should use spaces
  - [ ] remove templates and xml definition files
  - [x] Format (tabs to spaces, unnecessary this, cumbersome blank lines)
  - [x] Organize imports
  - [x] Remove `isRTL` method in impls
  - [x] Inconsistency variable declarations `static final/final static` -> use ``static final`
- [x] Checkstyle
- [ ] Provide unit test if possible to check consistency definition between taglib and java

Rapster:
- Generate abstract base component + component impl
- Remove "Generated, Do Not Modify"
- delete unnecessary PropertyKeys#ctor since SONAR will complain


tandraschko:
Yep i like it. Later, we could easily replace the content of the *Base classes with something like a CDK in the future.

If the PR is ready, we just need the GO from Cagatay.
I already discussed it with him last year and he was ok to do it this way, without CDK.

Rapster:
Glad to heat that ;-) I updated the roadmap for that ticket, if anything comes to your mind let me know.

-1 for CDK (Sounds like it's been generated by a tool while Base is just a convention name)

Rapster:
I'll start doing the cleaning part once I have a go (it requires that no one pushes on master, i might need half a day) @mertsincan @tandraschko are you planning to push on master today?

Rapster:
Quick question, since the base is some kind of "interface/definition" of the component (I prefer to see it that way than just a boilerplate container). Does it make sense to put the COMPONENT_TYPE, DEFAULT_RENDERER, etc. in the base class?

tandraschko:
I would probably put in the implementation class

Rapster:
I removed my coment since I realized it was okay:

- COMPONENT_FAMILY: I don't see how it could be different in PF context (and it can be easily overrided)
- DEFAULT_RENDERER: Same reason, easily overridable
- COMPONENT_TYPE: Not used in java files... but makes probably more sense in impl.

tandraschko:
COMPONENT_TYPE is only used by endusers
for example:

`ColorPicker colorPicker = (ColorPicker) FacesContext.getCurrentInstance().getApplication().createComponent(ColorPicker.COMPONENT_TYPE);`

Rapster:
Yes, i'll put this one in impl

Rapster:
Is Base class should be public? I feel that package visibility should be enough. I don't see end-users using that class

tandraschko:
Actually package visibility should be fine but not sure about some weird edge cases.
But we can still change it back to public if something shouldn't work.

Rapster:
Regarding comments, should we put as comment what there is in the taglib.xml? If yes, where? On PropertyKeys#value or getter/setter? The probme is that, it forces us to keep synchronize doc in taglib and java

tandraschko:
what exactly do you mean? what would you comment there?

Rapster:
Something like

```java
public enum PropertyKeys {
	    // Name of the client side widget
		widgetVar,
		// Background color, default is #ffffff
		backgroundColor,
		// Foreground color, default is #000000
		color,
		@// Thickness of lines, default is 2
		thickness,
		// Inline style of the component
		style,
	    // Style class of the component
		styleClass,
		// When enabled, signature is used for display purposes only
		readonly,
		// Adds a guideline when enabled, default is false
		guideline,
		// Color of the guideline, default is #a0a0a0
		guidelineColor,
		// Offset of guideline from bottom
		guidelineOffset,
		// Guide line indent from the edges
		guidelineIndent,
		// Client side callback to execute when signature changes
		onchange,
		// Write-only value used to pass the value in base64 to backing bean
		base64Value;
}
```

tandraschko:
I would skip it for now, i think it makes only sense If we can generate the taglib later

Rapster:
ok. Now, regarding imports, when there is too many imports from the same package. We should replace it by `.*` (after 5 consecutive import). E.g:

```java
import javax.faces.view.facelets.FaceletContext;
import javax.faces.view.facelets.FaceletException;
import javax.faces.view.facelets.TagAttribute;
import javax.faces.view.facelets.TagConfig;
import javax.faces.view.facelets.TagHandler;
```
becomes

```java
import javax.faces.view.facelets.*;
```

I think it's a good idea, it becomes way less cumbersome for a class like `DataTable`?

tandraschko:
It has pro and cons.... I usually do it in my applications but i avoid it in frameworks actually as it can lead to errors.
I'm not sure if i would do it because it only hurts in some big components.

Rapster:
Exactly, that's the reason why I would do it. It hurts only big components, so you'll see this very few times, but it gives so much "air" to the component! Honeslty, it looks so much nicer when I read DataTable impl (or other being said) I can make update the PR and it'd be ready to go :wink:

tandraschko:
i'm ok with it ;) But only after 5+ imports from the same package.

tandraschko:
It would be also nice if you can re-order the imports then

Rapster:
Cool. Yes I kept the sort like before

Rapster:
Is that class makes sense today?

```java
package org.primefaces.component.separator;

/**
 * Backward compatibility between PrimeFaces Mobile targeting 3.5 and PrimeFaces 4.0
 */
public interface Separator {

}
```

If I can delete it, it would help me to go faster in the migration...

tandraschko:
Is it used somewhere? XML, Java?

Rapster:
Nope, it's not referenced anywhere. It was added in 2013. According to the comment, I'd say it was only meant to be for PF Mobile

tandraschko:
+1 to remove then 😅

Rapster:
Checkstyle is getting tricky from this point...

Rapster:
Why those attributes 

- Tree#FILTER_CONSTRAINTS
- TreeTable#FILTER_CONSTRAINTS
- DataTable#FEATURES

Are using static initialization block instead of using `Collections.unmodifiableMap` inline initialization like `private static final Map<String, Class<? extends BehaviorEvent>> BEHAVIOR_EVENT_MAPPING = Collections.unmodifiableMap(new HashMap<String, ....>(){{}};`?

I think there is a typo here, since the visibility is package, you can't do much with it. So I would use `Collections.unmodifiableMap`, the reason I mention this, is because of the checkstyle.

Otherwise Checkstyle#Indentation does not really make sense to me, most the errors is due to those declarations:

```java
    private static final Map<String, Class<? extends BehaviorEvent>> BEHAVIOR_EVENT_MAPPING = Collections.unmodifiableMap(new HashMap<String, Class<? extends BehaviorEvent>>() {{
        put("change", null);
    }});
```

It wants me to add 4 more spaces at the put instruction... Besides that, I'm ready to merge that in master. @tandraschko let me know asap before more commits are pushed to master, I don't want to repeat that process too many times...

tandraschko:
If possible, we can remove the static constructor, no problem.

Rapster:
I added a comment on my last post, maybe you didn't see it

tandraschko:
about the commits?
I would have to talk with Mert. When are you ready? Currently the travis build still fails

Rapster:
Build fails mostly because of some strange indentation rule (as I explained in my previous post)

tandraschko:
If not possible to configure it via checkstyle especially for this case, i would switch back to static constructors probably
much better as 150 //CHECKSTYLE:OFF

Rapster:
I'm not pleased with both anyway, both has their disadvantages. The best solution is to use something like https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/ImmutableMap.Builder.html. But we don't want to add a dependency, so I was thinking creating a quick builder which does that job perfectly:

```java
    /**
     * Fluent builder for {@link Map}'s
     *
     * @param <K> key type
     * @param <V> value type
     */
    public static final class MapBuilder<K, V> {

        private final Map<K, V> map;

        private MapBuilder(Map<K,V> map) {
            this.map = map;
        }

        public static <K, V> MapBuilder<K, V> builder() {
            return new MapBuilder<>(new HashMap<K, V>());
        }

        public MapBuilder<K, V> put(K key, V value) {
            map.put(key, value);
            return this;
        }

        public Map<K, V> build() {
            return Collections.unmodifiableMap(map);
        }
    }
```

It has a lot of perks, first no ugly static initializer (which messing up which checkstyle also), using double brace might lead to memory leak (see http://minborgsjavapot.blogspot.com/2014/12/java-8-initializing-maps-in-smartest-way.html). Builder pattern is the best way to go. As an exemple:

```java
    static final Map<String, FilterConstraint> FILTER_CONSTRAINTS = MapBuilder.<String, FilterConstraint>builder()
            .put(STARTS_WITH_MATCH_MODE, new StartsWithFilterConstraint())
            .put(ENDS_WITH_MATCH_MODE, new EndsWithFilterConstraint())
            .put(CONTAINS_MATCH_MODE, new ContainsFilterConstraint())
            .put(EXACT_MATCH_MODE, new ExactFilterConstraint())
            .put(LESS_THAN_MODE, new LessThanFilterConstraint())
            .put(LESS_THAN_EQUALS_MODE, new LessThanEqualsFilterConstraint())
            .put(GREATER_THAN_MODE, new GreaterThanFilterConstraint())
            .put(GREATER_THAN_EQUALS_MODE, new GreaterThanEqualsFilterConstraint())
            .put(EQUALS_MODE, new EqualsFilterConstraint())
            .put(IN_MODE, new InFilterConstraint())
            .put(GLOBAL_MODE, new GlobalFilterConstraint())
            .build();
```

Checkstyle should be happy with this

tandraschko:
yup, +1
i would just call it "HashMapBuilder" or allow to pass a map instance

Rapster:
Alright, almost there! Last problem with indentation, using satic initialization array like this:
```java
private static final String[] SUPPORTED_TYPES = new String[]{"aif", "aiff", "aac", "au", "bmp", "gsm", "mov", "mid", "midi", "mpg", "mpeg",
            "mp4", "m4a", "psd", "qt", "qtif", "qif", "qti", "snd", "tif", "tiff", "wav", "3g2", "3pg"};
```

Will complain about indentation, it should be 8  spaces instead of 12, which is very strange because as mentionned earlier, it required 12 but 12 is for collection, while is 8 for arrays (kind of stupid I must say...) I worked-around the problem by writing the whole thing on one line, just like that:

```java
public static final String[] EVENTS = {START, SUCCESS, COMPLETE, ERROR};
```

But for a class like `QuickTimePlayer`, it won't work... We have three choices, either change the checkstyle rule, make indentation of 8 spaces or change  String[] into List<> (there is only one case like this, and using List instead of Array is not a bad idea IMO)

tandraschko:
If the occurence is below 5, lets add a //CHECKSTYLE:OFF
I can check it too after merging

Rapster:
Great! Now the very last one: LineLength! :D

Most of the component has these long style class attributes like:
```java
public static final String MULTIPLE_CONTAINER_WITH_DROPDOWN_CLASS = "ui-autocomplete-multiple-container ui-autocomplete-dd-multiple-container ui-widget ui-inputfield ui-state-default ui-corner-left";
```

Of course it's longer than 160 characters, but I don't think there is a simple way to solve that. I don't wanna to set these variables on several lines, otherwise it's gonna be a big mess. I think we shoudn't apply this rule on these attributes. I'm planning to add a rule to `Indetation` module that skip every attributes named `\w*_CLASS`, WDYT?

tandraschko:
yep, i think thats ok :)

Rapster:
Build finally passing :tada: I'll remove everything related to `maven-jsf-plugin` once I'm about to merge into master

tandraschko:
nice, great work!
will do a short review tomorrow and then we should merge it ASAP

tandraschko:
https://github.com/primefaces/primefaces/blob/9ca76ee5a30b27780d7b1ab190b41f06883c0207/src/main/java/org/primefaces/component/button/ButtonBase.java

`@Override
public boolean isDisableClientWindow() {`


many times in PF we didn't use @Override because it was added in newer JSF versions are not available in JSF2.0
Can you check if it still works on JSF2.0? (AFAIR it should work be we must test it)

Rapster:
I can have a try, but since `@Override` retention policy is `SOURCE` it won't be a problem at runtime

Rapster:
Works fine using Mojarra 2.0 :wink:

tandraschko:
all right!

Rapster:
If it's more convenient, I can see if I can push the whole thing in 1 commit (instead of 34)

Rapster:
@tandraschko Should I merge?

tandraschko:
+1 :)

Rapster:
Does it mean I should merge the PR or I make one commit (so that git history does not get poluted)?

tandraschko:
merging the PR is ok for me

PR #3983, opened 2018-08-22
Fix #3984 AutoComplete width for multiple=true

### LINKED ISSUE #3984: Autocomplete: improve panel width when the attribute multiple is true
Hi, the autocomplete panel width has a problem when the attribute multiple is true, you can check it in the showcase:
- https://www.primefaces.org/showcase/ui/input/autoComplete.xhtml

The panel width when we select the first item from the list looks fine, but when we select the second item the panel width is shorter, if we select the third item the panel width is even shorter.

I have created a pull request trying to fix it.

wsaca:
This change is to fix a problem with the autocomplete panel width when the attribute **multiple** is **true**. You could check this problem in the showcase:
- https://www.primefaces.org/showcase/ui/input/autoComplete.xhtml

The panel width when we select the first item from the list looks fine, but when we select the second item the panel width is shorter, if we select the third item the panel width is even shorter.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
+++ b/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
@@ -823,7 +823,7 @@ PrimeFaces.widget.AutoComplete = PrimeFaces.widget.BaseWidget.extend({
         var panelWidth = null;
 
         if(this.cfg.multiple) {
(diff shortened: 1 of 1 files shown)


melloware:
Please create an Issue to link this PR to.

wsaca:
The issue #3984 was created, thanks.

melloware:
Perfect. Can you update the title of this ticket to "Fix #3984 AutoComplete width for multiple=true" so it gets picked up by GitHub and automatically associates it to the Issue.

wsaca:
It's done...

PR #3988, opened 2018-08-23
Fix #3986

### LINKED ISSUE #3986: OutputLabel: add support for @AssertTrue
## 1) Environment
- PrimeFaces version: all

## 2) Expected behavior
Right now, the `indicateRequired="true"` attribute of `p:outputLabel` considers `@NotNull`, `@NotEmpty` and `@NotBlank` annotations. I think it should cover also the [`@AssertTrue`](https://docs.jboss.org/hibernate/beanvalidation/spec/1.1/api/javax/validation/constraints/AssertTrue.html) annotation. e.g.: A checkbox with the label "Do you accept terms and conditions?".

PR coming.

NicolaIsotta:
Add AssertTrue support


tandraschko:
hm not sure
we should probably check selectBooleanCheckbox && AssertTrue

PR #3997, opened 2018-08-28
Fix #2127: Picklist prevent client side manipulation.

### LINKED ISSUE #2127: Picklist: disabled items allow for client-side manipulation

melloware:
1. Cleaned up the example code
2. Chose to use javax.faces.component.UIInput.UPDATE as the error message which will be "An error occurred when processing your submitted information."
3. Refactored to make the validate method more readable now that it includes both REQUIRED validation and client side scrubbing validation.

### DIFF
--- a/src/main/java/org/primefaces/component/picklist/PickList.java
+++ b/src/main/java/org/primefaces/component/picklist/PickList.java
@@ -121,7 +121,7 @@ public class PickList extends PickListBase {
                 message = new FacesMessage(FacesMessage.SEVERITY_ERROR, requiredMessage, requiredMessage);
             }
             else {
-                message = MessageFactory.getMessage(REQUIRED_MESSAGE_ID, FacesMessage.SEVERITY_ERROR, new Object[] { label });
+                message = MessageFactory.getMessage(REQUIRED_MESSAGE_ID, FacesMessage.SEVERITY_ERROR, new Object[] {label});
             }
             facesContext.addMessage(clientId, message);
             setValid(false);
@@ -154,7 +154,7 @@ public class PickList extends PickListBase {
             boolean itemDisabled = isItemDisabled();
             // Check if disabled item has been moved from its former/original list
             if (itemDisabled && !oldEntries.contains(item)) {
-                FacesMessage message = MessageFactory.getMessage(UPDATE_MESSAGE_ID, FacesMessage.SEVERITY_ERROR, new Object[] { label });
+                FacesMessage message = MessageFactory.getMessage(UPDATE_MESSAGE_ID, FacesMessage.SEVERITY_ERROR, new Object[] {label});
                 facesContext.addMessage(clientId, message);
                 setValid(false);
                 break;
--- a/src/main/java/org/primefaces/component/picklist/PickList.java
+++ b/src/main/java/org/primefaces/component/picklist/PickList.java
@@ -105,10 +105,10 @@ public class PickList extends PickListBase {
         super.validateValue(facesContext, newValue);
 
         DualListModel<?> newModel = (DualListModel<?>) newValue;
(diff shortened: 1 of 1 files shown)


Rapster (changes requested):
Please remove these cumbersone "this.xxx". I got rid of all of them recently

melloware:
"this." has been removed.   Man I am not a big fan of the Whitepsace checkstyle rules.

Rapster:
What's wrong with it?

melloware:
@Rapster to me it seems overly cumbersome it just took me 3 attempts to understand why it was wrong I had this...

`new Object[] { label }`

And it wanted...

`new Object[] {label}`

To me that is not making the code better its just going to annoy committers like me who just had to spend 10 minutes fixing whitespace issues.  Just my two cents.

And Sonar is now failing for some reason that is past my CheckStyle checks something about couldn't connect to localhost:9000?

tandraschko:
Yup, looks like sonarcube still failing.
@Rapster can you fix it? Otherwise i have to revert this changes. It's annoying that maven-build fails because something additional like sonarcube fails.

tandraschko:
and @melloware 
i actually also use this in my applications:
`new Object[] { label }`

i just added the check to validate that we use the same whitespaces everywhere. We can also change the definition that we must add spaces after { and before }

Rapster:
No need to revert, it has been correctly built
`The command "mvn install -Dmaven.javadoc.skip=true --batch-mode --show-version" exited with 0.`

Something's wrong with SonarQube server, there might some custom configuration for PR (see https://sonarcloud.io/documentation/integrations/github)

Regarding formatting, we should share Eclipse/IntelliJ/Netbeans formatting preferences. I never had to complain about it since IDE do it for me...

Rapster:
@melloware Try updating your branch please. I tried to disable SonarCloud for PR

Rapster:
I was more thinking, you sycnhronizing your branch with master but that works too :wink:

tandraschko:
@rapster would you Like to retest it? Seems like @cnsgithub doesnt have much time currently 😪

Issue #2679, opened 2017-08-31
XML Exporter: Not Escaping Data


melloware:
## 1) Environment
- PrimeFaces version: 6.2-SNAPSHOT
- Does it work on the newest released PrimeFaces version? Version? NO
- Does it work on the newest sources in GitHub? NO
- Application server + version: Jetty 9
- Affected browsers: All

## 2) Expected behavior

Using the XML DataExporter with a column value like "BMW & Ford" should print in the XML document "BMW & amp; Ford"

## 3) Actual behavior

Value is printed "BMW & Ford" making it invalid XML

## 4) Steps to reproduce

Edited showcase and changed CarService to "BMW & Ford" in the payload.

## 5) Sample XHTML

Showcase dataexporter.xhtml with the above Java change

## 6) Sample bean

Showcase CarService.java edited in Step #4

melloware:
PR coming...

tandraschko:
@melloware we have now 2 escapeXml methods. 1 in XmlUtils, 1 in ComponentUtils. Could you please check which can be removed?

melloware:
My bad. I didn't realize their was one in Componentutils.  I would like to do the following:

1. Modify the one in ComponentUtils to the updated version I just added which adds unicode checks.

2. Move the XmlUtil.escapeTag to ComponentUtils as "escapeXmlTag" and remove the XmlUtil entirely.

Thoughts?

melloware:
Nevermind I see why they split in into XML util because it uses XML Beans which requires POI dependency where Componentutils does not.  I think I will just update the method in ComponentUtils and remove the new one I added in XmlUtil.  I will submit a PR shortly.

tandraschko:
Please note: ComponentUtils#escpeXML is used by our Ajax stuff. So please check first if we really need to escape every which is inside XMLUtils.

melloware:
The only thing added that was not being escaped before was double quote.

case '\"':
       sb.append("&quot;");
        break;

I can't see how that would be valid XML with a double quote even for Ajax.  Should we leave both flavors of the util?

tandraschko:
Not sure. Can you try if it breaks something? It's normally used for adding json to the ajax response via RequestContext.

melloware:
I will do some experimenting and report back.

melloware:
[...]

PR #4045, opened 2018-09-14
Fix #3520: Regression of Issue #2679 for XML column names.

### LINKED ISSUE #3520: XSS despite using ComponentUtils.escapeText

melloware:
When we converted to OWASP for encoders it introduce a regression issue for #2679.

This unit test shows my fix is correctly re-applied.

```java
    @Test
    public void testXml() {
        // Arrange
        final String input = "hello world & >";

        // Assert
        
        // invalid
        Assert.assertEquals("hello world &amp; &gt;", EscapeUtils.forXml(input)); 
        //invalid
        Assert.assertEquals("hello world &amp; >", EscapeUtils.forXmlAttribute(input));
        //valid!!!
        Assert.assertEquals("hello_world_.26._.gt.", EscapeUtils.forXmlTag(input));
    }
```


melloware:
To elaborate more the value of a column may be "Tom & Jerry".  Well escaped XML would make the XML tag:
```xml
<Tom &amp; Jerry>
```
which is not valid XML for a Tag and will fail XML parsers.  The method I restored makes it a valid tag name.
```xml
<Tom_.26_Jerry>
```

tandraschko:
@cnsgithub could you review plz?

PR #4046, opened 2018-09-15
Fix #4036: Picklist add "escape" property.

### LINKED ISSUE #4036: PickList: String values are encoded on submit
1) Environment
- PrimeFaces version: 6.2.8
- Application server + version: JBoss EAP 7.0.8.GA
- Affected browsers: Not depends on the browser

## 2) Expected behavior
`DualModel<String>` containing strings with `&` is supposed to contain same string after submit

## 3) Actual behavior
`DualModel<String>` contains encoded strings with `&amp;` instead of `&` after submit

## 4) Steps to reproduce
Create picklist with `DualModel<String>` filled with strings with `&`

## 5) Sample XHTML
```xml
      <h:form>
        <p:pickList value="#{pickListModel.model}"
                    var="value"
                    itemLabel="#{value}"
                    itemValue="#{value}">
        </p:pickList>

        <p:commandButton value="Submit" action="#{pickListModel.post}" />
      </h:form>
```
## 6) Sample bean
```java
package org.example;

import java.util.ArrayList;
import java.util.Collections;
import javax.el.MethodExpression;
import org.primefaces.model.DualListModel;
import org.springframework.stereotype.Component;

@Component
@ViewScoped
public class PickListModel {

    private DualListModel<String> model;

    public PickListModel() {
        model = new DualListModel<>();
        model.setSource(new ArrayList<>(Collections.singletonList("&source&")));
        model.setTarget(new ArrayList<>(Collections.singletonList("&target&")));
    }

    public DualListModel<String> getModel() {
        return model;
    }

    public void setModel(DualListModel<String> model) {
        this.model = model;
    }

    public void post() {
        System.out.println(model.getTarget());
    }
}
```

melloware:


Rapster (approved):

tandraschko:
Still in holiday but got some hours today ;)

I dont like the fix actually. Do we actually escape to much?
I mean, the & should be encoded to html amp and therefore submitted to the server correctly.
I feel like no extra attr should be required.

tandraschko:
And jfyi: $this is used instead _self since ages 😁

melloware:
Hmmm do you have a proposed fix?  I just looked at what we do in other places like SelectOne to solve the problem of whether someone wants something escaped or not. And I fixed _self to $this. :smile:

tandraschko:
No, i even dont have a computer currently. But how does the & Look on the Client Side html and in the http Post?

melloware:
OK I will create a PrimeFaces Test project and attach here with escape="true" and escape="false" so we can see.

melloware:
Cleaned up my fixes.

Test Case: 
[primefaces-4036.zip](https://github.com/primefaces/primefaces/files/2388558/primefaces-4036.zip)

**Escape=True**
image

image

**Escape=False**
image

image

tandraschko:
The case with HTML elemets is clear/expected! But & should work with escape=true, too. Thats the problem described in the issue, right?

melloware:
So I think you are saying whether or not escape="true" the value of the submitted value should always be & and not &amp;?

tandraschko:
I think so as i assume it works fine with a Inputtext?

tandraschko:
@melloware 
in generell it seems, that html is allowed in inputs and will also be posted un-encoded.
So we can probably remove the JS PrimeFaces.escapeHtml call when setting the input value.
But adding the escape attribute if of course an good enhancement to allow html elements for the picklist.

WDYT? We would introduce another security problem? As its working for a simple input, i have no problem to do it.

melloware:
@tandraschko This was the original fix in 2012 for this https://code.google.com/archive/p/primefaces/issues/4763

Cagatay definitely escaped the value on purpose so you are right I feel like we need to escape the value.  But what if the customer wants HTML in the value?  I pass the "escape" attribute in the WidgetBuilder so its available in the JS call.  Think I should wrap the value?

Back the way I had it like...

```java
itemValue = item.attr('data-item-value')
if (escape) {
   itemValue = PrimeFaces.escapeHTML(item.attr('data-item-value'));
}
```

tandraschko:
yep, lets do it
escape=false is the same behavior as currently and set per default
escape=true enabled html elements and a correctly &amp; handling

melloware:
OK fix committed.

PR #4048, opened 2018-09-18
Fix #4047: Add "aria-required" attribute and refactored ARIA attributes.

### LINKED ISSUE #4047: aria-required missing for input components
## 1) Environment
- PrimeFaces version: 6.2

## 2) Expected behavior

## 3) Actual behavior

## 5) Sample XHTML

melloware:
OK this was a bigger change than I thought but every time I stepped back and looked at it I could make the code cleaner and cleaner.  What this PR does...

1. Adds "aria-required" to any UIInput component that is marked required=true.
2. Cleans up ARIA handling and "disabled" and "readonly" attribute handling to prevent cut and paste code.
3. Add ARIA handling to components that were missing it like Signature, Rating, SelectOneMenu etc etc etc.

Note: I know there is client side JS for "PrimeFaces.skinInput()" which adds aria-disabled and aria-readonly but that is A) only for input components and B) rquires developers to know to call it in every  place.  This PR ensures those attributes are set on any Input component that doesn't even need to call skinInput like Signature and Rating for example but renders the proper ARIA attributes.

@Rapster @mertsincan @tandraschko Please review.  This change seems more invasive than I think it is but cleans up the ARIA code immensely IMHO.

### DIFF
--- a/src/main/java/org/primefaces/component/accordionpanel/AccordionPanelRenderer.java
+++ b/src/main/java/org/primefaces/component/accordionpanel/AccordionPanelRenderer.java
@@ -212,9 +212,9 @@ public class AccordionPanelRenderer extends CoreRenderer {
         writer.startElement("div", null);
         writer.writeAttribute("class", headerClass, null);
         writer.writeAttribute("role", "tab", null);
-        writer.writeAttribute("aria-expanded", String.valueOf(active), null);
-        writer.writeAttribute("aria-selected", String.valueOf(active), null);
-        writer.writeAttribute("aria-label", tab.getAriaLabel(), null);
+        writer.writeAttribute(HTML.ARIA_EXPANDED, String.valueOf(active), null);
+        writer.writeAttribute(HTML.ARIA_SELECTED, String.valueOf(active), null);
+        writer.writeAttribute(HTML.ARIA_LABEL, tab.getAriaLabel(), null);
         writer.writeAttribute("tabindex", tabindex, null);
         if (tab.getTitleStyle() != null) {
             writer.writeAttribute("style", tab.getTitleStyle(), null);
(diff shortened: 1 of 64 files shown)


Rapster:
The code might be invasive, but as you said it gives a better readabiltity, less duplication and it's optimized for JIT. So totally worth it :wink: Basically, I like it :+1: Great job @melloware!

tandraschko:
I wonder If the isreadonly and isdisabled method should only allow uiinput?
Other components should not have this attributes.

Rapster:
Agreed

Rapster (approved):

melloware:
Agreed will update.

melloware:
OK I have changed UIComponent to UIINput for shouldDecode(), isDisabled() and isReadOnly() methods.

I don't know why this still says I have conflicts when I rebased my branch and resolved them all?

tandraschko:
Maybe related because i unintenionally used rebase instead merging on my phone to apply PRs?

tandraschko:
I really Like the changes!
I would merge it If there are no conflicts.

Maybe its because of the Fileupload stuff? Its displayed as changes but its the same as trunk

melloware:
Maybe I rebased my branch incorrectly?  I did "git rebase origin/master" and then resolved the conflicts and updated my branch.  IS that not the right way to fix the conflicts?  I am a NOOB when it comes to this type of stuff with Git.

melloware:
You know what will be easier I will just make a new PR with all the changes for review and close this one when I am done.  I think it will be easier than figuring out what mess I made of this branch. 😄

Rapster:
> Maybe I rebased my branch incorrectly? I did "git rebase origin/master"

I guess origin is your repo, you have to add an upstream pointing on primefaces repo and then rebase from upstream

Issue #4054, opened 2018-09-20
SelectCheckboxMenu: Label defaults to '&nbsp;'


OBreidenbach:
## 1) Environment
- PrimeFaces version: 6.2
- Does it work on the newest released PrimeFaces version? Version? No, 6.2
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) No 
- Application server + version: Wildfly 11, Jetty
- Affected browsers: FF, Chrome

## 2) Expected behavior
If I do not define the label attribute, the label shut stay blank.

## 3) Actual behavior
'&nbsp;' is shown.

## 4) Steps to reproduce
Have a SelectCheckboxMenu without label, multiple attribute.

## 5) Sample XHTML
```
<h:form>
	<h:outputLabel for="basic" value="Basic:" />
	<p:selectCheckboxMenu id="basic" value="#{checkboxView.selectedCities}"
						  filter="true" filterMatchMode="startsWith" panelStyle="width:250px">
		<f:selectItem itemValue="Miami"/>
		<f:selectItem itemValue="London" />
		<f:selectItem itemValue="Paris" />
	</p:selectCheckboxMenu>
</h:form>
```

## 6) Sample bean
```
@ManagedBean
@ViewScoped
public class CheckboxView {
    private String[] selectedCities;

    public String[] getSelectedCities() {
        return selectedCities;
    }

    public void setSelectedCities(String[] selectedCities) {
        this.selectedCities = selectedCities;
    }
}
```

Screenshot:
![selectcheckboxmenuundefinedlabel](https://user-images.githubusercontent.com/11403755/45809743-419c2280-bcc9-11e8-9537-eb7bdb066d1a.png)

In 
https://github.com/primefaces/primefaces/blob/c10e5f8b51b76b4eb052bab00ec1416cba678c9d/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java#L210
the label shout be set to an empty string and not to &amp;nbsp;
or the label must not be escaped.

PR #4055, opened 2018-09-20
#4054 fix default label of SelectCheckboxMenu

OBreidenbach:
#4054


melloware (approved):
After reviewing this component it is the right fix because the Menu itself does not have an "escape" proper the Label can never be HTML in its current state.

tandraschko:
It would be great If you can check the js, too. We often had checks for nbsp

melloware:
I checked the JS and we are OK on this one.  I believe this to be the correct fix.

tandraschko:
As i said, there is a JS logic for &nbsp:
https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js#L181

OBreidenbach:
This code is in the renderItem method in the for loop for the items. So this logic is checking the label of an item not the label that is shown when the menu is closed.

Issue #4057, opened 2018-09-20
SelectCheckboxMenu: Label not updated for disabled component if updateLabel="true"


OBreidenbach:
## 1) Environment
- PrimeFaces version: 6.2
- Does it work on the newest released PrimeFaces version? Version? No, 6.2
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) No
- Application server + version: Wildfly 11, Jetty
- Affected browsers: FF, Chrome

## 2) Expected behavior
If a disabled SelectCheckboxMenu with updateLabel="true" is shown, the label presents the selected items and not the default label.

## 3) Actual behavior
The default label is shown.

## 4) Steps to reproduce
Have a disabled SelectCheckboxMenu with selected items.

## 5) Sample XHTML
```
<h:form>
	<h:outputLabel for="multiple" value="Enabled:" />
	<p:selectCheckboxMenu id="multiple" value="#{checkboxView.selectedCities}" label="test1" updateLabel="true">
		<f:selectItem itemValue="Miami"  itemLabel="Miami"  />
		<f:selectItem itemValue="London" itemLabel="London" />
		<f:selectItem itemValue="Paris"  itemLabel="Paris"  />
	</p:selectCheckboxMenu>
	<br/>
	<h:outputLabel for="multiple" value="Disabled:" />
	<p:selectCheckboxMenu id="multiple2" value="#{checkboxView.selectedCities}" label="test2" updateLabel="true" disabled="true">
		<f:selectItem itemValue="Miami"  itemLabel="Miami"  />
		<f:selectItem itemValue="London" itemLabel="London" />
		<f:selectItem itemValue="Paris"  itemLabel="Paris"  />
	</p:selectCheckboxMenu>
</h:form>
```

## 6) Sample bean
```
@ManagedBean
@ViewScoped
public class CheckboxView {

    private String[] selectedCities;

    public String[] getSelectedCities() {
        return selectedCities;
    }

    public void setSelectedCities(String[] selectedCities) {
        this.selectedCities = selectedCities;
    }
}
```
Screenshot:
![selectcheckboxmenulabelnotupdated](https://user-images.githubusercontent.com/11403755/45817931-7ff10c00-bce0-11e8-9f51-140aee240523.png)

Rapster:
Do you mean that you expect to see "London, Paris" on the disabled check menu after selecting items from enabled checkbox menu?

[...]

PR #4058, opened 2018-09-20
#4057 fix updateLabel of disabled SelectCheckboxMenu

OBreidenbach:
#4057; Update label even if the SelectCheckboxMenu is disabled.


tandraschko:
Setting the CSS directly via js is a -1 for me

Maybe someone can assist, i dont have time until next week

melloware:
Let me look at this one closer.

OBreidenbach:
I just copied some lines from the if to the else case. So their is nothing really new here.

tandraschko:
Oh i see, so we have to improve it later.

PR #4060, opened 2018-09-20
Fix #4047:  Add "aria-required" attribute and refactor ARIA attribute…

### LINKED ISSUE #4047: aria-required missing for input components

melloware:
OK this update is even better than the previous PR because:

1. It only adds code to InputRenderer not CoreRenderer since these should only be UIInput components.

2. Uses shouldDecode() in components which actually fixes a few places that were checking for "disabled" only.

3. PickList and Signature now properly extend InputRenderer.

### DIFF
--- a/src/main/java/org/primefaces/renderkit/InputRenderer.java
+++ b/src/main/java/org/primefaces/renderkit/InputRenderer.java
@@ -17,12 +17,15 @@ package org.primefaces.renderkit;
 
 import java.io.IOException;
 import javax.faces.component.UIComponent;
+import javax.faces.component.UIInput;
 import javax.faces.context.FacesContext;
+import javax.faces.context.ResponseWriter;
 import javax.faces.convert.Converter;
 import javax.faces.convert.ConverterException;
 import org.primefaces.component.api.RTLAware;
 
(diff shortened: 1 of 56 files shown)


src/main/java/org/primefaces/renderkit/InputRenderer.java
```diff
@@ -39,16 +42,74 @@ public Object getConvertedValue(FacesContext context, UIComponent component, Obj
...
     }
+
+    /**
+     * Adds "aria-required" if the component is required.
+     *
+     * @param context the {@link FacesContext}
+     * @param component the {@link UIInput} component to add attributes for
+     * @throws IOException if any error occurs writing the response
+     */
+    protected void renderRequired(FacesContext context, UIInput component) throws IOException {
```
Rapster: Rename into renderARIARequired if it only write HTML.ARIA_REQUIRED

melloware: Fixed!

PR #4067, opened 2018-09-21
Fix #4066: SavedState refactor UIRepeat/UITabPanel.

### LINKED ISSUE #4066: SavedState: Reuse existing model object
Class org.primefaces.component.api.SavedState can be modified and then duplicate internal copies can be removed from:

UITabPanel
UIRepeat

melloware:
Reduces copy and paste code to a class that already existed.

### DIFF
--- a/src/main/java/org/primefaces/component/api/SavedState.java
+++ b/src/main/java/org/primefaces/component/api/SavedState.java
@@ -17,8 +17,15 @@ package org.primefaces.component.api;
 
 import java.io.Serializable;
 
+import javax.faces.component.EditableValueHolder;
+
+/**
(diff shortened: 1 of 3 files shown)


tandraschko:
Looks good, i will apply it soon
i will just add a memo for me here to check the supresswarnings

melloware:
Yes I didn't understand the need for those SuppressWarnings either.

Rapster:
I understand the need of using suppress warning in this case (since Object is not serializable) so `@SuppressWarnings("serial")` should do. But I wonder in which case it needs to be serialized... (probably internally by jsf impl)

PR #4069, opened 2018-09-22
Fix #1689: Add "aria-invalid" to invalid components.

### LINKED ISSUE #1689: Improve component accessibility - aria-invalid
To improve the accessibility of input components you shoul add the attribute "aria-invalid" to the components, which should be set when the component is invalid (e.g. failed validation).

https://www.w3.org/WAI/GL/wiki/Using_Aria-Invalid_to_Indicate_An_Error_Field

It is necessary to make forms with validation accessible!

Thanks!

melloware:
This PR follows the ARIA spec exactly. it does the following.

1. Marks Input components with aria-invalid=true if they are invalid.

2. If a `<p:message>` is attached to the Input field for validation message then aria-describedby is added to the Input component pointing to the `<span>` containing the validation error message.

I had to slightly modify the Message component to accomplish this.

### DIFF
--- a/src/main/java/org/primefaces/component/message/MessageRenderer.java
+++ b/src/main/java/org/primefaces/component/message/MessageRenderer.java
@@ -155,7 +155,7 @@ public class MessageRenderer extends UINotificationRenderer {
     }
 
     protected void encodeScript(FacesContext context, Message uiMessage, UIComponent target) throws IOException {
-        boolean tooltip = uiMessage.getDisplay().equals("tooltip") ;
+        boolean tooltip = "tooltip".equals(uiMessage.getDisplay());
         if (tooltip || uiMessage.isShowDetail()) {
             String clientId = uiMessage.getClientId(context);
             String targetClientId = (target instanceof InputHolder) ? ((InputHolder) target).getInputClientId() : target.getClientId(context);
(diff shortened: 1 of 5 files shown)


src/main/java/org/primefaces/component/message/MessageRenderer.java
```diff
@@ -154,13 +155,15 @@ protected void encodeIcon(ResponseWriter writer, String severity, String title,
     }
 
     protected void encodeScript(FacesContext context, Message uiMessage, UIComponent target) throws IOException {
-        if (uiMessage.getDisplay().equals("tooltip")) {
+        boolean tooltip = uiMessage.getDisplay().equals("tooltip") ;
```
Rapster: `"tooltip".equals(uiMessage.getDisplay())` is better 😉

tandraschko:
We have to check this with CSV too

src/main/java/org/primefaces/component/message/MessageRenderer.java
melloware: Good suggestion. Change applied!

melloware:
Good idea @tandraschko I just tested with CSV and made the fixes.  

1. @NotNull was not adding data-p-required to the fields so thus aria-required would never be added.  Fixed this.

2. in validation.js now check for valid and required and add the correct aria attributes based on those fields just like we do on the server side.

tandraschko:
@melloware 
about 1):
@NotNull must NOT indicate that it's required. See OutputLabelRenderer line 180

melloware:
You are right.  I just reverted the@NotNull fix.

tandraschko:
Ok, we can solve this if we check for "isInterpretEmptyStringAsNull" but i would do it in a seperate ticket?!

PR #4071, opened 2018-09-23
Fix #4047: aria-required changes based on testing.

### LINKED ISSUE #4047: aria-required missing for input components

melloware:
Changes from @rho66  

Please review them as the Inplace needed special handling and ColorPicker I deviated from the suggested fix to put the attributes on both the button mode and inline mode.

### DIFF
--- a/src/main/java/org/primefaces/component/chips/ChipsRenderer.java
+++ b/src/main/java/org/primefaces/component/chips/ChipsRenderer.java
@@ -107,6 +107,7 @@ public class ChipsRenderer extends InputRenderer {
         if (inputStyle != null) {
             writer.writeAttribute("style", inputStyle, null);
         }
+        renderAccessibilityAttributes(context, chips);
 
         if (values != null && !values.isEmpty()) {
             Converter converter = ComponentUtils.getConverter(context, chips);
@@ -143,8 +144,13 @@ public class ChipsRenderer extends InputRenderer {
         writer.writeAttribute("class", "ui-widget", null);
         writer.writeAttribute("name", inputId, null);
         writer.writeAttribute("autocomplete", "off", null);
+        if (disabled) {
+            writer.writeAttribute("disabled", "disabled", "disabled");
+        }
+        if (chips.isReadonly()) {
+            writer.writeAttribute("readonly", "readonly", "readonly");
+        }
 
-        renderAccessibilityAttributes(context, chips);
         renderPassThruAttributes(context, chips, HTML.INPUT_TEXT_ATTRS_WITHOUT_EVENTS);
         renderDomEvents(context, chips, HTML.INPUT_TEXT_EVENTS);
 
--- a/src/main/java/org/primefaces/component/inplace/InplaceRenderer.java
+++ b/src/main/java/org/primefaces/component/inplace/InplaceRenderer.java
@@ -16,7 +16,7 @@
 package org.primefaces.component.inplace;
 
 import java.io.IOException;
-import java.util.Iterator;
+import java.util.List;
 
 import javax.faces.component.UIComponent;
 import javax.faces.component.UIInput;
@@ -200,9 +200,9 @@ public class InplaceRenderer extends CoreRenderer {
     }
 
     private boolean hasRequiredChildren(Inplace inplace) {
-        Iterator itr = inplace.getChildren().iterator();
-        while (itr.hasNext()) {
-            UIComponent c = (UIComponent) itr.next();
+        List<UIComponent> children = inplace.getChildren();
+        for (int i = 0; i < children.size(); i++) {
+            UIComponent c = children.get(i);
             if (c instanceof UIInput && ((UIInput) c).isRequired()) {
                 // at least one input component with attribute required found
                 return true;
--- a/src/main/java/org/primefaces/component/inplace/InplaceRenderer.java
+++ b/src/main/java/org/primefaces/component/inplace/InplaceRenderer.java
@@ -16,8 +16,10 @@
 package org.primefaces.component.inplace;
 
 import java.io.IOException;
+import java.util.Iterator;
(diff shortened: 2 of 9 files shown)


src/main/java/org/primefaces/component/inplace/InplaceRenderer.java
```diff
@@ -193,4 +198,16 @@ public void encodeChildren(FacesContext context, UIComponent component) throws I
     public boolean getRendersChildren() {
         return true;
     }
+
+    private boolean hasRequiredChildren(Inplace inplace) {
+        Iterator itr = inplace.getChildren().iterator();
+        while (itr.hasNext()) {
```
Rapster: Use index loop instead please

melloware: Fixed.

src/main/java/org/primefaces/component/chips/ChipsRenderer.java
```diff
@@ -143,8 +144,13 @@ protected void encodeMarkup(FacesContext context, Chips chips) throws IOExceptio
         writer.writeAttribute("class", "ui-widget", null);
         writer.writeAttribute("name", inputId, null);
         writer.writeAttribute("autocomplete", "off", null);
+        if (disabled) {
+            writer.writeAttribute("disabled", "disabled", "disabled");
+        }
+        if (chips.isReadonly()) {
+            writer.writeAttribute("readonly", "readonly", "readonly");
+        }
```
melloware, line 153: I put these back because the accessibility attributes were moved to the UL instead of each LI

src/main/java/org/primefaces/component/inplace/InplaceRenderer.java
```diff
@@ -193,4 +198,16 @@ public void encodeChildren(FacesContext context, UIComponent component) throws I
     public boolean getRendersChildren() {
         return true;
     }
+
+    private boolean hasRequiredChildren(Inplace inplace) {
+        List<UIComponent> children = inplace.getChildren();
+        for (int i = 0; i < children.size(); i++) {
+            UIComponent c = children.get(i);
+            if (c instanceof UIInput && ((UIInput) c).isRequired()) {
```
Rapster, line 206: I think it needs support of composite also. See https://github.com/primefaces/primefaces/blob/271a52100de06d8c8a339025e9a0b4c5500e31a0/src/main/java/org/primefaces/component/outputlabel/OutputLabelRenderer.java#L70

melloware: I think you might be right, I have never seen that ContextCallback code before.  I would have to study it some more to understand what it's doing.

tandraschko:
also i don't think the changes of Chips e.g. are correct
you would now render disabled and readonly attributes on UL
please review everything and provide a second PR to fix such cases

tandraschko:
Thats actually the case for many components - this must be fixed as it produced invalid html
@melloware  :P

often disabled/readonly is also rendered on divs.
probably the aria attributes should also only applied to the input elements.

rho66:
[MultiSelectListboxRenderer.aria_required.patch.gz](https://github.com/primefaces/primefaces/files/2411712/MultiSelectListboxRenderer.aria_required.patch.gz)
Probably it would be better to use `renderARIARequired(context, chips);` where we only want to render aria-required attribute!

I am with you that aria attributes should be applied to input elements only, but this seems to me as the easiest solution to fullfill my requirement of highlighting required input elements! Better way would be to create a separate style which could also be applied to renderd divs for some elements!

Find below a patch for MulitSelectListbox which applies aria-required to div element.

melloware:
@rho66 but is this for something you are writing or how the ScreenReader is reading it?  Because according to the ARIA spec aria-required=true is NOT allowed on DIVs which is what you previous patches had me doing?

tandraschko:
i don't care about styling for now - this is something special like adding a css class and is another feature.
we should only care about the accessability attributes and ARIA for now.

readonly and disabled are only allowed on inputs BUT i'm not sure about the aria-required as it seems to be allowed on UL's. So not sure about DIVs.

melloware:
Agreed.  I only care about following the ARIA spec.

rho66:
please see my comment on https://github.com/primefaces/primefaces/issues/4047

should i create a new PR for a required styling css class?

tandraschko:
I dont like to blow up the markup for such a special case.
But i can post a snippet later, which you can include in your application.
If there are more requests in the Future, we can still add it to PrimeFaces

rho66:
Thanks, I would appreciate it to receive your snippet!

tandraschko:
https://gist.github.com/tandraschko/6875a3484b69f3010396370305713e84
something like this should work

`        <system-event-listener>
            <system-event-listener-class>xxx.Blub</system-event-listener-class>
            <system-event-class>javax.faces.event.PreRenderComponentEvent</system-event-class>
        </system-event-listener>`

rho66:
Looks interesting - where do you find such things?!

melloware:
@tandraschko has been doing this so long I learn something new almost every week on how things work. :smile:

tandraschko:
:D yep :D

rho66:
@tandraschko I tried your approach with the SystemEventListener and it worked for most input components, but not for all! Especially not for those, which are rendered with some DIV elements. The listener applies the additional style class only to the input components but has no influence on the real rendering of the component. So this approach is unfortunately not a perfect solution for my requirement to highlight required input fields with a special border color!

tandraschko:
Should work for every component as every component has a styleClass on the root element.
Check your css rules ;)

PR #4072, opened 2018-09-23
Fix #3019: Add ARIA properties to SelectCheckBoxMenu.

### LINKED ISSUE #3019: selectCheckboxMenu: accessibility
## 1) Environment
- PrimeFaces version: Primefaces 6.1
- Application server + version: JBoss EAP 7.0
- Affected browsers: Firefox ESR 52.5.0 (64-Bit)
- Screenreader: JAWS 18.0

## 2) Expected behavior
<p:selectCheckboxMenu> component should be read by screenreader similiar like <p:selectOneMenu> component. There it says "`label` dropdown, `selection`, enter text"

## 3) Actual behavior
<p:selectCheckboxMenu> component is read as "Read only input" by screenreaders.

## 4) Steps to reproduce
Read <p:selectCheckboxMenu> component with screenreader

melloware:
Fix #3019: Add ARIA properties to SelectCheckBoxMenu.

### DIFF
--- a/src/main/java/org/primefaces/renderkit/InputRenderer.java
+++ b/src/main/java/org/primefaces/renderkit/InputRenderer.java
@@ -112,4 +112,19 @@ public abstract class InputRenderer extends CoreRenderer {
         }
     }
 
+    /**
+     * Adds ARIA attributes if the component is "role=combobox".
+     *
+     * @param context the {@link FacesContext}
+     * @param component the {@link UIInput} component to add attributes for
+     * @throws IOException if any error occurs writing the response
+     * @see https://www.w3.org/TR/wai-aria-practices/#combobox
+     */
+    protected void renderARIACombobox(FacesContext context, UIInput component) throws IOException {
+        ResponseWriter writer = context.getResponseWriter();
+        writer.writeAttribute("role", "combobox", null);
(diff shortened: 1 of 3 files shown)


src/main/java/org/primefaces/renderkit/InputRenderer.java
```diff
@@ -112,4 +112,19 @@ protected void renderARIARequired(FacesContext context, UIInput component) throw
         }
     }
 
+    /**
```
Rapster, line 115: If it's really meant for combobox, what about putting it in `SelectRenderer`? Or you think it might be needed by renderer inheriting from InputRenderer?

tandraschko:
@melloware i removed the necessary generics stuff - i had a warning because of a missing documentation of \<T\>

melloware:
Thanks @tandraschko!  And @Rapster from reviewing the ARIA docs and the PF codebase i think the only two components this specific ARIA role applies to is these two.

Rapster:
Yeah so I think this method `renderARIACombobox` should be in SelectRenderer instead. Since it's been merged already, I'll do it

tandraschko:
Actually we should think about moving to CoreRenderer or something as even DataTable e.g. has comboboxes.
I would leave it as it is for now.

Issue #4073, opened 2018-09-24
InputMask: wrong description for "lang" attribute in primefaces-p.taglib.xml


OBreidenbach:
The lang attribute has the same description as the label attribute.
https://github.com/primefaces/primefaces/blob/271a52100de06d8c8a339025e9a0b4c5500e31a0/src/main/resources/META-INF/primefaces-p.taglib.xml#L8727-L8738

PR #4074, opened 2018-09-24
#4073 fix description of lang attribute for inputMask component

OBreidenbach:
#4073


tandraschko:
I would remove "for this component" as the context is always "this component"

tandraschko:
will remove it after applying the PR

PR #4079, opened 2018-09-24
Fix #3935 Autocomplete: avoid to focus a hidden element.

### LINKED ISSUE #3935: AutoComplete: pressing tab focuses invisible component in multiple mode
## 1) Environment
- PrimeFaces version: 6.2.7
- Affected browsers: At least Chrome and Edge.

## 2) Expected behavior
Focus changed to next input component.

## 3) Actual behavior
Focus seems to "disappear". Actually changed to internal select html element. This select does not seem to exist whenever multiple is set to _false_.

## 4) Steps to reproduce
Reproduce simply be using showcase: https://www.primefaces.org/showcase/ui/input/autoComplete.xhtml

Click the "Multiple" variant and press the tab key.

wsaca:
This PR is to resolve the issue #3935, please check it.

### DIFF
--- a/src/main/java/org/primefaces/component/autocomplete/AutoCompleteRenderer.java
+++ b/src/main/java/org/primefaces/component/autocomplete/AutoCompleteRenderer.java
@@ -274,6 +274,7 @@ public class AutoCompleteRenderer extends InputRenderer {
         writer.writeAttribute("name", id, null);
         writer.writeAttribute("multiple", "multiple", null);
         writer.writeAttribute("class", "ui-helper-hidden-accessible", null);
+        writer.writeAttribute("tabindex", "-1", null);
 
         if (ac.isDisabled()) {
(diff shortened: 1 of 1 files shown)


melloware:
@tandraschko @Rapster  I wonder if this is something we can generalize where if the element is being styled for "ui-helper-hidden-accessible" it should always add tabindex=-1 ?

wsaca:
There are elements rendered like:
```
  <input type="checkbox">
</div>
```
For this example the input should be able to take the focus, render tabindex on a DIV is not useful.

tandraschko:
AFAIR it's a helper for making a component focusable/tabbable, so not sure if the fix is correct.

wsaca:
Yes, I think like you. but why has the select this style applied? was it applied only to hide it? I tested it with tabindex=-1 without to find any problem.

tandraschko:
@mertsincan can you check this? I don't want to break something

PR #4103, opened 2018-10-03
donate ajax session timeout listener

eduarddrenth:


Rapster:
Please create an issue first, so you can explain what this is about

eduarddrenth:
fixes #4109

tandraschko:
Will close the PR for now. You first need to check the root problem.
And a -1 for me to include such a solution to PF - a ViewExpired thought be thrown.

eduarddrenth:
This: https://myfaces.apache.org/wiki/core/user-guide/jsf-and-myfaces-howtos/managing-errors---infos---warnings/viewexpiredexception-and-session-expiry.html is interesting. I'll see how payara does this and what config options I have there and how that cooperates with PF exception handling.

PR #4118, opened 2018-10-04
Fix #4110 Use css class to hide scroll buttons.

### LINKED ISSUE #4110: TabView: Allow override the ui-tabs-navscroller padding-left
## 1) Environment
- Affected browsers: Any

## 2) Expected behavior
A way to easily override the "ui-tabs-navscroller" padding-left class to be able to increase the width in the "ui-tabs-navscroller-btn".

## 3) Actual behavior
It is not posible because the "padding-left" is set using javascript in the next function:
https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/tabview/tabview.js#L225

It is always 18px and we can't override that value because there is not way to know if the navigation buttons should be shown or not.

wsaca:
I propose this change to resolve #4110, I'm trying to avoid to break custom themes or templates. You could test it or give any suggestion.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/tabview/tabview.css
+++ b/src/main/resources/META-INF/resources/primefaces/tabview/tabview.css
@@ -152,6 +152,11 @@
 
 .ui-tabs .ui-tabs-navscroller {
     overflow: hidden;
+    padding-left: 18px;
+}
+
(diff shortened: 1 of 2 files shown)


tandraschko:
Like it, thanks

Issue #4122, opened 2018-10-05
PanelGrid: component causes invalid HTML and broken AJAX communication


abelet:
## Environment ##
- PrimeFaces version: tested with 6.2, 6.2.9
- JSF 2.3

## Maven project demonstrating the problem ##
[jsf-ajax-failure.zip](https://github.com/primefaces/primefaces/files/2450527/jsf-ajax-failure.zip)

## Problem with the implementation ##

JSF components are not perfect. Their implementations can have some faults. For example the primefaces PanelGrid component is liable for not closing table rows. It does not close the last row if “actual number of columns” cannot be divided by the “value of columns attribute” without a remainder.   For example the following JSF fragment:

```xml
<p:panelGrid columns="2" >
    <h:outputLabel value="x"/>
    <h:outputLabel value="x"/>
    <h:outputLabel value="x"/>
</p:panelGrid>
```
produces a HTML table missing a closing TR tag:

```html
<table id="form:j_idt5" class="ui-panelgrid ui-widget" role="grid">
    <tbody>
        <tr class="ui-widget-content" role="row">
            <td role="gridcell" class="ui-panelgrid-cell"><label>x</label></td>
            <td role="gridcell" class="ui-panelgrid-cell"><label>x</label></td>
        </tr>
        <tr class="ui-widget-content" role="row">
            <td role="gridcell" class="ui-panelgrid-cell"><label>x</label></td>
    </tbody>
</table>
```

The JSF fragment cannot be blamed for the problem. It is a natural need to have 2 components in the first and 1 in the second row. So the PanelGrid implementation is wrong, but the improper HTML will be rendered perfectly by the browser. Unfortunately the JSF implementation can make from this mistake an unrecoverable error. But before describing how and in in which circumstances it is done, I have to outline the relevant elements of the JSF implementation.

## Description of the JSF implementation relevant parts ##

The com.sun.faces.renderkit.html_basic.HtmlResponseWriter provides a functionality for renaming HTML tag names generated by JSF Components. With this implementation the following JSF code:

```xml
<ui:composition xmlns="http://www.w3.org/1999/xhtml"
   xmlns:h="http://java.sun.com/jsf/html"
   xmlns:ui="http://java.sun.com/jsf/facelets"
   xmlns:pass="http://xmlns.jcp.org/jsf/passthrough"
>
      <h:outputLabel value="Welcome" pass:label_text="Welcome" pass:elementName="span"/>
</ui:composition>
```
[...]

PR #4124, opened 2018-10-05
Fixing issue #4122

abelet:
Fixing the issue descibed here: https://github.com/primefaces/primefaces/issues/4122

### DIFF
--- a/src/main/java/org/primefaces/component/panelgrid/PanelGridRenderer.java
+++ b/src/main/java/org/primefaces/component/panelgrid/PanelGridRenderer.java
@@ -111,6 +111,7 @@ public class PanelGridRenderer extends CoreRenderer {
     }
 
     public void encodeDynamicBody(FacesContext context, PanelGrid grid, int columns) throws IOException {
+        boolean rowIsClosed = true;
         ResponseWriter writer = context.getResponseWriter();
         String columnClassesValue = grid.getColumnClasses();
         String[] columnClasses = columnClassesValue == null ? new String[0] : columnClassesValue.split(",");
@@ -123,6 +124,7 @@ public class PanelGridRenderer extends CoreRenderer {
 
             int colMod = i % columns;
             if (colMod == 0) {
+                rowIsClosed = false;
                 writer.startElement("tr", null);
                 writer.writeAttribute("class", PanelGrid.TABLE_ROW_CLASS, null);
                 writer.writeAttribute("role", "row", null);
@@ -143,6 +145,10 @@ public class PanelGridRenderer extends CoreRenderer {
             if (colMod == 0) {
                 writer.endElement("tr");
             }
+        }
+
+        if (!rowIsClosed) {
+            writer.endElement("tr");
         }
     }
 


tandraschko:
I would do a small change.
1) rowIsClosed = false per default
2) add rowIsClosed=true in line 146

WDYT?

abelet:
I guess, if there is no row, just an empty tbody ( when grid.getChildren() is empty ), a closing tr will be placed on the output, which will be invalid again

abelet:
I hope I understand well, what do you mean.

If I undertand well, the tdody will be always writen out, and the encodeDynamicBody will be always called, if the columns attribute is larger then 0. But it is not the actual number of elements in the PanelGrid component. So if you specify columns=2, and do not put any single component inside the PanelGrid, wrong output is generated, if you do your suggestions. 

```java
    public void encodeTableBody(FacesContext context, PanelGrid grid, int columns) throws IOException {

        writer.startElement("tbody", grid);

        if (columns > 0) {
            encodeDynamicBody(context, grid, grid.getColumns());
        }
        else {
            encodeStaticBody(context, grid);
        }

        writer.endElement("tbody");
    }
```

melloware (approved):

tandraschko:
Not sure, i just don't like the naming probably. It feels not correct as the default value is true (and we don't know yet if the row will be closed really)

abelet:
Maybe the name isn't perfect, but this variable is for signaling if we should close a row in the end or not. I’m quite sure that the modification you suggested would result a closing tr without an opening one, if there is no item in the panelGrid. I haven't tested it, but seeing the code, I'm quite certain.

tandraschko:
WDYT about if we count the closing and opening trs?
and then do:

if (opening > closing) {
instead of
if (!rowIsClosed) {

I think it's much easier to read and to understand the actual logic.

abelet:
Your suggestion seems right to me.

I’d like to ask your opinion on a situation.

The error in the panelGrid is obvious. But I’m even more concerned about the JSF 2.3 implementation, which I think is very dangerous with this incorrectly implemented tag renaming functionality. Besides my case there is a similar one (https://github.com/jboss/mojarra/issues/21), and I’m sure as people starting to move towards JSF 2.3 it is likely  that the number problems it causes will rise. I’m worried about the future magical errors it will cause in our systems.  I opened an issue about this bug at the JSF project:

https://github.com/eclipse-ee4j/mojarra/issues/4488

The JSF development team closed this issue despite the fact how vulnerable JSF become to component rendering problems, like this one we are about to fix. Do you think the behavior of the JSF implementation should be accepted?

tandraschko:
I always have 2 opinions in such cases :D

1) as user:
it would be great if Mojarra would be more safe

2) as framework dev (i'm also working on MyFaces):
i can understand the Mojarra team and the component libs must produce correct html

I will apply my change now, hope you can retest it then.

abelet:
The problem with MyFaces does not exists. Unfortunately wildfly uses mojarra.

tandraschko:
Yes, i'm aware of that. But the goal should be that PrimeFaces always produces correct HTML and no "hacks/failsafe features" should be required in the JSF impls.

abelet:
Actually there is no need for doing hacks. The HtmlResponseWriter’s endElement method do not receive the componentForElement parameter, and has no information about passthrough attributes, so the elementNames stack away was introduced as a hack, causing troubles. I have several ideas how to modify the current implementation to become safer. But my idea isn’t perfect, and I still hope the JSF team will solve this problem better than I can.
Using the JSF 2.3 implementation, which mixes the HTLM response up, the application developer won't understand what happened, what went wrong. For example @erickdeoliveiraleal 
had to switch back to JSF 2.2.

Issue #4142, opened 2018-10-16
Schedule: Cannot render events with special characters like ampersand (&), or minus(-)


anhquande:
## 1) Environment
- PrimeFaces version: 6.3.0-SNAPSHOT
- Does it work on the newest released PrimeFaces version? Version? yes (work on 6.2.0)
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) no (does not work on 6.3.0-SNAPSHOT)
- Application server + version: does not matter
- Affected browsers: does not matter

## 2) Expected behavior
- All events should be visible in calendar
## 3) Actual behavior
- No events are shown
- `<p:log/>` shows a crash when parsing JSON: 

> SyntaxError: Unexpected token x in JSON at position
![bug-schedule](https://user-images.githubusercontent.com/4417654/47013350-26e19000-d147-11e8-8142-c22656f1d994.png)

## 4) Steps to reproduce
- Checkout the primefaces 6.3 and deploy locally. Read [HOW-TO-BUILD-FROM-SOURCE](https://github.com/primefaces/primefaces/wiki/Building-From-Source)

- Checkout from the test project
`git clone git@github.com:anhquande/primefaces-test.git && cd primefaces-test && git checkout calendar-test
`

- execute from command line
`mvn jetty:run-exploded
`

## 5) Sample XHTML

```
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml"
      xmlns:ui="http://java.sun.com/jsf/facelets"
      xmlns:f="http://java.sun.com/jsf/core"
      xmlns:p="http://primefaces.org/ui"
      xmlns:h="http://java.sun.com/jsf/html">

<h:head>
    <title>PrimeFaces Test</title>
    <h:outputScript name="js/calendar-config.js" library="myprimefaces"/>
</h:head>
<h:body>

    <h:form>

        <p:log/>
        <p:growl id="messages" showDetail="true"/>
        <p:schedule id="schedule" value="#{scheduleView.eventModel}" widgetVar="myschedule"
                    timeZone="#{scheduleView.timeZone}"
                    locale="de"
                    rightHeaderTemplate="month,agendaWeek,agendaDay"
                    leftHeaderTemplate="today,prev,next"
                    centerHeaderTemplate="title"
                    showHeader="true"
                    aspectRatio="1"
                    draggable="false"
                    initialDate="#{scheduleView.getInitialDate()}"
[...]

PR #4151, opened 2018-10-22
#4142, fix JSON encoding for scheduler events

OBreidenbach:
#4142, Schedule: Cannot render events with special characters like ampersand (&), or minus(-)


tandraschko:
what about the "!= null" checks? i would re-add them to avoid unnecessary json content.

OBreidenbach:
There is already a not null check in the JSONObject. So they not needed here. The JSON-Response does not increase. Only the order of the attributes changes. They are now alphabetical ordered.

cnsgithub (approved):

melloware (approved):

Issue #4154, opened 2018-10-23
schedule: URL events enable phishing attacks


cnsgithub:
## 1) Environment
- PrimeFaces version: 6.3-snapshot

## 2) Expected behavior
Clicking at a schedule URL event should not enable the target site to access the window opener.

## 3) Actual behavior
By default, the target site is opened in a new browser window/tab by using `window.open(<url>, '_blank')` javascript. Therefore the target site has access to `window.opener` and is able to modify the `location ` for example, leading to potential phishing attacks.

## 4) Steps to reproduce

1. Just point the birthday party event in the showcase's `ScheduleView.java` to the example target site https://cnsgithub.github.io/foo/:
```java
	    DefaultScheduleEvent birthdayParty = new DefaultScheduleEvent("Birthday Party", today1Pm(), today6Pm());
	    birthdayParty.setUrl("https://cnsgithub.github.io/foo/");
```

2. Open up the [schedule showcase](http://localhost:8080/showcase/ui/data/schedule.xhtml) and click at the birthday party event.

3. Enjoy the party: the target site will show ![image](https://user-images.githubusercontent.com/13161711/47355535-ec856f00-d6c1-11e8-91c2-71dc3d41b537.png)
 while the showcase's location will have changed to a potential phishing site (in this case the genuine PrimeStore).

## 5) Sample XHTML
n/a

## 6) Sample bean
See 4).

## 7) Mitigation
Whenever links to untrusted sites are rendered with `target="_blank" ` or opened via javascript `window.open(<url>, '_blank')` accordingly, we must prevent access to `window.opener` by either setting `rel="noopener noreferrer" `or `opener=null`.

Further reading:

- https://cwe.mitre.org/data/definitions/1022.html
- https://mathiasbynens.github.io/rel-noopener/

Stay tuned, I'll try to provide a PR...

Could someone please add the security label?

cnsgithub:
PR: https://github.com/primefaces/primefaces/pull/4156

melloware:
Nice work.  Searching for window.open in the PF codebase returned these places..

ButtonRenderer.java
contentflow.js
gmap.js
printer.js

Not sure if it's vulnerable anywhere but ButtonRenderer definitely looks like a candidate?

cnsgithub:
Yep, that would have been the next step. :+1: `ButtonRenderer` is indeed another candidate. Feel free to fix those as well.

cnsgithub:
[...]

PR #4156, opened 2018-10-23
fix: schedule: URL events enable phishing attacks (#4154)

cnsgithub:
After applying this PR you can perform the reproduction steps from the issue once again. Then you'll notice the redirect on the opener is gone and the target site is showing 
image

### DIFF
--- a/src/main/java/org/primefaces/component/schedule/ScheduleBase.java
+++ b/src/main/java/org/primefaces/component/schedule/ScheduleBase.java
@@ -68,7 +68,8 @@ abstract class ScheduleBase extends UIComponentBase implements Widget, ClientBeh
         weekNumberCalculator,
         nextDayThreshold,
         slotEventOverlap,
-        urlTarget
+        urlTarget,
+        noOpener
     }
 
     public ScheduleBase() {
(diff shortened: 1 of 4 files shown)


melloware (approved):

cnsgithub:
Successfully tested with latest Firefox, Chrome and IE 11.

tandraschko:
Why not doing it like in the owasp cheatsheet?
https://www.owasp.org/index.php/HTML5_Security_Cheat_Sheet#Tabnabbing

cnsgithub:
@tandraschko Thanks for pointing out this `windowFeatures ` parameter of `window.open`. However, I just tested with this piece of code and it doesn't seem to work correctly:
```javascript
                var targetWindow = window.open(calEvent.url, $this.cfg.urlTarget, 'noopener,noreferrer');
                targetWindow.opener = null;
```
Both windows are getting redirected to https://cnsgithub.github.io/foo/ then in latest Firefox and Chrome. It just seems to be working in Internet Explorer.

tandraschko:
I wonder if we should make it configurable in general (but enabled per default) as some portals may depend on it (same for the link rel="noopener noreferrer" later)

cnsgithub:
Hi @tandraschko, what do you want to enable per default - security or phishing? :jack_o_lantern: 

Do you have any preferences for attribute names we could introduce for `p:schedule`, e.g. `noopener="true|false"` or `opener="false|true"`? Or do you prefer it even more fine-grained, having it attached to the `ScheduleEvents`?

Or were you thinking of having it globally configured just once in `web.xml`? I think that might be too coarse-grained.

tandraschko:
pishing per default ofc :D

I would call it "noopener" with "true" per default

Hmm, i don't know. I would do it on component level for now, not for each event.

cnsgithub:
@tandraschko Added attribute `p:schedule noOpener="true"`. This needs to be documented in the user guide.

PR #4160, opened 2018-10-23
Fix #4159 Enable cancel button by file

### LINKED ISSUE #4159: FileUpload: Enable cancel button by file
## 1) Environment
- Affected browsers: Any

## 2) Expected behavior
When the upload button is pressed the cancel button of each file should be enabled.

## 3) Actual behavior
When the upload button is pressed the cancel button of each file is disabled.

## 4) Steps to reproduce
1. This case can be tested in the showcase.

This issue is an extension to improve the PR [#3862](https://github.com/primefaces/primefaces/pull/3862)

wsaca:
This is a proposal to resolve #4159, the changes are related to:
- Enable the cancel button by file.
- Add an identifier for each file to avoid problems when:
  - Same file is added twice and one of these is canceled.
  - We use the jquery index function, if we use it we could have problems when two files are canceled at same time because of the fadeOut animation time.
- The abort function is called only when the upload button was pressed.
- The cancel button is disabled when the file is being removed to avoid it is clicked needlessly.

I would like you verify this change.


tandraschko:
@melloware are you interessted in retesting?

melloware (approved):

melloware:
I poked around testing it and nothing seems to be broken.   The button enabled and disabled as I expected it to.

PR #4162, opened 2018-10-23
Fix #4161 Append overlay panel in the body by default.

### LINKED ISSUE #4161: LightBox: Overlay Panel is not displayed
## 1) Environment
- Does it work on the newest released PrimeFaces version? Yes Version? 6.2
- Affected browsers: Any

## 2) Expected behavior
The panel should be displayed.

## 3) Actual behavior
The panel is not displayed.

## 4) Steps to reproduce
1. This case can be tested in the showcase with PF 6.3-SNAPSHOT.

This issue was introduced in this [commit](https://github.com/primefaces/primefaces/commit/c5bd47f4c8edaa0b987272bd704840b41f170d11#diff-2782592d3b0e81a72383e3d23536c97f)

wsaca:
This is a fix to #4161. The panel should be appended to the body.


tandraschko:
Thanks but fixed it slightly different - see the issue

Issue #3391, opened 2018-03-01
DataList and DataGrid should support multiViewState


christophs78:
## 1) Environment
- PrimeFaces version: 6.2
- Does it work on the newest released PrimeFaces version? no
- Does it work on the newest sources in GitHub? no
- Application server + version: does not matter
- Affected browsers: does not matter

## 2) Expected behavior

DataTable supports multiViewState since Version 6.1.
DataList and DataGrid should support multiViewState too

## 3) Actual behavior

DataList and DataGrid do not support multiViewState

## 4) Steps to reproduce

Attribute/Feature multiViewState does not exist for DataList and DataGrid

## 5) Sample XHTML

Attribute/Feature multiViewState does not exist for DataList and DataGrid

## 6) Sample bean

Attribute/Feature multiViewState does not exist for DataList and DataGrid

Saturos88:
Vote for this feature too!

Rapster:
@christophs78 I'd like to have your thoughts on #4329 The idea is to make the impl more concise, instead of using different keys in the session map, the same one will be used either it's a datatable, datagrid, datalist etc. 

The problem with this, is that I don't know which component hides behind the key, so methods like `PrimeFaces#clearDataListStates` can't really be implemented instead you'd use `PrimeFaces#clearMultiViewStates` which clears up everything.

Do you have a use case where you want to clear states of datalist but not datatable for example?

PR #4165, opened 2018-10-24
#3391 multiViewState for dataList V2

christophs78:
I screwed my existing PR during rebase. Now the clean PR again.

### DIFF
--- a/src/main/java/org/primefaces/component/datalist/DataList.java
+++ b/src/main/java/org/primefaces/component/datalist/DataList.java
@@ -16,6 +16,7 @@
 package org.primefaces.component.datalist;
 
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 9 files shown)


src/main/java/org/primefaces/component/datalist/DataList.java
```diff
@@ -195,4 +196,42 @@ protected void processChildren(FacesContext context, PhaseId phaseId) {
         }
     }
 
+    @Override
+    public boolean isMultiViewState() {
```
Rapster: This method has no use. Please delete it

src/main/java/org/primefaces/component/datalist/DataList.java
```diff
@@ -195,4 +196,42 @@ protected void processChildren(FacesContext context, PhaseId phaseId) {
...
+        return super.isMultiViewState();
+    }
+
+    public void restoreListState() {
+        ListState ls = getListState(false);
+        if (ls != null) {
+            if (isPaginator()) {
```
Rapster: Make one condition using "&&"

src/main/java/org/primefaces/component/datalist/DataList.java
```diff
@@ -195,4 +196,42 @@ protected void processChildren(FacesContext context, PhaseId phaseId) {
...
+        }
+    }
+
+    public ListState getListState(boolean create) {
+        FacesContext fc = getFacesContext();
+        Map<String, Object> sessionMap = fc.getExternalContext().getSessionMap();
+        Map<String, ListState> dlState = (Map) sessionMap.get(Constants.DATALIST_STATE);
+        String viewId = fc.getViewRoot().getViewId().replaceFirst("^/*", "");
+        String stateKey = viewId + "_" + getClientId(fc);
+        ListState ls;
```
Rapster: Write `ListState ls = dlState.get(stateKey);` instead, before the condition (e.g remove that cumbersome declaration `ListState ls;`)

Rapster (commented):
@christophs78 I started a review, please see the comments

christophs78:
@Rapster Thank you for reviewing! Good point´s! Also made one modification to the existing DataTable-class which had the same issue.

tandraschko:
i would call the ListState class DataListState

christophs78:
Ok, I renamed ListState to DataListState.
Originally named it ListState because DataTable stores it´s multiViewState in org.primefaces.component.datatable.TableState. Tried to do replicate this naming for multiViewState of DataList.

tandraschko:
Yep, i know your intensions but even TableState is to generic as we also have a TreeTable.

tandraschko:
Thanks for your (great) first PR ;)

Issue #4004, opened 2018-08-29
SubMenu: does not have attribute disabled


brunoaandrade:
## 1) Environment
- PrimeFaces version: 6.2
- Affected browsers: all

## 2) Expected behavior
it should be possible to set the SubMenu disabled with an attribute. 
If i am using a p:submenu inside a p:menubar i should be able to disable a specific one.

## 3) Actual behavior
SubMenu does not have an attribute disabled

## 4) Steps to reproduce
Example in https://www.primefaces.org/showcase/ui/menu/menubar.xhtml
.. 

## 5) Sample XHTML
<h:form>  
    <p:growl id="messages"/>
    <p:menubar>
        <p:submenu label="File" icon="ui-icon-document">
            <p:submenu label="New" icon="ui-icon-contact" disabled="true">
                <p:menuitem value="Project" url="#" />
                <p:menuitem value="Other" url="#" />
            </p:submenu>
            <p:menuitem value="Open" url="#" />
            <p:separator />
            <p:menuitem value="Quit" url="#" />
        </p:submenu>
 
        <p:submenu label="Edit" icon="ui-icon-pencil">
            <p:menuitem value="Undo" url="#" icon="ui-icon-arrowreturnthick-1-w" />
            <p:menuitem value="Redo" url="#" icon="ui-icon-arrowreturnthick-1-e" />
        </p:submenu>
   </p:menubar>
</h:form>

## 6) Sample bean

..

tandraschko:
Please provide more description to your tickets, otherwise i will close it as invalid.
Should it be possible via xhtml, model or what do you want to archive...

PR #4169, opened 2018-10-28
#4004 Disabled attribute for SubMenu

EnmanuelReyes:
Adds disabled attribute for SubMenu as requested in #4004


tandraschko:
I like the change, thanks!

tandraschko:
Had to revert it, sorry.
You write the href attribute after the closing the "a" tag, thats not correct?!

Please provide new PR.

PR #4172, opened 2018-10-29
#4004 Disabled attribute for SubMenu

EnmanuelReyes:
This is the new PR after #4169  was reverted.

Adds disabled attribute for SubMenu as requested in #4004.


tandraschko:
It still produces invalid html / MyFaces throws a error.
I will fix it now for you but next time please test your code better.

PR #4192, opened 2018-10-31
Fix #4191 tabIndex shouldn't be rendered when component is disabled.

### LINKED ISSUE #4191: TriStateCheckbox: tabIndex shouldn't be rendered when component is disabled
## 1) Environment
- Affected browsers: Any

## 2) Expected behavior
TabIndex shouldn't be rendered when component is disabled

## 3) Actual behavior
TabIndex is rendered when component is disabled

## 4) Steps to reproduce
Simply include this component disabled, the tabindex is rendered, then this component take the focus even disabled.

wsaca:
Fix issue #4191.

### DIFF
--- a/src/main/java/org/primefaces/component/tristatecheckbox/TriStateCheckboxRenderer.java
+++ b/src/main/java/org/primefaces/component/tristatecheckbox/TriStateCheckboxRenderer.java
@@ -158,7 +158,9 @@ public class TriStateCheckboxRenderer extends InputRenderer {
         }
 
         writer.startElement("div", null);
-        writer.writeAttribute("tabIndex", checkbox.getTabindex() == null ? 0 : checkbox.getTabindex(), "tabindex");
+        if (!disabled) {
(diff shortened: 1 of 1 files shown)


tandraschko:
currently we always render the tabindex, so not sure if the change is correct

wsaca:
The component is receiving the focus even when it is disabled. Another option is to set tabIndex -1 or work with the input element to control the focus and states instead of the div.

tandraschko:
why does it work with other components?

wsaca:
If you want to verify this issue in other component you can do it with SelectOneButton, they both use tabindex on a div to control the focus and they have a similar issue.
The component SelectBooleanCheckbox works fine because the focus is controlled with the input element and when the disabled attribute is added then tabIndex is ignored, that's why we don't need to remove/change it if the component is disabled.

tandraschko:
Wouldn't it be a better fix to move the tabindex to a hidden input like done in other components?

wsaca:
I think so, but that change affects the way how to control the three states, should I give a proposal creating another PR or do it in this one?

wsaca:
I will create another issue.

PR #4195, opened 2018-11-01
Fix #2811: Calendar change events fired when date did not change.

### LINKED ISSUE #2811: Spurious valueChange events in a Calendar with a datetime pattern

melloware:
Added the fix the right way based on @mertsincan comments.


mertsincan:
Thanks a lot, @melloware ;)

PR #4213, opened 2018-11-06
Fix #4212: Fix Touch detection on Chrome 70 and latest MS Edge.

### LINKED ISSUE #4212: Signature:Not working on touchscreen since latest Windows update.
Just want to let you know that the signature is not working using the screen since latest windows update and Chrome update. 
In Chrome and Microsoft Edge.

This problem also occurred with jSignature and it was corrected changing some javascript code in the js file.
Here is the link to the thread that corrected the problem.

https://github.com/brinley/jSignature/issues/157

Hope this helps....
Regards

melloware:
Implemented the fix documented here to update detection of Touch Events on desktops based on removal of events.

https://github.com/furf/jquery-ui-touch-punch/issues/309

https://blog.chromium.org/2018/09/chrome-70-beta-shape-detection-web.html

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.touch-punch.js
+++ b/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.touch-punch.js
@@ -12,7 +12,7 @@
 (function ($) {
 
     // Detect touch support
-    $.support.touch = 'ontouchend' in document || navigator.maxTouchPoints > 0; //PF GitHub #4212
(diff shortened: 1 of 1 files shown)


melloware:
OK this version is a huge improvement over the one that has not been touched in 5 years.  it implemented much of the patches over the years to this component including:
- Better handler of Touch vs Pointer events
- Uses .on() instead of .bind().
- Better detection of browsers that have Touch events or Pointer Events.

PR #4215, opened 2018-11-06
Fix #4214 TriStateCheckbox: Move tabIndex attribute to the input element

### LINKED ISSUE #4214: TriStateCheckbox: move tabIndex attribute to the input element
## 1) Environment
- Affected browsers: Any

## 2) Expected behavior
TabIndex shouldn't work when the component is disabled. It should be rendered on the input element.

## 3) Actual behavior
TabIndex is rendered on the div element and working when the component is disabled.

## 4) Steps to reproduce
Simply include this component disabled, the tabindex is rendered, then this component take the focus when it's disabled.

## 5) Changes
To resolve this problem related to the focus, the next changes were done:
- Move tabIndex attribute to the input element.
- Use the class "ui-helper-hidden-accessible" to make visible and focusable the input element.
- Make the input element readonly because we don't need values added by the user.

wsaca:
This PR resolves #4214. The event keydown was necessary to disable the auto scrolling when the spacebar is pressed in Chrome.


tandraschko:
looks good, thanks

PR #4221, opened 2018-11-07
Fix #3668 Tree dynamic="true": now expand event gets triggered on each expand

### LINKED ISSUE #3668: Tree: expand/collapse events not symmetrical with 'dynamic="true"'
## 1) Environment
- PrimeFaces version: 6.1.x ,6.2.x
- Does it work on the newest released PrimeFaces version? No, Version? Tried 6.2 Elite
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) 6.2-SNAPSHOT no
- Application server + version: N/A (tried Tomcat 8 and Wildfly 10
- Affected browsers: ALL

## 2) Actual behavior
When a node is opened for the first time, an 'expand' ajax event is fired and the child nodes are populated client side. When the node is closed, a 'collapse' event is send. So far so good. But when the node is opened again, there is no ajax event send. Since the content is present no in the client, it can in a way be understood that there is no ajax event sent. What is strange than is that on closing this node, an collapse event **is** sent. 

When dynamic is false, each open and close results in a new ajax event, even if all content is already present on the client. 

The same behaviour can be achieved with `dynamic="false" when the `cache` attribute, which defaults to true, is set to false. 

## 3) Expected behavior
At least symmetrical behaviour for the expand and collapse events. Two options:
- Only send an ajax event on the initial open and close?
- Like with dynamic="false" keep sending ajax events all the time. 

The latter would be what is most in line with expected behaviour I think. 

## 4) Steps to reproduce

Add `dynamic="true"` to /ui/data/tree/events.xhtml in the showcase

## 5) Sample XHTML

See steps to reproduce (much easier than posting xhtml here

## 6) Sample bean

Use the beans from the /ui/data/tree/events.xhtml case in the showcase. 

## 7) Fix 

If fixing it like stated in 3 would be the way to go, adding  `this.fireExpandEvent(node); ` in line [63 in the tree.js in trunk](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/tree/tree.js#L63)  is what should be done.

## 8) Workaround: 

Add  the following javascript in your page:

```
<h:outputScript>
        
//<![CDATA[

	PrimeFaces.widget.BaseTree.prototype.expandNode = function(node) {
        var $this = this;

        if(this.cfg.dynamic) {
            if(this.cfg.cache && $this.getNodeChildrenContainer(node).children().length > 0) {
                this.showNodeChildren(node);
  	        this.fireExpandEvent(node);
                return;
            }

            if(node.data('processing')) {
                PrimeFaces.debug('Node is already being expanded, ignoring expand event.');
                return;
            }

            node.data('processing', true);

            var options = {
                source: this.id,
                process: this.id,
                update: this.id,
                formId: this.cfg.formId,
                params: [
                    {name: this.id + '_expandNode', value: $this.getRowKey(node)}
                ],
                onsuccess: function(responseXML, status, xhr) {
                            widget: $this,
                            handle: function(content) {
                                var nodeChildrenContainer = this.getNodeChildrenContainer(node);
                                nodeChildrenContainer.append(content);

                                this.showNodeChildren(node);

                                if(this.cfg.draggable) {
                                    this.makeDraggable(nodeChildrenContainer.find('span.ui-treenode-content'));
                                }

                                if(this.cfg.droppable) {
                                    this.makeDropPoints(nodeChildrenContainer.find('li.ui-tree-droppoint'));
                                    this.makeDropNodes(nodeChildrenContainer.find('span.ui-treenode-droppable'));
                                }
                            }
                        });

                    return true;
                },
                oncomplete: function() {
                    node.removeData('processing');
                }
            };

            if(this.hasBehavior('expand')) {
                var expandBehavior = this.cfg.behaviors['expand'];
                expandBehavior.call(this, options);
            }
            else {
                PrimeFaces.ajax.Request.handle(options);
            }
        }
        else {
            this.showNodeChildren(node);
            this.fireExpandEvent(node);
        }
    }
	
	

//]]>
        </h:outputScript>```

VsevolodGolovanov:


tandraschko:
I will close the PR for now.
I don't see any reason to fire an ajax event when cache=true, it just slow down the gained performance that someone would expect from using cache=true.

Issue #2791, opened 2017-10-09
FileUpload: check allowTypes on server side


rcroonenberghs:
## 1) Environment
- PrimeFaces version: 6
- Application server + version: wildfly-10.1.0.Final
- Affected browsers: All

## 2) Expected behavior

When uploading a file which is not in the allowedtypes, a error is shown on screen and upload is blocked. But if you tamper with the post, you can change a valid file to a file extension that would not be allowed. (tested via Burpsuite) Expected behaviour would be that on server side, the allowedtypes would be checked as well.

## 3) Actual behavior

The allowedtypes is not checked on server side.

## 4) Steps to reproduce

submit upload and tamper the post and change the file extension.

## 5) Sample XHTML
```xml
<p:fileUpload id="fileupload" fileUploadListener="#{documentlist.handleFileUpload}"
		            mode="advanced"
		          allowTypes="/(\.|\/)(gif|png|jpe?g)$/i"					            
		            />
```
## 6) 
public void handleFileUpload(FileUploadEvent event) {

kukel:
checking  allowed *types* based on filename extensions (not on content) is confusing. It should in the documentation (and showcase) be explicitly stated that content is not checked! (allowed 'extensions' would have been a better 'name')

melloware:
@tandraschko @cnsgithub Can you add the security label to this.  Our Pentration testers also reported this issue.

cnsgithub:
@melloware Just for info, I am not allowed to add any labels. :wink: Btw, are you allowed to share what your penetration testers found besides this issue? I'd be glad having those vulnerabilities fixed.

Just checking the extensions will always be unsafe. However, I agree with @kukel that `allowTypes `should be renamed to clarify that we are not doing any content type checks. We could even use a more abstract attribute name like `allowFilenameRegex` (or something shorter), as the check is not limited to just the extensions. I also agree that this regex should be validated against at the backend side once again.

The more tricky part would be to introduce content type checking. We could use something like `java.nio.file.Files.probeContentType(Path) (`Java 7), however that would most probably not work without having installed additional `FileTypeDetectors`, e.g. Apache Tika.

Some further reading: https://www.owasp.org/index.php/Unrestricted_File_Upload

melloware:
[...]

PR #4242, opened 2018-11-15
fix: check filename of uploaded file at server side as well (#2791)

cnsgithub:
add: content type checking of uploaded file based on java 7 FileTypeDetector (#2791)


cnsgithub:
@tandraschko @melloware This is a first draft only and needs thorough testing. 

Unfortunately, I have not yet managed to get `org.apache.tika.filetypedetector.TikaFileTypeDetector `picked up by the `java.nio.file.spi.FileTypeDetector` service loader despite adding the dependency to the showcase's `pom.xml`:
```xml
        <dependency>
            <groupId>org.apache.tika</groupId>
            <artifactId>tika-java7</artifactId>
            <version>1.19.1</version>
            <scope>runtime</scope>
        </dependency>  
```
Might be a classloading problem? Webapp vs. system classloading?

cnsgithub:
Btw, my last commit to this branch upgrades junit and mockito to work with java 11, too. Should be okay? I need this for providing a unit test...

tandraschko:
it should actually work but if you are on Java11, maybe a modules problem?

the dependency upgrade is ok, yes.

1) Could you please add 2-3 simple unit tests for the isValidType method?
2) Should we make this feature deactivatable? If tika is not present, the content check is not reliable and should be deactivatbale. Otherwise is tika is >actually< required dependency.

cnsgithub:
> it should actually work but if you are on Java11, maybe a modules problem?

After upgrading the versions everything works fine with java 11, thus my question. It was due to a method in `sun.misc.Unsafe` being removed. This method however was frequently used by code generators, see https://dzone.com/articles/jdk-11-and-proxies-in-a-world-past-sunmiscunsafe for an explanation.

> Could you please add 2-3 simple unit tests for the isValidType method?

Yes, it was my intention to do that.

> Should we make this feature deactivatable? If tika is not present, the content check is not reliable and should be deactivatbale. Otherwise is tika is >actually< required dependency.

Usually there is always one default implementation for `java.nio.file.spi.FileTypeDetector`. On my Windows machine with JRE 11 it's `sun.nio.fs.RegistryFileTypeDetector `that simply picks up the file extension from the Windows registry for example. That means if you upload a file `test.png `it'll lookup `image/png`. However, if you upload an executable and rename it to test.png it'll still lookup `image/png ` wrongly. In very rare cases `Files.probeContentType `may return `null`. Then we have to decide what to do. My current implementation will strictly refuse the upload.

tandraschko:
Yup, i know ;) But that should be unrelated to the serviceloader.


I see. I would make it deactivatable.

cnsgithub:
> Could you please add 2-3 simple unit tests for the isValidType method?

Added some more unit tests. Some advanced tests making use of Apache Tika are still required...

tandraschko:
JFYI:

`ng.Boolean</type>re`

the "re" is to much :D

EDIT: you already changed it

tandraschko:
Thanks for the good PR, it's a good improvement.

cnsgithub:
Continued this PR in https://github.com/primefaces/primefaces/pull/4249

PR #4243, opened 2018-11-15
Fix #4240: ChartJS model objects are all now serializable.

### LINKED ISSUE #4240: Charts: Model classes should implement Serializable
## 1) Environment
- PrimeFaces version: 6.3.0
- Does it work on the newest released PrimeFaces version? Version? No, 6.3.0
- Application server + version: does not matter
- Affected browsers: does not matter

In the new chart package (org.primefaces.model.charts)  Any class that can be converted to a JSON object should implement Serialiable.

melloware:
All model objects are now Serializable and imports have been properly organized.

### DIFF
--- a/src/main/java/org/primefaces/model/charts/ChartData.java
+++ b/src/main/java/org/primefaces/model/charts/ChartData.java
@@ -23,8 +23,8 @@ import java.util.List;
  * The data configuration for Charts
  */
 public class ChartData implements Serializable {
-	
-	private static final long serialVersionUID = 1L;
+
+    private static final long serialVersionUID = 1L;
 
     private List<ChartDataSet> dataSet;
     private Object labels;
--- a/src/main/java/org/primefaces/model/charts/ChartData.java
+++ b/src/main/java/org/primefaces/model/charts/ChartData.java
@@ -23,6 +23,8 @@ import java.util.List;
  * The data configuration for Charts
  */
 public class ChartData implements Serializable {
+	
+	private static final long serialVersionUID = 1L;
 
     private List<ChartDataSet> dataSet;
     private Object labels;
--- a/src/main/java/org/primefaces/model/charts/ChartDataSet.java
+++ b/src/main/java/org/primefaces/model/charts/ChartDataSet.java
@@ -16,8 +16,11 @@
 package org.primefaces.model.charts;
 
 import java.io.IOException;
+import java.io.Serializable;
 
-public abstract class ChartDataSet {
+public abstract class ChartDataSet implements Serializable {
+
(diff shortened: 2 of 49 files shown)


tandraschko:
Please remove the serialVersionUID with default value, thats just boilerplate.
Also travis seems to be broken?

melloware:
Yep fixing Checktyle.  Are you sure the serialVersionId is boilerplate Eclipse claims its not valid without it?

tandraschko:
It is valid without
serialVersionId just makes sense if used correclty with versioning (instead of default value) for persistent objects (e.g. if you save objects per file)

melloware:
For serialVersionId: https://stackoverflow.com/questions/285793/what-is-a-serialversionuid-and-why-should-i-use-it

> If a serializable class does not explicitly declare a serialVersionUID, then the serialization runtime will calculate a default serialVersionUID value for that class based on various aspects of the class, as described in the Java(TM) Object Serialization Specification. However, it is strongly recommended that all serializable classes explicitly declare serialVersionUID values, since the default serialVersionUID computation is highly sensitive to class details that may vary depending on compiler implementations, and can thus result in unexpected InvalidClassExceptions during deserialization. Therefore, to guarantee a consistent serialVersionUID value across different java compiler implementations, a serializable class must declare an explicit serialVersionUID value. It is also strongly advised that explicit serialVersionUID declarations use the private modifier where possible, since such declarations apply only to the immediately declaring class serialVersionUID fields are not useful as inherited members.

So it looks like a computation performance hit of recalculating serialVersionId every time?

melloware:
Also from Joshua Bloch: 

> It is worth noting that Joshua Bloch advices that for every Serializable class it's worth specifying the serial version uid. Quote from chapter 11: Regardless of what serialized form you choose, declare an explicit serial version UID in every serializable class you write. This eliminates the serial version UID as a potential source of incompatibility (Item 74). There is also a small performance benefit. If no serial version UID is provided, an expensive computation is required to generate one at runtime. 

So to not be doing this makes our "for loop" optimization moot.  This is an easy microoptimization.

tandraschko:
ok

> There is also a small performance benefit. If no serial version UID is provided, an expensive computation is required to generate one at runtime.

thats the only reason then ;)

pzygielo:
Out of curiosity - what bad happens if these classes do not implement `Serializable`? I see no error described in 4240. What's the reasoning for this change?

tandraschko:
A exception is thrown if you use with session/viewscoped, if they are serialized.

pzygielo:
@tandraschko - thanks!

PR #4250, opened 2018-11-16
fix p:ajax not firing server side listener

paolopiccinini:
p:ajax inside p:textEditor is broken. It want not fire listener method.


tandraschko:
Please create a valid issue first

melloware:
Also you have a CheckStyle issue that needs to be resolved.

> [ERROR] /home/travis/build/primefaces/primefaces/src/main/java/org/primefaces/component/texteditor/TextEditorRenderer.java:36: 'method def' child has incorrect indentation level 16, expected level should be 8. [Indentation]

tandraschko:
we should probably also add 
`
        if (!shouldDecode(inputText)) {
            return;
        }
`
to the decode method

tandraschko:
Thanks for the PR

PR #4259, opened 2018-11-19
add: fileUpload: virus scan (closes #4256)

### LINKED ISSUE #4256: fileUpload: virus scan
Just an idea for further enhancement of `p:fileUpload` in terms of security: automatic virus scanning by providing an opt-in `performVirusScan` attribute.

Therefore I would recommend to use the [VirusTotal file-report API](https://developers.virustotal.com/v2.0/reference#file-report) that may be used after free registration and obtaining an API key (→ `primefaces.PRIVATE_VIRUSTOTAL_KEY` context param) in non-commercial applications.

The `file-report `endpoint is expecting a file hash only instead of the whole file content which would be subject to disclosure and asynchronously.

If there is some positive feedback, I could provide a pull request.

cnsgithub:
- Virus scan can be enabled via `p:fileUpload performVirusScan="true"` (default is `false`)
- Virus scanners can be plugged in and will be picked up automatically by implementing the SPI `org.primefaces.virusscan.spi.VirusScanner`.
- PrimeFaces bundles a default SPI provider that requests the file hash from VirusTotal. To make this work the user has to configure the context parameter `primefaces.virusscan.VIRUSTOTAL_KEY`.
- When plugging in custom implementations VirusTotal can either kept enabled or manually disabled by setting the context parameter `primefaces.virusscan.VIRUSTOTAL_ENABLED `to `false`.
- If there are more than one virus scanners ALL of them must be passed to proceed.

@melloware What do you think how easy it would be to implement a ClamAV provider? Maybe we could provide an example on the wiki? (see https://github.com/primefaces/primefaces/wiki/More-secure-file-upload)

Please note that the current implementation is actually not sending and scanning any file contents. That might probably imply long running tasks and make it necessary to poll VirusTotal until scanning is finished.

### DIFF
--- a/src/main/java/org/primefaces/component/fileupload/CommonsFileUploadDecoder.java
+++ b/src/main/java/org/primefaces/component/fileupload/CommonsFileUploadDecoder.java
@@ -24,6 +24,7 @@ import org.primefaces.event.FileUploadEvent;
 import org.primefaces.model.DefaultUploadedFile;
 import org.primefaces.model.UploadedFileWrapper;
 import org.primefaces.util.FileUploadUtils;
+import org.primefaces.virusscan.VirusException;
 import org.primefaces.webapp.MultipartRequest;
 
 import java.io.IOException;
@@ -89,8 +90,17 @@ public class CommonsFileUploadDecoder {
     }
 
     private static boolean isValidFile(FileUpload fileUpload, DefaultUploadedFile uploadedFile) throws IOException {
-        return (fileUpload.getSizeLimit() == null || uploadedFile.getSize() <= fileUpload.getSizeLimit()) && FileUploadUtils.isValidType(fileUpload,
+        boolean valid = (fileUpload.getSizeLimit() == null || uploadedFile.getSize() <= fileUpload.getSizeLimit()) && FileUploadUtils.isValidType(fileUpload,
                 uploadedFile.getFileName(), uploadedFile.getInputstream());
+        if (valid) {
+            try {
+                FileUploadUtils.performVirusScan(fileUpload, uploadedFile.getInputstream());
+            }
+            catch (VirusException ex) {
+                return false;
+            }
+        }
+        return valid;
     }
 
 }
--- a/src/main/java/org/primefaces/component/fileupload/FileUploadBase.java
+++ b/src/main/java/org/primefaces/component/fileupload/FileUploadBase.java
@@ -62,7 +62,8 @@ abstract class FileUploadBase extends UIInput implements Widget {
         uploadIcon,
         cancelIcon,
         onAdd,
-        validateContentType
+        validateContentType,
+        performVirusScan
     }
 
     public FileUploadBase() {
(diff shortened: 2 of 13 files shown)


tandraschko:
1) I would put all related files in the virusscan root directory. we already have so many packages, i would not introduce packages for only 1 class
2) i would not register the VirusTotalScanner per default, so we also don't need the primefaces.virusscan.VIRUSTOTAL_ENABLED parameter
If the user would like to use a VirusScanner, he need to configure which one and maybe also the virustotal_key
3) we should talk about it, if we should really use a serviceloader here. Ofc the benefit is that a user could use multiple virusscanner, but i'm not sure if thats required. We decided in the past to use a context param for such cases (see CacheProvider), we should think about using the same way here.

cnsgithub:
> I would put all related files in the virusscan root directory. we already have so many packages, i would not introduce packages for only 1 class

The current package structure looks like this:
image

I'd keep the `spi `sub package to clearly denote that it is a SPI, like `java.nio.FileTypeDetector` for example.

> i would not register the VirusTotalScanner per default, so we also don't need the primefaces.virusscan.VIRUSTOTAL_ENABLED parameter
> If the user would like to use a VirusScanner, he need to configure which one and maybe also the virustotal_key

Hmm, this would mean that `p:fileUpload performVirusScan="true"` cannot be used out of the box without having to specifiy additional context params. That's a matter of taste. :wink: 

> we should talk about it, if we should really use a serviceloader here. Ofc the benefit is that a user could use multiple virusscanner, but i'm not sure if thats required. We decided in the past to use a context param for such cases (see CacheProvider), we should think about using the same way here.

Like the benefit of chaining several scanners. :wink: 

Are there any chances to get this in PF 6.3?

tandraschko:
1) it's just your view - i develop with netbeans default settings ;)
    also both VirusException and VirusScannerService is part of the SPI - i would do it this way:
    virusscan/VirusException
    virusscan/VirusScanner
    virusscan/VirusScannerService 
    virusscan/impl/....

2) yep, but better as 2 places for the "same" kind of configs
    maybe later we have multiple build-in implementations, we would need to add a context-param for each... thats a -1 for me

3) have to talk with PrimeTek, for me it's a -0,1 as it's different with other SPIs (CacheProvider); we could also align other places but we should do it similar everywhere

cnsgithub:
> 1. it's just your view - i develop with netbeans default settings ;)
>    also both VirusException and VirusScannerService is part of the SPI - i would do it this way:
>    virusscan/VirusException
>    virusscan/VirusScanner
>    virusscan/VirusScannerService
>    virusscan/impl/....

Refactored the packages.



> 2\. yep, but better as 2 places for the "same" kind of configs
>     maybe later we have multiple build-in implementations, we would need to add a context-param for each... thats a -1 for me

I see your point. However, I think that we most probably will never provide a second default implementation. Let us remove the `primefaces.virusscan.VIRUSTOTAL_ENABLED` context param and just check the existence of `primefaces.virusscan.VIRUSTOTAL_KEY`. Existence of this context param would imply that we want to use the `VirusTotalVirusScanner` default implementation. Would that be okay for you?



> 3\. have to talk with PrimeTek, for me it's a -0,1 as it's different with other SPIs (CacheProvider); we could also align other places but we should do it similar everywhere

Any news or responses from PrimeTek? I think `primefaces.CACHE_PROVIDER` is the only case we have so far where an concrete implementation is defined. But the use case is different. More than one cache provider will never make sense, but having more than one virus scanner could perfectly make sense, e.g. `VirusTotalVirusScanner` is just verifying checksums while other user-supplied implementations might perform deeper analyses.

tandraschko:
1) +1
2) yep, sounds good
3) not yet, sorry

cnsgithub:
> 2. yep, sounds good

Removed  ``primefaces.virusscan.VIRUSTOTAL_ENABLED`` context param in favor of implicitly enabling ``VirusTotalVirusScanner`` by specifying ``primefaces.virusscan.VIRUSTOTAL_KEY``.

tandraschko:
yep, looks good now
i will do a deep review and probably merge it until the RC1

PR #4262, opened 2018-11-20
add: experimental, basic support for Content-Security-Policy scripts (#3505) - supersedes #3884

cnsgithub:
Copied from #3884 

> Hey guys,
> 
> As a proof of concept I implemented basic support for `Content-Security-Policy `scripts. Do not worry, it's just experimental and of course disabled by default. I tested some rudimentary things in the showcase and that worked for me. Please note, that support for `partial-response` is not yet implemented.
> 
> I would like you to give it a try by just adding the following to the showcase's `web.xml` - feedback is very welcome:
> 
> ```
>     <context-param>
>         <param-name>primefaces.csp.ENABLED</param-name>
>         <param-value>true</param-value>
>     </context-param>
>     <context-param>
>         <param-name>primefaces.csp.SUPPORTED_DIRECTIVES</param-name>
>         <param-value>script-src</param-value>
>     </context-param>
> ```
> 
> When looking at the browser console you will notice something like that:
> 
> `Refused to load the script 'https://www.google-analytics.com/analytics.js' because it violates the following Content Security Policy directive: "script-src 'nonce-OWNiNjI0ZWEtOWM2My00MDI4LTk2NzYtYWM2ZTQ5MTk0OGEz' 'nonce-Y2Q2NTAxZGQtODYzNi00MGJiLTg5ZGItYmQ2ZDljNzFlZDJl' 'nonce-NTk5NzEzMjgtN2E3OS00MDlmLTg5ZTgtZTI3YTY3MjMyOGE0' 'nonce-ZjM5NzQ0MTEtODk1Mi00OWE2LTlmZTgtNjk0OWY4YWRlOGNh' 'nonce-YzFhNDZlNGMtYzI0NS00OGVlLWJmNTctYTZkMmJiZWZlMTIw' 'nonce-MGMzMDg0MzgtZWI1Yy00ZjJjLTg1OTgtODY1OTgwMGU0ZDQ0' 'nonce-MjEyMWViMzktZDY1Yy00ZmQ0LTg1NzMtMDkyNTFjMTg4NzM0' 'nonce-ZDRmMjA5ODEtNWU3Ni00YjgzLWEwNTctMmEwYmViZGRkYWY2' 'nonce-ZDI4NGY1NjYtY2RkOC00MDUyLWI4YTUtZTFhZDcxMzJkOGMw'".`
> 
> When looking at the jetty console you will notice that the CSP violation has been reported:
> 
> `	WARNING: CSP violation reported by browser: {"csp-report":{"referrer":"","status-code":200,"disposition":"enforce","script-sample":"","violated-directive":"script-src","original-policy":"script-src 'nonce-MWYyNjlhMzgtNjgyYy00MTlhLTg2NGQtOWVlNjJkM2NjNmM3' 'nonce-NmJhZGI2YzktZGYwZC00ZjZhLTlkMjctZTk1YTZjMTI4YWUy' 'nonce-MzMzOTY3ZmMtYzJjYS00ZGFiLThjY2UtZjUxNTk0ZWQ5NmYz' 'nonce-Y2I0YTE4NDktZTk5OS00ZWUzLTkxYTQtZWIzMTU2NTI5NTNj' 'nonce-OTZhMmFlZDEtZWU0NC00NmQyLThhZDgtYWU3NjUzNzQxNDcw' 'nonce-MTZmYmUzNDUtNmY3NC00OWE2LWEzMmQtZWI5MTA1ZjRhNDlh' 'nonce-ZmM5NWNmMjctNWI1ZC00MWI2LWEyMjYtNTExODg3Y2ZlZTZi' 'nonce-YjM1NTM4NzEtZWVhZC00NzdlLTk2YmMtZjcwNDc1YjI5NWMy' 'nonce-ZWMxNTA0MmYtMjdlNy00MTNiLTlmMWMtM2M2MTI0Mzc3ODg5'; report-uri /showcase/csp-report;","source-file":"http://localhost:8080/showcase/","column-number":34,"document-uri":"http://localhost:8080/showcase/","line-number":511,"effective-directive":"script-src","blocked-uri":"https://www.google-analytics.com/analytics.js"}}`
> 
> Then just go back to your `web.xml` and add these lines:
> 
> ```
>     <context-param>
>         <param-name>primefaces.csp.ALLOWED_HOSTS</param-name>
>         <param-value>https://www.google-analytics.com</param-value>
>     </context-param>
> ```
> 
> Your browser will be happy again.
> 
> When inspecting the DOM you will notice the nonces that have been added to all `script` elements. Furthermore you will not find any `onclick` etc. attributes anymore. Those are all merged into the last `script` block:
> 
> ```
> <script nonce="">var pf=PrimeFaces;
> pf.csp1=function(id,evt,js){document.getElementById(id).addEventListener(evt,js);}
> pf.csp2=function(id,attr,jsUri){document.getElementById(id).setAttribute(attr,jsUri);}
> pf.csp1("SubMenu-Ajax","click",function(){Showcase.openSubMenu(this);});
> pf.csp1("SubMenu-Input","click",function(){Showcase.openSubMenu(this);});
> pf.csp1("SubMenu-Button","click",function(){Showcase.openSubMenu(this);});
> pf.csp1("SubMenu-Data","click",function(){Showcase.openSubMenu(this);});
> pf.csp1("SubMenu-Panel","click",function(){Showcase.openSubMenu(this);});
> pf.csp1("SubMenu-Overlay","click",function(){Showcase.openSubMenu(this);});
> pf.csp1("SubMenu-Menu","click",function(){Showcase.openSubMenu(this);});
> pf.csp1("SubMenu-Charts","click",function(){Showcase.openSubMenu(this);});
> pf.csp1("SubMenu-Message","click",function(){Showcase.openSubMenu(this);});
> pf.csp1("SubMenu-Multimedia","click",function(){Showcase.openSubMenu(this);});
> pf.csp1("SubMenu-File","click",function(){Showcase.openSubMenu(this);});
> pf.csp1("SubMenu-DragDrop","click",function(){Showcase.openSubMenu(this);});
> pf.csp1("SubMenu-CSV","click",function(){Showcase.openSubMenu(this);});
> pf.csp1("SubMenu-DF","click",function(){Showcase.openSubMenu(this);});
> pf.csp1("SubMenu-Misc","click",function(){Showcase.openSubMenu(this);});
> </script>
> ```
> 
> Please note that in contrast to Firefox, Chrome will not show the nonces. But they are in place, check this by inspecting the headers:
> 
> image
> 
> **Some TODOs**
> 
> * [x]  support `partial-response`
> * [ ]  support `partial-response eval`
> * [x]  support hashes as well
> * [x]  provide JUnit tests
> * [ ]  find better solution than increasing the buffer size
> * [ ]  test, test, test... wink
> * [x]  provide showcase
> * [ ]  get PrimeFaces team to integrate it... tongue
> 
> And as I said, I would be glad to get some feedback...


cnsgithub:
Copied from #3884, comment from @tandraschko 

> Short review:
> 
> * It would be great if we can avoid the the checkstyle exclude
> * could you patch the jquery method via overwrite in another file? Otherwise your change will probably be lost on the next update
> * i'm not sure about the HeaderWriterResponse and the ServletFilter
>   Whats the meaning of it? Can't we plugin in the JSF lifecycle? Feels a bit to hacky.
> * CspReportServlet - is this something optional, right? Does the browser report back to the server if CSP fails?

cnsgithub:
Copied from #3884, comment from @cnsgithub 

> > It would be great if we can avoid the the checkstyle exclude
> 
> I think we could do that. However, I just thought because `OnCommittedResponseWrapper `is a copy from `spring-security-web` it would be better to not change it to make it easier to compare with newer versions in the future. There are already other candidates as well.
> 
> > could you patch the jquery method via overwrite in another file? Otherwise your change will probably be lost on the next update
> 
> That makes sense, absolutely. Okay, can you give me a hint how to do that? Btw, I hope that next jquery version 3.4.0 will include that fix.
> 
> > i'm not sure about the HeaderWriterResponse and the ServletFilter
> > Whats the meaning of it? Can't we plugin in the JSF lifecycle? Feels a bit to hacky.
> 
> Do you mean a JSF `PhaseListener `(render response complete) or something like that? I tested that in the first place. Cannot remember right now why I discarded this idea...
> 
> > CspReportServlet - is this something optional, right? Does the browser report back to the server if CSP fails?
> 
> Yes, the `report-uri` directive and hence the servlet is optional, albeit very reasonable in my opinion. Please see @melloware's review comments and my answer also regarding `@WebServlet `annotation. I thought it would be nice to have the endpoint already built-in. Users may override the endpoint in `web.xml`.

cnsgithub:
Showcase: https://github.com/primefaces/showcase-facelift/pull/8

cnsgithub:
Currently I have no desire to work on this on my own. It's been too much time. It would be cool if someone could join.

melloware:
@cnsgithub Can you lay out the tasks remaining and I am sure we can jump in an help.

cnsgithub:
@melloware Of course. It's primarily the open tasks that can be found in https://github.com/primefaces/primefaces/pull/4262#issue-232214085. And then the things @tandraschko mentioned in his review: https://github.com/primefaces/primefaces/pull/4262#issuecomment-440170250

cnsgithub:
>  support partial-response eval

The basic structure of partial-response is as follows:
```xml
<partial-response>
   <changes>
      <update id="form:j_idt...">...</update>
      <update id="...">...</update>
      <insert>...</insert>
      <eval><![CDATA[ /* CALL SOME JAVASCRIPT here */]]</eval>
   </changes>
</partial-response>
```

The javascript that's included within `eval` cannot be executed because there is no nonce being allowed in the CSP header. The JSF specification lacks to define an attribute like `nonce` for the `eval` node. We could however try to move the javascript code from `eval `to `update`, embedding it within `script `tags and use a `nonce` attribute there (don't using `eval` at all). Therefore it might be necessary to override `PrimePartialResponseWriter`'s `startEval `and `endEval` methods in `CspScriptsPartialResponseWriter`. Eval scripts are triggered by calls to JSF's `RequestContext.getCurrentInstance().execute`.

src/main/java/org/primefaces/application/resource/csp/CspFilter.java
```diff
@@ -0,0 +1,186 @@
...
+
+        /**
+         * Content-Security-Policy header cannot be set before collection of nonce/hash information in {@link CspScriptsResponseWriter}.
+         * Therefore as a workaround we have to increase the response buffer size.
+         * Otherwise, sending headers after the first flush occurred will not have any effects.
+         * See {@link javax.servlet.ServletResponse#setBufferSize(int)}.
+         * @param bufferSize the buffer size to use
+         */
+        private void configureResponse(int bufferSize) {
+            setBufferSize(bufferSize);
```
cnsgithub, line 100: >  find better solution than increasing the buffer size

@melloware This is some kind of hack that the open task addresses.

tandraschko:
About the eval:
could we add a nonce attribute to the eval tag and execute the script then correctly?

src/main/java/org/primefaces/application/resource/csp/CspFilter.java
```diff
@@ -0,0 +1,186 @@
+/**
```
cnsgithub, line 1: @melloware @tandraschko prefers using JSF `PhaseListener `here.

src/main/java/org/primefaces/application/resource/csp/CspFilter.java
melloware: Yep I see what you are doing here.  This is a tricky one I can see why you solved it this way

cnsgithub:
> About the eval:
> could we add a nonce attribute to the eval tag and execute the script then correctly?

Please see my comment above:

> The JSF specification lacks to define an attribute like nonce for the eval node.

I think the execution of the code contained within `eval `is somehow executed by the JSF implementation, maybe `jsf.js` in mojarra?

tandraschko:
It depends.
If p:ajax/p:commandButton etc. is used, then we handle the eval node (see https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js#L819) and we can easily add a attribute
we can just not support it when f:ajax is used... Maybe we can hook in to mojarra and myfaces impl, don't know
but for me it's ok if we only support p:ajax for now.

src/main/resources/META-INF/resources/primefaces/jquery/jquery.js
```diff
@@ -1,2 +1,2 @@
...
...e(n,n),n}},w.each(w.expr.match.bool.source.match(/\w+/g),function(e,t){var n=ht[t]||w.find.attr;ht[t]=function(e,t,r){var i,o,a=t.toLowerCase();return r||(o=ht[a],ht[a]=i,i=null!=n(e,t,r)?a:null,ht[a]=o),i}});var gt=/^(?:input|select|textarea|button)$/i,yt=/^(?:a|area)$/i;w.fn.extend({prop:function(e,t){return z(this,w.prop,e,t,arguments.length>1)},removeProp:function(e){return this.each(function(){delete this[w.propFix[e]||e]})}}),w.extend({prop:function(e,t,n){var r,i,o=e.nodeType;if(3!==o&&8!==o&&2!==o)return 1===o&&w.isXMLDoc(e)||(t=w.propFix[t]||t,i=w.propHooks[t]),void 0!==n?i&&"set"in i&&void 0!==(r=i.set(e,n,t))?r:e[t]=n:i&&"get"in i&&null!==(r=i.get(e,t))?r:e[t]},propHooks:{tabIndex:{get:function(e){var t=w.find.attr(e,"tabindex");return t?parseInt(t,10):gt.test(e.nodeName)||yt.test(e.nodeName)&&e.href?0:-1}}},propFix:{"for":"htmlFor","class":"className"}}),h.optSelected||(w.propHooks.selected={get:function(e){var t=e.parentNode;return t&&t.parentNode&&t.parentNode.selectedIndex,null},set:function(e){var t=e.parentNode;t&&(t.selectedIndex,t.parentNode&&t.parentNode.selectedIndex)}}),w.each(["tabIndex","readOnly","maxLength","cellSpacing","cellPadding","rowSpan","colSpan","
```
cnsgithub, line 2: comment from @tandraschko 

> could you patch the jquery method via overwrite in another file? Otherwise your change will probably be lost on the next update

my answer:

> That makes sense, absolutely. Okay, can you give me a hint how to do that? Btw, I hope that next jquery version 3.4.0 will include that fix.

melloware: @cnsgithub can you show me the bug or exactly what you fixed and i can probably help with where to put it in the pf overrides JS file.

cnsgithub: @melloware See this jquery pull request: https://github.com/jquery/jquery/pull/3766/files, basically the `nonce` attribute must not be lost.

melloware: Thanks!  It looks like that PR is not getting updated by the original developer.  I wonder if we should  submit a separate Jquery PR?

tandraschko:
I will probably take over.
However, if we can't support all jsf/PF core features (like eval nodes in the ajax response), i'm not sure if i will work further on it or apply it.

tandraschko:
@cnsgithub @melloware i just wonder - coudln't we use 1 nonce per request instead of multiple ones? Or is a new nonce required for every script tag? I just wonder as we have a set of nonces.

cnsgithub:
Actually I switched from multiple different nonces to just one nonce per session, see https://github.com/primefaces/primefaces/pull/3884#issuecomment-408334160 for the reason. The set is most probably a relict.

cnsgithub:
> However, if we can't support all jsf/PF core features (like eval nodes in the ajax response), i'm not sure if i will work further on it or apply it.

Let's try to move the `eval` script to `update` (see https://github.com/primefaces/primefaces/pull/4262#issuecomment-440245001) and additionally request a change at https://github.com/javaee/javaserverfaces-spec for future releases? Otherwise the spec would further prevent JSF users from securing their web apps.

tandraschko:
does CSP require the hashing of the inline script? Otherwise a implementation would be quite easy....

cnsgithub:
Which implementation do you think of? PrimeFaces or JSF in general? Hashing is just an alternative to nonces.

tandraschko:
In generell
There are just a few steps with a few classes:
1) generate a nonce per view
2) add this in the headers via a phaselistener (without all those OnCommittedResponseWrapper stuff, buffering...), just a one liner
3) catch all rendered scripts block and add the nonce
4) move the rendered on* to a script block
5) try to find a solution for the ajax eval

right?
After that we can still add the reporting

cnsgithub:
In principle yes. 

I think the tricky part is to

> catch all rendered scripts block and add the nonce

Where exactly do you want to add the nonce to the scripts?

You cannot modify data that has already been written and flushed to the output stream. That's why I increased the buffer size (that's bad of course).

> generate a nonce per request

The nonce must be sent together with non-ajax requests. `Content-Security-Policy `headers seem to be ignored with XHR responses.

tandraschko:
I think it's not to hard:
- Create a ResponseWriterWrapper
- listen to startElement for "scripts"
- wait until child content will be written and do writeAttribute(nonce, ...)

cnsgithub:
> I think it's not to hard:
> 
> * Create a ResponseWriterWrapper
> * listen to startElement for "scripts"
> * wait until child content will be written and do writeAttribute(nonce, ...)

Hmm, that's exactly the way how it is currently implemented in this PR, isn't it?

tandraschko:
probably. i dindn't look at the ResponseWriter in detail, i just checked everything around.
And if we don't need hashes, we don't need that stuff like buffers and OnCommittedResponseWrapper. As we just add the nonce as header before render everything + add this noonce to all script tags.

melloware:
Looks like Jquery finally merged that CSP patch.  https://github.com/jquery/jquery/pull/4269#event-2072576252

tandraschko:
@cnsgithub 
why did you add "if (event.cancelable) { event.preventDefault();}" before call the actional listener?

cnsgithub:
@tandraschko Puh, cannot remember right now. This was the initial commit to the old branch: https://github.com/primefaces/primefaces/pull/3884/commits/ac61ed16fa9716009480cd9a777cac6c0265108d

The commit comment is saying 
> fix: after clicking a button the XHR got aborted and the default action (form submit) was performed resulting in two requests

I think it's necessary.

tandraschko:
This was actually prevented by "return false;" which is rendered by the client behavior stuff. but ok, i know the reason now :)

cnsgithub:
Yes. I think `return false` didn't work in this context and a general solution came out to be the `preventDefault` stuff.

tandraschko:
I will close the PR for now as we can't just apply it - Lets use the ticket for further discussion

PR #4273, opened 2018-11-22
Fix #4253: BubbleChartSeries switch x,y,radius to double from int.

### LINKED ISSUE #4253: BubbleChart: allow fractional X, Y, and radius values in BubbleChartSeries
BubbleChartSeries takes only integers for the X and Y values.  This is very limiting, especially when the numbers are fractional over a small range, for example, between 0 and 2.  Of course it is less limiting when the numbers are large, as in the showcase example.

## 1) Environment
- PrimeFaces version:
6.2 (Community)

No.

Not tested. However, the latest version in github still has ints.

Not relevant, but Tomcat 8.5.x

- Affected browsers:
All

## 2) Expected behavior
I'd like the BubbleChartSeries to have a fractional data type for X, Y, and Radius.  I presume the drawn sizes of the Radius would be scaled based on the maximum Radius, and similar to the Axes.

The values could then be float, double, or BigDecimal; any of these would be far better than int.

## 3) Actual behavior
Any attempt to add fractional data is clipped to integer values, which defeats the purpose for small values (e.g. less than 5 or so).

## 4) Steps to reproduce

Modify the showcase example to have small decimal values for prices (and Labels); potentially just put /100f behind each x, y, and radius value.

## 5) Sample XHTML

Same as in the showcase example.

## 6) Sample bean

Same as in the documentation example.  But here it is so you can save some CTRL-V pasting.

```
public class Bean {
    private BubbleChartModel model;
    public ChartBean() {
        model = new BubbleChartModel();
        model.add(new BubbleChartSeries("Acura", 70/100f, 183/100f, 55/100f));
        model.add(new BubbleChartSeries("Alfa Romeo", 45/100f, 92/100f, 36/100f));
        model.add(new BubbleChartSeries("AM General", 24/100f, 104/100f, 40/100f));
        model.add(new BubbleChartSeries("Bugatti", 50/100f, 123/100f, 60/100f));
        model.add(new BubbleChartSeries("BMW", 15/100f, 89/100f, 25/100f));
        model.add(new BubbleChartSeries("Audi", 40/100f, 180/100f, 80/100f));
        model.add(new BubbleChartSeries("Aston Martin", 70/100f, 70/100f, 48/100f));
        model.setTitle("Bubble Chart");
        model.getAxis(AxisType.X).setLabel("Price");
        Axis yAxis = model.getAxis(AxisType.Y);
        yAxis.setMin(0);
        yAxis.setMax(250/100f);
        yAxis.setLabel("Labels");
    }
    public BubbleChartModel getModel() { return model; }
}
```

melloware:


melloware:
@tandraschko resubmitted. I was getting that strange line ending error again with HtmlSanitizer so Git was requiring me to fix it to check it in.

PR #4274, opened 2018-11-22
Fix #4129: Add Legend customization to ChartJS models.

### LINKED ISSUE #4129: Chartjs: Make Legend customizable
## 1) Environment
- PrimeFaces version: 6.2.9
- Does it work on the newest released PrimeFaces version? No Version? 6.2.10
- Does it work on the newest sources in GitHub? No because there are no commits relatet to this issue
Payara 5.183
- Affected browsers:
All

## 2) Expected behavior

It should be possible to set parameters for the options.legend namespace inside chartsjs. 
Reference chartjs Documentation: [Chartjs legend namespace docs](https://www.chartjs.org/docs/latest/configuration/legend.html).

Example:
```
var chart = new Chart(ctx, {
    type: 'bar',
    data: data,
    options: {
        legend: {
            display: true,
            labels: {
                fontColor: 'rgb(255, 99, 132)'
            }
        }
}
});
```

## 3) Actual behavior

Actually it's not possible to set legend position, font color or type and if its showing up or not.  I know in most cases i can set this with the extender, but sometime i like to change some settings dynamicly for example by the user. This is difficult to made because its hard to control javascript content from Managed Bean.

## 4) Steps to reproduce

.. 

## 5) Sample XHTML

..

## 6) Sample bean

..

## 7) Forum
[Primefaces Forum link](https://forum.primefaces.org/viewtopic.php?f=3&t=57092)

melloware:
Following @mertsincan excellent patterns it was pretty easy to add Legend customization.  I also updated the BarChart example in the showcase to use the legend customization.

### DIFF
--- a/src/main/java/org/primefaces/component/barchart/BarChartRenderer.java
+++ b/src/main/java/org/primefaces/component/barchart/BarChartRenderer.java
@@ -16,11 +16,12 @@
 package org.primefaces.component.barchart;
 
 import java.io.IOException;
+
 import javax.faces.component.UIComponent;
 import javax.faces.context.FacesContext;
 import javax.faces.context.ResponseWriter;
+
 import org.primefaces.component.charts.ChartRenderer;
-import org.primefaces.model.charts.ChartModel;
 import org.primefaces.model.charts.bar.BarChartOptions;
 import org.primefaces.util.WidgetBuilder;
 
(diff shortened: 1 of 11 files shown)


mertsincan:
Good PR! Thanks @melloware ;)

PR #4277, opened 2018-11-23
Fix #953: Schedule convert to double first in case browser sends decimal.

### LINKED ISSUE #953: p:schedule throw NumberFormatException when resize the ScheduleEvent box to a minimum fast
1:
p:schedule throw NumberFormatException when resize the ScheduleEvent box to a minimum fast
Exception message: 
java.lang.NumberFormatException: For input string: "-37.4"
    at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
    at java.lang.Integer.parseInt(Integer.java:580)
    at java.lang.Integer.valueOf(Integer.java:766)
    at org.primefaces.component.schedule.Schedule.queueEvent(Schedule.java:449)
    at org.primefaces.behavior.ajax.AjaxBehaviorRenderer.decode(AjaxBehaviorRenderer.java:47)
    at javax.faces.component.behavior.ClientBehaviorBase.decode(ClientBehaviorBase.java:132)
    at org.primefaces.renderkit.CoreRenderer.decodeBehaviors(CoreRenderer.java:530)

2:
The moment.js file was not loaded when the page(with p:schedule in it) rendered partially by using  ui:include .
p:printer has the same problem. The printer.js file not loaded.

melloware:
We have seen this with other components where some browsers like Firefox decide to send 37.4 instead of 37.  So its safer to convert to Double first and then convert to int.

### DIFF
--- a/src/main/java/org/primefaces/component/schedule/Schedule.java
+++ b/src/main/java/org/primefaces/component/schedule/Schedule.java
@@ -154,8 +154,8 @@ public class Schedule extends ScheduleBase {
             else if (eventName.equals("eventResize")) {
                 String resizedEventId = params.get(clientId + "_resizedEventId");
                 ScheduleEvent resizedEvent = getValue().getEvent(resizedEventId);
-                int dayDelta = Integer.parseInt(params.get(clientId + "_dayDelta"));
-                int minuteDelta = Integer.parseInt(params.get(clientId + "_minuteDelta"));
+                int dayDelta = Double.valueOf(params.get(clientId + "_dayDelta")).intValue();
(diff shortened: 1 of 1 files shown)


melloware:
@OBreidenbach please review.

OBreidenbach:
As the same is already down for 'eventMove', that should be fine. 

- In case of 'eventMove' a cast '(int)' is used. Lines: 135+136
- In case of 'eventResize' you call '.intValue(). Lines 157+158

Maybe you could use the same version for both cases?
Beside that, everything is fine.

melloware:
Great catch.   I just updated my PR so both are consistent withe Double.valueOf().intValue() only because I i think its cleaner and more explicit.

PR #4281, opened 2018-11-27
Chartjs.js uses tooltips instead of tooltip

efrenperez:


tandraschko:
Please create a valid issue before, otherwise PR's will not be applied.

tandraschko:
@mertsincan Can you review this?

mertsincan:
Good catch ;) Thanks a lot for the PR!

PR #4286, opened 2018-11-27
Fix #3760: Calendar date select event should respect locale.

### LINKED ISSUE #3670: p:multiSelectListBox  doesn't have ajax event available.
- Expected behavior:  MultiSelectListBox will use to change event listener.
- Actual behavior:       It throws NullPointerException.

### LINKED ISSUE #3760: Calendar: Ajax event DateSelect not properly supplying locale specific value
I am using the p:calendar with mode="inline" and supplying pattern "d-MMM-yyyy HH:mm" and locale of "nl".  I am also using the p:ajax within the component for event "dateSelect" with a process="@this".

If I choose the month of 'mei' and select the 9th day I get the following error "Value '9-May-2018 00:00' could not be understood as a date and time.".  The issue is the component is not properly using 'mei' rather it is using 'May'.  But when trying to parse to a date it is using the passed in Locale which doesn't jive with the value 'May'.  So the ajax event somehow is not translating the selected value per the locale

## 1) Environment
- PrimeFaces version: 6.2.1
- Does it work on the newest released PrimeFaces version? No Version? 6.2.5
- Affected browsers: ALL

## 2) Expected behavior
The value that is parsed should be '9-mei-2018 00:00'
...

## 3) Actual behavior
The value that is parsed is '9-May-2018 00:00' using 'nl' locale which doesn't understand 'May'
..

## 4) Steps to reproduce
Create the component in step #5

.. 

## 5) Sample XHTML
<p:calendar widgetVar="fromCalendar"  mode="inline" id="startDateFilter" value="#{form.startDate}" pattern="d-MMM-yyyy HH:mm" locale="nl">
						<p:ajax listener="#{form.search}"  event="dateSelect" process="@this"  global="false" />
</p:calendar>
..

## 6) Sample bean
Create a form bean that has search method and startDate field with getters and setters.
..

melloware:
Thanks to @CJBinder!

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/calendar/calendar.js
+++ b/src/main/resources/META-INF/resources/primefaces/calendar/calendar.js
@@ -194,7 +194,7 @@ PrimeFaces.widget.Calendar = PrimeFaces.widget.BaseWidget.extend({
                 }
             }
             else {
-                // GitHub #3670 Pass the config settings to TimePicker
+                // GitHub #3760 Pass the config settings to TimePicker
                 var settingsObj = {
                        settings : $this.cfg
                 };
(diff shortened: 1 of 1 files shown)


melloware:
Sorry fixes #3760

PR #4287, opened 2018-11-27
Fix #4285: BreadCrumb missing span for disabled menu items.

### LINKED ISSUE #4285: BreadCrumb: missing </span> when item is disabled
When I create a BreadCrumb with disabled MenuItem the generated xml is not valid, one span is not closed.

Test code:
```
<p:breadCrumb>
	<p:menuitem value="Categories" url="#" disabled="true"/>
</p:breadCrumb>
```

Renders like this:
```
<div id="j_idt124:j_idt125" class="ui-breadcrumb ui-module ui-widget ui-widget-header ui-helper-clearfix ui-corner-all" role="menu">
<ul>
	<li role="menuitem">
		<span class="ui-menuitem-link ui-corner-all ui-icon ui-icon-home ui-state-disabled">
			<span class="ui-menuitem-text">
				Categories
			</span>
	</li>
</ul>
</div>
```

One `</span>` is missing.

melloware:
Disabled menu item had a mismatch of span elements.


tandraschko:
Thanks for the very fast pr :D

PR #4288, opened 2018-11-27
Fix #3765: Calendar Inline mode not setting time sliders correctly.

### LINKED ISSUE #3765: Calendar: Inline mode with date/time pattern, sliders don't render properly when supplied with Date object with date and time
We use the calendar in mode="inline" and use it to allow users to enter valid date ranges.  If the user select a date and lets say changes the time to 10 for hour and 10 for minute, the date supplied is accurate.  However when redisplaying the component the sliders end up with 0 for hour and 0 for minute using the same date.

## 1) Environment
- PrimeFaces version: 6.2.1
- Does it work on the newest released PrimeFaces version? No Version? 6.2.5
- Affected browsers: ALL

## 2) Expected behavior
When displaying the calendar as inline and using a pattern that includes HH:mm and the date supplied for the value includes time of 10:10, the sliders show 0 and 0.  There is an javascript error stating it can't parse the date time value
...

## 3) Actual behavior
When displaying the calendar as inline and using a pattern that includes HH:mm and the date supplied for the value includes time of 10:10, the sliders should show 10 and 10. 
..

## 4) Steps to reproduce
Create a button on a form that displays a dialog popup that shows the value of a date using the following xhtml.  Change the value of the date by supplying hour and minute.  Click ok on the form to submit the value.  Next click the button again which should render the calendar with the date/time you previously specified and notice the time sliders are at zero and if you look at the javascript console in a browser you will see a parsing error.

## 5) Sample XHTML
<p:calendar mode="inline" id="endDateFilter" value="#{cc.attrs.filterForm.selectedFilter.endDate}" pattern="d-MMM-yyyy HH:mm" ></p:calendar>

## 6) Sample bean
Basic pojo that has Date with getter and setter

melloware:
Thanks to @CJBinder for his suggested fix: https://github.com/primefaces/primefaces/issues/3765#issuecomment-395407187

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/jquery/extensions.txt
+++ b/src/main/resources/META-INF/resources/primefaces/jquery/extensions.txt
@@ -28,3 +28,4 @@ jquery.ui.pfextensions:
 - Added gotoToday selection fix (https://github.com/primefaces/primefaces/issues/1420)
 - Sortable performance fix (/https://github.com/primefaces/primefaces/issues/3675)
 - Timepicker Fix (https://github.com/primefaces/primefaces/issues/2811)
+- Timepicker Fix (https://github.com/primefaces/primefaces/issues/3765)
--- a/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.js
+++ b/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.js
@@ -307,6 +307,16 @@ $.widget( "ui.sortable", $.ui.sortable, {
                     this.$input.trigger("change"); // PrimeFaces https://github.com/primefaces/primefaces/issues/2811
                 }
             },
+            
+            // PrimeFaces https://github.com/primefaces/primefaces/issues/3765
+            _addTimePicker: function (dp_inst) {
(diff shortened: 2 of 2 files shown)


tandraschko:
`PrimeFaces is not initalizing that alt field`
Wouldn't it be possible to initialize the alt field, like mentioned in the issue?

melloware:
Like initialize it somewhere in the widget component?

tandraschko:
maybe yes

please investigate a bit, so we can find the best solution

melloware:
OK let me look.

melloware:
OK I have investigated this.  Because PrimeFaces (Optimus) specifically chose to implement this component using a single input submit field and not separate time and date inputs.  TimePicker wants you to have 2 input fields 1 for date and 1 for the time component.  To make it work with PF they decided to use a single Input field and not store the time in the altinput field.

Switching to use the AltInput field I think would be a pretty massive change for this component.  We would need to..
a) create a second '_input' field called '_timeinput'.
b) make that new Input the altInput field.
c) Modify the PF Calenderrender on Submit to combine the two input submitted values into a date + time.
c) revisit all other Calendar code that is expecting a single Input submit value

So to me @CJBinder's solution works quite well for how PF decided to implement this.

tandraschko:
Thanks

PR #4298, opened 2018-11-29
Fix #4289

### LINKED ISSUE #4289: InputMask: server-side validation of mask attribute skips required-validation
## 1) Environment
- PrimeFaces version: 6.3 SNAPSHOT as of 28.11.2018 09:30 UTC+1
- Application server + version: Wildfly 10.1, JSF Mojarra 2.2.13.SP1 20160303-1204
- Affected browsers: All

## 2) Expected behavior
Input mask with `required = true` should validate and set `validationFailed = true` when nothing or something invalid (not matching the mask) was inserted into the field.

## 3) Actual behavior
Submitted value of the input mask will be set to `null` and therefore the whole validation is skipped. UIInput#validate(FacesContext) contains the code
```
// Submitted value == null means "the component was not submitted
// at all".  
Object submittedValue = getSubmittedValue();
if (submittedValue == null) {
    return;
}
```

This behavior was introduced by  #3234 / #3237

## 4) Steps to reproduce
Add an <p:inputMask requiired="true" ... /> to your page and insert nothing / something invalid according to the mask. Submit the field. The input is accepted as valid and the whole JSF lifecycle is executed including invoke-application, etc

## 5) Sample XHTML
```
<h:form>
    <p:inputMask value="#{myBean.maskedValue}" mask="99:99" required="true" />
    <p:commandButton value="execute" process="@form" update="@form" actionListener="#{myBean.doProcessMaskedValue}" />
</h:form>
```

## 6) Sample bean
```
@ViewScoped
@Named

    private String _maskedValue;

    public void doProcessMaskedValue() {
        System.out.prinln("Masked value should not be NULL, current value is = " + _maskedValue);
    }

    public String getMaskedValue() {
        return _maskedValue;
    }

    public void setMaskedValue(String maskedValue) {
        _maskedValue = maskedValue;
    }
}
```

..

fanste:
The submittedValue should not be set to null when the  submittedValue does not match the mask attribute. Instead set it to empty string. This ensures that the validation of the component is correctly executed.

### DIFF
--- a/src/main/java/org/primefaces/component/inputmask/InputMaskRenderer.java
+++ b/src/main/java/org/primefaces/component/inputmask/InputMaskRenderer.java
@@ -46,10 +46,10 @@ public class InputMaskRenderer extends InputRenderer {
         String clientId = inputMask.getClientId(context);
         String submittedValue = context.getExternalContext().getRequestParameterMap().get(clientId);
 
-        if (submittedValue != null) {
+        if (submittedValue != null && !submittedValue.isEmpty()) {
             Pattern pattern = translateMaskIntoRegex(context, inputMask);
             if (!pattern.matcher(submittedValue).matches()) {
-                submittedValue = null;
+                submittedValue = "";
             }
 
             inputMask.setSubmittedValue(submittedValue);


melloware (approved):

tandraschko:
So the regex-validation on serverside is still working + the default validators are also executed?

fanste:
At least in the application that I have here to test it.
 - no input + required = valid false
 - invalid input + required = valid false
 - correct input = full lifecycle executed

PR #4301, opened 2018-11-29
Fix #4300 - Calendar: Change _self to $this

### LINKED ISSUE #4300: Calendar: Typo in javascript
## 1) Environment
- Affected browsers: All

## 2) Description
Calendar with `mode=inline` + `event=dateSelect` will throw a JS error because of a typo in the JS file introduced in the bugfix of #3760

calendar.js line 201 => _self instead of $this

PR #4301

fanste:


melloware:
good catch.  So weird My browser did not throw a JS error when I tested this.  I have a plugin to notify me of JS errors and it didn't throw it.  I wonder if my JS was cached.  Thanks for fixing.

melloware (approved):

PR #4304, opened 2018-12-01
Fix #3270: Autocomplete fire clear event when forceselection=true.

### LINKED ISSUE #3270: AutoComplete: Trigger clear event when input is cleared by forceSelection=true
## 1) Environment
- PrimeFaces version: 6.2.RC1
- Affected browsers:

## 2) Expected behavior
Fire the `clear` event when the input is emptied by the setting `forceSelection=true.

## 3) Actual behavior
The newly added `clear` event ( #2846 ) is trigger if the input is emptied using backspace, etc. But if the autoComplete is set to `forceSelection=true` the event is not triggered if no valid item was selected.

## 4) Steps to reproduce
- Setup an autoComplete with `forceSelection=true` and add an `p:ajax event="clear"` to it.
1. 
    - Type something (in this example: TEST) and select an item
    - Remove it again using the backspace key. The `clear` event is fired
2.
    - Type something (in this example: TEST) again and select an item
    - Remove some chars from the input but not all. Leave the input field. `forceSelection` will clear the input field but NO `clear` event is fired
3.
    - Type something (in this example: TEST) again and NOT select an item
    - Leave the input field. `forceSelection` will clear the input field but NO `clear` event is fired

## 5) Sample XHTML
```
<p:autoComplete value="#{bean.txt1}" completeMethod="#{bean.completeText}" forceSelection="true">
     <p:ajax event="clear" listener="#{bean.onClear}" update="msgs" />
</p:autoComplete>
```

## 6) Sample bean
```
@RequestScoped
@Named
public class Bean {
    private String txt1;

    public List<String> completeText(String qry) {
        if(qry != null && qry.startsWith("TEST")) {
            return Arrays.asList("TEST", "TEST1", "TEST2");
        }

        return Collections.emptyList();
    }

    public String getTxt1() {
        return this.txt1;
    }

    public void setTxt1(String txt1) {
        this.txt1 = txt1;
    }
}
```

melloware:
@fanste can you review this fix. Basically when forceselection=true when it checks whether there is a match an there is no match it sets the Input to ''.  I just added the fireClearEvent to happen when that occurs.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
+++ b/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
@@ -941,6 +941,7 @@ PrimeFaces.widget.AutoComplete = PrimeFaces.widget.BaseWidget.extend({
             if(!this.cfg.multiple) {
                 this.hinput.val('');
             }
+            this.fireClearEvent();
         }
 
         return valid;


fanste:
Seems to work. I've added the fix to the application and the clear event is now fired when forceSelection clears the input.

Thank you.

fanste (approved):

Issue #4263, opened 2018-11-20
OverlayPanel: Top position out of view


tdtappe:
1) Environment
Reproduceable with current showcase (6.2.11), primefaces-test-master sample project (6.2 - see below) and my app (6.2.1 with Wildfly 13.0.0.Final)
Browsers: Firefox 63.0.3 and Chrome 70.0.3538.110

2) Expected behaviour
The top of the overlay panel should always be visible (no negative top-value)

3) Actual behaviour
If the browser window is too small the overlay panel is moved to the top of the bound element. But too far so that the top of the overlay panel is no longer visible (exceeds the top of the browser window).

4) Steps to reproduce
See sample project: Decrease the size (height) of the browser window and click the test button. There is enough space to put the overlay panel at the bottom of the panel but it's moved to the top. But even is there wasn't enough space below the button the top of overlay panel should never get out of view.

5) Sample XHTML
See sample project

6) Sample bean
See sample project

Sample project: https://github.com/tdtappe/primefaces-test/tree/overlay-panel-position

fanste:
The fix in #4309 works in my local application. But I can't tell you if this is the best / performant fix.

melloware:
Looks good to me but I will let @tandraschko review it and see.

PR #4308, opened 2018-12-03
Fix #4307: OverlayPanel - broken when dynamic=true and re-rendered

### LINKED ISSUE #4307: OverlayPanel: Dynamic panel is broken when parent container is re-rendered
## 1) Environment
- Affected browsers: ALL
- Similar issue: #1671

## 2) Expected behavior
OverlayPanel with `dynamic = true` is inside a DataTable. The pagination of the DataTable should not break the OverlayPanel once the panel content was loaded.

## 3) Actual behavior
An empty panel is shown, if the OverlayPanel was part of an AJAX re-render

## 4) Steps to reproduce
Hover over the icon, wait till the panel is shown. Change the page of the DataTable and again back to the old page. Hover the same icon. Empty panel is shown.

## 5) Sample XHTML
```
<p:dataTable value="#{[1,2,3,4,5,6,7,8,9,10]}" var="trow" pagination="true" rows="2">
	<p:column>row #{trow}</p:column>
	<p:column>
		<p:commandLink id="clNotes"><i class="fa fa-a fa-file-o"></i></p:commandLink>
		<p:overlayPanel for="clNotes" dynamic="true" showDelay="1000" showEvent="mouseover">
			<dl>
				<ui:repeat value="#{[1, 2, 3, 4]}" var="note">
				<dt><b>[date-#{note}]</b> FROM</dt>
				<dd>note-#{note} note-#{note} note-#{note} note-#{note} note-#{note} note-#{note}</dd>
				</ui:repeat>
			</dl>
		</p:overlayPanel>
	</p:column>
</p:dataTable>
```

## 6) Sample bean
Not needed

fanste:


melloware (approved):

melloware:
Since you were just looking at OverlayPanel want to investigate this one too?  https://github.com/primefaces/primefaces/issues/4263

I think its just a matter of making sure the top value is never negative?

PR #4309, opened 2018-12-03
Fix #4263: OverlayPanel - Ensure top is not negative

### LINKED ISSUE #4263: OverlayPanel: Top position out of view

fanste:


tdtappe:
Does not work as I would expect it to work. I think the top position has to be checked with its "absolute" value. If the overlay panel is for example inside a parent at the bottom of the page then a negative relative top position should be allowed. But the "absolute" value should never be negative.

tandraschko:
@fanste Can you rework your PR based on the input from @tdtappe ?

fanste:
This week I'm out of time. 

@tdtappe Could you create a test case that reproduces the behavior from your comment? That would make it easier for me.

tdtappe:
In my sample project I added another button at the bottom of the page that also shows an overlay panel. And this panel is mostly out of view though it doesn't have to be that way because there's a lot of space available above it.

tandraschko:
@fanste can you work on @tdtappe s Feedback?

fanste:
I will work on this again when I'm back from holidays.

tandraschko:
i will close the PR for now. If you reworked it, you can still create a new one.

Issue #2358, opened 2017-05-12
The form attribute in p:menuItem is ignored


zhizhideng:
1) Environment
- PrimeFaces version: 6.1

2) Expected behavior
According to document, if I specify the the "form" attribute of a menuItem, it should use the specified form for submission (It works for p:ajax). 

In our case, we are using jQuery.load loading external contents for a panel, to avoid nested forms we have to remove any outer form and use the "form" attribute to specify the actual submission form.  

3) Actual behavior
If the menuItem has a parent form, it will will use that parent form for submission. If the menuItem doesn't have a parent form, a FacesException saying "MenuItem must be inside a form element" will be thrown.

4) Quick inspection

In org.primefaces.component.menu.BaseMenuRenderer#encodeMenuItem, there is null check preventing menuItem outside a form to be rendered.

```
                UIComponent form = ComponentTraversalUtils.closestForm(context, menu);
                if(form == null) {
                    throw new FacesException("MenuItem must be inside a form element");
                }
```
While in  org.primefaces.component.menu.BaseMenuRenderer#buildAjaxRequest method, the form attribute of the menuItem is not used.

```
        builder.init()
                .source(clientId)
                .process(menu, source.getProcess())
                .update(menu, source.getUpdate())
                .async(source.isAsync())
                .global(source.isGlobal())
                .delay(source.getDelay())
                .timeout(source.getTimeout())
                .partialSubmit(source.isPartialSubmit(), source.isPartialSubmitSet(), source.getPartialSubmitFilter())
                .resetValues(source.isResetValues(), source.isResetValuesSet())
                .ignoreAutoUpdate(source.isIgnoreAutoUpdate())
                .onstart(source.getOnstart())
                .onerror(source.getOnerror())
                .onsuccess(source.getOnsuccess())
                .oncomplete(source.getOncomplete())
                .params(params);
        
        if(form != null) {
            builder.form(form.getClientId(context));
        }
```

melloware:
You are right the documentation on Page 340 of the User Guide for MenuItem says "form null String Form to serialize for an ajax request. Default is the enclosing form."

So to men your Menu always has to be wrapped in a parent form however the problem is it is not respecting the form="" on the menu item.  Let me look closer at it.

melloware:
OK I think the documentation is wrong.  MenuItem doesn't even have a "form" attribute.

```java
public interface MenuItem extends MenuElement, Confirmable {
        
    public String getIcon();
    
    public String getIconPos();
[...]

PR #4319, opened 2018-12-09
RemoteCommand: consider AjaxSource.getForm()

### LINKED ISSUE #4317: RemoteCommand: The form attribute is ignored
## 1) Environment
- PrimeFaces version: 6.2
- Application server + version: WildFly 14
- Affected browsers: Any

## 2) Expected behavior

Using `<p:remoteCommand form=":rootForm" />` outside a `h:form` should render fine.

## 3) Actual behavior

Using `<p:remoteCommand form=":rootForm" />` outside a `h:form` causes a `javax.faces.FacesException: RemoteCommand 'doStuff'must be inside a form.`

## 4) Steps to reproduce

* create a plain view file `remote-command-test.xhtml`
* insert `<p:remoteCommand form=":rootForm" />` (*not* nested inside a `h:form`, see below)
* try to render the view.

## 5) Sample XHTML

```
<!DOCTYPE html>
<h:body>

  <h:form id="rootForm" />

  <p:remoteCommand form=":rootForm" />

</h:body>
</html>
```

Like #2358 this is caused by the [renderer](https://github.com/primefaces/primefaces/blob/ae1e952d897e310de6c5b6acb09aa3851b22f870/src/main/java/org/primefaces/component/remotecommand/RemoteCommandRenderer.java#L62) which does not consider the form attribute and instead just looks for the enclosing form. Right now only the [AjaxBehaviorRenderer](https://github.com/primefaces/primefaces/blob/ae1e952d897e310de6c5b6acb09aa3851b22f870/src/main/java/org/primefaces/behavior/ajax/AjaxBehaviorRenderer.java#L97) considers the form attribute at all.

schulzp:
fixes #4317

### DIFF
--- a/src/main/java/org/primefaces/component/remotecommand/RemoteCommandRenderer.java
+++ b/src/main/java/org/primefaces/component/remotecommand/RemoteCommandRenderer.java
@@ -15,8 +15,6 @@
  */
 package org.primefaces.component.remotecommand;
 
-import java.io.IOException;
-
 import javax.faces.FacesException;
 import javax.faces.component.UIComponent;
 import javax.faces.component.UINamingContainer;
@@ -25,8 +23,11 @@ import javax.faces.context.ResponseWriter;
 import javax.faces.event.ActionEvent;
 import javax.faces.event.PhaseId;
 
+import java.io.IOException;
+
 import org.primefaces.component.api.AjaxSource;
 import org.primefaces.context.PrimeRequestContext;
+import org.primefaces.expression.SearchExpressionFacade;
 import org.primefaces.renderkit.CoreRenderer;
 import org.primefaces.util.AjaxRequestBuilder;
 import org.primefaces.util.ComponentTraversalUtils;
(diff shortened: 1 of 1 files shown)


melloware (commented):
Nice work.  You might be able to
Simplify it some more by only checking traverse utils if the formId is null. Right now you are always executing traverse utils and search facade every time.

tandraschko:
Thanks for the PR.
The PR in generell is very good but could we skip the "closest form lookup", if a form was specified on the component?
I think thats the same for other components, it's a very small but good perf improvement.

tandraschko:
I will apply it now, i will come back with my idea.

Rapster:
The comment of your commit sounds actually much better

melloware:
@tandraschko I had the same thought wouldn't this be a better fix?  Or maybe we could wrap it it up in ComponentUtils so we can reuse it.

```java
 String formId = SearchExpressionFacade.resolveClientId(context, component, source.getForm());
        if (formId == null) {
            UIComponent form = ComponentTraversalUtils.closestForm(context, command);
            if (form == null) {
                throw new FacesException("RemoteCommand '" + name + "' must be inside a form or reference a form via its form attribute.");
            }
            formId = form.getClientId(context);
        }
...
        String request = builder.init()
                .source(clientId)
                .form(formId)
```

tandraschko:
Please review: https://github.com/primefaces/primefaces/commit/c7f8c9539ff07ce291a7f6e9f24fe4ad33431b79

melloware:
That is a fantastic fix.  Huge improvement.

Rapster:
Regarding this ticket, why not calling `CoreRenderer#buildAjaxRequest` looks like it does the same no?

melloware:
@tandraschko I believe your refactoring now fixes this issue as well:  https://github.com/primefaces/primefaces/issues/2358

tandraschko:
@Rapster because it uses preventDefault.
It probably needs a refactor after the release. the CoreRendere#buildAjaxRequest should not return a string, it should return the builder.
So the CoreRenderer could do the default mapping from AjaxSource<>AjaxRequestBuilder. And each renderer could "enhance" the builder with stuff like preventDefault e.g.

Rapster:
Ah ok, fair enough. We should create a ticket for the refactoring part then :wink:

schulzp:
Cool, thanks a lot @tandraschko, @melloware, and @Rapster!

PR #4321, opened 2018-12-10
Fix #4320: Picklist performance improvement. Remove animation effect …

### LINKED ISSUE #4320: Picklist: filter is causing script error and browser crash for more than 4000 items
# IMPORTANT :
Picklist filter is not working properly.It is taking too much time to filter the data.We have tried to change the browser setting to resolve this issue like setting the value of dom.max_chrome_script_run_time;0 and
dom.max_script_run_time;0, but it didn't help.It just avoided the script error message but browser was still getting crashed and it is taking 2 min 47 seconds to filter data from 4000 items..
It is an very urgent issue for me. Could anyone please look into it and provide me a solution or workaround for it ASAP.

## 1) Environment
- PrimeFaces version:6.2.9./6.2.12.(In 6.1 it works properly)
- Does it work on the newest released PrimeFaces version?No Version?6.2.12
- Does it work on the newest sources in GitHub?No (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source)
- Affected browsers: All

## 2) Expected behavior: After typing in picklist filter, it should filter the value from picklist.It should not take much time to filter and browser should not crash.

...

## 4) Steps to reproduce : 
1- Add more than 4000 items in the picklist (primeface version 6.2.9 or 6.2.12).
2- Add showSourceFilter="true" showTargetFilter="true" filterMatchMode="contains".
3- Try to filter the picklist items using filter text area.
4 - It will take approx 3 minutes to filter the picklist or script error(see attachment) or Browser crash.

.. 

## 5) Sample XHTML--
```xml
<p:pickList id="pickList"  value="#{userActivation.users}" var="users" showCheckbox="true" effect="none" itemLabel="#{users}"  itemValue="#{users}" showSourceFilter="true" showTargetFilter="true" filterMatchMode="contains" filterEvent="enter">
		<p:ajax event="transfer" onstart="PF('bui').show()" oncomplete="PF('bui').hide()"> </p:ajax>
			
								
			<f:facet name="sourceCaption">#{msg['msg.inactiveUser']}</f:facet>
        	<f:facet name="targetCaption">#{msg['facet.ActiveUsers']}</f:facet>
        	
        	<p:column style="width:10%">
								<h:outputLabel id="img_#{users}" styleClass="ui-user"/>
			</p:column>
			<p:column style="width:90%">
							<h:outputLabel id="txt_#{users}" value="#{users}"/>
			</p:column>
        </p:pickList>
```

..

## 6) Sample bean--

```java
@ManagedBean
@ViewScoped
public class UserActivation {

	/**
	 * Logger Object to print log
	 */
	private static Logger logger = Logger
			.getLogger("");

	private DualListModel<String> users = new DualListModel<String>(
			Collections.<String> emptyList(), Collections.<String> emptyList());
	private List<String> usersSource;
	private List<String> usersTarget;
	
	@PostConstruct
	public void init() {
		final String lBuf = " ";
		logger.info(lBuf + "");
		try {
			IDfSession session = SessionManager.instance()
					.getWEBSession();

			
			String superUserID = DAOFactory
					.getContextConfig()
					.getString(Constants.DOCBASE_SUPERUSER_NAME).trim();
			
			AdminAPI adminApi = new AdminAPI();
			
			String superUserName= adminApi.getFullNameByUserID(superUserID, session);
			usersSource = adminApi.getUsersAllByGroupName(
					AdminConstants.grp_global_none, session);
			usersTarget = adminApi.getUsersAllByGroupName(
					AdminConstants.grp_global_users, session);
			if(usersTarget.contains(session.getLoginUserName())){
				usersTarget.remove(session.getLoginUserName());
				usersTarget.remove(superUserName);
			}
			if (usersSource == null) {
				usersSource = new ArrayList<String>(
						Collections.<String> emptyList());
			}
			if (usersTarget == null) {
				usersTarget = new ArrayList<String>(
						Collections.<String> emptyList());
			}
			users = new DualListModel<String>(usersSource, usersTarget);
		} catch (AdminException exp) {
			logger.error(lBuf + exp);
		} catch (Exception exp) {
			logger.error(lBuf + exp);
		} finally {
			logger.info(lBuf + "");
		}
	}
}	
```

Please let me know if you need any further information.

melloware:
Two problems:

1. UI Update was being done INSIDE the for loop every time on the filter.

2. Animation effect on every loop of the filter.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/picklist/picklist.js
+++ b/src/main/resources/META-INF/resources/primefaces/picklist/picklist.js
@@ -519,28 +519,14 @@ PrimeFaces.widget.PickList = PrimeFaces.widget.BaseWidget.extend({
                 matches = this.filterMatcher(itemLabel, filterValue);
 
                 if(matches) {
-                    if(animated) {
-                        item.fadeIn('fast', function() {
-                            $this.updateListRole();
-                        });
-                    }
-                    else {
-                        item.show();
-                        this.updateListRole();
-                    }
+                    item.show();
                 }
                 else {
-                    if(animated) {
-                        item.fadeOut('fast', function() {
-                            $this.updateListRole();
-                        });
-                    }
-                    else {
-                        item.hide();
-                        this.updateListRole();
-                    }
+                    item.hide();
                 }
             }
+            
+            this.updateListRole();
         }
 
     },


tandraschko:
It completely removes the effect now?

melloware:
I guess I can restore it.  I felt it kinda useless but let me put it back in after the for loop.

melloware:
actually @tandraschko any suggestions it was doing a Fade IN every time it added an item and Fade Out every time it removed an item.  What effect should I use considering in 4000 items there could be 700 Fade Ins and 3300 Fade Outs?  That was why I removed it.

melloware:
Should I just pick Fade Out?

tandraschko:
Could the user should set effect=none on the p:picklist?

melloware:
YEs the user can set the effect="none" I tested that and the performance is still horrible because its calling `this.updateListRole();` in the loop. If you look at `this.updateListRole();` its doing this expensive operation!

```javascript
updateListRole: function() {
        this.sourceList.children('li:visible').length > 0 ? this.sourceList.attr('role', 'menu') : this.sourceList.removeAttr('role');
        this.targetList.children('li:visible').length > 0 ? this.targetList.attr('role', 'menu') : this.targetList.removeAttr('role');
    }
```

melloware:
Should we have @mertsincan review just to get another set of eyes on it?  I have the reproducible example included in the ticket.

tandraschko:
What about doing updateListRole outside, if the effect is none? Should fix the performance but the effect should still work.

melloware:
I thought of that but if you look at the effect code what its doing is calling `this.updateListRole();' inside the effect to animate it on every for loop.  Can I do an animation if none of the items are actually updated until AFTER the for loop finishes and `this.updateListRole();` is called.

tandraschko:
I think we can just move the updateListRole outside, when !animated?

The animation itself is ok. If someone uses 4k items, he should be able to set effect=none and the speed should be good.

melloware:
Actually can you look at that code again.  If I am seeing it right, it calls Animate if animated=true but if animated= false it calls item.show() or item.hide(). So why is it even animating if it hasn't even shown or hidden anything in the list.  It doesn't make any sense?  

```javascript
                   if(animated) {
                        item.fadeIn('fast', function() {
                            $this.updateListRole();
                        });
                    }
                    else {
                        item.show();
                        this.updateListRole();
                    }
```

Am I reading that right that if its animated just repaint the list but if its not animated then update the visibility of the item and then repaint the list?  

The animation per for loop just doesn't make any sense.  IF anything all the items should be changed and then an animation effect applied after the for loop?

mertsincan:
Thanks @melloware and @tandraschko. I'll check the PR

mertsincan:
Thanks a lot for the PR, @melloware ;) I made some changes in it. Please see;
https://github.com/primefaces/primefaces/commit/ce42383127fca7154dbebc9b54a2091dd3e9fad6

fadeIn/fadeOut are async methods. Therefore, we need to update role attribute in oncomplete callback of these methods.

Thanks again.

melloware:
Ahh nice fix.  Thanks for checking @mertsincan !

Issue #4332, opened 2018-12-17
Modernize Documentation


cagataycivici:
Documentation is currently done offline using a word processor, we should make expose it to markdown or similar to committer can maintain it.

tandraschko:
Finally! :) Thats a good decision

Rapster:
Sounds awesome! Does the final support has to be a PDF? If not, we could simply use github pages. We could use primefaces/primefaces repo itself, I can work on a first draft if that's okay?

cnsgithub:
Good decision!

Maybe we could also define some templates (e.g. freemarker or velocity) to automatically get attributes and descriptions out of `primefaces-p.taglib.xml`. This would ensure code and documentation to be in sync.

melloware:
So just throwing my two cents in Spring Boot projects use something called AsciiDoc.  http://asciidoc.org/

The files are all stored in GitHub, but the output can be HTML, PDF, ePUB etc.  For an example of what Spring boot documentation looks like check out this: https://cloud.spring.io/spring-cloud-netflix/2.0.x/

It generates single page HTML or MultiPage and I think it looks quite nice.

cnsgithub:
> So just throwing my two cents in Spring Boot projects use something called AsciiDoc. http://asciidoc.org/

I'd like that! Introducing a table of content or things like that work like a charm.

We could have included static as well as dynamically generated contents.

cnsgithub:
Automatically publishing generated docs via `mvn site `to GitHub Pages would be cool: https://github.github.com/maven-plugins/site-plugin/

cnsgithub:
There's the [Asciidoctor ](https://asciidoctor.org/) (also available via `asciidoctor-maven-plugin`) that is able to generate different formats like HTML from `adoc` files.

melloware:
@cnsgithub this is an example Spring Boot project doc so you can see the layout of the .adoc pages and the Maven pom.xml to generate.  https://github.com/codecentric/spring-boot-admin/tree/master/spring-boot-admin-docs

tandraschko:
TomEE also uses adoc + JBake.
Here is a small history what TomEE did in the past: http://tomee-openejb.979440.n4.nabble.com/Merging-Old-and-New-Websites-td4685488.html

mertsincan:
![image](https://user-images.githubusercontent.com/11868120/50093010-93007300-0220-11e9-81cd-913addc9938b.png)

[...]

PR #4334, opened 2018-12-17
Proposal #4332 - Modernize Documentation

Rapster:
Solution for the new documentation:

It's text based so the community will finally be able to contribute, the syntax is easy and has a lot of nice features. Also, we think it's a good approach to generate few infos from taglib.xml, faces-config.xml, js files instead of writing it manually. To give you an idea, I think most of the issues labelled documentation could be closed thanks to this new approach. 

The whole thing is handled by the plugin primefaces-asciidoc (only available on my profile for now). But as I suggested to Thomas, instead of having repos for primefaces, we could managed them as maven modules. It'll be a lot more convenient for us but also for the community. Why a new plugin? Because it can't be done otherwise, and I definitely prefer this way. Though, adoc files are stored in primefaces repo.

So in few words, primefaces-asciidoc is a parsing & transformation engine. It expects few arguments from primefaces to configure it, basically paths to folder or files. 

Under the hood, it uses:

- AsciiDoctor: Obviously...
- JAXB: Parse XML files into Java objects, makes life easier
- Freemarker: That's how we're able to generate component properties dynamically and also it defines what a component doc should look like (see component.ftl). 

That whole processing is called by calling a simple command: include::p-inputText[] from index.adoc (the main "page"), so what does it do exactly?

1. First, make sure the tag name does exist in primefaces-p.taglib.xml
2. Create a new file in src/main/components/asciidoc/ folder (in PF repo), following this convention `_<tagName>.adoc`
3. Put whatever you want in it, don't worry about what has to be generated, the plugin will take care of that.  But it's important for you to know, that, the plugin will look for that file and inject the "content" in component.ftl to render the component output

If your curiosity is craving for it, have a look here to understand how it works exactly: https://github.com/Rapster/primefaces-asciidoc It's pretty simple you'll see 😉

So far, primefaces-asciidoc handles these cases, parsing:
- Component properties and attributes
- JSF Functions
- Component Client API (it's been a hell of a work, but it's done now 😉)

Today, there are three major tasks to complete the documentation:

- Migrating sections into asciidoc (I did a few already)
- Importing new pictures with the brand new theme
- Thomas made me realize recently, there is no default value tags in taglib.xml. I've been using defaultValue tag introduced by PrimeFaces, but it's not recognized as a valid tag. To workaround this, in order to parse the default value, a convention has to be set: "Default value is: <value>". It means, rework all the descriptions of all attributes of each components. 
- Doc client API for every components: use of @public and follow jsDoc style (see few examples in dataTable.js, ajaxStatus.js)

My thought on another topic: the CSS classes. I don't see the added value of indicating which CSS classes are used for a component (plus it's incomplete). Personnally, I go on the showcase or my company app what's there and end of the story. My suggestion is to delete these, as it's not really useful and you'll save some time for the migration process (believe me you will)

In case you feel like there are things we could do better, feel free to suggest, but so far, I think we covered most of what we need 😉

Quick sample here: [index.txt](https://github.com/primefaces/primefaces/files/3018242/index.txt) _(rename into html)_


cnsgithub (changes requested):
Looks good for a first draft. How did you convert it?

The tables are messed up.

image

cnsgithub (changes requested):
Missing code formatting/indentation:

image

mertsincan:
Maybe, http://pdf2md.morethan.io/ :D I used it :D Looks good @Rapster ;)

cnsgithub (commented):
How can we derive these component and attribute information automatically from sources?

image

Suggestion:
1. Use [Freemarker](https://github.com/apache/freemarker) templates, iterate over components/tags/attributes/renderers/etc. defined in `primefaces-p.taglib.xml`, `faces-config.xml `or somewhere else and dynamically create snippets.
```freemarker
<#list components as component>
   ...
   ${component.name} ${component.description}
   <#list component.attributes as attribute>
      ...
   </#list>
</#list>
```
2. Include those snippets in `index.adoc` during doc generation.
````adoc
include::{snippets}/comp1.adoc[]
````
3. Use `asciidoctor-maven-plugin` to generate HTML.
4. Publish generated HTML to GitHub Pages.

**Comment by Rapster** [comment]
Thanks @cnsgithub I simply did a quick conversion pdf to markdown, and clean up the table of content so far. I'll figure out something about the generation of component properties, thanks for sharing your thoughts. Once done, I'll start cleaning up the whole adoc (indentation, table, etc.)

**Comment by christophs78** [comment]
Realy happy to hear someone tries to modernize the Documentation!

**Comment by melloware** [comment]
@Rapster also it looks like everything is in 1 giant ADOC file?  Should we have an ADOC file per component to make future updates and commits easier?  I know this your first cut so I apologize if that was in your plans.

**Comment by Rapster** [comment]
Yup, that's the plan. Don't worry, I'm glad you're thinking the same :wink:

File: src/main/asciidoc/ftl/component.ftl [hunk 1]
```diff
@@ -0,0 +1,5 @@
+# ${tag.description}
```
cnsgithub: @Rapster I like your approach very much! 👍  Please consider escaping for asciidoc within ftl templates.

Rapster: I'll have a look, thanks for saying it :wink:

tandraschko:
Thanks for your great work so far!

I just wonder if we could move code from the new dependency:
`                     <dependency>
                         <groupId>org.primefaces</groupId>
                         <artifactId>primefaces-asciidoc</artifactId>
                         <version>1.0-SNAPSHOT</version>
                     </dependency>`
inside PF itself? 
I don't like to introduce new dependencies, it was always annoying with the maven plugin....

Rapster:
That was my first intent, I tried and didn't work and from the doc it says clearly:

> Note that processors must be included in the plugin’s execution classpath, not in the project’s.

But I kinda like it this way, as it gives is a clear distinction between PF core and PF asciidoc API. And don't worry, it's not a proper new dependency, just used during plugin execution.

Rapster:
@cagataycivici I recently add a look on your last commit. I'm not sure what has ben decided... But I think this PR might answer to #4332, you'll find a sample here https://github.com/primefaces/primefaces/issues/4332#issuecomment-451780963

tandraschko:
yep please check our discussion and share your thoughts @cagataycivici :D
i think the community including mert really agreed on a cool way with very much autogeneration.

Rapster:
I just feel the whole conversation we had about what's the best for documentation is completely being ignored. It's the shame really, I took a lot of my free time for this as I was quite convinced (and I still am) it was the right way to solve the documentation problem...

From what I can see right now:
- one folder per version: I don't think it's good idea as you'll congest the repo
- nothing generated: a lot of incoherences can happen. We (and the community) will have to handle the documentation tickets in case of adding/updating/deleting attributes.
- docsify

melloware:
I am with @Rapster this is the way to go for documentation.  Its what all the Spring Boot projects are doing.

Rapster:
I just feel the whole conversation we had about what's the best for documentation is completely being ignored. It's a shame really, I took a lot of my free time for this (after finding a common ground) as I was quite convinced (and I still am) it was the right way to solve the documentation problem...

From what I can see right now:
- one folder per version: This is something we talked about already, I don't think it's good idea as you'll congest the repo with time, specially that git does that for you
- nothing generated: a lot of incoherences can happen. We (and the community) will have to handle the documentation tickets in case of adding/updating/deleting attributes. I thought we all agreed that it's a pretty cool feature.
- docsify: sounds nice but very limited, handles only static content and the project is dying. AsciiDoctor handles custom directive. The final result on AsciiDoctor looks better IMO.
- markdown: Since the core is asciidoctor, using asciidoc makes sense (and it has more features than markdown). But in your case, docsify id markdown based.
- migration: you've been rewriting what I've already done so far... Kind of a waste of time on both sides.

But again, as said Thomas, sharing your thoughts would be a great help for the community to understand what led to that decision. This lack of communication is very frustrating...

tandraschko:
As i already stated, i can't change it and don't like the decision of PrimeTek and that you even don't get an statement.
However, i can't change it and will close the PR.

Thanks for your hard work.

Issue #3415, opened 2018-03-06
Theoretical EL injection via unencrypted client side state saving


cnsgithub:
## 1) Environment
- PrimeFaces version: 6.3-snapshot
- Does it work on the newest released PrimeFaces version? Version? Yes, 6.2
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) Yes
- Application server + version: all
- Affected browsers: all

## 2) Expected behavior
EL expressions should never be serialized and sent to the client if client side state saving is enabled, because they could be tampered with. This could lead to serious EL injections and remote code execution in rare cases, i.e. when encryption is disabled, older JSF implementations are used or something is misconfigured or turned on/off intentionally (e.g. state passivation, clustering). PrimeFaces should not rely on any server configurations or JSF implementations since there have been massive vulnerabilities: https://stackoverflow.com/questions/28231372/com-sun-faces-clientstatesavingpassword-recommendations-for-actual-password

## 3) Actual behavior
PrimeFaces serializes `MethodExpressions `and `ValueExpressions `which can be tampered with in client side state saving mode without encryption. See [CWE-502](https://cwe.mitre.org/data/definitions/502.html) also. 

The following component classes are potentially vulnerable since they implement `Serializable `and contain non-transient `ValueExpressions`:
- `TableState`
- `ResetInputActionListener`

`MethodExpressions `are probably of less interest for attackers since they are very limited, however I would not give a hasty all-clear, e.g. `#{guestPreferences.changeTheme}` in the showcase's `template.xhtml `can be found in the serialized view state.

You don't need to have a special gadget like `InvokerTransformer `in obsolete `commons-collections` implementations in your classpath as in older attacks.

## 4) Steps to reproduce
- Enable client side state saving in web.xml of showcase:
```
    <context-param>
        <param-name>javax.faces.STATE_SAVING_METHOD</param-name>
        <param-value>client</param-value>
    </context-param>
```
- Disable client state encryption in web.xml of showcase - note that setting the `context-param `to false will also **disable** encryption which in my opinion is very dangerous (tested at least with Mojarra 2.3.2)
```
    <context-param>
        <param-name>com.sun.faces.disableClientStateEncryption</param-name>
        <param-value>true</param-value>
    </context-param>
```
- Go to the table sort showcase: http://localhost:8080/showcase/ui/data/datatable/sort.xhtml
- Click at the first column header to sort by id
[...]

PR #4338, opened 2018-12-19
Fix #4335: TreeNode interface extend Serializable.

### LINKED ISSUE #4335: TreeTable: Make TreeNode interface serializable
## 1) Environment
- PrimeFaces version: 6.1.12
- Does it work on the newest released PrimeFaces version? Version? - No
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) - No
- Application server + version: Apache Tomcat 8.5
- Affected browsers:

## 2) Expected behavior
`TreeNode` interface should extend `Serializable` so we can use it as a member variable in our backing beans.

## 3) Actual behavior
`TreeNode` interface does not extend `Serializable` although both its implementation classes `CheckboxTreeNode` and `DefaultTreeNode` implement the `Serializable` interface.
This produces warnings like `Make "treeTableRoot" transient or serializable.` and gives problems during undeploy and redeploy phases of the web application.

## 4) Steps to reproduce
see showcase: https://www.primefaces.org/showcase/ui/data/treetable/basic.xhtml

## 5) Sample XHTML
see showcase: https://www.primefaces.org/showcase/ui/data/treetable/basic.xhtml

## 6) Sample bean
see showcase: https://www.primefaces.org/showcase/ui/data/treetable/basic.xhtml

melloware:
@tandraschko You can reject this if you want, but this does seem like a harmless change that I can definitely see being useful.

### DIFF
--- a/src/main/java/org/primefaces/model/CheckboxTreeNode.java
+++ b/src/main/java/org/primefaces/model/CheckboxTreeNode.java
@@ -15,10 +15,9 @@
  */
 package org.primefaces.model;
 
-import java.io.Serializable;
 import java.util.List;
 
-public class CheckboxTreeNode implements TreeNode, Serializable {
+public class CheckboxTreeNode implements TreeNode {
 
     public static final String DEFAULT_TYPE = "default";
 
--- a/src/main/java/org/primefaces/model/DefaultTreeNode.java
+++ b/src/main/java/org/primefaces/model/DefaultTreeNode.java
@@ -15,10 +15,9 @@
  */
 package org.primefaces.model;
 
-import java.io.Serializable;
 import java.util.List;
 
-public class DefaultTreeNode implements TreeNode, Serializable {
+public class DefaultTreeNode implements TreeNode {
 
     public static final String DEFAULT_TYPE = "default";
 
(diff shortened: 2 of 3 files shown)


cnsgithub (commented):
Since `DefaultTreeNode` and `CheckboxTreeNode` contain the field `Object data `which is not necessarily serializable, it was wrongly implemented before and still remains wrong. `Object data `must be changed to `Serializable data `to be conform with the `Serializable` interface anyway. However, that might be a breaking change. Please also note that it might bring in some vulnerabilities if custom `TreeNode` implementations contain `ValueExpression` fields for example, see related, theoretical issue https://github.com/primefaces/primefaces/issues/3415 also.

melloware:
Good point.  The problem is if these are not serializable you cannot use them in ViewScoped beans.  So I am not sure what is worse making the Object data field Serializable data, or to remove Serializable period and have all downstream consumers break and have to add Serializable every time they use DefaultTreeNode or worse have to change their code to always write their own custom TreeNode.

Probably worth a PrimeTek discussion with Optimus weighing in.

tandraschko:
As i said, the whole thing isn't 100% nice. This PR would just prevent a stupid IDE warning, nothing more.

Issue #1836, opened 2016-10-16
Dialog: modal should block page scroll


JaviRpo:
# IMPORTANT
- If you open a issue, fill every item. Otherwise the issue might be closed as invalid.
- Before you open a issue, test it with the current/newest version.
## 1) Environment
- PrimeFaces version: 6.0
- Does it work on the newest PrimeFaces version?
- Application server + version: Tomcat 8
- Affected browsers: All
## 2) Expected behavior

When i open a p:dialog with modal="true" attribute the page scroll should be blocked
...
## 3) Actual behavior

Actually, when i open a p:dialog with modal="true" attribute i can scroll the page in background causing some problems with p:sticky component
..
## 4) Steps to reproduce

I have a p:sticky component:
![image](https://cloud.githubusercontent.com/assets/8562436/19421520/cd61a280-93c8-11e6-8fbf-df4470fd8e90.png)

When i open a p:dialog component this appears above the p:sticky component:
![image](https://cloud.githubusercontent.com/assets/8562436/19421531/09388abc-93c9-11e6-9db7-b23998d64269.png)

The page scroll is still active, when I scroll the page to the top the p:sticky component desappears:
![image](https://cloud.githubusercontent.com/assets/8562436/19421541/55f1dcf0-93c9-11e6-9248-fdce9cb9fc69.png)

But, if I scroll down the page, the p:sticky component appears again, this time above the p:dialog:
![image](https://cloud.githubusercontent.com/assets/8562436/19421561/b81f6938-93c9-11e6-8bd2-858a9d551509.png)

One solution for this is block the page scroll.
Other solution should be, when the p:stick component appears again, this do the p:dialog below.

.. 
## 5) Sample XHTML

`<p:toolbar id="tb2">
    <f:facet name="left">
      <p:commandButton type="button" value="Open Dialog" icon="ui-icon-document" onclick="PF('dlg').show()" />
    </f:facet>
  </p:toolbar>
  <p:sticky target="tb2" />
  <div style="height: 1000px; border: 1px solid;">Any Content</div>
  <p:dialog id="dlg" widgetVar="dlg" dynamic="true" header="Title" modal="true" showEffect="fade" hideEffect="fade" resizable="false" closeOnEscape="true">
    <div style="height: 1000px; border: 1px solid;">Some Content</div>
  </p:dialog>`
..
## 6) Sample bean

Not necessary
..

tandraschko:
[...]

PR #4347, opened 2018-12-28
Fix #4140 / #1836: Spotlight and any modal dialog/overlays prevent scrolling.

### LINKED ISSUE #4140: Spotlight: Body is still scrollable when the spotlight is shown
## 1) Environment
- PrimeFaces version: 6.3.0
- Application server + version: Tomcat 8.5.15
- Affected browsers: ALL

## 2) Expected behavior
When the spotlight is visible, it should be already attached to the corresponding item.

## 3) Actual behavior

When the spotlight is visible, if the user scroll the page using mouse wheel then the position of the highlight area is fixed, but the content underneath is moved. It makes the spotlight mark the wrong area.

## 4) Steps to reproduce
- Go to current primefaces showcases: https://www.primefaces.org/showcase/ui/misc/spotlight.xhtml
- Do not enter anything in the form
- Click button SAVE

primefaces-spotlight

melloware:
Much better fix thanks to @tandraschko !!

Tested:
 - OverlayPanel modal and non-modal
 - Dialog modal and non-modal
 - SideBar blockScroll="true" and blockScroll="false"
 - Spotlight modal and non-modal

### DIFF
--- a/src/main/java/org/primefaces/component/dialog/DialogBase.java
+++ b/src/main/java/org/primefaces/component/dialog/DialogBase.java
@@ -37,6 +37,7 @@ abstract class DialogBase extends UIPanel implements Widget, RTLAware, ClientBeh
         draggable,
         resizable,
         modal,
+        blockScroll,
         visible,
         width,
         height,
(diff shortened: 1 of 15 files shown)


tandraschko:
Imo we should introduce the same attribute as on sidebar, so everything can be configurable!?

melloware:
OK no problem.  Should it be TRUE by default on dialog, spotlight and overlaypanel?   I feel like that is a good default where as Sidebar False is the correct default.

tandraschko:
Dont know, i would make it false as overlaypanel could be used without preventscroll and its probably the default

melloware:
OK made major improvements to the utility and added blockScroll attribute to Lightbox, Dialog, DialogFramework, OverlayPanel, and Spotlight.

tandraschko:
@melloware its used for modal overlays, right? So i think we could set the default to true. Also for sidebar.
I'm just not sure if this could break some layout types?

melloware:
Correct it it’s used on any modal calls. I made it false in all situations by default to not affect any current behavior.  But we could consider making in true in certain scenarios

Issue #4397, opened 2019-01-10
DataExporter: Support custom Export class


ravihariharan:
## 1) Environment
- PrimeFaces version: 6.2
- Does it work on the newest released PrimeFaces version? Version? NA
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) NA
- Application server + version: NA
- Affected browsers: NA

Raising this issue to discuss possible solution to support a Custom Export class for DataExporter.

My Problem : Primefaces defines a hard dependency on for e.g POI version. The issue I ran into is our application has Birt report 4.31 which works only with poi 3.9. 

The proposed PR solution could also help to support other file format if required.

PS : i have raised it for discussion and could do addition test if the enhancement is accepted.

PR #4398, opened 2019-01-10
DataExporter: Support custom Export class #4397

ravihariharan:
The proposed PR solution could also help to support other file format if required.

PS : i have raised it for discussion and could do addition test if the enhancement is accepted.

### DIFF
--- a/src/main/java/org/primefaces/component/export/DataExporter.java
+++ b/src/main/java/org/primefaces/component/export/DataExporter.java
@@ -56,7 +56,7 @@ public class DataExporter implements ActionListener, StateHolder {
 
     private MethodExpression onTableRender;
 
-    private ValueExpression customExporter;
+    private ValueExpression customExporterClass;
 
     public DataExporter() {
     }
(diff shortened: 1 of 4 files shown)


melloware:
I like it.

tandraschko:
why not pass a bean-instance to it instead of a className?

ravihariharan:
> why not pass a bean-instance to it instead of a className?

I used String for fully qualified name of the custom Exporter  to be used in place of DataExporter.  

@tandraschko  can you elaborate on you suggestion? sorry did not clearly understand how it can be  done with bean-instance.

tandraschko:
customExporter="#{myCustomExporterNamedBean}"

ravihariharan:
@tandraschko can you review the new fix?

tandraschko:
please also remove tryNewInstanceOfClass

ravihariharan:
@tandraschko  i have removed tryNewInstanceOfClass

tandraschko:
Please also rename customExporterClass, after that i can merge it

ravihariharan:
@tandraschko  in addition to the rename , i also added customExporterClass to restoreState and saveState.

tandraschko:
Why customExporterClass? It's just "customExporter". It's the instance, not a class.

ravihariharan:
> Why customExporterClass? It's just "customExporter". It's the instance, not a class.

Sorry.. my bad.. need to get some extra strong coffee :P

PR #4399, opened 2019-01-10
Fix #4396: ProgressBar XSS in labelTemplate.

### LINKED ISSUE #4396: progressBar: XSS in labelTemplate
## 1) Environment

## 2) Expected behavior
Proper escaping of `p:progressBar labelTemplate`.

## 3) Actual behavior
labelTemplate is not escaped in progressbar.js

## 4) Steps to reproduce
1. Modify showcase accordingly:
`            <p:progressBar widgetVar="pbAjax" ajax="true" value="#{progressBarView.progress}" labelTemplate="{value}%&lt;script&gt;alert('xss');&lt;/script&gt;" styleClass="animated" global="false">
`
2. Open the [showcase](http://localhost:8080/showcase/ui/misc/progressBar.xhtml).
3. Click at the _Start_ button below _Ajax ProgressBar_.
image

melloware:


cnsgithub (changes requested):
`PrimeFaces.escapeHTML` should be called after `replace` since you cannot know how escaping is implemented. It might break the placeholder replacement otherwise.

Anyway, using a safe jQuery API should be preferred over keeping the `html` call.

melloware:
Fixed.

PR #4401, opened 2019-01-10
Fix #4389: Slider XSS in displayTemplate.

### LINKED ISSUE #4389: slider: XSS in displayTemplate
## 1) Environment

## 2) Expected behavior
Proper client side escaping of `p:slider displayTemplate`.

## 3) Actual behavior
XSS vulnerabilities in `slider.js`.

## 4) Steps to reproduce
1. Change `slider.xhtml` of the showcase accordingly:
`                <p:slider for="txt2" display="output" style="width: 200px" displayTemplate="Set ratio to %{value}&lt;script&gt;alert('xss');&lt;/script&gt;" />
`
2. Open the [showcase](http://localhost:8080/showcase/ui/input/slider.xhtml)
3. Modify the slider value
image

melloware:


cnsgithub (changes requested):
`PrimeFaces.escapeHTML` should be called after `replace` since you cannot know how escaping is implemented. It might break the placeholder replacement otherwise.

Anyway, using a safe jQuery API should be preferred over keeping the `html` call.

melloware:
Fixed.

PR #4406, opened 2019-01-10
Fix #4386: SelectCheckBoxMenu XSS.

### LINKED ISSUE #4386: selectCheckboxMenu: XSS when using group label
## 1) Environment

## 2) Expected behavior
Proper escaping of group's label.

## 3) Actual behavior
`SelectItemGroup.label` is correctly written in `SelectCheckboxMenuRenderer` but used without escaping in [forms.selectcheckboxmenu.js](https://github.com/primefaces/primefaces/blob/f47f2f34ea4d65f902b31199be3f7d84b013d78c/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js#L173).

## 4) Steps to reproduce
1. Rename a group label in the [showcase's](http://localhost:8080/showcase/ui/input/checkboxMenu.xhtml) `CheckboxView` like so:
`        SelectItemGroup germanCars = new SelectItemGroup("<script>alert('German Cars');</script>");
`
2. Go to the [showcase](http://localhost:8080/showcase/ui/input/checkboxMenu.xhtml).

melloware:


cnsgithub (changes requested):
I would prefer to use `text()` here.

melloware:
Fixed.

PR #4427, opened 2019-01-14
fix #4421 - chips: lack of user input validation (max)

### LINKED ISSUE #4421: chips: lack of user input validation (max)
## 1) Environment

## 2) Expected behavior
When using `p:chips max` it should not be possible to enter more than `max` values.

## 3) Actual behavior
Client side code may be bypassed to submit more than `max `values. Server side validation is missing.

## 4) Steps to reproduce
1. Put sample XHTML code shown below in `chips.xhtml `of showcase.
2. Put sample bean shown below as new `ChipsView.java` to showcase.
3. Open modified [showcase](http://localhost:8080/showcase/ui/input/chips.xhtml).
4. Execute `$("select[id$=chips_hinput]").append('<option value="chips" selected="selected"/><option value="sausage" selected="selected"/><option value="beer" selected="selected"/>')`
5. Model is updated with three instead of two (expected) values.

## 5) Sample XHTML
```xml
        <h:form>
            <h:panelGrid columns="2">
                <p:outputLabel for="chips" value="Values: " />
            </h:panelGrid>

            <p:chips id="chips" value="#{chipsView.chips}" max="2" />
            <p:commandButton type="submit" value="Submit" />
        
        </h:form>
```

## 6) Sample bean
```java
package org.primefaces.showcase.view.chips;

import java.util.List;

@ManagedBean
public class ChipsView {
    
    private List<String> chips;

    public List<String> getChips() {
        return chips;
    }

    public void setChips(List<String> chips) {
        this.chips = chips;
    }
}
```

cnsgithub:


src/main/java/org/primefaces/component/chips/ChipsRenderer.java
```diff
@@ -54,6 +54,10 @@ public void decode(FacesContext context, UIComponent component) {
             submittedValues = LangUtils.concat(submittedValues, new String[]{inputValue});
         }
 
+        if (submittedValues.length > chips.getMax()) {
+            return;
+        }
```
melloware, line 59: Should we throw an exception instead?

cnsgithub: Good point, considered that as well, but decided to silently ignore. There are both (or even more) variants in already existing code. And we don't have any guidelines on that, do we?

melloware: Yeah I noticed it was not consistent either.  In this case its an unexpected scenario so I thought maybe a FacesException would be better.

tandraschko: I just remember that we switched from throw a exception to silent ignore it for another case. 
But can't remember, maybe we should work on a guideline.

cnsgithub: @tandraschko Should we change it to throw a `FacesException `instead of silently ignoring it?

tandraschko: I would leave it for now. But it think we should create a issue for 7.1 to align all validations.

melloware: Agreed. Thanks guys.

PR #4442, opened 2019-01-16
Fix #4422: Datatable right click should not select checkbox.

### LINKED ISSUE #4422: DataTable: rowSelectMode="none" ContextMenu is selecting row and checkbox
I have a dataTable with multiple selection and atrribute rowSelectMode="checkbox" and contextMenu. So I want to select rows only with checkbox. But in this combination, I get row selected using RMB together with contextMenu

1) Environment
- PrimeFaces version: **6.2.14**
- Application server + version: **WLS 12.1.3.0.0**
- Affected browsers: **All, ie. FirefoX, IE, Chrome**

2) Expected behavior

**Checkbox should stay unchecked**

3) Actual behavior

**Ceckbox is being checked.**

4) Steps to reproduce

**Create dataTable with rowSelectMode="checkbox" and selectionMode set to multiple and add contextMenu to it. Right click on a row.**

melloware:


tandraschko:
i think it's currently by design (but not sure) - so thats up to @mertsincan.

PR #4462, opened 2019-01-22
Fix #4461: Add ariaLabel to CommandLink and update taglib docs for ariaLabel

### LINKED ISSUE #4461: Add ariaLabel attribute to commandLink without using title
Add aria-label to commandLink without using title attribute as commandButton has.
It's an accesibility issue.

melloware:
Fixed the taglib docs in all the places where ariaLabel was used to add description.
Added ariaLabel to CommandLink.


pedrodelgado81:
Great! Thanks

PR #4475, opened 2019-01-28
Fix #4395: Autocomplete added escapeResults attribute.

### LINKED ISSUE #4395: Autocomplete: Ampersand in suggestions returns escaped markup
## 1) Environment
- PrimeFaces version: 6.2
- Does it work on the newest released PrimeFaces version? Version? Yes - 6.2
- Does it work on the newest sources in GitHub? Not tested
- Application server + version: JBoss EAP 7.* - tested on multiple versions
- Affected browsers: Chrome 71.0.3578.98

## 2) Expected behavior

The suggestions of the autocomplete component should not display nor highlight the escaped ampersand.

## 3) Actual behavior

The suggestions of the autocomplete component display an escaped ampersand, if the item matching the search query contains any character in the escaped ampersand (&amp;).
See Screenshot.
primefaces_github

## 4) Steps to reproduce

The search query from the autocomplete component must contain any character in the escaped ampersand (&amp;). e.g. 'M' in the sample provided.

## 5) Sample XHTML

```
    <h:head>
        <title>Web Client Test</title>
    </h:head>
    <h:body>
        <h:form>
            <p:outputLabel value="Simple:" for="acSimple" />
            <p:autoComplete id="acSimple" value="#{mainController.txt1}" completeMethod="#{mainController.completeText}" />
        </h:form>
    </h:body>
</html>
```

## 6) Sample bean

```
public class MainController implements Serializable {
    private List<String> list;
    private String txt1;

    @PostConstruct
    public void init() {
        list = new ArrayList<>();
        list.add("Metro & Co.");
        list.add("McKesson");
        list.add("Morgan Stanley");
        list.add("Merrill Lynch");
        list.add("Marathon Oil");
    }

    public List<String> completeText(final String query) {
        return list.stream()
                .filter(item -> StringUtils.containsIgnoreCase(item, query))
                .collect(Collectors.toList());
    }

    public String getTxt1() {
        return txt1;
    }

    public void setTxt1(final String txt1) {
        this.txt1 = txt1;
    }
}
```
[web-client-test.zip](https://github.com/primefaces/primefaces/files/2745664/web-client-test.zip)

Related issue:
https://github.com/primefaces/primefaces/issues/1385

melloware:
Fixed & in search results.
image

### DIFF
--- a/src/main/java/org/primefaces/component/autocomplete/AutoCompleteBase.java
+++ b/src/main/java/org/primefaces/component/autocomplete/AutoCompleteBase.java
@@ -71,7 +71,8 @@ abstract class AutoCompleteBase extends HtmlInputText implements Widget, InputHo
         moreText,
         unique,
         dynamic,
-        autoSelection
+        autoSelection,
+        escapeResults
     }
 
     public AutoCompleteBase() {
@@ -409,6 +410,14 @@ abstract class AutoCompleteBase extends HtmlInputText implements Widget, InputHo
 
     public void setAutoSelection(boolean autoSelection) {
         getStateHelper().put(PropertyKeys.autoSelection, autoSelection);
+    }
+
+    public boolean isEscapeResults() {
+        return (Boolean) getStateHelper().eval(PropertyKeys.escapeResults, true);
+    }
+
+    public void setEscapeResults(boolean escapeResults) {
+        getStateHelper().put(PropertyKeys.escapeResults, escapeResults);
     }
 
     @Override
--- a/src/main/java/org/primefaces/component/autocomplete/AutoCompleteBase.java
+++ b/src/main/java/org/primefaces/component/autocomplete/AutoCompleteBase.java
@@ -72,7 +72,7 @@ abstract class AutoCompleteBase extends HtmlInputText implements Widget, InputHo
         unique,
         dynamic,
         autoSelection,
-        escapeResults
+        escape
     }
 
     public AutoCompleteBase() {
@@ -412,12 +412,12 @@ abstract class AutoCompleteBase extends HtmlInputText implements Widget, InputHo
         getStateHelper().put(PropertyKeys.autoSelection, autoSelection);
     }
 
-    public boolean isEscapeResults() {
-        return (Boolean) getStateHelper().eval(PropertyKeys.escapeResults, true);
+    public boolean isEscape() {
+        return (Boolean) getStateHelper().eval(PropertyKeys.escape, true);
     }
 
-    public void setEscapeResults(boolean escapeResults) {
-        getStateHelper().put(PropertyKeys.escapeResults, escapeResults);
+    public void setEscape(boolean escape) {
+        getStateHelper().put(PropertyKeys.escape, escape);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


tandraschko:
Could this break something other?
Is it possible to use html in the item text? escape flags? and so on?

melloware:
I think we would have to introduce an "escapeSuggestions" property if we wanted to make sure this was fixed properly.  Want me to do that?

tandraschko:
i don't know.
I just fear to break other cases. I think e.g. adding a html "i" tag should be possible.

melloware:
OK pushed another change to add escapeResults property which by default is true. This leaves the code exactly the way it works today with the ability to turn it off if you desire.

tandraschko:
I case that escapeResults is false, we call both .text() and .html() - i would reduce it to only execute code thats really required.

cnsgithub (changes requested):
@melloware I just wonder if escape logic is inverted when comparing to component's `escape` attribute?

`escape="false"` => user wants to explicitly use HTML and bypass JSF escaping

melloware:
@tandraschko just submitted the one line fix to only use one or the other.

@cnsgithub I tested it all and its all working.   if escape = true what its doing is converting the suggestions to HTML which turns & into amp; and with escape = false it converts it to text() which leaves the & as an & and then when its replaces and put back into the HTML its rendered properly.  I have my test case in the ticket if you want to try it out?

cnsgithub:
@melloware I trust you. :laughing:

melloware:
HA! You don't trust anyone.  If you were American I would be convinced you were in the CIA. 😃

tandraschko:
what does the escape on autocompelte currently? isn't it built for the suggestions?

melloware:
There is in the Renderer an encodeSuggestionsAsList() method.

I think its doing this... `writer.writeText(ac.getItemLabel(), null);`

tandraschko:
so isn't escapeResults the same then? just not implemented on client side correctly?

melloware:
Good question.  Now you have me question how its not working.  I definitely know that & is turning to &amp; in the client side.

melloware:
OK thoroughly debugged this and wanting to know next steps.

1. User puts in SelectItem "Metro & Co".

2. Renderer using `writer.writeText()` of "Metro & Co"

3. Client gets it and it does `item.html()` which converts it to "Metro amp; Co"

4. The code then replaces "M" with highlighting `highlighedText = text.replace(re, '<span class="ui-autocomplete-query">$&</span>')`

5. The highlighted text is output to HTML with `item.html(highlighedText)`; is output with "amp;"

With my fix now  step#2 is replaced if you don't want escaping with item.text() which returns the exact inner text of the element which is "Metro & Co".

tandraschko:
I still don't understand why we should have escape and escapeResults

melloware:
So unless I am missing something obvious I don't see any escape property on AutoComplete?   So are you just suggesting i renamed escapeResults to just escape?

Rapster:
I think Thomas thought AutoComplete had an escape property while it doesn't. I actually thought the same when you introduced this new property escapeResults. IMO, renaming into `escape` sounds better

melloware:
I am definitely ok with switching it to just escape.

tandraschko:
i still wonder about:
`                    var text = $this.cfg.escape ? item.html() : item.text();`

for me it sounds wrong. escape should actually not use html?!

cnsgithub:
> i still wonder about:
> ` var text = $this.cfg.escape ? item.html() : item.text();`
> 
> for me it sounds wrong. escape should actually not use html?!

That confused me, too.

melloware:
Yep look at my steps above I document exactly what is going on.

tandraschko:
maybe we should invest more time here after the release.
Just merging it now feels wrong.

melloware:
Agreed.  Let's take a closer look after the release.

tandraschko:
Thanks!

PR #4520, opened 2019-02-08
Fix #4505: Correct both up and down propagation on backing bean

### LINKED ISSUE #4505: Tree with propagateSelectionDown to false still propagates selection on update
## 1) Environment
- PrimeFaces version: 6.2.0
- Does it work on the newest released PrimeFaces version? Version? Not working in 7.0.RC2
- Does it work on the newest sources in GitHub? Not working
- Application server + version: Wildfly 10.1.0.Final
- Affected browsers: All

## 2) Expected behavior

When setting attribute 'propagateSelectionDown' to 'false' in checkbox selection, I expect there is no selection propagation to child nodes.

## 3) Actual behavior

If I select a node with some childs, child nodes are selected or unselected in backing model. 
This behavior can be seen on updating the tree component.

## 4) Steps to reproduce

First, check a node with some childs like this : 
image

After this, click on ajax button which updates the tree, it will results on this : 
image

## 5) Sample XHTML

```xhtml
<h:form>
	<p:tree value="#{testTreeBean.root}"
		var="data"
		selectionMode="checkbox"
		selection="#{testTreeBean.selectedNodes}"
		propagateSelectionDown="false"
		propagateSelectionUp="false">
				
		<p:treeNode>
			<h:outputText value="#{data}"/>
		</p:treeNode>
	</p:tree>
			
	<p:commandButton value="test" update="@form" />
</h:form>
```

## 6) Sample bean

```java
@Named
@ViewScoped
public class TestTreeBean implements Serializable {	
	private TreeNode root;
	private TreeNode[] selectedNodes;
	
	@PostConstruct
	public void init() {
		root = new CheckboxTreeNode();
		
		// First Level
		TreeNode fruits = new CheckboxTreeNode("Fruits", root);
		TreeNode vegetables = new CheckboxTreeNode("Vegetables", root);
		
		// Fruits Childs
		TreeNode apple = new CheckboxTreeNode("Apple", fruits);
		TreeNode strawberry = new CheckboxTreeNode("Strawberry", fruits);
		
		// Apple Childs
		TreeNode greenApple = new CheckboxTreeNode("Green Apple", apple);
		TreeNode yellowApple = new CheckboxTreeNode("Yellow Apple", apple);
	}

	public TreeNode getRoot() {
		return root;
	}

	public void setRoot(TreeNode root) {
		this.root = root;
	}

	public TreeNode[] getSelectedNodes() {
		return selectedNodes;
	}

	public void setSelectedNodes(TreeNode[] selectedNodes) {
		this.selectedNodes = selectedNodes;
	}
}
```

brivalin:
class UITree: Get values of propagateSelectionDown and propagateSelectionUp. If selectionMode is 'checkbox', call setSelected method of CheckboxTreeNode with propagateSelectionDown and propagateSelectionUp.

class CheckboxTreeNode: Add setSelected method with the two parameters propagateSelectionDown and propagateSelectionUp to process independently up and down.

function initSelection: Add constraint propagateUp to not propagate minus icons to parents if propagateUp is false

### DIFF
--- a/src/main/java/org/primefaces/component/api/UITree.java
+++ b/src/main/java/org/primefaces/component/api/UITree.java
@@ -200,7 +200,7 @@ public abstract class UITree extends UIComponentBase implements NamingContainer 
     }
 
     public boolean isPropagateSelectionDown() {
-        return (Boolean) getStateHelper().eval(PropertyKeys.propagateSelectionDown, false);
+        return (Boolean) getStateHelper().eval(PropertyKeys.propagateSelectionDown, true);
     }
 
     public void setPropagateSelectionDown(boolean _propagateSelectionDown) {
@@ -208,7 +208,7 @@ public abstract class UITree extends UIComponentBase implements NamingContainer 
     }
 
     public boolean isPropagateSelectionUp() {
-        return (Boolean) getStateHelper().eval(PropertyKeys.propagateSelectionUp, false);
+        return (Boolean) getStateHelper().eval(PropertyKeys.propagateSelectionUp, true);
     }
 
     public void setPropagateSelectionUp(boolean _propagateSelectionUp) {
--- a/src/main/java/org/primefaces/component/api/UITree.java
+++ b/src/main/java/org/primefaces/component/api/UITree.java
@@ -32,6 +32,7 @@ import javax.faces.event.*;
 
 import org.primefaces.component.columns.Columns;
 import org.primefaces.component.tree.UITreeNode;
+import org.primefaces.model.CheckboxTreeNode;
 import org.primefaces.model.TreeNode;
 import org.primefaces.util.MessageFactory;
 import org.primefaces.util.SharedStringBuilder;
@@ -64,7 +65,9 @@ public abstract class UITree extends UIComponentBase implements NamingContainer 
         requiredMessage,
         skipChildren,
         showUnselectableCheckbox,
-        nodeVar;
+        nodeVar,
+        propagateSelectionDown,
+        propagateSelectionUp;
     }
 
     public String getRowKey() {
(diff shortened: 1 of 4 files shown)


melloware:
So I think from this PR you are saying that CheckBoxTreeNode never supported the propogate up or down that it was missing functionality?

brivalin:
Yes, a part of code was missing but I think it looks like more as a bug because propagation methods was already implemented in CheckboxTreeNode but not called properly.

melloware (approved):

src/main/java/org/primefaces/component/api/UITree.java
```diff
@@ -64,7 +65,9 @@
         requiredMessage,
         skipChildren,
         showUnselectableCheckbox,
-        nodeVar;
+        nodeVar,
+        propagateSelectionDown,
+        propagateSelectionUp;
```
melloware, line 70: If you put these propagate values in UITree shouldn't you remove them from TreeBase.java?

melloware (changes requested):

src/main/java/org/primefaces/component/api/UITree.java
brivalin: I have added these changes and I have correct another mistake I made by setting default propagate values to true.

melloware: Thanks for fixing.

trandbert37 (approved):

bvandaele:
Any chance for this to be merged before next release ?

brivalin:
Is there something else to do on this Pull Request to be merged ?

tandraschko:
Looks good for me. Thanks for the PR.

Issue #4516, opened 2019-02-07
DataTable: clearFilter does not work if only globalFilter ist used


OBreidenbach:
## 1) Environment
- PrimeFaces version: 6.2
- Does it work on the newest released PrimeFaces version? Version? No, 7.0RC2
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) No
- Application server + version: wildfly 11
- Affected browsers: FF, Chrome, ...

## 2) Expected behavior

Having a table with a globalFilter but without any column filter. I expect the global filter field to be cleared and the table to be reloaded without filter.

## 3) Actual behavior

Nothing happens because of the early exit in, 
https://github.com/primefaces/primefaces/blob/f57cfd531815afc8ac6b7edbc5d497bee5fd0ae2/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js#L2921-L2928

## 4) Steps to reproduce

Having a table with a globalFilter but without any column filter.

The early exit was introduced with: #1818
But it seams, as it not necessary any more. Locally I do not get any failure without it. Maybe that is because of #2965?

So my suggestion would be to remove the early exit. I can provide a PR if desired.
The same would apply to the clearFilter method of TreeTable: 
https://github.com/primefaces/primefaces/blob/f57cfd531815afc8ac6b7edbc5d497bee5fd0ae2/src/main/resources/META-INF/resources/primefaces/treetable/treetable.js#L118-L122

tandraschko:
@mertsincan WDYT?

mertsincan:
@OBreidenbach, Could you please attach a PR for this issue? Thanks a lot!

Issue #1818, opened 2016-10-08
calling ClearFilters on DataTable without filter - No error was thrown but something failed internally 


jxmai:
# IMPORTANT
- If you open a issue, fill every item. Otherwise the issue might be closed as invalid.
- Before you open a issue, test it with the current/newest version.
## 1) Environment
- PrimeFaces version: Primefaces v5.3
- Does it work on the newest PrimeFaces version? No, it failed on v6.0.x
- Application server + version: Tomcat v8.0
- Affected browsers: Firefox v49.0 (Browser version is irrelevant for this issue actually)
## 2) Expected behavior

Calling clearFilters on DataTable without filter **should do nothing as if nothing happened**. 
## 3) Actual behavior

No error was thrown, but the table was frozen. It will give developers a hard time to trace the bug ...
## 4) Steps to reproduce

Step 1: Click **Add a new row** footer button on the table, and a new record will be added

![capture](https://cloud.githubusercontent.com/assets/10467831/19209688/80d232b2-8cd5-11e6-8bec-e43a092b2aba.PNG)

Step 2: Click **Clear Filter** footer button on the table

Step 3: Repeat Step 1, and **add button stops working from UI perspective**. No error was thrown

**_on Browser Side (No error):**_

![capture](https://cloud.githubusercontent.com/assets/10467831/19209697/f8dcc786-8cd5-11e6-9b12-360ac6321185.PNG)

**_on Server Side (No error)**_:

![capture](https://cloud.githubusercontent.com/assets/10467831/19209701/2c9037ac-8cd6-11e6-995a-5c2f22176d56.PNG)
## 5) Sample XHTML

**table.xhtml**

``` html
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml"
    xmlns:ui="http://java.sun.com/jsf/facelets"
    xmlns:h="http://java.sun.com/jsf/html"
    xmlns:f="http://java.sun.com/jsf/core"
    xmlns:p="http://primefaces.org/ui"
    xmlns:fn="http://java.sun.com/jsp/jstl/functions">

<h:head>

</h:head>

<h:body>
    <h:form id="form">
        <p:growl id="msgs" showDetail="true" />

[...]

PR #4527, opened 2019-02-11
#4516, remove early exit in clearFilters in datatable.js and treetabl…

OBreidenbach:
…e.js

As sayed in #4516, the early exit was introduced with #1818. I noticed no failure without it, but maybe it is still necessary.


tandraschko:
@mertsincan WDYT?

Rapster:
I think this test  `if (columnFilters.length == 0) {` was there for a reason no?

OBreidenbach:
How can we progress?

tandraschko:
thats up to @mertsincan 
DataTable changes are up to PrimeTek, it's the most complex component.

PR #4531, opened 2019-02-12
Fix #4528: Autocomplete remove placeholder when items added.

### LINKED ISSUE #4528: p:autocomplete: multiple="true" does not hide placeholder on element added
## 1) Environment
- PrimeFaces version: 6.2
- JSF: 2.2.17
- Affected browsers: all

## 2) Expected behavior

p:autocomplete with multiple enabled, expecting placeholder to disappear when element is added.

## 3) Actual behavior

Placeholder remains visible regardless of number of elements added.

## 4) Steps to reproduce

Self explanatory from the description.

## 5) Sample XHTML

`								 <p:autoComplete id="autocomplete" multiple="true" 
								 				 placeholder="Employees" value="#{sendEmailController.model.employees}"
								 				 completeMethod="#{sendEmailController.onCompleteEmployee}"
												 var="employee" itemLabel="#{employee.email}"
												 itemValue="#{employee}">
								 </p:autoComplete>`

## 6) Sample bean

Complete method queries employee list from backing bean.
Employees is a list of entities. 

Problem showcased below: 

image

Edit: applies to p:chips component also.

melloware:


tandraschko:
what happen if the placeholder is null,empty or no placeholder attr defined? do we add a empty palceholder attr then in ln 764?

melloware:
Let me fix.

melloware:
OK just checked in a change where it will only put the placeholder back if there was one to begin with.

tandraschko:
Cool, could you also so it for Chips?

melloware:
Done I submitted that as a different PR so it would be lined up with its ticket.

PR #4544, opened 2019-02-14
Fix incomplete PR #2898

fdellwing:
#2898 implemented the ability to click the label in a `selectManyMenu` to select an additional option, but missed the deselection part. This PR fixxes that.


tandraschko:
@jxmai WDYT?

jxmai:
Thanks. I will check :+1: . @fdellwing, ~~why build failure for this PR~~ ? (most likely upstream issue at that time. Please merge the current upstream codebase to remove the error )

fdellwing:
Should be done, let's see if it fixes the travis build.

melloware (approved):

jxmai (approved):
I see. Good catch by the way.

PR #4548, opened 2019-02-17
Fix #2276: Added lastItemDisabled to BreadCrumb.

### LINKED ISSUE #2276: BreadCrumb: add option to disable the last item by default
Allowing a h:outputText in p:breadCrumb would be nice, or a current / last item text attribute on p:breadCrumb.

jepsar:
This seemed the most reasonable way to fix #2276. In most cases you just want to disable the last (current) item in the bread crumb, and this attribute will help you to do so without you having to determine what the last item is (when you are working with a dynamic list of elements).


tandraschko:
could you please update the docs and move "item.isDisabled()" as first statement in the IF?

jepsar:
@tandraschko done.

tandraschko:
could you please also not sort the table?
we didn't decide about the rules but IMO a must have is to have common attribute at the top (id, binding, widgetVar)

jepsar:
@tandraschko Okay, I checked with data table where the attributes were ordered. I've now added the new attribute at the end of the list.

tandraschko:
Thanks!

PR #4554, opened 2019-02-18
Fix #4553 : Updated the javascript to allow swipe left/right only

### LINKED ISSUE #4553: Carousel : Page Scroll not working (only mobile devices)
## 1) Environment
Primefaces official Demo and the issue is present in the GIT version also.

## 2) Expected behavior
The page should be scrollable

## 3) Actual behavior
The page is not scrollable

## 4) Steps to reproduce
1. Open the Carousel component in the Primefaces Official demo in any mobile device.
2. Try swiping up/down to scroll.
3. The page can be scrolled only when it is swiped on the area which is not covered by the carousel component (When the carousel is combined with a scroll-able datatable in each slide, this becomes tricky as the scroll in that datatable doesn't work).

Skipping the example code sections as this can be replicated in the Primefaces official demo itself.

ghost:


melloware (approved):

tandraschko:
Thanks for the PR!

PR #4558, opened 2019-02-19
 Fix #4263: OverlayPanel - Ensure top/left is not out of view

### LINKED ISSUE #4263: OverlayPanel: Top position out of view

fanste:
This fix uses the ``offset`` of the ``offsetParent`` of the to be positioned overlay as maximum allowed negative values for top and left. The ``offsetParent`` will respect e,g, ``position:fixed`` parents.

It is based on the current 7.0.RC3


fanste:
@tdtappe please run a test with this fix

fanste:
BTW: I did not find the option ``offset`` for jQuery UI's version of ``position``. Is this still supported / needed?`I did not see any difference when I removed it for testing.

tdtappe:
> @tdtappe please run a test with this fix

Works for me :-) Thanks a lot!!!

melloware (approved):

tandraschko:
Can you fix the conflicts?

fanste:
Should be good for now.
( A lot to learn about how to work with github :D )

tandraschko:
@fanste Could you check this case here? https://github.com/tdtappe/primefaces-test/tree/overlay-panel-position
The "bottom test" is always in background now
not sure if releated to your fix but would be great if we can fix both issues for the testcase.

fanste:
I had some time to take a look at this issue again. I don't think that we can fix that 'automatically'.

The button with the overlay is inside a div element (i call it parent). The modal's div element (i call it modal) is on the same level as the parent. Chancing the z-index on the overlay and modal has no effect as long as the parent does not have a higher z-index as the modal.

But automatically applying a z-index to each parent may destroy the look of the website. Therefore I suggest to either let it as it is and the user has to apply the z-index manually to each parent element that needs it or we introduce a new attribute for `p:overlayPanel` like `applyZindexToParent=true|false` with `false` as default. Additionally with an other new attribute `zindexParentSelector` (with * as default) the user could even decide which elements require an automatically applied z-index.

Another possibility may be to move the overlay out of the parent and parallel to modal. But i think that may require far more calculation and detection logic.

tandraschko:
@fanste can you rebase it?

fanste:
Looks like it should be correct again.

tandraschko:
Can someone test some cases?

tandraschko:
i tried to retest it by myself but something is wrong.
the second overlaypanel is covered by a modal.
can you check this?
it might be a different bug.

fanste:
What do you mean with "second overlaypanel"? The one that is opend with the button "bottom-test"? Then please have a look at your comment from the 24th april 2019 and my answer form the 27th may 2019. I already described the problem and two possibile solutions.

An alternative solution may be to move the model parallel to the overlay (inside the `position:fixed`-div). With the correct z-index for the button, overlay and modal it will be correct again. But that won't work when any parent has `position:relative`.

tandraschko:
oh, i see
need to check it in detail

src/main/resources/META-INF/resources/primefaces/overlaypanel/overlaypanel.js
```diff
@@ -199,7 +199,9 @@ PrimeFaces.widget.OverlayPanel = PrimeFaces.widget.DynamicOverlayWidget.extend({
     },
 
     align: function(target) {
-        var win = $(window);
+        var win = $(window),
+        allowedNegativeValuesByParentOffset = this.jq.offsetParent().offset(),
+        _self = this;
```
melloware: we typically use $this instead of _self.

melloware (changes requested):

melloware (approved):

melloware:
@tandraschko do you think this is safe to merge.  We are past 8.0 release and so far my testing for this has been good.

PR #4584, opened 2019-02-26
Fix #4583 JavaScript ScriptEngine not available via Thread.currentThread().getContextClassLoader() in certain environments

### LINKED ISSUE #4583: JavaScript ScriptEngine not available via Thread.currentThread().getContextClassLoader() in certain environments
## 1) Environment

In Liferay 7.0+ the JavaScript `ScriptEngine` is not available for the context `ClassLoader` which means that `new ScriptEngineManager().getEngineByName("javascript")` returns `null`.

## 2) Steps to reproduce:

1. Create a PrimeFaces Portlet Application for Liferay using the following `form` and `@ViewScoped` bean:

    ```
			<p:dataTable value="#{bean.uploadedFiles}" var="uploadedFile">
				<p:column>
					<f:facet name="header">
						<h:outputText value="File Name" />
					</f:facet>
					<h:outputText value="#{uploadedFile.fileName}" />
				</p:column>
				<p:column>
					<f:facet name="header">
						<h:outputText value="Size" />
					</f:facet>
					<h:outputText value="#{uploadedFile.size}" />
				</p:column>
				<p:column>
					<f:facet name="header">
						<h:outputText value="Content Type" />
					</f:facet>
					<h:outputText value="#{uploadedFile.contentType}" />
				</p:column>
			</p:dataTable>
			<h:messages for="fileUpload" />
			<p:fileUpload id="fileUpload" fileUploadListener="#{bean.handleFileUpload}"
				mode="advanced" update="@form"
				allowTypes="/(\.|\/)(txt)$/"/>
		</h:form>
    ```

    ```
	@ManagedBean
	@ViewScoped
	public class Bean {

	private List<UploadedFile> uploadedFiles;

	public List<UploadedFile> getUploadedFiles() {

		if (uploadedFiles == null) {
			uploadedFiles = new ArrayList<UploadedFile>();
		}

		return uploadedFiles;
	}

	public void handleFileUpload(FileUploadEvent fileUploadEvent) {

		List<UploadedFile> uploadedFiles = getUploadedFiles();
		UploadedFile uploadedFile = fileUploadEvent.getFile();
		uploadedFiles.add(uploadedFile);
	}

	}
    ```

2. Deploy and navigate to the portlet.
3. Upload a file with the **`.txt`** extension.

If the bug still exists the file will not be uploaded and no error messages will appear.

If the bug is fixed the file will be uploaded.

stiemannkj1:


tandraschko:
I would log a warning instead exception and skip filename validation when the engine is null. I dont trust that its available in all Environments

stiemannkj1:
@tandraschko the exception is caught and logged here: https://github.com/primefaces/primefaces/blob/c3d86342527a308e6220979459824dc4e69038ed/src/main/java/org/primefaces/util/FileUploadUtils.java#L153-L158

I think it's better to throw an exception so that devs see the failure immediately rather than hoping they see the issue in the logs.

stiemannkj1:
If you skip validation without throwing an exception, devs might presume that they are still protected when they actually aren't. It's a potential security vulnerability.

tandraschko:
Thats true but does every jvm implement it?

stiemannkj1:
No idea.

stiemannkj1:
But in the case that the JVM doesn't implement it, it'd still be better to throw an exception too.

melloware (approved):

PR #4586, opened 2019-02-26
Fix #4585 PrimeApplicationContext is not thread safe

### LINKED ISSUE #4585: PrimeApplicationContext is not thread safe
## 1) Environment

## 2) Explanation

The `PrimeApplicationContext` is an application scoped object that is lazily initialized. However, its initialization is not protected by any `synchronized` blocks or other thread safety code. So if X threads call `PrimeApplicationContext.getCurrentInstance(facesContext)` at the same time, they may instantiate as many as X `PrimeApplicationContext` instances.

Another aspect of `PrimeApplicationContext` that isn't thread safe is the lazy initialization of member data (such as `VirusScanner`). [This article](https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html) explains the issue in detail, but essentially the current implementation may allow threads to access partially initialized objects.

stiemannkj1:


stiemannkj1:
@tandraschko, sorry for the massive changes, but they were pretty much required to make things work in a thread-safe way. Let me know if you want me to explain anything. I thoroughly tested on Liferay, but changes like this probably warrant another RC or at least more testing from PrimeFaces folks.

tandraschko:
I dont like the API changes and such big changes are to riski before final.
Cant we just fix the double sync? Im still not sure if a real bug could occur in this lazy init.

 I dont Care about multiple short living applicationcontext instance, which can maybe only happen on first request if multiple users hit the application the same milliseconds...

stiemannkj1:
@tandraschko, okay yeah, I think I over-thought some of the changes. Lemme try and reduce some of them.

tandraschko:
Many of these changes looks Like refactoring, which is to much and to risky now.
I would like to see only a fix for the possible double sync problem, everything else ist 7.1. we already had 3 RC.

stiemannkj1:
@tandraschko, I just pushed an update to the branch.

I think this is the smallest I can make these changes. The problem is that many classes such as `PrimeExternalContext`, `PrimeConfiguration`, `PrimeEnvironment`. `VirusScannerService`, all `CacheProvider`s etc. need to be immutable and have all their inner data be `final` to prevent partial initialization as described [here](https://stackoverflow.com/questions/12611366/constructor-synchronization-in-java) and [here](https://wiki.sei.cmu.edu/confluence/display/java/TSM03-J.+Do+not+publish+partially+initialized+objects), All the aforementioned classes may be accessed by multiple threads, so they need to be thread-safe. I tried to remove all refactoring and reduce the changes as much as possible. Let me know if I should make any more changes.

tandraschko:
Can you provide an unittest to replicate the issue?
I would like to check it by myself on friday. Its still to many changes imo.

stiemannkj1:
@tandraschko, since partial initialization is by definition an intermittent issue, I cannot create a unit test for it unfortunately. All I can do is point to resources that explain why double-checked locking doesn't work: https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

stiemannkj1:
@tandraschko, if this change is too big for PrimeFaces 7.0, then I still **strongly** recommend that you make `PrimeEnvironment` and `PrimeConfiguration` effectively immutable by removing all setter methods from these classes. Then you can merge my changes immediately after the release so they can go into PF 7.0.1 and PF 7.1 without changing the API.

stiemannkj1:
@tandraschko, setter methods also need to be removed from `CacheProvider` implementations.

stiemannkj1:
@tandraschko mostly fixed this in https://github.com/primefaces/primefaces/issues/4585. However, I have a minor improvement that fixes partial object publication for `PrimeApplicationContext`: https://github.com/primefaces/primefaces/issues/4585

stiemannkj1:
@tandraschko mostly fixed this in https://github.com/primefaces/primefaces/issues/4585. However, I have a minor improvement that fixes partial object publication for `PrimeApplicationContext`: https://github.com/primefaces/primefaces/pull/4599

PR #4599, opened 2019-03-01
Fix #4585 PrimeApplicationContext is not thread safe

### LINKED ISSUE #4585: PrimeApplicationContext is not thread safe

stiemannkj1:


stiemannkj1:
@tandraschko, this should fix the issue where `PrimeApplicationContext` can be accessed while it is partially initialized. [When all data is `final` the object must be fully constructed before it can be referenced and the JVM cannot reorder the instructions](https://wiki.sei.cmu.edu/confluence/display/java/TSM03-J.+Do+not+publish+partially+initialized+objects). I *think* it may also prevent access to partially initialized `PrimeEnvironment` and `PrimeConfiguration` as well.

Also `ServiceLoader` is explicitly not thread safe so it shouldn't be used by multiple threads.

PR #4614, opened 2019-03-08
Added basic support of LazyDataModel to DataView

ngwaambe:
Add basic support for LazyDataModel to Dataview


tandraschko:
+1 basically
@ngwaambe Could you please provide also a PR in the showcase for a example? And also please update the docs (see git root docs/7.1)

PR #4620, opened 2019-03-11
Fix #4619: CSS font-display.

### LINKED ISSUE #4619: Controlling Font Performance with font-display
Based on findings with Google Lighthouse I think we need to add the CSS property "font-display" to the CSS.

You can read what Google says about this property here so I am not quite sure the correct setting should be "swap" or "fallback" but I figured you guys would do some internal testing.

```css
@font-face {
    font-family: 'PrimeIcons';
    font-display: swap;
    src: url('fonts/primeicons.eot');
    src: url('fonts/primeicons.eot?#iefix') format('embedded-opentype'), url('fonts/primeicons.ttf') format('truetype'), url('fonts/primeicons.woff') format('woff'), url('fonts/primeicons.svg?#primeicons') format('svg');
    font-weight: normal;
    font-style: normal;
}
```

It looks line FontAwesome chose "font-display: auto;".

https://github.com/FortAwesome/Font-Awesome/blob/master/css/all.css#L4262

melloware:


tandraschko:
I dont like to modify the original fa files

melloware:
Its a pretty harmless change and FA has made this change themselves in their 5.x branch.  They have stopped patching the 4.7 branch we are on.  I think also Primetek plans on eventually removing FontAwesome from PF and having users just include it on their own.  I could be wrong though...

mertsincan:
Thanks @melloware ;)

PR #4638, opened 2019-03-17
Fix #4637: ChartJS Radar/Polar must only have singular scale.

### LINKED ISSUE #4637: ChartsJS: Polar and Radar have incorrect scale
From this forum post: https://forum.primefaces.org/posting.php?mode=reply&f=3&t=58338

Looking at Polar and Radar charts the JS should not be "scales" it should be singular "scale".
https://www.chartjs.org/docs/latest/charts/radar.html

The radar chart supports only a single scale. The options for this scale are defined in the scale property.
```javascript
options = {
    scale: {
        // Hides the scale
        display: false
    }
};
```

Attached is runnable sample: 
[pf-4637.zip](https://github.com/primefaces/primefaces/files/2975686/pf-4637.zip)

melloware:


melloware:
@mertsincan @tandraschko I think this should be merged before 7.0 release as this definitely decreases the usability of the Polar and Radar charts.

mertsincan:
Thanks a lot @melloware ;)

PR #4640, opened 2019-03-18
fix #4636 - support `ShowAll` in rowsPerPageTemplate validation

### LINKED ISSUE #4636: DataTable: "IllegalArgumentException: Unsupported rows per page value" with ShowAll in rowsPerPageTemplate
## 1) Environment
- PrimeFaces 7.0.RC3
- WildFly Full 16.0.0.Final
- Affected browsers: Firefox 65.0.2

## 2) Expected behavior

When I select "All" in DataTable's rows per page drop-down list, all available rows should be shown.

## 3) Actual behavior

When the "All" option is selected, empty table is shown and there is an _java.lang.IllegalArgumentException: Unsupported rows per page value: 15_ error in server's log.

## 4) Steps to reproduce

Load the page, select the "All" option in the DataTable's rows per pages drop-down list.

## 5) Sample XHTML

    <h:form>
        <p:dataTable value="#{dataTableController.rows}" var="row" paginator="true"
                     rowsPerPageTemplate="5,10,{ShowAll|'All'}" rows="5">
            <p:column>
                <h:outputText value="#{row}"/>
            </p:column>
        </p:dataTable>
    </h:form>

## 6) Sample bean

    @Named
    @ApplicationScoped
    public class DataTableController implements Serializable {
        private static final int ROW_COUNT = 15;
        private static final List<String> ROWS = new ArrayList<>(ROW_COUNT);

        static {
            for (int i = 0; i < ROW_COUNT; i++) {
                ROWS.add("Row " + i);
            }
        }

            return ROWS;
        }
    }

cnsgithub:


tandraschko:
Thank you! Good PR.
It could just happen that the rowCount could change betweens requests, therefore it would fail.
A more safer fix would be that the Client Side posts "all" instead of a rowCount.
But not sure if can postpone to 7.1?

cnsgithub:
> It could just happen that the rowCount could change betweens requests

Hmm, you're right, we should have taken this into account. You mean when adding rows like in http://localhost:8080/showcase/ui/data/datatable/addRow.xhtml?

> But not sure if can postpone to 7.1?

@mertsincan already commited his own fix very similar to this PR: https://github.com/primefaces/primefaces/commit/a3c2bfa8390558b4c7ed0aecf3aac947aac06efc. So I am closing this one...

tandraschko:
Yep exactly.

tandraschko:
Maybe best example are Databases. Another User might add a entry.

PR #4641, opened 2019-03-18
update to PF 7.0

cnsgithub:


mertsincan:
:D Thanks :)

cnsgithub:
@mertsincan That was too fast. :wink: The artifacts are not yet available at maven central...

melloware:
It's OK they will be in another hour or two. :)

PR #4648, opened 2019-03-19
Fix #4611: add columns support to columnGroup 

### LINKED ISSUE #4611: TreeTable: add columns support to columnGroup
In 2014 support for ``p:columns`` was introduced inside ``p:columnGroup`` for ``p:dataTable`` - see commit 8779900c956afbbd9abc7cfa5946f3e212ca93bb.

Same should be done for tree table. A PR is currently in our internal testing.

fanste:


tandraschko:
should we move the the content of the ColumnGroupHelperRendererForTreetable to ColumnGroupHelperRenderer and decide there if dataTable or treeTable?

fanste:
You mean something like that? (see last commit)

tandraschko:
yep
but why is encodeColumnFooter public?

fanste:
Because it is used by `ColumGroupHelperRenderer`. `encodeColumnHeader` and `encodeColumnFooter` have to be public, otherwise they cannot be used from another package. `TreeTableRenderer` uses the same aproach like `DataTableRenderer`. Both methods int the datatable renderer are public too.

tandraschko:
all right, thanks!

Issue #3778, opened 2018-06-12
[Documentation] Schedule: update attributes for new FulLCalender version


tandraschko:
See: https://github.com/primefaces/primefaces/pull/3739

tandraschko:
@OBreidenbach Could you please change the documentation? Its here now: https://github.com/primefaces/primefaces/tree/master/docs

melloware:
@OBreidenbach this was merged i think its safe to close?

OBreidenbach:
@melloware:  @tandraschko opend this issue. So I can not close it.

Issue #3740, opened 2018-05-30
[Documentation] Schedule: explain the usage of the extender method


OBreidenbach:
## 1) Environment
- PrimeFaces version: 6.2

I learned about the extender and how to use it by accident. I just found it mentioned in some github ticket.
So it would be nice, if it could be explained and given an example either in the pdf user-guide or/and in the showcase. 
What do you think?
I do know how to provide PRs for the showcase. 
But I do not know how to provide content to the user guide besides the contents of the schedule.xml file.

tandraschko:
showcase -> https://github.com/primefaces/showcase

tandraschko:
@OBreidenbach Could you please change the documentation? Its here now: https://github.com/primefaces/primefaces/tree/master/docs

melloware:
@OBreidenbach this was merged i think its safe to close?

PR #4653, opened 2019-03-19
#3778, #3740 update schedule documentation

OBreidenbach:
#3778, #3740 update schedule documentation


tandraschko:
Sorry, renamed the dir to 7_0

melloware (approved):

tandraschko:
How can you Approve when it conflicts? :p

melloware:
I just reviewed it for its contents of what he changed not for what was conflicting.

OBreidenbach:
I have a questions. In the documentation the attributes of the components are in the order when they were added. That makes it hard to find a specif attribute in the table, especially if the component does have a lot of them.
Wouldn't it be much nicer, if the attributes would be in a more predictable order? Maybe alphabetically, maybe some basic attributes (like: id, value, ...) at the top.

tandraschko:
Sure

mertsincan:
> Wouldn't it be much nicer, if the attributes would be in a more predictable order? Maybe alphabetically, > maybe some basic attributes (like: id, value, ...) at the top.

Yes, unfortunately we need to do this manually in all *.md files :( But, maybe we can use markdown table sort plugin -> https://marketplace.visualstudio.com/items?itemName=simonguo.vscode-markdown-table-sort

I'll try it.

mertsincan:
WDYT?

https://github.com/primefaces/primefaces/blob/master/docs/7_0/components/datatable.md

tandraschko:
Thats also weird 😅
Id Binding widgetvar are much more common

mertsincan:
@tandraschko, could you please change them :D

melloware:
Just my two cents and maybe we just fix pages as the community offers a new PR on a component that we make that part of the checklist rather than editing all 100 pages by hand right now...just edit them as we make changes.

**Common:**
ID
WidgetVar
Rendered
Binding

**Attributes:**
(alphabetical order)

**Callbacks:**
onStart
onComplete
onClick
etc...

OBreidenbach:
@melloware I like your suggestion

tandraschko:
+0.75
Common + Callbacks as name makes sense. But this are all Attribute, so "Attributes" is not a good label


WDYT mert?

melloware:
Yep I couldn't think of what to call it maybe "Custom" or "Component Specific"

tandraschko:
I also wonder if it makes sense to add facets + behaviors the same table?
At lest we should make it the same in all widgets (the order - but we can make is as sections of course):
- short description
- renderer class and this stuff
- attributes
- behaviors
- facets
- client side API

PR #4663, opened 2019-03-20
 Fixes #4662 - LinkButton: 'target' and 'title' attributtes not rendered

### LINKED ISSUE #4662: LinkButton: 'target' and 'title' attributtes not rendered
## 1) Environment
- PrimeFaces version: 7.0
- Application server + version: Jetty 9.4 from https://github.com/primefaces/primefaces-test
- Affected browsers: ALL

## 2) Expected behavior

'Target' and 'title' attributes should be rendered because there are in `primefaces-p.taglib.xml`.
```XHTML
<p:linkButton outcome="test2" target="_BLANK" title="Test Title" value="Link Button" />
```
Should be render as

```HTML
<span id="j_idt6" class="ui-linkbutton ui-button ui-widget ui-state-default ui-corner-all ui-button-text-only" role="button" title="Test Title">
     <a href="/primefaces-test/test2.xhtml" target="_BLANK"><span class="ui-button-text ui-c">Link Button</span></a>
</span>
```

## 3) Actual behavior

Doesn't render 'target' nor 'title' attributes.

Got render as

```HTML
<span id="j_idt6" class="ui-linkbutton ui-button ui-widget ui-state-default ui-corner-all ui-button-text-only" role="button">
     <a href="/primefaces-test/test2.xhtml"><span class="ui-button-text ui-c">Link Button</span></a>
</span>
```

## 4) Sample XHTML

```XHTML
<p:linkButton outcome="test2" target="_BLANK" title="Test Title" value="Link Button" />
```

jasonex7:
This pull request fixes #4662. LinkButton is an excellent component to fill the gap of bookmarkable button.


tandraschko:
whats the intention of this PR?
It should actually work because its in: LINK_ATTRS_WITHOUT_EVENTS, right?

jasonex7:
Yes, with [this line](https://github.com/primefaces/primefaces/pull/4663/commits/574e8ed7df3210f4ded0a2b13f71b887472a97d9#diff-4f77beff7ed9bc85d9b0dc137ef4a1f2R77) that include LINK_ATTRS_WITHOUT_EVENTS in `<a>` tag it should work. The 'title' attribute is excluded because it was already rendered in `<span>` tag.

tandraschko:
why should it be rendered on the span instead a tag?
and i can't see any changes for the "target" attribute.

jasonex7:
> why should it be rendered on the span instead a tag?

Because the main `<p:linkButton>` component is rendered on a `<span>` tag, the `<a>` tag is a descendant, so the title should be on main tag. 

> and i can't see any changes for the "target" attribute.

The target attribute is now included en `<a>` thanks to rendering LINK_ATTRS_WITHOUT_EVENTS in this PR. Additionally, others attributes related to links (for example `rel`, `rev`, etc) are also rendered in  `<a>` tag.

melloware (approved):
I have studied this and I see exactly what @jasonex7 is doing and why.  I think this is a good fix.

PR #4668, opened 2019-03-20
fix: broken example for fileUpload's allowTypes attribute

cnsgithub:
- the pipe as well as the escape character itself must be escaped within markdown tables


tandraschko:
Conflicts :/

cnsgithub:
Resolved the conflict.

PR #4669, opened 2019-03-21
Fix #4655: JQplot device ratio issues.

### LINKED ISSUE #4655: JQPlot: Device Pixel Ratio
This is my first issue report.
metergauge is not displayed as expected.
Even if I set the style="width:400px;height:250px" the output is not readable.
You can check what I am talking about in the demo page :
https://www.primefaces.org/showcase/ui/chart/metergauge.xhtml

melloware:
No harm in fixing this since we have patched other JQplot 1.0.9 issues.  I know its being deprecated but many users are still using this functionality. I documented the fix in jqplot.txt.


melloware:
@mertsincan This should probably go in 7.0.1 as well.

PR #4677, opened 2019-03-23
Fix #4676: OutcomeTargetRender refactor for HREF client window handling.

### LINKED ISSUE #4676: menuItem: disableClientWindow does not apply to url
When use  with "url" attribute 
When I use the <p:menuitem with the "url" attribute and the "disableClientWindow='true'" attribute, the window parameter is still rendered.

                        <p:menuitem
                            value="Users"
                            url="faces/web/users/list-users.xhtml"
                            disableClientWindow="true"
                            ajax="false"
                            />
Rendered:
    faces/web/users/list-users.xhtml?jfwid=SFnBQ.....

1) Environment

    PrimeFaces version: 7.0
    Application server + version: wildflay 15
    Affected browsers: firefox
   JSF 2.3

melloware:
Please review this PR.  I simply moved the try {} finally {} handling of the client window code around the whole method so the URL's are evaluated.

The one thing I was not sure of is in the previous code the code
```java
if (outcomeTarget.getFragment() != null) {
                    url += "#" + outcomeTarget.getFragment();
                }
```

Was done outside the finally{} clause and I wasn't sure if that was on purpose or not.

### DIFF
--- a/src/main/java/org/primefaces/renderkit/OutcomeTargetRenderer.java
+++ b/src/main/java/org/primefaces/renderkit/OutcomeTargetRenderer.java
@@ -154,54 +154,52 @@ public class OutcomeTargetRenderer extends CoreRenderer {
     protected String getTargetURL(FacesContext context, UIOutcomeTarget outcomeTarget) {
         String url;
 
-        String href = outcomeTarget.getHref();
-        if (href != null) {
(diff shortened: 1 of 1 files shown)


tandraschko:
we have to check if url must have clientWindowId or not. 
In generell i would say, that the clientWindowId is only required for internal links (outcome) and not for external (url).

PR #4688, opened 2019-03-25
Fix #4683: SelectCheckboxMenu generate ARIA label for="" correctly.

### LINKED ISSUE #4683: SelectCheckboxMenu: add accessibility to items checkbox and label
Hello,

For accessibility purpose (see WCAG2.1 rules), is it possible to add the "for" attribute to the selectCheckboxMenu's labels to reference the left checkbox ?

POC (left=PF 7.0 code, right=POC code) : 
https://i.postimg.cc/qBhpHdTv/primefaces-Select-Checkbox.png

With this change, a screenreader reads "Other - checkbox not checked"
With PF 7.0, it reads "checkbox not checked" without the label, which is not accessible.

NB: possibly linked to : https://github.com/primefaces/primefaces/issues/3019 but screenreaders still doesn't read correctly the component's labels.

melloware:
Found an efficient UUID generator on stack overflow to generate unique ids in JS for these hidden input and labels.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/core/core.js
+++ b/src/main/resources/META-INF/resources/primefaces/core/core.js
@@ -589,6 +589,24 @@
             var ariaLocaleSettings = this.getLocaleSettings()['aria'];
             return (ariaLocaleSettings&&ariaLocaleSettings[key]) ? ariaLocaleSettings[key] : PrimeFaces.locales['en_US']['aria'][key];
         },
+        
+        /**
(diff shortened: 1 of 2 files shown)


tandraschko:
looks good, thanks
did you check the performance with 100+ items?

melloware:
I will test with 200 items and let you know.  I did pick this guys UUID algorithm because he performance tuned it vs others.  You can see his results if you scroll down to 

https://stackoverflow.com/a/21963136/502366

tandraschko:
Ok, should be fine then
thanks

Issue #4702, opened 2019-03-28
pickList:  Call to String.prototype.startsWith ES6 JS which causes errors on IE11


koan00:
## 1) Environment
- PrimeFaces version: 7.0
- Does it work on the newest released PrimeFaces version? No 
- Does it work on the newest sources in GitHub? No
- Application server + version: Tomcat 7
- Affected browsers: Internet Explorer 11 (IE11)

## 2) Expected behavior

No errors in Javascript console

## 3) Actual behavior

Many errors in javascript console: 
SCRIPT438: Object doesn't support property or method 'startsWith' 
components.js.xhtml, Line: 59, Column: 5360

## 4) Steps to reproduce

- Using IE 11, navigate to PF showcase: https://www.primefaces.org/showcase/ui/data/pickList.xhtml
- Click on a single item in left list
- Hold Shift, select second item on left list

PR #4703, opened 2019-03-28
Use indexOf instead of ES6 method startsWith to restore IE11 compatibility

koan00:
Fix for #4702


melloware (approved):

melloware:
@mertsincan this should probably go in 7.0.1 Elite release.

mertsincan:
Fixed now ;)

PR #4717, opened 2019-04-02
Fix #4712: SelectManyRenderer better defaults for empty array/list.

### LINKED ISSUE #4712: SelectCheckboxMenu: SelectItem with null value checked by default
## 1) Environment
- PrimeFaces version: 6.2.0
- Does it work on the newest released PrimeFaces version? Version? Not working in 7.0
- Does it work on the newest sources in GitHub? Not working
- Application server + version: Wildfly 10.1.0.Final
- Affected browsers: All

## 2) Expected behavior

Items with null values don't have to be checked by default.

## 3) Actual behavior

Items with null values are checked.

## 4) Steps to reproduce

No Steps to reproduce, only create SelectCheckboxMenu with code samples and you will get this : 

selectCheckboxMenu

## 5) Sample XHTML
```xhtml
	<h:head></h:head> 
	<body>
		<h:form>
			<p:selectCheckboxMenu value="#{testPrimeBean.values}" label="Test" updateLabel="true">
				<f:selectItems value="#{testPrimeBean.items}" />
			</p:selectCheckboxMenu>
		</h:form>
	</body> 
</html>
```
## 6) Sample bean
```java
@Named
@ViewScoped
public class TestPrimeBean implements Serializable{
	
	private List<SelectItem> items;
	private String[] values;

	@PostConstruct
	public void init() {
		items = new ArrayList<>();
		values = new String[0];
		
		items.add(new SelectItem(null, "One"));
		items.add(new SelectItem("two","Two"));
		items.add(new SelectItem("three"));
		items.add(new SelectItem(null));
	}

	public List<SelectItem> getItems() {
		return items;
	}

	public void setItems(List<SelectItem> items) {
		this.items = items;
	}

	public String[] getValues() {
		return values;
	}

	public void setValues(String[] values) {
		this.values = values;
	}
}
```
..

melloware:
This was the cleanest way to fix it by overriding the method in SelectRenderer and just tweaking this check.  

The base method is used by SelectOne and some other non many selectors so that I believe is the correct behavior if both are null.

### DIFF
--- a/src/main/java/org/primefaces/renderkit/SelectManyRenderer.java
+++ b/src/main/java/org/primefaces/renderkit/SelectManyRenderer.java
@@ -82,7 +82,7 @@ public abstract class SelectManyRenderer extends SelectRenderer {
     @Override
     protected boolean isSelected(FacesContext context, UIComponent component, Object itemValue, Object valueArray, Converter converter) {
         // GitHub #4712
-        if (itemValue == null && valueArray == null) {
+        if (valueArray == null) {
             return false;
         }
         return super.isSelected(context, component, itemValue, valueArray, converter);
--- a/src/main/java/org/primefaces/renderkit/SelectManyRenderer.java
+++ b/src/main/java/org/primefaces/renderkit/SelectManyRenderer.java
@@ -30,6 +30,7 @@ import javax.faces.FacesException;
 import javax.faces.component.UIComponent;
 import javax.faces.component.UISelectMany;
 import javax.faces.context.FacesContext;
+import javax.faces.convert.Converter;
 
 public abstract class SelectManyRenderer extends SelectRenderer {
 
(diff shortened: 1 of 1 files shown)


tandraschko:
Why not just "valueArray == null"?
Really not sure, maybe we should collect some cases first before applying it.

melloware:
Good question.  Was just following the pattern from the original method. I think that would be pretty safe to do.

tandraschko:
Still really unsure about this PR.
valueArray are the submittedValues and not the selected value expresison (value=...)

melloware:
well i have my test case in the ticket that shows the issue or lets you tweak different combinations of the issue.   I don't mind leaving it as is but to me String[] and List<String> should be equivalent and right now they definitely are not.

tandraschko:
let me also check it next week :)

tandraschko:
See the comment in the issue

PR #4728, opened 2019-04-09
Fix #4727: Dialog only use aria-labelledby is showHeader=true.

### LINKED ISSUE #4727: Dialog: improve accessibility when showHeader=false
## 1) Environment
- PrimeFaces version: 7.0
- Affected browsers: all

## 2) Expected behavior

The attribute `aria-labelledby` of p:dialog is hidden when there is no header. (`showHeader="false"`)

## 3) Actual behavior

An `aria-labelledby` is present and it references a not existing ID, which is forbidden. It should be generated only when showHeader is true.

## 4) Steps to reproduce

Add a p:dialog with `showHeader="false"` into a page.

## 5) Sample XHTML

```
    <h:head />
    <h:body>
        <h:form>
        	<p:dialog widgetVar="dia" showHeader="false">Hello World !</p:dialog>
        	<p:commandButton onclick="PF('dia').show()" />
        </h:form>
    </h:body>
</html>
```

## 6) Sample bean

None is necessary.

melloware:
I simply check for the existence of the title element and only set aria-labelledBy if the title is found.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/dialog/dialog.js
+++ b/src/main/resources/META-INF/resources/primefaces/dialog/dialog.js
@@ -561,8 +561,8 @@ PrimeFaces.widget.Dialog = PrimeFaces.widget.DynamicOverlayWidget.extend({
         });
         
         // GitHub #4727
-        var title = '#' + this.id + '_title';
-        if ($(title).length) {
+        var title = this.id + '_title';
+        if ($('#' +title).length) {
             this.jq.attr('aria-labelledby', title);
         }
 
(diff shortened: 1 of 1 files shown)


tandraschko:
Don't you miss a "#" in the selector?

melloware:
I sure did. Committed.

tandraschko:
Still not sure if this works as this.id can contain ":", which must be escaped.
We use PrimeFaces.escapeCLientID for this

melloware:
You are right updated to PrimeFaces.escapeClientId.

PR #4735, opened 2019-04-12
Fix #3139: Datatable refresh() adds multiple sort and page events.

### LINKED ISSUE #3139: DataTable: Same event handlers registered multiple times after DataTable.refresh() is called
## 1) Environment
- PrimeFaces version: 6.1.2
- Application server + version: irrelevant
- Affected browsers: irrelevant (all)

## 2) Expected behavior
When JS function DataTable.refresh() is called the different JS event handlers like those bound in Paginator.bindPageLinkEvents() and DataTable.bindCheckboxEvents() are registered twice - the old one stay bound and new listeners which do the same logic are registered. This is harmless in some cases but harmful in others. 

## 3) Actual behavior

Example for harmful cases:
a) in case of a DataTable with a paginator: when clicking one of the pageLinks like "prev" or "next" the same AJAX request is processed multiple times because in bindPaginator() "this.paginator = new PrimeFaces.widget.Paginator(this.cfg.paginator);" registers new event handlers without deregistering the old ones.
b) in case of table with a checkAllToggler the click handler is registered multiple times in bindCheckboxEvents(). If it's an even number (e.g. after calling DataTable.refresh() exactly once) clicking the "togle all" check box seems to doesn't function because it's implicitly clicked twice.
c) in case of sortable columns the click handler on the column headers is registered multiple times in bindSortEvents(). This causes (similar to a)) multiple identical AJAX requests when clicking on a sort header.

There may be more harmful cases. I haven't walked through all bind...() functions.

## 4) Steps to reproduce

for example a)
- open www.primefaces.org/showcase/ui/data/datatable/paginator.xhtml
- open JS console
- type "PrimeFaces.widgets['widget_j_idt117_j_idt118'].refresh(PrimeFaces.widgets['widget_j_idt117_j_idt118'].cfg)"
- jump to "Network" tab of DevTools
- click "next page" link inside the paginator
- in "Network" tab you can see that two AJAX requests were submitted
- type "PrimeFaces.widgets['widget...'].refresh(PrimeFaces.widgets['widget...'].cfg)" again
- in "Network" tab you can see that three AJAX requests were submitted

## 5) Sample XHTML

see www.primefaces.org/showcase/ui/data/datatable/paginator.xhtml

## 6) Sample bean

no special bean necessary

melloware:


mertsincan:
Good PR, thanks a lot @melloware ;)

Issue #4226, opened 2018-11-08
Galleria: Improve the animation


wsaca:
## 1) Environment
- PrimeFaces version: 6.3-SNAPSHOT
- Does it work on the newest released PrimeFaces version? Not Version? 6.2
- Does it work on the newest sources in GitHub? Not
- Application server + version: WildFly 13
- Affected browsers: Any

## 2) Expected behavior
When we click the next button the previous animation should be controlled. 

## 3) Actual behavior
When we click the next button many times and very fast, the previous animation is not controlled correctly. Actually there are the next problems related:
- The caption panel is still animated after we stop pressing the next button.
- The caption content is changed before the panel is hidden, it should be changed when the panel is hidden and before show it again. Resolve the issue #155.
- We could see more than one thumbnail with opacity = 1.

## 4) Steps to reproduce
1. Go to: https://www.primefaces.org/showcase/ui/multimedia/galleria.xhtml
2. Click the arrow "Next" many times and very fast.
3. We can see the problems described above.

melloware:
@wsaca are you saying this fixed issue #155 because that issue is still open?

wsaca:
@melloware yes, the issue #155 should be closed because the animation problem with the caption text was resolved.

melloware:
Excellent thanks for the update!

PR #4766, opened 2019-04-29
Fix #155: Galleria improve animation sync with caption text

### LINKED ISSUE #155: Improve Galleria animation sync with caption text
A minor thing, but it's annoying when you spot it.
When you have a Galleria with a longer effectSpeed you can see that the caption text changes when the fade down starts.

```
        <p:galleria value="#{MyController.banners}" var="b"
                    panelWidth="800" panelHeight="450"
                    showCaption="true"
                    showFilmstrip="false"
                    autoPlay="true"
                    effect="slide" effectSpeed="1000" transitionInterval="10000">
```

Delaying the caption change by the effectSpeed is enough to fix this in galleria.js line 201 looks like this:

```
        //caption
        if(this.cfg.showCaption) {
            this.showCaption(newPanel);
        }
```

Please can it get change to do something like this?

```
        //caption
        if(this.cfg.showCaption) {
            var _self = this;
            setTimeout(function() {_self.showCaption(newPanel);},this.cfg.effectSpeed);
        }
```

melloware:


melloware:
waiting for @wsaca to comment on whether this was fixed by his PR for #4226

tandraschko:
@wsaca @melloware do we need to apply this or already fixed by 4226?

Issue #4239, opened 2018-11-13
TreeTable: support filterFunction


crocimi:
## 1) Environment
- PrimeFaces version: 6.2
- Does it work on the newest released PrimeFaces version? Version? No
- Does it work on the newest sources in GitHub? -
- Application server + version: Wildfly 10.0.0
- Affected browsers: all

## 2) Expected behavior
filterFunction can be used in treeTable (like dataTable)

## 3) Actual behavior
Funtion is not called

## 4) Steps to reproduce
1. Build a tree table with dynamic columns and a filter
2. write some text in the filter field
.. 

## 5) Sample XHTML
```
<h:form id="form">
    <h:panelGrid columns="3" style="margin:10px 0">
        <p:outputLabel for="template" value="Template:" style="font-weight:bold"/>
        <p:inputText id="template" value="#{ttColumnsView.columnTemplate}" size="50"/>
        <p:commandButton update="tt" action="#{ttColumnsView.createDynamicColumns}" value="Update" process="@parent" icon="pi pi-refresh" />
    </h:panelGrid>

    <p:treeTable id="tt" value="#{ttColumnsView.root}" var="document">
        <f:facet name="header">
            Document Viewer
        </f:facet>
        <p:columns value="#{ttColumnsView.columns}" var="column" headerText="#{column.header}" filterBy="#{document[column.property]}"  filterFunction="#{ttColumnsView.filterWithWildcard}">
            <h:outputText value="#{document[column.property]}" />
        </p:columns>
    </p:treeTable>
    </h:form>
..
```

## 6) Sample bean

```
import org.primefaces.model.TreeNode;

import javax.annotation.PostConstruct;
import javax.faces.bean.ManagedBean;
import javax.faces.bean.ManagedProperty;
import javax.faces.bean.ViewScoped;
import javax.faces.component.UIComponent;
import javax.faces.context.FacesContext;
import java.io.Serializable;
import java.util.ArrayList;
import java.util.Arrays;
import java.util.List;
import java.util.Locale;
@ManagedBean(name="ttColumnsView")
@ViewScoped
public class ColumnsView implements Serializable {

  private final static List<String> VALID_COLUMN_KEYS = Arrays.asList("name", "size", "type");

  private String columnTemplate = "name size type";

  private List<ColumnModel> columns;

  private TreeNode root;

  @ManagedProperty("#{documentService}")
  private DocumentService service;

  @PostConstruct
  public void init() {
    root = service.createDocuments();

[...]

PR #4775, opened 2019-05-02
Add support for filterFunction

crocimi:
Add support for custom filterFunction (like DataTable) . Fix Issue #4239


tandraschko:
Thanks!

Issue #453, opened 2015-06-18
Locale support for p:spinner


fcanevar:
The spinner component (http://www.primefaces.org/showcase/ui/input/spinner.xhtml) should support different locales to enable input like "1,1" or "1.000,582" instead of "1.1".

This problem was first discovered in 2010 (http://forum.primefaces.org/viewtopic.php?f=3&t=6398), then an issue was opened in Jan 2015 (https://code.google.com/p/primefaces/issues/detail?id=7852) but still no luck.

Can anyone fix this? 

Thanks

cagataycivici:
What options should be defined by locale?

fcanevar:
Mainly the decimal and the thousands separator, for example you could see how this one works: http://jqueryui.com/spinner/#decimal . If you use the German locale the decimal separator changes. (try input 1.000,99 and then press '+').
Thank you very much for your prompt reply

fante76:
Hi! I've the same problem in the web application I'm developing. I'll stay tuned for a good solution.

cagataycivici:
I think we can get it from PrimeFaces.locales maybe. We already have decimal seperator and grouping separator. Will keep this in our radar for 5.3.

fcanevar:
Thanks a lot, is there a roadmap for the 5.3 release?

fcanevar:
This is the patch that I've made based on this one: http://jsf.hatemalimam.com/DailyLab/spinner/main.xhtml , keep in mind that it's tailored on our needs and that I'm not a JavaScript developer.

String.prototype.replaceAll = function(search, replace) {
    if (replace === undefined)
        return this.toString();
    else
        return this.split(search).join(replace);
}

PrimeFaces.widget.Spinner.prototype.getDecimalSeparator = function() {
    return '#{bean.decimalSeparator}';
};

PrimeFaces.widget.Spinner.prototype.getGroupingSeparator = function() {
    return '#{bean.groupingSeparator}';
};

PrimeFaces.widget.Spinner.prototype.getRegex = function() {
    return /(\d+)(\d{3})/;
};

PrimeFaces.widget.Spinner.prototype.format = function() {
    if (this.value !== null) {
        var res = (this.value+'').split('.'),
            result = res[0];
        while(this.getRegex().test(result))
            result = result.replace(this.getRegex(), '$1'+this.getGroupingSeparator()+'$2');
        if(res.length==2)
            result += this.getDecimalSeparator()+res[1];
        this.input.val(result);
    }
};

[...]

PR #4780, opened 2019-05-03
Locale support for p:spinner

rrsln:
Locale support for p:spinner #453


cnsgithub (commented):
@rrsln Are there really use cases where you need to specify `decimalSeparator` and `thousandSeparator` in a way that differs from the locale's default?

src/main/resources/META-INF/primefaces-p.taglib.xml
```diff
@@ -23902,6 +23902,22 @@
             <name>decimalPlaces</name>
             <required>false</required>
             <type>java.lang.String</type>
+        </attribute>
+		<attribute>
+            <description>
+                <![CDATA[Decimal separator char. Default is taken from the view's locale.]]>
```
cnsgithub: `Character` or `String`? `DecimalFormatSymbols` uses `Character`, so this should be sufficient for us as well?

cnsgithub (changes requested):

docs/7_1/components/spinner.md
```diff
@@ -35,6 +35,8 @@ min | null | Double | Minimum boundary value
 max | null | Double | Maximum boundary value
 prefix | null | String | Prefix of the input
 suffix | null | String | Suffix of the input
+decimalSeparator | Default is taken from the view's locale | String | Decimal separator char.
+thousandSeparator | Default is taken from the view's locale | String | Thousand separator char.
```
cnsgithub, line 40: I would prefer `groupingSeparator` instead of `thousandSeparator`, see `DecimalFormatSymbols`.

cnsgithub (changes requested):

tandraschko:
i dont like the new style of:

`ComponentUtils.getValue(getStateHelper(), "thousandSeparator",
            () -> LocaleUtils.getThousandSeparator(getFacesContext()))`

we should leave it as it is and everywhere the same.

src/main/resources/META-INF/primefaces-p.taglib.xml
rrsln: Character is enough for me. However I think that string is more flexible (for example someone may need to use space + separator) and it is consistent with p:inputNumber.

docs/7_1/components/spinner.md
rrsln: we use "thousandSeparator" in p:inputNumber. Using "thousandSeparator" there and "groupingSeparator" here is confusing.

cnsgithub: Agreed, should be consistent. Didn't know that.

rrsln:
> 
> 
> @rrsln Are there really use cases where you need to specify `decimalSeparator` and `thousandSeparator` in a way that differs from the locale's default?

well, locale's default is sufficient for me. However it is easy to imagine usecase when there is a need to use something different (for example when you select locale from the combobox or would like to use "space + separator" rather than simply "separator".

cnsgithub:
> it is easy to imagine usecase

Sorry, cannot imagine. :alien:  I personally would not support those weird, non-standard formattings. Anyways, it should be consistent with the rest of PrimeFaces.

rrsln:
> 
> 
> i dont like the new style of:
> 
> `ComponentUtils.getValue(getStateHelper(), "thousandSeparator", () -> LocaleUtils.getThousandSeparator(getFacesContext()))`
> 
> we should leave it as it is and everywhere the same.

It is possible to use:
`new Supplier() {
   bla bla bla
}`
instead but in my opinion it is more readable with lambda and functional interfaces have been mostly introduced for lambdas. BTW we already use lambdas in PrimeApplicationContext

tandraschko:
It's not about the lambda, why not just use:

return (String) getStateHelper().eval(PropertyKeys.thousandSeparator, LocaleUtils.getThousandSeparator(getFacesContext()));

because LocaleUtils.getThousandSeparator is slow?

rrsln:
I doubt that LocaleUtils.getThousandSeparator is slow. For some reason it was written that way for InputNumber and I didn't want to change that. Also it doesn't harm when some method is not invoked assuming that readability is not sacrificed.

tandraschko:
Yep, i see.
Also created a feature request for JSF 3.x, so we can use a lambda for the future: https://github.com/eclipse-ee4j/mojarra/issues/4568

tandraschko:
Would be great if you can do following refactoring:
1) Move InputNumber get/setDecimalSeparator get/setThousandSeparator to InputNumberBase
2) Create PropertyKeys for this 2 properties instead of Strings and also use it in the getter/setter
3) Rename ComponentUtils#getValue - it's to generic for such a generic class. Maybe something evalLazy or something... no idea. There should be a reference to "stateHelper#eval" in the name

tandraschko:
it also conflicts now

rrsln:
fixed all comments. Renamed ComponentUtils#getValue to ComponentUtils#eval, it seems the best option to me.However feel free to rename if you believe that some other name is better.

tandraschko:
Looks good for me now
Can someone retest it?

melloware:
@tandraschko from my testing this PR looks good I didn't find anything wrong.

Issue #4646, opened 2019-03-18
Move to Java8 LocalDate/Time


tandraschko:

tandraschko:
If someone would like to provide a PR, it would be great!

christophs78:
What do we expect?

Which components should work with LocalDate/Time?
* Calendar
* Datepicker
* Timeline
* Schedule

Do we expect backward-compatibility?
Should they support java.util.Date and LocalDate?
When I look into BaseCalendarRenderer, method getConvertedValue we see some existing approaches.
Replacing SimpleDateFormat with DateTimeFormatter and (eg) LocalDate.parse would be simple. But is this the way to go?

When we expect backward-compatibility... How should we detect whether we should use the old or the new? Do we need some kind of (global?) switch?

JSF 2.3 offers LocalDate/Time-Support to some degree: https://javaee.github.io/glassfish/doc/5.0/vdldoc/f/convertDateTime.html

tandraschko:
IMO if older types like Date are used, they can simple add a converter but internally we should not use the old APIs.

melloware:
I think we should support old and new dates.  There are millions of lines of code using java.util.Date in applications and asking people to touch code that is working just to change to a LocalDate or introduce the converter feels wrong.   Its a pretty big breaking change.

tandraschko:
BIG -1 to maintain multiple versions
everyone can put a simple converter in it

christophs78:
Maintaing multiple versions within the PF-codebase does not seem to be easy task.
But i think we should provide a easy way to migrate existing code utilizing java.util.Date. (Maybe deliver a Converter OOTB with PF which provides backward-compatibility. Which regonices things like the pattern-attribute of Calendar and DatePicker.)
From my perspective step 1 would be to build a POC for Calendar and DatePicker und look how this works for new projects using LocalDate/Time. And check how to offer a easy migration path.

Step 2 would be to look at Timeline and Schedule. This is something we can´t solve with a converter. We have to look at classes like DefaultScheduleEvent.

tandraschko:
Yep, i would create a cool converter and add them to the migration guide. 

+1 for step1 and 2

christophs78:
Played around a bit with moving DatePicker-component to LocalDate/Time.
What should we do with attributes like disabledDates or disabledDays? (disabledDates expects currently List<java.util.Date>, disabledDays a list of int-values)
```
<p:outputLabel for="disabledDD" value="Disabled Date/Day:" />
[...]

PR #4782, opened 2019-05-05
#4646 Move to Java8 LocalDate/Time (Calendar + Datepicker)

christophs78:
See also https://github.com/primefaces/showcase-facelift/pull/41. The updated Showcase contains separated xhtml-pages and ViewBeans for DatePicker and Calendar with java.util.Date and java.time.LocalDate/LocalDateTime/LocalTime.

Backward-compatibility with java.util.Date is provided OOTB. So no migration work necessary. 
OOTB-backward-compatibility may be removed with a future (major) release.


melloware:
Looks like a lot of work went into this.  Nice work.  I guess you need answers on all the TODO comments?

christophs78:
Yes, it would help to get answers/opinions on the TODO comments.

src/main/java/org/primefaces/component/calendar/BaseCalendarRenderer.java
```diff
@@ -144,12 +153,14 @@ public Object getConvertedValue(FacesContext context, UIComponent component, Obj
...
-        //Delegate to global defined converter (e.g. joda or java8)
+        //Delegate to global defined converter (e.g. joda or java.util.Date)
         try {
             ValueExpression ve = uicalendar.getValueExpression("value");
             if (ve != null) {
                 Class type = ve.getType(context.getELContext());
-                if (type != null && type != Object.class && type != Date.class) {
+                //if (type != null && type != Object.class && type != Date.class) {
+                if (type != null && type != Object.class && type != LocalDate.class && type != LocalDateTime.class && type != LocalTime.class) {
+                    //TODO: right decision to check Java8-Types instead of java.util.Date?
```
tandraschko: why not? :D

src/main/java/org/primefaces/component/api/UICalendar.java
```diff
@@ -153,19 +155,49 @@ public boolean hasTime() {
...
+            /*
+            Locale locale = calculateLocale(getFacesContext());
+            //TODO: SimpleDateFormat --> DateTimeFormatter
+            String localePattern = ((SimpleDateFormat) DateFormat.getDateInstance(DateFormat.SHORT, locale)).toPattern();
             String userTimePattern = getPattern();
 
             timeOnlyPattern = localePattern + " " + userTimePattern;
+             */
+
+            //TODO: Default-Pattern when no Pattern is specified via Pattern-Attribute?
```
tandraschko: shouldn't we just add a default pattern in the stateHelper call?

src/main/java/org/primefaces/component/api/UICalendar.java
```diff
@@ -217,6 +249,7 @@ public void setLabelledBy(String labelledBy) {
     public Converter getConverter() {
         Converter converter = super.getConverter();
 
+        //TODO: Why does it matter whether Client-Side-Validation is enabled?
```
tandraschko, line 224: no idea, could you try to debug and/or remove this and check what happens?

src/main/java/org/primefaces/component/calendar/BaseCalendarRenderer.java
```diff
@@ -163,20 +174,49 @@ public Object getConvertedValue(FacesContext context, UIComponent component, Obj
             throw e;
         }
 
-        //Use built-in converter
-        format = new SimpleDateFormat(uicalendar.calculatePattern(), uicalendar.calculateLocale(context));
-        format.setLenient(false);
-        format.setTimeZone(CalendarUtils.calculateTimeZone(uicalendar.getTimeZone()));
         try {
-            return format.parse(submittedValue);
+            //TODO: is there some way to dynamically determine the type?
```
tandraschko: which type?

christophs78: The type of the value which is bound to the component.
This would allow us to do something like the following pseudocode:
```
if (typeOfValue == java.util.Date) {
   //convert to java.util.Date
}
else if (typeOfValue == LocalDate)
{
   //convert to java.time.LocalDate
}
else if (typeOfValue == java.time.LocalDateTime)
{
   //convert to java.time.LocalDateTime
}
else if (typeOfValue == java.time.ZonedDateTime)
{
   //convert to java.time.ZonedDateTime
}
```
So we won´t need a converter for java.util.Date, we would be able to support ZonedDateTime.

src/main/java/org/primefaces/component/api/UICalendar.java
christophs78: I tried both variant´s (via primefaces.CLIENT_SIDE_VALIDATION) but got no real idea. Thought about throwing this out.
Maybe there is some missalignement between the client-side-validation and the server-side-validation. And this code tries to work arround this issue. Hoped someone reminds why this was implemented this way some time ago.

src/main/java/org/primefaces/component/calendar/BaseCalendarRenderer.java
christophs78: Maybe i simply have to look a few lines up.
```
ValueExpression ve = uicalendar.getValueExpression("value");
            if (ve != null) {
                Class type = ve.getType(context.getELContext());
```
I will try this.

christophs78:
Backward-Compatibility with java.util.Date:
My first iteration required a converter.
`<p:datePicker id="inline" value="#{calendarView.date1}" inline="true" converter="pf.DateBackwardCompatiblityConverter" />`

My second iteration determines the type of the value expression dynamical. So no migration work necessary. Only exception is DatePicker with selectionMode = multiple and selectionMode = range.

src/main/java/org/primefaces/component/api/UICalendar.java
christophs78: Checked how this was implemented in 6.2. (Before Datepicker)
https://github.com/primefaces/primefaces/blob/6_2/src/main/java/org/primefaces/component/calendar/CalendarRenderer.java
6.2 did not had this (from my perspective) awkward behaviour depending on Client-Side-Validation.
So i´ll throw this out. At least as long until someone come´s up with an argument, why this needs to be this way.
(This allows me to rollback org.primefaces.convert.DateTimeConverter to the original version.)

src/main/java/org/primefaces/component/api/UICalendar.java
christophs78: Compared to java.util.Date - Code and think the easy solution is equivalent to the old behavior.

tandraschko:
AFAICS if we can determine the generic type of the list, we could also remove DateBackwardCompatiblityConverter?

christophs78:
Yes.

tandraschko:
Ok, just google a bit. This should be possible via TypeArgument API.
If not, let me know

christophs78:
I already googled a few days ago...
https://stackoverflow.com/questions/1942644/get-generic-type-of-java-util-list
At runtime there is no typing Information. Only the compiler knows how the list is typed. At runtime it´s a untyped list.
The only way to find out the type  of the list at runtime would be reflections. But I do not think this works together with value-binding via EL.

tandraschko:
Hmmm, not "easy" but it's possible:
Get the real value expression / reference via: ValueExpressionAnalyzer.getReference and get the field via reflection. Then you can also get the type as you already mentioned.

This way it will also work with a null list.

christophs78:
Sounds good. I will try this tomorrow. :-)

christophs78:
It works! :-)
Hope we are able to merge this PR within the next days.
After this is done i´ll work on a second PR which adds LocalDate/Time-Support to Schedule and Timeline.

tandraschko:
great :)
I will try to do a big review on sunday

christophs78:
Is there a chance to make some progress on this PR? I´d realy like to continue with Timeline, Schedule, ... after this is done. :-)

tandraschko:
Will review next month, sry

tandraschko:
@christophs78 the changes are looking fine! I assume that i you tested everything (and also CSV)?

christophs78:
@tandraschko: Thank you for reviewing! I tested everything that came into my mind. (As you can see in the PR in the showcase-project, quite a good amount of testcases.)
CSV??? Do you mean comma separated values? Hm. During CSV-export of a DataTable which contains Calendar or Datepicker? Please help me. :-)

melloware:
No CSV = Client Side Validation.   https://www.primefaces.org/showcase/ui/csv/basic.xhtml

christophs78:
https://github.com/primefaces/primefaces/blob/4c435187cf75befa2a53e4b923198fe9359531ba/src/main/resources/META-INF/resources/primefaces/validation/validation.js#L435

First assessment:
On the client-side it should not matter, which type we have on the server. Relevant is the data-p-con-Attribute:
`<input id="form:popup_input" ... data-p-con="javax.faces.DateTime"`

Seems CSV for Calendar/DatePicker required to explicit use <f:convertDateTime pattern="xxx" />. This does not change.

I´lll create a testcase to be totally sure.

christophs78:
CSV is fixed. :-)
Thanks for drawing my attention to this!

NicolaIsotta:
@christophs78 would it be possible to add `java.time.YearMonth` too? I think it combines well with `<p:datePicker view="month"/>`

christophs78:
@NicolaIsotta: Done :-)

PR #4785, opened 2019-05-07
mod: validate script nonce to prevent injections, use `addHeader` instead of `setHeader` to not override pre-existing headers (#3505)

cnsgithub:


tandraschko:
wouldn't it be better to move the validation to CspState#getNonce?

cnsgithub:
@tandraschko For the moment I omitted `strict-dynamic` because when looking at the overview image (https://github.com/primefaces/primefaces/issues/3505#issuecomment-489038655), this might have negative impact on DOM XSS prevention. Example: with `strict-dynamic` in place the last test in http://localhost:8080/showcase/ui/misc/csp.xhtml (_XSS_ tab) leads to an injection.

On the other hand, when omitting `strict-dynamic`, issues like https://github.com/primefaces/showcase-facelift/issues/39 do not get solved and we are in a situation where we would need script domain whitelists again - which is unsafe.

cnsgithub:
> wouldn't it be better to move the validation to CspState#getNonce?

You're right, I moved it.

tandraschko:
Does Base64.getDecoder().decode(nonce); throw a exception on empty/blank string, too?

cnsgithub:
> Does Base64.getDecoder().decode(nonce); throw a exception on empty/blank string, too?

`null` -> NPE, `""` -> no exception, `"<invalidBase64>"` -> `IllegalArgumentException` :cry:

cnsgithub:
@tandraschko Sorry, added another commit to https://github.com/cnsgithub/primefaces/tree/csp-3505

tandraschko:
PR please :)

PR #4792, opened 2019-05-08
Better performance for TreeTable (update TreeNodeChildren)

crocimi:
Changes for better performance (no need to recalculate rowIndex for all nodes)


melloware:
@Mich2424 all PR's must be associated to an Issue ticket.  Is this related to a ticket? I also noticed you have two similar PR's open?  Are they different?

crocimi:
ok sorry, I will procede the correct way

PR #4794, opened 2019-05-09
Fix #4777: Floating Label

### LINKED ISSUE #4777: InputText: Float Label for Primefaces
It would be great to have support for floating labels on inputTexts or other inputs like there is already in primeng or primereact:
https://www.primefaces.org/primeng/#/inputtext

:)

melloware:


tandraschko:
@mertsincan looks ok for me. WDYT?

PR #4796, opened 2019-05-09
fixes #4791: NPE in range validation of slider within dynamic tabView

### LINKED ISSUE #4791: Slider: NullPointerException when using in dynamic TabView
## 1) Environment
- PrimeFaces version: 7.0.1
- Affected browsers: IE, Chrome, Firefox

## 2) Expected behavior
When submitted the given form I do not expect a NullPointerException to be thrown.

## 3) Actual behavior
A NullPointerException is thrown

## 4) Steps to reproduce
Use the given example (extended example from PrimeFaces Demo of Slider), when clicking the "ok" button then following error is thrown:
```
Caused by: java.lang.NullPointerException
        at deployment.infoChantier-wildfly16-PROD.war//org.primefaces.component.slider.Slider.validate(Slider.java:106)
```
The error seems to be coming from the "getSubmittedValue" funtion in the Slider.java.
The line:
`return input.getSubmittedValue() == null && input.isLocalValueSet() ? input.getValue() : input.getSubmittedValue();`
returns "NULL" because the "isLocalValueSet" return "false" and so the "getSubmittedValue" is returned which is "NULL" whereas the "getValue" would return the right number.

I do not know if this helps but this is as far as I came.

## 5) Sample XHTML
```
    <f:view>
        <h:head>
            <title>InfoChantier</title>
        </h:head>
        <h:body>
            <h:form>
                <p:tabView dynamic="true" rows="20" value="#{sliderView.tabs}" var="tab">
                    <p:tab title="#{tab}">
                            <h:outputText id="displayRange" value="Between #{sliderView.number8} and #{sliderView.number9}" />
                            <p:slider for="txt6,txt7" display="displayRange" style="width: 400px" range="true" displayTemplate="Between {min} and {max}" />
                        </h:panelGrid>
                        <h:inputHidden id="txt6" value="#{sliderView.number8}" />
                        <h:inputHidden id="txt7" value="#{sliderView.number9}" />
                    </p:tab>
                </p:tabView>
                <p:commandButton value="Ok" />
            </h:form>
        </h:body>
    </f:view>
</ui:composition>
```

## 6) Sample bean
```
package lu.etat.pch.infoChantier.ejb.controllers;

import javax.inject.Named;

@Named
public class SliderView {
    private int number8 = 30;
    private int number9 = 80;

    public int getNumber8() {
        return number8;
    }

    public void setNumber8(int number8) {
        this.number8 = number8;
    }

    public int getNumber9() {
        return number9;
    }

    public void setNumber9(int number9) {
        this.number9 = number9;
    }

    public String[] getTabs() {
        String[] result = new String[2];
        result[0] = "Tab 1";
        result[1] = "Tab 2";
        return result;
    }
}
```

cnsgithub:
This PR is probably just fixing the symptoms (NPE) instead of the real cause.

### DIFF
--- a/src/main/java/org/primefaces/component/slider/Slider.java
+++ b/src/main/java/org/primefaces/component/slider/Slider.java
@@ -112,8 +112,13 @@ public class Slider extends SliderBase {
             String[] inputIds = getFor().split(",");
             UIInput inputFrom = (UIInput) SearchExpressionFacade.resolveComponent(context, this, inputIds[0]);
             UIInput inputTo = (UIInput) SearchExpressionFacade.resolveComponent(context, this, inputIds[1]);
-            String valueFromStr = getSubmittedValue(inputFrom).toString();
-            String valueToStr = getSubmittedValue(inputTo).toString();
+            Object submittedValueFrom = getSubmittedValue(inputFrom);
+            Object submittedValueTo = getSubmittedValue(inputTo);
+            if (submittedValueFrom == null || submittedValueTo == null) {
+                return;
+            }
+            String valueFromStr = submittedValueFrom.toString();
+            String valueToStr = submittedValueTo.toString();
             double valueFrom = Double.valueOf(valueFromStr);
             double valueTo = Double.valueOf(valueToStr);
             if (valueTo < valueFrom) {


cnsgithub:
@tandraschko Did you notice this PR's comment?

PR #4799, opened 2019-05-09
Optimize node add/remove

crocimi:
for better performance

### DIFF
--- a/src/main/java/org/primefaces/model/TreeNodeChildren.java
+++ b/src/main/java/org/primefaces/model/TreeNodeChildren.java
@@ -54,7 +54,7 @@ public class TreeNodeChildren extends TreeNodeList {
             eraseParent(node);
             boolean result = super.add(node);
             node.setParent(parent);
-            updateRowKeys(parent);
+            updateRowKeys(parent.getChildCount() - 1, this.parent);
             return result;
         }
     }
@@ -71,13 +71,14 @@ public class TreeNodeChildren extends TreeNodeList {
             eraseParent(node);
             super.add(index, node);
             node.setParent(parent);
-            updateRowKeys(parent);
+            updateRowKeys(index, this.parent);
         }
     }
 
     @Override
     public boolean addAll(Collection<? extends TreeNode> collection) {
         Iterator<TreeNode> elements = (new ArrayList<TreeNode>(collection)).iterator();
+        int index = collection.size();
         boolean changed = false;
         while (elements.hasNext()) {
             TreeNode node = elements.next();
@@ -93,7 +94,7 @@ public class TreeNodeChildren extends TreeNodeList {
         }
 
         if (changed) {
-            updateRowKeys(parent);
+            updateRowKeys(index, parent);
         }
 
         return (changed);
@@ -117,7 +118,7 @@ public class TreeNodeChildren extends TreeNodeList {
         }
 
         if (changed) {
-            updateRowKeys(parent);
+            updateRowKeys(index, parent);
         }
 
         return (changed);
@@ -140,7 +141,7 @@ public class TreeNodeChildren extends TreeNodeList {
             super.set(index, node);
             previous.setParent(null);
             node.setParent(parent);
-            updateRowKeys(parent);
+            updateRowKeys(this.parent, node);
             return previous;
         }
     }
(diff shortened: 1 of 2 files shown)


melloware:
Much better thanks!

src/main/java/org/primefaces/model/TreeNodeChildren.java
```diff
@@ -54,7 +54,7 @@ public boolean add(TreeNode node) {
             eraseParent(node);
             boolean result = super.add(node);
             node.setParent(parent);
-            updateRowKeys(parent);
+            updateRowKeys(parent.getChildCount() - 1, this.parent);
```
Rapster: Please, don't use unnecessary `this` references

src/main/java/org/primefaces/model/TreeNodeChildren.java
```diff
@@ -71,13 +71,14 @@ else if ((index < 0) || (index > size())) {
...
             node.setParent(parent);
-            updateRowKeys(parent);
+            updateRowKeys(index, this.parent);
         }
     }
 
     @Override
     public boolean addAll(Collection<? extends TreeNode> collection) {
         Iterator<TreeNode> elements = (new ArrayList<TreeNode>(collection)).iterator();
+        int index = collection.size();
```
Rapster, line 81: Coud you move index declaration into changed if statement, it's not used anywhere else

src/main/java/org/primefaces/model/TreeNodeChildren.java
```diff
@@ -140,7 +141,7 @@ else if ((index < 0) || (index >= size())) {
             super.set(index, node);
             previous.setParent(null);
             node.setParent(parent);
-            updateRowKeys(parent);
+            updateRowKeys(this.parent, node);
```
Rapster: Please, don't use unnecessary this references

src/main/java/org/primefaces/model/TreeNodeChildren.java
```diff
@@ -168,7 +169,7 @@ else if ((index < 0) || (index >= size())) {
             TreeNode previous = get(index);
             super.set(index, node);
             node.setParent(parent);
-            updateRowKeys(parent);
+            updateRowKeys(this.parent, node);
```
Rapster: Please, don't use unnecessary this references

src/main/java/org/primefaces/model/TreeNodeChildren.java
```diff
@@ -214,4 +216,24 @@ private void updateRowKeys(TreeNode node) {
             }
         }
     }
+
+    private void updateRowKeys(int index, TreeNode node) {
+        int childCount = node.getChildCount();
+        if (childCount > 0) {
+            for (int i = index; i < childCount; ++i) {
```
Rapster: `++i`: use `i++` instead please

src/main/java/org/primefaces/model/TreeNodeChildren.java
```diff
@@ -214,4 +216,24 @@ private void updateRowKeys(TreeNode node) {
...
+                TreeNode childNode = (TreeNode) node.getChildren().get(i);
+                String childRowKey = node.getParent() == null ? String.valueOf(i) : node.getRowKey() + "_" + i;
```
Rapster: In these 3 updateRowKeys methods, these lines are duplicated, could you please refactor that a bit so it can be reused:
```java
String childRowKey = node.getParent() == null ? String.valueOf(i) : node.getRowKey() + "_" + i;
childNode.setRowKey(childRowKey);
updateRowKeys(childNode);
```

Rapster (changes requested):

src/main/java/org/primefaces/model/TreeNodeChildren.java
Rapster: Still there

crocimi: sorry, fixed

Rapster (approved):
Thank you, great job! :wink:

melloware (approved):

tandraschko:
@Mich2424 Could you please provide some unittests, so we can be sure that we don't break something?

tandraschko:
@Mich2424 ping :D

crocimi:
I'll do it soon :-)

tandraschko:
Thanks!

PR #4814, opened 2019-05-16
WebXmlParser: fix xpath-queries for namespace-aware XPath implementations

elbird:
To get the error-pages from the web.xml the WebXmlParser uses the javax.xml.xpath.XPathFactory to create xpath-queries. Which implementation of the XPathFactory is actually used can be globally defined via system-properties or Java service provider configuration. If a namespace-aware XPathFactory-implementation such as the net.sf.saxon.xpath.XPathFactoryImpl is configured during runtime the xpath-queries for the error-pages fail and no error-pages are available when the PrimeExceptionHandler handles an exception. 
This pull request improves the xpath-queries so they will work on both namespace-aware and namespace-unaware XPath implementations and adds a unit test which verifies the fix.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -200,6 +200,13 @@
             <scope>provided</scope>
         </dependency>
 
+        <dependency>
+            <groupId>net.sf.saxon</groupId>
+            <artifactId>Saxon-HE</artifactId>
+            <version>9.9.0-2</version>
+            <scope>test</scope>
+        </dependency>
+
         <!-- will be included and relocated by shade plugin -->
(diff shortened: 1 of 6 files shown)


melloware:
Please create and issue first and link it to this PR.  They only accept PR's with an associated issue. Thanks!

elbird:
Thanks for the heads up!  I created an issue for this pull request: #4815

Rapster (approved):

tandraschko:
Thanks!

PR #4836, opened 2019-05-24
Fix #4818: DatePicker better handling of range and multiple.

### LINKED ISSUE #4818: Add rangeSeparator attribute to DatePicker
With pattern set to "yyyy-MM-dd" it's impossible to use selectionMode "range". DatePickerRenderer has hardcoded separator string in line 186, and since the number of elements in the parts array isn't equal to 2, it doesn't even try to parse those dates. Please allow to specify different separator string for range and multiple selectionModes.

## 1) Environment
- PrimeFaces version: 7.0.2
- Application server + version: WildFly 16.0.0
- Affected browsers: All

## 2) Expected behavior

Getting list with two elements.

## 3) Actual behavior

Getting empty list.

melloware:


tandraschko:
@mertsincan can you review?

melloware:
It looks like something similar was already done for the current calendar "multiple" mode.
https://github.com/primefaces/primefaces/blob/3617b0d9f1e8fbd61d946dfa3ce09b36a049ee7c/src/main/java/org/primefaces/util/CalendarUtils.java#L139-L142

PR #4845, opened 2019-05-29
refactoring encodeFilter() method

matinh:
This is just refactoring of the encodeFilter() method from the DataTableRenderer into smaller methods to make life easier if one has to override the renderer.

See also forum post: https://forum.primefaces.org/viewtopic.php?f=3&t=59011&p=175505

### DIFF
--- a/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
+++ b/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
@@ -842,110 +842,11 @@ public class DataTableRenderer extends DataRenderer {
     }
 
     protected void encodeFilter(FacesContext context, DataTable table, UIColumn column) throws IOException {
-        Map<String, String> params = context.getExternalContext().getRequestParameterMap();
         ResponseWriter writer = context.getResponseWriter();
         UIComponent filterFacet = column.getFacet("filter");
 
         if (filterFacet == null) {
-            String separator = String.valueOf(UINamingContainer.getSeparatorChar(context));
-            boolean disableTabbing = table.getScrollWidth() != null;
-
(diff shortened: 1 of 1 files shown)


melloware:
Please create an Issue and link it to this PR: https://github.com/primefaces/primefaces/issues

PR's are usually not accepted without issues unless its a documentation only change.

src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
```diff
@@ -842,122 +842,151 @@ else if (headerText != null) {
...
+
+        writer.endElement("input");
+    }
+
+    protected Object findFilterValueForColumn(FacesContext context, DataTable table,
+        UIColumn column, String filterId) {
+
+        Object filterValue;
+        if (table.isReset()) {
+            filterValue = "";
```
melloware: Use Constants.EMPTY_STRING instead of "".

melloware (changes requested):

src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
matinh: Well, I just did refactoring, no changes to code otherwise. But I will do if you want me to.

melloware: Yep may as well, last one to touch it owns it! 😄

melloware (approved):

tandraschko:
@melloware would you Like to check the PR and retest with the showcase If nothing is broken?

melloware:
I just pulled down this PR and ran it through all my expected scenarios and the Showcase Filter examples and everything is working properly.  I would say good to merge.

PR #4854, opened 2019-06-04
Fix #3192: InputNumber do not fire change event with suffix

### LINKED ISSUE #3192: InputNumber: auto send request when focus out if use attribute symbol
# Usecase
I'm now using Primeface 6.1, as i know that Primefaces already moved some elements from Primefaces Extension to Primefaces and it include p:inputNumber.
Now i have the xhtml like this:
```xml
<p:inputNumber   value="#{myBean.amount}"
			minValue="0" decimalPlaces="0">
		 <p:ajax event="change" global="false" listener="#{myBean.doValueChangeEvent}"/>
</p:inputNumber>
```
It work fine, when i focus in then enter value and focus out, it will fire request to server.
But when i define the symbol for my value (because it's about currency), like this:
```xml
<p:inputNumber symbol=" CHF" symbolPosition="s"
 			  value="#{myBean.amount}"
			minValue="0" decimalPlaces="0">
		 <p:ajax event="change" global="false" listener="#{myBean.doValueChangeEvent}"/>
</p:inputNumber>
```
it mean it will show "CHF" as suffix of my value. When i only focus in and focus out, this element will fire event although i did not input any value (still empty).

## 1) Environment
- PrimeFaces version: 6.1
- It does not work on Primeface 6.2.RC1 as well
- Application server + version: Tomcat 8.5
- Affected browsers: IE 11

## 2) Expected behavior
Should not fire any request in case user does not enter value to inputNumber. 

## 3) Actual behavior
It fire request if i only focus in and focus out without changing value
..

## 4) Steps to reproduce
1. Focus in to InputNumber, the suffix will appear
2. Focus out then request is fired to server

## 5) Sample XHTML
```xml
<p:inputNumber symbol=" CHF" symbolPosition="s"
 			  value="#{myBean.amount}"
			minValue="0" decimalPlaces="0">
		 <p:ajax event="change" global="false" listener="#{myBean.doValueChangeEvent}"/>
</p:inputNumber>
```
Thanks

melloware:


tandraschko:
please notice it somewhere that we modified this script and what exactly
otherwise all infos are likely lost if we upgrade the script someday

melloware:
Good catch!  Added "extensions.txt" similar to how we do it for other components.

PR #4856, opened 2019-06-04
Fix #4855 - Refactor AjaxBehaviorHandler

### LINKED ISSUE #4855: Refactor AjaxBehaviorHandler
Since I'm working on https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/706 and bringing a new tag handler on top of `AjaxBehaviorHandler`. I realized `AbstractBehaviorHandler` could be a lot more simplified and extendable to ease overriding existing tag handlers

PR on his way, as I'd like to read your comments before merging into master

Rapster:


tandraschko:
+1 in generell but -1 for the "I" prefix
I would call it "BehaviorAttributes" as component properties can be received via component.getAttributes

Rapster:
Thanks Thomas. I applied your suggestion

tandraschko:
Cool!
Should we name it "BehaviorAttribute" or "BehaviorAttributeS"?

Rapster:
BehaviorAttribute is fine, I understand your question but it's because we use enum that makes us questionning the name.

Rapster:
Though should I rename `AbstractBehavior#getAllProperties` into `AbstractBehavior#getAllAttributes`?

tandraschko:
+1

Rapster:
If that's okay, i'll merge that PR into master then

melloware:
+1

Issue #349, opened 2015-05-27
Navigation with <p:autoComplete> in <p:dataTable>


Pooum:
Hello,

In a datatable with cell editor it is possible to navigate from a cell to the next one using the tab key.
Except when a column is an autocomplete, in this case we can navigate from a cell to an autocomplete but not from an autocomplete to another one.

Example, take the editable datatable fom promefaces showcase ( http://www.primefaces.org/showcase/ui/data/datatable/edit.xhtml ) , and replace : 

``` html
<p:column headerText="Brand">
           <p:cellEditor>
               <f:facet name="output"><h:outputText value="#{car.brand}" /></f:facet>
               <f:facet name="input">
                   <h:selectOneMenu value="#{car.brand}" style="width:100%">
                       <f:selectItems value="#{dtEditView.brands}" var="man" itemLabel="#{man}" itemValue="#{man}" />
                   </h:selectOneMenu>
               </f:facet>
           </p:cellEditor>
       </p:column>
```

by 

``` html
<p:column headerText="Brand">
            <p:cellEditor>
                <f:facet name="output">
                    <h:outputText value="#{car.brand}" />
                </f:facet>
                <f:facet name="input">
                    <p:autoComplete value="#{car.brand}" var="p" itemLabel="p" style="width:100%">
                    </p:autoComplete>
                </f:facet>
            </p:cellEditor>
        </p:column>
```

Thanks

cagataycivici:
This is a question, please use forum instead.

Pooum:
Hello,

I don't understand your message, it's not a question, may be I haven't explain enough. 

Actually navigate with "tab", in a datatable containing some autocompletes as input doesn't work.
My point is that it could be good to do it ; actually I need it and I think I'm not alone in this case.

Think of a use case where the user want to edit many data in a datatable.
The best way to do it is to click on the first cell, fill it then press "tab" to go to the next cell ; fill it etc.

It works with checkbox, inputText, selectOneItem and some others. 
Make it work with autocomplete seems logical too.

cagataycivici:
Ok, we'll check again, thanks.

jederfreitas:
Hi Cagatay, Pooum,

Is there a solution or a workaround about it? I have the same problem.

Thanks

Prismapanda:
Hi there,

we experienced this behaviour even if a cell is not editable at all (no p:cellEditor present, v6.0.3). Would be nice to hear of a fix.

Rapster:
[...]

Issue #3314, opened 2018-02-10
TAB navigation in autocomplete inside a datatable


denizque:
I have the newest version of primefaces elite.
I put an editable p:dataTable with various columns in it.
When I press the tab key to switch from one column to the other it switches as long as it not an autocomplete component.
The column which has p:inputText or p:selectOneMenu switches fine.
But if I put an autocomplete in that column and I press tab key nothing happens.
Is this a bug? Or what should I do to make it navigating through columns when it has autocomplete in it?

melloware:
Related to this #3087 ?  Looks like something was tweaked in PF 6.1.11.

https://github.com/primefaces/primefaces/issues/3087

tandraschko:
@mertsincan could you check that?

huseyinturkmenoglu:
Hi, 

I don't see any problem. 

Could you share your version of primefaces ? 
And 
Please, you could send your sample code or  create  mvn projet from  : https://github.com/primefaces/primefaces-test

melloware:
Fixed with PR: https://github.com/primefaces/primefaces/pull/4858

If you set `autoSelection="false"` on AutoComplete the TAB key will work like normal and not select the item.

Issue #344, opened 2015-05-25
AutoComplete: Item selection on focusout


codingPF:
Item is only selected with a click on the result item - not by just typing the text (which exactly matches the result) and just leaving the filed / leaving focus.

Reproducing the issue:

- Change AutoCompleteView line 64 to: if(skin.getName().toLowerCase().startsWith(query.toLowerCase())) {

- /ui/input/autoComplete.xhtml

- Type Afterdark into the "Pojo" autocomplete - do not leave the box (no enter/no tab/no selecting by mouse); Just leave focus by clicking somewhere on the screen

- Press Submitt > Conversion error (internally the ID exchange did not happen)

cagataycivici:
I don't understand anything, please edit your description with a clear one.

melloware:
PR submitted. If you have autoSelection="false" the TAB key will work like normal and not select the item.

PR #4858, opened 2019-06-05
Fix #4857,#344,#349,#3314: TAB key will not select item if autoSelection="false"

### LINKED ISSUE #4857: AutoComplete: Item from dropdown selected when pressing "tab"
## 1) Environment
- PrimeFaces version: 7.0.3
- Affected browsers: Chrome, Firefox, IE

## 2) Expected behavior
First element of dropdown should not be selected when using "tab" button.

## 3) Actual behavior
When setting focus to the next component using the "tab" button, the first element of the dropdown is selected.

## 4) Steps to reproduce
Go to the primefaces showcase ([https://www.primefaces.org/showcase/ui/input/autoComplete.xhtml]) now go to the first Autocomplete and start typing for example a "d". Now suggestions are shown (d0, d1, ...). Now you have to type a "1" and then press the tabulator before the dropdown list is updated.
-> "d0" is selected nevertheless you typed "d1".
I think when exiting the autocomplete "d1" should stay (which is the case when you exist it using the mouse)
I also tried to set the "autoSelection" to "false" but this didn't helped.

## 5) Sample XHTML
Use primefaces showcase

## 6) Sample bean
Primefaces showcase

melloware:
If AutoComplete has  `autoSelection="false"` it makes the TAB key work like normal and just focus the next field.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
+++ b/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
@@ -316,10 +316,14 @@ PrimeFaces.widget.AutoComplete = PrimeFaces.widget.BaseWidget.extend({
                         break;
 
                     case keyCode.TAB:
-                        if(highlightedItem.length) {
+                        if(highlightedItem.length && $this.cfg.autoSelection) {
                             highlightedItem.trigger('click');
+                        } else {
+                            $this.hide();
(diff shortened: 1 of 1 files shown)


melloware:
I tested and this also fixes #349 and #3314

tandraschko:
@mertsincan can you please review?

PR #4861, opened 2019-06-05
Fix #4860: Add aria attributes to overlay panel

### LINKED ISSUE #4860: OverlayPanel: Add ARIA attributes to target
## 1) Environment
- PrimeFaces version: latest
- Affected browsers: all

## 2) Expected behavior
The overlay panel adds ARIA `aria-expanded` and `aria-controls` attributes to the target component and toggles `aria-expanded` if the panel visibility changes.

## 3) Actual behavior
Nothing
***
PR will be added soon

Rawi01:


melloware (approved):

tandraschko:
why you do the if statement (if (this.target) {) not in the first code change, too?

Rawi01:
It gets set in line 26 and will always be there. Even if the jquery selector can not find the target it should be safe to call the method. The same happens in `bindTargetEvents()`. 
`postShow`/`postHide` needs that check because the target will be `undefined` if the `for` attribute of the `overlayPanel` is empty.

PR #4869, opened 2019-06-07
Fix #4868: Hide/Show all items at once

### LINKED ISSUE #4868: SelectOneMenu: Improve filter performance
## 1) Environment
- Application server + version: **all**
- Affected browsers: **Internet Explorer, Firefox**

## 2) Expected behavior
Using the filter in a SelectOneMenu with many entrys should be fast and without noticable lag.

## 3) Actual behavior
It lags...

## 4) Steps to reproduce
- Create a SelectOneMenu with 2000 items and filtering enabled
- Start typing
***
PR will be added soon

Rawi01:
This improves the filter performance. A full refactoring might speed up this even more but requires way more testing than these minimal changes.

## Before
**Firefox:**
FF_before
**Internet Explorer:**
IE_before
## After
**Firefox:**
FF_after
**Internet Explorer:**
IE_after

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/forms/forms.selectonemenu.js
+++ b/src/main/resources/META-INF/resources/primefaces/forms/forms.selectonemenu.js
@@ -874,21 +874,29 @@ PrimeFaces.widget.SelectOneMenu = PrimeFaces.widget.DeferredWidget.extend({
             this.itemsContainer.children('.ui-selectonemenu-item-group').show();
         }
         else {
+            var hide = [];
+            var show = [];
+
             for(var i = 0; i < this.options.length; i++) {
                 var option = this.options.eq(i),
                 itemLabel = this.cfg.caseSensitive ? option.text() : option.text().toLowerCase(),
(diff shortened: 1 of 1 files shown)


melloware (approved):

tandraschko:
why is it faster?

melloware:
That is a good question, from my research so far I can't explain why it's faster? @Rawi01 
?

Rawi01:
It is faster because it doesn't trigger layout events of the browser engine after each hide/show but only once in the end. I attached two additional profiler screenshots to illustrate this.

**Before:**
grafik
**After:**
grafik

melloware:
That is what I suspected.   I was looking for documentation about it anywhere about that the Layout would only be triggered once the $.each was done but couldn't find any.

Rapster:
I have a feeling we've met this problem before... Worth to have a look and apply the same fix

melloware:
@Rapster I was thinking the same thing. We should look for similar for loops in the JS code.

PR #4871, opened 2019-06-07
Fix #4870: Remove value attribute

### LINKED ISSUE #4870: Paginator: RowsPerPageDropdownRenderer adds invalid value attribute
## 1) Environment
- Application server + version: **all**
- Affected browsers: **all**

## 2) Expected behavior
PrimeFaces creates beautiful and correct HTML code

## 3) Actual behavior
RowsPerPageDropdownRenderer adds the value attribute to the `<select>` which triggers a validation error. (https://html.spec.whatwg.org/multipage/form-elements.html#the-select-element)

***
PR will be added soon

Rawi01:


tandraschko:
not sure if this change is correct for all cases.

Rawi01:
Can you explain why you think that there is a usecase for this one?

tandraschko:
it think thats the current value for rowsPerPage, where is the usecase without the current selection value?

Rawi01:
The `value` attribute is not allowed for a `<select>` (https://validator.w3.org/nu/?doc=https%3A%2F%2Fwww.primefaces.org%2Fshowcase%2Fui%2Fdata%2Fdatatable%2Fpaginator.xhtml)

Yes, it is obviously the current value but it does not fulfill any purpose over there. The currently selected option gets selected later:
https://github.com/primefaces/primefaces/blob/37530a1630edcb8ae32adffb5d4b70497cf4a403/src/main/java/org/primefaces/component/paginator/RowsPerPageDropdownRenderer.java#L101-L103

melloware:
i wonder if we should use a "data" attribute instead of "value".

Rawi01:
Sure, you could move the value to a data attribute if you prefer that. In almost every case you can simply query the value of the select (`$(rpp).val()`) if you need it for some reason.

If you decide to do that you should also change the name to something that describes that value properly e.g. `data-initalRowsPerPage`. The only case I can imagine where this might be usefull is some kind of client side reset button that resets this value but that could be solved on server side way easier.

tandraschko:
can someone other verify if still everything works as expected?

PR #4879, opened 2019-06-11
SonarQube - Optimizing String indexOf() and lastIndexOf() for single char

dvmarcilio:
These methods calls can be more performant when called with a `char` argument instead of a single letter `String`.

Examples: Replacing `String.indexOf("[")` with `String.indexOf(']')`, and `String.lastIndexOf("_")` with `String.lastIndexOf('_')`.

This fixes SonarQube violations of the rule: [String function use should be optimized for single characters](https://rules.sonarsource.com/java/RSPEC-3027)


tandraschko:
Thanks for the PR!
It would be great if you could create a issue for each PR, we use this as tracking

melloware (approved):
Thanks for the PR this is a good change.

dvmarcilio:
Created #4881 for this specific PR.

PR #4880, opened 2019-06-11
SonarQube - Using StringBuilder for concatenating String in for loops

dvmarcilio:
This pull request fixes violations of the SonarQube rule [Strings should not be concatenated using '+' in a loop](https://rules.sonarsource.com/java/RSPEC-1643) that says:

> Strings are immutable objects, so concatenation doesn't simply add the new String to the end of the existing string. Instead, in each loop iteration, the first String is converted to an intermediate object type, the second string is appended, and then the intermediate object is converted back to a String. Further, performance of these intermediate operations degrades as the String gets longer. Therefore, the use of StringBuilder is preferred.

### DIFF
--- a/src/main/java/org/primefaces/expression/impl/RowExpressionResolver.java
+++ b/src/main/java/org/primefaces/expression/impl/RowExpressionResolver.java
@@ -55,9 +55,9 @@ public class RowExpressionResolver implements SearchExpressionResolver, ClientId
 
         int row = validate(context, source, last, expression);
         UIData data = (UIData) last;
-        char seperatorChar = UINamingContainer.getSeparatorChar(context);
+        char separatorChar = UINamingContainer.getSeparatorChar(context);
 
-        String clientIds = "";
+        StringBuilder clientIds = new StringBuilder();
 
         for (UIComponent column : data.getChildren()) {
             // handle dynamic columns
@@ -69,11 +69,18 @@ public class RowExpressionResolver implements SearchExpressionResolver, ClientId
                     for (UIComponent comp : column.getChildren()) {
 
                         if (clientIds.length() > 0) {
-                            clientIds += " ";
+                            clientIds.append(" ");
                         }
 
-                        clientIds += data.getClientId(context) + seperatorChar + row + seperatorChar
-                                + dynamicColumn.getId() + seperatorChar + i + seperatorChar + comp.getId();
+                        clientIds.append(data.getClientId(context));
+                        clientIds.append(separatorChar);
+                        clientIds.append(row);
+                        clientIds.append(separatorChar);
+                        clientIds.append(dynamicColumn.getId());
+                        clientIds.append(separatorChar);
+                        clientIds.append(i);
+                        clientIds.append(separatorChar);
+                        clientIds.append(comp.getId());
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


tandraschko:
AFAIK the JVM optimizes it itself to a StringBuilder?

src/main/java/org/primefaces/expression/impl/RowExpressionResolver.java
```diff
@@ -57,7 +57,7 @@ public String resolveClientIds(FacesContext context, UIComponent source, UICompo
         UIData data = (UIData) last;
         char seperatorChar = UINamingContainer.getSeparatorChar(context);
 
-        String clientIds = "";
+        StringBuilder clientIds = new StringBuilder("");
```
melloware, line 61: Use Constants.EMPTY_STRING instead of "".

src/main/java/org/primefaces/util/CalendarUtils.java
```diff
@@ -138,19 +138,19 @@ public static final String getTimeOnlyValueAsString(FacesContext context, UICale
 
     public static final String getValueAsString(FacesContext context, UICalendar calendar, Object value, String pattern) {
         if (value instanceof List) {
-            String valuesAsString = "";
+            StringBuilder valuesAsString = new StringBuilder("");
```
melloware: Use Constants.EMPTY_STRING instead of "".

melloware (changes requested):

dvmarcilio:
> AFAIK the JVM optimizes it itself to a StringBuilder?

I thought so too. SonarQube still reports it.

melloware:
Don't forget to create and issue and see my code review comments.

tandraschko:
I would leave it as it is

cnsgithub:
> AFAIK the JVM optimizes it itself to a StringBuilder?

Not the JVM but the compiler. :wink: 

However, I am not sure if it just considers simple cases or even string concatenations within loops and other more complex stuff, see https://dzone.com/articles/string-concatenation-performacne-improvement-in-ja.

To be sure, you could write a small program, compile it with different compilers and take a look at the bytecode.

Anyways, I think it's good practice to explicitly use `StringBuilder`.

cnsgithub (approved):

src/main/java/org/primefaces/expression/impl/RowExpressionResolver.java
```diff
@@ -69,27 +70,27 @@ public String resolveClientIds(FacesContext context, UIComponent source, UICompo
                     for (UIComponent comp : column.getChildren()) {
 
                         if (clientIds.length() > 0) {
-                            clientIds += " ";
+                            clientIds.append(" ");
                         }
 
-                        clientIds += data.getClientId(context) + seperatorChar + row + seperatorChar
-                                + dynamicColumn.getId() + seperatorChar + i + seperatorChar + comp.getId();
+                        clientIds.append(data.getClientId(context) + seperatorChar + row + seperatorChar
```
cnsgithub: @dvmarcilio From a compiler perspective it should be okay to mix up `append` and `+` operators, since the latter will most probably get replaced automatically. However, in my opinion I wouldn't mix them anyway.

dvmarcilio: What is the best way in your opinion? one `append()` for each `String`, in the same line? multiple lines?

cnsgithub: Yes, I'd use one `append` per `String`. Regarding the lines: checkstyle will tell you its preference. :stuck_out_tongue_winking_eye:

src/main/java/org/primefaces/util/CalendarUtils.java
Rapster: IMO, you should keep it simple: `StringBuilder valuesAsString = new StringBuilder();` In both case, initial capacity is the same

dvmarcilio:
Addressed @Rapster comments to keep it simple: `StringBuilder valuesAsString = new StringBuilder();` instead of `StringBuilder valuesAsString = new StringBuilder(Constants.EMPTY_STRING);`

Issue #4888, opened 2019-06-11
LinkButton renders style attribute twice causing layout issues


jepsar:
I just ran into this page at the showcase (7.1-SNAPSHOT):

https://www.primefaces.org/showcase/ui/button/linkButton.xhtml

The styling does not seem to be right to me (result is more or less the same for all the free themes):

![Screenshot from 2019-06-11 14-28-38](https://user-images.githubusercontent.com/7500178/59272030-59e80f00-8c55-11e9-996b-301d32c80e2a.png)

The image was taken with Chrome. Firefox also looks wrong (there is a slight difference between it and Chrome).

tandraschko:
a PR would be great

jepsar:
@tandraschko I haven't looked into it yet. I'll have a look tonight.

jepsar:
It is caused by https://github.com/primefaces/primefaces/blob/4551eaa37ca975430bc08b8f7927cf7458165ec9/src/main/java/org/primefaces/component/linkbutton/LinkButtonRenderer.java#L77

This will render the style attribute again (should only be rendered in the outer span).

https://github.com/primefaces/primefaces/blob/36a455f9a0ff32cc77e6a746f7c8b6f3486272c7/src/main/java/org/primefaces/util/HTML.java#L81

In the showcase it will cause an extra 20px right margin in the `a` tag, which causes the styling issue.

PR #4889, opened 2019-06-11
LinkButton renders style attribute twice causing layout issues

jepsar:
This will partly deal with issue #4888 

I've found that there's also an issue with the theme CSS, caused by:

```
body .ui-button.ui-button-icon-only {
  overflow: hidden;
}
```


melloware (approved):

tandraschko:
What about Claas/Styleclass? Will it be rendered only once?

jepsar:
Yes, styleClass was and will be rendered once. It used to be like this (see comment):

https://github.com/primefaces/primefaces/blob/36a455f9a0ff32cc77e6a746f7c8b6f3486272c7/src/main/java/org/primefaces/util/HTML.java#L69-L85

jepsar:
I'll create a new issue to fix the issue with the themes. Those new free themes are not on GitHub, right? I'll just create it here, okay?

PR #4890, opened 2019-06-12
Fix #4882: ColumnToggler duplicate id's.

### LINKED ISSUE #4882: ColumnToggler: Duplicate id
ColumnToggler produces duplicate identifiers. Verified with the current showcase. 

Generated output contains an empty DIV element, e. g.:
`<div id="cars:j_idt703"></div>`
And a div element containing the links:
`<div id="cars:j_idt703" role="dialog" class="ui-columntoggler ui-widget ui-widget-content ui-shadow ui-corner-all" style="z-index: 1004; width: 101.719px; display: none; top: 554.984px; left: 2391.89px;"><a href="#" class="ui-columntoggler-close" aria-label="Close"></div>`

In my opinion the root cause is https://github.com/primefaces/primefaces/blob/dc0a8c69979e2f2aa0e8a334272b2b4485ffa44a/src/main/resources/META-INF/resources/primefaces/columntoggler/columntoggler.js#L46

This script just appends DIV elements instead of operating on DIV created in the ColumnTogglerRenderer

melloware:
The DIV was already created in the Renderer so instead of creating a new one simply find the existing one.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/columntoggler/columntoggler.js
+++ b/src/main/resources/META-INF/resources/primefaces/columntoggler/columntoggler.js
@@ -43,7 +43,7 @@ PrimeFaces.widget.ColumnToggler = PrimeFaces.widget.DeferredWidget.extend({
 
     render: function() {
         this.columns = this.thead.find('> tr > th:not(.ui-static-column)');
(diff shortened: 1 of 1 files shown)


melloware:
tested using the showcase and verified there is no duplicate ID and everything is still working properly.

PR #4906, opened 2019-06-18
Fix #4329 - Refactor PrimeFaces#clearTableState

### LINKED ISSUE #4329: Refactor PrimeFaces#clearTableState
Quoting myself on Gitter:

> I've been using multiviewstate feature on DataTable lately, and I kinda don't like the signature of this method: PrimeFaces#clearTableState. First, name of the parameter is "clientId" which makes me suppose that it expects the clientid of a datatable, but it's wrong. The key set on a table state is actually the combination of a viewId and a clientId (e.g pages/record/search/recordListSearchResult.xhtml_table_form:results:table). And this combination is not a simple concatenation, see DataTable#getTableState
> 
> ```java
>         String viewId = fc.getViewRoot().getViewId().replaceFirst("^/*", "");
>         String stateKey = viewId + "_" + this.getClientId(fc);
> ```
> 
> I would change the signature by clearTableState(String viewId, String clientId), and the impl would do the fiddling. Today I have to make it by myself which is not good thing

@tandraschko coulld we do this on 6.3?

Rapster:


src/main/java/org/primefaces/PrimeFaces.java
```diff
@@ -176,39 +179,51 @@ public void resetInputs(String... expressions) {
...
+     *
+     * @deprecated Use {@link MultiViewState#clearMultiViewState(String, String)} instead
      */
+    @Deprecated
     public void clearDataListState(String key) {
-        Map<String, Object> sessionMap = getFacesContext().getExternalContext().getSessionMap();
-        Map<String, TableState> dtState = (Map) sessionMap.get(Constants.DATALIST_STATE);
-        if (dtState != null) {
-            dtState.remove(key);
+        Pattern p = Pattern.compile("(.*\\.(?:xhtml|jsf|jsp))_(.*)");
```
melloware, line 219: I like to move Patterns to private static final constant so that way the pattern is compiled only once which is the expensive part.

src/main/java/org/primefaces/PrimeFaces.java
```diff
@@ -359,4 +383,112 @@ public void update(String... expressions) {
...
+            Set<String> states = multiViewStates.keySet().stream()
+                    .filter(s -> s.startsWith(stateKey))
+                    .collect(Collectors.toSet());
+            if (!states.isEmpty()) {
+                multiViewStates.keySet().removeAll(states);
+
+                if (clientIdConsumer != null) {
+                    states.stream()
+                            .map(s -> s.replace(stateKey + SEPARATOR, ""))
+                            .forEach(clientIdConsumer);
```
melloware, line 425: Use Constants.EMPTY_STRING instead of "".

melloware (changes requested):

src/main/java/org/primefaces/PrimeFaces.java
Rapster: I like to do that too, but I can't declare a static variable in inner class (though the compiler seems to be if it's a String, I didn't find an explanation for that). Since this method is deprecated and should be removed in the future, I didn't want to spend too much time on it

src/main/java/org/primefaces/PrimeFaces.java
Rapster: Good point

src/main/java/org/primefaces/PrimeFaces.java
melloware: No problem!  Approving PR.

melloware (approved):

PR #4911, opened 2019-06-20
Fix #4305 - Improve FileUpload API

### LINKED ISSUE #4305: Improve FileUpload API
After fixing #4038 it brings some design flaws. By definition UploadedFile is either simple or multiple but can't be both. This is what the current design is implying. It's quite confusing... See also @cnsgithub coment: https://github.com/primefaces/primefaces/issues/4038#issuecomment-438169951

> Depending on the CTOR being called we either have the single vs. multiple file upload situation. Since there are two separate field members, part and parts, the other will stay null and therefore is predestinated for NPEs thrown later on. E.g. methods getSize or getInputstream are suitable for the single file upload case only.

A proper design might imply some breaking changes of the current model. We have to make a clear distinction between simple and multiple uploaded file.

I had a quick thought about it:
image

Rapster:


melloware:
I like the changes but this is definitely a pretty significant change for end users. So now when users get an UploadFile in their FileUploadEvent instead of being able to get the Name and Stream etc they will then have to cast or detect that its a Single vs Multiple file?

Can you show an example of what the new Upload receiving code for an end user would look like?

Rapster:
First, regarding `FileUploadEvent`, as stated in the doc:

> Note that multiple mode is for selection only, it does not send all files in one request. FileUpload component always uses a new request for each file.

So instead of returning an `UploadedFile` it returns a `SingleUploadedFile`, they'll have a compiler error and it's easy to fix. That's the biggest change in my opinion, instead of using `UploadedFile`, they'll have to replace it by `SingleUploadedFile`

Second, `MultipleUploadedFile` is used in the newly feature where multiple files can be uploaded using simple mode of FileUploaded. In 95% cases, the user know in advance whether or not multiple files are expected, either they can straightaway cast `UploadedFileWrapper#getWrapped` or used `FileUploadUtils#getUploadedFile`. In case where fileupload mode is set dynamically, yes they'll have to manually check the `UploadedFile` type (either single or multiple)

I'm fully aware of the breaking changes, but I don't think it's very major as the compiler will alert end users

melloware:
Perfect I am happy with a compiler alert I just didn't want it to be something that had to be casted.

Rapster:
I added `FileUploadUtils#consume` to hide the casting complexity

Rapster:
Another thought, FileUploadUtils needs some brush up, I really don't like the fact that the component is passed as a parameter

tandraschko:
Will review next month

Rapster:
To limit the breaking changes, I can keep UploadedFile (known as SingleUploadedFile in this PR) and rename the top interface (UploadedFile) with some other name?

tandraschko:
Shouldn't we rename DefaultUploadedFile into DefaultSingleUploadedFile?
Also FileUploadUtils#consume isn't called?

Rapster:
Indeed, I should rename into `DefaultSingleUploadedFile`. `FileUploadUtils#consume` is an utility for end users to avoid to make themselves explicit casting. Basically, whatever what the input is, at the end you know you're working with `SingleUploadedFile`

Rapster:
Fix `DefaultSingleUploadedFile`, and also since I refactored few things already, I corrected method names of `UploadedFile` into `getInputStream` _(instead of `getInputstream`)_ and `getContent` _(instead of `getContents`)_

tandraschko:
Ok, cool!
If #consume is for public usage, please also add some javadoc. Then +1 from my side.

I will talk with mert then.

Rapster:
Javadoc fixed

Rapster:
Should I merge?

tandraschko:
+1
could you please rename isValidFile with the List<files> to something like "allFilesValid"?

PR #4916, opened 2019-06-20
Fix #3815 - OutputPanel: New View-ID is created when updating a form …

### LINKED ISSUE #3815: OutputPanel: New View-ID is created when updating a form that contains a p:outputpanel in deffered mode.
New javax.faces.ViewState is created after updating @form in deffered load of the outputpanel

## 1) Environment
   Primefaces Community Version 6.2
   Glassfish 3.2 & Payara 4

## 2) Expected behavior
 The behavior of updating h:form or p:outputpanel should be same.

## 3) Actual behavior
 When updating h:form the deferred request returns with a new view-id.

## 4) Steps to reproduce
  Click first time on button1 to initialize the viewscopeBean.
  The second click works fine without re-initialization of viewscopeBean but a new view-id is applied.
  On the third click, the deferred loading of the panel initializes the Postconstruct method again.

  This cannot be reproduced with button2 that updates only the p:outputpanel inside the form.

## 5) Sample XHTML

```
    <h:form id="form1" >
      <p:outputPanel id="test">
        <p:commandButton id="button1" value="update form" process="@this" update="form1" actionListener="#{viewscopeBean.action()}"/>
        <p:commandButton id="button2" value="update panel" process="@this" update="test" actionListener="#{viewscopeBean.action()}"/>
        <p:outputPanel id="filterButtons" deferred="true">        
        </p:outputPanel>
      </p:outputPanel>
    </h:form>
```

## 6) Sample bean

.. @ManagedBean
  @ViewScoped
  public class ViewscopeBean implements Serializable {

    @PostConstruct
    public void initValues() {
    }

    public void action() {
    }
  }

Rapster:


melloware (approved):

tandraschko:
deffered is now always "true" when rendering the script? seems wrong for me

Rapster:
I think it was already the case, `OutputPanel#encodeScript` is called only if deferred

tandraschko:
ahh, yep

PR #4920, opened 2019-06-21
Fix #4296 - Move component CSS classes into Renderers

### LINKED ISSUE #4296: Move component CSS classes into Renderers
99% of the CSS classes are used by renderers, and since these are responsible of rendering, make more sense that CSS classes are defined in renderers and not in component

Rapster:


melloware (approved):
Nice changes, this was a lot of work!

Rapster:
Nah, I did that bits by bits :wink: Though, don't hesitate to check on whatever components inheriting from BaseMenuRenderer or OutcomeTargetRenderer, that was the trickiest part of the migration

melloware:
Looks like you have a merge conflict now with DataTableRenderer.

src/main/java/org/primefaces/component/autocomplete/AutoCompleteRenderer.java
```diff
@@ -44,6 +44,28 @@
 
 public class AutoCompleteRenderer extends InputRenderer {
 
+    public static final String COMPONENT_TYPE = "org.primefaces.component.AutoComplete";
```
Rapster: Note for myself: COMPONENT_TYPE should stay in AutoComplete class

Rapster:
@mertsincan when do you think we could merge this into master?

Rapster:
@mertsincan could you please tell me when do you think we could merge this into master, before the merge of this PR turns into a nightmare?

Rapster:
@mertsincan ?

Rapster:
@mertsincan ?

Rapster:
@tandraschko I thought Mert could have an opinion on this, but never heard from him, so I'm trying my luck with you :sweat_smile: you might have an idea on when I could merge this

tandraschko:
I will need to talk with him. Also got no response for this topic. Seems he has a good workload currently.

I'm also on holiday this week, i will try to check the AbstractMenu/DialogReturn stuff next week.

mertsincan:
Hi,
> @mertsincan could you please tell me when do you think we could merge this into master, before the merge of this PR turns into a nightmare?
- @Rapster, Thanks a lot for your great effort! But, this is a very big PR. It is very difficult for us to keep track of what has changed. Therefore, it would be better for PrimeTek to make this change and frankly, PrimeTek doesn't like to apply such big PRs from the community.

Sorry for the delayed response!

Thanks again, @Rapster ;)

Rapster:
I don't see the difference between from community or PrimeTek since the final result should be the same (it's just moving variable from one class to another). No matter how big the PR is, as long the changes are relevant, it should be done (and PrimeTek approved PR much bigger than that in the past)

tandraschko:
will close it, based on the decision of mert
it's just cosmetics, so i dont have a problem with it

PR #4924, opened 2019-06-23
Fix #4893 LinkButton icon only variant has incorrect theme styling

### LINKED ISSUE #4893: LinkButton icon only variant has incorrect theme styling
See #4888 and #4889
I think it's caused by the following rule in the theme CSS:

```
body .ui-button.ui-button-icon-only {
  overflow: hidden;
}
```

jepsar:
Fix #4893 LinkButton icon only variant has incorrect theme styling


tandraschko:
We cant apply this, we need to do it in the sass files.
@mertsincan can you do that?

mertsincan:
Hi,
Unfortunately, we can not remove overflow:hidden rule from all buttons that have ui-button-icon-only class. It causes FireFox issue on focus. Also, with your fix;
image
I'll fix it now.

Thanks a lot for the PR!

mertsincan:
Please try this css;
```
body .ui-linkbutton.ui-button-icon-only {
  overflow: visible; }
  body .ui-linkbutton.ui-button-icon-only .ui-button-text {
    -webkit-transform: scale(0);
    -moz-transform: scale(0);
    -o-transform: scale(0);
    -ms-transform: scale(0);
    transform: scale(0); }
```

Issue #4847, opened 2019-05-29
DataTable: improve support for filterOptions in dynamicColumns


matinh:
See forum post https://forum.primefaces.org/viewtopic.php?f=3&t=59011&p=175510 for details.

In short: <p:dataTable> with <p:columns> inside has support for filterOptions, but it is not possible to have columns with and without filterOptions.

matinh:
I'll submit a PR, once #4846 is accepted and closed.

PR #4938, opened 2019-06-25
Support for  #4847:  DataTable: improve support for filterOptions in dynamicColumns

matinh:
This PR adds support for filterOptions EL that evaluates to null or empty array, as requested by #4847.

### DIFF
--- a/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
+++ b/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
@@ -889,12 +889,18 @@ public class DataTableRenderer extends DataRenderer {
     protected void encodeFilterInput(UIColumn column, ResponseWriter writer, boolean disableTabbing,
         String filterId, String filterStyleClass, Object filterValue, String ariaLabelId) throws IOException {
 
-        if (column.getValueExpression(Column.PropertyKeys.filterOptions.toString()) == null) {
+        if (hasFilterOptions(column)) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
```diff
@@ -889,14 +889,20 @@ protected void encodeDefaultFilter(FacesContext context, DataTable table, UIColu
...
         else {
             encodeFilterInputSelect(column, writer, disableTabbing, filterId, filterStyleClass, filterValue, ariaLabelId);
         }
     }
 
+    private boolean hasFilterOptions(UIColumn column) {
+        return column.getValueExpression(Column.PropertyKeys.filterOptions.toString()) == null
+            || getFilterOptions(column) == null
+            || getFilterOptions(column).length == 0;
+    }
```
melloware, line 908: My only comment is you could assign `getFilterOptions(column)` in a variable and use that variable twice.

melloware (changes requested):

src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
matinh: Your're right. And in addition, the method name `hasFilterOptions()` was misleading as it was actually checking if no filterOptions were present (logic was correct, though). I'll update the PR.

matinh: Done, PR has been updated.

PR #4943, opened 2019-06-26
Add files via upload

wdjjinan:
fixed bug: first to use column filter then click page size=All, it will throw javax.servlet.ServletException: Unsupported rows per page value: xxx


melloware:
I thin your PR messed up line endings the entire file is showing as changed.  We only want to see the bare minimum changes made.

Issue #4647, opened 2019-03-18
DataTable: fix "All" pagesPerRow validation


tandraschko:
See: https://github.com/primefaces/primefaces/issues/4636

melloware:
Related to? https://github.com/primefaces/primefaces/issues/3228

tomlux:
commented a possible workaround here:
 https://github.com/primefaces/primefaces/issues/4636#issuecomment-498144063

PR #4944, opened 2019-06-26
fixed bug: Unsupported rows per page value: xxx

wdjjinan:
fixed bug: first to use column filter then click page size=All, it will throw javax.servlet.ServletException: Unsupported rows per page value: xxx


tandraschko:
Please See: https://github.com/primefaces/primefaces/issues/4647

We need to do a better fix probably

tandraschko:
will commit it now, please see #4647

PR #4947, opened 2019-06-27
Fix #3563 - DefaultMenuItem: use builder pattern

### LINKED ISSUE #3563: DefaultMenuItem: use builder pattern

Rapster:
- Use builder pattern for every pojos
- Remove `public` keyword in interfaces
- Remove `MenuModel#addElement` since not used internally (instead do `MenuModel.getElements.add()`)
- Default impl for `MenuGroup#getElementsCount()`
- Removal of `DynamicMenuModel`


tandraschko:
i dont like to remove methods (getElementsCount) and deprecate the constructor as we should not force someone to use the builder. I would just see it as optional.

src/main/java/org/primefaces/model/menu/DefaultMenuColumn.java
```diff
@@ -37,6 +37,9 @@
     private List<MenuElement> elements;
     private boolean rendered = true;
 
+    /**
+     * @deprecated use DefaultMenuColumn#Builder instead
+     */
```
melloware: So by marking deprecated you are forcing people to the builder pattern?  Lombok allows you to use Fluent builder or regular setters and getters depending on dev preference so why not just leave them in non-deprecated?

Rapster:
- `MenuGroup#getElementsCount()` is not removed, has a default impl though
- I think it's good practise to minimize entry points, allowing constructor and builder results into inconsistent code.

melloware (approved):

melloware:
> * `MenuGroup#getElementsCount()` is not removed, has a default impl though
> * I think it's good practise to minimize entry points, allowing constructor and builder results into inconsistent code.

OK I can live with that.  Less code.  Although if we are thinking about making more things Builder Patternable I really love Lombok.  Allows you to use both and minimizes the code we are actually writing!

src/main/java/org/primefaces/model/menu/DefaultMenuColumn.java
Rapster: First, because I believe what is written in Effective Java is right and second, forcing does not mean "no using constructor is bad", it means "whatever you want to dp, you might want to you use the builder pattern :wink:"

Though, if by any chance building a default `DefaultMenuItem` (without any initialization) would make sense, I would remove the `@Deprecated` on no args constructor.

melloware: Yep no problem.  The only thought of leaving the no arg constructor is if people are somehow producing and injecting these using CDI it will require a no arg constructor to be proxied.

Rapster:
Lombok has already been part of a discussion somewhere here. And IMO, using such library (which is basically a hack of the compiler) would probably make sense during application dev., but for a library like PF, I wouldn't take that risk. Builder and other patterns are easy to generated with IDE, fair enough, Lombok is a lot more easier but it's not complicated at all

src/main/java/org/primefaces/model/menu/DefaultMenuColumn.java
Rapster: That would be a very strange way to create menus, but yes that's true

melloware:
Agreed I didn't see the value in Lombok at all.  But then entire Spring Boot community embraced it and I have been doing a lot of Spring Boot lately and I was surprised how quickly I got used to using it!  But yes definitely not necessary.

tandraschko:
> 
>     * `MenuGroup#getElementsCount()` is not removed, has a default impl though
> 

all right, fine

> 
>     * I think it's good practise to minimize entry points, allowing constructor and builder results into inconsistent code.
> 

everything is matter of taste ;) also, as melloware said, we can never remove it because serialization. So if you remove the @deprecated, i'm fine with the changes.

tandraschko:
Yup, +1 for "without lombok" for now

Rapster:
I'm okay removing `@Deprecated` on no args constructor

Rapster:
Also Thomas, something I already discussed on gitter. I think `MenuItem#getFunction` should be renamed into `MenuItem#getAction` using a `Supplier`, not a function. Using a Function, it's like you were defining a method: `MenuItem#getFunction(MenuItem item)` in `MenuItem` which does not really make sense in this context right?

In a functional programmatic way, you'd do: `item.setAction(() -> doSth(item));` if item was really needed.

I think it's a very good initiative to be able to workaround `MethodExpression` that way

tandraschko:
I used Function for easier access because it's not possible to use "this" and "item" would need to be final, right?

I thought about the name as well but i wasn't sure as "action" is kind of "reserved" for a MethodExpression.
For me it's ok to rename.

Rapster:
> I used Function for easier access because it's not possible to use "this" and "item" would need to be final, right?

Yes indeed, `final` becomes required

> I thought about the name as well but i wasn't sure as "action" is kind of "reserved" for a MethodExpression.

Indeed, `UICommand#getAction` exists already :/

tandraschko:
Lets discuss this later/different issue

+1 for renaming into action and maybe command into actionExpression

+0 for the supplier, im not a final Lover 😅

Rapster:
Got it

tandraschko:
@Rapster 
talked with mert and PrimeTek doesn't like to deprecate the old constructors. Will revert those @Deprecated annotations

tandraschko:
@Rapster 
JFYI
i tried to rename function to action but its not possible as UIMenuItemBase extends from UICommand which already has a MethodExpression getAction.

If we would like to rename, we would need to search another name, maybe something like:
outcome -> actionAsExpression
function -> actionAsFunction?

Not sure....

Rapster:
Yeah I know this is what I told you in https://github.com/primefaces/primefaces/pull/4947#issuecomment-506387032

Regarding the name, `getFunction` might be our best bet...

erickdeoliveiraleal:
I want to try 8.0 and OmegaMenu and it stopped working with 8.0 because the code below: buildAjaxRequest(context, (AjaxSource) menuitem, form)

How can I change it to work with 8.0?

Rapster:
First, how is it related to this ticket? Second, your statement is pretty vague, could you elaborate (showing the error message is a good start :wink:) in another issue please.

FYI, community does not have access to Premium themes code

tandraschko:
Just checked OmegaMenu + OmegaMenuRenderer and it compiles fine with 8.0.

Does it throw a Exception? if yes, please post the stacktrace.

erickdeoliveiraleal:
Em qua., 4 de dez. de 2019 às 11:29, Thomas Andraschko <
notifications@github.com> escreveu:

erickdeoliveiraleal:
Line 222

Imagem 3

tandraschko:
Which class?
It would be really great if you could post all possible informations on the first post next time ;)

erickdeoliveiraleal:
OmegaMenuRenderer

Rapster:
Or on a new post...

From MenuItemAwareRenderer:

```java
 String command;
            if (menuitem.isDynamic()) {
                String menuClientId = source.getClientId(context);
                Map<String, List<String>> params = menuitem.getParams();
                if (params == null) {
                    params = new LinkedHashMap<>();
                }
                List<String> idParams = Collections.singletonList(menuitem.getId());
                params.put(menuClientId + "_menuid", idParams);

                command = menuitem.isAjax()
                        ? buildAjaxRequest(context, source, (AjaxSource) menuitem, form, params)
                        : buildNonAjaxRequest(context, source, form, menuClientId, params, true);
            }
            else {
                command = menuitem.isAjax()
                        ? buildAjaxRequest(context, (UIComponent & AjaxSource) menuitem, form)
                        : buildNonAjaxRequest(context, ((UIComponent) menuitem), form, ((UIComponent) menuitem).getClientId(context), true);
            }
```

Please adapt

tandraschko:
The OmegaMenuRenderer in our SVN master has a bit different code:

```java

                String command;
                if(menuitem.isDynamic()) {
                    String menuClientId = menu.getClientId(context);
                    Map<String,List<String>> params = menuitem.getParams();
                    if(params == null) {
                        params = new LinkedHashMap<String, List<String>>();
                    }
                    List<String> idParams = new ArrayList<String>();
                    idParams.add(menuitem.getId());
                    params.put(menuClientId + "_menuid", idParams);

                    command = menuitem.isAjax() ? createAjaxRequest(context, menu, (AjaxSource) menuitem, form, params) : buildNonAjaxRequest(context, menu, form, menuClientId, params, true);
                } 
                else {
                    command = menuitem.isAjax() ? createAjaxRequest(context, (AjaxSource) menuitem, form) : buildNonAjaxRequest(context, ((UIComponent) menuitem), form, ((UIComponent) menuitem).getClientId(context), true);
                }
```

tandraschko:
And create createAjaxRequest is defined in the OmegaMenuRenderer.
Just try to update to the newest version.

erickdeoliveiraleal:
thank you!

Rapster:
@tandraschko You have access to SVN premium themes code?

tandraschko:
Yep

PR #4416, opened 2019-01-11
DataExporter: Enhance DataExporter to support other components #4414

ravihariharan:
Creating a PR for discussion for support for DataExporter for use with other components

`Exporter<TABLE extends UIComponent>`
i could only extends it to UIComponent  since UIData would have excluded all UITree components. 

Open to review and suggestion to make this PR better.

Issue #4414, opened 2019-01-11
DataExporter: Enhance DataExporter to support Other Data Components


ravihariharan:
PrimeFaces version: 6.2
Does it work on the newest released PrimeFaces version? Version? NA
Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) NA
Application server + version: NA
Affected browsers: NA

Once DataExporter: Support custom Export class #4397 is in place we would be able to support other Data Components.

Primefaces by default provide out of box support for Excel , CVS , XML for DataTable.

A minor code refactor will allow user to support custom implementation of export for any component.  
The below is suggested approach considering TreeTable.

1.  Refactor and Rename Exporter to DataTableExporter
2. Create a new interface called Exporter

```
public interface Exporter<TABLE> {
    void export(FacesContext facesContext, TABLE table, String outputFileName, boolean pageOnly, boolean selectionOnly,
           String encodingType, MethodExpression preProcessor, MethodExpression postProcessor, ExporterOptions options,
           MethodExpression onTableRender) throws IOException;

    void export(FacesContext facesContext, List<String> clientIds, String outputFileName, boolean pageOnly,
            boolean selectionOnly, String encodingType, MethodExpression preProcessor, MethodExpression postProcessor,
            ExporterOptions options, MethodExpression onTableRender) throws IOException;

    void export(FacesContext facesContext, String outputFileName, List<TABLE> tables, boolean pageOnly,
            boolean selectionOnly, String encodingType, MethodExpression preProcessor, MethodExpression postProcessor,
            ExporterOptions options, MethodExpression onTableRender) throws IOException;

    String getSheetName(FacesContext context, TABLE table);
}
```

3. DataTableExporter implements Exporter

`public abstract class DataTableExporter implements Exporter<DataTable>`

4. A developer could then create a TreeTableExporter  Exporter<TreeTable>.

`public abstract class TreeTableExporter implements Exporter<TreeTable>`

Rapster:
> 4. A developer could then create a TreeTableExporter Exporter.

But how the user would reuse it? So far `ExporterFactory#getExporterForType` is the decision maker

ravihariharan:
![image](https://user-images.githubusercontent.com/15189434/60360475-cf6c1300-99a9-11e9-90a9-685378e6366f.png)

Rapster:
[...]

PR #4949, opened 2019-06-28
DataExporter: Enhance DataExporter to support other components #4414

ravihariharan:
DataExporter: Enhance DataExporter to support other components #4414 

Closed Old PR #4416


Rapster:
Please don't change the format

ravihariharan:
@Rapster  which format ? i cherry picked the commit and merge the code in a new branch. I also had to fix the License issue.

Rapster:
See `Exporter#export()` parameters were previously well formatted

Rapster:
I guess exporters should be renamed then, for instance `ExcelExporter` -> `ExcelDataTableExporter`?

ravihariharan:
> I guess exporters should be renamed then, for instance `ExcelExporter` -> `ExcelDataTableExporter`?

The current changes are backward compatible. 

Since ExcelExporter and other were not final class i did not rename them.  
An example of breaking change would be extending ExcelExporter  and using a customExporter.
https://github.com/primefaces/primefaces/issues/4397

Rapster:
Sorry I don't understand your point

ravihariharan:
@Rapster  I made only the minimum changes that were required to support extending customExporter to be used with other components.  

We can rename all the class or 
We can move the DataTable specific classes to subpackage of .export.datatable
image

Lets see what @tandraschko / @melloware / @mertsincan   has to say about rename or moving these classes.

Rapster:
+1 for creating the appropriate package, but I don't think all of them need to move in that new package. Few of these classes can be reused no matter what the component is

ravihariharan:
> +1 for creating the appropriate package, but I don't think all of them need to move in that new package

All the class in the sub-package are specific to DataTable. (moved *Options class back)
image

ExporterFactory can be renamed to DataTableExporterFactory.

Rapster:
Why callbacks and factory should be different from the rest? (This is a trick question btw :yum: )

ravihariharan:
> Why callbacks and factory should be different from the rest?



image


We will have to remove <T extends UIComponent> from Exporter.
Will have to change the methods to have UIComponent  and cast when required.

Let me make few changes for factory  and in another branch and confirm.

Rapster:
Well, nothing prevents you to create a `PDFExporter` interface, your "PDFDataTableExporter" implements it and that should solve your problem no?

melloware:
I like all these changes.  And it will make future flexibility better.

tandraschko:
+1 for the sub-package
and we should also name e.g. XMLExporter -> XMLDataTableExporter

ravihariharan:
There are few discussion in https://github.com/primefaces/primefaces/issues/4414

@Rapster  approach was
Keep in mind, that my objective is to be able to reuse these exporters by doing so:
<p:dataExporter type="xls" target="tbl" fileName="cars" pageOnly="true"/>


I was going in the direction of using customExporter for other components. 

e.g. if we want to export a quickbooks format then we could do something like
<p:dataExporter type="quickbooks" customExporter="#{quickbooksExporter}" />
type="quickbooks" is only added for readability


i did not plan for a generic export for treetable , but allow developer to use the customExporter to implement the needed feature.

ravihariharan:
image


CSVExporter ,  ExcelExporter , PDFExporter are now interface that are referred with corresponding VisitCallback

ravihariharan:
@tandraschko / @melloware  / @Rapster i have made all the changes as discussed above. Please review and let me know if we can improve something.

1 task for future would be to create a Exporter factory.

melloware:
This looks great to me.

Rapster:
Since a package `datatable` has been created, I would have probably keep the original name. It's a lot of datatable-something in the datatable package if you know what I mean. Otherwise looks good

ravihariharan:
This is what i would like to do @Rapster 
1. Move the datatable specific to org.primefaces.component.**datatable.export** package.
2. Rename the files back

Only reference in org.primefaces.component.export.DataExporter will be to the DataTableExporterFactory

image


i have the changes in another branch and will merge on confirmation :)

Rapster:
I like your second suggestion in terms of packages. Indeed we can't have CSVExporter twice, and we can't use fully qualified class name... Let's keep `DataTable` as a prefix

src/main/java/org/primefaces/component/export/datatable/DataTableExporterBase.java
```diff
@@ -0,0 +1,333 @@
...
+import javax.faces.component.html.HtmlGraphicImage;
+import javax.faces.context.FacesContext;
+import javax.faces.convert.Converter;
+
+import org.primefaces.component.celleditor.CellEditor;
+import org.primefaces.component.datatable.DataTable;
+import org.primefaces.component.overlaypanel.OverlayPanel;
+import org.primefaces.util.ComponentUtils;
+
+public abstract class DataTableExporterBase {
```
Rapster: 1. Make this implements `Exporter<DataTable>` (and remove this from its children)
2. Rename into `DataTableExporter` (xxxBase is for definition of UIComponent)`

src/main/java/org/primefaces/component/export/datatable/DataTableCSVExporter.java
```diff
@@ -0,0 +1,251 @@
...
+                    case FOOTER:
+                        textValue = col.getExportFooterValue() != null ? col.getExportFooterValue() : col.getFooterText();
+                        break;
+                    default:
+                        textValue = null;
+                        break;
+                }
+
+                if (textValue != null) {
+                    this.addColumnValue(builder, textValue);
```
Rapster: Remove unnecessary `this` references

src/main/java/org/primefaces/component/export/datatable/DataTableExporterBase.java
ravihariharan: Rename into DataTableExporter (xxxBase is for definition of UIComponent) - done

ravihariharan: > Make this implements Exporter<DataTable> (and remove this from its children)

even if we do it , we will need the the children to implement specific interface.
e.g.

`public class DataTablePDFExporter extends DataTableExporter implements PDFExporter<DataTable>`

PDFExporter methods are reference from PDFExportVisitCallback

Rapster: I know but still, DataTableExporter is an exporter (abstract but still an exporter :wink:)

ravihariharan:
@Rapster / @tandraschko  PR ready for merge.

Rapster:
I'm cool with it, do you think by any chance you could provide an example in the showcase?

tandraschko:
@ravihariharan please also update the documentation (with samples how to use it)

ravihariharan:
Migration Guide updated

ravihariharan:
https://github.com/primefaces/showcase-facelift/pull/54

I reused a existing image to use for export for text.

image

PR #4951, opened 2019-06-28
Fix #4647: EOL fixes

### LINKED ISSUE #4647: DataTable: fix "All" pagesPerRow validation

melloware:
@tandraschko Looks like there were some EOL issues so submitting this fix.


Rapster:
Thomas, I think you need to configure your git options (for this repo at least)

PR #4957, opened 2019-07-01
Fix #4954: Remove min-width from DatePicker.

### LINKED ISSUE #4954: DatePicker: annoying min-width
## 1) Environment
- PrimeFaces version: 7.0.4
- Application server + version: WildFly 16.0.0
- Affected browsers: all

## 2) Expected behavior

DatePicker's text input is as flexible as text inputs from other components.

## 3) Actual behavior

DatePicker's text input is forced to have a min-width of 17em by '.p-datepicker' class (components.css.xhtml line 5973). This is way too long for date only (or year-month only) patterns.

## 4) Steps to reproduce

Use DatePicker :)

melloware:


tandraschko:
thats up to @mertsincan. datepicker sources are shared somehow between projects.

mertsincan:
Thanks @melloware ;)

PR #4961, opened 2019-07-02
Fix #4969 - Method TimelineModel#addAllGroups broken

### LINKED ISSUE #4969: Method TimelineModel#addAllGroups broken
TimelineModel#addAllGroups doesn't add anything in TimelineModel#groups attribute

maasvdberg:
fix broken method TimelineModel.addAllGroups, adds to parameter instead of instance variable


melloware:
Please create an issue so we can link this defect to the problem.  Make sure to describe the problem you are having.

Rapster:
See #4969

Issue #3172, opened 2018-01-18
TreeTable: reopens closed nodes


devAnky:
Please see https://stackoverflow.com/questions/48316675/ptreetable-sorting-reopens-closed-nodes-that-were-open-before

## 1) Environment
PrimeFaces-6.2.RC1 on Mojarra-2.3.2 -> PrimeFaces ShowCase!

## 2) Expected behavior
If the user closes nodes and then sorts by a header, the state of the nodes should stay the same.

## 3) Actual behavior
All previous closed nodes will be reopened.

## 4) Steps to reproduce
Go to https://www.primefaces.org/showcase/ui/data/treetable/sort.xhtml
Open and close a node.
Sort by a header.

## 5) Sample XHTML
See https://www.primefaces.org/showcase/ui/data/treetable/sort.xhtml

## 6) Sample bean
See https://www.primefaces.org/showcase/ui/data/treetable/sort.xhtml

The same seems to happen with primefaces.org/showcase/ui/data/treetable/paginator.xhtml
If you open an close a node and then go back and forth on the pages, the previously closed node will be reopened.

EDIT: It seems like adding an collapse event listener like:
<p:ajax event="collapse" 
onstart="someMethod()"
oncomplete="someOtherMethod()" />
Seems to help (the methods in m case show a loading icon)

tandraschko:
I applied your changes but without collapsing all child nodes - thanks!

PR #4975, opened 2019-07-08
#3172: TreeTable: propagate collapse to sever side

maasvdberg:
While **expand** was propagated to the server side TreeNode... **collapse** was not...

Note: now also collapses recursively...


tandraschko:
can someone test it?

tandraschko:
should we also add a expand method?

tandraschko:
Also i'm not sure if we should always fire a expand/collapse request - maybe just if the user defines it via p:ajax?
Also it feels weird that the #queueEvent not collapse the nodes but doesn't expand them in the expand event?

PR #4981, opened 2019-07-10
Fix #4979: refactored columntoggler checkboxes to be accessible

### LINKED ISSUE #4979: ColumnToggler: Inaccessible for screen-readers and voice commands
# Summary
The ColumnToggler component is not correctly read out by screen-readers. It is also not accessible by voice commands of speech recognition software. 
- There is no relation between checkbox and label
  - The hidden `<input>` does not have an `id` attribute
  - The `<label>` does not have a `for` attribute
- The `<div>` that represents the checkbox doesn't have a `role` attribute
- The `<div>` that represents the checkbox doesn't have `aria-checked` attribute

Screen-reader and voice command issue are combined here because the fix for this the same.
I've used NVDA and Dragon for verification.

## 1) Environment
- Does not work with 7.0.4
- Does not work with the showcase
https://www.primefaces.org/showcase/ui/data/datatable/columnToggler.xhtml

## 2) Expected behavior - Screen-Reader
Consider the Year column checkbox of the showcase example
Text read out on focus: 
> Year checkbox checked

Text read out on toggle:
> not checked

Text read out on toggle:
> checked

## 3) Actual behavior - Screen-Reader
Consider the Year column checkbox of the showcase example
Text read out on focus: 
> checkbox checked

Text read out on toggle:
> not checked

Text read out on toggle:
> checked

As one can see you are not able to distinguish between checkboxes when you rely on the screen reader.
## 4) Expected behavior - Voice command
Checkbox is listed when calling a command to interact with checkboxes
> Click checkbox

## 5) Actual behavior -  Voice command
UI Elements are not recognized as checkboxes, therefore nothing happens. You are not able to interact with checkboxes

## 6) Steps to reproduce
- Use a screen reader and listen to the generated output of the ColumnToggler
- Try to use voice commands to access the components checkboxes

## 7) Illustration of fixes considering the generated output
One list element of a ColumnToggler checkbox generates the following output:

Tog1

A fixed version looks like this:
Tog2

MarkusChur:
Here is the fix for https://github.com/primefaces/primefaces/issues/4979

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/columntoggler/columntoggler.js
+++ b/src/main/resources/META-INF/resources/primefaces/columntoggler/columntoggler.js
@@ -59,7 +59,8 @@ PrimeFaces.widget.ColumnToggler = PrimeFaces.widget.DeferredWidget.extend({
             boxClass = hidden ? 'ui-chkbox-box ui-widget ui-corner-all ui-state-default' : 'ui-chkbox-box ui-widget ui-corner-all ui-state-default ui-state-active',
             iconClass = (hidden) ? 'ui-chkbox-icon ui-icon ui-icon-blank' : 'ui-chkbox-icon ui-icon ui-icon-check',
             columnChildren = column.children('.ui-column-title'),
-            columnTitle = columnChildren.text();
+            columnTitle = columnChildren.text(),
+            columnTogglerCheckboxId = this.tableId + "_columnTogglerChkbx" + i;
             
             var label = columnChildren.find('label');
             if (label.length) {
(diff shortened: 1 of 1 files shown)


melloware (approved):
I just checked this out locally and all is working perfectly.  Really nice PR.

MarkusChur:
Hey guys, what's your opinion on this PR? If you agree with the fix, I would like to provide similar fixes for other components affected by the same problem.

melloware:
@MarkusChur I really like this fix  just waiting on @mertsincan to review it.  I think it will get approved.  One place I did not test did you test this in IE11/Edge ?

I would say it is safe to start submitting other PR's for fixes like this.

MarkusChur:
checked that it works with IE11, Edge, and Chrome.

melloware:
I have thoroughly tested this PR so I am going to merge it.

PR #4986, opened 2019-07-12
Fix #1281: Added PrimeFaces.env.getMediaSize() function.

### LINKED ISSUE #1281: Add javascript 'responsiveness' function to read media size
In more than one situation (like https://github.com/primefaces/primefaces/issues/394), but also for  , it is very useful to be able to access the 'media size' in javascript. 

Adding the following script to the new primefaces-core.js.

```
getMediaSize : function() {
    var size;
    if (window.matchMedia('(min-width: 960px) and (max-width:1200px)').matches) {
        /* Intermediate sized desktop */
        size = 5;
    } else if (window.matchMedia('(min-width: 960px)').matches) {
        /* Large desktop */   
       size = 4;
    } else if (window.matchMedia('(min-width: 640px) and (max-width: 960px)').matches) {   
        /* Portrait tablet to landscape <???> and small desktop */
        size = 3
    } else if (window.matchMedia('(min-width:480px) and (max-width: 640px)').matches) {   
        /* Landscape phone to portrait tablet */
        size = 2;
    } else  if (window.matchMedia('(min-width:320px) and (max-width: 480px)').matches) {
        /* Portrait phones and down */
        size = 1;
    } else {
        alert('Are you seriously using this?');
        size = 0;
    }
    return size;
},
```

Keeping this 'in sync' with the css might be 'work' but the advantage imo outweighs that.

It might also be possible have this achieved automagically by settting [css variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables) and [reading those in javascript](https://www.broken-links.com/2014/08/28/css-variables-updating-custom-properties-javascript/)

melloware:


tandraschko:
Why? -1 for unused code in pf

tandraschko:
I will revert it for now. As we long as we don't use it, i dont see any benefit of it and i dont like to have unused SPI in PF.
If you need it in PFE, please add it there and lets talk about it later.

Issue #3542, opened 2018-03-29
Remove CoreRenderer


Rapster:
Lately, I was having a look to CoreRenderer and realized it's more an utility than a Renderer per say. I'll go straight to the point, but what about moving all methods in RendererUtils, and delete CoreRenderer?

> Thomas Andraschko
> +0
> moving to a util makes nothing better
> You would have a SelectRendererUtils instead of SelectRenderer...

> Sébastien Lepage
> Well, first it gives less complexity to the class hierarchy. Second, let's say I'm creating a custom Renderer which already inherits from something, I can't reuse utility method of CoreRenderer, only choice I'll have left is copy/paste. SelectRendererUtils? CheckRendererUtils (which btw is RendererUtils)? I don't know... why not? Or put everything in RendererUtils... Another reason is, when I read a class I expect some of these method to be used as Template Method pattern, but none of these are, which is kind of confusing... it's not when you realized that, these methods are utils and not instance based. Last reason is, on a performance level it might have a difference, since the method is not related to an object but a class, i won't put my money it, but it might be faster. Saying all this, I think it will be better to remove CoreRenderer and move these methods in RendererUtils

> Thomas Andraschko
> For me there is no reason as

> - all of our component renderers inherit from CoreRenderer
> - there is no performance difference
> - we already have soo many utils, i prefer inheritance
> 
> but lets talk later, i currently don't have time to do big refactorings
> maybe in ~2 months
> as i said, i currently don't like that much utils and with the escaping stuff, they get even more

> Sébastien Lepage
> - all of our component renderers inherit from CoreRenderer: IMO, this is actually a design mistake
> - there is no performance difference, it won't make 5x faster sure, but I'm not an export in memory allocation either so I can't really argue
> - we already have soo many utils, i prefer inheritance: I don't think this is a good reason to not doing it, if it makes more sense to be static method so be it. Number of utils shouldn't be a problem as long their responsability are well defined

melloware:
I think I am with @tandraschko on this one. Not enough "bang for the buck" for me to endorse doing this without a good refactoring reason.

Rapster:
I understand what you're saying, but my concern is about code quality, and this might be big. Probably minor to some other things but it deserves some attention, as I said to Thomas, I'm quite curious about performance as well, I have to investigate on this...

tandraschko:
There is no measureable performance difference! ;)

Rapster:
[...]

PR #5003, opened 2019-07-17
Fix #52 - Dialog Framework: DialogReturn event for Menuitem

### LINKED ISSUE #52: Dialog Framework: DialogReturn event for Menuitem
In addition to commandButton and commandLink, also add dialogReturn event for menuitem.

Rapster:
To sum up:
- Fix #52 hopefully 😄 
- Refactoring to avoid copy/paste using default interface
    - `DialogReturnHolder` contains all the common stuff after analysing which component supporting `dialogReturn` event
    - Most of default methods might look useless, but having these just make sense to me as it implements `ClientBehaviorHolder` (FYI, I think `PrimeClientBehaviorHolder `should extend `ClientBehaviorHolder`)

One thing I don't like is the call of `CoreRenderer#decodeBehaviors` in `UIMenuItem#decode`, see my long comment here #3542

Feedbacks are welcome 😉

### DIFF
--- a/src/main/java/org/primefaces/application/DialogReturnHolder.java
+++ b/src/main/java/org/primefaces/application/DialogReturnHolder.java
@@ -65,9 +65,7 @@ public interface DialogReturnHolder extends ClientBehaviorHolder, PrimeClientBeh
         return DEFAULT_EVENT;
     }
 
-    default void handleEvent(FacesEvent event, UIComponent source, Consumer<FacesEvent> queueEvent) {
-        FacesContext context = event.getFacesContext();
-
+    default void handleEvent(FacesEvent event, FacesContext context, UIComponent source, Consumer<FacesEvent> queueEvent) {
         if (event instanceof AjaxBehaviorEvent) {
             Map<String, String> params = context.getExternalContext().getRequestParameterMap();
             String eventName = params.get(Constants.RequestParams.PARTIAL_BEHAVIOR_EVENT_PARAM);
--- a/src/main/java/org/primefaces/application/DialogReturnHolder.java
+++ b/src/main/java/org/primefaces/application/DialogReturnHolder.java
@@ -39,7 +39,7 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.function.Consumer;
 
-public interface DialogReturn extends ClientBehaviorHolder, PrimeClientBehaviorHolder {
+public interface DialogReturnHolder extends ClientBehaviorHolder, PrimeClientBehaviorHolder {
 
     Map<String, Class<? extends BehaviorEvent>> BEHAVIOR_EVENT_MAPPING = MapBuilder.<String, Class<? extends BehaviorEvent>>builder()
             .put("dialogReturn", SelectEvent.class)
@@ -65,8 +65,6 @@ public interface DialogReturn extends ClientBehaviorHolder, PrimeClientBehaviorH
         return DEFAULT_EVENT;
     }
 
-    void queueEvent(FacesEvent event);
-
     default void handleEvent(FacesEvent event, UIComponent source, Consumer<FacesEvent> queueEvent) {
         FacesContext context = event.getFacesContext();
 
@@ -74,7 +72,7 @@ public interface DialogReturn extends ClientBehaviorHolder, PrimeClientBehaviorH
             Map<String, String> params = context.getExternalContext().getRequestParameterMap();
             String eventName = params.get(Constants.RequestParams.PARTIAL_BEHAVIOR_EVENT_PARAM);
 
-            if (eventName.equals("dialogReturn")) {
+            if ("dialogReturn".equals(eventName)) {
                 AjaxBehaviorEvent behaviorEvent = (AjaxBehaviorEvent) event;
                 Map<String, Object> session = context.getExternalContext().getSessionMap();
                 String dcid = params.get(source.getClientId(context) + "_pfdlgcid");
@@ -84,7 +82,7 @@ public interface DialogReturn extends ClientBehaviorHolder, PrimeClientBehaviorH
                 event = new SelectEvent(source, behaviorEvent.getBehavior(), selectedValue);
                 queueEvent.accept(event);
             }
-            else if (eventName.equals("click")) {
+            else if (DEFAULT_EVENT.equals(eventName)) {
                 queueEvent.accept(event);
             }
         }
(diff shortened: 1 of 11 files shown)


src/main/java/org/primefaces/util/MapBuilder.java
```diff
@@ -56,4 +57,8 @@ private MapBuilder(Map<K, V> map) {
     public Map<K, V> build() {
         return map;
     }
+
+    public Map<K, V> ibuild() {
+        return Collections.unmodifiableMap(map);
+    }
```
melloware, line 71: What call it iBuild?  What does the "i" stand for?  Maybe Javadoc this method too?

melloware (changes requested):
Awesome PR I started working on this last week and got stuck in the implementation. This is a very nice fix that cleans up the code!  I made one minor comment.

src/main/java/org/primefaces/util/MapBuilder.java
Rapster: ahah I expected that 😄 It means "immutable". I like to keep things short ><

melloware (approved):

src/main/java/org/primefaces/application/DialogReturnHolder.java
```diff
@@ -0,0 +1,91 @@
...
+public interface DialogReturnHolder extends ClientBehaviorHolder, PrimeClientBehaviorHolder {
+
+    Map<String, Class<? extends BehaviorEvent>> BEHAVIOR_EVENT_MAPPING = MapBuilder.<String, Class<? extends BehaviorEvent>>builder()
+            .put("dialogReturn", SelectEvent.class)
+            .put("click", null)
+            .ibuild();
+
+    Collection<String> EVENT_NAMES = Collections.unmodifiableSet(BEHAVIOR_EVENT_MAPPING.keySet());
+
+    String DEFAULT_EVENT = "click";
```
melloware: Can you use this "click" constant above where the mappings are declared?

melloware (changes requested):

melloware (approved):

melloware:
@Rapster this PR is a huge win as the ticket has 15+ upvotes.  I think you have done a really nice job and I am comfortable with you merging this PR!  I think this is a big win for PrimeFaces.

Rapster:
Thank you @melloware I'm hoping to hear from @tandraschko soon

tandraschko:
really not sure if i like it as DialogReturnHolder really smells for me like a multi-inheritance missuse.
I need to think about, how i would solve it.

Rapster:
Well, if I were in java 7 I would have used a strategy pattern (interface + default impl) and every eligible component would have a `DialogReturnHolder` as attribute, but default method in java 8 just make it easier. Let me know your thoughts

tandraschko:
Still no better idea but what i dont like:
- DialogReturnHolder now acts as "base class" between CommandLinkBase and HtmlCommandLink, with all those events stuff but actually it's used in CommandLink
- the events stuff is owned by the DialogReturnHolder but actually dialogReturn should be just something additional and not replace all event stuff

it's not bad but the relations between all those classes are not clean and DialogReturnHolder should just something additional, not replace something

Rapster:
Mmmh not sure what you're saying... But it does not act as a "base class" but as a default contract (avoiding copying/pasting). It does not replace anything. I think having an interface `DialogReturnHolder` is a nice way to define a behavioral contract (again see pattern strategy).

I don't see which relation you're referring to, the only one that exists is `OutcomeTargetRenderer#encodeOnClick()`

One thing I don't like about my PR is the signature of `DialogReturnHolder#handleQueueEvent`, the consumer attribute is kinda unsual but I don't see how to call `super#queueEvent` otherwise. And the other thing is `CoreRenderer#decodeBehaviors`

But please make your suggestion and talk about this again

melloware:
I think its a nice use of the Default Interface in Java 8.

src/main/java/org/primefaces/application/DialogReturnHolder.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.Collections;
+import java.util.Map;
+import java.util.function.Consumer;
+
+public interface DialogReturnHolder extends ClientBehaviorHolder, PrimeClientBehaviorHolder {
+
+    String DEFAULT_EVENT = "click";
+
+    Map<String, Class<? extends BehaviorEvent>> BEHAVIOR_EVENT_MAPPING = MapBuilder.<String, Class<? extends BehaviorEvent>>builder()
+            .put("dialogReturn", SelectEvent.class)
```
melloware, line 47: Maybe make "dialogReturn" also a constant?  It can be used down below

src/main/java/org/primefaces/renderkit/OutcomeTargetRenderer.java
```diff
@@ -268,6 +271,15 @@ protected void encodeOnClick(FacesContext context, UIComponent source, MenuItem
                 writer.writeAttribute("onclick", onclick, null);
             }
         }
+
+        if (menuitem instanceof DialogReturnHolder) {
+            List<ClientBehaviorContext.Parameter> behaviorParams = new ArrayList<>();
+            behaviorParams.add(new ClientBehaviorContext.Parameter(Constants.CLIENT_BEHAVIOR_RENDERING_MODE, ClientBehaviorRenderingMode.UNOBSTRUSIVE));
+            String dialogReturnBehavior = getEventBehaviors(context, (ClientBehaviorHolder) menuitem, "dialogReturn", behaviorParams);
```
melloware, line 278: Can use the constant here.

melloware (changes requested):

tandraschko:
I created a small change, how it would refactor it (without changes on the menu stuff): https://gist.github.com/tandraschko/26bc2c75af01c959cbb402b47f64f6be

i like it much more as the dialogreturn could be called from any event and is not related to a click event or any other event mapping stuff.

Rapster:
Tough to read your changes, you can make a PR on my branch so I can see clearly the changes

tandraschko:
won't have time the next hours but you can simple apply the patch local

tandraschko:
commited it for now

tandraschko:
Applied your fix for MenuItem. 

Sorry to not apply your change direclty. Hope you don't feel cheated.
Your contribution was great and thanks therefore.
I just had enough time today and i like the way more with a less-abstract way of the DialogReturnHolder as the default event or event mapping does not belong to the single "dialogReturn" event.

PR #5007, opened 2019-07-18
Fix #5006 - Refactor BaseCalendarRenderer#getConvertedValue

### LINKED ISSUE #5006: Refactor BaseCalendarRenderer#getConvertedValue
Method's complexity is way too high. The way of how exception are handled are too cumbersome (I blame java for that...) Using Try monad could help. PR on this way (SPOILER alert: I took a lot on liberty on this)

Feedbacks are welcome 😉 

I tagged it as enhancement but it's also experimental 😋

Rapster:
To sum up:
- "Fix" #5006 I did my best to have the lowest complexity
- For that, I used Try monads from https://github.com/jasongoodwin/better-java-monads and then shade it as it is a lightweight library, and I honestly think it helps a lot in terms of readability

### DIFF
--- a/src/main/java/org/primefaces/component/calendar/BaseCalendarRenderer.java
+++ b/src/main/java/org/primefaces/component/calendar/BaseCalendarRenderer.java
@@ -29,6 +29,7 @@ import org.primefaces.el.ValueExpressionAnalyzer;
 import org.primefaces.renderkit.InputRenderer;
 import org.primefaces.util.CalendarUtils;
 import org.primefaces.util.HTML;
+import org.primefaces.util.LangUtils;
 import org.primefaces.util.MessageFactory;
 
 import javax.el.ValueExpression;
@@ -45,11 +46,16 @@ import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
-import java.time.*;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.YearMonth;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
 import java.time.format.DateTimeParseException;
 import java.time.temporal.ChronoField;
+import java.time.temporal.Temporal;
+import java.util.Collection;
 import java.util.Date;
 import java.util.TimeZone;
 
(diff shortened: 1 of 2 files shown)


christophs78:
Always interesting to look at code written by a experienced developer and learing some new tricks. :-) 
IMHO the refactored code ist way better structured and readable. (Would also allow for (new/better) unit-tests.)
I hope @tandraschko or @melloware have a opinion about using https://github.com/jasongoodwin/better-java-monads. From my perspective i see the benefits. But i´m not the one to decide whether it´s worth the "price" of adding a additional dependency. And different people have different opinions in terms of using lambdas and functional style programming.

tandraschko:
Will review next week

melloware:
My two cents is the code is a huge improvement over the old code however I hate introducing new libraries into PF.  Is there a clean way it can be written without including Monads?

tandraschko:
yup, -1 for dependencies

Rapster:
Thank you all for your comments.

@christophs78 Thank you, but from my experience, I always felt that, when someone is reviewing your code he will always come up with good ideas and you'd be like "Why didn't I think about that?!" but it doesn't mean he would have done a better job (I have no merits in this, you're the original author :wink:), you would have come back on this a month later, and come up with similar ideas, sometimes you just need to take some distance. I'm glad you're keeping an open mind and appreciating the PR

@christophs78 @ @melloware @tandraschko I don't considerer bringing any new dependency in this PR, as I'm shading it :smile: When you'll have PF in your dependencies running in your servlet container, no `better-java-monads.jar` will be deployed so basically it's not a dependency :yum: 

> Is there a clean way it can be written without including Monads?

If you mean:
- Not using `better-java-monads` but okay with using [monad](https://en.wikipedia.org/wiki/Monad_(functional_programming)), yes I could copy all the content from the library and paste in one class (there only 4 of them). But what's the point while I can shade it.
- Using classic try/catch is IMO an ugly instructions, In a few words: makes the code hard to read and hard to write. Read this https://blog.softwaremill.com/exceptions-no-just-try-them-off-497984eb470d and if you can, this: https://medium.com/@johnmcclean/when-functional-try-outperforms-try-catch-c44e83ec7939

You might have a better idea of what Try monad is now. For me, it's a "life" changer, and yet I'm introducing a very lightweight library for PF while others like cyclops or vavr are really great but heavier. If code readability matters to you, you should definitely opt for it (I know sounds a bit like an advertisment :sweat_smile: )

melloware:
@Rapster 1. Looks like you have a conflict you need to rebase?

2. I am not arguing against Monads but to me at the end of the day its all just syntax sugar.  I would argue regular try...catch are more readable than the lamba's.  For me personally when I see lambda's I have to stop and stare at the code to try and figure out what it is doing.  I know I am in the minority so I am not fighting the Java8+Lambda way, its just for libraries like PF is it truly necessary versus writing clean imperative code?  At the end of the day the compiler doesn't care its all getting compiled to byte code.

That just my two cents I am also fine with leaving it and shading it if its truly 4 classes.

Rapster:
Ahah I know that feeling, I'm still struggling sometimes, you'll always be able to find an alternative using lambdas but it doesn't mean it's a good one. I guess you need to get use to it, and for that, you need to read and write to practise. Regarding Try monad, it's more than just "a better way to handle exceptions", look also at the PR and you see how the code is much more concise. 

I'll ignore that conflict it as it doesn't apply anymore since the refactoring.

melloware:
Either way this is a great PR.

tandraschko:
I don't like to introduce new yet-another-random-lib dependency (doesn't matter if shaded or not) - and the last commit was 4 years ago.
I would like to see a refactoring without monads before (i think the biggest improvement is the splitting of the methods) and after that, we can still check if we can use lambdas there or not.

Rapster:
Just for fun I had a try with `MenuItemHolder#handleBroadcast`:

Before:
```java
Object invokeResult = null;
try {
	MethodExpression me = expressionFactory.createMethodExpression(elContext, command, String.class, new Class[0]);
	invokeResult = me.invoke(elContext, null);
}
catch(MethodNotFoundException mnfe1) {
	try {
		MethodExpression me = expressionFactory.createMethodExpression(elContext, command, String.class, new Class[] {ActionEvent.class});
		invokeResult = me.invoke(elContext, new Object[] {event});
	}
	catch(MethodNotFoundException mnfe2) {
		MethodExpression me = expressionFactory.createMethodExpression(elContext, command, String.class, new Class[] {MenuActionEvent.class});
		invokeResult = me.invoke(elContext, new Object[] {event});
	}
}
finally {
	String outcome = (invokeResult != null) ? invokeResult.toString() : null;

	context.getApplication().getNavigationHandler().handleNavigation(context, command, outcome);
}
```

After:
```java
// Alias to ExpressionFactory#createMethodExpression
BiFunction<Class[], Object[], Object> getOutcome = (paramTypes, params)
                        -> ef.createMethodExpression(elContext, command, String.class, paramTypes).invoke(elContext, params);

String outcome = (String)Try.ofFailable(() ->  getOutcome.apply(new Class[0], null))
                    .orElseTry(() -> getOutcome.apply(new Class[]{ActionEvent.class}, new Object[]{event}))
                    .orElseTry(() -> getOutcome.apply(new Class[]{MenuActionEvent.class}, new Object[]{event}))
                    .recover(e -> null);

context.getApplication().getNavigationHandler().handleNavigation(context, command, outcome);
```

Not surprising it's no longer maintened vavr and cyclops are already doing an awesome job, anyway i'll see what I can do...

Rapster:
PR updated

tandraschko:
+1, i like it
should we rename getConvertedValueFromXXX to convertValueToXXX?

tandraschko:
and could we add some unittests?

Rapster:
Naming is fine, regarding unit tests @christophs78 do you wanna give it a try?

christophs78:
@Rapster - Yes, i´ll write some unittest. I´ll provide a extra PR for this, after this PR was merged. Ok?

Rapster:
Sounds good to me

PR #5008, opened 2019-07-19
Fix name of constructor

pzygielo:
It's just in docs.


melloware (changes requested):
Couldn't you just change the name of the class to ChartBean instead of changing the constructors?

pzygielo:
Wouldn't I have to update references of `*bean*.model` in `p:chart`?

melloware:
Ahh I got it.  Looks good.

Issue #4923, opened 2019-06-23
Primefaces-Events (eg SelectEvent): use generics


christophs78:
## 1) Environment
- PrimeFaces version: 7.1

## 2) Expected behavior
`org.primefaces.event.SelectEvent` is currently untyped.
For some usages it would IMO make sense to change it to 
`org.primefaces.event.SelectEvent<T>`.

Rapster:
Do you have an example as why it'd make more sense?

christophs78:
It would be some kind of syntactic sugar. :-)

Old (1):
```
public void onNodeSelect(SelectEvent event) {
   MindmapNode node = (MindmapNode) event.getObject();
```
New (1):
```
public void onNodeSelect(SelectEvent<MindmapNode> event) {
    MindmapNode node = event.getObject();
```
Old (2):
```
public void onDateSelect(SelectEvent event) {
    FacesContext facesContext = FacesContext.getCurrentInstance();
    LocalDate date= (LocalDate)event.getObject();
    DateTimeFormatter formatter =  DateTimeFormatter.ofPattern("dd/MM/yyyy");
    facesContext.addMessage(null, new FacesMessage(FacesMessage.SEVERITY_INFO, "Date Selected", date.format(formatter)));
}
```
New (2):
```
public void onDateSelect(SelectEvent<LocalDate> event) {
    FacesContext facesContext = FacesContext.getCurrentInstance();
    DateTimeFormatter formatter =  DateTimeFormatter.ofPattern("dd/MM/yyyy");
    facesContext.addMessage(null, new FacesMessage(FacesMessage.SEVERITY_INFO, "Date Selected", event.getObject().format(formatter)));
}
```

Old (3):
```
public void onRowSelect(SelectEvent event) {
    FacesMessage msg = new FacesMessage("Car Selected", ((Car) event.getObject()).getId());
    FacesContext.getCurrentInstance().addMessage(null, msg);
}

New (3):
public void onRowSelect(SelectEvent<Car> event) {
    FacesMessage msg = new FacesMessage("Car Selected", event.getObject().getId());
    FacesContext.getCurrentInstance().addMessage(null, msg);
}
```

Rapster:
Ok that's what I thought, I understand what you're saying but I'm not sure if the change would bring any added value (syntaxically speaking). 

Since you have no control over event and listener instanciation, having a generic does not guarantee anything in terms of what the `FacesEvent#getObject` type it will be.

[...]

PR #5012, opened 2019-07-21
#4923: Primefaces-Events (eg SelectEvent): use generics

christophs78:


christophs78:
showcase-PR: https://github.com/primefaces/showcase-facelift/pull/51

tandraschko:
Thanks!

Issue #5016, opened 2019-07-23
SplitButton: lacking margin-right


maasvdberg:
SplitButton: lacking margin-right
<img width="285" alt="buttons" src="https://user-images.githubusercontent.com/910975/61732934-0a7e1e00-ad7f-11e9-9449-aa06e3fc4d91.png">

melloware:
Assigning to @mertsincan to review.

Issue #3496, opened 2018-03-19
SelectoneButton: cumbersome margin-right


Rapster:
## 1) Environment
- PrimeFaces version: 6.2
- Does it work on the newest released PrimeFaces version? Version? No
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) No
- Application server + version:
- Affected browsers: Chrome

## 2) Expected behavior
No margin-right should be set

## 3) Actual behavior
margin-right set with 7px

## 4) Steps to reproduce
It's not visible in the showcase, but you can inspect sources to see it. In my application, it looks like this:
https://www.primefaces.org/showcase/ui/input/oneButton.xhtml
![image](https://user-images.githubusercontent.com/7521311/37609602-c9858262-2b9d-11e8-8995-f5b54870511b.png)

## 5) Sample XHTML

```xml
<p:selectOneButton value="#{selectOneView.option}">
            <f:selectItem itemLabel="Xbox One" itemValue="Xbox One" />
            <f:selectItem itemLabel="PS4" itemValue="PS4" />
            <f:selectItem itemLabel="Wii U" itemValue="Wii U" />
</p:selectOneButton>
```

## 6) Proposal
Remove cumbersome `.ui-buttonset` style, I don't see the reason why some component whould have a margin-right (button does not have it for instance)

PR #5017, opened 2019-07-23
SplitButton: lacking margin-right #5016

maasvdberg:
SplitButton: lacking margin-right #5016


melloware (approved):

tandraschko:
not sure... our inputtext also doesn't have a margin

maasvdberg:
agree... but at least *all* buttons will

Rapster:
-1 for adding a default margin, similar to a ticket I opened a while back #3496

Or, the CSS selector should be smarter, something like if button followed by another widget then add margin

Or up to the end users to define his own style

PR #5022, opened 2019-07-24
#4646: Calendar/Datepicker; additional UnitTests

christophs78:
My first try to add UnitTest´s for Calendar and Datepicker.
Maybe there are further suggestions how to do this better. (@tandraschko, @Rapster, @melloware ?)
I added a few test-scoped dependencies for getting this UnitTest´s done.


tandraschko:
why not just using plain mockito, it's already importet

i really like to avoid multiple testing libs and we agreed for mockito in the past
in myfaces there were 3 mocking libs used over 1800 unittests..., it was a nightmare to refactor it

christophs78:
Because i need to mock a static method. Google-search suggested to do this with PowerMock which is built on top of Mockito. (I do most things with Mockito, except the static method.)

melloware:
Yep PowerMockito is the only way to get at static methods in unit tests.

tandraschko:
cant you mock #createConverterException - to avoid mocking the static method?

Also can we avoid harmcrest? As i said, i like to avoid so many dependencies

christophs78:
I will try tomorrow.
Hamcrest is easy to avoid. (It allows to check the message of the expected exception with startsWith, without Hamcrest I have to compare the whole message.)

christophs78:
I removed the extra hamcrest-dependency. It comes already with JUnit.

christophs78:
Mocking #createConverterException to avoid PowerMock is a bit more complex.
- It implies that we do not test this method any more.
- It would require to mock DatePickerRenderer (the class we test). This has a few further implications. (At least as far is i understand this now.)

tandraschko:
1) yep
2) yep, right - a partial mock/spy isn't bad ;)

christophs78:
Question - tried to add one additional test, which fails....
```
    @Test
    public void getConvertValue_Date() {
        Class type = Date.class;
        setupValues(type, Locale.ENGLISH);
        when(renderer.getConvertedValue(eq(context), eq(calendar), any())).thenCallRealMethod();
        Object object = renderer.getConvertedValue(context, calendar, "7/23/19");
        ...
    }
```

This fails with
```
java.lang.ClassCastException: org.primefaces.component.calendar.Calendar$MockitoMock$54864688 cannot be cast to org.primefaces.component.datepicker.DatePicker
	at org.primefaces.component.datepicker.DatePickerRenderer.getConvertedValue(DatePickerRenderer.java:176)
	at org.primefaces.component.calendar.CalendarAndDatePickerTest.getConvertValue_Date(CalendarAndDatePickerTest.java:267)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	...
```
Is there a  way to solve this?

christophs78:
I´ll try to add some tests for DatePickerRenderer#getConvertedValue.
--> Please wait with merging.

christophs78:
Finished. Ready for merging.
Had to add a @Ingore for one test, because i can´t avoid, that it call´s a static method.
But no matter. We have now way more tests than before.

christophs78:
@tandraschko - Tried to add some tests for DatePicker#validateValue. (Because i want to refactor DatePicker#validateValue and DatePicker#validateDateValue in a further step.)
I have come a long way, but now i fail at the static method-call MessageFactory.getMessage:
```
            if (!isValid()) {
                FacesMessage msg = null;
                String validatorMessage = getValidatorMessage();
                Object[] params = new Object[] {MessageFactory.getLabel(context, this)};
                if (validatorMessage != null) {
                    msg = new FacesMessage(FacesMessage.SEVERITY_ERROR, validatorMessage, validatorMessage);
                }
                else if (isDisabledDate) {
                    msg = MessageFactory.getMessage(DATE_INVALID_MESSAGE_ID, FacesMessage.SEVERITY_ERROR, params);
                }
                else if (!isRangeDatesSequential) {
                    msg = MessageFactory.getMessage(DATE_INVALID_RANGE_MESSAGE_ID, FacesMessage.SEVERITY_ERROR, params);
                }
                else {
                    msg = MessageFactory.getMessage(DATE_OUT_OF_RANGE_MESSAGE_ID, FacesMessage.SEVERITY_ERROR, params);
                }
                context.addMessage(getClientId(context), msg);
            }
```
Can we discuss about Powermock as Test-Only-Dependency?

christophs78:
Stop. :-) I´ll try to first refactor DatePicker#validateValue in a way, that it´s easier to unit-test.

tandraschko:
I would probably make ValidationResult static and the name the fields without "is".
Java beans convention is disabledDate (field) -> isDisabledDate (getter)

christophs78:
Done. :-)

tandraschko:
Thanks!

PR #5026, opened 2019-07-27
Fix #5032: Menu Add properties collision and maxHeight

### LINKED ISSUE #5032: Add properties collision and maxHeight to menu and menuButton
Sometimes I use a layout where the body of the HTML has a fixed height and does not scroll. When I add a menu or menu button with more than a few items, that often results in an overlay menu with items that are off-screen, with no way for the user to scroll to them.

This could be solved by supporting the option `collision` provided by JQueryUI.

To ensure the overlay does not grow taller than the screen, it becomes necessary to set the max height of the overlay menu as well, in addition to changing the overflow property of the widget to show scrollbars when needed.

Sometimes an picture is worth a thousand word. A menu button with the new attributes set might look like this:

```xml
<p:menuButton value="Options" collision="flipfit" maxHeight="90vh">
    <!-- ... ... -->
</p:menuButton>
```

image

Likewise for a menu:

```xml
<p:menu overlay="true" trigger="dynaButton" my="left top" at="left bottom" collision="flipfit" maxHeight="90vh">
    <!-- ... ... -->
</p:menu> 
```

image

blutorange:
Sometimes I use a layout where the body of the HTML has a fixed height and does not scroll. When I add a menu or menu button with more than a few items, that often results in an overlay menu with items that are off-screen, with no way for the user to scroll to them.

This adds an option `collision` to the components `menu` and `menuButton` (same as JQueryUI position.collision). Also adds an option `maxHeight` to ensure the overlay does not grow taller than the screen. I also updated the taglib description:

## menu
* `collision`: Applied only when overlay is set to true. When the overlay menu overflows the window in some direction, move it to an alternative position. Supported values are flip, fit, flipfit and none. See https://api.jqueryui.com/position/ for more details. Defaults to flip. When you the body of your layout does not scroll, you may also want to set the option maxHeight.
* `maxHeight`: The maximum height of the menu. May be either a number (such as 200), which is interpreted as a height in pixels. Alternatively, may also be a CSS length such as 90vh or 10em. Often used when overlay is set to true, but also works when it is set to false. Useful in case the body of your layout does not scroll, especially in combination with the collision property.

## menuButton

* `collision`: For the overlay menu that shows up when the menu button is clicked. When the overlay menu overflows the window in some direction, move it to an alternative position. Supported values are flip, fit, flipfit and none. See https://api.jqueryui.com/position/ for more details. Defaults to flip. When you the body of your layout does not scroll, you may also want to set the option maxHeight.
* `maxHeight`: The maximum height of the overlay menu that shows up when the menu button is clicked. May be either a number (such as 200), which is interpreted as a height in pixels. Alternatively, may also be a CSS length such as 90vh or 10em. Useful in case the body of your layout does not scroll, especially in combination with the collision property.

## Screens

Sometimes an picture is worth a thousand words:

```xml
<p:menuButton value="Options" collision="flipfit" maxHeight="90vh">
    <!-- ... ... -->
</p:menuButton>
```

image

```xml
<p:menu overlay="true" trigger="dynaButton" my="left top" at="left bottom" collision="flipfit" maxHeight="90vh">
    <!-- ... ... -->
</p:menu> 
```

image


melloware:
@blutorange I really like it.  However 2 things...

1. Please create an Issue and link it to this PR.  You can cut and paste your exact details above into the ticket.

2. You need to update the /docs/7.1 folder to add these 2 new properties to the documentation files.

tandraschko:
+1 for the colission

just not sure about the max-height... why not just use css?

blutorange:
@tandraschko My thinking was that (1) just setting max-height is not enough - you also need to change the overflow property, (2) being able to just set the property and have PrimeFaces work out the details  is more convenient than having to add a bunch of CSS property manually. Also (3) it seemed to me it's similar to the `width` / `height` properties of many elements that could also be set via CSS.

But I can also your point, keeping the number of properties low and don't add redundant ones.

I've created an issue (see #5032). Will update the documentation as mentioned by @melloware  once we've decided whether or not to include the `maxHeight` property.

blutorange:
Alright, so I went ahead and update the docs. I could take out the `maxHeight` property if you still feel it should be removed.

melloware:
I am indifferent on max-height so I will leave that up to @tandraschko .

tandraschko:
Can you please use Character.isDigit instead the >0, <9 check?

melloware (approved):

tandraschko:
Thanks!

PR #5031, opened 2019-07-28
#4646 Schedule: add java.time.LocalDateTime-support

christophs78:
This PR replaces java.util.Date with java.time.LocalDateTime for the Schedule-component.
see also https://github.com/primefaces/showcase-facelift/pull/53

Additionaly i added a few documentation-updates for Calendar and Datepicker.

### DIFF
--- a/docs/7_1/components/calendar.md
+++ b/docs/7_1/components/calendar.md
@@ -121,7 +121,7 @@ Value of the calendar should be a java.util.Date.
 ```
 ```java
 public class DateBean {
-    private Date date;
+    private LocalDate date;
     //Getter and Setter
 }
 ```
@@ -206,8 +206,8 @@ _org.primefaces.event.SelectEvent_ instance.
 <p:messages id="msg" />
 ```
 ```java
-public void handleDateSelect(SelectEvent event) {
-    Date date = (Date) event.getObject();
+public void handleDateSelect(SelectEvent<LocalDate> event) {
+    LocalDate date = event.getObject();
     //Add facesmessage
 }
 ```
--- a/docs/7_1/components/calendar.md
+++ b/docs/7_1/components/calendar.md
@@ -39,7 +39,7 @@ ajax selection and more.
 | pattern | MM/dd/yyyy | String | DateFormat pattern for localization
 | locale | null | Object | Locale to be used for labels and conversion.
 | navigator | false | Boolean | Enables month/year navigator
-| timeZone | null | Time Zone | String or a java.util.TimeZone instance to specify the timezone used for date conversion, defaults to TimeZone.getDefault()
+| timeZone | null | Time Zone | String or a java.time.ZoneId instance or a java.util.TimeZone instance to specify the timezone used for date conversion, defaults to TimeZone.getDefault()
 | readonlyInput | false | Boolean | Makes input text of a popup calendar readonly.
 | showButtonPanel | false | Boolean | Visibility of button panel containing today and done buttons.
 | effect | null | String | Effect to use when displaying and showing the popup calendar.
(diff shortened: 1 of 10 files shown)


melloware:
Hmm you removed and deprecated all of the java.util.Date handling?  I thought we were allowing both LocalDate and util.Date to still be viable.

christophs78:
Yes. Schedule internally works now with LocalDateTime. But this is something we should discuss.

In my opinion there a the following options:
- Keep Schedule java.util.Date only.
- Make Schedule java.time.LocalDateTime only.
- Keep some kind of backward-compatibility as i do.

Keeping some kind of backward-compatibilty raises the following questions:
- Do we deprecate the old (java.util.Date) constructors and getter/setter?
- Which type does Schedule (DefaultScheduleEvent) use internally. (Date? LocalDateTime? Both?)

christophs78:
Schedule is a different beast compared to Calendar and Datepicker, because it has its own model. So IMO we need to take a somewhat different approach.

melloware:
Understood.  I have never used Schedule so I trust your judgement here.

tandraschko:
I (personally) would migrate it to DateTime API and add a hint to the migration guide, that a user has to migrate, too.

IMO it would be also much better to remove Date support from Calendar and just proide a converter / force a converter from the user.
The codebase is really bad if we always need to support both Date APIs.

melloware:
Yep I guess as long as there is a DateConverter for backward compatibility it shouldn't be a problem and we could keep the code base clean.

christophs78:
A converter does not help there. You have to put your data into ScheduleModel and add DefaultScheduleEvent-instances to the ScheduleModel.

tandraschko:
Yep, for Schedule it's not possible. I would just migrate to DateTime API and we should realy think about removing Date support from Calendar/DatePicker (converters are supported there, so no problem actually)

christophs78:
Ok, i´ll remove java.util.Date - support from Schedule within the next week.
When i move Timeline to java.time, i´ll do the same. (only support java.time)
Schedule and Timeline are IMO both components that are not used that often.
For Calendar and DatePicker IMO we have to keep java.util.Date - support for the PF 7.x - Releases.
Calendar and DatePicker are used very frequent. So IMO this is a breaking change we only should do with am major release. (eg 8.0)

christophs78:
@tandraschko: Please look at @Rapster latest comment in https://github.com/primefaces/primefaces/issues/4646.
What should we do?
IMO making a breaking change to Schedule and Timeline should not harm to that many people. But i´m also totally fine with keeping Backward-Compatibility in Schedule (and Timeline) as currently suggested in this PR.

christophs78:
Breaking Schedule (and Timeline) would at least result in compile-time-errors.
Breaking Calendar and DatePicker would be much more perfidious because this results in runtime-errors.

tandraschko:
+1 for the breaking chance for this Not often used components

If we can rename attributes for cosmetics, we can also do this (and you fail compile time, not silently)

christophs78:
Ok. I removed java.util.Date - support. Schedule is now java.time.LocalDateTime-only.
I also updated the showcase-PR.
@tandraschko: One of-topic-question: JetBrains offers free IntelliJ-licences for open-source-projects. (https://www.jetbrains.com/community/opensource/) Is there a way to get some of this for PrimeFaces. (Would save some money for the IntelliJ-licencse on my home-PC.)

tandraschko:
Looks like not possible (PrimeFaces is sponsored) - If you would become a PrimeFaces Extensions commiter, it should work

christophs78:
No problem. Just tried to ask before i pay 150 for my home-licence.

christophs78:
@tandraschko: Are we able to make some progress with this PR? (I´d like to continue with Timeline after Schedule is done. And finally i´ll do some cleanup and maybe refactoring (as suggested by @Rapster) with CalendarUtils.)

tandraschko:
i will have time again mid/end next week
i had to start some releases @ apache this week.

tandraschko:
If we only have LocalDate in the models now, we should not name them "getStartLocalDateTime". We should use the same names as before

christophs78:
Ok. I updated this PR and showcase-PR according to your suggestions.

tandraschko:
Thanks!

PR #5034, opened 2019-07-29
Fix #5033: Overlay Panel aria-label for Close.

### LINKED ISSUE #5033: OverlayPanel: Missing aria-label on close icon
## 1) Environment
- Application server + version: **All**
- Affected browsers: **All**

## 2) Expected behavior

The close icon of the overlay panel sets an additional ARIA label (`aria-label="Close"`) to support assistive technology like screenreaders. Thats how it works for other overlay components too e.g. column toggler.

## 3) Actual behavior

Right now there is no label and the screenreader can not explain the user what this close link will do. 

The whole code gets rendered using jQuery here
https://github.com/primefaces/primefaces/blob/938d06d1b680612c55ab9beca3bc6ae128d7bea5/src/main/resources/META-INF/resources/primefaces/overlaypanel/overlaypanel.js#L20-L22

A possible fix for this one would be to add 
`.attr('aria-label', PrimeFaces.getAriaLabel('overlaypanel.CLOSE'))` 
to the construction code (thats how it works for column toggler right now). There also is a duplicate `href` attribute on the link that could be removed too :smile:

melloware:


tandraschko:
+1

PR #5044, opened 2019-08-01
Fix #5035: Menu wrong position when using trigger element

### LINKED ISSUE #5035: Menu: wrong position when using trigger element
## 1) Environment
- PrimeFaces version: 7.0.

An element with `float: right` which is a trigger for the menu. The menu has the following position attributes `my="right bottom" at="right top"`.

## 2) Expected behaviour
Menu positioned by the right side of the element.

## 3) Actual behaviour
The offset position is wrong. The problem in this code
https://github.com/primefaces/primefaces/blob/251ab3b91622d8187ad29b201166019be71350d8/src/main/resources/META-INF/resources/primefaces/menu/menu.base.js#L92-L100

First, it set `display: block` and `visibility: hidden` for the element, and at that point, as the menu container is located outside of content the scrollbar (attached to the `body`) appears (and changes the viewport size). After that, the code calculates the correct location for the element and relocate it, the scrollbar is not needed anymore and it disappears.

As a result, the menu position is not right. We need to add the size of the scroller to offset value of the target element. Or we can calculate position first and then change `display` style.

melloware:


tandraschko:
why removing "this.jq.css('visibility', '');"? it don't think it's there just for fun :D

melloware:
I poked around all the Showcase menu examples and I didn't see anything that was not working.  I was stumped as to why it was doing the visibility;hidden trick to begin with?  Let me put the visibility stuff back and see if it works.

melloware:
OK just pushed a fix.  What I do now is make the behavior like the current menu just changed the order.
Current Behavior:
1. Change CSS
2. Show Menu
3. Align it
4. Remove Visibility

New:
1. Change CSS
2. Align it
3. Show Menu
4. Remove Visibility

PR #5054, opened 2019-08-06
Fix #5053 - Refactor Exporter

### LINKED ISSUE #5053: Core: Refactor Exporter - Part 1
Recently I had a deeper look on newly introduced `Exporter` interface, and to me that kinda of definition is wrong:
```java
public interface Exporter<T extends UIComponent> {

    void export(FacesContext facesContext, T component,
            String outputFileName, boolean pageOnly, boolean selectionOnly,
            String encodingType, MethodExpression preProcessor,
            MethodExpression postProcessor, ExporterOptions options, MethodExpression onTableRender) throws IOException;

    void export(FacesContext facesContext, List<String> clientIds,
            String outputFileName, boolean pageOnly, boolean selectionOnly,
            String encodingType, MethodExpression preProcessor,
            MethodExpression postProcessor, ExporterOptions options, MethodExpression onTableRender) throws IOException;

    void export(FacesContext facesContext,
            String outputFileName, List<T> components, boolean pageOnly, boolean selectionOnly,
            String encodingType, MethodExpression preProcessor,
            MethodExpression postProcessor, ExporterOptions options, MethodExpression onTableRender) throws IOException;

}
```

1. 3 methods that does 99% the same thing
1. methods having 10 parameters

As a result, there is a lot of duplicates (recently sonarcloud complained about it) which makes it even tougher to read the code.

I think there is a need to be able to export a datable inside an iterator component such as `ui:repeat`, the reason why we have these `xxxExportVisitCallback`. But IMO, we should be able to solve this more simply.

PR on his way

Rapster:
To sum up:

- Clean/Refactoring code (less duplications and simpler code)
- Removal of `DataExporter#repeat` as it can be deduce. Originally, I wanted to reuse `DataTable#isNestedWithinIterator` but I thought it could be useful in other scenarios, so I moved the algo into `ComponentUtils#invokeOnClosestIteratorParent`. @tandraschko you might have a better proposal regarding the name or expected params (I know what I did is a bit unconventionnal)
- Removal of callbacks used by exporters (only one is used now)
- Adding two methods in `DataTableExporter`: `preExport/postExport` allowing the export of multiple datatable (being said, xml should be able to do that also)
- The 10 expected parameters has been wrapped into one class `ExportConfiguration` to stop the congestion of differents methods


src/main/java/org/primefaces/component/datatable/export/DataTablePDFExporter.java
```diff
@@ -23,198 +23,115 @@
...
-import com.lowagie.text.FontFactory;
-import com.lowagie.text.Paragraph;
-import com.lowagie.text.pdf.PdfPCell;
-import com.lowagie.text.pdf.PdfPTable;
-import com.lowagie.text.pdf.PdfWriter;
+import javax.faces.component.UIComponent;
+import javax.faces.component.UIPanel;
+import javax.faces.context.ExternalContext;
+import javax.faces.context.FacesContext;
+import java.awt.*;
```
melloware: Why is java.awt.* instead of just java.awt.Color?

melloware (changes requested):

src/main/java/org/primefaces/component/datatable/export/DataTablePDFExporter.java
Rapster: I don't know, I'll change that

melloware:
Nice PR.  Big improvement and cleans up the code.

tandraschko:
Just my 0.02$:
I would not refactor the nesting iterator thing. Our UIData was copied from mojarra, which requires a nested iterator check, which is not required in the UiTabPanel, which is a copy from MyFaces UiData.
The whole thing actually needs a big refactoring.
Same as with p:repeat. Its a nasty copy which is required because of the nested iterator thing

Rapster:
Yeah that's something I wanted to talk about (but in another issue). So far, we?/I have no idea why we copied these classes, it doesn't say from which version it has been copied, what does it fix, does it contain any custom code  etc. Even `p:repeat` is an obscur component to me...

But again, I'm gonna create a ticket for that

tandraschko:
I will close it for now, feel free to create a new one then

Rapster:
Why creating a new one? and why closing?

tandraschko:
The build was red and you didn't work on it for a month. Or im wrong?

melloware:
I think the build was broken at the time and I have since fixed the travis.yml to fix the JDK8 build.

tandraschko:
Oh, then lets apply it.

Rapster:
I'll update the wiki

melloware:
@Rapster the Showcase Facelift is broken now too for the TextExporter example.  Can you fix please?

Rapster:
TextExporter?

Rapster:
How did I miss that one?! I'll check

melloware:
yeah its org.primefaces.showcase.view.data.dataexporter.TextExporter in the showcase.

melloware:
@Rapster thanks for fixing the showcase but I think something in the latest refacoring or in @ravihariharan original commit the PDF export is broken.  run the Showcase locally and on the Basic example try to export each of the 4 types.  All 4 work except the PDF is a 0 byte download and trying to open it its obviously not a PDF.  

Can you please investigate?

The current deployed Showcase the PDF is working fine.

melloware:
@Rapster nevermind I found the issue and I will fix it.  document.close() must be called before writeResposne.  You switched it around to call it afterwards.

PR #5062, opened 2019-08-12
Documentation : ColumnToggler: Label Issue on ColumnToggler

ravihariharan:


melloware (changes requested):
Technically since this is supported in 7.0 also shouldn't you update the 7.0 docs?

ravihariharan:
> Technically since this is supported in 7.0 also shouldn't you update the 7.0 docs?

Fixed

PR #5067, opened 2019-08-14
Fix #5064: Do not increase z-index on every AJAX request.

### LINKED ISSUE #5064: Datatable: Partial rendering of datatable with stickyHeader="true" and stickyTopAt="..." increases z-index of div
## 1) Environment
- PrimeFaces version: 7.0.2
- Application server + version: JBoss EAP 7.0.7.GA (Mojarra 2.2.12.SP4)
- Affected browsers: all
-  example with existing serenity theme showcase 

## 2) Expected behavior
Partial rendering of datatable with sticky table header should keep z-index stable
## 3) Actual behavior
z-index of div-Element above table-element increases with each ajax request which causes the datatable to be partially rendered.
example in our application:
before partial rendering:
<pre>
&lt;div class="ui-datatable ui-datatable-sticky ui-widget ui-shadow ui-sticky" style="position: fixed; width: 1629px; top: 0px; left: 17px; z-index: 1001;"&gt;
</pre>
after partial rendering
<pre>
&lt;div class="ui-datatable ui-datatable-sticky ui-widget ui-shadow ui-sticky" style="position: fixed; width: 1629px; top: 0px; left: 17px; z-index: 1003;"&gt;
</pre>

This behaviour leads to the effect that the drop downs of the primefaces menubar appear under the sticky table header after some partial renderings.

## 4) Steps to reproduce
Put a p:datatable on a sample page and a commandbutton which causes the datatable to rerender on ajax request.
## 5) Sample XHTML
...
<pre>
&lt;div class="ui-sticky"&gt;&lt;/div&gt;
&lt;p:dataTable id="id" var="item" value="#{controller.list}"
            	      stickyHeader="true" stickyTopAt=".ui-sticky"
                      rowSelectMode="checkbox" 
                      selection="#{controller.selectedRows}" rowKey="#{item.id}"&gt;
...
</p:dataTable>
</pre>
...

## 6) Cause
This part of the setupStickyHeader-function of datatable.js is responsible for the increasing z-index:
<pre>
        this.stickyContainer.css({
            position: "absolute",
            width: d.outerWidth(),
            top: g.top,
            left: g.left,
            "z-index": ++PrimeFaces.zindex
        });
</pre>
By the way: It is not the only code fragment where "++PrimeFaces.zindex" is found in the primefaces-javascript. In my oppinion this sort of dynamic increase of the zindex auf html-Elements leads to a lot of issues.

melloware:
What I did to fix this was the following:

1. Use global boolean variable Primefaces.isAjaxRequest to track whether or not this code is executing inside an ajax request or not.

2. Changed all ++zIndex calls to use a function PrimeFaces.getZindex(); which will return either the same index inside AJAX requests or +1 if loading the page for the first time to assign zIndexes.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
+++ b/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
@@ -524,7 +524,7 @@ PrimeFaces.widget.AutoComplete = PrimeFaces.widget.BaseWidget.extend({
                         .css({
                             'left':'',
                             'top':'',
-                            'z-index': ++PrimeFaces.zindex,
+                            'z-index': PrimeFaces.getZindex(),
                             'width': content.outerWidth()
                         })
                         .position({
@@ -850,7 +850,7 @@ PrimeFaces.widget.AutoComplete = PrimeFaces.widget.BaseWidget.extend({
         this.panel.css({'left':'',
                         'top':'',
                         'width': panelWidth,
-                        'z-index': ++PrimeFaces.zindex
+                        'z-index': PrimeFaces.getZindex()
                 });
 
         if(this.panel.parent().is(this.jq)) {
--- a/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
+++ b/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
@@ -59,7 +59,7 @@ PrimeFaces.widget.BlockUI = PrimeFaces.widget.BaseWidget.extend({
     },
     
     show: function() {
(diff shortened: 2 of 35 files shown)


tandraschko:
unsure but this will probably introduce many problems with nested components which rely on the zindex

melloware:
I agree any thoughts on how we can solve this problem for the ticket?   Should I constrain this PR to fix only the 1 place where they are having the issue?

tandraschko:
currently no time sorry, mert can also better help regarding datatable sticky stuff.

tandraschko:
I will close it for now, lets see if mert can check the issue

PR #5074, opened 2019-08-17
Fix #5075: SelectOneMenu fixed p:column rendered property

### LINKED ISSUE #5075: SelectOneMenu: column rendered attribute doesn't work when using advanced mode
## 1) Environment
- PrimeFaces version: 7.0 and 7.1-SNAPSHOT
- Does it work on the newest released PrimeFaces version? Version? No, it doesn't.
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) No, it doesn't.
- Application server + version: JBoss EAP 7.2.2
- Affected browsers: Any.

## 2) Expected behavior

When setting `<p:column rendered="false">`, the column shouldn't be rendered.

## 3) Actual behavior

The column is rendered even if rendered is set to false.

## 4) Steps to reproduce

This issue can be reproduced in [Primefaces Showcase](https://www.primefaces.org/showcase/ui/input/oneMenu.xhtml), if you set `<p:column rendered="false">` in the advanced version of SelectOneMenu. 

## 5) Sample XHTML

```
        <p:outputLabel for="advanced" value="Advanced:" />
        <p:selectOneMenu id="advanced" value="#{selectOneMenuView.theme}" converter="themeConverter" panelStyle="width:180px"
                         effect="fade" var="t" style="width:160px" filter="true" filterMatchMode="startsWith">
            <f:selectItems value="#{selectOneMenuView.themes}" var="theme" itemLabel="#{theme.displayName}" itemValue="#{theme}" />
 
            <p:column style="width:10%" rendered="false">
                <h:graphicImage name="showcase/images/themeswitcher/themeswitcher-#{t.name}.png" alt="#{t.name}" styleClass="ui-theme" />
            </p:column>
 
            <p:column>
                <h:outputText value="#{t.displayName}" />
            </p:column>
        </p:selectOneMenu>
```

## 6) Sample bean

Not needed.

hfluz:
When using advanced version of SelectOneMenu with a inner columns, the column rendered property is ignored.

This commit fix that issue.

### DIFF
--- a/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
+++ b/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
@@ -429,19 +429,21 @@ public class SelectOneMenuRenderer extends SelectOneRenderer {
             else {
                 for (int j = 0; j < columns.size(); j++) {
                     Column column = columns.get(j);
-                    String style = column.getStyle();
-                    String styleClass = column.getStyleClass();
+                    if (column.isRendered()) {
+                        String style = column.getStyle();
+                        String styleClass = column.getStyleClass();
 
-                    writer.startElement("td", null);
-                    if (style != null) {
-                        writer.writeAttribute("style", style, null);
-                    }
-                    if (styleClass != null) {
-                        writer.writeAttribute("class", styleClass, null);
-                    }
+                        writer.startElement("td", null);
+                        if (style != null) {
+                            writer.writeAttribute("style", style, null);
+                        }
+                        if (styleClass != null) {
+                            writer.writeAttribute("class", styleClass, null);
+                        }
 
-                    renderChildren(context, column);
-                    writer.endElement("td");
+                        renderChildren(context, column);
+                        writer.endElement("td");
+                    }
                 }
             }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
```diff
@@ -429,19 +429,21 @@ protected void encodeOptionsAsTable(FacesContext context, SelectOneMenu menu, Li
             else {
                 for (int j = 0; j < columns.size(); j++) {
                     Column column = columns.get(j);
```
melloware, line 431: A better fix would be...
```
Column column = columns.get(j);
if (!column.isRendered()) {
   continue;
}
```

melloware (changes requested):

src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
hfluz: That's a nice suggestion. I refactored my PR.

tandraschko:
Thanks!

PR #5078, opened 2019-08-19
Fix #3925: SelectCheckboxMenu placeholder with multiple=true

### LINKED ISSUE #3925: SelectCheckboxMenu: placeholder not supported with multiple=true
## 1) Environment
PrimeFaces version: 6.2
Affected browsers: all

## 2) Expected behavior
A stylable hint should be displayed when label as value.

## 3) Actual behavior
The attribute "label" does not work with multiple="true"

melloware:


tandraschko:
what happens on IE11?

melloware:
@tandraschko Tested in IE11, Edge, Firefox and Chrome all working great.

PR #5081, opened 2019-08-20
Fix #5061: TouchPunch 1.0.5 and extensions.txt doc.

### LINKED ISSUE #5061: Dialog does not close in Chrome on touch devices
**Environment**
PrimeFaces version: 7.0.5
Did work on 6.X

Affected browsers: Chrome Mobile
Tested on Samsung TabActive and S10

**Expected behavior**
Dialog should close on touch off the close icon in the top right panel

**Actual behavior**
Dialog does not close

**Steps to reproduce**
[primefaces-test-dialog.zip](https://github.com/primefaces/primefaces/files/3491654/primefaces-test-dialog.zip)

**Sample XHTML**
```
<h:body>
        <p:commandButton value="Basic" type="button" onclick="PF('dlg1').show();" />
        <p:dialog header="Basic Dialog" widgetVar="dlg1" minHeight="40">
            <h:outputText value="Resistance to PrimeFaces is futile!" />
        </p:dialog>
</h:body>


```

melloware:
- Updated to 1.0.5 of RWAP's fork where he accepted my Lenovo TouchScreen patch
- Addded doc to extensions.txt to document this plugin and the fork URL we are using.
- Original Touch Punch has been inactive for 3+ years
- RWAP's fork has been keeping it modern and fixing issues like Apple Pencil and the original bug we had for #5061

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/jquery/extensions.txt
+++ b/src/main/resources/META-INF/resources/primefaces/jquery/extensions.txt
@@ -10,6 +10,7 @@ List of Plugins
 - Cursor Position (https://github.com/beviz/jquery-caret-position-getter)
 - Browser (https://github.com/gabceb/jquery-browser-plugin/releases)
 - Autosize (https://github.com/jackmoore/autosize)
+- Touch Punch (https://github.com/RWAP/jquery-ui-touch-punch custom fork of https://github.com/furf/jquery-ui-touch-punch)
 
 
 
--- a/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.touch-punch.js
+++ b/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.touch-punch.js
@@ -11,10 +11,6 @@
  * Depends:
  * jquery.ui.widget.js
  * jquery.ui.mouse.js
- *
- * Depends:
- *  jquery.ui.widget.js
- *  jquery.ui.mouse.js
  */
 (function( factory ) {
     if ( typeof define === "function" && define.amd ) {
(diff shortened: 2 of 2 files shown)


Rapster:
I think you should prefer the minified version, and put your comments in extensions.txt

melloware:
@Rapster  Won't it get minified as part of the build process in Maven?

Rapster:
You can exclude js files if necessary, see https://github.com/primefaces/primefaces/blob/3130c198f7f56bf0e54afc4dcb39a4b154a3050f/pom.xml#L934

melloware:
I guess I prefer the extensions model where the JS is uncompressed at rest which makes developing and debugging easier.  The only time is gets compressed is when a release build is cut.

tandraschko:
+1
better to avoid include/exclude configs

Rapster:
I understand, but when the original repo proposed a minified version I'd rather go for it.

First, why bothering doing something that someone else have already done? Second, with a minified version developers are not encouraged to make custom changes and having a tough time to see differences in the future. Also, you shouldn't make modifications in these imported files, even if it's only comments, this should be done in an appropriate file like extensions.txt, btw you duplicate comments in the js file

Regarding debugging, you can easily have a pretty version of minified version on web browser nowadays.

And I don't see any reasons why avoiding excludes? Actually, we should avoiding minified js files as much as we can

melloware:
Thanks I removed the duplicate comment.

As for Min version the reason I don't use it is look at that fork: https://github.com/RWAP/jquery-ui-touch-punch

They don't seem to be updating the MIN version only the unminified version.  I have seen this a LOT in open source Git projects where the minified version is not in sync with the unminified version.  In fact I have been burned by it before.

The only time I prefer the minified version if the library is like 200K+ like Raphael or FullCalendar. But for a file that is 200 lines I don't see the need when the build tool can do it.

Rapster:
In this case, makes sense to keep the unminified then. But our comments should be in extensions.txt

PR #5082, opened 2019-08-20
Fix #4998: Allow custom animation duration for show/hide

### LINKED ISSUE #4998: Dialog/BlockUI: allow custom animation duration for show/hide
## 1) Environment
I currently use the latest primefaces Elite/Pro Version 7.0.5.
As this function currently does not exists this is an extension 
for the existing release, but i think very easy to implement.

## 2) Expected behavior
I am using **e.g. the primefaces "ajaxStatus"** component
to show a please wait dialog and also lock the entire page while Ajax is not ready.
But when Ajax is ready i hide the "Please wait dialog" and then remove the page 
lock also by hiding them.

Now here's my problem/improvement idea:
In summary the hidding of the 2 components needs about 400ms because
the dialog and also the lock page will be faded out.
So it seems that in the background jQuery is working here which 
is able to set the speed in ms (milliseconds) or as text when i don't won't a 
default fadein/fadeout.

So is it possible to set the parameters of primefaces show/hide 
exact like i can do in jquery ?

Here an example to understand what i currently do
and what i want:

_My default form template i use on each form:_
```
<h:form id="DefaultFormId" prependId="false" autocomplete="off">
	<!-- Global Ajax please wait dialog -->
	<p:dialog widgetVar="bui" id="bui" modal="true" draggable="false" closable="false"
			  resizable="false" showHeader="false" position="center" style="z-index:10000; opacity: 0.85;">
		<p:panel style="margin-top:10px;">
			<f:facet name="header">
				<h:outputText value="Please wait" style="color:#000000;"/>
			</f:facet>
			<h:outputText style="color:#ffa500;" value="Loading, please wait..."/>
		</p:panel>
	</p:dialog>
    <!-- Inline javascript functions -->
    <h:outputScript>
        function LockPage() {
            PF('BlockDefaultLayoutWrapper').show();
            PF('bui').show();
        }
        function UnLockPage() {
                PF('bui').hide();
                PF('BlockDefaultLayoutWrapper').hide();
        }
        // Execute script on page ready
        $(document).ready(function()
        {
            UnLockPage();
        });
    </h:outputScript>
	<!-- Ajax status -->
	<p:blockUI block="DefaultLayoutWrapper" widgetVar="BlockDefaultLayoutWrapper"/>
	<p:ajaxStatus onstart="LockPage();" oncomplete="UnLockPage();" onerror="UnLockPage();" onsuccess="UnLockPage();"/>
</h:form>
```

The show() and the hide() functions fade the dialog and the blockUI page in/out
which is visual very slow for a user.

What i want is to call the show/hide like in jquery:
```
PF("...").show("fast");
PF("...").show("slow");
PF("...").show(); // Default 400ms like in jQuery
PF("...").show(100); // 100ms
```
... and the same for the hide function.

As i think that the function calling jquery inside
the only thing to do here is to pass the parameters directly to the jQuery functions ?!

melloware:
Added new optional param [duration] to .show and .hide function of dialog and block UI.

If the parameter is left off the component behaves like .show() and .hide() always have.

### DIFF
--- a/docs/7_1/components/blockui.md
+++ b/docs/7_1/components/blockui.md
@@ -72,8 +72,8 @@ Widget: _PrimeFaces.widget.BlockUI_
 
 | Method | Params | Return Type | Description | 
 | --- | --- | --- | --- | 
-| show() | - | void Blocks the UI.
-| hide() | - | void Unblocks the UI
+| show(duration) | duration: (optional) duration of the animation | void | Blocks the UI.
+| hide(duration) | duration: (optional) duration of the animation | void | Unblocks the UI
 
 ## Skinning
 Following is the list of structural style classes;
--- a/docs/7_1/components/dialog.md
+++ b/docs/7_1/components/dialog.md
@@ -146,8 +146,8 @@ Widget: _PrimeFaces.widget.Dialog_
 
 | Method | Params | Return Type | Description | 
 | --- | --- | --- | --- | 
-| show() | - | void | Displays dialog.
-| hide() | - | void | Closes dialog.
+| show(duration) | duration: (optional) duration of the animation | void | Displays dialog.
+| hide(duration) | duration: (optional) duration of the animation | void | Closes dialog.
 | isVisible() | - | void | Returns visibility as a boolean.
 | resetPosition | - | void | Reset the dialog position based on the configured "position".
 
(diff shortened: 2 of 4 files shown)


tandraschko:
can't we apply the duration for the animations, too?

melloware:
Let me try that.

melloware:
Yep it works great. Just updated the PR.

tandraschko:
then +1
but please also add the methods and the new param to the docu

melloware:
7.1 Docs updated!

tandraschko:
cool thanks!

PR #5083, opened 2019-08-21
Fix #5084: IdleMonitor: support multiple on the same page

### LINKED ISSUE #5084: IdleMonitor: support multiple on the same page
# IdleMonitor: support multiple on the same page
When using multiple idle monitor tags, they influence each other. I needed to use separate idle timers on the same page with different timeout times.

I also made a pull request with my solution #5083. I developed this for primefaces 6.1 but rebased and adapted it to the current 7.1.SNAPSHOT master.

Note that the idle timer is from [https://github.com/thorst/jquery-idletimer](https://github.com/thorst/jquery-idletimer).
## 1) Environment
- PrimeFaces version: 7.1
- Issue is present on 6.1 and current master as well
- Application server + version: Wildfly 17.0.1.Final, but affects all
- Affected browsers: Chrome 76.0.3809.100, but should affect all

## 2) Expected behavior
When using multiple p:idleMonitor tags with different timeouts, they should all work independently of each other.

## 3) Actual behavior
The idleMonitors all store data under the same key and overwrite the timers of each other.

## 4) Steps to reproduce

Use two p:idleMonitor tags on the same page with different timeouts. Check when onidle is executed.

## 5) Sample XHTML

This reproduces the issue. Check console log when onidle timers are executed.
[https://github.com/pazepaze/primefaces-test/commit/0785d96c1b04be240558e7fbba1208c74fd930b9](https://github.com/pazepaze/primefaces-test/commit/0785d96c1b04be240558e7fbba1208c74fd930b9)

## 6) Sample bean

not applicable

pazepaze:
When using multiple idle monitor tags, they influence each other. I needed to use separate idle timers on the same page with different timeout times.

Client id is used to make several strings that are used as keys and event names unique.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/idlemonitor/0-jquery.idletimer.js
+++ b/src/main/resources/META-INF/resources/primefaces/idlemonitor/0-jquery.idletimer.js
@@ -16,7 +16,7 @@
 */
 (function ($) {
 
-    $.idleTimer = function (firstParam, elem) {
+    $.idleTimer = function (firstParam, elem, uniqueId) {
         var opts;
         if ( typeof firstParam === "object" ) {
             opts = firstParam;
@@ -37,14 +37,14 @@
         }, opts);
 
         var jqElem = $(elem),
-            obj = jqElem.data("idleTimerObj") || {},
+            obj = jqElem.data("idleTimerObj" + uniqueId) || {},
 
             /* (intentionally not documented)
              * Toggles the idle state and fires an appropriate event.
              * @return {void}
              */
             toggleIdleState = function (e) {
-                var obj = $.data(elem, "idleTimerObj") || {};
+                var obj = $.data(elem, "idleTimerObj" + uniqueId) || {};
 
                 // toggle the state
                 obj.idle = !obj.idle;
@@ -53,7 +53,7 @@
                 obj.olddate = +new Date();
 
                 // create a custom event, with state and name space
-                var event = $.Event((obj.idle ? "idle" : "active") + ".idleTimer");
+                var event = $.Event((obj.idle ? "idle" : "active") + ".idleTimer" + uniqueId);
 
                 // trigger event on object with elem and copy of obj
                 $(elem).trigger(event, [elem, $.extend({}, obj), e]);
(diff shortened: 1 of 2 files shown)


melloware:
Please create an issue first in the Issues section.  No PR is accepted without an Issue.

tandraschko:
does multiWindow still work?

pazepaze:
You're right I need to adjust that, too. To be honest, multi window didn't quite work as I expected in the first place though. But I can adjust it so that it should work the same as before.

tandraschko:
What did you expect and what was the result?
I developed it for a customer, exactly like it is currently

pazepaze:
I accept the possibility that I just didn't understand how it works.

All I did was open two windows in chrome on a page with a p:idleMonitor with multi window set to true and a console.log on idle and active. I expected to see the same log output on both windows. I didn't.

pazepaze:
> does multiWindow still work?

works the same as before now

tandraschko:
really not sure as you append the id now to the globalLastActiveKey
this means if you have 2 tabs with the same view (and therefore same idleMonitor ids), it should work
however if this are other views (other idleMonitor ids), it will probably fail

pazepaze:
You are right in that it behaves different now if you had different idleMonitors in different views and expected them to still be connected somehow with multiWindowSupport. Although I am not sure how one would even specify how that was supposed to behave since you could also chose different timeouts there.

I would argue that the expected behavior is that idleMonitors in different views (with different ids) should not influence each other, whether multiWindowSupport is active or not.

tandraschko:
Not sure.
Depends on the usecase; my customers usecase:
If someone is idle for X minutes in all tabs for the same context, automatically redirect to a logout page.

in this case the idleMonitor is defined in a template probably. So your change should not break it.

tandraschko:
@pazepaze 
can you try to get your changes applied in https://github.com/thorst/jquery-idletimer first?

melloware:
@pazepaze @tandraschko I submitted a ticket on the IdleTimer page: https://github.com/thorst/jquery-idletimer/issues/53

pazepaze:
btw. I had also made a PR [thorst/jquery-idletimer/pull/54](https://github.com/thorst/jquery-idletimer/pull/54) there

Issue #3245, opened 2018-01-29
Timeline: Migrate to vis.js


cnsgithub:
http://almende.github.io/chap-links-library/

> CHAP Links Library now has a successor: vis.js. A couple of fundamental limitations in the current architecture of CHAP Links Library have been reached. Therefore, a new initiative has been started, aiming at more flexible data manipulation, better performance, and expandable visualization tools.

http://visjs.org/

markusg80:
Are here any news? I think it would be a good idea to change to vis.js! This way primefaces eill get a more powerfull Timeline Component

melloware:
I think I will do this for PF 7.1  My client started using this component and it would be good to update but its too close to 7.0 release.

jasonex7:
What should be the strategy to migrate to vis.js? Create a new component based on vis.js or update existing timeline component to use vis.js?

melloware:
I would prefer to upgrade the current component to vis.js to make the upgrade for users as seamless as possible.

jasonex7:
I've revised current vis.js [documentation](https://visjs.github.io/vis-timeline/docs/timeline/) and how to migrate current `<p:timeline>` to vis.js and see, until now, one main difference: in vis.js the events (or items) are identified by a new field called `id` and not by positional `index` as current timeline.js is. So I think a new field called `id` must be created in primefaces class `TimelineEvent` and modify classes TimelineEvent (equals and hash methods), Timeline, TimelineModel, TimelineUpdater and javascript file `1-timeline.js` to work by that `id` field instead of `index`, in fact I recommend remove all references to index field because there are no such property on items data of vis.js and I think no safe manner to obtain it.

Maybe that change affect existing projects that use `<p:timeline>` and the migration to the new version was not so seamless.

melloware:
Couldn't we just use the "Index" field in the current event but just map it to the new "ID" field i nthe vis.js?   Basically its what makes a timeline even unique right?

jasonex7:
That would be a good option, having Primefaces work by positional index and when rendered the list of `TimelineEvent` instances then map positional index to id property. There are more subtle differences like more fine grained control over editable property of an item (editable.remove, editable.updateTime, editable.updateGroup) but I think that wouldn't be a major problem.

[...]

PR #5098, opened 2019-08-28
Fix #5097 - Timeline: move event between groups change its type

### LINKED ISSUE #5097: Timeline: move event between groups change its type
## 1) Environment
- PrimeFaces version: 7.0
- Application server + version: WildFly 10.1
- Affected browsers: all

## 2) Expected behavior
When user moves a timeline event between groups, the event should maintain its type ('range', 'box', etc). 

## 3) Actual behavior
When user moves a timeline event between groups, the event change its type to 'box' always.

## 4) Steps to reproduce

1. Go to timeline/custom showcase: https://www.primefaces.org/showcase/ui/data/timeline/custom.xhtml
2. Move an event to another group

     image

3. The event change its type from 'range' to 'box'.

     image

jasonex7:


melloware (approved):
@tandraschko Since this library is the last version and we already have a backlog ticket to move to the new vis.js library I am OK with making this change directly in the timeline.js code.  Thoughts?

https://github.com/primefaces/primefaces/issues/3245

jasonex7:
@melloware I've downloaded timeline.js version [2.9.1](http://almende.github.io/chap-links-library/js/files/timeline-2.9.1.zip) from the site http://almende.github.io/chap-links-library/timeline.html and that file already had the fix. I don't know why primefaces timeline.js differs from the original on that line.

melloware:
Oh wow did you do a diff to see if there any other differences?  I thought we were using 2.9.1?

jasonex7:
Yes, I did a diff. There are only 2 more differences:

#### Diff 1 
##### PF timeline.js starting from line 774
````javascript
links.Timeline.prototype.setVisibleChartRangeAuto = function() {
    if (this.options && (this.options.start || this.options.end)) {
        this.setVisibleChartRange(this.options.start, this.options.end);
    } else {
        var range = this.getDataRange(true);
        this.setVisibleChartRange(range.min, range.max);
    }
};
````
##### Original timeline.js starting from line 774
````javascript
links.Timeline.prototype.setVisibleChartRangeAuto = function() {
    var range = this.getDataRange(true);
    this.setVisibleChartRange(range.min, range.max);
};
````

#### Diff 2 
##### PF timeline.js starting from line 5336
````javascript
if (axisOnTop) {
      groupBase += Math.max(options.groupMinHeight, group.labelHeight) + eventMargin;
} else {
       groupBase -= (Math.max(options.groupMinHeight, group.labelHeight) + eventMargin);
}
````
##### Original timeline.js starting from line 5332
````javascript
if (axisOnTop) {
       groupBase += options.groupMinHeight + eventMargin;
} else {
       groupBase -= (options.groupMinHeight + eventMargin);
}
````

melloware:
OK so Diff#2 the "groupbase" fix was done on purpose: https://github.com/primefaces/primefaces/issues/1995

Diff#1 I can't explain.

jasonex7:
I've investigated about my commit proposed on this PR and I found that this was the original behavior until change was introduced in primefaces-extensions on this commit: https://github.com/primefaces-extensions/core/commit/60e9188b2182e15492ebda6bfd1599044643d61a on [line 4931](https://github.com/primefaces-extensions/core/blob/04311a7197309c3dab9ce9cac3f32aa18768497b/src/main/resources/META-INF/resources/primefaces-extensions/timeline/0-timeline.js#L4931) related to [issue 298](https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/298). I think that change was mistakenly introduced.

jasonex7:
> Diff#1 I can't explain.

I've investigated that Diff#1 comes from this commit on primefaces-extensions https://github.com/primefaces-extensions/core/commit/0d9d7fe395f2dc8578181b8e3d63cf6860b92daa to resolve [issue 323](https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/323)

melloware:
OK cool. great detective work.

jasonex7:
> OK cool. great detective work.

Thank you.

> I think I will do this for PF 7.1  My client started using this component and it would be good to update but its too close to 7.0 release.
> _Originally posted by @melloware in https://github.com/primefaces/primefaces/issues/3245#issuecomment-463615310_


By the way @melloware how does issue #3245 is going? :smile: :+1:

melloware:
HA I think it was bigger than i thought so I put it to the side.  As a note we are successfully using Timeline in production and its quite solid!

jasonex7:
Thanks for that comment, I hope some day that issue was closed. By now I only miss from vis.js the capability of hide some ranges of time (perhaps non labor hours/days) using `hiddenDates` [configuration option](https://visjs.github.io/vis-timeline/docs/timeline/#Configuration_Options).

melloware:
Have you looked into it at all on how difficult it would be to migrate to vis.js. I didn’t know how drastically rewritten it was?

jasonex7:
I've only compared javascript sources (timeline.js vs vis-timeline-graph2d.esm.js version 5.1.0 and also timeline.js vs vis.js version 4.21) superficially and saw that there are plenty of differences.

PR #5103, opened 2019-08-30
Fix #688: Paginator allow disabling of focus.

### LINKED ISSUE #688: Paginator: Refocus the paginator after pagination should be optional
In Paginator.updatePageLinks(), there is the following code that refocuses on the page button:

``` javascript
....
        focusContainer, tabindex;

        if(focusedElement.hasClass('ui-paginator-page')) {
            var pagesContainerIndex = this.pagesContainer.index(focusedElement.parent());
            if(pagesContainerIndex >= 0) {
                focusContainer = this.pagesContainer.eq(pagesContainerIndex);
                tabindex = focusedElement.index();
            }
        }
...
     if(focusContainer) {
            focusContainer.children().eq(tabindex).trigger('focus');
        }
```

This makes sense, but in some cases (like when the row content is really tall inside of a DataList or DataGrid), you may want to set the focus elsewhere. It'd be nice if the current behavior was optional.

melloware:
By default the behavior is not changing as it defaults to refocusing the paginator as it always has.  This just creates a widget method to give downstream users the ability to turn off refocus with...

```javascript
PrimeFaces.widget.Paginator.prototype.allowRefocus =  function() {
        return false;
 };
```

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/paginator/paginator.js
+++ b/src/main/resources/META-INF/resources/primefaces/paginator/paginator.js
@@ -304,7 +304,7 @@ PrimeFaces.widget.Paginator = PrimeFaces.widget.BaseWidget.extend({
             this.pagesContainer.append('<a class="' + styleClass + '" aria-label="' + ariaLabel + '" tabindex="0" href="#">' + (i + 1) + '</a>');
         }
 
-        if(focusContainer) {
+        if(focusContainer && this.allowRefocus()) {
             focusContainer.children().filter('.ui-state-active').trigger('focus');
         }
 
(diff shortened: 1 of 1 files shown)


tandraschko:
@mertsincan WDYT?

melloware:
As a note I decided to do this as a widget override because the paginator is used by so many components it gives and easy way to enable the functionality per widget per all the different components that use it.

mertsincan:
It's by design. I think it is necessary for accessibility. Therefore, users can override updatePageLinks widget method to solve this issue.

melloware:
I agree they could override the entire updatePageLinks method but i thought this solution gave them a more precision piece of code to override without having to override a larger method.  I can close this PR if you want but then I recommend we close the main ticket as well.

tandraschko:
+0.1 for choose "implemented by design"

melloware:
OK closing both the PR and the ticket as it can be done by overriding updatePageLinks.

PR #5107, opened 2019-09-03
Fix #4863: Removed check requiring file extension.

### LINKED ISSUE #4863: fileUpload: Unable to upload a file without extension
## 1) Environment
- PrimeFaces version: 7.0.3
- Affected browsers: All

## 2) Expected behavior
It should be possible to upload any existing file.

## 3) Actual behavior
Files without extensions cause a FacesExeption.

## 4) Steps to reproduce
Create a file named 'foobar' and try to upload it

Reason for this is the following code in FileUploadUtils.getValidFilename:
```
        String extension = FilenameUtils.EXTENSION_SEPARATOR_STR + FilenameUtils.getExtension(filename);

        if (extension.equals(FilenameUtils.EXTENSION_SEPARATOR_STR)) {
            throw new FacesException("File must have an extension");
        }
```
This seems to enforce the presence of a filename extension for later checks. I am unsure if this is really needed.

The most pressing problems are:

- A user of a primefaces based application is not aware that this restriction exists
- Any primefaces based application is not able to catch that exception properly

melloware:
Tested and everything is working properly without an extension.

### DIFF
--- a/src/main/java/org/primefaces/util/FileUploadUtils.java
+++ b/src/main/java/org/primefaces/util/FileUploadUtils.java
@@ -88,12 +88,8 @@ public class FileUploadUtils {
         }
 
         String name = FilenameUtils.getName(filename);
-        String extension = FilenameUtils.EXTENSION_SEPARATOR_STR + FilenameUtils.getExtension(filename);
-
-        if (extension.equals(FilenameUtils.EXTENSION_SEPARATOR_STR)) {
-            throw new FacesException("File must have an extension");
-        }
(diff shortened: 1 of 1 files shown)


tandraschko:
if its well tested -> +1
@cnsgithub would you like to review?

tandraschko:
no feedback. Should we apply the fileupload changes and let the users also test?

melloware:
I think so its better to publish these fixes and let people test and provide feedback.  I will merge. We can always fix again if anymore reports come in.

PR #5108, opened 2019-09-03
Fix #4843: Upload allowTypes must also check contentType.

### LINKED ISSUE #4843: fileUpload: Update works sporadically
1) Environment
- PrimeFaces version:
Tried on versions 7.0, 7.0.2. 7.0.3

Browsers: I initially discovered while using IE 11, but also was able to reproduce in the showcase using Chrome 74.

 2) Expected behavior
After uploading a file, the component given by "update" should always update

 3) Actual behavior
The component given by "update" does update sometimes, but frequently, it does not.

 4) Steps to reproduce
I was able to reproduce using the showcase. https://www.primefaces.org/showcase/ui/file/upload/single.xhtml

I select a file then click upload. The growl message does not always appear as it should. It usually takes me 3 or less tries to reproduce.

5) Sample XHTML
6) Sample bean
See showcase code

melloware:
Now the server side check matches the UI check.

### DIFF
--- a/docs/7_0/gettingstarted/dependencies.md
+++ b/docs/7_0/gettingstarted/dependencies.md
@@ -10,6 +10,7 @@ any 3rd part work incorporated are compatible with the PrimeFaces Licenses.
 | itext | 2.1.7 | Optional | DataExporter (PDF) |
 | apache poi | 3.17 | Optional | DataExporter (Excel) |
 | rome | 1.9.0 | Optional | FeedReader |
+| apache tika | 1.22 | Optional | FileUpload |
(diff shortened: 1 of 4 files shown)


tandraschko:
if its well tested -> +1
@cnsgithub would you like to review?

tandraschko:
will this work fine without tika or would it break every upload?

melloware:
Yep this fix works with or without Tika.  Especially Since Tika uses accept="image/jpg" property of the file upload.  This just simply adds the same check as the UI to check that the allowTypes="jpeg|gif" regex matches either the file name or the type.

PR #5115, opened 2019-09-05
Fix #3378: DialogFramework allow widgetVar option.

### LINKED ISSUE #3378: DialogFramework: add ability to define a custom widgetVar
Recently, I have posted some remarks about the interception of closing/hiding DialogFramework.

I have now a solution when some parts are explain in
https://github.com/primefaces/primefaces/issues/3373
https://github.com/primefaces/primefaces/issues/2538

This work well, but it is very complicated and tricky to find the DialogFramework Primefaces widget.
I explain that in forum https://forum.primefaces.org/viewtopic.php?f=3&t=54492

The code used to find the PF('widget') is

```
var wgDialog = jQuery(".ui-dialog.ui-overlay-visible",window.parent.document)
                .each
                    (function(nIndex)
                        {
                        var sWidgetName = $(this).attr('data-widgetvar');
                        var wgDialog = window.parent.PF(sWidgetName);
```

It would be simpler if a "widgetvar" attribute exists for DialogFramework widget.

Can somebody at Primefaces team implement that request ?

melloware:
Tested and if widgetVar is not assigned it defaults to current behavior of auto assignment.

Other than that a widget var can be assigned with...

```java
options.put("widgetVar", "wgtChooseDialog");
```

### DIFF
--- a/docs/7_1/core/dialogframework.md
+++ b/docs/7_1/core/dialogframework.md
@@ -99,7 +99,7 @@ Here is the full list of configuration options;
 
 | Name | Default | Type | Description | 
 | --- | --- | --- | --- |
-widgetVar | null | String | Custom widgetVar of the dialog, if not declared it will be automatically created as id+'_dlgWidget'.
+widgetVar | null | String | Custom widgetVar of the dialog, if not declared it will be automatically created as "id+dlgWidget".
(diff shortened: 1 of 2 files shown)


tandraschko:
+1, thanks

PR #5121, opened 2019-09-06
Fix #5119: CSV respect Growl summary/detail attributes.

### LINKED ISSUE #5119: Client Side Validation: skip values to false of sumary and detail attributes of growl component
1) Environment
- PrimeFaces version: 7.0
- Does it work on the newest released PrimeFaces version? No Version? 7.0
- Does it work on the newest sources in GitHub? NO - 7.1-SNAPSHOT (build)
- Application server + version: Apache Tomcat 9.5, JSF 2.3 (Mojarra 2.3.2)
- Affected browsers: ALL

## 2) Expected behavior
When validating the input with Client Side Validation, the growl component should only show the detail and not the summary.

## 3) Actual behavior
When validating the input with Client Side Validation, the growl component shows both the detail and the summary, in spite of having false attributes. When both attributes are set to false or combinations are made with both true and false it also always shows both.

## 4) Steps to reproduce
1.  Create a growl with showSumary="false" and showDetal="true"
2. Create a p:messages with showSumary="false" and showDetal="true"
2. Create a input with required="true"
3. Create commandbutton with validateClient="true" ajax="false"
4. Press commandbutton and view sumary and detail on growl, p:messages show correctly
 
I have reviewed the Client Side Validation JS code and I have been able to verify that in the renderMessages function: function (container) {on line 869, when iterating to get the values ​​of the growl attributes, the attribute values ​​are not obtained showSumary and showDetail, which means that the two values ​​are always sent in the component. For example, for the p: messages component, if the values ​​of these attributes are obtained, the p: messages works correctly.

The implementation of the summary and detail based on the attributes defined in the JS code of Client Side Validation is missing, in addition to collecting the values ​​and putting them in the html.
In the generated html code, these attributes are also not shown for the growl and if for the p: messages.

HTML Generated GROWL:
`<span id="form:messages" class="ui-growl-pl" data-widget="widget_form_messages" data-severity="all,error" data-redisplay="true"></span>`

datadetail is missing

HTML Generated MESSAGES:
`<div id="form:mesinput" class="ui-messages ui-widget" aria-live="polite" data-global="false" data-detail="data-detail" data-severity="all,error" data-redisplay="true"><div class="ui-messages-error ui-corner-all"><span class="ui-messages-error-icon"></span><ul><li><span class="ui-messages-error-detail">INPUT: Validation Error: Value is required.</span></li></ul></div></div>`

## 5) Sample XHTML
`<h:body>
	<h:form id="form">
		<p:growl id="messages" showSummary="false" showDetail="true" />
		<h:outputLabel value="Value:"/>
		<p:inputText id="input" required="true" label="INPUT" styleClass="ui-state-highlight" />
		<p:messages id="mesinput" showDetail="true" showSummary="false"/>
		<p:commandButton update="@form" validateClient="true" value="BOTON"></p:commandButton>
	</h:form>
</h:body>`

## 6) Sample bean
NONE

melloware:


tandraschko:
+1 if well tested

melloware:
Using the CSV showcase Custom example here is what it it looks like with...
```xml
<p:growl showDetail="true" showSummary="true"/>
```
image

And now with...
```xml
<p:growl showDetail="false" showSummary="true"/>
```
image

Issue #3585, opened 2018-04-15
DataExporter: PDF export does not work for Chinese data


Mararsh:
## 1) Environment
- PrimeFaces version: 6.1
- Does it work on the newest released PrimeFaces version? Version? No
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) No
- Application server + version: GlassFish 4.1.2
- Affected browsers: Firefox

## 2) Expected behavior
PDF dataExporter should export Chinese correctly with UTF-8 encoding.

## 3) Actual behavior
PDF dataExporter shows messed text in exported PDF file for data in Chinese. 

## 4) Steps to reproduce
Create a p:dataTable with data containing Chinese.

## 5) Sample XHTML
Chinese are correct in exported XLS and CSV, but messed in exported PDF:
```
<p:commandLink value="pdf" ajax="false"
	actionListener="#{userFeedbackBean.exporter.checkExportFilename2('TUserFeedback')}">
<p:dataExporter type="pdf"  encoding="UTF-8"
	target=":TUserFeedbackListForm:datalist"
	pageOnly="#{userFeedbackBean.exporter.exportContent eq 'page'}"
	selectionOnly="#{userFeedbackBean.exporter.exportContent eq 'selected'}"
	fileName="#{userFeedbackBean.exporter.exportFilename}"/>
</p:commandLink>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;

<p:commandLink value="xls" ajax="false"
	actionListener="#{userFeedbackBean.exporter.checkExportFilename2('TUserFeedback')}">
<p:dataExporter type="xls"  encoding="UTF-8"
	target=":TUserFeedbackListForm:datalist"
	pageOnly="#{userFeedbackBean.exporter.exportContent eq 'page'}"
	selectionOnly="#{userFeedbackBean.exporter.exportContent eq 'selected'}"
	fileName="#{userFeedbackBean.exporter.exportFilename}"/>
</p:commandLink>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;

<p:commandLink value="csv" ajax="false"
	actionListener="#{userFeedbackBean.exporter.checkExportFilename2('TUserFeedback')}">
<p:dataExporter type="csv"  encoding="UTF-8"
	target=":TUserFeedbackListForm:datalist"
	pageOnly="#{userFeedbackBean.exporter.exportContent eq 'page'}"
[...]

PR #5125, opened 2019-09-06
Fix #146,#2363,#3585: Exporter set font for PDF/Excel

### LINKED ISSUE #146: Customizable font in PDFExport
Add option to change font in pdf export

melloware:


tandraschko:
not sure about the default font.... I would make it null or something, is that possible?
JFYI: arial doesn't exist on linux

melloware:
Interesting.  Those were the 2 default fonts PrimeTek had in the code before I refactored.
PDF = Times
Excel = Arial

I don't feel like a NULL font is a good idea so maybe we should find a font that is common to Windows/Linux?

tandraschko:
no idea
what is the default in POI e.g.?

melloware:
OK POI has this in the docs.
https://poi.apache.org/apidocs/dev/org/apache/poi/xssf/usermodel/XSSFFont.html#DEFAULT_FONT_NAME

> By default, Microsoft Office Excel 2007 uses the Calibri font in font size 11

So I will switch to use this constant XSSFFont.DEFAULT_FONT_NAME

DEFAULT_FONT_NAMEBy default, Microsoft Office Excel 2007 uses the Calibri font in font size 11
--

melloware:
No definitely NOT as Cilibri will never be on a linux system as its a Microsoft font.  I can't believe no one noticed this before I just downloaded an XLSX using PrimeFaces Exporter in my production app using Red Hat Linux and the XLSX downloads fine and when I open it the font is "Arial".  So on Red Hat Linux it appears to be working.  

What do you recommend?  Or maybe Excel is smart enough to downgrade if it doesn't find the font listed when you open the excel doc?

tandraschko:
ok, then i think its ok to not change the defaults for now
+1 for the PR

PR #5132, opened 2019-09-10
Fix #1602: GMap expose fitBounds client API.

### LINKED ISSUE #1602: Google Maps Map.fitBounds Overridden by GMap Attribute
When using the Google Maps JavaScript API on a GMap component, Map.fitBounds throws the following error: 

> home.xhtml:48 Uncaught TypeError: PF(...).getMap(...).fitBounds is not a function 

It appears that this is because fitBounds is a GMap attribute that can be set in the XHTML. I found [this workaround](http://stackoverflow.com/questions/23060823/fitbounds-in-primefaces-gmap-component) on Stack Overflow. This fixes the problem for me but it's not obvious or noted in the documentation.

melloware:
Tested and working perfectly.  Also added some missing Ajax Behavior Events to the docs.


tandraschko:
+1

PR #5134, opened 2019-09-10
Fix #933: SelectOneMenu RTL support.

### LINKED ISSUE #933: SelectOneMenu: add direction support
I am working on application which also supports arabic (rtl) but primefaces selectOneMenu does not support dir="rtl".

bug_selectonemenu_rtl

melloware:
Tested in Chrome, IE11, Edge, Firefox.

image


tandraschko:
+1

PR #5138, opened 2019-09-11
Fix #3787: AjaxStatus add delay attribute.

### LINKED ISSUE #3787: ajaxStatus: add attribute "delay"
You can use ajaxStatus to block user from interacting with UI while an ajax call is being made [1]. However, it is very annoying for the user if the ajax call is very fast (it will look like a blink). A delay attribute would make "onstart" to run only after the delay time has passed and only if the ajax is still in progress.
This solution [2] works, but it would be very good to have it native.

[1] https://www.primefaces.org/showcase/ui/ajax/status.xhtml
[2] https://stackoverflow.com/questions/37782947/how-to-delay-primefaces-ajaxstatus-on-jsf

melloware:
Modeled after how AutoComplete handles the delay.
Tested in IE, Chrome, Firefox, and Edge.
Default is 0 ms which is identical to current behavior.

### DIFF
--- a/docs/7_1/components/ajaxstatus.md
+++ b/docs/7_1/components/ajaxstatus.md
@@ -24,6 +24,7 @@ AjaxStatus is a global notifier for ajax requests.
 | oncomplete | null | String | Client side callback to execute after ajax requests complete. |
 | onsuccess | null | String | Client side callback to execute after ajax requests completed succesfully. |
 | onerror | null | String | Client side callback to execute when an ajax request fails. |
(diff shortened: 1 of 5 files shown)


tandraschko:
i would only call setTimeout when delay > 0

tandraschko:
+1 if its working fine

melloware:
Yep it works great and with that default of delay 0 it really doesn't change the behavior of the component at all from previous versions until you give it a delay >0.  Thanks for reviewing and suggestions.

PR #5139, opened 2019-09-11
Fix #97: SelectOneMenu footer facet.

### LINKED ISSUE #97: Footer Facet for SelectOneMenu
Add support for optional footer facet in selectOneMenu.

melloware:


tandraschko:
+1 if well tested
but please add it to the documentation
not sure if we have a list of facets for other components already

melloware:
It looks like we don't really have documentation for facets but I did add a showcase example showing the use of the footer facet.

tandraschko:
feel free to apply the PR but i would really like to have a documentation about facets
selectOneMenu would be great to start (it only has header and footer)

melloware:
OK let me add some docs for the facets

tandraschko:
did you add css classes instead facets to the docu?

melloware:
Good catch forgot to hit Save in my editor.  Added the RTL and Footer Styles.

tandraschko:
+1

PR #5140, opened 2019-09-11
Fix #3165: Spinner add rotate attribute.

### LINKED ISSUE #3165: Spinner: not rotating the values
# IMPORTANT 

- Before you open an issue, test it with the current/newest version.
- Try to find an explanation to your problem by yourself, by simply debugging. This will help us to solve your issue 10x faster
- Clone this repository https://github.com/primefaces/primefaces-test.git in order to reproduce your problem, you'll have better chance to receive an answer and a solution.
- Otherwise the example must be as small and simple as possible! It must be runnable without any other dependencies (like Spring,..., or project/company internal classes)!
- Feel free to provide a PR (Primefaces is an open-source project, any fixes or improvements are welcome.)

## 1) Environment
- PrimeFaces version:6.1
- Affected browsers:

## 2) Expected behavior
in inputspinner field , value has to be rotating(spinning). once the "max" value is reached after incrementing the value , the new value has to be updated to "min" , this is the actual defination of spinner. the values inside spinner should spin.
...

## 3) Actual behavior
once the value is reached to max, after incrementing the value , value is not reset to min.
..

## 4) Steps to reproduce

.. 

## 5) Sample XHTML
<p:spinner min="0" max="100" value="#{numberSpinnerBean.spinnerValue}"/>
..

## 6) Sample bean
private int spinnerValue;

	/**
	 * get spinner value
	 * @return int
	 */
	public int getSpinnerValue() {
		return spinnerValue;
	}

	/**
	 * set spinner value
	 * @param spinnerValue
	 */
	public void setSpinnerValue(int spinnerValue) {
		this.spinnerValue = spinnerValue;
	}	
	
..

melloware:
Tested and working great with rotate=true and rotate=false.  Added showcase example PR.

### DIFF
--- a/docs/7_1/components/spinner.md
+++ b/docs/7_1/components/spinner.md
@@ -33,6 +33,7 @@ widgetVar | null | String | Name of the client side widget.
 stepFactor | 1 | Double | Stepping factor for each increment and decrement
 min | null | Double | Minimum boundary value
 max | null | Double | Maximum boundary value
+rotate | false | Boolean | Rotate to the minimum value when maximum value is reached and vice versa
 prefix | null | String | Prefix of the input
 suffix | null | String | Suffix of the input
 decimalPlaces | null | String | Number of decimal places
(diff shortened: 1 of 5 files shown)


tandraschko:
+1

PR #5142, opened 2019-09-11
Fix #3078: TabView add swipe events

### LINKED ISSUE #3078: Add swipe functionality to TabView
## 1) Environment
- PrimeFaces version: 6.1.10
- Affected browsers:

## 2) Expected behavior
When you swipe left on a TabView, the screen should move one tab to the right and when you swipe right the screen should move one tab to the left.

Note: the TabView menu should also scroll to the right or left so that the currently displayed tab is visible in the menu.

I documented how I added swipe to TabView in the following Forum entry.

https://forum.primefaces.org/viewtopic.php?f=3&t=53503

Note:  In my solution, the TabView menu does not scroll with the page swipe so if the entire menu is not visible you can display a page not visible/selectable on the menu. 

## 3) Actual behavior

At the moment, TabView does not support swipe left or swipe right.

## 4) Steps to reproduce

N/A

## 5) Sample XHTML

N/A

## 6) Sample bean

N/A

melloware:
Follow Carousel as the pattern for adding swipe events.

Test in Chrome mobile.

### DIFF
--- a/src/main/java/org/primefaces/component/tabview/TabView.java
+++ b/src/main/java/org/primefaces/component/tabview/TabView.java
@@ -46,7 +46,8 @@ import org.primefaces.util.MapBuilder;
         @ResourceDependency(library = "primefaces", name = "jquery/jquery.js"),
         @ResourceDependency(library = "primefaces", name = "jquery/jquery-plugins.js"),
         @ResourceDependency(library = "primefaces", name = "core.js"),
-        @ResourceDependency(library = "primefaces", name = "components.js")
+        @ResourceDependency(library = "primefaces", name = "components.js"),
+        @ResourceDependency(library = "primefaces", name = "touch/touchswipe.js")
 })
 public class TabView extends TabViewBase {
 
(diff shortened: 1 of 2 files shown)


tandraschko:
should we make it configurable?

melloware:
My thought was Carousel does not and it lets TouchSwipe detect whether your browser has touch events like IOS or Android.  So on desktop systems swipe is disabled by default.  I can be convinced to make it configurable ?

tandraschko:
no idea :D it was just a question
for me its also ok without configuration

fanste:
This change breaks components like `p:selectCheckboxMenu` inside `p:tab`in IE11. The component uses `mousedown` to show the panel which seems to be affected by the `pointerdown` event that is used by the touchswipe plugin.

Maybe it is enough to add `.ui-selectcheckboxmenu` to the `excludedElements`, I have not tested it

melloware:
Investigating...  For now to temporarily fix it add styleClass="noSwipe" to your selectCheckBoxMenu.

PR #5146, opened 2019-09-12
Fix #97: SelectOneMenu footer facet rendering.

### LINKED ISSUE #97: Footer Facet for SelectOneMenu

melloware:
@Rapster this should do it.


melloware:
Shouldn't theoretically every single call to a facet in the code be doing instead of...

```java
if (facet!= null) {
            facet.encodeAll(context);
}
```
be doing...
```java
if (ComponentUtils.shouldRenderFacet(facet)) {
            facet.encodeAll(context);
}
```

Rapster:
yup

PR #5148, opened 2019-09-12
Fix #910: Galleria accessibility for keyboard and swipe.

### LINKED ISSUE #910: Galleria Component Accessibility with keyboard
It should be possible to navigate through the Galleria images using the keyboard ( at least with the arrow keys ). It would be a nice addition to PF 5.4, as it is mainly focused on accessibility.

melloware:
- Added LEFT and RIGHT keyboard events if the galleria has focus
- Added tabIndex=0 so the galleria can receive focus and keyboard events
- Add touch swipe events for mobile browsers to swipe back and forth.


tandraschko:
should we make the tabindex confiruable?

melloware:
Absolutely let me add tabindex as an attribute and document.

tandraschko:
+1

PR #5157, opened 2019-09-13
Fixes #5149 - Timeline: TimelineUpdater is null when javax.faces.PARTIAL_STATE_SAVING is false

### LINKED ISSUE #5149: Timeline: TimelineUpdater is null when javax.faces.PARTIAL_STATE_SAVING is false
## 1) Environment
- PrimeFaces version: 7.0.7
- Does it work on the newest released PrimeFaces version? No Version? 7.0.7

## 2) Expected behavior
I expect timelineUpdater not to be null

## 3) Actual behavior
timelineUpdater is null

## 4) Steps to reproduce
https://github.com/mores/primefaces-test/tree/timeline001-nullUpdater

         <context-param> 
                <param-name>javax.faces.PARTIAL_STATE_SAVING</param-name> 
        </context-param>

## 5) Sample XHTML
see github

## 6) Sample bean
Timeline source = (Timeline)event.getSource();
TimelineUpdater timelineUpdater = TimelineUpdater.getCurrentInstance( source.getClientId() );
see also github

jasonex7:
Fixes #5149 by moving the logic of setting the map of TimelineUpdaters in FacesContext attribute from `TimelineListener` to `DefaultTimelineUpdater` which is a phase listener, specifically before the APPLY_REQUEST_VALUES phase which it gets always executed no matter what value javax.faces.PARTIAL_STATE_SAVING property has.

### DIFF
--- a/src/main/java/org/primefaces/component/timeline/DefaultTimelineUpdater.java
+++ b/src/main/java/org/primefaces/component/timeline/DefaultTimelineUpdater.java
@@ -106,29 +106,22 @@ public class DefaultTimelineUpdater extends TimelineUpdater implements PhaseList
 
     @Override
     public void beforePhase(PhaseEvent event) {
-        if (event.getPhaseId().equals(PhaseId.RENDER_RESPONSE)) {
-            processCrudOperations(event.getFacesContext());
-        }
-        else if (event.getPhaseId().equals(PhaseId.APPLY_REQUEST_VALUES)) {
+        if (event.getPhaseId().equals(PhaseId.APPLY_REQUEST_VALUES)) {
             populateTimelineUpdater(event.getFacesContext());
+        }
+        else if (event.getPhaseId().equals(PhaseId.RENDER_RESPONSE)) {
+            processCrudOperations(event.getFacesContext());
         }
     }
 
     private void populateTimelineUpdater(FacesContext context) {
-        //This method assumes TimelineListener.processEvent(cse) was called before.
-        Map<String, TimelineUpdater> map
-                = (Map<String, TimelineUpdater>) context.getAttributes().get(TimelineUpdater.class.getName());
+        Map<String, TimelineUpdater> map = (Map<String, TimelineUpdater>) context.getAttributes().get(TimelineUpdater.class.getName());
         if (map == null) {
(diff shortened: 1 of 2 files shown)


melloware (approved):
This looks like a great fix.  @mertsincan , @Rapster @tandraschko can one of you get eyes on this?  This bug has been out there for years even from the PF Extensions days.

jasonex7:
I refactored the method `populateTimelineUpdater` to only add `this` instance instead of traverse viewRoot phase listeners.

src/main/java/org/primefaces/component/timeline/DefaultTimelineUpdater.java
```diff
@@ -101,21 +101,39 @@ public void clear() {
 
     @Override
     public PhaseId getPhaseId() {
-        return PhaseId.RENDER_RESPONSE;
+        return PhaseId.ANY_PHASE;
     }
 
     @Override
     public void beforePhase(PhaseEvent event) {
+        if (event.getPhaseId().equals(PhaseId.APPLY_REQUEST_VALUES)) {
```
melloware: I would flip these around like `PhaseId.APPLY_REQUEST_VALUES.equals(event.getPhaseId())` so the constant is always on the left.

melloware (changes requested):

src/main/java/org/primefaces/component/timeline/DefaultTimelineUpdater.java
jasonex7: OK, good observation. 👍

melloware (approved):

tandraschko:
Will be in holiday for 2 weeks, would be great if @rapster could review

Just one thing from my side: please test that the phaselistener does nothing/is not registred on a View without timeline

jasonex7:
> Just one thing from my side: please test that the phaselistener does nothing/is not registred on a View without timeline

Hello @tandraschko, the phase listener (DefaultTimelineUpdater) is registered by TimelineListener only on PostAddToView event of the Timeline component. This behavior is the same as before within this PR

tandraschko:
+1 then

PR #5164, opened 2019-09-17
Fix #5163: TextEditor add secure attribute.

### LINKED ISSUE #5163: TextEditor: SONATYPE-2018-0365 Issue
Original Security Issue: https://github.com/primefaces/primefaces/issues/3214

Using Sonatype IQ Server they are reporting PF as vulerable with the following explanation.

**Explanation**
PrimeFaces is vulnerable to Cross-Site Scripting (XSS). The decode() method in the TextEditorRenderer class, which is partially responsible for rendering text input into an HTML document, allows arbitrary JavaScript to be contained in input before it is rendered to the page. An attacker can exploit this vulnerability by submitting input containing malicious JavaScript and enticing a victim to visit the resulting page. When the victim visits the page the malicious JavaScript will be invoked, resulting in Cross-Site Scripting.

**Detection**
The application is vulnerable by using this component while org.owasp.html.PolicyFactory from com.googlecode.owasp-java-html-sanitize:owasp-java-html-sanitizer is not on the classpath.

**Recommendation**
There is no non vulnerable version of this component. We recommend investigating alternative components or a potential mitigating control.

_Note:_ Versions 7.RC1 and higher in the 7.x branch contain code to mitigate this vulnerability that is enabled only while org.owasp.html.PolicyFactory from com.googlecode.owasp-java-html-sanitize:owasp-java-html-sanitizer is on the classpath.

**Root Cause**
primefaces-7.0.jar <= org/primefaces/component/texteditor/TextEditorRenderer.class : [7.0.RC1 , )

@mertsincan @tandraschko @cnsgithub @jxmai @Rapster How do you guys feel about us making it secure by default and actually failing to load this component with an error ("OWASP Sanitizer not detected on classpath for security reasons this component is disabled").  Something along those lines.

Basically this CVE is saying we should be secure by default.   And I certainly don't want to add the OWASP Sanitizer by default as noted in that thread because of the guava bloat.  Thoughts?

melloware:
Makes the component secure by default with the ability to opt-opt of security.

### DIFF
--- a/docs/7_1/components/texteditor.md
+++ b/docs/7_1/components/texteditor.md
@@ -36,6 +36,7 @@ style | null | String | Inline style of the editor container.
 styleClass | null | String | Style class of the editor container.
 placeholder | null | String | Placeholder text to show when editor is empty
 toolbarVisible | true | Boolean | Whether the toolbar of the editor is visible.
+secure | true | Boolean | Secure the component with the OWASP HTML Sanitizer library on the classpath.
 allowFormatting | true | Boolean | Whether to allow formatting to be included.
 allowBlocks | true | Boolean | Whether to allow blocks to be included.
 allowStyles | true | Boolean | Whether to allow styles to be included.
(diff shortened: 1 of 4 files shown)


tandraschko:
+1
But also add it to the Migration guide

melloware:
Migration guide updated!

PR #5166, opened 2019-09-17
Fix #5105: InputMask validation enabled by default

### LINKED ISSUE #5105: Add validateMask attribute to InputMask
Defines whether mask pattern would be validated or not on the server side.

validateMask="true|false(default)"

melloware:
Enables security by default and allows the user to opt-out rather than opt-in.


tandraschko:
+1

PR #5167, opened 2019-09-17
Fix #5162: Better fix for resolveStaticField. Fixed unit test.

### LINKED ISSUE #5162: Change in resolveStaticField is not suitable for lazyloaded datatable
Reported By PRO User;
> We have a datatable with lazyloading.
> We use code like
> filterBy="#{orderType}"
> because it does not matter, if property named "orderType" do not exist. It is just constant with which load method works.
> 
> In PF 7 it stopped working. I found this change
> https://github.com/primefaces/primefaces/issues/4185
> which cause #{orderType} to return null as name of filtered field.

Test code;
test.xhtml
```
<h:form id="form">

    <p:dataTable var="car" value="#{dtLazyView.lazyModel}" paginator="true" rows="10"
                 paginatorTemplate="{RowsPerPageDropdown} {FirstPageLink} {PreviousPageLink} {CurrentPageReport} {NextPageLink} {LastPageLink}"
                 rowsPerPageTemplate="5,10,15" selectionMode="single" selection="#{dtLazyView.selectedCar}" id="carTable" lazy="true">
        </p:column>
        <p:column headerText="Year" sortBy="#{car.year}">
        </p:column>
        <p:column headerText="Year from" filterBy="#{anything.yearFrom}">
        </p:column>    
        <p:column headerText="Year to" filterBy="#{anything.yearTo}">
        </p:column>     
        <p:column headerText="Brand">
        </p:column>
        <p:column headerText="Color">
        </p:column>
    </p:dataTable>

</h:form>
```
LazyCarDataModel.java
```
public class LazyCarDataModel extends LazyDataModel<Car> {

  private List<Car> datasource;

  public LazyCarDataModel(List<Car> datasource) {
    this.datasource = datasource;
  }

  @Override
  public Car getRowData(String rowKey) {
    for (Car car : datasource) {
      if (car.getId().equals(rowKey))
        return car;
    }

    return null;
  }

  @Override
  public Object getRowKey(Car car) {
    return car.getId();
  }

  @Override
  public List<Car> load(int first, int pageSize, String sortField, SortOrder sortOrder, Map<String, Object> filters) {
    List<Car> data = new ArrayList<Car>();

    int yearFrom = 0;
    int yearTo = 4000;

    if (filters.containsKey("yearFrom")) {
      String filterProperty = "yearFrom";
      Object filterValue = filters.get(filterProperty);

      if (filterValue != null) {
        yearFrom = Integer.valueOf(filterValue.toString());
      }
    }

    if (filters.containsKey("yearTo")) {
      String filterProperty = "yearTo";
      Object filterValue = filters.get(filterProperty);

      if (filterValue != null) {
        yearTo = Integer.valueOf(filterValue.toString());
      }
    }

    // filter
    for (Car car : datasource) {
      if (yearFrom <= car.getYear() && car.getYear() <= yearTo) {
        data.add(car);
      }
    }

    // sort
    if (sortField != null) {
      Collections.sort(data, new LazySorter(sortField, sortOrder));
    }

    // rowCount
    int dataSize = data.size();
    this.setRowCount(dataSize);

    // paginate
    if (dataSize > pageSize) {
      try {
        return data.subList(first, first + pageSize);
      } catch (IndexOutOfBoundsException e) {
        return data.subList(first, first + (dataSize % pageSize));
      }
    } else {
      return data;
    }
  }
}
```
The other codes; https://www.primefaces.org/showcase/ui/data/datatable/lazy.xhtml

melloware:


melloware:
@rapster I am merging this for now to fix the broken tests.  We can always provide another PR later but it sounds like @martin654 had a real world use case that PrimeTek had to restore.

Rapster:
Yup no worries @melloware it is temporary anyway, I'll have a look at several things once I'm back from vacation. I need to read up about how filters work in PF, I might be missing something but the use case provided by Mert is strange, does not feel right... Having 3 times the same column is UI/UX speaking: wrong. And I don't think filters was meant to be used that way, at least I didn't find anything in the doc or in stackoverflow

Issue #5159, opened 2019-09-16
PanelGrid: new layout(-mode) "flex"


christophs78:
PanelGrid has currently the layout-modes "table" and "grid".
Grid is based on the "old" grid-CSS, which has a few corner-cases because it´s based on the CSS-float-mechanism.
Think it would make sense to provide a additional "flex-grid" - layout-mode. (I´ll provide a PR for this within the next month.)

One question: PrimeFlex is currently not provided OOTB with PrimFaces. We may document that this new layout-mode requires PrimeFlex. A other option would be to provide PrimeFlex OOTB starting with PF 7.1.

tandraschko:
Related to #1820

melloware:
@christophs78 Much appreciated for the future PR.  To answer your questions...

1. Assume PrimeTek does not include PrimeFlex out of the box we can add it to the documentation possible even be clever and put a JS Console PrimeFaces.error() if we detect that primeflex.css is not loaded.

2. I would just call the layout mode "flex".

3. I would be amazing if you could fix #1820 while you are adding PrimeFlex!

christophs78:
Probably https://github.com/primefaces/primefaces/issues/1820 was already fixed:
https://github.com/primefaces/primefaces/commit/9f880385fa26287737eb92f79467d232057f40de#diff-d7a99593a8e92d9c69f098585d87dcfa
https://github.com/primefaces/primefaces/issues/2677

melloware:
You are right I just tested and confirmed.  Thanks I closed the other ticket!

martin654:
Hi, can I have a question why **p-nogutter** is used for rows? ( FLEX_ROW_CLASS = "p-grid p-nogutter"; ) What if user wants some gutter? Shouldn't this be optional? Thanks

christophs78:
Do you have a concrete usecase where it would be better without p-nogutter?
When you look at the showcase-page https://www.primefaces.org/showcase/ui/panel/panelGrid.xhtml i think we are fine with p-nogutter.
From a technical point of view it would be easy to remove p-nogutter from FLEX_ROW_CLASS. After introducing layout="flex" we also added the new attribute "contentStyleClass". So it would be easy the add p-nogutter if necessary.

martin654:
Well, in our case we used layout="grid" and switched to layout="flex". In "grid" layout we had configured padding between "cells" of the grid and we wanted to use them as well in "flex" mode. 

[...]

PR #5168, opened 2019-09-17
#5159: PanelGrid: new layout(-mode) "flex"

christophs78:
Showcase-PR: https://github.com/primefaces/showcase-facelift/pull/65

### DIFF
--- a/src/main/java/org/primefaces/component/panelgrid/PanelGridRenderer.java
+++ b/src/main/java/org/primefaces/component/panelgrid/PanelGridRenderer.java
@@ -250,10 +250,15 @@ public class PanelGridRenderer extends CoreRenderer {
         ResponseWriter writer = context.getResponseWriter();
         String columnClassesValue = grid.getColumnClasses();
         String[] columnClasses = columnClassesValue == null ? new String[0] : columnClassesValue.split(",");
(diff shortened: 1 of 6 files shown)


src/main/java/org/primefaces/component/panelgrid/PanelGridRenderer.java
```diff
@@ -39,15 +39,19 @@
...
+    public static final String LAYOUT_GRID = "grid";
+    public static final String LAYOUT_FLEX = "flex";
+
     @Override
     public void encodeEnd(FacesContext context, UIComponent component) throws IOException {
         PanelGrid grid = (PanelGrid) component;
 
-        if (grid.getLayout().equals("tabular")) {
+        if (grid.getLayout().equals(LAYOUT_TABULAR)) {
             encodeTableLayout(context, grid);
```
melloware, line 51: Reverse like you have below LAYOUT_TABLE.equals(grid.getLayout()).  Also should we do equalsIgnoreCase() everywhere on these so if someone accidentally puts "Flex" instead of "flex they don't get dinged?

melloware (changes requested):

melloware (approved):

tandraschko:
+1

Issue #4626, opened 2019-03-13
DatePicker: pattern including date and time


carlbradwell17:
## 1) Environment
- PrimeFaces version: 7.0-RC3

## 2) Expected behavior
If i want to use a datePicker to set a date object which includes the time, I should set the showTime attribute to true to allow the user to select a time. I then should be able to set a pattern parameter, to format the datetime in the input. 
It is possible to pass a pattern for the time when the datePicker is set to timeOnly="true".
The behaviour of accepting a pattern with date and time is how the calendar component works,

## 3) Actual behavior

Setting showtime to true adds the time to the end of the value. The pattern attribute therefore only seems to affect the date portion. I therefore have no way of customising the formatting of the time portion. 

## 4) Steps to reproduce

create a datePicker with showTime set to true and a pattern including time. E.g. if I click 12th march 2019, and set the time to 11:09, the value displayed in the input is "12-Mar-2019 HH:03 11:09". The value cannot be set on the date object, i.e. nothing happens when submitting the form.

## 5) Sample XHTML

<h:form>
 <p:datePicker value="#{testBean.date1}" showTime="true" pattern="dd-MMM-yyyy HH:mm" />
 <p:commandButton action="#{testBean.submit}" value="submit" />	
			
</h:form>
## 6) Sample bean

@Named
@ViewScoped
public class TestBean {
private Date date1;

public String submit() {
		return null;
	}
public Date getDate1() {
        return date1;
    }
 
    public void setDate1(Date date1) {
        this.date1 = date1;
    }
}

christophs78:
This was solved with https://github.com/primefaces/primefaces/pull/5191.

Issue #4625, opened 2019-03-13
DatePicker: Time is not set on Date Object


carlbradwell17:
## 1) Environment
- PrimeFaces version: 7.0.0-RC3

## 2) Expected behavior
When I use a datePicker to set a Date object on a bean, both the date and time should be set.

## 3) Actual behavior

The date is set, but the time is always 00:00:00.
E.g. if I set the datepicker to "14/03/19 11:57", when I inspect the Date object, the value is "Thu Mar 14 00:00:00 GMT 2019".

## 4) Steps to reproduce

Create a datepicker, with no options set other than showTime="true". Set a date and time and then check the Date object after submission.

## 5) Sample XHTML
<h:form>
<p:datePicker value="#{testBean.date1}" showTime="true" />
<p:commandButton action="#{testBean.submit}" value="submit" />	
</h:form>

## 6) Sample bean

@Named
@ViewScoped
public class TestBean {
private Date date1;

	public String submit() {
		return null;
	}
public Date getDate1() {
        return date1;
    }
 
    public void setDate1(Date date1) {
        this.date1 = date1;
    }

}

christophs78:
Applied a fix/work-around for this (for LocalDateTime) during my work on java.time-support.

christophs78:
Additional improvements with https://github.com/primefaces/primefaces/pull/5191

melloware:
@christophs78 Is this safe to mark closed?

PR #5172, opened 2019-09-18
Add support for InputNumber and additional constraints. Fixes #5165

### LINKED ISSUE #5165: InputNumber and Spinner: support additional bean validation constraints
## 1) Environment
- PrimeFaces version: 7.0
- Affected browsers: all

## 2) Issue Description
Right now, only Spinner is considered in [BeanValidationInputMetadataTransformer](https://github.com/primefaces/primefaces/blob/70a71f9524dd68d065198c9c442d80253274c9c1/src/main/java/org/primefaces/metadata/transformer/impl/BeanValidationInputMetadataTransformer.java) :
https://github.com/primefaces/primefaces/blob/70a71f9524dd68d065198c9c442d80253274c9c1/src/main/java/org/primefaces/metadata/transformer/impl/BeanValidationInputMetadataTransformer.java#L91-L102
InputNumber should be included too.
Adding `@DecimalMax` and `@DecimalMin` (and maybe `@Digits`) annotations support would be useful too.

NicolaIsotta:
Fixes #5165


src/main/java/org/primefaces/metadata/transformer/impl/BeanValidationInputMetadataTransformer.java
```diff
@@ -91,23 +99,64 @@ protected void applyConstraint(ConstraintDescriptor constraintDescriptor, UIInpu
...
                 Min min = (Min) constraint;
                 spinner.setMin(min.value());
             }
+            if (annotationType.equals(DecimalMax.class) && spinner.getMax() == Double.MAX_VALUE) {
+                DecimalMax max = (DecimalMax) constraint;
+                try {
+                    spinner.setMax(Double.parseDouble(max.value()));
+                }
+                catch (NumberFormatException ex) {
+                }
```
melloware, line 117: I would add `LOGGER.log(Level.WARN, "Failed setting Spinner max value: " + ex.getMessage());` I don't like empty try catches.

src/main/java/org/primefaces/metadata/transformer/impl/BeanValidationInputMetadataTransformer.java
```diff
@@ -91,23 +99,64 @@ protected void applyConstraint(ConstraintDescriptor constraintDescriptor, UIInpu
...
+            }
+            if (annotationType.equals(DecimalMin.class) && spinner.getMin() == Double.MIN_VALUE) {
+                DecimalMin min = (DecimalMin) constraint;
+                try {
+                    spinner.setMin(Double.parseDouble(min.value()));
+                }
+                catch (NumberFormatException ex) {
+                }
```
melloware, line 126: I would add `LOGGER.log(Level.WARN, "Failed setting Spinner min value: " + ex.getMessage());` I don't like empty try catches.

melloware (changes requested):

tandraschko (commented):
Is DecimalMin in BVal 1.0?

melloware:
> Is DecimalMin in BVal 1.0?

Yes DecimalMin and Max are in 1.0.

tandraschko:
+1
But please also change the Date to LocalDate in this class

NicolaIsotta:
> +1
> But please also change the Date to LocalDate in this class

Sure. Two questions:
- Since time could be in the calendar pattern, maybe LocalDateTime is better?
- Should I consider BV 2.0 new time constraints (`@PastOrPresent` and `@FutureOrPresent`)? We can use class names like in OutputLabelRenderer so no need to change dependency.

tandraschko:
1) Not sure. CC @christophs78
2) yes. We can not upgrade as PF would not work on JavaEE6/7 then

christophs78:
Calendar and DatePicker currently (7.1-dev; Master-Branch) supports LocalDate as minDate and maxDate-values. (And java.util.Date for backward-compatibility-reasons.)
See [https://primefaces.github.io/primefaces/7_1/#/components/datepicker?id=datepicker](https://primefaces.github.io/primefaces/7_1/#/components/datepicker?id=datepicker)
Maybe we should add support for LocalDateTime for minDate and maxDate-Attributes for convenience-reasons. And for edge-cases where Calendar and DatePicker are used with showTime=true. And maybe we should support LocalTime for timeOnly=true.

NicolaIsotta:
Updated with LocalDate e BV 2.0 new time contraints.

tandraschko:
+1 
Can someone retest it?
I assume it also works with DatePicker?

melloware:
It should work with DatePicker, @NicolaIsotta can you confirm and then I will take a pass at testing it.

NicolaIsotta:
It's not ready to be tested as DatePicker and LocalDateTime/java.util.Date combination seems a bit broken ATM (see #4625 #4626 and relative fixes)

tandraschko:
Whats happening?
@christophs78 can you assist pls?

christophs78:
@NicolaIsotta: Applied a fix/work-around for 4625 (for LocalDateTime) during my work on java.time-support. Please look into [https://github.com/primefaces/showcase-facelift/blob/master/src/main/webapp/ui/input/datePickerJava8.xhtml](https://github.com/primefaces/showcase-facelift/blob/master/src/main/webapp/ui/input/datePickerJava8.xhtml). Compare this with [https://github.com/primefaces/showcase-facelift/blob/master/src/main/webapp/ui/input/datePicker.xhtml](https://github.com/primefaces/showcase-facelift/blob/master/src/main/webapp/ui/input/datePicker.xhtml)
Could you look please look into this and give me feedback? (I know this solution is not ideal, but it´s a work-around without needing to to a deepdive into the clientside JS.) 

@tandraschko, @mertsincan: Is it possible to deploy the 7.1-dev-Showcase? Think it would be ideal to have the 7.0-Release Showcase deployed and in parallel the 7.1-dev-Showcase?

@NicolaIsotta: I know there is some room for improvement (as already mentioned) for Type-Support for minDate/maxDate-Attributes. I will work on this sometime in october.

NicolaIsotta:
I'll open a new issue for datepicker to avoid cluttering this conversation.

> @NicolaIsotta: I know there is some room for improvement (as already mentioned) for Type-Support for minDate/maxDate-Attributes. I will work on this sometime in october.

mindate and maxdate already support LocalDateTime, no need for additional work.

tandraschko:
+1 
Can you merge plz melloware?

PR #5173, opened 2019-09-18
Fix #4787: CSP replaced illegal use of new Function()

### LINKED ISSUE #4787: ConfirmDialog: uncaught EvalError if CSP is enabled
## 1) Environment
- PrimeFaces version: 7.1-SNAPSHOT

## 2) Expected behavior
`confirmDialog` working correctly if CSP is enabled.

## 3) Actual behavior
image

image

## 4) Steps to reproduce
- Enable CSP
- Go to http://localhost:8080/showcase/ui/overlay/confirmDialog.xhtml
- Click at _Yes_

melloware:
Tested with CSP enabled and disabled.  ConfirmDialog is working exactly as it was before and CSP violation is now fixed.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
+++ b/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
@@ -833,18 +833,11 @@ if (!PrimeFaces.ajax) {
             doEval : function(node, xhr) {
                 var textContent = node.textContent || node.innerText || node.text;
 
+                var nonce;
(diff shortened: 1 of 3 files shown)


tandraschko:
Will review it after holiday

tandraschko:
What i dont like is that a widget needs to call "PrimeFaces.csp.*"
Maybe we should create a eval method in core.js directly, need to think about it.

melloware:
No problem.   I didn't know where else to put it and putting it in CSP felt like the logical place since it deals with a NONCE?  Let me know what you decide.

PR #5176, opened 2019-09-20
Fix #5175: Apache Tika 1.22. Fixed Tika classloading issue.

### LINKED ISSUE #5175: FileUpload: Tika Issues
Based on this forum thread: https://forum.primefaces.org/viewtopic.php?f=3&t=59996

I can confirm that just having Tika on the classpath is not currently enough to force it to load and be used in `Files.probeContentType`.

Also currently Tika 1.19.1 is being used and Tika 1.22 is the latest that fixes multiple CVE's.
CVE's fixed in 1.22: https://www.cvedetails.com/vulnerability-list/vendor_id-45/product_id-35375/year-2019/Apache-Tika.html

melloware:
- Upgraded Tika 1.22
- Command Lang was removed from Tika so I fixed GenerateAjaxBehaviorDoc to not use it
- xml-apis conflict with new Tika was causing WebXmlParserTest to fail so in the pom.xml I had to force the version to 1.4.01 to have the test pass.  Its TEST scope so it's not affecting anything else.

@cnsgithub If you are around please review.


tandraschko (commented):
I think you need to cache the serviceloader result for perf? Please check
PrimeApplication probably

tandraschko:
Maybe also move "tika available" to PrimeEnvironment

melloware:
Great suggestion!  I had to rework the unit test but I got both Tika and TikaDetector cached in PrimeEnvironment.

tandraschko (commented):
I think the filetypedetector belongs more to PrimeApplication

melloware:
Made the unit test a little trickier but you are right PrimeApplicationContext was the right place for it.

tandraschko:
1) TIKA_CLASS is to common for me
2) fileTypeDetector in PrimeApplication doesnt need to be  type of Lazy, if' its no lazy implemented :P

melloware:
I changed to TIKA_FILE_DETECTOR_CLASS and removed the Lazy.  I opened a different PR and closing this one so I could incorporate all the latest FileUpload changes.

PR #5177, opened 2019-09-20
Fix #3020: SplitButton ARIA accessibility.

### LINKED ISSUE #3020: SplitButton: accessibility
## 1) Environment
- PrimeFaces version: Primefaces 6.1
- Application server + version: JBoss EAP 7.0
- Affected browsers: Firefox ESR 52.5.0 (64-Bit)
- Screenreader: JAWS 18.0

## 2) Expected behavior
<p:splitButton> main button should be read like a standard button. Dropdown button should be read as "Dropdown". Dropdown menu should be accessible only by keyboard.

## 3) Actual behavior
<p:splitButton> main button is read as something like "UI button switch". Dropdown toggler is not read as "Dropdown". Also I can not select any of the dropdown values with keyboard because the dropdown is immediately closed.

## 4) Steps to reproduce
Read <p:splitButton> component with screenreader

melloware:


tandraschko:
+1

PR #5180, opened 2019-09-23
Fix #5179: Document CSP.

### LINKED ISSUE #5179: Document CSP
@melloware interessted?

melloware:
@cnsgithub @tandraschko this was my starting point for CSP doc.  Feel free to suggest any changes or any other information I am leaving out.

### DIFF
--- a/docs/7_1/core/contentsecuritypolicy.md
+++ b/docs/7_1/core/contentsecuritypolicy.md
@@ -19,6 +19,32 @@ CSP is disabled by default and a global parameter is required to turn it on.
     <param-value>true</param-value>
 </context-param>
 ```
+## Policy
+There are many ways to configure CSP for different levels of security. Currently, PrimeFaces has chosen to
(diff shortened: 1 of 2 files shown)


tandraschko:
I think we should mention that we decided for the "nonce" way instead hashes. Also we currently only do it for Scripts?

AFAIR f:ajax doesnt work, right? (We added some support to MyFaces 3.x but one thing is still not working)

melloware:
I will add sections about NONCE vs hashes and that we currently only support script-src.

As for `f:ajax` should I mention that its a limitation of the Faces implementations and that `p:ajax` must be used?

tandraschko:
I would do it for now, yep

tandraschko:
+1 i Like it
Please put in core

Thanks, really appreciate your many contributions

melloware:
No problem.  You did all the hard work...is a huge security win for PrimeFaces!

PR #5191, opened 2019-09-26
Fix #5186: Datepicker improve pattern handling

### LINKED ISSUE #5186: DatePicker: pattern messed with LocalDateTime and showTime="true"
## 1) Environment
- PrimeFaces version: trunk
- Application server + version: jetty 9.20
- Affected browsers: all

## 2) Expected behavior
The user-provided pattern is not modified.

## 3) Actual behavior
HH:mm is added to the pattern, causing conversion errors.

## 4) Steps to reproduce

.. 

## 5) Sample XHTML
```xhtml
<h:form>
	<div>
		<p:outputLabel for="date1" value="Date1" style="display: inline-block; width: 120px;"/>
		<p:datePicker id="date1" value="#{testViewCdi.date1}" showTime="true"/>
		<p:message for="date1"/>
	</div>
	<div>
		<p:outputLabel for="date2" value="Date2" style="display: inline-block; width: 120px;"/>
		<p:datePicker id="date2" value="#{testViewCdi.date2}" showTime="true" pattern="dd/MM/yy HH:mm"/>
		<p:message for="date2"/>
	</div>
	<div>
		<p:outputLabel for="date3" value="Date3" style="display: inline-block; width: 120px;"/>
		<p:datePicker id="date3" value="#{testViewCdi.date3}"/>
		<p:message for="date3"/>
	</div>
	<div>
		<p:outputLabel for="date4" value="Date4" style="display: inline-block; width: 120px;"/>
		<p:datePicker id="date4" value="#{testViewCdi.date4}" pattern="dd/MM/yy"/>
		<p:message for="date4"/>
	</div>
	<p:commandButton id="sub" value="submit" update="@form"/>
</h:form>
```

## 6) Sample bean
```java
@Named
@ViewScoped
public class TestViewCdi implements java.io.Serializable {

	private java.time.LocalDateTime date1 = java.time.LocalDateTime.now();

	private java.time.LocalDateTime date2 = java.time.LocalDateTime.now();

	private java.time.LocalDateTime date3 = java.time.LocalDateTime.now();

	private java.time.LocalDateTime date4 = java.time.LocalDateTime.now();

	// getters and setters
}
```

NicolaIsotta:
Fixes #5186


melloware:
@christophs78 please review!

src/main/java/org/primefaces/component/datepicker/DatePickerBase.java
```diff
@@ -386,4 +387,35 @@ public void setOnYearChange(String onYearChange) {
     public boolean hasTime() {
         return this.isShowTime() || this.isTimeOnly() || super.hasTime();
     }
-}
\ No newline at end of file
+
+    @Override
+    public String calculatePattern() {
```
christophs78, line 392: Please add some test-methods for this method. Think we should cover usecases like usedefinied dateFormat with time (-> removeTime) and the calculated TimeOnlyPatterns for 12 and 24 hour-format.

christophs78: Please add a few tests for this method. Eg userdefinied pattern where we need to remove time. TimeOnlyPattern with 12 and 24 hours.

christophs78 (changes requested):
I am on holidays with my family. So only limited time and code-review on my iPad. :-)
First of all thanks for your work. And thanks for fixing a few things i missed when moving Calendar and DatePicker to java.time!
Overall your work looks really excellent!
Could you add a few mutations of getValueAsString_LocalDateTime to CalendarUtilsTest?

christophs78:
Please add the issue-number to the name/label of this PR.

melloware:
Title updated.

NicolaIsotta:
I'll add some test as soon as I can.
While testing I discovered the original java 8 PR ( #4782 ) broke `p:calendar` when `timeOnly="true"`.
I'll try to fix that too.

christophs78:
Seems like i broke UICalendar#calculateTimeOnlyPattern with the java 8 PR.

Generated client-side JS with 7.0:
`pf.cw("Calendar","widget_form_time",{id:"form:time",popup:true,locale:"en",dateFormat:"m\/d\/y HH:mm",timeOnly:true,`

Generated client-side JS with 7.1-dev:
`pf.cw("Calendar","widget_form_time",{id:"form:time",popup:true,locale:"en",dateFormat:"HH:mm",timeOnly:true`

christophs78:
Maybe we should use your new method calculateWidgetPattern also for CalendarRenderer. (Maybe with some modifications.)

NicolaIsotta:
I've already fixed it, but I've not commited yet as it's still broken with LocalTime (p:calendar always expects a date part). I'll have to make some special workaround

src/main/java/org/primefaces/util/CalendarUtils.java
```diff
@@ -241,18 +254,12 @@ else if (value instanceof LocalDate || value instanceof LocalDateTime || value i
...
-                    pattern += " HH:mm";
-                }
-                dateTimeFormatter = DateTimeFormatter.ofPattern(pattern, calendar.calculateLocale(context));
-
                 return ((LocalDateTime) value).format(dateTimeFormatter);
             }
             else if (value instanceof LocalTime) {
+                if (forCalendarInput) {
+                    return LocalDateTime.of(LocalDate.now(), (LocalTime) value).format(dateTimeFormatter);
+                }
```
christophs78: I don´t think, it´s necessary to add the new forCalendarInput parameter.
We should be able to to something like `if (calendar instanceof Calendar) {`

NicolaIsotta: Yeah you're right, it simpler with instanceof

christophs78 (approved):
Tested your PR on my machine against the showcase. Everything looks fine.
We should add a few test´s to ensure this a bit complex "construct" does not break in the future. But we do not necessarily need to to this in this PR. I (or we) can add additional test´s with a future PR.

NicolaIsotta:
I'll commit some tests later today

christophs78:
@melloware - From my perspective this PR is finished and ready to be merged.

melloware:
@tandraschko any objection?  I have looked it over and it looks good and they added a bunch of unit tests to it as well.

melloware (approved):

PR #5192, opened 2019-09-26
Fix #5189: CSP self directive.

### LINKED ISSUE #5189: CSP: Not working in MS Edge
In the Console it is reporting: 

> CSP14312: Resource violated directive 'script-src 'nonce-ZGEyMTlhODAtMjY4Yi00OTMyLWIwMTAtM2E5YjcxMThlMzMy'' in Content-Security-Policy: http://localhost:8080/showcase/javax.faces.resource/jquery/jquery.js.xhtml?ln=primefaces&v=7.1-SNAPSHOT. Resource will be blocked.

See this post for why: https://github.com/elastic/kibana/issues/30988#issuecomment-463786740

Apparently CSP is broken in Edge and Microsoft probably won't fix it until they release their new browser.   So the questions are:

1. Should we fix it like that post IF browser is Edge add all those unsafe directives?

2. Should we just do nothing since all the other browsers work and leave Edge unsupported until they fix the issue?

melloware:
It turns out it does work with the latest Edge we were just missing the "self" directive to apply the nonce the page's own domain.

See: https://blogs.windows.com/msedgedev/2017/01/10/edge-csp-2/

I tested in IE11, Edge, Chrome, and Firefox and CSP is working correctly.

### DIFF
--- a/src/main/java/org/primefaces/csp/CspPhaseListener.java
+++ b/src/main/java/org/primefaces/csp/CspPhaseListener.java
@@ -63,7 +63,7 @@ public class CspPhaseListener implements PhaseListener {
             CspState state = PrimeFacesContext.getCspState(context);
 
             HttpServletResponse response = (HttpServletResponse) externalContext.getResponse();
-            response.addHeader("Content-Security-Policy", "script-src 'nonce-" + state.getNonce() + "'");
+            response.addHeader("Content-Security-Policy", "script-src 'self' 'nonce-" + state.getNonce() + "'");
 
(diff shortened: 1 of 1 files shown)


tandraschko:
What about external scripts like g-analytics with nonce? Does it still work?

melloware:
Well there is still this open issue you discussed about whitelisting with @cnsgithub that you guys never resolved. : https://github.com/primefaces/showcase-facelift/issues/39
I was going to ask you about that.

tandraschko:
OK
Lets talk about csp Missing stuff in 2 weeks
Please Ping me

For now i would merge this PE

melloware:
Understood.  Go back to your holiday!

PR #5194, opened 2019-09-27
Fix #5183: CSP .trigger('change') call native DOM event.

### LINKED ISSUE #5183: CSP: Spinner not registering change event
PrimeFaces version: 7.1-SNAPSHOT

**CSP Disabled:**
- Go to the Showcase spinner example: http://localhost:8080/showcase/ui/input/spinner.xhtml
- On the Ajax Spinner press up or down and you will see the Ajax fire and the label updated

**CSP Enabled:**
- Go to the Showcase spinner example: http://localhost:8080/showcase/ui/input/spinner.xhtml
- On the Ajax Spinner press up or down and you will see no Ajax events fired.

I can see PrimeFaces registered the change event for the spinner.
```javascript
PrimeFaces.csp.register('j_idt710:ajaxSpinner_input',
  'onchange',function(event){PrimeFaces.ab({s:"j_idt710:ajaxSpinner",e:"valueChange",f:"j_idt710",p:"j_idt710:ajaxSpinner",u:"j_idt710:ajaxSpinnerValue"});}
);
```

However in the spinner code when its triggering the 'change' event nothing is happening.
```javascript
$this.input.trigger('change');
```

**Browsers:**
_IE11, Chrome, Firefox, _Edge:_ are all not firing the event

melloware:
There might be a better way to do this but here is what is happening.

First when `jq.trigger('change');` is being called that only looks for Jquery events to trigger or any registered "onChange=''" handlers on a DOM element. It does **NOT** call native DOM registered events.

So what this fix does is..

1. override the native .trigger() call in Jquery so we don't have to touch any code.

2. It detects whether there is a JQuery event registered for "focus" or "change" etc and will call the Jquery .trigger().

3. If no Jquery event is detected it then tries to call a Native DOM event using .triggerNative().

I am sure there might be a better way to do this but I have tested that this works so far to fix the CSP issue.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/core/core.csp.js
+++ b/src/main/resources/META-INF/resources/primefaces/core/core.csp.js
@@ -3,6 +3,7 @@ if (!PrimeFaces.csp) {
     PrimeFaces.csp = {
 
         NONCE_INPUT : "primefaces.nonce",
+        TRIGGERS_HANDLED: false,
 
         init : function(nonce) {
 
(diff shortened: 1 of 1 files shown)


tandraschko:
Need to check it next week, feels quite hacky

melloware:
I totally agree.  will appreciate the help next week.

Issue #2086, opened 2017-02-15
Dialog Framework: dialogReturn is not called back in some case


zsuiqiang:
# IMPORTANT 
- If you open an issue, fill every item. Otherwise the issue might be closed as invalid.
- Before you open an issue, test it with the current/newest version.
- Try to find an explanation to your problem by yourself, by simply debugging. This will help us to solve your issue 10x faster
- Clone this repository https://github.com/primefaces/primefaces-test.git in order to reproduce your problem, you'll have better chance to receive an answer and a solution.
- Feel free to provide a PR (Primefaces is an open-source project, any fixes or improvements are welcome.)

## 1) Environment
- PrimeFaces version:6.0.15
- Does it work on the newest released PrimeFaces version? Version?No
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source)
- Application server + version: Glassfish v4.1
- Affected browsers: Chrome or Edge

## 2) Expected behavior
"dialogReturn" callback correctly.

## 3) Actual behavior
    In some case,it work incorrectly (not callback).
## 4) Steps to reproduce
    If i open  host page Independently，all work correctly. But if i open host page in iframe, there are some problems.

- Case one:
     Open dialog by "openDlgLink" ,when i click "returnDlgBtn" in dialog page, method "createPlanFromStd" is invoked,but method "dialogReturn" is not called back.

- Case two:
     Open dialog by "openDlgBtn" ,when i click "returnDlgBtn" in dialog page, method "createPlanFromStd" is not invoked,and method "dialogReturn" is not called back.
In this case,i get error info: 
```
core.js.jsf?ln=primefaces&v=6.0.15:1  Widget for var 'openDlgBtn' not available!
components.js.jsf?ln=primefaces&v=6.0.15:1 

Uncaught TypeError: Cannot read property 'cfg' of undefined
    at Object.closeDialog (components.js.jsf?ln=primefaces&v=6.0.15:1)
    at Object.closeDialog (core.js.jsf?ln=primefaces&v=6.0.15:1)
    at eval (eval at <anonymous> (jquery.js.jsf?ln=primefaces&v=6.0.15:14), <anonymous>:1:12)
    at eval (<anonymous>)
    at jquery.js.jsf?ln=primefaces&v=6.0.15:14
    at Function.globalEval (jquery.js.jsf?ln=primefaces&v=6.0.15:14)
[...]

PR #5204, opened 2019-09-28
Fix #5193, #2086: Capture current frame position of source widget for dialog framework

### LINKED ISSUE #5193: Dialog Framework: Closing dialog doesn't work when in iframe
I'm on Primefaces 7.0 and this issue is related to https://github.com/primefaces/primefaces/issues/1403 and https://github.com/primefaces/primefaces/issues/1776

I have a JSF page that contains an iframe with a JSF page as src. The nested JSF page now opens a dialog, which works, but when closing, it tries to find the sourceWidget which it can't because the dialog is attached to the root window, but the sourceWidget is in a sub-frame.

main.xhtml
```
<iframe src="some.xhtml"></iframe>
```

some.xhtml
```
<p:commandButton value="open dialog" actionListener="#{bean.openDialog()}"/>
```

dialog.xhtml
```
<p:commandButton value="close dialog" actionListener="#{bean.closeDialog()}"/>
```

Bean.java
```
public void openDialog() {
  PrimeFaces.current().dialog().openDynamic("dialog.xhtml", options, null);
}

public void closeDialog() {
  PrimeFaces.current().dialog().closeDynamic(null);
}
```

Here the sourceWidget is `undefined`: https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/core/core.dialog.js#L165

I'm not sure how this could be fixed, maybe capture the frame position of the source widget and resolve the frames form the root?

beikov:


tandraschko:
wouldn't it be easier to remember the src of the iframe? 
can it change during the "conversation" until dialogReturn?

beikov:
Not sure how that helps. You need to locate the source widget within the frame which could be deeply nested.

tandraschko:
Then please explain what exactly sourceFrames does :D

beikov:
Let's see if I can visualize it

```
----------------------
| Window             |
|  ------------      |
|  | Frame 1  |      |
|  ------------      |
|                    |
|  ----------------- |
|  | Dialog Frame  | |
|  ----------------- |
----------------------
```

When the button(source widget) that opens the dialog is part of _Frame 1_ sourceFrames is the frame path from the root window to the frame content. In the test case, it's e.g. `html > body > iframe` but when the iframe has an id it could simply be `#iframeId`. We use this to get access to the source widget object when clicking on close in the dialog frame. The current implementation assumes the source widget is always part of the root window, which is not the case. That's why we need to capture the path from the root window to the frame in which the source widget is contained so that we can resolve that then.

tandraschko:
What about nested frames?
Like:
-window
--iframe1
---iframe1.1
---iframe1.2
--iframe2

beikov:
It will capture the full path. If the source button is in iframe 1.1 it could be `['html > body > iframe', '#iframe1_1']`

tandraschko:
ok, fine
can you add a method docu do it, what it actually does?
I will merge then. Can you also give it a try @melloware ?

beikov:
Done

melloware:
@tandraschko I can confirm this code works to fix the issue. I think its safe to merge.

melloware:
Tested on IE11, Edge, Firefox, and Chrome.

src/main/resources/META-INF/resources/primefaces/core/core.dialog.js
```diff
@@ -159,6 +226,10 @@ if (!PrimeFaces.dialog) {
             }
             else {
                 windowContext = rootWindow;
+                var frames = dlgWidget.cfg.sourceFrames;
+                for (var i = 0; i < frames.length; i++) {
+                    windowContext = $(windowContext.document).find(frames[i]).get(0).contentWindow;
```
tandraschko, line 271: why is this executed in a loop?

src/main/resources/META-INF/resources/primefaces/core/core.dialog.js
```diff
@@ -12,6 +12,72 @@ if (!PrimeFaces.dialog) {
...
+
+                    var s = pieces.slice(1).join('');
+
+                    sourceFrames.unshift(s);
+                    w = parent;
+                };
+
+                return sourceFrames;
+            }();
+
```
tandraschko, line 118: its still very hard for me to understand the whole thing. So many arrays, shifts unshifts.
Why sourceFrames is array?

In generell we need a bit more readable code, better comments, it would be unmaintable for me.

tandraschko:
@beikov could you please try to work on it?
I would like to get it into 7.1RC1, which is in ~2 weeks.

beikov:
IMO this is done @tandraschko. Do you have anything in mind that still needs to be done?

beikov:
IMO this is done @tandraschko. Do you have anything in mind that still needs to be done?

melloware (approved):

tandraschko:
Please see my inline comments.

beikov:
I'm sorry, but can you post a link? Which inline comments?

melloware:
I am not seeing any inline comments either?

tandraschko:
better? :D

tandraschko:
first times used the review function and didnt submit it

src/main/resources/META-INF/resources/primefaces/core/core.dialog.js
beikov: Because there might be multiple nested frames from the top window to the source widget. Think of a frameset containing another frameset. If the source button is in frame1->frame1_1, we need to first resolve form the top window the frame with the id `frame1`, go into the content window of that frame, resolve `frame1_1` and use that content window as windowContext for the source widget.

src/main/resources/META-INF/resources/primefaces/core/core.dialog.js
beikov: I can try to add some more comments and examples if you like, but that's just how resolving nested frames works. Not sure how else this could be implemented. It's complex because the scenario this handles is complex. I'm open for concrete suggestions, but I just don't see a simpler version of this. I already reduced to to the bare minimum IMO.

tandraschko: yes please

beikov:
Allright, I added the documentation, rebased and pushed. Hope this is all good now @tandraschko ?

tandraschko:
Yep, thats a nice documentation! Thank you.

PR #5210, opened 2019-10-02
Fix #5209: Tree filter behave same as other components.

### LINKED ISSUE #5209: Tree: Filter not always triggered
## 1) Environment
- PrimeFaces version: PrimeFaces Showcase (7.1-SNAPSHOT)
- Does it work on the newest released PrimeFaces version? No (7.0)
- Affected browsers: Firefox, Chrome

## 2) Expected behavior

Filtering of the tree should be triggered by:
- Cutting (ctrl+x)
- Pasting (ctrl+v)
- Deleting (del key)

## 3) Actual behavior

If you enter a filter text by pasting, or remove text by cutting or deleting with the del key, filtering of the tree is not triggered.
Deleting with backspace works. In other components with filtering, e.g. DataTable or PickList, this problem does not exist.

melloware:
Also fixed the line endings in the PartialResponseWriter.


melloware:
@tandraschko @Rapster do you mind reviewing this one.  I tested it thoroughly.

After we merge this PR I think we can write some utility JS method because it looks like DataList, PickList, Tree etc that use filters all use the same copy and pasted code for ignoring the ENTER key and have a list of ignored keys for the filter etc.

melloware:
Sorry merging this to fix  the Line endings issue.

PR #5214, opened 2019-10-03
Fix #5213: CSP_POLICY allow custom policy override.

### LINKED ISSUE #5213: CSP: Customize Policy
We cannot know every Javascript usage on every PrimeFaces website. You may have custom code in your application that you need to allow other CSP directives such as `unsafe-inline` or whitelist a websitesuch as `https://www.google-analytics.com`.  By using the `primefaces.CSP_POLICY` context parameter you can override the default policy.

```xml
<context-param>
    <param-value>script-src 'self' 'unsafe-inline' 'unsafe-eval' https://www.google-analytics.com</param-value>
</context-param>
```

This will give developers ultimate flexibility and it seems to be what a lot of other frameworks are allowing so you can customize the policy.

melloware:


tandraschko:
+1
i would just mention our default with description in the docu

melloware:
Thanks, just added a section in the doc about "Default Policy".

PR #5220, opened 2019-10-04
Fix #5219: Filter refactor JS methods.

### LINKED ISSUE #5219: Filter: Refactor JS filtering
There are 8 places in the code doing the identical check to block an enter key.  

Refactor into `PrimeFaces.utils.blockEnterKey` method.

There are 3 places doing identical key filtering for filters.

Refactor into `PrimeFaces.utils.ignoreFilterKey` method.

melloware:


Rapster (approved):

tandraschko:
+1
Good idea

PR #5222, opened 2019-10-07
Fix #433: Allow ENTER/ESC to submit/cancel row editor.

### LINKED ISSUE #433: DataTable: Allow Enter and Esc to be used in editing to confirm/cancel editing
Having a `<p:datatable>` with a `<p:rowEditor>`, it would be quite natural to be able to use Enter to confirm editing and Esc to cancel it. Instead, these keys are currently ignored and captured by other elements in the page.

melloware:
1. Use `:input` instead of `input` so if you are are a select dropdown you can still press ENTER or ESC.  Previously it would only work if you were in an input box.

2. Modified the ROW finder to use the more specific styles `.ui-row-editor-check` and `.ui-row-editor-close` instead of `.ui-row-editor .ui-icon-close`

3. used Jquery constants for ENTER and ESC.


tandraschko:
+1
i would just give the "case"s more indentation

melloware:
I added an `off()` call to be consistent with others in the datatable and also fixed the indenting to be in line with the rest of the datatable.

Issue #5129, opened 2019-09-09
Calendar/DatePicker out of range validation message not customizable enough


tandraschko:
Currently, if the input exceeds the range (mindate and maxdate), we add the following message:

> primefaces.calendar.OUT_OF_RANGE = {0}: Validation Error: Value is out of range.

1) We don't even pass the min/maxdate as param. So per default the message is not usefull.
2) If a user overwrites the messge, a user can not differ between exeeds max or mindate. Still, we don't have the params
3) A user can not even define a good validation message via attribute as the "validationMessage" is not only used for range validation

tandraschko:
About 1) and 2)
We can simple pass params to the message, but even then, there might be only one attribute defined (min or max) and not both
So, how could a user use a global message, if some calendars only have mindate, others also have maxdate?

Is it a good solution to split this message into 2 messages?


@melloware @Rapster

tandraschko:
@christophs78

christophs78:
3 messages?
1. Date must be between X and Y
2. Date must be X or later
3. Date must be Y or before

Plan B (or C?) would be to prevent the user on the client side from typing/selecting a date that is not allowed. (Yes we have to check this one the server side too, but maybe the message is not that important any more.)

tandraschko:
3 sounds actually good
1) if min and max is defined
2) else-if min is defined
3) else-if max is defined

christophs78:
I´m working on this and will provide a PR.

PR #5231, opened 2019-10-08
#5129 Calendar/DatePicker out of range validation message not customizable enough

christophs78:
Implemented this for DatePicker.
Will look into Calendar as a second step.


tandraschko:
like it! seems even bit more readable

tandraschko:
Thanks for your work.
@melloware please also review and apply if it's ok from your side

melloware:
Reviewing it now.

melloware:
@christophs78 The ticket says Calendar/DatePicker validation but this PR looks like it is only for DatePicker? Am I seeing that correctly?

melloware:
Oh I see you comment you are still investigating Calendar.  I will wait until you have that in.

christophs78:
Now also covers Calendar. And did some refactoring to reduce duplicated code.

melloware:
This is a really nice refactoring.  Thanks for the PR.

PR #5232, opened 2019-10-08
Fix #3078: Swipe refactoring and exclude role=combobox.

### LINKED ISSUE #3078: Add swipe functionality to TabView

melloware:
@fanste Please test this fix. I just tested in IE11 Phone mode and proved it was an issue and its fixed.

- It also allowed me to refactor this to common swiped exlcuded since that was cut and pasted.
- I chose role='combobox' so that will fix SelectOneMenu, SelectCheckBox menu and any others that are rendered with a DIV.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/carousel/carousel.js
+++ b/src/main/resources/META-INF/resources/primefaces/carousel/carousel.js
@@ -159,7 +159,7 @@ PrimeFaces.widget.Carousel = PrimeFaces.widget.DeferredWidget.extend({
                     $this.setPage($this.page - 1);
                 }
             },
-            excludedElements: "button, input, select, textarea, a, .noSwipe"
+            excludedElements: PrimeFaces.utils.excludedSwipeElements()
         });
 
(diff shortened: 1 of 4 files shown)


fanste:
Seems to be working, thanks.

Issue #3549, opened 2018-04-03
DataTable: (inline-)edit - row-mode: set focus to the first input-field in the row


christophs78:
## 1) Environment
- PrimeFaces version: 6.2

## 2) Expected behavior
When the user clicks on the row-edit-icon the focus should be set to the first input-field in the row.

## 3) Actual behavior
Focus ist not set. User has to click into the input-field.

## 4) Steps to reproduce
https://www.primefaces.org/showcase/ui/data/datatable/edit.xhtml
First example. (Row Editing) 
Click on the row-edit-icon.

christophs78:
Currently we are able to do work-arrounds like
`<p:ajax event="rowEditInit" oncomplete="dataTableRowEditFocusFirstInput();" />`

```
<script>
function dataTableRowEditFocusFirstInput() {
   var input=jQuery('#formEdit\\:automarkeTable .ui-datatable-data tr.ui-row-editing td input');
   var firstInput=input.first();
   firstInput.focus();
}
</script>
```

This is not perfect because it has shortcommings when multiple row´s are set to edit-mode.

Think it should be realtiv easy to add this to DataTable-Primefaces-Component without breaking exiting implementations. (Use CSS-selector to select the first input-field within the row and set the focus to the first input-field.)

PR #5233, opened 2019-10-09
#3549 DataTable: (inline-)edit - row-mode: set focus to...

christophs78:
… the first input-field in the row

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
+++ b/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
@@ -2399,11 +2399,6 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
         else {
             this.showRowEditors(row);
 
-            let inputs=$(row).find(':input:enabled');
-            if (inputs.length > 0) {
-                inputs.first().focus();
-            }
-
             if(this.hasBehavior('rowEditInit')) {
                 var rowIndex = this.getRowMeta(row).index;
 
@@ -2423,6 +2418,11 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
             column.find('.ui-cell-editor-output').hide();
             column.find('.ui-cell-editor-input').show();
         });
+
+        var inputs=row.find(':input:enabled');
+        if (inputs.length > 0) {
+            inputs.first().focus();
+        }
     },
 
     getCellMeta: function(cell) {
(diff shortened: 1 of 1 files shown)


src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
```diff
@@ -2399,6 +2399,12 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
         else {
             this.showRowEditors(row);
 
+            let inputs=$(row).find('input, select, textarea, button');
```
melloware: I think a better selector would be `:input:enabled` which finds the first input,select,textarea radio button or any input element that is actually NOT disabled.

src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
```diff
@@ -2399,6 +2399,12 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
...
+            if (inputs.length > 0) {
+                let firstInput = inputs.first();
+                firstInput.focus();
```
melloware: I think you should just make this one line with `inputs.first().focus();` no need to declare variable for this yet.

melloware (changes requested):

tandraschko:
let is supported in IE?

melloware:
It looks like Let is supported in IE11 https://stackoverflow.com/a/2356926/502366

christophs78:
> 
> 
> It looks like Let is supported in IE11 https://stackoverflow.com/a/2356926/502366

Yep. :-) https://caniuse.com/#feat=let

christophs78:
@melloware : Thanks for your feedback. Done.

mertsincan:
Hi @christophs78,

Thanks a lot for the PR! But, I think you need to make these changes in showRowEditors. Because, row editor has lazy feature. 
```
      if(this.cfg.rowEditMode === "lazy") {
            this.lazyRowEditInit(row);
        }
```
When users use this feature, your code won't work for them.

Also, could you please use 'var' instead of 'let'? We don't use ES6 syntax in PrimeFaces.

Maybe, 
```
showRowEditors: function(row) {
        var isFirstItemFocused = false;
        row.addClass('ui-state-highlight ui-row-editing').children('td.ui-editable-column').each(function() {
            var column = $(this);
    /** Maybe **/
            if (!isFirstItemFocused) {
               var inputs = column.find('...');
               if (inputs.length > 0) {
                    ...
                    isFirstItemFocused = true;
                }
            }
    /**********/
            column.find('.ui-cell-editor-output').hide();
            column.find('.ui-cell-editor-input').show();
        });
    },
```

Best Regards,

christophs78:
@mertsincan: Thanks for your valuable feedback.

melloware (approved):

melloware:
Good teamwork on this PR. :)

tandraschko:
is the feedback from mert about lazy already implemented?

christophs78:
Yes, it is. (https://github.com/primefaces/primefaces/pull/5233/commits/b03082954ede0e4552554e34a6bf3a16f63633cc)

PR #5238, opened 2019-10-11
Fix #4000: ContextMenu add mobile long press event to trigger.

### LINKED ISSUE #4000: ContextMenu: support touch events
## 1) Environment
- PrimeFaces version: **6.2.8**
- Does it work on the newest released PrimeFaces version? **No** Version? **6.2.8**
- Does it work on the newest sources in GitHub? (Build by source ->**not tested because no commit related to that**
- Application server + version: **Payara 5.182**
- Affected browsers: **all**

## 2) Expected behavior

**When pressing long time on the screen the contectmenu should popup**

## 3) Actual behavior

**When pressing long time on the screen default context of ios shows up.**

## 4) Steps to reproduce

**Showcase Datatable ContextMenu example with ios and press long time inside the table** 

## 5) Sample XHTML

**not needed**

## 6) Sample bean

**not needed**

melloware:


tandraschko:
how do you bind the swipe exactly to this single element? looks global

melloware:
You are right, in the bottom second I changed to `$(this.jqTargetId).swipe` to bind to the target element only.

In the top section it is global on the document the same as the right click.

tandraschko:
ok. tested in all browsers i assume? if yes -> please apply it

PR #5239, opened 2019-10-11
Fix #608: Datatable add swipe support if paginated.

### LINKED ISSUE #608: DataTable: add swipe support
It would be nice to have the same swipe functionalities that are available in the mobile-renderkit also in the standard one so that responsive pages could be swipable.

Fabio

melloware:


tandraschko:
did you test it in a android and ios?

melloware:
I tested in Chrome mobile, and Firefox mobile but not an actual IOS or Android device.

mohsen:
I have a custom build of primefaces, downloaded from git repos on 20191112. Any datatable with pagination throws js exception on chrome/firefox mobile emulator. When mobile emulator is off, no error is thrown. I believe this is related to swipe feature which is only enabled on mobile devices.

```
jquery.js.xhtml?ln=primefaces&v=8.0-20191112:2 Uncaught TypeError: owner.swipe is not a function
    at Class.bindSwipeEvents (components.js.xhtml?ln=primefaces&v=8.0-20191112:17084)
    at Class.bindPaginator (components.js.xhtml?ln=primefaces&v=8.0-20191112:3547)
    at Class.init (components.js.xhtml?ln=primefaces&v=8.0-20191112:3397)
    at Class.prototype.<computed> [as init] (core.js.xhtml?ln=primefaces&v=8.0-20191112:2562)
    at new Class (core.js.xhtml?ln=primefaces&v=8.0-20191112:2575)
    at Object.createWidget (core.js.xhtml?ln=primefaces&v=8.0-20191112:574)
    at Object.cw (core.js.xhtml?ln=primefaces&v=8.0-20191112:551)
    at HTMLDocument.<anonymous> (home.xhtml?nocache:324)
    at e (jquery.js.xhtml?ln=primefaces&v=8.0-20191112:2)
    at t (jquery.js.xhtml?ln=primefaces&v=8.0-20191112:2)
```

tandraschko:
@melloware can you check that?

melloware:
Investigating!

melloware:
@moshen I can't reproduce.  Can you create a reproducible example or use the PrimeFaces 8.0-SNAPSHOT built in Jitpack instead of your custom built one?

melloware:
@moshen in fact go to the Showcase here: https://www.primefaces.org/showcase/ui/data/datatable/paginator.xhtml

Open Chrome or Firefox mobile emulator and you can see the swipe left and right work fine.  Can you verify?

mohsen:
@melloware Thanks for quick investigation.
I have no problem with sample showcase and its swipe feature. It works fine. But I believe the showcase is simpler than my case. My datatable has filtering enabled, so the following code is generated by PF. I cannot see such a method `PrimeFaces.cw("DataTable", ...)` in the showcase generated html.

```
<script id="form:inboxDataTable_s" type="text/javascript">
    $(function() {
        PrimeFaces.cw("DataTable", "widget_form_inboxDataTable", {
            id: "form:inboxDataTable",
            paginator: {
                id: ['form:inboxDataTable_paginator_top', 'form:inboxDataTable_paginator_bottom'],
                rows: 20,
                rowCount: 12000,
                page: 0,
                currentPageTemplate: '{startRecord} - {endRecord} / {totalRecords}',
                pageLinks: 7
            },
            filter: true,
            filterEvent: "enter",
            filterDelay: 500,
            groupColumnIndexes: [],
            disableContextMenuIfEmpty: false,
            behaviors: {
                sort: function(ext, event) {
                    PrimeFaces.ab({
                        s: "form:inboxDataTable",
                        e: "sort",
                        f: "form",
                        p: "form:inboxDataTable"
                    }, ext);
                }
            }
        });
    });
</script>

```

melloware:
So can you try your local sample and pull out filtering and see if that is the difference?  Anything else you can narrow down that might be different from the showcase that works?

mohsen:
Yes, I'm trying to remove anything I can. but `PrimeFaces.cw("DataTable",...)` is still there... Investigating more...

melloware:
cw means CreateWidget. That is the script that creates the Datatable JS.

melloware:
@moshen I found the problem and fixed with this PR: 
https://github.com/primefaces/primefaces/pull/5344

Thanks for reporting!

mohsen:
@melloware thanks for the quick help.

Issue #4128, opened 2018-10-08
Datatable: Row selection causes rendering to go in an infinite loop, when updating header from onRowSelect ajax event, and when rows attribute is not set


ghost:
## 1) Environment
- PrimeFaces version: 6.2
- Does it work on the newest released PrimeFaces version? Version? No
- Does it work on the newest sources in GitHub? No
- Application server + version: jetty from primefaces-test
- Affected browsers: all

## 2) Expected behavior

 When row is selected, expect the row to be selected, and expect the application to continue running smoothly. 

## 3) Actual behavior

When row is selected, no response is received back from the ajax event, and the page is unresponsive.

On debugging, found that it is stuck in an infinite loop in UIData.visitRows
An observation is that this is because the rows attribute is missing from the datatable. However, it was not expected that this be stuck in an infinite loop. Also, everything else works without the rows attribute, for lazy datatable, as the scrollRows is set instead of rows.

## 4) Steps to reproduce

To reproduce use this test repo - https://github.com/amathew121/primefaces-test 

http://localhost:8080/primefaces-test/

On the datatable, click on any row to make a selection. 
At this point it will be stuck in an infinite loop, and other ajax actions (like scrolling) on the page will not work. You can also check the jvm taking up all the CPU.

## 5) Sample XHTML

```
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:ui="http://java.sun.com/jsf/facelets" xmlns:f="http://java.sun.com/jsf/core"
	xmlns:p="http://primefaces.org/ui" xmlns:h="http://java.sun.com/jsf/html">

<h:head>
	<title>PrimeFaces Test</title>
</h:head>
<h:body>

	<h1>#{testView.testString}</h1>

	<h1>#{testViewCdi.testString}</h1>

	<h:form id="tableForm">
		<p:dataTable var="car" value="#{dtLazyView.lazyModel}" scrollRows="20" scrollable="true" virtualScroll="true" scrollHeight="200" lazy="true"
			style="margin-bottom:0" selection="#{dtLazyView.selectedCars}" rowIndexVar="rowIndex">
			<f:facet name="header">
				<h:panelGroup id="usersSelection">
					<h:outputText value="#{dtLazyView.selectedCars.size()} cars selected" />
				</h:panelGroup>
			</f:facet>
[...]

PR #5240, opened 2019-10-12
Issue #4128

mrswadge:
https://github.com/primefaces/primefaces/issues/4128#issuecomment-541213394

### DIFF
--- a/src/main/java/org/primefaces/component/api/UIData.java
+++ b/src/main/java/org/primefaces/component/api/UIData.java
@@ -1029,7 +1029,7 @@ public class UIData extends javax.faces.component.UIData {
 
         while (true) {
             if (visitRows) {
-                if ((rows > 0) && (++processed > rows)) {
+                if (rows <= 0 || ((rows > 0) && (++processed > rows))) {
                     break;
                 }
 


melloware (approved):

tandraschko:
this breaks processing of a normal DataTable, i need to revert it.

I just have a p:selectBooleanCheckbox inside a row and the listener is not invoked anymore:


`            <p:dataTable var="r" value="#{bean.myListWithEntities}" id="table">
                <p:column>
                    <p:selectBooleanCheckbox id="chk"
                                             value="#{r.enabled}"
                                             valueChangeListener="#{....}">
                        <p:ajax update="@composite:@composite"/>
                    </p:selectBooleanCheckbox>
                </p:column>`

mrswadge:
Ok, understood.

We will have to revisit the solution as the infinite loop causes an OutOfMemoryException which takes down the server, so it's a severe bug when triggered. 

It creates an infinite number of org.primefaces.component.api.SavedState objects as it iterates through UIData.restoreDescendantState(UIComponent component, FacesContext context).

This stack trace is running on PrimeFaces 7.0.

```
"[ACTIVE] ExecuteThread: '6' for queue: 'weblogic.kernel.Default (self-tuning)'" #62 daemon prio=5 os_prio=0 tid=0x000000002a2f4000 nid=0xe2c runnable [0x000000003747a000]
   java.lang.Thread.State: RUNNABLE
	at javax.faces.component.UIComponentBase.getRenderer(UIComponentBase.java:1496)
	at javax.faces.component.UIComponentBase.getClientId(UIComponentBase.java:378)
	at org.primefaces.component.api.UIData.restoreDescendantState(UIData.java:746)
	at org.primefaces.component.api.UIData.restoreDescendantState(UIData.java:773)
	at org.primefaces.component.api.UIData.restoreDescendantState(UIData.java:728)
	at org.primefaces.component.api.UIData.setRowIndexWithoutRowStatePreserved(UIData.java:581)
	at org.primefaces.component.api.UIData.setRowIndex(UIData.java:640)
	at org.primefaces.component.api.UIData.visitRows(UIData.java:1010)
	at org.primefaces.component.api.UIData.visitTree(UIData.java:889)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1747)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1747)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1747)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1747)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1747)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1747)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1747)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1747)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1747)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1747)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1747)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1747)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1747)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1747)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1747)
	at javax.faces.component.UIForm.visitTree(UIForm.java:395)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1747)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1747)
	at com.io_software.catools.jspacc.bb.UI.findComponentById(UI.java:358)
        ...
```

mrswadge:
Before committing any code, perhaps worth a brief discussion. 

I trapped the issue in debug, I was wrong to think it was the `processed` counter, it's the `rowIndex` which is counting up infinitely. I was thinking I could simply alter my change to look at whether `rowIndex` is greater than `rows`. 

Do you think this is ok?

image

tandraschko:
i think this breaks other things as well as we have to visit all rows probably, not only the "rows per page"

mrswadge:
The LazyDataModel has a rowIndex set to 14, which is the size of the data array. This is despite the UIData.visitRows(..) setting the value well above that (debug trapped at over 10,000). When it calls through UIData.isRowAvailable() [next line in screenshot above] it's saying that the row *is* available. This check appears to be ignoring the rowIndex value that was set in UIData.visitRows(..).

Row count is zero which is interesting too?

image

mrswadge:
In UIData.getRows() it appears to get the value from a property stored on the component. For some reason this is coming back as zero. Under the LazyModel it's using data.size() instead.

```
UIData:
    public int getRows() {
        return (Integer) getStateHelper().eval(PropertyKeys.rows, 0);
    }
```
vs.
```
LazyDataModel:
    @Override
    public boolean isRowAvailable() {
        if (data == null) {
            return false;
        }

        return rowIndex >= 0 && rowIndex < data.size();
    }
```

Is there a reason for UIData using a property value instead of using the data model to get the row count?

Observations:

 - When UIData.rowCount is returned as 0 and using a LazyDataModel
   - UIData.processed does not increment due to left-hand-side of `if ((rows > 0) && (++processed > rows)) {`
   - UIData.isRowsAvailable() does not return false as LazyDataModel.rowIndex does not increase beyond LazyDataModel.data.size()

mrswadge:
We think we discovered the cause, but it was in our own code.

We have _LazyDataModel_ with a subclass implementation _GenericLazyModel_. Both have member _data_ as an ArrayList. The _GenericLazyModel.data_ list was empty, but the _LazyDataModel.data_ list was not because it had previously been populated. 

The _LazyDataModel.isRowAvailable()_ was looking at _LazyDataModel.data_, which was populated despite the _GenericLazyModel_ having an empty list. We were able to fix the issue with a single line of code that ensures that the data lists on both data models were in sync with each other.

I've not spent the time to prove it, but presumably the _UIData.getRows()_ row count was being calculated via the _GenericLazyModel.data_ or _LazyDataModel.rowCount_ value - either way it was being set to 0.

The two methods of calculating the number of rows caused the infinite loop. I don't know if this is something that Primefaces could protect against better. As I don't know the Primefaces code well, I probably will leave it here as we have identified a workaround.

melloware:
@mrswadge Thanks for posting back.  

@tandraschko Do you think this is likely the cause of the original ticket?  Is this jut misuse of the LazyDataModel?

mrswadge:
There are definitely some similarities in the circumstances between the OP and our setup as we both had checkboxes in data tables. The OP also mentioned the visit contexts, so it appears to land in the same ball park. There is no stack trace, so it's hard to be 100% sure.

I think our case is a misuse of the LazyDataModel as we duplicated the data list into our own implementation. I can't see a good reason for this as we could have used LazyDataModel.getWrappedData() to populate LazyDataModel.data in the subclass implementation directly.

mrswadge:
The code in the OP is quite close to our LazyDataModel code.

```
    @Override
    public List<Car> load(int first, int pageSize, String sortField, SortOrder sortOrder, Map<String,Object> filters) {
        ...
        //rowCount
        int dataSize = data.size();
        this.setRowCount(dataSize);
        ...
```

Our fix was simply to add another line to ours to call _this.setWrappedData( data );_

```
    @Override
	public List<Object> load(int first, int pageSize, List<SortMeta> multiSortMeta, Map<String,Object> filters) {
        ...
        //rowCount
        int dataSize = data.size();
        this.setRowCount(dataSize);
        this.setWrappedData( data ); // ensure LazyDataModel.data and GenericLazyModel.data are in sync
        ...
```

We're going to remove GenericLazyModel.data and use the one on the LazyDataModel instead in future.

mrswadge:
The question that remains for me is whether when LazyDataModel.load(int first, int pageSize, List<SortMeta> multiSortMeta, Map<String,Object> filters) is called whether it should set the returned list to LazyDataModel.data. I've not looked at the callers, so up to you guys.

PR #5245, opened 2019-10-14
Fix #5038: DatePicker time picker only respect left click.

### LINKED ISSUE #5038: DatePicker: timepicker stuck at incrementing on right click or releasing outside
## 1) Environment
- PrimeFaces version: 7.0.5
- Application server + version: WildFly 17
- Affected browsers: Tested with Chrome and Safari

## 2) Expected behavior
When right clicking a timepicker, it would not start incrementing at all.
When releasing the mousebutton outside of the timepicker it would stop incrementing.

## 3) Actual behavior
When right clicking a timepicker, it starts incrementing just like the left click. But it will never release because the context menu popping up prevents the mouseup event.

When left click + hold, then drag your mouse outside of the box it keeps incrementing, even after releasing the mousebutton.

## 4) Steps to reproduce
See actual behavior.

## 5) Sample XHTML
```
<p:datePicker showTime="true"/>
```
## 6) Sample bean
No bean required

melloware:
Tested in IE11, Edge, Chrome, Firefox.

@mertsincan this is a fix in core datepicker code which probably needs to go to PrimeReact, Vue, etc.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
+++ b/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
@@ -1658,6 +1658,10 @@
         },
 
         repeat: function (event, interval, type, direction) {
+            // GitHub #5038 only allow left click
+            if (event.which !== 1) {
+                return;
+            }
             var i = interval || 500,
                 $this = this;
 


mertsincan:
Thanks a lot, @melloware ;)

PR #5249, opened 2019-10-17
Fix #5229: Junit 5

### LINKED ISSUE #5229: Update JUnit from 4.x to 5.x

melloware:
Added some generics to BaseCalendarRenderer like Class<?> instead of raw Class.


tandraschko:
+1

PR #5253, opened 2019-10-18
Fix #3624: Escape jQuery selectors.

### LINKED ISSUE #3624: Escaping of jquery selectors
In https://github.com/primefaces/primefaces/issues/3612 @xtron found one specific issue regarding improperly escaped strings used within jquery selectors.

We've found that there are various similar locations where escaping is also missing, see https://github.com/primefaces/primefaces/issues/3612#issuecomment-384518318

Just do a regex search  `\[[A-Za-z0-9-_]+=["'].*?\+` (naive approach - may return false positives) over all primefaces *.js files to reveal potential erroneous locations.

Please keep in mind that the regex mentioned above only searches for attribute selectors. Other patterns could be of interest as well.

Generally, jquery's `$.escapeSelector` should be used to avoid such issues. See https://github.com/primefaces/primefaces/pull/3611/files for an example.

melloware:


tandraschko:
i wonder if it isn't used already internally by jquery? So do we really have to use it?

melloware:
I don't think it is since the documentation would list it as deprecated if it was being done automatically?
https://api.jquery.com/jQuery.escapeSelector/

tandraschko:
i dont know if we should apply it everything.
take e.g. this case: $.escapeSelector(this.id)
what unallowed character could be in the id? just ':' - but it still works fine?!

melloware:
I was wondering about that.  I can remove it from ID's.  But anywhere a NAME is being passed feels like it should be done.

tandraschko:
yep
i would also skip esacping the other IDs

tandraschko:
+1 now

PR #5127, opened 2019-09-08
#4646: Timeline: move to java.time.LocalDateTime

christophs78:
Some kind of "preview". I´ll have to rebase/update/redo this PR after https://github.com/primefaces/primefaces/pull/5113 is merged.

Showcase-PR: https://github.com/primefaces/showcase-facelift/pull/59

Noteable:
- Changed Client-JS to send toISOString() instead of getTime() to Server
- so browserTimeZone-attribute should not be necessary any more
- maybe we should merge DateUtils with CalendarUtils (DateUtils is only used by Timeline; there is left only one method)

PR #5113, opened 2019-09-04
Fix #3245,#2193,#2941,#5143  - Timeline: Migrate to vis.js

jasonex7:
Fixes #3245 

- Using vis.js 5.1. It was recommended to upgrade to v6 which it should be out in the next few days. It was told that it will be 95% compatible to current 5.1 release so I have only to replace 0-timeline.min.js. (https://github.com/visjs/vis-timeline)
- TimelineEvent, TimelineGroup, TimelineModel, TimelineModificationEvent, TimelineSelectEvent now use generics.
- TimelineEvent has now 3 more properties to fine tunning editable: editableTime, editableGroup and editableRemove.
- TimelineGroup has now styleClass property similar to the one in TimelineEvent.
- Properties deprecated in favor of others: axisOnTop, groupsChangeable, groupsWidth, groupMinHeight, snapEvents, timeChangeable.
- Properties deprecated with NO replacement: dragAreaWith (use css classes instead), unselectable, groupsOnRight, animate, animateZoom, showButtonNew, showNavigation.
- New properties: maxHeight, orientationAxis, orientationItem, editableAdd, editableTime, editableGroup, editableRemove, editableOverrideItems, eventHorizontalMargin, eventVerticalMargin, groupStyle, snap, clickToUse.
- New extender property to specify a javascript function to extend configuration of vis.js timeline component similar to extender property in <p:chart> component.

PR #5257, opened 2019-10-19
#4646: Timeline: move to java.time.LocalDateTime ("Version 2")

christophs78:
redone after https://github.com/primefaces/primefaces/pull/5113, we may close the old PR https://github.com/primefaces/primefaces/pull/5127

Showcase-PR: https://github.com/primefaces/showcase-facelift/pull/59

Noteable:
- Changed Client-JS to send toISOString() instead of getTime() to Server
- so browserTimeZone-attribute should not be necessary any more


melloware:
@jasonex7 Please review this PR if you can.

tandraschko:
why page and sort event has been removed?

christophs78:
This two events are not implemented. (see https://github.com/primefaces/primefaces/blob/master/src/main/java/org/primefaces/component/timeline/Timeline.java)
IMO this is a documentation error.

tandraschko:
looks good to me
but would be great if @jasonex7 can also review and test it.

melloware (approved):

src/main/resources/META-INF/primefaces-p.taglib.xml
```diff
@@ -26310,15 +26310,6 @@
...
         </attribute>
-        <attribute>
-            <description>
-                <![CDATA[Time zone the user's browser / PC is running in. This time zone allows to correct the conversion of start / end dates to the target timeZone for displaying.
-                The attribute can be either a String or TimeZone object or null. Note: browserTimeZone should be provided if the target timeZone is provided. If null, browserTimeZone defaults to the server's timeZone.]]>
-            </description>
-            <name>browserTimeZone</name>
-            <required>false</required>
-            <type>java.lang.Object</type>
-        </attribute>
```
jasonex7, line 26321: I recommend not delete this property from this file, but deprecate in comments instead as I did in other properties like `groupsChangeable` in order to don't break existing xhtml.

src/main/java/org/primefaces/util/CalendarUtils.java
```diff
@@ -477,4 +413,19 @@ public static final String removeTime(String pattern) {
         return pattern.trim();
     }
 
+    /**
+     * Convert ISO-String (@see <a href="https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString">https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString</a>)
+     * to LocalDateTime.
+     * @param isoDateString
```
jasonex7, line 420: Missed @param zoneId

docs/7_1/components/timeline.md
```diff
@@ -93,12 +92,8 @@ public class BasicTimelineView implements Serializable {
 
     @PostConstruct
     protected void initialize() {
-        model = new TimelineModel();
```
jasonex7, line 96: The `model` variable has to be initialized.

jasonex7 (changes requested):
This PR seems good to me and I've tested it with my existing code that is already in production using vis-timeline and everything is OK. I'm just suggesting some cosmetic changes.

PD: Sorry for the late response, I was on vacations.

christophs78:
@jasonex7 : Thanks for reviewing!

docs/7_1/components/timeline.md
christophs78: fixed

src/main/java/org/primefaces/util/CalendarUtils.java
christophs78: fixed

src/main/resources/META-INF/primefaces-p.taglib.xml
christophs78: I removed this, because it´s not necessary anymore.
LocalDateTime-values are sent from the server to the client as `ISO_OFFSET_DATE_TIME`.
eg `new Date('2019-10-29T12:00:00+01:00'), end: new Date('2019-10-29T17:00:00+01:00')`
They are sent from the client to the server as isoString (`item.start.toISOString()`)
eg `2019-10-29T20:30:00.000Z`
IMO browserTimeZone was some kind of work-around because until now java.util.Date-values were transfered without any timezone-info.
eg `"start":new Date(1572387021292),"end":new Date(1572415821292)`
IMO the new solution is better than the old one. The browser know´s his timezone. He does not need some kind of awkward help from the server-side.

jasonex7: I understand your solution, is seems a good solution for me and I'm OK with that. My recommendation was only to don't remove the property `browserTimeZone` from taglib file in order to avoid breaking existing xhtml that already uses that property and document the deprecation in the taglib instead of removing the property. What do you think @melloware @tandraschko ?

melloware: I think its OK to mark it in the documentation as deprecated but can probably be removed from the taglib if it truly is not doing any good.  Anybody that breaks will be forced to lookup the documentation.

melloware: Sorry I meant Migration Guide!

melloware (approved):

melloware:
I updated the Migration Guide mentioning that browserTimeZone is deprecated with NO replacement.

PR #5269, opened 2019-10-25
Fix #5268: Security EL functions allow arrays and collections.

### LINKED ISSUE #5268: EL functions: Add array/list support for page authorization related functions
When checking page authorization with p:ifAllGranted(), p:ifAnyGranted() or p:ifNotGranted() you have to build a String containing all roles. When working with constant imports by e.g. using p:importConstants you have to concat one role with another by e.g. #{Role.A += ',' += Role.B += ...}.

It would be very nice if one could just pass an array or list of roles by e.g. #{[Role.A, Role.B, ...]}.

Even if one prepares the valid roles in backing bean he would benefit. Just array or list initialization instead of error-prone string concatination.

melloware:


tandraschko:
basically +1
but please test older containers if its working fine

melloware:
@Rapster is still testing i won't merge until i hear from him if he finds a better solution.

tandraschko:
i would apply it
i also had some problems using overloading on facelet functions for resolveComponent/WidgetVar
The Object param is probably the safer way.

Rapster:
What kind of problem?

tandraschko:
it didn't call the method correclty / throwed a method, can't remember exactly
therefore i introduced a second method with "fromContext" suffix

Issue #5267, opened 2019-10-24
Schedule: Upgrade Fullcalendar (4.3.1) 


christophs78:
succeeds https://github.com/primefaces/primefaces/issues/4783
Migration-Guide: https://fullcalendar.io/docs/upgrading-from-v3

christophs78:
@melloware : Started to work on this. At this point in time i switched to the new version and got the rendering up and running. I´ll have to put more work on this. (events etc) But i hope to have a complete PR within some weeks. (spare time is limited...)

melloware:
fantastic!  Much appreciated.

ProfEibe:
Will this include the List-View of the calendar-events?
Or is this already in the 3.10 version?

christophs78:
Yes, we will include https://fullcalendar.io/docs/list-view with 4.x.

Issue #4876, opened 2019-06-10
Add columnHeaderFormat attribute to Schedule


mertsincan:
Format for column headers. the columnFormat is deprecated; https://fullcalendar.io/docs/v3/columnFormat

PR #5274, opened 2019-10-26
#5267: Schedule: update to new version of Fullcalendar (4.3.1) 

christophs78:
Showcase-PR: https://github.com/primefaces/showcase-facelift/pull/79

TODO´s:
- [x] check each feature available via PF-schedule
- [x] maybe expose some more (new) features
- [x] cleanup code
- [x] update PF - migration guide
- [x]  https://github.com/primefaces/primefaces/issues/4876, https://fullcalendar.io/docs/v3/columnHeaderFormat, https://fullcalendar.io/docs/v3/view-specific-options

CHANGED: 
- [x] localization comes with FullCalendar; does not require/use PF-localization
- [x] new view-names (dayGridMonth, dayGridWeek, dayGridDay, timeGridWeek, timeGridDay, listYear, listMonth, listWeek, listDay); old ones are translated inside the Schedule-component
- [x] ScheduleEntryMoveEvent; new properties yearDelta and monthDelta
- [x] ScheduleEntryResizeEvent; new properties yearDelta and monthDelta; splitted properties between start and end (up to PF 7.0 only end)
- [x] property slotLabelFormat was removed (only available in commerical version of FullCalendar - see https://fullcalendar.io/docs/slotLabelFormat)
- [x] ScheduleEvent; new properties groupId and overlapAllowed


melloware:
Looking good. Don't forget to update this Doc MD as well as part of the PR: https://primefaces.github.io/primefaces/7_1/#/core/thirdpartylibs

pom.xml
```diff
@@ -783,7 +783,24 @@
...
-                                <include>${project.build.directory}/classes/META-INF/resources/primefaces/schedule/0-fullcalendar.min.js</include>
+                                <include>${project.build.directory}/classes/META-INF/resources/primefaces/schedule/fullcalendar-core.js</include>
+                                <include>${project.build.directory}/classes/META-INF/resources/primefaces/schedule/fullcalendar-locales-all.js</include>
+                                <include>${project.build.directory}/classes/META-INF/resources/primefaces/schedule/fullcalendar-interaction.js</include>
+                                <include>${project.build.directory}/classes/META-INF/resources/primefaces/schedule/fullcalendar-daygrid.js</include>
+                                <include>${project.build.directory}/classes/META-INF/resources/primefaces/schedule/fullcalendar-timegrid.js</include>
+                                <include>${project.build.directory}/classes/META-INF/resources/primefaces/schedule/fullcalendar-list.js</include>
+                                <include>${project.build.directory}/classes/META-INF/resources/primefaces/schedule/fullcalendar-moment.js</include>
```
melloware: Are you sure you need moment.js as moment.js already exists in https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/moment/moment.js .  You could probably just include this in your component since moment is used by multiple components.

And I would move moment-timezone.js to the /moment folder of PF.

pom.xml
```diff
@@ -968,7 +985,22 @@
...
+                                <exclude>**/fullcalendar-locales-all.min.js</exclude>
+                                <exclude>**/fullcalendar-interaction.js</exclude>
+                                <exclude>**/fullcalendar-interaction.min.js</exclude>
+                                <exclude>**/fullcalendar-daygrid.js</exclude>
+                                <exclude>**/fullcalendar-daygrid.min.js</exclude>
+                                <exclude>**/fullcalendar-timegrid.js</exclude>
+                                <exclude>**/fullcalendar-timegrid.min.js</exclude>
+                                <exclude>**/fullcalendar-list.js</exclude>
+                                <exclude>**/fullcalendar-list.min.js</exclude>
+                                <exclude>**/fullcalendar-moment.js</exclude>
```
melloware: Do you need the regular and the .min files?

src/main/java/org/primefaces/component/schedule/Schedule.java
```diff
@@ -48,11 +47,13 @@
 import org.primefaces.util.MapBuilder;
 
 @ResourceDependencies({
-        @ResourceDependency(library = "primefaces", name = "schedule/schedule.css"),
+        @ResourceDependency(library = "primefaces", name = "schedule/fullcalendar-core.css"),
+        @ResourceDependency(library = "primefaces", name = "schedule/fullcalendar-timegrid.css"),
+        @ResourceDependency(library = "primefaces", name = "schedule/fullcalendar-daygrid.css"),
+        @ResourceDependency(library = "primefaces", name = "schedule/fullcalendar-list.css"),
         @ResourceDependency(library = "primefaces", name = "components.css"),
         @ResourceDependency(library = "primefaces", name = "moment/moment.js"),
```
melloware, line 52: AHH yes see here you are including moment properly.

melloware (changes requested):

pom.xml
christophs78: FullCalendar 4.x is splitted into modules. fullcalendar-moment.js is the plugin which connects FullCalendar to momentJs. It´s only "glue".

pom.xml
christophs78: I´ll clean this up before we merge this PR.

christophs78: Done

src/main/java/org/primefaces/component/schedule/Schedule.java
Rapster: Why not aggregating these css files all together into `schedule.css`?

christophs78: > 
> 
> Why not aggregating these css files all together into `schedule.css`?

Good idea. Done.

Rapster: I think we should stick to the name `schedule.css` instead of `fullcalendar.css`, I know that's the js widget name but PF named it `schedule.css`

christophs78: done

src/main/java/org/primefaces/component/panelgrid/PanelGridRenderer.java
```diff
@@ -305,8 +305,10 @@ public void encodeGridBody(FacesContext context, PanelGrid grid, int columns, St
             }
         }
 
-        if (i != 0 && (i % columns) != 0) {
-            writer.endElement("div");
+        if (LAYOUT_GRID.equalsIgnoreCase(layout)) {
+            if (i != 0 && (i % columns) != 0) {
+                writer.endElement("div");
+            }
```
melloware: I think you can remove this fix as Mert just refactored the panelGrid for Flex option in https://github.com/primefaces/primefaces/issues/5275

melloware (changes requested):

christophs78:
Made some mistake with rebase. I´am looking into this.

melloware (approved):

melloware:
Since I merged your other changes not it looks like you need to rebase a little bit.

christophs78:
Yes, i´ll need to do some rebase. I also have to look at columnHeaderFormat. This requires some work. I´ll update you after this is done. (Maybe later today or tomorrow.)

christophs78:
IMO this PR is finished. We only need to update the Migration Guide afterwards.

melloware:
@tandraschko I have been following this PR and it looks good to me.  let me know if you want me to merge then we can merge showcase PR and Migration Guide.

melloware:
@tandraschko did you want to merge this or did you want to wait until 7.2?

tandraschko:
Please update the migration guide, thanks :D

melloware:
Mirgation guide updated!

PR #5278, opened 2019-10-28
Fix #5277: Browser plugin 0.1.0

### LINKED ISSUE #5277: jQuery: Browser Plugin Update 0.1.0
Currently using Browser plugin 0.0.6 upgrade to latest 0.1.0.

melloware:


tandraschko:
please also chech core.env.js
we forked the browser plugin there, can't remember why

We should check if
1) every component has a jquery-plugins include
or
2) move jquery-browser to core.js instead jquery-plugins

melloware:
@tandraschko  you are right this was when Jquery pulled the browser out of Jquery core PF copied and pasted the old code here.  Then we realized too many places in the code were using $.brower including plugins so we added the browser plugin back in.

So I just updated the PR to:
1. Make browser part of core.js not jquery-plugins.js
2. Removed the old copy and pasted code from PF.core.env.

Tested in IE and Chrome and everything is working fine.

tandraschko:
+1

PR #5292, opened 2019-10-30
Fix #5291: FileUpload 10.2.0

### LINKED ISSUE #5291: FileUpload: Update to 10.2.0
Current Version 9.4.5
New Version: 10.2.0

Also replacing the 300KB gif ie-anim with 3KB progressbar.gif in from FileUploader.

melloware:
Just getting us on the latest version.  tested in IE11 Chrome and Firefox.

Also replaced 300KB image with 3KB image which is identical I have no idea why the old one was so large for an animated gif.


tandraschko:
+1 if well tested

melloware:
Tested in all browers and all showcase examples and everything is looking good.

PR #5322, opened 2019-11-12
Fix #5321

### LINKED ISSUE #5321: Excel(X)Exporter: Improve performance for huge exports
## 1) Environment
- PrimeFaces version: 7.1-SNAPSHOT from the past

## 2) Expected behavior
A datatable with thousands of rows and tens of columns should be exported in a reasonable time.

## 3) Actual behavior
The export takes a really long time and even may crash the application server because of the feature `autoSizeColumn` that is always invoked at the end of the excel export.

A PR is coming that allows to disable this feature (`autoSizeColumn`) using exporter options.

fanste:


melloware (approved):

melloware:
@tandraschko I think this is a good change as @fanste is correct autosizing can be expensive if you have a very large amount of data.

PR #5324, opened 2019-11-12
Fix #5323: ContextMenu not attaching to document

### LINKED ISSUE #5323: ContextMenu: Not attaching to page
PrimeFaces 8.0-SNAPSHOT

Visit Showcase: https://www.primefaces.org/showcase/ui/menu/contextmenu/basic.xhtml

The Context menu should be attached to the page but its not working at all.

melloware:
No idea why this fixes it but it does.


melloware:
Oh is `var document` a reserved word in Javascript so I bet I can't assign a variable to it. DUH!

PR #5325, opened 2019-11-12
Fix #5213: CSP when using unsafe directives a NONCE is not allowed

### LINKED ISSUE #5213: CSP: Customize Policy

melloware:
Based on the spec when using unsafe-inline or unsafe-eval using a custom policy a NONCE must not be included.

### DIFF
--- a/src/main/java/org/primefaces/csp/CspPhaseListener.java
+++ b/src/main/java/org/primefaces/csp/CspPhaseListener.java
@@ -40,6 +40,7 @@ import org.primefaces.util.Lazy;
 public class CspPhaseListener implements PhaseListener {
 
     private static final long serialVersionUID = 1L;
+    private static final String UNSAFE_DIRECTIVES = "unsafe-";
 
     private Lazy<Boolean> enabled;
     private Lazy<String> customPolicy;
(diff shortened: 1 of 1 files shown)


tandraschko:
why should one activate our CSP with unsafe?

melloware:
CKEDitor does not work unless you use the unsafe-inline directive.  See: https://ckeditor.com/docs/ckeditor5/latest/builds/guides/integration/csp.html

So i am heavily testing the PrimeFaces Extensions showcase with CSP and resolving issues.  We can't constrict what PF users may need for a CSP Policy because of scenarios like this where you use 3rd party JS that needs to use unsafe-eval or unsafe-inline.

melloware:
So far on the PFE showcase to make GChart work and CKEDitor work I am using this custom policy.
```xml
	<context-param>
	</context-param>
	<context-param>
		<param-value>script-src 'self' 'unsafe-inline' https: *.googleapis.com</param-value>
	</context-param>
```

tandraschko:
Thats what i mean.
If you use unsafe-inline, you don't need to use our CSP. 
99% of our CSP code is exactly to avoid inline event handlers.

melloware:
Ahh you are saying just send back a CSP header in the response instead of using PF CSP?

tandraschko:
yep, exactly
when using unsafe-*, you would completely remove our nonce handling, which means that our whole CSP stuff is useless

in theory a user could use unsafe-inline but not use unsafe-eval - then our nonce would be usefull
but if you say that you can't combine nonce and unsafe-* by specs, the whole thing is useless?

melloware:
Yep here is the error Chrome reports when I just use the NONCE with CKEDITOR.
```
Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' https: *.googleapis.com 'nonce-NzI4MzdjNjItODViNy00MmIxLThlNTEtZjkzZDZiNjkyZGZj'". 

Either the 'unsafe-inline' keyword, a hash ('sha256-km7Z7Q/deuGnP1CMlC9+RCOTa0uat5E0irIAoiuErFk='), or a nonce ('nonce-...') is required to enable inline execution.
```

tandraschko:
yep, then -1 for the PR
please document that PF CSP should not be used for unsafe stuff, then a user should use set their own response headers

melloware:
Will do

PR #5328, opened 2019-11-13
Fixes #5320 - Timeline: timeZone and browserTimeZone are ignored when manually set

### LINKED ISSUE #5320: Timeline: timeZone and browserTimeZone are ignored when manually set
I have an issue using Timeline component in different timezones: we want to force the visualization of the events in an specific timezone, but if we force the timezone in Timeline component, is completely ignored (both _timeZone_ and _browserTimeZone_ parameters). BTW I use Primefaces 7.0.8.

The issue can be simply reproduced using any timeline demo like this: https://www.primefaces.org/showcase/ui/data/timeline/allEvents.xhtml

```
<p:timeline id="timeline" value="#{allEventsTimelineView.model}"
            editable="true" eventMargin="10" eventMarginAxis="0" 
            start="#{allEventsTimelineView.start}" 
            end="#{allEventsTimelineView.end}" 
            timeZone="Europe/Paris"
            browserTimeZone="Europe/Paris">
```

Start the application and access to the app through another device with another timezone on its computer (like American or Asian timezone). You will see that all events will be displaced in time as they are displayed with the device timezone instead of the server timezone as specified.

jasonex7:
Fixes #5320

### DIFF
--- a/src/main/java/org/primefaces/component/timeline/TimelineBase.java
+++ b/src/main/java/org/primefaces/component/timeline/TimelineBase.java
@@ -52,6 +52,7 @@ public abstract class TimelineBase extends UIComponentBase implements Widget, RT
         varGroup,
         locale,
         timeZone,
+        clientTimeZone,
         height,
         minHeight,
         maxHeight,
@@ -177,6 +178,14 @@ public abstract class TimelineBase extends UIComponentBase implements Widget, RT
 
     public void setTimeZone(Object timeZone) {
         getStateHelper().put(PropertyKeys.timeZone, timeZone);
+    }
+
+    public Object getClientTimeZone() {
+        return getStateHelper().eval(PropertyKeys.clientTimeZone, null);
+    }
+
+    public void setClientTimeZone(Object clientTimeZone) {
+        getStateHelper().put(PropertyKeys.clientTimeZone, clientTimeZone);
     }
 
     public String getHeight() {
--- a/src/main/java/org/primefaces/component/timeline/TimelineBase.java
+++ b/src/main/java/org/primefaces/component/timeline/TimelineBase.java
@@ -142,6 +142,7 @@ public abstract class TimelineBase extends UIComponentBase implements Widget, RT
         getStateHelper().put(PropertyKeys.var, var);
     }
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/primefaces/component/timeline/TimelineBase.java
```diff
@@ -571,6 +572,7 @@ public void setDropScope(String dropScope) {
         getStateHelper().put(PropertyKeys.dropScope, dropScope);
     }
 
+    @Override
     public String getDir() {
         return (String) getStateHelper().eval(PropertyKeys.dir, "ltr");
     }
```
melloware, line 587: You can remove getDir() entirely as RTLAware interface is a default interface that does exactly this right?

melloware (changes requested):

melloware (approved):

christophs78:
I think we should bring back browserTimeZone (or clientTimeZone when we look at Schedule) for this. IMO the timeZone used server-side should be decoupled from the clientTimeZone. (eg we want to set server-side timeZone but do not want to set a fixed clientTimeZone)

melloware:
we should standardize the name like clientTimeZone for both Schedule and Timeline

jasonex7:
So `clientTimeZone` will affect vis Timeline component and `timeZone` will affect TimelineEvent start and end dates, is that correct?

christophs78:
Thanks, looks good. 
(server-side) timezone is the "anchor". 
We send `start: new Date('2019-11-14T14:00:00+01:00'), end: new Date('2019-11-14T22:00:00+01:00')` from the server to the client.
OOTB the client (=browsers) transfers this into his (client)TimeZone. As an exmple: you want to know when the formular 1 race start. In this case you expect the browser to transfers this into your timezone. So you know when to turn on the tv.
When you attend an conference in madrid and look on the website of this converence you expect to see the starting-time in the (client)TimeZone of Madrid. For this usecase you would set clientTimeZone.

christophs78:
@melloware : IMO this PR is ready to be merged.

tandraschko:
all right
please also change the migration guide

Issue #5337, opened 2019-11-15
TimelineEvent and DefaultScheduleEvent: provide builder


christophs78:
deprecate args constructors; like https://github.com/primefaces/primefaces/blob/master/src/main/java/org/primefaces/model/menu/DefaultMenuItem.java

also see https://github.com/primefaces/primefaces/issues/5335

PR #5338, opened 2019-11-15
#5337 TimelineEvent and DefaultScheduleEvent: provide builder

christophs78:
... and deprecate args constructor

### DIFF
--- a/src/main/java/org/primefaces/model/timeline/TimelineEvent.java
+++ b/src/main/java/org/primefaces/model/timeline/TimelineEvent.java
@@ -300,6 +300,24 @@ public class TimelineEvent<T> implements Serializable {
         return new Builder<T>();
     }
 
+    public static <T> Builder<T> builder(TimelineEvent<T> event) {
+        Builder<T> builder = builder();
+
+        builder.id(event.id);
+        builder.data(event.data);
+        builder.startDate(event.startDate);
+        builder.endDate(event.endDate);
+        builder.editable(event.editable);
+        builder.editableTime(event.editableTime);
+        builder.editableGroup(event.editableGroup);
+        builder.editableRemove(event.editableRemove);
+        builder.group(event.group);
+        builder.title(event.title);
+        builder.styleClass(event.styleClass);
+
+        return builder;
+    }
+
     @Override
     public int hashCode() {
         int hash = 5;
--- a/src/main/java/org/primefaces/model/timeline/TimelineEvent.java
+++ b/src/main/java/org/primefaces/model/timeline/TimelineEvent.java
@@ -92,46 +92,90 @@ public class TimelineEvent<T> implements Serializable {
         this.id = UUID.randomUUID().toString();
     }
 
+    /**
+     * @deprecated Use {@link TimelineEvent.Builder} instead.
+     */
+    @Deprecated
     public TimelineEvent(T data, LocalDateTime startDate) {
         this(data, startDate, null, null, null, null);
     }
 
+    /**
+     * @deprecated Use {@link TimelineEvent.Builder} instead.
+     */
+    @Deprecated
     public TimelineEvent(T data, LocalDate startDate) {
         this(data, startDate.atStartOfDay(), null, null, null, null);
     }
 
+    /**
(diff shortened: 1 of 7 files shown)


Rapster:
My bad I wasn't clear enough, I think event shoud be immutable object unless there are exceptional cases, and indeed I see we use setters sometimes... I'm not an expert in this component, so if you know it well, feel free to do some adjustments.

One thing for sure, is that we can't mix builder pattern and multi-constructors, for 8.0 we should deprecate all of the constructors with args.

By the way, if event has to be mutable maybe you can choose chain methods over build pattern: it's a  lot more simple to implement

christophs78:
Ok, i´ll look into making the events immutable objects in a separate PR.

christophs78:
@Rapster: Added one more commit - is this what you suggested/expected?

christophs78:
Method-Chaining does not seem to be a good idea...
```
15-Nov-2019 21:51:58.186 SCHWERWIEGEND [http-nio-8080-exec-10] org.primefaces.application.exceptionhandler.PrimeExceptionHandler.logException Property [title] not writable on type [org.primefaces.model.DefaultScheduleEvent]
 javax.el.PropertyNotWritableException: Property [title] not writable on type [org.primefaces.model.DefaultScheduleEvent]
	at javax.el.BeanELResolver$BeanProperty.write(BeanELResolver.java:315)
	at javax.el.BeanELResolver$BeanProperty.access$100(BeanELResolver.java:275)
	at javax.el.BeanELResolver.setValue(BeanELResolver.java:120)
```

tandraschko:
i would not change the setter. It's defined that setter are void, we should just add a additional Builder.

christophs78:
I reverted the "method chaining" commit.

Also tried to update showcase and run into the following:

Old code:
```
DefaultScheduleEvent event = new DefaultScheduleEvent("Champions League Match", previousDay8Pm(), previousDay11Pm());
event.setDescription("Team A vs. Team B");
eventModel.addEvent(event);
```

New code:
```
DefaultScheduleEvent event = DefaultScheduleEvent.builder()
		.title("Champions League Match")
		.startDate(previousDay8Pm())
		.endDate(previousDay11Pm())
		.description("Team A vs. Team B")
		.build();
eventModel.addEvent(event);
```

This requires more code than before. I understand that we should deprecate/remove the args constructors from an architecture-standpoint. But from an usability and backward-compatibility-view - should we really deprecate/remove ALL the args constructors?

tandraschko:
i think so
maybe we can also leave the all-args constructor?

src/main/java/org/primefaces/model/timeline/TimelineEvent.java
```diff
@@ -146,6 +190,10 @@ public TimelineEvent(String id, T data, LocalDateTime startDate, LocalDateTime e
         this.styleClass = styleClass;
     }
 
+    /**
+     * @deprecated Use {@link TimelineEvent.Builder} instead.
+     */
+    @Deprecated
     public TimelineEvent(TimelineEvent<T> event) {
```
Rapster, line 197: Add a `TimelineEvent#builder(TimelineEvent)` method to replace copy constructor

Rapster (commented):
> Method-Chaining does not seem to be a good idea...
> 
> ```
> 15-Nov-2019 21:51:58.186 SCHWERWIEGEND [http-nio-8080-exec-10] org.primefaces.application.exceptionhandler.PrimeExceptionHandler.logException Property [title] not writable on type [org.primefaces.model.DefaultScheduleEvent]
>  javax.el.PropertyNotWritableException: Property [title] not writable on type [org.primefaces.model.DefaultScheduleEvent]
> 	at javax.el.BeanELResolver$BeanProperty.write(BeanELResolver.java:315)
> 	at javax.el.BeanELResolver$BeanProperty.access$100(BeanELResolver.java:275)
> 	at javax.el.BeanELResolver.setValue(BeanELResolver.java:120)
> ```

That's funny, the return type shouldn't be a problem here, did you modify the interface also? Anyway, let's not change the interface. Builder should be good enough

christophs78:
Done. Showcase-PR: https://github.com/primefaces/showcase-facelift/pull/88

Issue #5335, opened 2019-11-14
Improve quality gate before 8.0


Rapster:
Before 8.0 gets released, I'd like to clean up few things based on sonarcloud analysis. The idea is just to apply the simplest fix to the squids. I'm planning to start with [maintainability section](https://sonarcloud.io/component_measures?id=org.primefaces%3Aprimefaces&metric=Maintainability). Of course, I'll do whatever I can in the most reasonable way. I'll certainly ping you for advice.

tandraschko:
please also check why our sonar-project.properties isn't used correctly
i always deactived the rules in our properties first but doesn't work, so i had to use the UI to disable checks.

tandraschko:
What should we do about:
https://sonarcloud.io/project/issues?id=org.primefaces%3Aprimefaces&open=AWWwlzIHlgbYnohYUhY0&resolved=false&types=VULNERABILITY

Rapster:
2 options: disable the rule or turning into immutable list.

I don't think sonarcloud consider the properties file, see https://stackoverflow.com/a/45095506/4605161

Rapster:
@tandraschko I created that ticket so we can link commits, and also I think it'd be better to commit per squid and not randomly...

Rapster:
squid:S2176 I think it has a good point and we shouldn't disable it, mostly we should rename all into ClientXXXConverter or ClientXXXValidator

Rapster:
squid:S1611 "Parentheses should be removed from a single lambda input parameter when its type is inferred" Why disable it?

christophs78:
Ok. I´ll look into TimelineEvent und ScheduleEvent. (introduce builder, deprecate args constructors)

Rapster:
@christophs78 Could you have a look at [this one](https://sonarcloud.io/project/issues?id=org.primefaces%3Aprimefaces&open=AWxM5-Nqdg2jiL0yV2Wf&resolved=false&rules=squid%3AS2175) please

tandraschko:
i think it's right to disable squid:S2176; also think about SelectOneMenu or InputText.

Rapster:
What about them? Their hierarchy looks good to me

tandraschko:
We will also have classes with the same name as in the JSF API/Impl, therefore S2176 doesn't make sense for me.

Rapster:
Well, I'd say using same names as api was not the best thing to do in a first place... It's definitely an unusual move... Anyway I'll check it out later

[...]

PR #5341, opened 2019-11-15
#5335 DatePicker SonarCloud-Issue

christophs78:
https://sonarcloud.io/project/issues?id=org.primefaces%3Aprimefaces&open=AWxM5-Nqdg2jiL0yV2Wf&resolved=false&rules=squid%3AS2175

### DIFF
--- a/docs/8_0/components/datepicker.md
+++ b/docs/8_0/components/datepicker.md
@@ -67,7 +67,7 @@ ajax selection and more.
 | appendTo | null | String | Appends the dialog to the element defined by the given search expression.
 | triggerButtonIcon | null | String | Icon of the datepicker element that toggles the visibility in popup mode.
 | disabledDates | null | List<java.time.LocalDate>, List<java.util.Date> (deprecated) | List of dates that should be disabled.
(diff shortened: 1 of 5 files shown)


tandraschko:
-1
this just changes 1 place
if you look in the DatePicker base, it's defined as non-generic list.
We should first define what can be inside the disabledDays list and then change everything.

christophs78:
I had not been happy with this change too. Do you think it got better with my second commit. (I´m still not totally happy myself.) Or should we revert this and use the old non-generic list for disabledDays in DatePicker to keep SonarCloud happy?

tandraschko:
why you are not happy?
If it can only contain ints, the change seems to be good?

christophs78:
We stilll have
- `return (List<Integer>) getStateHelper().eval(PropertyKeys.disabledDays, null);` and
- and we have to keep other attributes like disabledDates as untyped list because it may contain String or java.util.Date or java.time.LocalDate

But it´s not a perfect world. :-) I think this PR makes the code a bit better than it was before and we can merge it.

tandraschko:
yep :)

PR #5346, opened 2019-11-19
Fix #5345: SplitButton support of p:confirm

### LINKED ISSUE #5345: SplitButton: Support for p:confirm
Issue #507 introduced support for `p:confirm` in the menu items of the split button. This issue is about to support the split button itself.

PR is on the way.

fanste:
Add support for p:confirm on the default button of p:splitButton


melloware (approved):

Rapster:
Safe to merge this one no? WDYT?

melloware:
Looks good to me

tandraschko:
I think we have to use the statehelper instead a var, to be compatible for MyFaces View pooling

tandraschko:
ok, we have a general problem here. we sometimes have normal variables, especially in DataTable.

For now it would be good to change all confirmationScripts to the stateHelper. Can you do it @fanste ?

tandraschko:
will do it

PR #5347, opened 2019-11-19
Fix #3602: Datatable input fields

### LINKED ISSUE #3602: DataTable: RowSelect with any input field in row without using rowEditor => some keys won't work
## 1) Environment
- PrimeFaces version: 6.2
- Affected browsers: IE 11 (FF works)

## 2) Expected behavior
Pressing space, enter or arrow keys should do the desired actions inside the input field

## 3) Actual behavior
Pressing space, enter or arrow keys change the focused row of the selectable data table.

## 4) Steps to reproduce

- Setup a dataTable with selectionMode=single and add input fields as column children. Do not use the feature `rowEditor`.
- Enter something in the input field while no row is selected in the datatable. Every character will work, even space, enter, etc..
- Select a row and again enter something in the input field. Space, enter, etc won't work anymore.

## 5) && 6)
See related issues #2000 #2015

The fix for #2015 added the following code lines
```
if($(e.target).is(':input') && $this.cfg.editable) {
    return;
}
```

Why not just drop `&& $this.cfg.editable` and always ignore space, enter, etc when inside an input field?

fanste:


melloware:
Assigned to @mertsincan to review

mertsincan:
Thanks a lot for the PR!

PR #5360, opened 2019-11-21
Add correct articles

fdenzer:
As turkish has no grammatical articles, this commit fixes some too-open, article-free cases in readme.md by adding the/a as I saw fit.


tandraschko:
Thanks

PR #5366, opened 2019-11-22
squid:S2386 Immutable Sets instead of String[]

melloware:
Code is frozen for RC1 but i wanted to put this out there for review.

### DIFF
--- a/src/main/java/org/primefaces/component/api/MenuItemAware.java
+++ b/src/main/java/org/primefaces/component/api/MenuItemAware.java
@@ -34,15 +34,17 @@ import javax.faces.event.AbortProcessingException;
 import javax.faces.event.ActionEvent;
 import javax.faces.event.FacesEvent;
 import java.util.List;
+import java.util.Set;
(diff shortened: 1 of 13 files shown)


Rapster:
-1 for `MenuItemAware` These attributes are here so they're not instanciate over and over again, but making that change is countering the initial idea. Either we leave it as it is, or we don't make them as attributes

melloware:
Agreed I didn’t love that either. Any suggestion on a better way to do it?  Or just revert that and leave that class as is?

Rapster:
Yeah right now, I don't have a better suggestion except leave it as it is

melloware:
Reverted MenuItemAware as suggested.

melloware:
@tandraschko now that 8.0RC1 is out the door is this safe to merge?

tandraschko:
Nope, we should so it after the final release

Rapster:
Safe to merge yes, don't see why not

melloware:
I marked it for 9.0

Issue #5363, opened 2019-11-22
DatePicker: inline="true" broken


christophs78:
I broke it with https://github.com/primefaces/primefaces/commit/24ea45ed86d3e8f4d865e3f38d28338c9e3ffb03.

I have to add a fix to ignore appendTo="@(body)" for inline="true".

PR #5367, opened 2019-11-22
#5363 DatePicker: inline="true" broken; refactor

christophs78:
@Rapster : this way better?


tandraschko:
IMO it was better before :D

maybe its better to move those checks to the renderer and leave @(body) the default in the datepickerbase

Rapster:
Yes +1 to move that new method in the renderer

christophs78:
satisfied?

mertsincan:
Hi,

Thanks a lot for the PR, @christophs78 ;) But, I think we can add if-check to datepicker core widget to control appendTo and inline. If the calendar has inline mode, the appendTo is unnecessary. 

Thanks a lot again ;)

Issue #4182, opened 2018-10-31
The input field doesn't update when the "current date" button is clicked on Calendar


mertsincan:
Reported By PRO User;
> I would like to use p:calendar for keep datefrom and dateto (date and time) with maxdate. ... When you click on current date, change are apply on the calendar but not on the textbox like it's do when you slide with mouse ?

PR #5375, opened 2019-11-26
Fix #5374: Calendar disabled style

### LINKED ISSUE #5374: Calendar: inline mode is not styled correctly with disabled=true
## 1) Environment
- PrimeFaces version: lastest trunk

## 2) Expected behavior
`<p:calendar mode="inline" disabled="true" />` should get the styleClass `ui-state-disabled`

## 3) Actual behavior
The component is internally disabled but the styleClass `ui-state-disabled` is not applied

PR #5375

fanste:


melloware (approved):

tandraschko:
also whats with validation failed style?
I'm sure we had it. Can you check the history, if it was removed while refactoring both calendar and datepicker?

tandraschko:
also see BaseCalendarRenderer#encodeInput

melloware:
It looks like DatePicker and Calendar work slightly differently in "popup" mode. The Calendar uses an extra DIV wrapper while the DatePicker does not.

fanste:
The inline calendar is rendered at runtime by jQuery UI and I can't find a options that allows me to set a style class on the generated DOM. If I add the style class to the DIV wrapper it simply looks wrong.
grafik


I can't find a revision where `disabled` and `valid` styles should have been set for the inline calendar.

fanste:
Well.. found it - at least for `disabled`.

The fix for #4182 introduced a short delay in generating the HTML of the calendar (`setTimeout`). The affected method is `_updateDatepicker` from jQuery UI datepicker.

This method is used by `_inlineDatepicker` which again calls `_disableDatepicker`. `_disableDatepicker` tries to set `ui-state-disabled` on the not yet generated HTML. We remember the delay in `_updateDatepicker`.

@mertsincan  you created #4182, how should we fix the wrong behavior?

fanste:
Provided a alternative fix for `disabled` based on the same ugly way ;)

The fix of 4182 seems to be copied from the timepicker addon which is full of such hacks...

src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.js
```diff
@@ -160,6 +160,13 @@
                 }, 0);
         }
     };
+
+    $.datepicker._base_disableDatepicker = $.datepicker._disableDatepicker;
+    $.datepicker._disableDatepicker = function(target) {
+        setTimeout( function() {
+        	$.datepicker._base_disableDatepicker(target);
+        }, 0);
+    }
```
melloware, line 170: I would put a comment above this "GitHub #5374 Calender inline=true not styled correctly" and also update the file "extensions.txt" which is located in that same directory.

melloware (changes requested):

melloware (approved):

Issue #4627, opened 2019-03-13
DatePicker : when showTime true and maxdate set, cannot select time today


carlbradwell17:
## 1) Environment
- PrimeFaces version: 7.0.-RC3

## 2) Expected behavior
If I use the datepicker component, and set showTime="true" and I set maxdate="#{now}", I should be able to select any date/time up until the minute the page was loaded.

## 3) Actual behavior

If I select the current day, the hour selector is disabled, but I can advance the minutes beyond the maxdate. I cannot decrease the hours. If I select a different day prior to maxdate, it works as expected.

## 4) Steps to reproduce

create a datepicker with showTime="true" maxdate="#{now}", try to select a time today before/after #{now}. E.g. if the current time is 12:30, it will be impossible to select 11:30 (today), but possible to select 12:40 (today).

## 5) Sample XHTML

<h:form>
 <p:datePicker showTime="true" maxdate="#{now}" />
</h:form>

PR #5377, opened 2019-11-27
Fix #5334, #4627: Datepicker timeOnly/showTime-issues in combination with minDate/maxDate

### LINKED ISSUE #5334: DatePicker: timeOnly; default time should respect minDate and maxDate
# Issue 1
DatePicker in timeOnly-mode defaults to the current time; this does not make sense when minDate and/or maxDate are set
--> DatePicker should check wether current time is inside minDate/maxDate; if not it should default to minDate (when current time is before minDate) respectively maxDate (wehn current time is after maxDate)

# Issue 2
eg DatePicker-input has value 04:00. minDate is 07:00. You can´t use the arrows to change time. DatePicker does not allow this because eg 05:00 ist still before 07:00.
There is no feedback to the user, that there´s a minDate or maxDate defined.

christophs78:


christophs78:
Issues: https://github.com/primefaces/primefaces/issues/4627, https://github.com/primefaces/primefaces/issues/5334

Showcase-PR: https://github.com/primefaces/showcase-facelift/pull/93

Thinks this is one to be reviewed by @mertsincan.

christophs78:
@NicolaIsotta Can you have a look at this one and the Showcase-PR?

christophs78:
@carlbradwell17: Can you test this PR?

melloware:
Assigned to @mertsincan since this touches the core DatePicker code used by PrimeReact/NG/Vue

carlbradwell17:
> @carlbradwell17: Can you test this PR?

Tested this on snapshot build this morning, confirmed it satisfies the Expected Behaviour I raised in #4627.
The maximum selectable time when on today is the time when the page loaded. Changing to a previous day with a greater time, then going back to today resets the time to the maximum.

NicolaIsotta:
I took a quick test.
Consider this datepicker:
`<p:datePicker maxdate="10:00" timeOnly="true"/>`
Select "10:00" using the arrows. Then click the down arrow under the minutes. The value will now be 10:59. Should this be prevented?

christophs78:
> `<p:datePicker maxdate="10:00" timeOnly="true"/>`
> Select "10:00" using the arrows. Then click the down arrow under the minutes. The value will now be 10:59. Should this be prevented?

Yes, this should be prevented.

christophs78:
@NicolaIsotta: Think i catched your corner-case. With the last commit i only allow to use the arrows outside the allowed range when the currently selected value is outside the allowed range.

mertsincan:
Thanks a lot for the PR!

PR #5386, opened 2019-11-29
Fix #5385 - Deprecate DefaultStreamedContent ctors/setters

### LINKED ISSUE #5385: Deprecate DefaultStreamedContent ctors/setters
Should be a immutable object. Also, I think we should remove `ByteArrayContent`, and instead use `DefaultStreamedContent` with a `ByteArrayInputStream`

Rapster:


tandraschko:
hmm, not sure
the setter are used by the Builder, so we can't deprecate it?

Rapster:
Good catch! Fixed :wink:

melloware (approved):

PR #5396, opened 2019-12-03
Fix #5395: Messages allow array of ids

### LINKED ISSUE #5395: Messages: "for" should take an array of ids
p:messages takes just one id at the moment but it would be great if we had the opportunity to refer to more than one id.
The problem is that it's not possible to have a "global" p:messages in the template with a non-global on an included page with more than one validation field. The validation message of the n-th field will not show in the non-global p:messages.

Location: MessagesRenderer - collectFacesMessages
I have already tested the following code:
```java
// key case
            if (forType == null || forType.equals("key")) {
                String[] fors = _for.split(" ");
                for (String messageFor : fors) {
                    Iterator<FacesMessage> messagesIterator = context.getMessages(messageFor);
                    while (messagesIterator.hasNext()) {
                        if (messages == null) {
                            messages = new ArrayList<>();
                        }
                        messages.add(messagesIterator.next());
                    }
                }
            }
            // clientId / SearchExpression case
                String[] fors = _for.split(" ");
                for (String messageFor : fors) {
                    UIComponent forComponent = SearchExpressionFacade.resolveComponent(context, uiMessages, messageFor, SearchExpressionHint.IGNORE_NO_RESULT);
                    if (forComponent != null) {
                        String forComponentClientId = forComponent.getClientId(context);
                        if (!messageFor.equals(forComponentClientId)) {
                            Iterator<FacesMessage> messagesIterator = context.getMessages(forComponentClientId);
                            while (messagesIterator.hasNext()) {
                                FacesMessage next = messagesIterator.next();
                                if (messages == null) {
                                    messages = new ArrayList<>();
                                }
                                if (!messages.contains(next)) {
                                    messages.add(next);
                                }
                            }
                        }
                    }
                }
            }
```

melloware:


tandraschko:
the "key" case was missing, so i did it myself
please see my last commit

PR #5405, opened 2019-12-05
Fixes #5404: Default command multiple instances

### LINKED ISSUE #5404: DefaultCommand: Multiple DefaultCommand instances in same form broken
## 1) Environment
- PrimeFaces version: 7.0.10 (Elite)
- Still worked on the previous version: 7.0.9 (Elite) and 7.0
- Does it work on the newest released PrimeFaces version? Version 7.0.10? No
- Affected browsers: Firefox/Chrome

## 2) Expected behavior
When you define multiple `<p:defaultCommand />` instances in one form, you can define the subsection of the form the DefaultCommand should apply to with the optional `scope` attribute.
If you have two non-overlapping panels in one form, and create a DefaultCommand instance for each one, with the `scope` attribute set to the `id` attribute of the panel, one should expect that pressing ENTER in the first panel would trigger the given defaultCommand of this panel, and doing the same in the second panel would result in the second defaultCommand being executed.

The Feature for multiple DefaultCommands in one form was introduced in #4324 and worked fine until now.

## 3) Actual behavior
Always the first declared DefaultCommand action is triggered, and all the other ones are never triggered. This completely ignored the scope from which the event originated.

## 4) Steps to reproduce
See this example repository: https://github.com/metaxmx/PrimeFacesExampleDefaultCommand

## 5) Sample XHTML
```html
<h:form id="form">

    <p:growl id="growl"/>

    <p>
        Focus a Textbox and press ENTER.
        The associated default action button for the section you are in should be triggered.
    </p>

    <h:panelGroup id="panel1" layout="block">

        <h2>Panel 1</h2>

        <p:inputText id="text1"/>

        <p:commandButton id="actionOne" value="Action One"
                         action="#{example.actionOne}" process="@this" update="growl"/>

        <p:defaultCommand target="actionOne" scope="panel1"/>

    </h:panelGroup>

    <h:panelGroup id="panel2" layout="block">

        <h2>Panel 2</h2>

        <p:inputText id="text2"/>

        <p:commandButton id="actionTwo" value="Action Two"
                         action="#{example.actionTwo}" process="@this" update="growl"/>

        <p:defaultCommand target="actionTwo" scope="panel2"/>

    </h:panelGroup>

    <h2>Outside Panel (no scope)</h2>

    <p:inputText id="text3"/>

    <p:commandButton id="actionThree" value="Action Three"
                     action="#{example.actionThree}" process="@this" update="growl"/>

    <p:defaultCommand target="actionThree" />

</h:form>
```

## 6) Sample bean
```java
@Named("example")
@ApplicationScoped
public class ExampleBean {

    public void actionOne() {
        context.addMessage(null, new FacesMessage("Action One",  "Action One Called") );
    }

    public void actionTwo() {
        context.addMessage(null, new FacesMessage("Action Two",  "Action Two Called") );
    }

    public void actionThree() {
        context.addMessage(null, new FacesMessage("Action Three",  "Action Three Called") );
    }

}

```

## 7) Problem analysis
I believe the problem was introduced by the commit 86250d25a58946bfcc17e85fa6ed335a29588dcb for issue #4984 
Prior to this commit, the keydown-Event for the explicit id of the originating DefaultCommand was triggered.
```javascript
closestForm.trigger('keydown.' + $this.id, {scopeEnter: true});
```
This made sure, that when multiple keydown events were registered, the event handlers for DefaultCommands in different scopes were not triggered by accident.

After the change, now the KeyDown-Event-Object is forwarded to the Form:
```javascript
closestForm.trigger(e, {scopeEnter: true});
```
This results in any registered DefaultCommand event handlers being called, without any information about the originating scope.
As a result, this first registered DefaultCommand event handler will execute (because `{scopeEnter: true}` is given) and then call `e.stopImmediatePropagation();` on the event, so all other defaultCommands are never triggered.

metaxmx:
Fix for issue #5404

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/forms/forms.defaultcommand.js
+++ b/src/main/resources/META-INF/resources/primefaces/forms/forms.defaultcommand.js
@@ -22,7 +22,8 @@ PrimeFaces.widget.DefaultCommand = PrimeFaces.widget.BaseWidget.extend({
             var keyCode = $.ui.keyCode;
 
             data = data || e.data;
-            if (($this.scope && data.scopeEnter) || (!$this.scope && (e.which == keyCode.ENTER))) {
+            if (($this.scope && data.scopeEnter && data.scopeDefaultCommandId === $this.id)
+                    || (!$this.scope && !data.scopeEnter && (e.which == keyCode.ENTER))) {
                 //do not proceed if target is a textarea,button or link
                 if ($(e.target).is('textarea,button,input[type="submit"],a')) {
                     return true;
@@ -40,7 +41,7 @@ PrimeFaces.widget.DefaultCommand = PrimeFaces.widget.BaseWidget.extend({
             this.scope.off('keydown.' + this.id).on('keydown.' + this.id, function (e) {
                 var keyCode = $.ui.keyCode;
                 if (e.which == keyCode.ENTER) {
-                    closestForm.trigger(e, {scopeEnter: true});
+                    closestForm.trigger(e, {scopeEnter: true, scopeDefaultCommandId: $this.id});
                     //e.preventDefault();
                     e.stopPropagation();
                 }


alfonx:
This regression hit us with 7.0.10, a fix in 7.0.11 would be great!

mertsincan:
Thanks a lot @metaxmx ;)

PR #5406, opened 2019-12-06
FileUpload - close input streams to prevent file handlers leak

pawwal:
It happens that UploadedFile.getInputStream() points to a real temporary file (eg. under Undertow). Therefore if this InputStream is not being closed file handler leaks, which may eventually lead to "too many open files" error.


Rapster:
Please create an issue first, following the issue template (don't forget to show the stacktrace)

pawwal:
Issue created: https://github.com/primefaces/primefaces/issues/5408

PR #5407, opened 2019-12-06
Revert commit #5162

### LINKED ISSUE #4185: Datatable: Wrong resolveStaticField result used in facelet tag
## 1) Environment
- PrimeFaces version: 6.2
- Affected browsers: Any

## 2) Expected behavior
Condering this facelet tag:
```xml
                ...>

    <o:tagAttribute name="value"/>
    <o:tagAttribute name="sortBy"/>
    
    <p:dataTable id="#{id}"
                 widgetVar="wv_#{component.id}"
                 value="#{value}"
                 var="item"
                 sortBy="#{sortBy}">

        <ui:insert/>

    </p:dataTable>
</ui:composition>
```

Intilization of this `<p:dataTable />` should not be a problem

## 3) Actual behavior
Later on, a business exception will be thrown since as DataTable#sortField is evaluated as "sortBy" (and it's a not a property of a bean). See DataTable#resolveStaticField:

```java
    public String resolveStaticField(ValueExpression expression) {
        if(expression != null) {
            String expressionString = expression.getExpressionString();
            expressionString = expressionString.substring(2, expressionString.length() - 1);      //Remove #{}

            return expressionString.substring(expressionString.indexOf(".") + 1);                //Remove var
        }
        else {
            return null;
        }
    }
```

The result of this algorithm will be "sortBy"

## 4) Steps to reproduce
Create a facelet tag, see above

## 5) Sample XHTML
See above

## 6) Proposal
First I came up with this (using regex):
```java
    @Override
    public String resolveStaticField(ValueExpression expression) {
        if (expression != null) {
            String expressionString = expression.getExpressionString();
            Matcher matcher = Pattern.compile("^#\\{\w+\\.(.*)\\}$").matcher(expressionString);
            if (matcher.find()) {
                return matcher.group(1);
            }
        }

        return null;
    }
```

But I'm afraid performance could be lower... Or should I make a simple test "if expressionString contains var"?

This benchmark makes me wonder though: https://www.logicbig.com/tutorials/core-java-tutorial/java-regular-expressions/performance.html

Improvements

- Move Pattern as a member (should save a lot of time)
- Should we create/update an util to parse that kind of expression?

Rapster:
@mertsincan after sorting out #5162 and realize there was no issue in 7.0. Can you please apply this PR, it revert commits done in the issue and close #4185


mertsincan:
@Rapster 
We have to make this change to support the special o:tagAttribute feature of omnifaces ???

Rapster:
no, tagAttribute is just a nice way to write `<c:set var="foo" value="#{empty foo ? 'sthg' : foo}" />`. I could have used that instead and problem would be the same

Rapster:
Hey @mertsincan!

Happy new year by the way :wink: Can you merge that PR please?

tandraschko:
Can someone please give a short description why we need this PR and what will be broken then?
It's quite hard to follow this discussions.

Rapster:
It fixes #4185 originally merged in 7.0. In short, using `<c:set var="foo" value="#{empty foo ? 'sthg' : foo}" />` and `sortBy="#{foo}` will result as sorting with foo field which is wrong.

In #5162, the user uses EL expression to pass a static value while he should have used `Column#field`, this very same user and I confirmed it works. Basically, reverting the commit was not necessary

tandraschko:
hmm.... Puh..... not easy

To clarify this, we have to exactly define what `sortBy` is and what `field` is and why we have both

Rapster:
I'll be honest with you, I think we should remove that trick that `sortBy` does on the long run. `field` should be meant to indicate what field of `var` bean should be displayed, if `sortable` set to true then you can deduce `sortBy`, and `sortBy` should be used as custom sorting like i18n for example.

As a proof, read https://github.com/primefaces/primefaces/issues/5162#issuecomment-559795286  the user would expect: `field` + `sortable` to enable sorting. He's right, it should but it does not.

I started to work on a new version of DataTable to clean up few things, if I manage to make 75% dataTable working. I'll submit an issue listing the "improvements" to be done. Though I'm not sure how to introduce that since it's a hell of breaking changes already from what I started... And I know where PF stands on that front. But the result is pretty cool, the definition and implementation of `DataTable`/`DataTableRenderer` is a lot of more simple.

Meantime, this PR should be merged...

melloware:
Looks like you have `1 checkstyle issue:
```
[ERROR] /home/travis/build/primefaces/primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java:185: Line matches the illegal pattern 'Trailing whitespace'. [Regexp]
```

Rapster:
Can we merge that PR? It's been on hold for a while now...

melloware (approved):

tandraschko:
@mertsincan WDYT?

melloware:
@Rapster this PR conflicts now or is it no longer valid?

Rapster:
I'll merge it later once I'm done with #6402

tandraschko:
we (i) first need to talk with mert about that.

PR #5412, opened 2019-12-07
Fix #5411 - DataTable#resolveDynamicField: better impl

### LINKED ISSUE #5411: DataTable#resolveDynamicField: better impl
Use regex instead just like `DataTable#resolveStaticField`

Rapster:


src/main/java/org/primefaces/component/datatable/DataTable.java
```diff
@@ -665,31 +668,29 @@ public String resolveStaticField(ValueExpression expression) {
...
-        // new syntax is:
-        // #{column.property} or even a method call
-        if (expressionString.startsWith("#{" + getVar() + "[")) {
-            expressionString = expressionString.substring(expressionString.indexOf('[') + 1, expressionString.indexOf(']'));
-            expressionString = "#{" + expressionString + "}";
-
-            ValueExpression dynaVE = context.getApplication()
-                    .getExpressionFactory().createValueExpression(elContext, expressionString, String.class);
-            return (String) dynaVE.getValue(elContext);
+        Matcher matcher = DYNAMIC_FIELD_REGEX.matcher(exprStr );
```
melloware: @Rapster I feel like some Javadoc is necessary here about what the REGEX is doing and examples of what it is parsing.  It will help future devs because REGEX can be hard to distinguish what it is doing without examples.

src/main/java/org/primefaces/component/datatable/DataTable.java
```diff
@@ -665,31 +668,29 @@ public String resolveStaticField(ValueExpression expression) {
...
+        if (matcher.find()) {
+            exprStr = matcher.group(1);
+            exprVE = context.getApplication().getExpressionFactory()
+                    .createValueExpression(elContext, "#{" + exprStr  + "}", String.class);
         }
```
melloware, line 695: Maybe this whole block can be a sub method and you can write unit tests that show all the permutations of of the REGEX being parsed.  Or maybe just break the regex parse and matcher.group(1) return into a method so we can unit test just that?

melloware (changes requested):

src/main/java/org/primefaces/component/datatable/DataTable.java
Rapster: That's something you can do already in `DataTableTest#testResolveDynamicField`

src/main/java/org/primefaces/component/datatable/DataTable.java
Rapster: I added javadoc on the field iteself, and rename the regex to help understand what it does

Rapster:
I wonder if we shouldn't simply drop the backward compatibility on this? @melloware @tandraschko

melloware:
Do you mean drop the syntax of `#{car[column.property]}`?  If so I would be OK with it.

Rapster:
Yes that's what I meant :wink:

melloware (approved):

melloware:
@Rapster I am good with this PR but i think we should leave the backwards compatibilty.

PR #5430, opened 2019-12-11
Fix #5429: CSS replace zoom with transform

### LINKED ISSUE #5429: CSS: Replace non-standard zoom property
Go to the showcase page in Firefox 71+: https://www.primefaces.org/showcase/ui/panel/accordionPanel.xhtml

In the console it reports:

> This page uses the non standard property “zoom”. Consider using calc() in the relevant property values, or using “transform” along with “transform-origin: 0 0”.

They are correct if you look here its a non standard CSS which used to be used to fix IE8 issues.
https://developer.mozilla.org/en-US/docs/Web/CSS/zoom

> This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.

It should be replaced with:
```css
transform: scale(1); /* Standard Property */
transform-origin: 0 0;  /* Standard Property */
```

**Components affected:**
- [x] Accordion
- [x] AutoComplete
- [x] Dialog
- [x] Forms
- [x] Menu
- [x] Paginator
- [x] Tabview
- [x] Tree
- [x] Omega Theme (@mertsincan)

melloware:
Everything but the Omage theme which has to be done by @mertsincan


tandraschko:
does it work correctly in all other browsers?

melloware:
I didn't notice anything different in IE11, Firefox, Chrome, Edge but wanted @mertsincan or @onursenture  or one of the PrimeTek CSS guys to weigh in on this.

mertsincan:
Thanks a lot, @melloware ;)

mertsincan:
Reverted this PR.

PR #5436, opened 2019-12-12
Schedule: Fix ScheduleRenderer bug - only LocalDateTime properties ar…

mnicorici68:
After removing support for Date type in Schedule component, into ScheduleRenderer.encodeEventsAsJSON(), the dynamicProperties from ScheduleEvent are all handled as LocalDateTime.

I have readed the type check on putting those dynamic properties into the JSON object.


tandraschko:
Thanks. Merged it but please create a issue next time. Otherwise we dont apply PRs

PR #5448, opened 2019-12-16
Fix #5431: Fileupload pointer issue in Firefox

### LINKED ISSUE #5431: FileUpload: button doesn't work for 100% in advanced mode
http://primefaces.org/showcase/ui/file/upload/auto.xhtml

The "hand" cursor isn't shown everywhere in the button and sometimes the click doesn't open the upload window (at least in harmony e.g.).

melloware:


tandraschko:
Does it Work in all browsers and all Fileupload modes?

melloware:
Tested Basic, Single, and Auto modes in IE11, Edge, Chrome and Firefox and the pointer is correct in all browsers.

tandraschko:
Then merge plz

PR #5457, opened 2019-12-18
Fix #5429: CSS replace zoom with transform

### LINKED ISSUE #5429: CSS: Replace non-standard zoom property

melloware:
Fixed including the SelectOne RTL layout.


mertsincan:
For me, Ok ;) Thanks a lot for the PR ;)

PR #5462, opened 2019-12-18
Fix #5461: destroy() call correct super

### LINKED ISSUE #5461: DatePicker: Calling incorrect super.destroy().
DatePicker is calling the wrong super method for destroy.

https://github.com/primefaces/primefaces/blob/b1cfde71c09f32ad3711dfb66a3bafddf843cee9/src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js#L104-L110

It should be `this._super();` not `this._super(cfg);`

melloware:


tandraschko:
+1

PR #5467, opened 2019-12-23
Fix #5466: InputTextArea global=false for autocomplete

### LINKED ISSUE #5466: InputTextArea: Introduce "query" behavior to allow to set global="false"
See forum post: https://forum.primefaces.org/viewtopic.php?f=3&t=60712

When using "completeMethod" of InputTextArea it should send its AJAX with global=false to not trigger the AJAX status similar to the AutoComplete component search method.

See AutoComplete: https://github.com/primefaces/primefaces/blob/1ba97c624b579088c2890f7f80232948eb39d7ea/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js#L1008

melloware:


tandraschko:
9.0 + add it to the migration guide

melloware:
OK I moved it to 9.0 Milestone. I will wait to update the Migration Guide until we merge this in and start a new 9.0 section.

tandraschko:
i think we should make it false per default AND add a possibility to overwrite it with p:ajax

melloware:
And we should be consistent with AutoComplete component.

tandraschko:
thanks for the PR but introduced the query event

Issue #5483, opened 2019-12-28
DataView (and other UIData-impls): add multiViewState


christophs78:
DataTable and DataList have it already. Now add it to...
- [x] DataView
- [x] DataGrid
- [ ] TreeTable

Rapster:
I'm working on a PR for the reset thing, and I realized almost every UIData component should support multiview state, so I think you can make this ticket bigger 😉

christophs78:
Yep. I´d would be nice to have multiViewState for all UIData-implementations. Just the question, where to start and where to stop. Where to put our effort, ...

tandraschko:
added TreeTable, not sure how easy it is

tandraschko:
I will create a new issue for the DataView tracking for RC3

melloware:
@tandraschko is working on TreeTable now: https://github.com/primefaces/primefaces/issues/6592

tandraschko:
Will be handled in #6592

Issue #5482, opened 2019-12-28
multiViewState: add component reset option


christophs78:
## 1) Environment
- PrimeFaces version: 8.0-SNAPSHOT
- Does it work on the newest released PrimeFaces version? Version? No
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) No
- Application server + version: all
- Affected browsers: all

## 2) Expected behavior
`PrimeFaces.current().multiViewState().clear...` should also reset the DataTable/DataList.

## 3) Actual behavior
MultiViewState is reseted, but it has no impact on the current view/XHTML.

## 4) Steps to reproduce

1. https://www.primefaces.org/showcase/ui/data/datatable/tableState.xhtml
2. Click on "clear table state"

## 5) Sample XHTML
https://www.primefaces.org/showcase/ui/data/datatable/tableState.xhtml

## 6) Sample bean
`org.primefaces.showcase.view.data.datatable.TableStateView#clearTableState`

(With 8.0 we explicit need to call reset() on the DataTable after clearing the multiViewState. Maybe we should reset DataTable/DataList per default or we should offer an additional parameter/method to control whether DataTable/DataList should be reset.)

Rapster:
I wonder if I should remove the consumer, I introduced this so I can do later a reset... I think I should have introduced reset option instead of consumer, my bad...

christophs78:
IMO you did overall a realy good job in improving the handling of MultiViewState with 8.0.
Related to replacing the consumer with a reset-option: Is there something als than reseting we may do with the consumer?

tandraschko:
@Rapster i commited a small refactoring, please review
i also changed the showcase
could someone please adjust the docus? i dont have much time this week

Rapster:
Arf... I was actually working on those changes, please next time let me know you plan to do something :wink:

PR #5484, opened 2019-12-28
#5483: DataView: add multiViewState

christophs78:
Showcase-PR: https://github.com/primefaces/showcase-facelift/pull/110


Rapster:
I would go differently about it. Have an interface `MultiViewStateAware<T>` with methods:
- isMultiViewState
- restoreState: void
- getState(boolean create): T (I think you should be able to have a default impl here)
- reset: void (will be called by PrimeFaces#multiViewState facade)

On top of my head, that should be enough

Rapster:
Don't bother, i'll do the interface thingy as I need it for #5482

tandraschko:
@christophs78 can you adjust your PR? as we had many changes now regarding multiViewState, i think we can apply it

christophs78:
Yes, i´ll update this PR.

christophs78:
Updated this PR and showcase-PR.

tandraschko:
Can you call it #saveMultiViewState? Otherwise it's to much relation with the JSF viewstate

tandraschko:
and #isMultiViewState probably misses a @Override

christophs78:
Good points. Done.

tandraschko:
Thanks, perfect!

PR #5487, opened 2019-12-29
Fix #5482 - multiViewState: add component reset option

### LINKED ISSUE #5482: multiViewState: add component reset option

Rapster:


tandraschko:
does it consider that only components should be reseted, when the current view is equals?

Rapster:
No I give the option to reset it or not

tandraschko:
I mean:

`public void clear(String viewId, String clientId, boolean reset) {`
calls
`private void reset(String clientId) {`

independent of the viewId - this is not correct and a performance loss if you always try to lookup components, which will never be in this view

Rapster:
Good catch :wink: Fixed!

christophs78:
Excellent work!
One minor addition: Should we note within the JavaDoc, it´s fine to pass null for clientIdConsumer?

Rapster:
Just a thought, since MVS API is brand new, maybe it worth the shot to merge this in 8.0 (instead of reintroducing another API in 9.0 which implies deprecation, backward compatibility, migration guid etc.)

christophs78:
Same opinion as @rapster: We should merge this for 8.0. Change MVS only once. And not some changes with 8.0 and some more with 9.0.

christophs78:
IMO putting this PR into 8.0 would be compareable to what happend to LazyDataModel between 8.0 RC1 and RC2. LazyDataModel was modified for 8.0. We saw the LazyDataModel-modification was not yet perfect with RC1. So we changed it with RC2.
From a theoretic point of view it may make sense to first do something like beta-versions before we start with RCs. I agree it´s not ideal to make API-changes during RC-phase.

tandraschko:
i think we can merge it, we just need to do a detailed review.

@rapster shouldnt MultiViewStateAware#reset be something like MultiViewStateAware#resetMultiViewState? or can we remove it from the interface?

christophs78:
`MultiViewStateAware#reset` had been extraced from the existing `DataTable#reset`.
#reset resets the actual state of the DataTable, DataList ... -instance. (Back to first page, clear filter, clear sort, ...)
IMO it needs to be part of the MultiViewStateAware-interface, so `PrimeFaces.current().multiViewState().clear` is able to not only reset MVS but also reset (triggered via reset-Parameter) the actual state of the component. (The topic we discussed for https://www.primefaces.org/showcase/ui/data/datatable/tableState.xhtml; `TableStateView#clearTableState`.)
IMO it may help to add some JavaDoc to the methods required by `MultiViewStateAware`-Interface.

tandraschko:
yeah but it only resets all things releated to the multiviewstate, so it think it should be in the name

christophs78:
Usecase: We have a XHTML which contains a searchform and DataTable which shows the search-results. (DataTable does not use multiviewstate.) When the user changes the searchform and pushes the search-button the datasource of the DataTable changes. We want to reset the state of the DataTable. (Back to page 1, ...)

Rapster:
I understand both point of view, I must say I thought about renaming reset... Though it can't be removed. I kept reset as a name because it was easier, so I don't have to introduce a new method.

~~I think it depends of the definition we give to `DataTable#reset`, what is it suppose to do? If MultiViewStateAware is meant to do exactly the same, then maybe we should keep the same name.~~ Being said, I just feel this is fiddling... So having a different name is probably better. Even though in  both scenarios, they do the same thing, the goal is slightly different...

tandraschko:
Also not sure but i think resetMultiViewState is for now ok
if we really need a own `reset` function, we can create one later and `resetMultiViewState` can delegate to it.

tandraschko:
@rapster: why not storing a Map<MVSKey, Object> instead <String, Object> in the session?

Rapster:
See https://docs.oracle.com/javaee/7/api/javax/faces/context/ExternalContext.html#getSessionMap--

tandraschko:
what do you mean exactly?
We store a submap in the sessionmap, we can store whatever we want?

Rapster:
Oh yeah my bad, I mixed it up with what I've sent you... So, yes I guess we can store Map<MVSKey, Object> instead of <String, Object>. I'll modify that asap

PR #5488, opened 2019-12-29
Fix #1151: DataExporter calls load method twice on lazy DataTables; "Plan A"

### LINKED ISSUE #763: Exporting from Data Table with lazy data model generates incorrect file if table is manipulated while exporting
The `org.primefaces.component.export.Exporter.exportAll(FacesContext, DataTable, Object)` method manipulates the `DataTable` object it receives, establishing the first row as 0, the page size as the total number of records, then it loads the "page" (the whole list) and it exports it, row by row. Finally, it restores the original values of the different variables and loads again the original page the user was seeing.

``` java
if(lazy) {
        if(rowCount > 0)
             {
                table.setFirst(0);
                table.setRows(rowCount);
                table.clearLazyCache();
                table.loadLazyData();
             }

            for(int rowIndex = 0; rowIndex < rowCount; rowIndex++) {
                exportRow(table, document, rowIndex);
            }

            //restore
            table.setFirst(first);
            table.setRowIndex(-1);
            table.clearLazyCache();
            table.loadLazyData();
        } 
```

The problem is that sometimes a lazy export can take long time to execute (why would we use lazy tables if not?) and, while the file is being generated, the user could change the filter or change the page, which would change the same `DataTable` object the export is using, thus generating an incorrect file.

Right now, my workaround is to create an identical hidden DataTable and linking the exporter to that one (since the user cannot manipulate it).

I'm guessing a good solution would be to clone the `LazyDataModel` object and use this clone only for that particular export. I guess the main challenge should be to evaluate the EL expressions over items of the clone instead of over the original one (maybe cloning also the `DataTable` object?).

I'm using PF 5.1 but checking [https://github.com/primefaces/primefaces/blob/master/src/main/java/org/primefaces/component/export/Exporter.java](https://github.com/primefaces/primefaces/blob/master/src/main/java/org/primefaces/component/export/Exporter.java), the bug seems to be also present on the latest version.

Thank you very much.

### LINKED ISSUE #1151: DataExporter calls load method twice on lazy DataTables
I'm using a DataTable with `lazy="true"` to paginate a big datasource. It works perfect, but, if you include a `<p:dataExporter>` for that table, the query is executed twice while exporting the data (without pagination).

You can test it with the lastest version of the Showcase, just include a dataExporter on \src\main\webapp\ui\data\datatable\lazy.xhtml

``` xml
<p:dataTable var="car" value="#{dtLazyView.lazyModel}" paginator="true" rows="10" id="carTable" lazy="true">
    </p:column>
    <p:column headerText="Brand" sortBy="#{car.brand}" filterBy="#{car.brand}">
    </p:column>
</p:dataTable>

<h:commandLink value="Export">
    <p:dataExporter type="xls" target="carTable" fileName="test" />
</h:commandLink>
```

In this case, `LazyCarDataModel.load` is called twice (iterating 200 cars each time).

Related to issue #763 
## 

Primefaces 5.3

christophs78:
Showcase-PR: https://github.com/primefaces/showcase-facelift/pull/111

Please have a look whether i copy all relevant attributs to the DataTable-Clone. Think so, but it would help if someone else checks this too.

This does not fix https://github.com/primefaces/primefaces/issues/763. This would require some further work.

### DIFF
--- a/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
+++ b/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
@@ -234,7 +234,7 @@ public abstract class DataTableExporter implements Exporter<DataTable> {
             tmpTable.setFirst(0);
             tmpTable.setRows(rowCount);
 
-            //it would be ideal to also clone LazyDataModel, but that´s more complicated
+            //it would be ideal to also clone LazyDataModel, but that´s more complicated - see #763
 
             tmpTable.loadLazyData();
 
--- a/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
+++ b/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
@@ -42,10 +42,7 @@ import javax.faces.context.FacesContext;
 import javax.faces.convert.Converter;
 import java.io.IOException;
 import java.lang.reflect.Array;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.stream.Collectors;
 
 public abstract class DataTableExporter implements Exporter<DataTable> {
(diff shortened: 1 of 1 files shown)


tandraschko:
I think we should really try to use the LazyDataModel only; what we can't do this?

Rapster:
Why don't we just save wrappeddata from lazydatamodel on the side, and set it back after export?

christophs78:
I'll provide a "Plan B" - PR for this as suggested by @rapster

PR #5489, opened 2019-12-30
Fix #1151: DataExporter calls load method twice on lazy DataTables; "Plan B"

### LINKED ISSUE #763: Exporting from Data Table with lazy data model generates incorrect file if table is manipulated while exporting

### LINKED ISSUE #1151: DataExporter calls load method twice on lazy DataTables

christophs78:
Showcase-PR: primefaces/showcase-facelift#111

This does not fix #763. This would require some further work. (IMO we can´t just clone LazyDataModel because we don´t know how it was instantiated. The specific implementation of LazyDataModel may contain custom member-variables, ....)

### DIFF
--- a/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
+++ b/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
@@ -28,6 +28,7 @@ import org.primefaces.component.datatable.DataTable;
 import org.primefaces.component.export.ExportConfiguration;
 import org.primefaces.component.export.Exporter;
 import org.primefaces.component.overlaypanel.OverlayPanel;
+import org.primefaces.model.LazyDataModel;
 import org.primefaces.util.ComponentUtils;
 
 import javax.el.MethodExpression;
@@ -212,6 +213,16 @@ public abstract class DataTableExporter implements Exporter<DataTable> {
         boolean lazy = table.isLazy();
 
         if (lazy) {
+            LazyDataModel<?> lazyDataModel = null;
+            List<?> wrappedData = null;
+            int pageSize = -1;
+
+            if (table.getValue() instanceof LazyDataModel) {
+                lazyDataModel = (LazyDataModel<?>) table.getValue();
+                wrappedData = lazyDataModel.getWrappedData();
+                pageSize = lazyDataModel.getPageSize();
+            }
+
             if (rowCount > 0) {
                 table.setFirst(0);
                 table.setRows(rowCount);
@@ -228,7 +239,13 @@ public abstract class DataTableExporter implements Exporter<DataTable> {
             table.setRows(rows);
             table.setRowIndex(-1);
             table.clearLazyCache();
-            table.loadLazyData();
+
+            if (table.getValue() instanceof LazyDataModel) {
+                lazyDataModel.setWrappedData(wrappedData);
+                lazyDataModel.setPageSize(pageSize);
+                lazyDataModel.setRowIndex(-1);
+                lazyDataModel.setRowCount(rowCount);
+            }
         }
         else {
             for (int rowIndex = 0; rowIndex < rowCount; rowIndex++) {
--- a/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
+++ b/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
@@ -213,15 +213,8 @@ public abstract class DataTableExporter implements Exporter<DataTable> {
         boolean lazy = table.isLazy();
 
         if (lazy) {
-            LazyDataModel<?> lazyDataModel = null;
-            List<?> wrappedData = null;
-            int pageSize = -1;
-
-            if (table.getValue() instanceof LazyDataModel) {
-                lazyDataModel = (LazyDataModel<?>) table.getValue();
-                wrappedData = lazyDataModel.getWrappedData();
-                pageSize = lazyDataModel.getPageSize();
-            }
+            LazyDataModel<?> lazyDataModel = (LazyDataModel<?>) table.getValue();
+            List<?> wrappedData = lazyDataModel.getWrappedData();
 
             if (rowCount > 0) {
                 table.setFirst(0);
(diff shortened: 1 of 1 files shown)


src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
```diff
@@ -228,7 +239,13 @@ protected void exportAll(FacesContext context, DataTable table, Object document)
             table.setRows(rows);
             table.setRowIndex(-1);
             table.clearLazyCache();
-            table.loadLazyData();
+
+            if (table.getValue() instanceof LazyDataModel) {
+                lazyDataModel.setWrappedData(wrappedData);
+                lazyDataModel.setPageSize(pageSize);
```
christophs78: Not sure whether  it´s necessary to reset PageSize, RowIndex and RowCount.

christophs78: Not sure whether it´s necessary to reset PageSize, RowIndex and RowCount.

Rapster:
Here is what I did:
```java
        if (lazy) {
            LazyDataModel model = (LazyDataModel) table.getValue();
            List<Object> wrappedData = model.getWrappedData();
            if (rowCount > 0) {
                table.setFirst(0);
                table.setRows(rowCount);
                table.clearLazyCache();
                table.loadLazyData();
            }

            for (int rowIndex = 0; rowIndex < rowCount; rowIndex++) {
                exportRow(table, document, rowIndex);
            }

            //restore
            table.setFirst(first);
            table.setRows(rows);
            table.setRowIndex(-1);
            table.clearLazyCache();
            model.setPageSize(rows);
            model.setWrappedData(wrappedData);
        }
```

No need to check if it's a `LazyDataModel`, because if lazy = true then value must be `LazyDataModel`. That's why I think `DataTable#lazy` is useless and redundant.

Anyway, looks like it's "working". Whatever `DataTable#loadLazyData` is setting, we should do it so I think wr should re-set pagesize, first, etc. 

If it's looks good to you, we should prefer this solution as it's less boilerplate code

christophs78:
It looks good to me!

I tried to replicate what other Code-Fragments do to have consistent code. (eg `DataTable#loadLazyData`, `DataTable#clearLazyCache`, `DataTable#loadLazyScrollData`, ...)
IMO this methods don´t need the instanceof-check too.

I also think DataTable#lazy is useless and redundant. Should we open a issue for 9.0 to deprecate/remove the lazy-attribute (from UIData)?

Rapster:
I'll create a ticket regarding DataTable, this component needs some love 😄 It'll contain few ideas that came on top of my head to clarify and simplify some features. And one of them is of course, removing lazy attribute, but I remember Thomas mention something about not working completely, maybe we'll know what by, looking in the SVN history.

christophs78:
Maybe we should test features like selection or inline-editing against this modification. To be sure we don´t cause a regression a few (AJAX-)requests later because we modified LazyDataModel and only reset wrappedData...

Usecase:
1. User exports table-data
2. a) User selects a row from the current page (https://www.primefaces.org/showcase/ui/data/datatable/selection.xhtml)
b) User switches to a row from the current page into inline-edit-mode (https://www.primefaces.org/showcase/ui/data/datatable/edit.xhtml)

Maybe LazyDataModel need´s some love too. ;-)

Rapster:
Regarding #763 I think blockUI should be used (easiest solution IMO)

christophs78:
`DataTableExporter#exportAll`:

LazyDataModel before `DataTable#loadLazyData`
grafik

LazyDataModel after `DataTable#loadLazyData`
grafik

LazyDataModel after `foreach { DataTable#exportRow }`:
grafik

melloware (approved):

hades646:
> (IMO we can´t just clone LazyDataModel because we don´t know how it was instantiated. The specific implementation of LazyDataModel may contain custom member-variables, ....)

You could call an abstract method that must be implemented in the custom class. You clone your attributes and the rest must be cloned by the abstract method.

PR #5498, opened 2020-01-03
Fix #5497: Expose getDocument in PDFExporter

### LINKED ISSUE #5497: DataTablePDFExporter: Accessing Document Object
## 1) Environment
- PrimeFaces version: 8.0RC2
- Application server + version: Irrelevant
- Affected browsers: Irrelevant

## 2) Expected behavior

I like to extend the `DataTablePDFExporter` class to export a PDF document with landscape orientation.
For this, I must overload the `preExport` method as I need to call a different constructor for document.

I expected to see a protected `setDocument` method to set the document in `DataTablePDFExporter` from my overloaded method. Would this be possible to add this to the `DataTablePDFExporter` class?

## 3) Actual behavior

I need to overload all methods which are using the document field in DataTablePDFExporter. This is not the best solution here, since a lot of code has to be needlessly duplicated.
..

## 4) Steps to reproduce

Does not apply here.

## 5) Sample XHTML

Does not apply here.

## 6) Sample bean

Does not apply here.

melloware:
Exposes `getDocument()` to allow overriding it.
Follows Joshua Bloch's principal of using the getter instead of directly accessing the field variable.

### DIFF
--- a/src/main/java/org/primefaces/component/datatable/export/DataTablePDFExporter.java
+++ b/src/main/java/org/primefaces/component/datatable/export/DataTablePDFExporter.java
@@ -56,9 +56,16 @@ public class DataTablePDFExporter extends DataTableExporter {
     private Document document;
     private ByteArrayOutputStream baos;
 
+    protected Document getDocument() {
+        if (document == null) {
+            document = new Document();
+        }
+        return document;
+    }
+
     @Override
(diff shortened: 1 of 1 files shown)


tandraschko:
+1 if the user is happy with it

PR #5505, opened 2020-01-09
Fix #5504, Formatting of decimal values for AutoNumeric in InputNumber

### LINKED ISSUE #5504: InputNumber: Regression in handling large numbers for value/minValue/maxValue after AutoNumeric update
## 1) Environment
- PrimeFaces version: 7.0.11 (Elite)
- Affected browsers: Firefox tested

## 2) Expected behavior
The formatting/encoding of the decimal numbers in the `value`/`minValue`/`maxValue` is problematic for for large values (which are usually rendered in scientific notation for double) or for values which a large precision (BigDecimal). Also the check against the default min/max value in is not preformed before rendering, which results in an error.

Consider the following examples:

1. `maxValue` is a bean property with the double value `1_000_000_000_000_000_000.0`
2. `minValue` is a bean property with the double value `-1_000_000_000_000_000_000.0`
3. `minValue` and `maxValue` are not defined, and the value is `90_000_000_000_000.0` (above AutoNumeric default max value `10_000_000_000_000.0`)
4. `value` property is a bean property of type BigDecimal with value `new BigDecimal("123456789012345678901234567890.1234567890")` and `maxValue` of `new BigDecimal("9786543120978654312097865431209786543120.9786543120")`

Expected behaviour: The component is rendered without any JavaScript Errors and the original values and max/min values are considered as given. Also, the value can be changed according to the given constraints and the changed values will be decoded correctly in the component and written to the bound value expression as correctly.

## 3) Actual behavior

The actual behavior for the 4 examples above:
1. The `maxValue` is encoded as String `"1.0E18"` and not correctly reformatted by the InputNumberComponent before passing to the AutoNumeric JavaScript library. This results in a JavaScript exception `"The maximum possible value option 'maximumValue' is invalid ; it should be a string that represents a positive or negative number, [1.0E18] given."` and the Component is broken and doens't handle any change events.
2. same as 1, but with min instead of max
3. The value is formatted as `9.0E13` in the input field. Also, the component fails to perform a range check for the value against the default range of AutoNumeric, and so a JavaScript exception is shown: `"The value [90000000000000] being set falls outside of the minimumValue [-10000000000000] and maximumValue [10000000000000] range set for this element"`. As a result the component is broken and doesn't handle an change events.
4. The BigDecimal number is cut after 20 integer places, so only `12.345.678.901.234.567.890,12` is shown in the input field. When the value is edited, the long number seems to be always rounded as double, although AutoNumeric and `BigDecimal` both support a lot more places. When the value is edited into a large enough numer, the resulting value will be `0.00`, because the precision of the number is reduced to the precision of a double, and then the high digits will be removed by the "maximum 20 integer digits" formatting, leaving only 0 digits in the lower places.

## 4) Steps to reproduce
See example XHTML/Bean 

## 5) Sample XHTML
```html
<h:body>
    <h1>NumberInput Example</h1>
    <h:form>
        <p:growl id="grwl" />
        <h:panelGrid columns="2" cellpadding="5">

            <p:outputLabel for="largeMax" value="Large MaxValue:" />
            <p:inputNumber id="largeMax" value="#{numberExample.numberWithLargeMax}" maxValue="#{numberExample.largeMax}" inputStyle="width: 500px;">
                <p:ajax update="@form" process="@this" />
            </p:inputNumber>

            <p:outputLabel for="largeMin" value="Large MinValue:" />
            <p:inputNumber id="largeMin" value="#{numberExample.numberWithLargeMin}" minValue="#{numberExample.largeMin}" inputStyle="width: 500px;">
                <p:ajax update="@form" process="@this" />
            </p:inputNumber>

            <p:outputLabel for="aboveDefaultMax" value="Above Default Max:" />
            <p:inputNumber id="aboveDefaultMax" value="#{numberExample.numberAboveDefaultMax}" inputStyle="width: 500px;">
                <p:ajax update="@form" process="@this" />
            </p:inputNumber>

            <p:outputLabel for="bigDecimal" value="Big Decimal:" />
            <p:inputNumber id="bigDecimal" value="#{numberExample.numberBigDecimal}" maxValue="#{numberExample.maxBigDecimal}" decimalPlaces="5" inputStyle="width: 500px;">
                <p:ajax update="@form" process="@this" />
            </p:inputNumber>
        </h:panelGrid>
    </h:form>
</h:body>
```

## 6) Sample bean
```java
@ViewScoped
@Named("numberExample")
public class NumberExample implements Serializable {

    private static final double max_large = 1_000_000_000_000_000_000.0;

    private static final double min_large = -1_000_000_000_000_000_000.0;

    private double numberWithLargeMin = 123_456_789;

    private double numberWithLargeMax = 123_456_789;

    private double numberAboveDefaultMax = 90_000_000_000_000.0;

    private BigDecimal numberBigDecimal = new BigDecimal("123456789012345678901234567890.1234567890");

    private static final BigDecimal maxBigDecimal = new BigDecimal("9786543120978654312097865431209786543120.9786543120");

    public double getLargeMax() {
        return max_large;
    }

    public double getLargeMin() {
        return min_large;
    }

    public double getNumberWithLargeMin() {
        return numberWithLargeMin;
    }

    public void setNumberWithLargeMin(double numberWithLargeMin) {
        this.numberWithLargeMin = numberWithLargeMin;
        FacesContext.getCurrentInstance().addMessage(null, new FacesMessage(FacesMessage.SEVERITY_INFO,
                "Number Changed to " + numberWithLargeMin, "Number is now " + numberWithLargeMin));
    }

    public double getNumberWithLargeMax() {
        return numberWithLargeMax;
    }

    public void setNumberWithLargeMax(double numberWithLargeMax) {
        this.numberWithLargeMax = numberWithLargeMax;
        FacesContext.getCurrentInstance().addMessage(null, new FacesMessage(FacesMessage.SEVERITY_INFO,
                "Number Changed to " + numberWithLargeMax, "Number is now " + numberWithLargeMax));
    }

    public double getNumberAboveDefaultMax() {
        return numberAboveDefaultMax;
    }

    public void setNumberAboveDefaultMax(double numberAboveDefaultMax) {
        this.numberAboveDefaultMax = numberAboveDefaultMax;
        FacesContext.getCurrentInstance().addMessage(null, new FacesMessage(FacesMessage.SEVERITY_INFO,
                "Number Changed to " + numberAboveDefaultMax, "Number is now " + numberAboveDefaultMax));
    }

    public BigDecimal getMaxBigDecimal() {
        return maxBigDecimal;
    }

    public BigDecimal getNumberBigDecimal() {
        return numberBigDecimal;
    }

    public void setNumberBigDecimal(BigDecimal numberBigDecimal) {
        this.numberBigDecimal = numberBigDecimal;
        FacesContext.getCurrentInstance().addMessage(null, new FacesMessage(FacesMessage.SEVERITY_INFO,
                "BigDecimal Number Changed to " + numberBigDecimal, "Number is now " + numberBigDecimal));
    }
}
```

## 7) Problem analysis
After analyzing the class `InputNumberRenderer` and the corresponding `1-inputnumber.js` JavaScript file, I have identified multiple independent causes for the behaviour decribed above.

The main regression concerning the `minValue`/`maxValue` seems to be a result in the AutoNumeric update in #5279, because the old version if AutoNumeric didn't have such strong check for the format of the minimum and maximum number.

The problems with the precision seem to come from conversion to double in JavaScript:
```javascript
getValue: function () {
        var val = this.autonumeric.getNumericString();
        if (val && this.cfg.decimalPlaces) {
            return parseFloat(val).toFixed(this.cfg.decimalPlaces);
        }

        return val;
    }
```

Also in the renderer, some values are converted to double (`new BigDecimal(...).doubleValue()`) which reduces precision. 

Concerning the third point: there is a check against the given `minValue`/`maxValue` if they are specificed in the component. If the AutoNumeric default values for these are in place however, there is no such check that the value must not be outside this range.

And finally, in the function `formatForPlugin` a number formatter is used with fixed minimum/maximum digits, which is problematic for the use with `BigDecimal` with arbitrary scale and precision.
```
NumberFormat formatter = new DecimalFormat("#0.0#");
//autoNumeric jquery plugin max and min limits
formatter.setMinimumFractionDigits(15);
 formatter.setMaximumFractionDigits(15);
formatter.setMaximumIntegerDigits(20);
```

Contrary to the code comment, the neew AutoNumeric version seems to be fine with a lot larger precision/scale numbers.

metaxmx:
These changes should resolve the problems with the number formatting in InputNumber:

- Always use BigDecimal for internal value representation, so no digits are lost on conversion to double
- Format `minValue`/`maxValue` as plain decimal number without exponential notation
- Coerce the value inside the effective min/max range, even if the range is not explicitly given and the default minValue/maxValue of AutoNumeric are used
- render the formatted value also for the input / hidden input, so AutoNumeric never encounters a value in improper numeric format (with exponentials)
- No `parseFloat(val).toFixed(this.cfg.decimalPlaces)` in JavaScript, because autoNumeric already takes care of the decimal places and this would reduce the precision for large-precision decimal numbers
- Use `BigDecimal#toPlainString()` for decimal formatting instead of NumberFormat with fixed integer/decimal places

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js
+++ b/src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js
@@ -127,11 +127,6 @@ PrimeFaces.widget.InputNumber = PrimeFaces.widget.BaseWidget.extend({
     },
 
     getValue: function () {
-        var val = this.autonumeric.getNumericString();
-        if (val && this.cfg.decimalPlaces) {
-            return parseFloat(val).toFixed(this.cfg.decimalPlaces);
-        }
-
-        return val;
+        return this.autonumeric.getNumericString();
     }
 });
--- a/src/main/java/org/primefaces/component/inputnumber/InputNumberRenderer.java
+++ b/src/main/java/org/primefaces/component/inputnumber/InputNumberRenderer.java
@@ -25,9 +25,6 @@ package org.primefaces.component.inputnumber;
 
 import java.io.IOException;
 import java.math.BigDecimal;
-import java.math.RoundingMode;
-import java.text.DecimalFormat;
-import java.text.NumberFormat;
 
 import javax.el.ValueExpression;
 import javax.faces.FacesException;
@@ -41,6 +38,10 @@ import org.primefaces.renderkit.InputRenderer;
 import org.primefaces.util.*;
 
 public class InputNumberRenderer extends InputRenderer {
+
+    // Default values for "minValue"/"maxValue" properties of the AutoNumeric Plugin
+    private static final BigDecimal DEFAULT_MIN_VALUE = new BigDecimal("-10000000000000");
+    private static final BigDecimal DEFAULT_MAX_VALUE = new BigDecimal("10000000000000");
 
     @Override
     public Object getConvertedValue(FacesContext context, UIComponent component, Object submittedValue)
@@ -87,19 +88,9 @@ public class InputNumberRenderer extends InputRenderer {
                 }
             }
             else {
+                // Coerce submittedValue to (effective) range of [minValue, maxValue]
                 BigDecimal value = new BigDecimal(submittedValue);
-                if (!LangUtils.isValueBlank(inputNumber.getMinValue())) {
-                    BigDecimal min = new BigDecimal(inputNumber.getMinValue());
-                    if (value.compareTo(min) < 0) {
-                        submittedValue = String.valueOf(min.doubleValue());
-                    }
-                }
-                if (!LangUtils.isValueBlank(inputNumber.getMaxValue())) {
-                    BigDecimal max = new BigDecimal(inputNumber.getMaxValue());
-                    if (value.compareTo(max) > 0) {
-                        submittedValue = String.valueOf(max.doubleValue());
-                    }
-                }
+                submittedValue = coerceValueInRange(value, inputNumber).toString();
             }
         }
         catch (NumberFormatException ex) {
@@ -115,8 +106,20 @@ public class InputNumberRenderer extends InputRenderer {
 
         Object value = inputNumber.getValue();
         String valueToRender = ComponentUtils.getValueToRender(context, inputNumber, value);
-        if (valueToRender == null) {
+        if (isValueBlank(valueToRender)) {
             valueToRender = "";
+        }
+        else {
+            // Rendered value must always be inside the effective interval [minValue, maxValue],
+            // or else AutoNumeric will throw an error and the component will be broken
+            BigDecimal decimalToRender;
+            try {
+                decimalToRender = new BigDecimal(valueToRender);
+            }
+            catch (Exception e) {
+                throw new IllegalArgumentException("Error converting  [" + valueToRender + "] to a decimal value;", e);
+            }
+            valueToRender = formatForPlugin(coerceValueInRange(decimalToRender, inputNumber));
         }
 
         encodeMarkup(context, inputNumber, value, valueToRender);
@@ -222,14 +225,14 @@ public class InputNumberRenderer extends InputRenderer {
         WidgetBuilder wb = getWidgetBuilder(context);
         wb.init(InputNumber.class.getSimpleName(), inputNumber.resolveWidgetVar(context), inputNumber.getClientId());
         wb.attr("disabled", inputNumber.isDisabled())
-            .attr("valueToRender", formatForPlugin(valueToRender, inputNumber, value))
+            .attr("valueToRender", valueToRender)
             .attr("decimalCharacter", inputNumber.getDecimalSeparator(), ".")
             .attr("decimalCharacterAlternative", inputNumber.getDecimalSeparatorAlternative(), null)
             .attr("digitGroupSeparator", digitGroupSeparator, ",")
             .attr("currencySymbol", inputNumber.getSymbol())
             .attr("currencySymbolPlacement", inputNumber.getSymbolPosition(), "p")
-            .attr("minimumValue", inputNumber.getMinValue())
-            .attr("maximumValue", inputNumber.getMaxValue())
+            .attr("minimumValue", formatForPlugin(inputNumber.getMinValue()))
+            .attr("maximumValue", formatForPlugin(inputNumber.getMaxValue()))
             .attr("decimalPlaces", inputNumber.getDecimalPlaces())
             .attr("emptyInputBehavior", emptyValue, "focus")
             .attr("leadingZero", inputNumber.getLeadingZero(), "deny")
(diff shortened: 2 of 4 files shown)


src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js
```diff
@@ -127,11 +127,6 @@ PrimeFaces.widget.InputNumber = PrimeFaces.widget.BaseWidget.extend({
...
 
     getValue: function () {
-        var val = this.autonumeric.getNumericString();
-        if (val && this.cfg.decimalPlaces) {
-            return parseFloat(val).toFixed(this.cfg.decimalPlaces);
-        }
-
-        return val;
+        return this.autonumeric.getNumericString();
     }
```
melloware, line 131: @mertsincan You just fixed this to do this please review.

https://github.com/primefaces/primefaces/issues/5402

melloware (approved):

tandraschko:
@mertsincan can you please explain what you changed exactly and can review this PR, too?

src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js
mertsincan: Ok ! I'll check it. Thanks a lot!

mertsincan:
> No parseFloat(val).toFixed(this.cfg.decimalPlaces) in JavaScript, because autoNumeric already takes care of the decimal places and this would reduce the precision for large-precision decimal numbers

- Yes, but Users with the version prior to version 8.0 receive a float value on bean when using the decimalPlaces feature. Exp;
the entered value: 1
decimalPlaces="2"

before 8.0, output: 1.00
on 8.0, output: 1

I've added this for backward compatibility. @metaxmx Does your PR work when you use parseFloat(val).toFixed(this.cfg.decimalPlaces)? Could you please test it? I'll merge this PR according to your feedback.

Thanks a lot!

metaxmx:
Hi @mertsincan, thank you for your feedback.

First I wanted to mention, that the change in the JavaScript file is independent from all the other changes in the Java code, as they handle different problems of the current NumberInput widget. So even if you decide not to include the change in the JavaScript concerning the `parseFloat(val).toFixed(this.cfg.decimalPlaces)`, I would recommend merging all the other changes in this PR.

From my perspective the call of `parseFloat(val)` itself is a problem, when the stored number has higher precision, than a float can handle. 

E.g. if you have an AutoNumeric with the formatted value `"123.456.789.012.345.678.901.234.567.890,12346"`, the number will be butchered by `parseFloat()`:

```javascript
this.autonumeric.getFormatted(); // returns: "123.456.789.012.345.678.901.234.567.890,12346"
this.autonumeric.getNumericString(); // returns: "123456789012345678901234567890.12346"
this.autonumeric.getNumber(); // returns: 1.2345678901234568e+29
parseFloat(this.autonumeric.getNumericString()).toFixed(2); // returns: "1.2345678901234568e+29"
```
If you want t assign the value to a BigDecimal in Java, only the second value (`this.autonumeric.getNumericString()`) holds the full number.

However,I agree with your point: when you don't have as many digits as defined in "decimalPlaces", the returned number will not contain padded zeroes:

```javascript
this.autonumeric.getFormatted(); // returns: "3,30"
this.autonumeric.getNumericString(); // returns: "3.3"
this.autonumeric.getNumber(); // returns: 3.3
parseFloat(this.autonumeric.getNumericString()).toFixed(2); // returns: "3.30"
```

So the question that remains for me is: what do you do with the returned value? If it is stored in a Float/Double/BigDecimal, then the value returned by `this.autonumeric.getNumericString()` is fine, because it is parsed again in the backend, so the number of digits will be irrelevant.
However, if the value is stored in a String, then the behaviour will change, as the value `"3.3"` is returned, and not `"3.30"`. 

I propose one of three solutions:
1) If the ValueExpression bound to the `value` Property is a Number(Float,Double,BigInteger,BigDecimal), then an additional property is rendered to the JavaScript widget options. Then and only then the widget returns `this.autonumeric.getNumericString()`, and the original `parseFloat(val).toFixed(this.cfg.decimalPlaces)` in all other cases.
2) Instead of calling `parseFloat(...).toFixed()`, the alorithm to pad the decimal places with zeroes is implemented on String-level for the `this.autonumeric.getNumericString()`.
3) We introduce a new property to the JSF widget, which can be used to switch between the "old" (backwards-compatible to PF 7) and the "new" (return full-length decimal String, without padding to `decimalPlaces`) behaviour.

I hope, I could make the problem with the `parseFloat()` solution clear. If you agree to one of my proposed solutions, I would be happy to supply a commit for it.
Or maybe you will find a different solution, which will fix both situations...

melloware:
I think I like option 2.

Get the `this.autonumeric.getNumericString()` and then use string magic to determine if decimal padding is required.

mertsincan:
> So the question that remains for me is: what do you do with the returned value? If it is stored in a Float/Double/BigDecimal, then the value returned by this.autonumeric.getNumericString() is fine, because it is parsed again in the backend, so the number of digits will be irrelevant.
However, if the value is stored in a String, then the behaviour will change, as the value "3.3" is returned, and not "3.30".
- You are completely right. I agree with you, too. But some users want to use the value with decimal places according to the needs in their app. I cannot make a definitive comment as I do not know exactly how their app work. 

Option 2 seems appropriate. Can you add a code block about this solution into your next comment? We can talk about it all together.

Thanks again ;)

melloware:
@metaxmx How about something like this? (Not tested)
```javascript
getValue: function () {
        var val = this.autonumeric.getNumericString();

        // force decimal if decimalPlaces > 0
        if (this.cfg.decimalPlaces && val.indexOf('.') === -1) {
            val = val + '.';
            for (i = 0; i < this.cfg.decimalPlaces; i++) {
                val += '0';
            }
        }

        return val;
}
```

Only if decimal places is requested and NO decimal place found in the string append 0's.

melloware:
Although I guess mine only works if there is no decimal like "1" becomes "1.00" but not if Autonumeric reutnrs "1.1" it would not make it "1.10" which I think is the intent it always has the right amount of decimal places.

tandraschko:
to be honest: i don't understand why we should `parseFloat` in `getValue`.

`getValue` is mainly used to copy the value from the input to the hidden input and in this case `getNumericString` is perfectly fine and won't change anything on the server side. right?

So my question is @mertsincan: Why and where do they need a float? When using the JS widget API for custom JS? If yes, i think we should introduce a new method `getValueAsFloat` which can be used in JS by the customer.

metaxmx:
Yes @tandraschko I believe when the value is bound to a double/float/BigDecimal in the backend, then the number of printed digits is irrelevant, because it gets converted anyways.
Only when the value is bound to a String or with a custom converter, then the number of digits could possibly make a difference (and are in fact rendered without respect to the `decimalPlaces`).
But on the other hand, neither the `currencySymbol` nor the `decimalCharacter` etc. is reflected in the returned number in the backend, so I'm not sure why especially the `decimalPlaces` property is used in the returned number, but all other formatting properties only apply to the input widget. I suppose, it should do that only for backwards compatibility, right?

Also, for the new AutoNumeric library version the `decimalPlaces` property has a default value of 2 and needs to be a positive integer. In the Taglib it is still declared as String and the default value is not given.

@melloware I think we need to pad the zeroes, even if there is already a decimal point. Like you said in your seconds post.
Probably something like this? I tested it locally with some values and decimalPlaces settings.

```javascript
    getValue: function () {
        var val = this.autonumeric.getNumericString();
        if (val && this.cfg.decimalPlaces) {
            var decimalPlacesToPad;
            if (val.indexOf('.') === -1) {
                decimalPlacesToPad = this.cfg.decimalPlaces;
                val += '.';
            } else {
                var decimalPlacesAlreadyRendered = val.length - val.indexOf('.') - 1;
                decimalPlacesToPad = this.cfg.decimalPlaces - decimalPlacesAlreadyRendered;
            }
            while (decimalPlacesToPad-- > 0) {
                val += '0';
            }
        }
        return val;
    }
```

melloware:
@metaxmx I am OK with this change and can you also update the decimalPlaces taglib doc you pointed out as incorrect?  I already updated the Wiki Migration guide warning people of the decimalPlaces=2 default value.

tandraschko:
IMO we just should just merge your PR and introduce a additional `getValueAsFloat` if required by the user. We need to wait for Mert here.
I dont see any reason to do such complicated stuff in the `getValue` method.

metaxmx:
@melloware I added the new default value to the taglib documentation, as you requested

tandraschko:
Thanks, i will merge it for now

mertsincan:
Hi again,

This issue was reported again by a PRO User.
> If we input 19.950,000 in PF7 in the backing bean setter (server side) we receive a value (BigDecimal) with scale 3,
instead in PF8 we receive a value with scale 0.
> We need to have always receive a number with the same scale (in this case 3) also when decimal places are zeroes.

Exp;
on 7.x;
client: 19.950,000 server: 19950.000
on 8.x+
client: 19.950,000 server: 19950

WDYT?

melloware:
@mertsincan Did the customer provide a reproducer. I can definitely investigate this issue if I have an example?

mertsincan:
@melloware can try it in our showcase(BigDecimal example);
https://www.primefaces.org/showcase/ui/input/inputNumber.xhtml (with decimalPlaces="3")

melloware:
Perfect I will get back to you...

mertsincan:
@metaxmx 's suggestion looks good; https://github.com/primefaces/primefaces/pull/5505#issuecomment-578787754
Autonumeric API always removes decimal characters that are all zeros. Please see;
```
// Autonumeric API 
function trimPaddedZerosFromDecimalPlaces(numericString) {
      numericString = String(numericString);

      if (numericString === '') {
        return '';
      }

      var _numericString$split = numericString.split('.'),
          _numericString$split2 = _slicedToArray(_numericString$split, 2),
          integerPart = _numericString$split2[0],
          decimalPart = _numericString$split2[1];

      if (this.isUndefinedOrNullOrEmpty(decimalPart)) {
        return integerPart;
      }

      var trimmedDecimalPart = decimalPart.replace(/0+$/g, ''); // Please see this line
      var result;

      if (trimmedDecimalPart === '') {
        result = integerPart;
      } else {
        result = "".concat(integerPart, ".").concat(trimmedDecimalPart);
      }

      return result;
    }
```

melloware:
@mertsincan I created a new issue https://github.com/primefaces/primefaces/issues/6387

And PR: https://github.com/primefaces/primefaces/pull/6386

mertsincan:
Thanks @melloware ;)

PR #5512, opened 2020-01-10
Fix #5511: Duplicate mousewheel plugins

### LINKED ISSUE #5511: Mousewheel: Duplicate Plugins
It looks like we have duplicate mousewheel plugins.

1. Old plugin: https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/jquery/jquery.mousewheel.js

This plugin is really old and no longer maintained.

2. Offical Jquery Plugin: https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/mousewheel/jquery.mousewheel.min.js

Which is from here: https://github.com/jquery/jquery-mousewheel

It looks like #1 is being used and #2 is not actually being used.

@tandraschko I propose replacing #1 with #2 and removing the location of #2 entirely.  Thoughts?

melloware:
I updated the plugin in jquery plugins, and removed the duplicate in /mousewheel which looks like it was not referenced anywhere in the code that I could tell.


melloware:
@tandraschko unless you object i would like to merge this.

tandraschko:
if really not used (like pom.xml), +1

melloware:
confirmed the only one in the pom.xml is this one `/jquery/jquery.mousewheel.js` the other one was being compressed and put in the JAR but never used.  It was always using this legacy verison.

PR #5515, opened 2020-01-13
Fix #5514: Datatable unbind filter events

### LINKED ISSUE #5514: DataTable: Unregister filter events on refresh()
Similar to #3139

If you call `datatable.refresh(cfg)` it is not unbinding the old filter events thus when you filter the datatable it is calling multiple repeated filter events.   If you call refresh() 5 times and then filter you will see 5 filter AJAX events fire.  We do this for other events on the datatable.

PR coming.

melloware:


tandraschko:
+1

PR #5521, opened 2020-01-14
Fix #5513: DatePicker: Touch UI does not open correctly

### LINKED ISSUE #5513: DatePicker: Touch UI does not open correctly
## 1) Environment
- PrimeFaces version: PrimeFaces-8.0-SNAPSHOT
- Application server + version: Tomcat 9 + JDK 13
- Affected browsers: Opera 65.0.3467.69 / Chrome 79.0.3945.79

## 2) Expected behavior

The example can be reproduced by the showcase itself.

## 3) Actual behavior

As you can see, it seems to open and close !!! When you change the browser tab and return to the previous one the DatePicker is displayed !!

## 4) Steps to reproduce

[https://www.primefaces.org/showcase/ui/input/datepicker/datePicker.xhtml](url)

## 5) Sample XHTML

Not necessary

## 6) Sample bean

Not necessary

christophs78:


melloware:
Something like...
```javascript
var delayMs = 0;
if (this.options.touchUI) {
       //Bind click-listener for TouchUI delayed - see #5513
       delayMs = 100;
 }
         
setTimeout(function () { this.bindDocumentClickListener(); },  delayMs );
```

christophs78:
Now it's simplified. :-)

(There is one more (old) issue with DatePicker + Touch UI + IE11: When you select a date the overlay closes but the mask is still there. But i won´t analyse this for now.)

melloware (approved):

Issue #4931, opened 2019-06-24
DataGrid and DataTable: rows- and first-attributes ignored after use of paginator


matinh:
## 1) Environment
- PrimeFaces version: 7.0/7.1-SNAPSHOT
- Does it work on the newest released PrimeFaces version? Version? No, neither with 7.0, nor with 7.1-SNAPSHOT.
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source). No.
- Application server + version: Tested with Wildfly 11 and Jetty 9.4.15 from the primefaces-test project.
- Affected browsers: probably all (tested with Firefox 60.7.0esr, Opera 58 and Chromium 75). Note: Opera and Chromium seem to have other problems with 7.1-SNAPSHOT, which is probably unrelated)

## 2) Expected behavior

The `rows`-attribute from the `<p:dataGrid>` component should always use the actual value.

## 3) Actual behavior

After the use of the pager, the `rows`-attribute seems to be ignored and the last value before the pager was used is always used, no matter what the actual value of the rows EL is.

## 4) Steps to reproduce

* clone https://github.com/matinh/primefaces-test.git using the branch _dataGrid_ with `git clone 'https://github.com/matinh/primefaces-test.git' -b dataGrid`
* change directory `cd primefaces-test`
* run `mvn jetty:run-exploded`
* goto http://localhost:8080/primefaces-test/
* use the spinner to change number of rows in the dataGrid
* use pager to switch to another page
* try to use spinner again to change number of rows, which is no longer working.

## Further comments

* There are no JavaScript Errors in the browsers console.
* The setter of the columns-attribute is called with the correct value.

Rawi01:
I checked some primefaces sources and it seems like the problem is in L297 https://github.com/primefaces/primefaces/blob/e2062dc5671b10e9f3913b454d4c31b86a62c226/src/main/java/org/primefaces/component/api/UIData.java#L290-L307

A possible fix might be to check for an exisitng value expression as in L305 and only set that value to the new one e.g.

```java
ValueExpression rowsVe = data.getValueExpression("rows"); 

if (rowsVe != null && !rowsVe.isReadOnly(elContext)) { 
    rowsVe.setValue(context.getELContext(), Integer.valueOf(rowsParam)); 
} else {
    data.setRows(Integer.valueOf(rowsParam));
}
```

matinh:
[...]

Issue #4391, opened 2019-01-10
schedule: XSS in tooltips


cnsgithub:
## 1) Environment
- PrimeFaces version: 6.3-snapshot

## 2) Expected behavior
Proper escaping of `ScheduleEvent.description `when `p:schedule tooltip="true"` is used.

## 3) Actual behavior
The event descriptions are passed to jQuery's `html` function without being escaped which enables XSS.

See [1-schedule.js](https://github.com/primefaces/primefaces/blob/818f21d4fb8cfe5fc1bf94f222f67a5ec5778136/src/main/resources/META-INF/resources/primefaces/schedule/1-schedule.js#L137).

## 4) Steps to reproduce
1. Modify `ScheduleView.java `of showcase accordingly:
```java
	    DefaultScheduleEvent evil = new DefaultScheduleEvent("Champions League Match", previousDay8Pm(), previousDay11Pm());
	    evil.setDescription("<script>alert('xss');</script>");
	    eventModel.addEvent(evil);
```
2. Modify `schedule.xhtml` of showcase accordingly:
`                <p:schedule id="schedule" value="#{scheduleView.eventModel}" widgetVar="myschedule" timeZone="GMT+2" tooltip="true">`
3. Open the [showcase](http://localhost:8080/showcase/ui/data/schedule.xhtml).
4. Hover the _Champions League Match_ event of the first schedule.

![image](https://user-images.githubusercontent.com/13161711/50967072-59492200-14d7-11e9-9f2b-2bee61331a56.png)

PR #5524, opened 2020-01-15
Fix #5523: RowsPerPage always set data values

### LINKED ISSUE #5523: DataTable: rowsPerPage from paginator broken after using a filter
## 1) Environment
- PrimeFaces version: 8.0-SNAPSHOT (commit 6b6c2be from 2020-01-15)
- Does it work on the newest released PrimeFaces version? Version? It works on 7.0, can't test 7.0.x.
- Does it work on the newest sources in GitHub? No, it's broken.
- Application server + version: Jetty 9.4.15 from primefaces-test
- Affected browsers: Tested with Firefox 67 and 71 but probably independent from browser

## 2) Expected behavior
DataTable has a paginator which supports a _{RowsPerPageDropdown}_ to select number of rows displayed per page. This dropdown should work even after filtering for a column was used.

## 3) Actual behavior
Once filtering was used, the rowsPerPage-dropdown stops working. The user has to change filter again, so apply to changes in the drowdown.

## 4) Steps to reproduce

* clone https://github.com/matinh/primefaces-test.git using the branch _dataTable-rowsPerPage_ with `git clone 'https://github.com/matinh/primefaces-test.git' -b dataTable-rowsPerPage`
* change directory `cd primefaces-test`
* run `mvn jetty:run-exploded`
* goto http://localhost:8080/primefaces-test/
* try changing the number of rows per page, which works
* now apply a filter (for example "o" in the second column) which triggers an AJAX request and reduces displayed rows correctly
* remove the filter and try again to change the number of rows per page, which now has no effect.
 
## Further comments
* This issue is a regression bug. It used to work with 7.0 but no longer works with recent 8.0-SNAPSHOT versions.
* This issues is probably related to issue #4931 respectively to the changes from PR #4956.

melloware:
I think we need to always set the "data" values and not optionally set them. @matinh please test

### DIFF
--- a/src/main/java/org/primefaces/component/api/UIData.java
+++ b/src/main/java/org/primefaces/component/api/UIData.java
@@ -320,21 +320,18 @@ public class UIData extends javax.faces.component.UIData {
         }
 
         ValueExpression firstVe = data.getValueExpression("first");
+        int firstValue = Integer.valueOf(firstParam);
         if (isWriteable(elContext, firstVe)) {
-            firstVe.setValue(elContext, Integer.valueOf(firstParam));
+            firstVe.setValue(elContext, firstValue);
         }
-        else {
-            data.setFirst(Integer.valueOf(firstParam));
-        }
+        data.setFirst(firstValue);
 
         ValueExpression rowsVe = data.getValueExpression("rows");
         int newRowsValue = "*".equals(rowsParam) ? getRowCount() : Integer.valueOf(rowsParam);
         if (isWriteable(elContext, rowsVe)) {
             rowsVe.setValue(elContext, newRowsValue);
         }
-        else {
-            data.setRows(newRowsValue);
-        }
+        data.setRows(newRowsValue);
     }
 
     private boolean isWriteable(ELContext elContext, ValueExpression ve) {


matinh:
@melloware thanks for the quick response and the fix. Your suggested fix solves my concrete problem. And I was already thinking about the same fix. However, two issues come to my mind:
1. A potential NPE when the argument to `Integer.valueOf()` is `null`. Not sure if this could ever happen.
2. The semantic changes again, compared to the original fix, based on https://github.com/primefaces/primefaces/issues/4931#issuecomment-505080207. And I'm afraid this could re-introduce the issue #4931.

melloware:
yes can you try my fix and see if it regresses #4391 ? I didn't test that.

matinh:
Yes, your fix breaks #4391 again, so we need to find some other solution.

melloware:
Actually even with my fix this scenario is still broken.

1. Set RowsPerPage to 1.
2. Filter String field by character "o".
3. use the Pagintor to go to Page 2.
4. Now set RowsPerPage back to 5...nothing happens. That should cause the table to repaint with 5 rows.

matinh:
I just double-checked: your fix works for me (but re-introduces #4391). Maybe you mixed something up with used primefaces version?
However, as the fix introduces another problem, it won't be usable anyway.

Issue #5493, opened 2020-01-02
DataTable: currentPageReportTemplate {endRecord} incorrect calculation


melloware:
## 1) Environment 
- PrimeFaces version: 8.0.RC2
- Does it work on the newest released PrimeFaces version? NO
- Does it work on the newest sources in GitHub? NO
- Application server + version: Jetty Showcase
- Affected browsers: ALL

## 2) Expected behavior

Using currentPageReportTemplate reports the correct values.

## 3) Actual behavior

It is reporting incorrect values for the {endRecord} value.

## 4) Steps to reproduce

I just updated the showcase with this PR: https://github.com/primefaces/showcase-facelift/pull/112

1. Run the showcase and navigate to the Paginator Example: http://localhost:8080/showcase/ui/data/datatable/paginator.xhtml

2.  Note the starting value of the current report is "1-10 of 50 records".

3. Now page to the right it will display "11-20 of 50 records" correctly.

4. Page to the left it will display "1-10 of 50 records" correctly.

5. Now change the records per page from "10" to "15" the report will incorrectly display "1-015 of 50 records" notice the "015" instead of "15".

6. Now page to the right and it incorrectly displays "16-50 of 50 records" instead of "16-30 of 50" records.

## 5) Sample XHTML

```xml
<p:dataTable var="car" value="#{dtPaginatorView.cars}" rows="10" 
           paginator="true"
           paginatorTemplate="{CurrentPageReport} {FirstPageLink} {PreviousPageLink} {PageLinks} {NextPageLink} {LastPageLink} {RowsPerPageDropdown}"
          currentPageReportTemplate="{startRecord}-{endRecord} of {totalRecords} records" 
          rowsPerPageTemplate="5,10,15">
```

## 6) Sample bean

```java
@Named("dtPaginatorView")
@ViewScoped
public class PaginatorView implements Serializable {
     
    private List<Car> cars;
     
    @Inject
    private CarService service;
 
    @PostConstruct
    public void init() {
        cars = service.createCars(50);
    }
     
    public List<Car> getCars() {
        return cars;
    }
 
    public void setService(CarService service) {
        this.service = service;
    }
}
```

PR #5526, opened 2020-01-15
fix for issue #5523

matinh:
Ok, here is a PR that tries to fix issue #5523. It does not re-introduce issue #4931!

### DIFF
--- a/src/main/java/org/primefaces/component/api/UIData.java
+++ b/src/main/java/org/primefaces/component/api/UIData.java
@@ -147,7 +147,27 @@ public class UIData extends javax.faces.component.UIData {
         if (rows < 0) {
             throw new IllegalArgumentException(String.valueOf(rows));
         }
-        getStateHelper().put(PropertyKeys.rows, rows);
+
+        ELContext elContext = getFacesContext().getELContext();
+        ValueExpression rowsVe = getValueExpression("rows");
+        if (isWriteable(elContext, rowsVe)) {
+            rowsVe.setValue(elContext, rows);
+        }
+        else {
+            super.setRows(rows);
+        }
+    }
+
+    @Override
+    public void setFirst(int first) {
+        ELContext elContext = getFacesContext().getELContext();
+        ValueExpression firstVe = getValueExpression("first");
+        if (isWriteable(elContext, firstVe)) {
+            firstVe.setValue(elContext, first);
+        }
+        else {
+            super.setFirst(first);
+        }
     }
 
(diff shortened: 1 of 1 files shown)


melloware:
Unfortunately still doesn't fix all scenarios.  Just run the showcase locally and go to this example: http://localhost:8080/showcase/ui/data/datatable/paginator.xhtml

The paginator is not working at all.

melloware:
Or just switch your use case from `rows="#{testViewCdi.rowsPerPage}"` to `rows="2"` to see the broken default use case

matinh:
Ok, I fixed this issue. Now the mentioned show-case and the two mentioned issues work.
Anyway, I can't say I'm confident that everything is working now. I just don't know PrimeFaces components to little.

melloware:
You are going great.  let me test out your latest changes

melloware:
OK you definitely fixed the main use case without using an EL expression using the hard coded "2".

However try this...

1. Set rowsperPage to 1
2. Filter by String "o" which will produce 3 records and now you have 3 pages and the row being displayed is "One".
3. Page to page 2 and you will see the row being displayed is still "One" and it should be "Two".  Same for Page 3.

matinh:
You are just too fast ;-)

I was just reviewing my changes when noticing, that using stateHelper for `first` as well was just too much. I committed some more changes. Could you please try once more.

melloware:
OK that looks good.  Now I am just down to 1 scenario not working.

1. Set rowsPerPage to 1.
2. Filter by "o" should produce 3 pages.
3. click on Page 3 you properly see item "Four".
4. Now change rowsPerPage to "3" so you can see all 3 items.  Nothing happens.

matinh:
You are right, I can reproduce this. However, I've absolutely no idea why this happens. I noticed, that the last click on "rowsPerPage" doesn't even create an AJAX request in the browser. Very strange! Maybe this is some other, unrelated bug?

melloware:
Yeah it might be a problem in the paginator.js not sending the Ajax request if it thinks someting hasn't changed.   Might want to set a browser breakpoint there.

matinh:
It seems this only happens, when you increase rows in a ways the currently active page would no longer exist and should thus be reset/decremented.

melloware:
You are right this feels like a JS fix needs to be done.

melloware:
OK the regression was caused by #5493.  You can close this PR i am submitting one that fixes all issues for #5523,  #4931 and for  #5493

matinh:
Closing this PR, as PR #5527 includes this fix.

PR #5527, opened 2020-01-15
Fix #5523, #4931, #5493: Paginator issues

### LINKED ISSUE #5523: DataTable: rowsPerPage from paginator broken after using a filter

melloware:
i have studied the issues and worked with @matinh to test all 3 tickets and made sure all regression issues are now working.

### DIFF
--- a/src/main/java/org/primefaces/component/api/UIData.java
+++ b/src/main/java/org/primefaces/component/api/UIData.java
@@ -107,6 +107,7 @@ public class UIData extends javax.faces.component.UIData {
         paginator,
         paginatorTemplate,
         rows,
+        first,
         rowsPerPageTemplate,
         rowsPerPageLabel,
         currentPageReportTemplate,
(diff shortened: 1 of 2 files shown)


matinh:
I tested and reviewed this patch and it is working for all my use cases!

However, I have one comment/question regarding `PropertyKeys.first` [(line 100)](https://github.com/primefaces/primefaces/blob/0e00db3acbf6b6aee54dbf59b59d026acfaf4879/src/main/java/org/primefaces/component/api/UIData.java#L110). IMHO this is not used at all. Is this a left-over from an old version of my patch? I think we could remove this one line from the patch.

melloware:
You are right I can remove that first.

PR #5533, opened 2020-01-16
Fix #5530: innerHTML performance improvement

### LINKED ISSUE #5530: DataXXX: html() vs innerHTML perf improvement
Forum Post: https://forum.primefaces.org/viewtopic.php?f=3&t=60866

Components like the DataTable replacing the entire section of HTML using Jquery .html(replace) method can cause browser performance issues.

Stack Overflow: https://stackoverflow.com/questions/18393981/append-vs-html-vs-innerhtml-performance

@stolp tested it using a particular datatable he was seeing issues with and the result was _"this does indeed drop the time for script evaluation from 3200ms down to 540ms which is a roughly 500% speedup!"_

PR coming that will fix only a few places that do large HTML replacements.  We are not going to touch everywhere using .html() method.

melloware:
Only in DRAFT so it can be tested and review

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datagrid/datagrid.js
+++ b/src/main/resources/META-INF/resources/primefaces/datagrid/datagrid.js
@@ -40,7 +40,7 @@ PrimeFaces.widget.DataGrid = PrimeFaces.widget.BaseWidget.extend({
                 PrimeFaces.ajax.Response.handle(responseXML, status, xhr, {
                         widget: $this,
                         handle: function(content) {
-                            this.content.html(content);
+                            this.content[0].innerHTML = content;
                         }
                     });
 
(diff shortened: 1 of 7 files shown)


tandraschko:
as i said, please test by logging if the child elements are destroyed when replacing the content

tandraschko:
AND does innerHTML asignment execute scripts?! Maybe therefore the performance difference is that big

tandraschko:
probably not: https://stackoverflow.com/questions/1197575/can-scripts-be-inserted-with-innerhtml

melloware:
I will do some more testing and try and debug if children are getting destroyed.  leaving this in draft state so it can't be merged.

tandraschko:
i think the main problem is that new widgets, which are inside the new content, are not initialized as the scripts in thew content are not executed like with jQuery append; therefore the perf diff is so big as the scripts are just ignored completely.

melloware:
ahh you are right.  So theoretically if a column contained SelectOne menus their widget init would never be called after say Paging the datatable.

tandraschko:
exactly; thats the one and only reason probably why native is faster

PR #5538, opened 2020-01-17
Fix #5537: DatePicker click input more than once not showing popup

### LINKED ISSUE #5537: DatePicker: Unable to select date.
## 1) Environment
- PrimeFaces version: PrimeFaces-8.0-SNAPSHOT
- Does it work on the newest released PrimeFaces version? Version? not tested
- Does it work on the newest sources in GitHub? not tested
- Application server + version: not relevant
- Affected browsers: All

## 2) Expected behavior

After selecting a date, you can change it! 

## 3) Actual behavior

You cannot select another date when you already have one selected.

All examples of showcase have this inssues.

## 4) Steps to reproduce

You can test using the showcase. For example, select some date from 'Disabled Date / Day:' and then click again on the 'input' field. As you can see, it is no possible to select any other day. It will only be possible to change it when we click outside the 'input' field and then click again in 'input'.

## 5) Sample XHTML

Not necessary.

## 6) Sample bean

Not necessary.

melloware:


christophs78:
This creates some new issues: (tested with chrome; Popup does not close when we click outside the popup)
grafik

grafik

christophs78:
Think we need to do the following: (declare $this - variable)
```
        showOverlay: function () {
            if (this.options.onBeforeShow) {
                this.options.onBeforeShow.call(this);
            }

            this.panel.show();
            this.alignPanel();

            if (!this.options.touchUI) {
                var $this = this;
                setTimeout(function () {
                    $this.bindDocumentClickListener();
                }, 100);
            }
        },
```

christophs78:
And maybe 10ms instead of 100ms to reduce the lag of opening the popup.

melloware:
Trying it now..

melloware:
Seems to work great even at 10ms.

christophs78:
Ready to be merged. :-)
And hope there comes a day where no further DatePicker-bugs/limitations pop up. ;-)

melloware:
HA!  I feel like there are more and more bugs lurking with this component.    But you have done a great job squashing a lot of the big ones.

PR #5543, opened 2020-01-18
Fix #3857 - DataExporter: datatable with skipChildren is not working

### LINKED ISSUE #3857: DataExporter: datatable with skipChildren is not working
## 1) Environment
- PrimeFaces version: 6.2
- Affected browsers: All

## 2) Expected behavior
Using that example with Datatable#skipChildren="true" should export content into any supported export file. As the datatable does not contain any inputs, skipChildren make sense. 

## 3) Actual behavior
Content is not exported

## 4) Steps to reproduce
https://www.primefaces.org/showcase/ui/data/dataexporter/basic.xhtml
Set Datatable#skipChildren="true"

## 5) Sample XHTML
See showcase

## 6) Suggestions
In my use case, every columns contains a bunch of infos contained in a `pe:dynaForm` so it makes even more sense to use that attribute. Though, using that flag prevents the use of `p:dataExporter`. Having buttons in paginator works though.

See this code:
```java
    protected void processPhase(FacesContext context, PhaseId phaseId) {
        if (shouldSkipChildren(context)) {
            return;
        }

        setRowIndex(-1);
        processFacets(context, phaseId);
        if (requiresColumns()) {
            processColumnFacets(context, phaseId);
        }
        processChildren(context, phaseId);
        setRowIndex(-1);
    }
```

Do you think, it'd be possible to call UIData#shouldSkipChildren after processing facets? I don't think performance will be affected postponing that call

Rapster:


tandraschko:
can you describe the change?

Rapster:
Calling shouldSkipChildren too early prevents processing components like button (contained in facet for example), meaning actions won't be called. As an example, mentionned in the original ticket, export doesn't work since dataExporter is not processed at all

melloware:
This make sense to me you are basically skipping children right before the call to processChildren but still allowing facets to be processed.

melloware (approved):

src/main/java/org/primefaces/component/api/UIData.java
```diff
@@ -387,15 +387,16 @@ public void processUpdates(FacesContext context) {
     }
 
     protected void processPhase(FacesContext context, PhaseId phaseId) {
-        if (shouldSkipChildren(context)) {
-            return;
-        }
-
         setRowIndex(-1);
```
Rapster: Maybe I should move this line right before processChildren don't you think?

PR #5545, opened 2020-01-19
#5483: DataGrid: add multiViewState

christophs78:
ShowCase-PR: https://github.com/primefaces/showcase-facelift/pull/117


christophs78:
... and we should probably create a separate Issue for tracking this change for 8.0 RC3

Issue #5263, opened 2019-10-22
DataTable: invalid values with rowReorder of editable components like SelectBooleanCheckbox


stolp:
## 1) Environment
- PrimeFaces version: 7.0.8
- Does it work on the newest released PrimeFaces version? No
- Does it work on the newest sources in GitHub? Not tested
- Application server + version: Tomcat 9
- Affected browsers: All

## 2) Expected behavior
When dragging a row I expect that the values of input components do not change.

## 3) Actual behavior
After dragging a row containing false values are stored in the server side value representation.

In my understanding the following `process="@none"` should suppress changing values in a rowReorder event:

`<p:ajax partialSubmit="true" event="rowReorder" process="@none" />`

This works well for InputText, but fails with SelectBooleanCheckbox, SelectOneMenu and probably others.

Or am I missing something obvious here?

## 4) Steps to reproduce
Drag a row in the supplied test application. Although the visual state does not change, the server side values are set to false (in case of a SelectBooleanCheckbox) or null (in case of SelectOneMenu - not covered in the test here). This may apply to other components as well.

## 5) Sample XHTML
See attached project.

## 6) Sample bean
See attached project.

[primefaces-test.zip](https://github.com/primefaces/primefaces/files/3754612/primefaces-test.zip)

melloware:
Have you also tried this with `rowStatePreserved="true"` property on the Datatable?

melloware:
Your test works for me if I remove the partialSubmit="true".
```xml
<p:ajax event="rowReorder" process="@none" />
```

melloware:
OK nevermind I see what you mean.  If you watch the network traffic even with partialSubmit="true" its always including the Datatbale in the submit even though you said @none.

```
javax.faces.partial.execute: @none j_idt6:tblData
```

melloware:
OK submitted a fix.

PR #5560, opened 2020-01-23
Fix #2551: Treetable do not always add @this to process when AJAX eve…

### LINKED ISSUE #2551: TreeTable always including @this in cellEdit ajax request regardless of process setting
## 1) Environment
- PrimeFaces version: 6.1.3
- Application server + version: GlassFish 4.1.2
- Affected browsers: Chrome

## 2) Expected behavior
TreeTable should only send components included in the process attribute to the server as part of the cellEdit ajax request.

## 3) Actual behavior
TreeTable sends all the treeTable's editable components to the server as part of the cellEdit ajax request.

## 4) Steps to reproduce

Create a treeTable and set the process attribute to the following on the ajax cellEdit tag:

process="@([data-rk='1')"

I would expect to only see the data from row 1 sent to the server but I see the data from row 1 sent and then all the data from the treeTable.   In other words, if the treeTable as 4 rows, I see all 4 rows sent to the server.

## 5) Sample XHTML

```
<p:treeTable
    id="steps"
    value="#{checklistPC.treeNode}"
    var="step"
    paginator="false"
    editable="true"
    editMode="cell"
    reflow="true">
    <p:ajax
        event="cellEdit"
        listener="#{checklistPC.onCellEdit}"
        partialSubmit="true"
        process="@([data-rk='1')" />
    <p:ajax
        event="expand"
        listener="#{checklistPC.onNodeExpand}"
        oncomplete="ELIS.EditorDialogFactory.updateInputs();"
        partialSubmit="true" />
    <p:column
        headerText="#"
        styleClass="elis-dt-label"
        style="vertical-align: top;">
        <h:outputText value="#{step.number}" />
    </p:column>
    <p:column
        headerText="Instruction"
        styleClass="elis-dt-note"
        style="vertical-align: top;">
        <h:outputText
            value="#{step.instruction}"
            escape="false" />
    </p:column>
    <p:column
        headerText="Help"
        styleClass="elis-dt-note"
        style="vertical-align: top;">
        <h:outputText
            value="#{step.help}"
            escape="false" />
    </p:column>
    <p:column
        headerText="Owner"
        styleClass="elis-dt-label"
        style="vertical-align: top;">
        <p:cellEditor id="ceo">
            <f:facet name="output">
                <h:outputText value="#{step.owner.name}" />
            </f:facet>
            <f:facet name="input">
                <p:inputText
                    id="owner"
                    value="#{step.owner}"
                    placeholder="Owner"
                    requiredMessage="Owner is required"
                    label="Owner"
                    styleClass="elis-user_selector">
                    <f:converter converterId="elisUserConverter" />
                </p:inputText>
            </f:facet>
        </p:cellEditor>
    </p:column>
    <p:column
        headerText="Status"
        styleClass="elis-dt-label"
        style="vertical-align: top;">
        <p:cellEditor id="ces">
            <f:facet name="output">
                <h:outputText value="#{step.status.label}" />
            </f:facet>
            <f:facet name="input">
                <p:selectOneMenu
                    id="step-status"
                    value="#{step.status}"
                    requiredMessage="Status is required"
                    onchange="updateCheckbox(event);">
                    <f:selectItems
                        value="#{checklistPC.stepStatuses}"
                        var="item"
                        itemLabel="#{item.label}"
                        itemValue="#{item}" />
                </p:selectOneMenu>
            </f:facet>
        </p:cellEditor>
    </p:column>
    <p:column
        headerText=""
        styleClass="elis-dt-checkbox">
        <p:cellEditor id="cecb">
            <f:facet name="output">
                <p:selectBooleanCheckbox
                    id="ocb"
                    value="#{step.status == 'COMPLETED'}" />
            </f:facet>
            <f:facet name="input">
                <p:selectBooleanCheckbox
                    id="cb"
                    value="#{step.status == 'COMPLETED'}"
                    onchange="updateStatus(event);" />
            </f:facet>
        </p:cellEditor>
    </p:column>
    <p:column
        headerText="Note"
        styleClass="elis-dt-note"
        style="vertical-align: top;">
        <p:cellEditor id="cen">
            <f:facet name="output">
                <h:outputText
                    value="#{step.note}"
                    escape="false" />
            </f:facet>
            <f:facet name="input">
                <p:inputTextarea
                    id="note"
                    value="#{step.note}"
                    styleClass="elis-note_editor"
                    label="help" />
            </f:facet>
        </p:cellEditor>
    </p:column>
</p:treeTable>
```
## 6) Sample bean

..

## 7) Recommendations

I would like to see treeTable and dataTable support the following functionality:

1. During cellEdit, only send the data that changed.   Since you can only change a single field at a time, you would only need to send the current field.  This would minimize traffic between the client and server.

2. Support cellEdit with no Ajax (See issue 2030).   It would be nice to be able to edit the values of a datatable or treetable as a group and then send the changes to the server one time as a group.  Unfortunately, when you disable cellEdit, you break the cellEditor functionality.

melloware:
…nt is used.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/treetable/treetable.js
+++ b/src/main/resources/META-INF/resources/primefaces/treetable/treetable.js
@@ -1570,6 +1570,7 @@ PrimeFaces.widget.TreeTable = PrimeFaces.widget.DeferredWidget.extend({
         };
 
         if(this.hasBehavior('cellEdit')) {
+            delete options.process;
             this.callBehavior('cellEdit', options);
         }
         else {


tandraschko:
The fix looks weird and always removes all configured process values?
And i think we need to process the treetable to make the functionality work?

If the change makes sense, i think we should add the process in the else-if block instead of removing it

melloware:
I will test some more but it looks like when call callBehavior by default it will add the value from the p:ajax tag such as `process=@none` and if nothing is passed it makes it `@this`.  What the bad behavior that is happening now is that the code is always adding `process: this.id` to the JSON and then it goes to callBehavior and it makes it basically `@none this.id` so it has both.  So even though the user has said they want @none its always appending the this.id which is why I am removing it from the JSON when calling CallBehavior else it is needed when called with just the default PrimeFaces.Request.handle() method.

melloware:
Its the same issue as #5263

tandraschko:
If this is true, we should check and fix all places and move the change to 9.0.
Also, as i said, we should add the "process=@this" in the else-if-block instead of removing it in the if-block

melloware:
Yep I am looking through the global code a little more.  Like you mentioned i think we need a global fix for this situation.

PR #5563, opened 2020-01-24
Fix #5562: Calendar/DatePicker: pattern: support symbol "u"

### LINKED ISSUE #5562: Calendar/DatePicker: pattern: support symbol "u"
## 1) Environment
- Affected browsers: all

## 2) Expected behavior
https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html allows to use "u" as symbol for year. (similar to "y")
So it should be possible to do something like 
`<p:datePicker id="german" widgetVar="german" value="#{calendarJava8View.dateDe}" locale="de" monthNavigator="true" pattern="dd.MM.uuuu"/>`

## 3) Actual behavior
grafik

## 4) Steps to reproduce
Use ui/input/datepicker/datePickerJava8.xhtml from showcase and modify pattern.

christophs78:


christophs78:
The replacement is not ideal, because SimpleDateFormat interpreted "u" as "Day number of week (1 = Monday, ..., 7 = Sunday) ".
But i don´t expect someone used it in combination with Calendar/DatePicker-pattern.
And looking forward `java.time` is the way to go.

melloware (approved):

melloware:
Would it make it more focused to do something like.

```java
        if (pattern.contains("uu")) {
            pattern = pattern.replaceAll("u", "y");
        }
```

Basically if it is a single "u" in the pattern it is most likely SimpleDateFormats Day of Week symbol but once two uu's are together its most likely a year field of uu or uuuu so its safe to replace?

stolp:
Or this:

`pattern = pattern.replaceAll("uu", "yy");`

christophs78:
Good feedback!

PR #5570, opened 2020-01-29
Fix #5569: ConfirmDialog appendTo="@(body)" default

### LINKED ISSUE #5569: ConfirmDialog: Default to appendTo="@(body)"
1) Environment
- Does it work on the newest released PrimeFaces version? Version? Yes

## 2) Expected behavior

Using a non global confirm dialog should default to appendTo="@(body)".

## 3) Actual behavior

Default has no appendTo default.  We have a scenario like this in PrimeFaces 7 it is rendering the modal on top but in PrimeFaces 8.0 testing it is rendering the dialog behind the modal.  Manually putting appendTo="@(body)" fixes it but I think for confirm dialog that should be the default.

image

image

https://github.com/primefaces/primefaces/blob/4380e49adb34c4afd3b321000607652541db9ba6/src/main/resources/META-INF/resources/primefaces/dialog/dialog.js#L636-L638

## 5) Sample XHTML

```xml
<h:form id="theForm">
           <p:commandButton value="Show" type="button" onclick="PF('confirmDialog').show();" />

            <p:confirmDialog id="dialog" widgetVar="confirmDialog" message="Destroy The World?" modal="true" severity="alert" header="Test" closeOnEscape="true">
                <p:commandButton value="Yes" type="button" styleClass="ui-confirmdialog-yes" icon="pi pi-check" />
                <p:commandButton value="No" type="button" styleClass="ui-confirmdialog-no" icon="pi pi-times" />
            </p:confirmDialog>      
</h:form>
```

I think we should make switch it to an OR.
```javascript
if (!cfg.appendTo || cfg.global) {
    cfg.appendTo = '@(body)';
}
```

melloware:
Default to appendTo="@(body)" if either global="true" or if appendTo is not defined.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/dialog/dialog.js
+++ b/src/main/resources/META-INF/resources/primefaces/dialog/dialog.js
@@ -633,7 +633,7 @@ PrimeFaces.widget.ConfirmDialog = PrimeFaces.widget.Dialog.extend({
         cfg.resizable = false;
         cfg.modal = true;
 
-        if (!cfg.appendTo && cfg.global) {
+        if (!cfg.appendTo || cfg.global) {
         	cfg.appendTo = '@(body)';
         }
 


tandraschko:
Not sure about this.
AFAIR appendTo was mainly used because of older browsers
cant this be fixed with zindex?

melloware:
It's really strange our QA team is testing our apps on 8.0 and if you look at the screenshots in the ticket on 7.0 it works fine and the modal looks correct.  On 8.0 with no other code changes its hidden behind the modal.  So I fixed it with appendTo="@(body)" which let me to for confirm dialogs shouldn't they almost always default to BODY and only switch if overridden?  

Up to you though I am just afraid someone else if going to upgrade to 8.0 and run into this issue.

tandraschko:
but what is the reason that 7.0 worked? Is it appended to the body in 7.0 per default?

melloware:
No i stepped through the debugger on both 7.0 and 8.0 and the `cfg.appendTo value === undefined` in both.  So it has NO appendTo.   But it works on 7.0 and not on 8.0 until i forced it to appendTo Body. So I can't explain why it works on 7.0 but not 8.0.  Unless something else somewhere has changed that I am not seeing.

tandraschko:
so -1 from my side as it would just be a workaround
we need to find the root problem 
check the DOM or zindex and styles please

melloware:
Hmm but nothing else has changed just upgrading PF 7.0 to 8.0?  So how would the zindex change?  The only variable changed was upgrading from 7.0 to 8.0.

tandraschko:
Another code change in PF or even the theme, dont know
therefore we need to check it :D

PR #5585, opened 2020-02-04
Fix #4655: JQPlot ratio fix

### LINKED ISSUE #4655: JQPlot: Device Pixel Ratio

melloware:
Tested this fix locally from: https://github.com/jqPlot/jqPlot/issues/117

Was able to prove it was broken by Zooming in my browser and that this fix fixes it.


tandraschko:
+1 
please also test other browsers

melloware:
Tested in IE11, Chrome, Firefox and Edge!

Issue #4765, opened 2019-04-28
Columns: SortField and Filter params not passed to LazyDataModel


mafaul:
## 1) Environment
- PrimeFaces version: 7.0
- Application server + version: Payara 5.191
- Theme & Layout: Barcelona v 2.0.0
- Affected browsers: Chrome

## 2) Expected behavior
Filter values entered in globalFilter on column filters should be passed to the lazyDataModel like it did in PF6.2. Same with sortField.

## 3) Actual behavior
Sortfield: null. Filters: {null=[]}

## 4) Steps to reproduce
Enter a value to filter by on a lazy dataTable.
 Note that the values psecified in the datatable asstibutes (sortBy and filterBy) works, filtering after the initial table load has issues.

tandraschko:
Please provide a example or steps to reproduce in the showcase.

mafaul:
I will upload example soon. Working on it.

mafaul:
I got it working by specifying a value for the "field" property on the columns.

Docs: "Name of the field to pass lazy load method for filtering and sorting. If not specified, filterBy-sortBy values are used to identify the field name."

We used to pass the field name (filterBy-sortBy values) like filterBy=#{dbColumnName}. This does not work on PF7. I do however like using the dedicated "field" property more.

melloware:
Shouldn't this still work though?  So to me it sounds like a bug.

mafaul:
PF 4.0 did not have the "field" property on columns, so we had to find a _hacky_ workaround, I don't see this as a bug. 
It is confusing that specifying a value for "field" as well as setting "filterable" to true does not enable the filter for a column, you still have to give a dummy value for filterBy.

mafaul:
Columns highlighted in yellow are never set.
It is almost as if the var (subField) is null at the time that it is needed.
![image](https://user-images.githubusercontent.com/12007706/58962603-dee1ad00-87ab-11e9-8c01-b368b49af3bd.png)

mafaul:
It appears that this issue only appears on p:columns

mridgez59:
This still is an issue in PF 7.0. Any updates?

mdzaebel:
So the issue should be opened again.

tandraschko:
can you please check it with 8.0.RC2?

martin654:
Hi,
[...]

PR #5592, opened 2020-02-07
Fixed #4724: p:columns not working with p:cellEditor. 

### LINKED ISSUE #4724: CellEditor: Dynamic columns cellEdit event
The components _p:dataTable_ and _p:treeTable_ both have the following problem:
## 1) Environment
- All PF, browser and AppServer versions

## 2) Expected behavior
  * Event.old/new values should not be null
  * onCellEditInit should be called
  * Documentation of p:cellEditor should reference p:treeTable too
  * Documentation of p:treeTable should explain, why there are no cellEditInit and cellEditCancel events documented.

## 3) Actual behavior
  * "cellEdit" event.old/new values are null for value expressions like #{bean.values[index]}
  * cellEditInit not signaled?

## 4) Steps to reproduce
Change first "val" in the upper table to "val2".
grafik 
grafik This is correct.

Now, change "values1" to "values11".
grafik
The old and new value in the event is null, which is wrong. Additionally, there is no event fired for *cellEditInit*. I'd recommend, to add test cases for any event into the showcase, as just this case is missing.

Note that form of ValueExpression is absolute important for anyone, who has a data model with a dynamic length of values in a list. Note that this usecase - which seems to be quite normal - is not in you showcases, even if you have "dynamic columns", which however are fixed to possible fields. I don't see an other option to implement a p:dataTable with an unknown number of columns. If there is any workaround, I'd be very interested. Otherwise, could you give me some hints on how to fix it?

## 5) Sample XHTML
```xhtml
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:h="http://java.sun.com/jsf/html"
      xmlns:f="http://xmlns.jcp.org/jsf/core" xmlns:p="http://primefaces.org/ui">
<h:head />
<h:body>
  <h:form id="form">
  <p:growl id="growl" sticky="true"/>
    <p:dataTable value="#{[0]}" var="v" editMode="cell" editable="true" tableStyle="display: inline">
      <p:ajax event="cellEdit" listener="#{bean.onCellEdit}" update=":form:growl" />
      <p:ajax event="cellEditInit" listener="#{bean.onCellEditInit}" update=":form:growl" />
      <p:ajax event="cellEditCancel" listener="#{bean.onCellEditCancel}" update=":form:growl" />
      <p:columns var="c" value="#{[0,1]}">
        <p:cellEditor>
          <f:facet name="input"> 
            <p:inputText value="#{bean.val}" />
          </f:facet>
          <f:facet name="output">#{bean.val}</f:facet>
        </p:cellEditor>
      </p:columns>
    </p:dataTable>
    <p:dataTable value="#{[0]}" var="v" editMode="cell" editable="true" tableStyle="display: inline">
      <p:ajax event="cellEdit" listener="#{bean.onCellEdit}" update=":form:growl" />
      <p:columns var="c" value="#{[0,1]}">
        <p:cellEditor>
          <f:facet name="input">
            <p:inputText value="#{bean.values[c]}" />
          </f:facet>
          <f:facet name="output">#{bean.values[c]}</f:facet>
        </p:cellEditor>
      </p:columns>
    </p:dataTable>
  </h:form>
</h:body>
</html>
```
## 6) Sample bean
```java
@SessionScoped @Named
	String val="val";
	public String getVal() { return val; }
	public void setVal(String val) { this.val = val; }
	
        List<String>values=new ArrayList<>(Arrays.asList("values1","values2"));
        public List<String> getValues() { return values; }
        public void setValues(List<String> values) { this.values = values;  }
	
	public void onCellEdit(CellEditEvent e) { info("cellEdit", e); }
	public void onCellEditInit(CellEditEvent e) { info("cellEditInit", e); }
	public void onCellEditCancel(CellEditEvent e) { info("cellEditCancel", e); }
	
	public static void info(String method, CellEditEvent e) {
		String message=String.format("%s: old=%s, new=%s",
				method, e.getOldValue(), e.getNewValue());
		FacesContext.getCurrentInstance().addMessage(null, 
				new FacesMessage(message));
	}
}
```

TinusJ:
Applied stateless model when column is dynamic on value resolve

### DIFF
--- a/src/main/java/org/primefaces/event/CellEditEvent.java
+++ b/src/main/java/org/primefaces/event/CellEditEvent.java
@@ -29,6 +29,7 @@ import javax.faces.component.UIComponent;
 import javax.faces.component.UIPanel;
 import javax.faces.component.ValueHolder;
 import javax.faces.component.behavior.Behavior;
+import org.primefaces.component.api.DynamicColumn;
 import org.primefaces.component.api.UIColumn;
 import org.primefaces.component.api.UIData;
 import org.primefaces.component.api.UITree;
@@ -100,6 +101,11 @@ public class CellEditEvent<T> extends AbstractAjaxBehaviorEvent {
         else if (source instanceof UITree) {
             TreeTable data = (TreeTable) source;
             data.setRowKey(rowKey);
+        }
+
+        if (column.isDynamic()) {
+            DynamicColumn dynamicColumn = (DynamicColumn) column;
+            dynamicColumn.applyStatelessModel();
         }
 
         T value = null;


TinusJ:
This pull request will fix the following

- https://github.com/primefaces/primefaces/issues/4724
- https://forum.primefaces.org/viewtopic.php?t=34820
- https://forum.primefaces.org/viewtopic.php?t=43286
- https://github.com/primefaces/primefaces/issues/4765

image

melloware (approved):
I tested this definitely fixes #4724

mafaul:
> I tested this definitely fixes #4724

I also tested it, works 100%

melloware:
@mafaul Did you also test https://github.com/primefaces/primefaces/issues/4765  That is the one I did not get a chance to test?

PR #5599, opened 2020-02-14
Fix #5597: Do not execute disabled menu item

### LINKED ISSUE #5597: UIMenuItem: isDisabled() is not checked inside decode()
## 1) Environment
- PrimeFaces version: **ALL**
- Does it work on the newest released PrimeFaces version? Version? **NO**
- Does it work on the newest sources in GitHub? **NO**
- Application server + version: **ALL**
- Affected browsers: **ALL**

## 2) Expected behavior
`UIMenuItem` should conform to every other `UICommand` component, checking for `isDisabled()` before enqueuing an `ActionEvent` inside `decode()` method.

## 3) Actual behavior
It is possible to invoke a **DISABLED** `UIMenuItem` using javascript, leading to execution of unintended code on server

## 4) Steps to reproduce
create a 
```xml
<h:form id="form">
...
<p:menuitem id="menuitem" disabled="true" action="#{someBean.someAction}" value="disabled menuitem" />
...
</h:form>
```
and execute `PrimeFaces.ab({s:"form:menuitem",p:"form",u:"form",f:"form"})`.
Despite the **disabled="true"**, the menuitem is **decoded** and the action is invoked!

## 5) Sample XHTML
```xml
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:ui="http://xmlns.jcp.org/jsf/facelets"
  xmlns:h="http://xmlns.jcp.org/jsf/html" xmlns:f="http://xmlns.jcp.org/jsf/core" xmlns:p="http://primefaces.org/ui">

<h:head>
</h:head>

<h:body>
  <h:form id="form">
    <p:panel>
      <ul>
        <li>
          execute
          <pre>PrimeFaces.ab({s:"form:commandButton2",p:"form",u:"form",f:"form"})</pre>
          and the following timestamp will be resetted to "[NOT SET]".
          <br />
          This means that "#{'#{testMenuitemBean.updateTimestamp}'}" has not been invoked (testMenuitemBean is
          @RequestScoped).
        </li>
        <br />
        <br />
        <li>
          execute
          <pre>PrimeFaces.ab({s:"form:menuitem2",p:"form",u:"form",f:"form"})</pre>
          and the following timestamp
          <h1>will be updated!!</h1>
        </li>
      </ul>
    </p:panel>

    <br />

    <p:panel id="timestampPanel">timestamp: [#{testMenuitemBean.timestamp}]</p:panel>

    <br />

    <p:panel>
      <p:menuButton id="menuButton1" value="menuButton1">
        <p:menuitem id="menuitem1" action="#{testMenuitemBean.updateTimestamp}" process="@form" update="timestampPanel"
          value="enabled menuitem" />
        <p:menuitem id="menuitem2" disabled="true" action="#{testMenuitemBean.updateTimestamp}" process="@form"
          update="timestampPanel" value="disabled menuitem" />
      </p:menuButton>

      <p:commandButton id="commandButton1" action="#{testMenuitemBean.updateTimestamp}" process="@form"
        update="timestampPanel" value="enabled commandButton" />
      <p:commandButton id="commandButton2" disabled="true" action="#{testMenuitemBean.updateTimestamp}" process="@form"
        update="timestampPanel" value="disabled commandButton" />
    </p:panel>
  </h:form>
</h:body>
</html>
```

## 6) Sample bean
```java
import javax.enterprise.context.RequestScoped;
import javax.inject.Named;

@Named
@RequestScoped
public class TestMenuitemBean {
  private String timestamp = "NOT SET";

  public void updateTimestamp() {
    timestamp = String.valueOf(System.currentTimeMillis());
  }

  public String getTimestamp() {
    return timestamp;
  }
}
```

## 7) Solution
this is the current `decode()` of `https://github.com/primefaces/primefaces/blob/master/src/main/java/org/primefaces/component/menuitem/UIMenuItem.java`
```java
@Override
public void decode(FacesContext facesContext) {
  Map<String, String> params = facesContext.getExternalContext().getRequestParameterMap();
  String clientId = getClientId(facesContext);

  if (params.containsKey(clientId)) {
    queueEvent(new ActionEvent(this));
  }

  ComponentUtils.decodeBehaviors(facesContext, this);
}
```

just add:

```java
@Override
public void decode(FacesContext facesContext) {
  if(isDisabled()) {             // <------------   this check
    return;
  }

  Map<String, String> params = facesContext.getExternalContext().getRequestParameterMap();
  String clientId = getClientId(facesContext);

  if (params.containsKey(clientId)) {
    queueEvent(new ActionEvent(this));
  }

  ComponentUtils.decodeBehaviors(facesContext, this);
}
```

However, since `UIMenuItem` has to be used inside another "menu-enabled" component (like `TieredMenu` , `SlideMenu`, `MenuButton`, ...), it would be better to check also if the "menu-enabled" ancestor is disabled.
As of now, I didn't found a simpler way to reach that ancestor, so I'm doing:

```java
@Override
public void decode(FacesContext facesContext)
{
  if(isDisabled()) {
    return;
  }

  UIComponent ancestor = getParent(); 
  while(ancestor instanceof UICommand || ancestor instanceof AbstractMenu || ancestor instanceof Submenu) {
    if(!ancestor.isRendered() || Boolean.valueOf(String.valueOf(ancestor.getAttributes().get("disabled")))) {
      return;
    }
    
    ancestor = ancestor.getParent();
  }

  super.decode(facesContext);
}
```

Thank you and good work!

melloware:


tandraschko:
can't we use #isDisabled via casting or something?

melloware:
I couldn't find a clean or clever way to do it.
 I think its because Menus like TieredMenu , SlideMenu, MenuButton are  (MenuItemAware) and MenuItems and SubMenus are (MenuElement) and there is no common ancestry.  So the cleanest way was the getAttribute with the defensive BooleanUtils check like we do in InputRenderer isDisabled().

Issue #5547, opened 2020-01-20
FileUpload: support chunking and resume


christophs78:
PrimeFaces FileUpload (as of 8.0) is not ideal for uploading realy large files. (hundreds of MB or even larger)
- There is no support for resume
- (as far as i understand) we need a good amount RAM because the uploaded file has to fit into the memory

There are libraries like http://www.resumablejs.com/ and https://github.com/blueimp/jQuery-File-Upload which try to solve both requirements. What do you think about adding this with 9.0?

tandraschko:
If you would like to work on it - sure :)

christophs78:
I´ll look at this at some point in the future.

christophs78:
PrimeFaces already includes https://github.com/blueimp/jQuery-File-Upload.
Maybe we only need the server-side part for chunking. And some client-side configuration. (https://github.com/blueimp/jQuery-File-Upload/wiki/Chunked-file-uploads)

tandraschko:
@rapster @christophs78 could you please check our quality gate? 2 new entries here: https://sonarcloud.io/project/issues?id=org.primefaces%3Aprimefaces&resolved=false&sinceLeakPeriod=true&types=VULNERABILITY

PR #5607, opened 2020-02-17
Implement #5547: FileUpload: support chunking and resume

christophs78:
jQuery File Upload already supports chunked file upload.

Current state of implementation / limitations / ...
- [x] It´s only supported for mode="advanced"
- [x] ~~It´s only supported for native - Impl (NativeFileUploadDecoder). It should be possible to add this to CommonsFileUploadDecoder too.
Update: Implemented for CommonsFileUploadDecoder too.~~
- [x] ~~FileUpload-listener is called once per chunk. It´s up to the application to handle these chunks.~~
- [x] ~~NativeUploadedFile was extended with a few attributes which indicate which chunk was uploaded ->WIP~~
- [x] ~~VirusScanner-integration (after the whole file was uploaded?) -> not supported~~
- [x] ~~Whatˋs about ContentTypeVaildation via Tika? Does it need the whole file? Is it enough to look at the first (and/or last) chunk? -> not supported~~

TODOs
- [x] ~~i´ll provide a showcase-PR in a few days~~
- [x] CSP test with CSP enabled to make sure nothing is broken
- [x] automatic resume (which interval? how much retries? --> configurable)
- [x] (from chunks) merged file does not get removed automatically from tempdir
- [x] listener/chunkListener which one is called at which point in time?
- [x] remove chunk files (of aborted uploads) --> automatically at abort
- [x] add unique praefix to tempFile? (AbstractFileUploadDecoder#processLastChunk)

ShowCase-PR: https://github.com/primefaces/primefaces-showcase/pull/124


Rapster:
I'm actually working on this in my company project. Will definitely keep an eye on your PR :wink:

src/main/java/org/primefaces/component/fileupload/NativeFileUploadDecoder.java
```diff
@@ -97,10 +99,30 @@ private static void decodeAdvanced(FacesContext context, FileUpload fileUpload,
         Part part = request.getPart(clientId);
 
         if (part != null) {
+            String contentRange = request.getHeader("Content-Range");
+
+            Pattern pattern = Pattern.compile("bytes ([0-9]+?)-([0-9]+?)/([0-9]+?)");
```
melloware: Compile this pattern in a static final CONSTANT for performance so it doesn't have to recompile this pattern on every call.

src/main/java/org/primefaces/component/fileupload/NativeFileUploadDecoder.java
```diff
@@ -97,10 +99,30 @@ private static void decodeAdvanced(FacesContext context, FileUpload fileUpload,
...
+                    //chunking is active
+                    uploadedFile.setChunkRangeBegin(Long.parseLong(matcher.group(1)));
+                    uploadedFile.setChunkRangeEnd(Long.parseLong(matcher.group(2)));
+                    uploadedFile.setChunkTotalFileSize(Long.parseLong(matcher.group(3)));
+                    if ((uploadedFile.getChunkRangeEnd() + 1) == uploadedFile.getChunkTotalFileSize()) {
+                        uploadedFile.setLastChunk(true);
+                    }
+                }
+                else {
+                    //TODO: throw exception?
```
melloware: Probably safe to throw an exception here as its an expected case.

melloware (changes requested):

Rapster:
> FileUpload-listener is called once per chunk. It´s up to the application to handle these chunks.

I think current `listener` should be called once file is fully uploaded. Now I wonder if we should also have a listener for chunks, first use case that comes on top of my head is digest calculation

src/main/java/org/primefaces/component/fileupload/NativeFileUploadDecoder.java
Rapster: Based on https://tools.ietf.org/html/rfc7233#section-4.2, I would go for `^bytes (\d+)-(\d+)\/(\d+|\*)$`

christophs78:
> I think current `listener` should be called once file is fully uploaded. Now I wonder if we should also have a listener for chunks, first use case that comes on top of my head is digest calculation

I´m not sure about this point yet....
1. It should be possible to provide a specialized listerner for chunks. (Thats the easy thing.)
2. What do we expect from the "classic" fully-uploaded listener in chunked-upload-mode? We try to avoid to keep the whole file in memory. Which information should the fully-uploaded listener provide in chunked-upload-mode?
3. What should VirusScan do in chunked-upload-mode? We OOTB never have the whole file. We only have chunks. IMO it does not make sense to only scan chunks. Should this be handled by the application? Or should PF chunked-file-upload write one condensed temporary file in a configureable file-system-location? Do VirusScan based on this and provide this file to the fully-uploaded listener?

christophs78:
@Rapster: It´s still WIP but can you have a look at this PR? Any thoughts?

melloware:
@christophs78 Looks like you have some minor conflicts.

docs/9_0/components/fileupload.md
```diff
@@ -285,6 +287,15 @@ folder.
...
 
+## Chunking
+FileUpload supports chunked fileupload in advanced-mode via maxChunkSize - attribute.
+
+Chunked file upload comes with following restrictions:
+1. It is only supported for `mode="advanced"`
+2. It does not support `virusScan="true"` and `validateContentType="true"` (The application has to care about this itself. PrimeFaces can't handle this for chunked upload.)
+
+TODO: code-example, restrictions, ....
+
```
melloware, line 328: Still need examples here?

src/main/java/org/primefaces/component/fileupload/FileUpload.java
```diff
@@ -99,4 +121,8 @@ else if (newValue instanceof UploadedFiles) {
             }
         }
     }
+
+    public boolean isChunkedUpload() {
+        return (getMaxChunkSize() > 0);
```
melloware: You can remove the parenthesis around this.

melloware (changes requested):

Rapster:
Hey @christophs78,

Sorry for the late answer, but quite busy with work lately... Anyway, first and foremost: great job! I'm happy to see that feature implemented. Since I'm working on uploading large files as well, this will definitely help us to improve the way we upload files.

Though I have a couple of questions:
- Chunks are being merged by the end-user (based on your wiki doc) why that? I might be wrong but I would expect the component to do that by itself. 
- My idea of having a `chunkListener`  was to be able to do few things meantime the file is being uploaded (for example, computing digest file)
- I don't think chunks should be set as submittedValue, only the "final" file should be. Chunks are only temporary structures, they're not validated ,processed virus-scanned or anything so I wouldn't submit them as a proper value. Being said, that should simplify a lot your impl.
- Chunks upload should also be supported in "simple" mode, but since you're merging the final file through chunkListener, I understand why you can't.

One final thought, IMO the best thing about having chunk upload is to be able to resume an upload (see https://github.com/blueimp/jQuery-File-Upload/wiki/Chunked-file-uploads#resuming-file-uploads). We can talk about it once we're clear on how chunks should be handled by the component

christophs78:
Hi @Rapster!

Thanks for your feedback!

IMO there are two motiviations for chuncked file upload:
- Reduce the serverside memory-footprint when uploading large files (eg 1GB)
- Provide resume functionality. (Primarily for slow or not stable internet connections.)

My efforts to this point concentrate on the serverside memory-footprint because this is a issue for the Ops-colleagues in my organisation. 
My chunked - implementation is somewhat basic. We "stream" the (large) uploaded file(s) to the server. And the implementation on the server has to decide how it likes to further process the chunks.

Maybe we need something like to modes for chunked file upload:
- One which concentrates one low serverside memory-footprint (which is implemented in this PR) at the expense of increased developer-complexity.
- One which concentrates on resume-functionality and keeps a simple API at the expense of increased memory-footprint.

Do you see a way to catch both requirements with one implementation?

christophs78:
Is it an option to merge this PR after review and work on the resume-part in a separate PR?
(I already started playing around with resume.)

tandraschko:
It would be great if you can first agree with @rapster about the feature set and resolve the open tasks :)

Rapster:
Hi everyone,

Let me first answer to your last statement @christophs78:
> Reduce the serverside memory-footprint when uploading large files (eg 1GB)

Not sure what memory you're referring to, if it's RAM then chunks use more memory than a one long sequential upload (and that's okay! unless your chunks are 1GB 😅 ). Chunks are saved in memory (according to Jetty at least) while no-chunk write a file through InputStream. Though, network congestion is more likely to happen with no-chunk.

So basically, IMO what we want to solve with chunks are:
- avoiding network congestion
- and just like you said, the resume functionnality. It's actually the very first reason we want this, since it solves a lot of problem caused by long uploads (e.g network instability, timeouts etc.)

@christophs78 I can submit a PR on your branch to share my thoughts regarding my previous comments.

christophs78:
@Rapster: You are right. I tried to solve a phantom-issue. We had issues with uploading large files (up to 1 GB), where our application-server took a multiple of the filesize as memory.
Then we read the wrong internet-sources and my colleagues and i went into to wrong direction. 
And we did not debug what realy happens inside the application server / JSF / PrimeFaces until now. 
(Our real issue must have been the processing of the uploaded file.)

Even in our environment with OpenLiberty the upload-handling seems to be fine OOTB:
image

So i´m totally happy when you submit a PR based on my branch.

Rapster:
I'll try to push something this week, I managed to make chunks and resume working

docs/9_0/components/fileupload.md
```diff
@@ -285,6 +287,67 @@ folder.
...
+		FacesContext.getCurrentInstance().addMessage(null, msg);
+	}
+}
+
+public void handleFileUpload(FileUploadEvent event) {
+	FacesMessage msg = new FacesMessage("Successful", event.getFile().getFileName() + " is uploaded.");
+	FacesContext.getCurrentInstance().addMessage(null, msg);
+
+    try {
+        //TODO: process file
```
melloware: Still a TODO?

melloware (changes requested):

christophs78:
Many thanks to @Rapster who did great things in his "nightshift".
Tried to keep Checkstyle happy and added a bit of JSDoc to keep this happy too. (Still some things to learn regarding TypeScript and JSDoc. At least for me.)
I´ll come back and update the showcase-PR later this day.

src/main/java/org/primefaces/component/fileupload/FileUpload.java
```diff
@@ -97,8 +98,13 @@ else if (newValue instanceof UploadedFiles) {
                     throw new IllegalArgumentException("Argument of type '" + newValue.getClass().getName() + "' not supported");
                 }
 
-                if (newValue instanceof UploadedFile && "advanced".equals(getMode())) {
-                    queueEvent(new FileUploadEvent(this, (UploadedFile) newValue));
+                if ("advanced".equals(getMode())) {
```
Rapster: chunks are never submitted as a value (and that's on purpose), so this condition will never be verified

Rapster:
Thanks @christophs78 so far I think it's a good first draft, in a few words:
- chunk and resume features are supported
- chunks are not submitted as values
- merge chunks is supported by the component out of the box (but it's possible for the user to define is own through SPI)
- tika content type is not verified for chunks (but as mentionned @christophs78 earlier we could test the first chunk)

Though I'm not happy with one thing: `FileUploadChunkDecoder#getUploadDirectory`
We should be able to retrieve the upload directory from apache commons or servlet 3.0, and not hardcoded like I did. I'm still testing with servlet 3.0 to make sure I understand how the MultiPartConfig works, but so far it's quite confusing...

src/main/java/org/primefaces/component/fileupload/DefaultFileUploadChunkDecoder.java
```diff
@@ -0,0 +1,94 @@
...
+
+    protected UploadedFile processLastChunk(FileUpload fileUpload, UploadedFileChunk chunk, Path chunksDir) throws IOException {
+        Path wholePath = Paths.get(getUploadDirectory(), chunk.getFileName());
+        Files.deleteIfExists(wholePath);
+        Path whole = Files.createFile(wholePath);
+
+        List<Path> chunks = FileUploadUtils.listChunks(chunksDir);
+        for (Path p : chunks) {
+            Files.write(whole, Files.readAllBytes(p), StandardOpenOption.APPEND);
+            Files.delete(p);
```
Rapster: The chunkDir is deleted right after the final file is merged, I'd rather delete the chunks in one shot after making sure the final file is merged

Rapster (changes requested):

src/main/java/org/primefaces/component/fileupload/DefaultFileUploadChunkDecoder.java
christophs78: Deleting chunkDir with (chunk-)files within throws a DirectoryNotEmptyException.

src/main/java/org/primefaces/component/fileupload/FileUpload.java
christophs78: IMO the core question is do we want to to have a chunkListener? Or are we fine with the „normal“ listener which get´s fired after the last chunk.
The code you wrote last night fired the „normal“ listener after each chunk. With my modification we fire chunkListener after each chunk and the „normal“ listener is fired after the last chunk was uploaded and we have the whole file on the server.
From my side i do not think we need the chunkListener. IMO ist kind of nice2have.

christophs78: IMO the core question is do we want to to have a chunkListener? Or are we fine with the „normal“ listener which get´s fired after the last chunk.
The code you wrote last night fired the „normal“ listener after each chunk. With my modification we fire chunkListener after each chunk and the „normal“ listener is fired after the last chunk was uploaded and we have the whole file on the server.
From my side i do not think we need the chunkListener. IMO ist kind of nice2have.

src/main/java/org/primefaces/component/fileupload/DefaultFileUploadChunkDecoder.java
Rapster: Ah indeed... Last night I brought a lot of modifications just like this one, I guess it was too late to see it 😅 Anyway i'll have a look at it

src/main/java/org/primefaces/component/fileupload/FileUpload.java
Rapster: That's actually something have been wondering as well, do we need this chunkListener...? The only use case I have in mind is whenever you want and can calculate something over chunks (as mentionned digest for example) instead of processing the whole file. Have a look at https://docs.oracle.com/javase/7/docs/api/java/security/DigestInputStream.html to have a clearer of what I'm referring to.

Right now, that's probably something we don't really need, being said, the use case I mentionned might be solved using a custom `FileUploadChunkDecoder`

christophs78:
@Rapster: Thanks for your work from late monday!

Just added auto-resume. What do you think about this? Should it be configureable via (eg) a chunkUploadConfig-object?

Can you have a look at our TODO-list at the top of this PR. Any thoughts on them? What should we implement? Should we discard some of them?

Rapster:
I'll have to check the todo list. Have you ever had that issue: `Uncaught Error: cannot call methods on fileupload prior to initialization; attempted to call` calling `$.blueimp.fileupload.prototype.options.fail.call(this, e, data);`?

I see you're not using it while in the plugin example, it does. And I think we need it to support the auto upload feature

christophs78:
> I'll have to check the todo list. Have you ever had that issue: `Uncaught Error: cannot call methods on fileupload prior to initialization; attempted to call` calling `$.blueimp.fileupload.prototype.options.fail.call(this, e, data);`?
> 
> I see you're not using it while in the plugin example, it does. And I think we need it to support the auto upload feature

Just added auto="true" to https://github.com/primefaces/primefaces-showcase/blob/a0b8155ea4c9d4f158b7a08ab05104dba13c0821/src/main/webapp/ui/file/upload/chunked.xhtml. Seems to work fine with FF 75 and Chrome 81.
You got this error once or is it reproduceable? All browsers?

christophs78:
Added following things with my last commit:
- we create a subdirectory "PF_chunkedUpoad" below java.io.tmpdir and use this for upload (makes it easier for cleanup)
- the merged files get a UUID-prefix (users may upload files with the same names; these should not collide)
- a simple cleanup-mechanism which run´s once a day an removes all chunkDirs (from aborted uploads) which are older than one day; we may need to discuss this one, make it configureable, ...

Rapster:
Sorry I wrote that in the morning but forgot to press the button 😅 :

Yeah just adding auto attribute seems to be working...

Regarding the todolist:

> - [ ] automatic resume (which interval? how much retries?)

Introduce 2 new attributes

> - [ ] (from chunks) merged file does not get removed automatically from tempdir

When do they get removed usually?

- [ ] remove chunk files (of aborted uploads) after eg 1 day

I'm working on it, but ideally the component shouldn't have to check ellapsed time of the file and then delete it, a cron job would be more appropriate (just like the author of the plugin mention)

- [ ] add UUID-praefix to tempFile to wholeFile? (AbstractFileUploadDecoder#processLastChunk)

Simply prefix with the X-File-Id such as `[X-File-Id]Filename.ext`.

christophs78:
> > * (from chunks) merged file does not get removed automatically from tempdir
> 
> When do they get removed usually?

When i think about the temp files for upload handled by the servlet container (servlet 3.0), they are IMO removed after the request.
So i think about a PF-filter which implements javax.servlet.Filter, remembers the merged file in a static ThreadLocal-variable and removes it at the end of the request.

Plan B may be something like the developer has to remove the merged file manually after it´s processed.

christophs78:
Tomcat 9.0.12 throws the following error during startup:
01-May-2020 10:41:58.677 SCHWERWIEGEND [RMI TCP Connection(5)-127.0.0.1] com.sun.faces.config.ConfigureListener.contextInitialized Critical error during deployment: 
 java.lang.UnsupportedOperationException: **Section 4.4 of the Servlet 3.0 specification does not permit this method to be called from a ServletContextListener that was not defined in web.xml, a** web-fragment.xml file nor annotated with @WebListener
	at org.apache.catalina.core.StandardContext$NoPluggabilityServletContext.addListener(StandardContext.java:6593)
	at **org.primefaces.webapp.PostConstructApplicationEventListener.addUploadedFileCleanerListener(PostConstructApplicationEventListener.java:65)**
	at org.primefaces.webapp.PostConstructApplicationEventListener.processEvent(PostConstructApplicationEventListener.java:60)
	at javax.faces.event.SystemEvent.processListener(SystemEvent.java:123)
	at com.sun.faces.application.applicationimpl.Events.processListeners(Events.java:253)
	at com.sun.faces.application.applicationimpl.Events.invokeListenersFor(Events.java:231)
	at com.sun.faces.application.applicationimpl.Events.publishEvent(Events.java:112)
	at com.sun.faces.application.ApplicationImpl.publishEvent(ApplicationImpl.java:127)
	at javax.faces.application.ApplicationWrapper.publishEvent(ApplicationWrapper.java:788)
	at com.sun.faces.config.ConfigManager.publishPostConfigEvent(ConfigManager.java:548)
	at com.sun.faces.config.ConfigureListener.contextInitialized(ConfigureListener.java:265)
	at org.apache.catalina.core.StandardContext.listenerStart(StandardContext.java:4641)
...

src/main/resources/META-INF/resources/primefaces/fileupload/2-fileupload.js
```diff
@@ -225,13 +247,55 @@ PrimeFaces.widget.FileUpload = PrimeFaces.widget.BaseWidget.extend({
                 }
             },
             fail: function(e, data) {
+                if ($this.cfg.resumeContextPath) {
+                    $.ajax({
+                        url: $this.cfg.resumeContextPath + '?' + $.param({'X-File-Id' : $this.createXFileId(data.files[0])}),
+                        dataType: 'json',
+                        type: 'DELETE'
+                    });
+                }
```
christophs78: Not sure whether this should be at the start of fail-function. Think this would delete already uploaded chunks before we have the chance to resume.

Rapster: Not sure what you mean? fail function is called when user abort upload

Rapster:
Thanks for reporting that issue on Tomcat, I guess I can't add it programmatically that way then, so it'll be up to the user to add that listener, which is not a bad thing by the way since I don't fully understand why uploaded files should be right away deleted

src/main/resources/META-INF/resources/primefaces/fileupload/2-fileupload.js
christophs78: Fail is also called when the network-connection breaks. (eg when i simulate this via browser-devtools) 
We may move this block into the `if (data.errorThrown === 'abort')` { a few lines down.

christophs78:
> Thanks for reporting that issue on Tomcat, I guess I can't add it programmatically that way then, so it'll be up to the user to add that listener, which is not a bad thing by the way since I don't fully understand why uploaded files should be right away deleted

Is it a bad idea to add https://docs.oracle.com/javaee/7/api/javax/servlet/annotation/WebListener.html - annotation which exists since Servlet 3.0 or ship web-fragment.xml as part of PrimeFaces? (so only a few people still using 2.x - Servlet - containers would need to handle this themself)
Servlet 3.0 was introduced 2009. So i would hope most of our PF-userbase uses a servlet container which supports this. Or am I too optimistic?

src/main/resources/META-INF/web-fragment.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Rapster, line 1: If I'm not mistaken, web-fragment is servlet 3.0

tandraschko:
We should add a web-fragment and add a docu hint for serlvet 2.5.
I wouldn't use annotations for a filter as they were broken partly in older containers.

docs/9_0/components/fileupload.md
```diff
@@ -287,6 +290,43 @@ folder.
...
+```xml
+<listener>
+	<listener-class>org.primefaces.webapp.UploadedFileCleanerListener</listener-class>
+</listener>
+```
+
+The single uploaded chunks also get put into an internal (temporary) upload directory.
+These chunk-files get removed
+1. after the last chunk is uploaded and the merged file is created 
+2. when the users aborts the upload.
```
christophs78: @Rapster : Can you help out with your english-language-skills? I´m not totally happy with my text.

Rapster:
I think we good for now. Feel free to review that PR

christophs78:
`UploadedFileCleanerListener` does not delete the merged file anymore.
It´s maybe a regression due to https://github.com/primefaces/primefaces/pull/5607/commits/0fd106ca4b4b3837106481907cbd600412b41266.

Rapster:
Fixed :wink:

christophs78:
I´m happy with this PR.
@melloware, @tandraschko : do you want to have a look at this PR?

melloware (approved):
I have been following this PR the whole time and it looks great.  this is a nice improvement to PF.

tandraschko:
I wonder if we shouldn't use the servlet temp dir for uploading? (https://docs.oracle.com/javaee/6/api/javax/servlet/ServletContext.html#TEMPDIR)

Rapster:
I just had a try, `request.getServletContext().getAttribute(ServletContext.TEMPDIR)` returns a tmp folder contained in deployed app (target folder, e.g `D:\showcase-facelift\target\tmp`). I don't think we want to keep chunks in that location.

Besides that, I read this post https://stackoverflow.com/a/30111831/4605161 and BalusC answer. Made me realize that Servlet 3.0 is JEE 6.0, so how come do we still support Servlet 2.5? Being said, it does not mean we should remove code like CommonsFileUploader, but do we support Servlet 2.5 anywhere else?

tandraschko:
Actually the servlet-tempdir feels IMO more correct as the java tempdir.

We currently do support Servlet2.5 as there is not much difference. Just some auto-register features like web-fragment are missing but therefore we added it in the docu.
I think we could add a servlet25 flag in the PrimeConfig and therefore use servlet-tempdir or java-tempdir.

Rapster:
Yeah but see, even though there is not much of a difference between 2.5 and 3.0, PF codebase will still have to make a few adjustments to support 2.5 (so more tricks to detects which Servlet version are we running on...) Moreover, which does not really make clear where PF stands on which environnement it should run, JEE 6 but compatible Servlet 2.5? Sounds a bit confusing to me... I would rather keep things simple and clear, simply by saying it's JEE 6 therefore Servlet 3.0.

Regarding tempdir, I guess there are pros and cons, more cons than pros I think according to this: https://stackoverflow.com/a/18664715/4605161 But i'm opened to discuss what's best @melloware @christophs78

tandraschko:
As long we can easily support it, we should do it (like with adding docu hint to add the listener in the web.xml instead of rely on the web-fragment). Same with JSF 2.0.
If this feature will never run on 2.5 because of missing API parts, it's ok for me aslong we add it as hint to the docu. 

We probably MUST do a "modern" version of PrimeFaces after JSF 3.0, to use new API parts then.
3.0 will change just change the packages, JSF 4.0 will be with new features/deprecations and so on.

Then we also can remove support for older EE-specs.
We will probably have 3 submodules of primefaces then:
- core (maybe only JS?)
- javax / legacay
- jakarta / modern

christophs78:
> Regarding tempdir, I guess there are pros and cons, more cons than pros I think according to this: https://stackoverflow.com/a/18664715/4605161 But i'm opened to discuss what's best @melloware @christophs78

Just asked myself what my Ops-colleague would prefer. (He is a 100+ kilo man with a lot of temperament.) In our organisation we have a so called document-path (which is set via -D vm-argument) which points to a directory where applications may save their files.
So i would personally prefer to make tempdir configurable. And use something else (java.io.tmpdir) as fallback.
That would be similar to what https://primefaces.github.io/primefaces/8_0/#/components/fileupload?id=filter-configuration does.

Regarding supporting very old JSF-, Servlet-, ... versions. We have limited resources and IMO we should concentrate them on efforts that move PrimeFaces forward. (Running regression- and compatiblility-tests against Servlet 2.5 does not make much sense IMO. Servlet 3.0 was released in 2009)

As suggested by @tandraschko we should IMO do this cut for one the next PrimeFaces major releases.

Rapster:
> In our organisation we have a so called document-path (which is set via -D vm-argument) which points to a directory where applications may save their files.
> So i would personally prefer to make tempdir configurable. And use something else (java.io.tmpdir) as fallback.
> That would be similar to what https://primefaces.github.io/primefaces/8_0/#/components/fileupload?id=filter-configuration does.

It's already existing, either overriding `java.io.tmpdir` (which is more less meant for that) in your VM argument or this way:
```xml
<servlet>
        <servlet-name>Faces Servlet</servlet-name>
        <servlet-class>javax.faces.webapp.FacesServlet</servlet-class>

        <multipart-config>
            <location>D:\Test</location>
        </multipart-config>
    </servlet>
```

I don't consider chunks as a  "proper" uploaded file, they're just temporary data, so located in tmpdir fits more that purpose

tandraschko:
Im sure you can also set the servlet tempdir in Tomcat e.g.

Rapster:
You could, but that would change one thing: your war would be deployed into that directory. I don't think this is what we want

Rapster:
Mmmh I'll dig deeper regarding this constant, you're right that might be the right way to go 😉 i'll let you know

tandraschko:
then lets to for a context-param and fallback to the java tempdir?

Rapster:
Alright, I kept things very simple, chunks and merged file are located in same parent folder
1. Use `javax.servlet.context.tempdir` as primary location
2. If defined, it'll use either the dir from apache commons or servlet 3.0
3. If none of this is set, then fallback to `java.io.tmpdir` (which should probably never happen)

tandraschko:
Thanks for your effort - looks really good me to.
I will merge it now. If you can further improve it, just send a PR :D

Rapster:
Thank you Thomas, for such commit you should use the squash and merge option, I was planning to do the merge myself but you were faster than me :sweat_smile:

Issue #5608, opened 2020-02-18
FileUpload: Refactor decoders


Rapster:
I was having a look at #5594 and I figured the whole decoding thing could be simplified

PR #5615, opened 2020-02-21
Fix #5594 - FileUpload: Improve VirusScanner Interface

### LINKED ISSUE #5594: FileUpload: Improve VirusScanner Interface
I would suggest to improve the VirusScanner interface which is intented for implementing custom scan services, so that within the implementation file metadata about the uploaded file (name, MIME-type) and a reference to the fileupload component is available. This would make it possible to log the filename etc. and to inform the user about the problem if a virus threat was detected.

Also it would be great to change the `org.primefaces.virusscan.VirusException` in a way that it extends `javax.faces.validator.ValidatorException` and the message of the thrown exception is shown as error message in the FileUpload component for informing the user about the detected problem with the file.

I described my problem here: [https://forum.primefaces.org/viewtopic.php?f=3&t=61116](https://forum.primefaces.org/viewtopic.php?f=3&t=61116)

## 1) Environment
- PrimeFaces 7.0.11
- JSF: Mojarra 2.3.9.SP04
- JBoss Wildfly 18.0.1 Final
- Affected browsers: all Browsers

## 2) Expected behavior

The interface should perhaps look like this:

`    /**
     * Perform virus scan and throw exception if a virus has been detected.
     * @param inputStream input stream to perform virus scan on
     * @param facesContext the current FacesContext
     * @param uploadedFile The uploaded file the scan is conducted for containing the files metadata
     * @param fileUpload the fileUpload component
     * @throws VirusException if a virus has been detected by the scanner
     */
    void performVirusScan(InputStream inputStream, FacesContext facesContext, UploadedFile uploadedFile, FileUpload fileUpload) throws VirusException;

`

If it is not acceptable to provide the UploadedFile (which in turn exposes an InputStream of the uploaded file, what may not be intended.), at least the name and mime type should be provided.

## 3) Actual behavior

Currently only an FileInputStream is available:
`    /**
     * Perform virus scan and throw exception if a virus has been detected.
     * @param inputStream input stream to perform virus scan on
     * @throws VirusException if a virus has been detected by the scanner
     */
    void performVirusScan(InputStream inputStream) throws VirusException;

`

Rapster:
* Rename `Fileupload#performVirusScan` into `Fileupload#virusScan`
* Uploaded file validation moved to `FileUpload#validateValue`
* Refactor VirusTotalReportScanner

@Thomas-Schindler I changed `VirusTotalReportScanner` class so you can add your own error message. See `VirusTotalReportScanner#createErrorMessage`

This PR should be merged on top of #5608

### DIFF
--- a/src/main/java/org/primefaces/component/fileupload/CommonsFileUploadDecoder.java
+++ b/src/main/java/org/primefaces/component/fileupload/CommonsFileUploadDecoder.java
@@ -24,10 +24,7 @@
 package org.primefaces.component.fileupload;
 
 import org.apache.commons.fileupload.FileItem;
-import org.primefaces.event.FileUploadEvent;
-import org.primefaces.model.file.CommonsUploadedFile;
-import org.primefaces.model.file.UploadedFileWrapper;
-import org.primefaces.util.FileUploadUtils;
+import org.primefaces.model.file.*;
 import org.primefaces.webapp.MultipartRequest;
 
 import javax.faces.FacesException;
@@ -36,9 +33,6 @@ import javax.servlet.ServletRequestWrapper;
 import java.io.IOException;
 import java.util.List;
 import java.util.stream.Collectors;
-import org.primefaces.model.file.UploadedFile;
-import org.primefaces.model.file.UploadedFiles;
-import org.primefaces.model.file.UploadedFilesWrapper;
 
 public class CommonsFileUploadDecoder {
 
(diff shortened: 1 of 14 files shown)


tandraschko:
+1
i would just not rename "VirusTotalReportScanner", the old name matches the interface better

Rapster:
Besides it's a lot of "Virus" in one word. `VirusTotalVirusScanner` name is a bit confusing as it's implying it requesting for a scan while it's not, it calls `file/report` endpoint. So for now, I prefer `VirusTotalReportScanner`

tandraschko:
I dont feel that its right.
Its a implementation of the VirusScanner, which uses VirusTotal.
It doesn't matter if the internal VirusTotal API or URL is called "file/report".

An implementation for Kaspersky would be called KasperskyVirusScanner.

Thomas-Schindler:
I agree here with @Rapster, as the VirusTotal API actually does not perform a virus scan.

tandraschko:
why does it implement a VirusScanner then? :D

Thomas-Schindler:
It just sends the MD5, SHA-1 or SHA-256 Hash value of the file to the VirusTotal "file/report" API. If a report exist for this hash, it is returned, otherwise a response code "0" with the message "The requested resource is not among the finished, queued or pending scans" is returned. But no scan is performed.

tandraschko:
I see, then lets merge it

Issue #5619, opened 2020-02-24
DatePicker: ui-datepicker-prev does not show previous month on single click


DilyanaN:
### 1) Environment
- PrimeFaces version: 7.0.11
- Does it work on the newest released PrimeFaces version? Version? No
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) No
- Application server + version:
- Affected browsers: Chrome, IE, Firefox...
### 2) Expected behavior
On single click on ui-datepicker-prev (after submit), previous month is displayed, independent on the selected date.

### 3) Actual behavior
If a date is selected, which is not existent in the previous month, the user has to click twice on ui-datepicker-prev in order to see previous month.

### 4) Steps to reproduce
- Choose 31. March 2020
- Click on "Submit"
- Click on the datePicker field
- Select previous month via ui-datepicker-prev
- You have to click again on ui-datepicker-prev in order to see previous month

5) Sample XHTML

```
<f:view>
  <h:body>
      <h:form>
        <div class="ui-g">

          <div class="ui-g-6">
            <p:datePicker yearNavigator="true" showIcon="true" yearRange="1900:2150" pattern="dd.MM.yyyy"/>
          </div>

          <div class="ui-g-6">
            <p:commandButton value="Submit" ajax="false"/>
          </div>

        </div>
      </h:form>

  </h:body>
</f:view>
```

melloware:
Attached is a reproducer and I could reproduce this with 8.0RC3 this time.
[pf-5619.zip](https://github.com/primefaces/primefaces/files/4257305/pf-5619.zip)

PR #5622, opened 2020-02-26
Fix #5618,#5619: DatePicker prev/next month button calculation

### LINKED ISSUE #5618: DatePicker: ui-datepicker-next skips next month
## 1) Environment
- PrimeFaces version: 7.0.11
- Affected browsers: Chrome, IE, Firefox...

## 2) Expected behavior
On click on "Next month" (after submit), the next month is displayed, independent on the selected date.

## 3) Actual behavior
If a date is selected, which is *not* existent in the next month, by selecting next month via ui-datepicker-next (after submit) the next month is skipped and the month after is displayed.

## 4) Steps to reproduce
- Choose 31. January 2020
- Click on "Submit"
- Click on the datePicker field
- Select next month via ui-datepicker-next
- See month *March*

## 5) Sample XHTML

```
<f:view>
  <h:body>
      <h:form>
        <div class="ui-g">

          <div class="ui-g-6">
            <p:datePicker yearNavigator="true" showIcon="true" yearRange="1900:2150" pattern="dd.MM.yyyy"/>
          </div>

          <div class="ui-g-6">
            <p:commandButton value="Submit" ajax="false"/>
          </div>

        </div>
      </h:form>

  </h:body>
</f:view>
```

melloware:
The issue was the Javascript `setMonth(month-1)` to set to the previous or next month.

Date: March 31, 2020 if you clicked previous set the date to March 02, 2020.  This is because Feb doesn't have a 31st day so Javascript is confused.  Its the same issue for January 31st rolling next month goes to March.

The fix is simple use the overloaded `setMonth(month, day)` and always use day 1 since all months have a 1st day of the month.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
+++ b/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
@@ -1638,11 +1638,11 @@
 
             if (this.options.view === 'date') {
                 if (newViewDate.getMonth() === 0) {
-                    newViewDate.setMonth(11);
+                    newViewDate.setMonth(11, 1);
                     newViewDate.setFullYear(newViewDate.getFullYear() - 1);
                 }
                 else {
-                    newViewDate.setMonth(newViewDate.getMonth() - 1);
+                    newViewDate.setMonth(newViewDate.getMonth() - 1, 1);
                 }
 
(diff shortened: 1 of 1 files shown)


christophs78 (approved):
Thanks for fixing!

Issue #5627, opened 2020-03-01
Docu: show PF-version and add version-switch


christophs78:
People open documentation and sometime do not know which version of the documentation they opened. (This causes issues because sometimes due to refectorings newer documentation does not apply to older PF-versions.) Think it would make sense to explicit show the PrimeFaces - version - number. (Not only on the start-page.) Maybe somewhere near the PrimeFaces-documentation image in the upper left.

Or even better offer a drop-down to toggle between PrimeFaces-versions. (eg 7.0, 8.0 and a future 9.0)

PR #5632, opened 2020-03-02
Implement #5627:  Docu: show PF-version and add version-switch

christophs78:


christophs78:
To avoid confusion:
Docsify seems to have a limitation in it´s navbar-feature. Immediately after showing the coverpage the navbar containing the version-switch is not visible.
But it´s there on all "normal" pages.
grafik

melloware (approved):

PR #5633, opened 2020-03-02
Fix #5624: Datatable Excel exporter autosize incorrect cols

### LINKED ISSUE #5624: DataTableExcelExporter: DataTable.getColumnsCount() returns wrong value exporters
All implementations of [`DataTableExporter.exportCells()`](https://github.com/primefaces/primefaces/blob/8.0.RC3/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java#L304), e.g. DataTableExcelExporter, do column export if it is [rendered and **exportable**](https://github.com/primefaces/primefaces/blob/8.0.RC3/src/main/java/org/primefaces/component/datatable/export/DataTableExcelExporter.java#L117-L119).

But the [`DataTable.getColumnsCount()`](https://github.com/primefaces/primefaces/blob/8.0.RC3/src/main/java/org/primefaces/component/datatable/DataTable.java#L909) considers only columns which are [rendered](https://github.com/primefaces/primefaces/blob/8.0.RC3/src/main/java/org/primefaces/component/datatable/DataTable.java#L914) and [**visible**](https://github.com/primefaces/primefaces/blob/8.0.RC3/src/main/java/org/primefaces/component/datatable/DataTable.java#L922).

This discrepancy causes the ["autoSize" is not set for all exported columns](https://github.com/primefaces/primefaces/blob/8.0.RC3/src/main/java/org/primefaces/component/datatable/export/DataTableExcelExporter.java#L83-L87). Namely the columns with `visible="false"` are exported but not counted, therefore "autosize" is set for fewer columns than actually exported.

PrimeFaces 8.0 RC3

----------

Use case: exporting more columns than displayed in GUI by setting `visible="false"` to those additional columns.

----------

Current workaround: set CSS style `display: none;` for the `p:column`. In this way, it formally remains "visible" and therefore gets counted in `getColumnsCount()`.

melloware:


melloware:
Created showcase example and tested it and all is now working correctly with auto-sizing and the exportable flag.

PR #5634, opened 2020-03-03
Fix #5635: unauthorized ajax request redirect

### LINKED ISSUE #5635: Server requests that result in an unauthorized request (eg session timeout) should redirect to Location(aka) login
This is an improvement request. The use case is for a PrimeFaces app where the session has timed out, and the user then attempt to interact again with the app. When an AJAX request eventually happens the server will reply with a 401 (Unauthorized), and result in nothing. It would be desirable if the LOCATION header in the response was used to load that page via a window.location.

A fix has been identified and a PR submitted: https://github.com/primefaces/primefaces/pull/5634

## 1) Environment
- PrimeFaces version: v 7.0

The same block in core-ajax.js remains unchanged so it is expected the same behaviour results.

JBOSS EAP 7.1.6.GA

- Affected browsers: ALL

## 2) Expected behavior
This is a feature (improvement) request. A better solution would be to honour a 401 (Unauthorized) response and redirect the browser window location to this header.

## 3) Actual behavior
None. 401 Location is ignored.

## 4) Steps to reproduce
Interact with any app so that AJAX requests are made to the server, after waiting for the session to timeout.

## 5) Sample XHTML
N/A

## 6) Sample bean
N/A

mP1:


melloware:
Please submit an issue first describing the problem. PR will not be accepted without it.

mP1:
Closes https://github.com/primefaces/primefaces/issues/5635

tandraschko:
Please test with multiple containers and MyFaces and Mojarra

src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
```diff
@@ -538,6 +538,12 @@ if (!PrimeFaces.ajax) {
 
                 var jqXhr = $.ajax(xhrOptions)
                     .fail(function(xhr, status, errorThrown) {
+                        if (xhr.status == 401) {
+                            var location = xhr.getResponseHeader("Location");
+                            PrimeFaces.debug('Unauthorized status received. Redirecting to ' + location);
+                            window.location = location;
```
melloware, line 544: What if location is NULL?  Shouldn't it then just call `cfg.onerror.call` if no location URL is found?

melloware (changes requested):

src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
mP1:
```java
if (xhr.status == 401) {
```

becomes

```java
var location = xhr.getResponseHeader("Location");
if (xhr.status == 401 && locataion) {
```

this solves the null location header problem previously mentioned.

mP1:
Being selfish here, but when is the next release due ?

melloware:
Since 8.0 just come out don't expect 9.0 until March 2021 based on past history.  If you are a PRO member you can request this gets into 8.0.1 Elite release.

PR #5652, opened 2020-03-06
Fix #5288: Dock reimplemented using CSS

### LINKED ISSUE #5288: Dock: Reimplement using CSS
Based on this original suggestion: https://github.com/primefaces/primefaces/issues/3153

Problems with current dock:
- Uses third party Fisheye JS which has not been updated since 2009
- Does not look quite like the modern Max OSX dock
- Uses Javascript for all the animation

The suggestion to use pure CSS with a little JS is a good one and I have settled on this really well done CSS that can be modified heavily to fit PF requirements.
http://www.alessioatzeni.com/mac-osx-lion-css3/

melloware:
- Adds RTL support
- Removes all old Fisheye JS library
- Add animation bounce to the icon click just like OSX
- Looks more like modern OSX
- Tested in IE11, Edge, Chrome and Firefox


melloware:
Migration guide for 9.0 has been updated.

PR #5655, opened 2020-03-06
Fix #5654: ClamAV Daemon virus scanner implementation

### LINKED ISSUE #5654: VirusScanner: ClamAV Daemon implementation
A ClamD implementation of the virus scanner which scans your file in real time and reports viruses.

https://linux.die.net/man/8/clamd

It requires a ClamAV Daemon service running somewhere on your network and configured with 4 context params in `web.xml`.

**primefaces.virusscan.CLAMAV_HOST**
The IP Address of the machine where ClamAV is running. example 127.0.0.1

**primefaces.virusscan.CLAMAV_PORT**
The Port of the machine where ClamAV is running. Example 3310

**primefaces.virusscan.CLAMAV_TIMEOUT**
The socket timeout (in ms) for a call to ClamAV before it fails

**primefaces.virusscan.CLAMAV_BUFFER**
The buffer or chunk size (in bytes) for a call to ClamAV. Can be tweaked depending on size of your files.

Error message can be customizes in Messages.properties.
```
primefaces.fileupload.CLAM_AV_FILE = ClamAV has identified file '{0}' as potentially dangerous with message '{1}'
```

melloware:
Comments and suggestions welcome!


melloware:
I will fix CheckStyle issues. Weird that my Eclipse didn't warn me of them it usually does.

Rapster:
Looks good, but
* Would be better if client was not intanciate every time
* `US_ASCII`? Why not use UTF_8?

melloware:
US_ASCII was how the original code we used was and when I switched it to UTF-8 I think it did not work. 
See: https://github.com/solita/clamav-java/blob/master/src/main/java/fi/solita/clamav/ClamAVClient.java

I think it must be the clamd daemon encoding.

melloware:
Ok I cache the Client upon first creating it now.

PR #5657, opened 2020-03-07
Fix #5656: BaseCalendarRenderer improve reflection of Lists

### LINKED ISSUE #5656: DatePicker: Range not set when specified in abstract class
## 1) Environment
- PrimeFaces version: 8.0
- Application server + version: Payara 5.181
- Affected browsers: Firefox, Chrome, ...

## 2) Expected behavior

For multiple statistics views I have an abstract base class that handles the valid year range, the maximum date and the selected date range.
When a users selects a date range, then the setter should be called after selection of the start date **and after selection of the end date**.

## 3) Actual behavior

When a users selects a date range, then the setter is **only called after the selection of the start date**.

## 4) Steps to reproduce

See sample view and bean.

## 5) Sample XHTML

```
<p:datePicker id="dateDayRange"
              maxdate="#{testAbstractBacking.dayRangeMax}"
              monthNavigator="true"
              readonlyInput="true"
              selectionMode="range"
              size="16"
              showButtonBar="true"
              showIcon="true"
              value="#{testAbstractBacking.dayRange}"
              yearNavigator="true" yearRange="#{testAbstractBacking.yearRange}">
    <p:ajax event="dateSelect" />
</p:datePicker>
```

## 6) Sample bean

```
public abstract class AbstractTest
        implements Serializable {

    private String yearRange;

    private LocalDate dayRangeMax;

    private List<LocalDate> dayRange;

    @PostConstruct
    public void init() {
        // initialize the default year range
        final LocalDate localDateNow = LocalDate.now();
        yearRange = localDateNow.minusYears(10).getYear() + ":" + localDateNow.getYear();

        // set the current date as maximum date
        dayRangeMax = LocalDate.now();
    }

    public String getYearRange() {
        return yearRange;
    }

    public LocalDate getDayRangeMax() {
        return dayRangeMax;
    }

    public List<LocalDate> getDayRange() {
        return dayRange;
    }

    public void setDayRange(List<LocalDate> dayRange) {
        this.dayRange = dayRange;

        System.out.println("AbstractTest.setDayRange() -> " + dayRange);
    }

}

@Named
@ViewScoped
public class TestAbstractBacking
        extends AbstractTest {

    @Override
    public void setDayRange(List<LocalDate> dayRange) {
        System.out.println("TestAbstractBacking.setDayRange()");

        super.setDayRange(dayRange);
    }

}
```

melloware:
@christophs78 @tandraschko would like your eyes on this.  Had to test the main class if it had the Declared Field and then walk the superclass until it is found or not found.

I didn't want to modify LangUtils.getUnproxiedClass as that is from DeltaSpike and looks like its doing the right thing.

### DIFF
--- a/src/main/java/org/primefaces/component/calendar/BaseCalendarRenderer.java
+++ b/src/main/java/org/primefaces/component/calendar/BaseCalendarRenderer.java
@@ -309,14 +309,18 @@ public abstract class BaseCalendarRenderer extends InputRenderer {
             Object base = valueReference.getBase();
             Object property = valueReference.getProperty();
 
-            try {
-                Field field = LangUtils.getUnproxiedClass(base.getClass()).getDeclaredField((String) property);
-                ParameterizedType parameterizedType = (ParameterizedType) field.getGenericType();
-                Type listType = parameterizedType.getActualTypeArguments()[0];
-                type = Class.forName(listType.getTypeName());
-            }
-            catch (ReflectiveOperationException ex) {
-                //NOOP
+            Class<?> unproxiedClass = LangUtils.getUnproxiedClass(base.getClass());
+            while (unproxiedClass != null) {
+                try {
+                    Field field = unproxiedClass.getDeclaredField((String) property);
+                    ParameterizedType parameterizedType = (ParameterizedType) field.getGenericType();
+                    Type listType = parameterizedType.getActualTypeArguments()[0];
+                    type = Class.forName(listType.getTypeName());
+                    unproxiedClass = null;
+                }
+                catch (ReflectiveOperationException ex) {
+                    unproxiedClass = unproxiedClass.getSuperclass();
+                }
             }
         }
 


tandraschko:
Can't we assume that the assigned "value" VE is always the same type as the generic List type?
We would avoid reflection completely and only rely on the EL API

melloware:
That is a good question.  Not sure who wrote this code originally whether it was @christophs78 or @Rapster but your suggestion does sound better.

tandraschko:
it was @christophs78 but i suggested to do it like this :D
the code itself is ok but we should avoid reflection if possible.
Not sure if we can assume that "value" type is always the same as the "list" type? WDTY christoph?

christophs78:
Maybe i think a bit slow after a long day with my kids which started before 6am or i miss some facts about EL API...

VE looks like this:
grafik

We get the following type:
grafik

I totally agree avoiding reflections would be the prefered way. But don´t know how to figure out the type of this list without this Reflection-"magic".

Thomas - Do you think into an other direction?
We try to figure out the Type with `BaseCalendarRenderer#resolveDateType` before we convert the string-value which comes from the client to the proper type. At this point in time we don´t have necessary a value in the bound property from the bean. (eg the bound property from the bean is a empty list)

christophs78:
Whatever we do, IMO we should add some UnitTests for this. We put some effort over the last year into `DatePickerTest`. 

At some point we had the following test:
```
@Test
@Ignore
public void resolveDateType_ListLocalDate() {
	/*
		Unable to mock this with Mockito only.
		Cause:
		BaseCalendarRenderer#resolveDateType calls a static method.
		(ValueExpressionAnalyzer.getReference)
		It would require https://github.com/powermock/powermock to mock this static method.
	 */

	setupValues(List.class, Locale.ENGLISH);
	when(valueExpression.getValueReference(elContext)).thenReturn(null);
	Class type = renderer.resolveDateType(context, datePicker);
	assertEquals(LocalDate.class, type);
}
```
When i think about this today we should refactor `BaseCalendarRenderer#resolveDateType` in a way to make the "list-thing" testable. Do some "separation of concerns". Extract the code which figures out the type based on base(class) and property (field) and write some test´s for this.
It´s fine for me to add this refactoring and tests myself after this PR is merged.

tandraschko:
+1 for unit Testing

Nvm about my last comment, we already try to check the "value" Attribute first

melloware:
OK merged this fix as it fixes the presented test case.  @christophs78 appreciate any unit tests you decide to add!

PR #5658, opened 2020-03-08
#5656: DatePicker: List-value-binding; Refactoring & Tests

christophs78:


melloware (approved):
Nice refactor and tests

PR #5660, opened 2020-03-08
Fix #5659: IdleMonitor onactive called with widget

### LINKED ISSUE #5659: IdleMonitor: Change this context of onactive callback
Low priority. I came across this while adding types and docs for #5027 (yes, still working on this, got some time on my hands now and should be done soon).

The IdleMonitor widget [has got two callback the user can set](https://primefaces.github.io/primefaces/7_0/#/components/idlemonitor?id=attributes): `onactive` and `onidle`.

`onidle` is called with `this` [set to the widget instance](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/idlemonitor/1-idlemonitor.js#L14)

But for `onactive`, it is [set to the document element](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/idlemonitor/1-idlemonitor.js#L21).

That seems inconsistent, so I think that line 21 in `1-idlemonitor.js` should probably changed to:

```javascript
$this.cfg.onactive.call($this);
```

blutorange:
* IdleMonitor: Change the this context of the onactive callback to the widget instance, see #5659 
* Feel free to decline this pull request in case you feel that this change cannot be done due to backwards compatibility (or wait until PF 8.0)

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/idlemonitor/1-idlemonitor.js
+++ b/src/main/resources/META-INF/resources/primefaces/idlemonitor/1-idlemonitor.js
@@ -18,7 +18,7 @@ PrimeFaces.widget.IdleMonitor = PrimeFaces.widget.BaseWidget.extend({
         })
         .on("active.idleTimer" + this.cfg.id, function(){
             if($this.cfg.onactive) {
-                $this.cfg.onactive.call(this);
+                $this.cfg.onactive.call($this);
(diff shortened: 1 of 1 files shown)


melloware:
PF 8.0 is already released so i think this fix is safe.  I agree with your change.

PR #5666, opened 2020-03-09
Fix #5661: Repeat backward compatible with JSF2.2

### LINKED ISSUE #5661: Repeat: exception
## 1) Environment
- PrimeFaces version:
- Does it work on the newest released PrimeFaces version? Version? 8.0
- Affected browsers:

## 2) Expected behavior
click on button should not fire exception 
version 7.0 is ok

## 3) Actual behavior

an exception ocurred

java.lang.NoSuchMethodError: javax.faces.event.FacesEvent.getFacesContext()Ljavax/faces/context/FacesContext;
	

## 4) Steps to reproduce

click on any button

sample project: https://github.com/erickdeoliveiraleal/primefaces-test

melloware:
Tested this fix with the reproducer and its working. Shoul be a safe Fix since FacesEvent.getFacesContext() says it returns the target context or if NULL returns `FacesContext.getCurrentInstance()`.

### DIFF
--- a/src/main/java/org/primefaces/component/repeat/UIRepeat.java
+++ b/src/main/java/org/primefaces/component/repeat/UIRepeat.java
@@ -913,7 +913,7 @@ public class UIRepeat extends UINamingContainer {
             this.resetDataModel();
             int prevIndex = this.index;
             FacesEvent target = idxEvent.getTarget();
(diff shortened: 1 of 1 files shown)


tandraschko:
Plaese see my commit; UIComponentBase already has getFacesContext, which is a cached on in MyFaces

PR #5671, opened 2020-03-10
Fix #5670: CSP fix for Google reCaptcha

### LINKED ISSUE #5670: CSP: Google Recaptcha violation
## 1) Environment
- PrimeFaces version: 8.0
- Does it work on the newest released Prime Faces version? NO
- Application server + version: ALL
- Affected browsers: ALL

When PrimeFaces CSP is enabled Google reCaptcha is failing.

There is a 3 part fix.

1. Must update captcha.js to handle the NONCE value.

2. Must use the custom policy in your web.xml.
```xml
    <context-param>
    </context-param>
    <context-param>
        <param-value>script-src 'self' https: *.googleapis.com</param-value>
    </context-param>
```

3. Docs should be updated under the CSP page and Recaptcha page to document this.

melloware:
Updated 8.0 and 9.0 docs as this should be merged into 8.0.1 Elite.

### DIFF
--- a/docs/8_0/components/captcha.md
+++ b/docs/8_0/components/captcha.md
@@ -79,6 +79,16 @@ message bundle mechanism. Corresponding keys are;
 | Summary | primefaces.captcha.INVALID |
 | Detail | primefaces.captcha.INVALID_detail |
 
+## Content Security Policy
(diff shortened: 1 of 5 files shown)


tandraschko:
+1

PR #5672, opened 2020-03-10
Fix #5645: Docu: make sure up-to-date versions are delivered to the client

### LINKED ISSUE #5645: Docu: make sure up-to-date versions are delivered to the client
Due to this (old versions) of the documentation are loaded from "Cache Storage".

grafik
grafik

When we update documentation users still have the old versions in their "Cache Storage". Only pushing Ctrl + F5 (or cleanup all Browser-Cache) helps.

When we look at https://docsify.js.org/#/?id=docsify itself does not use a serviceworker.

There´s a docsify-issue: https://github.com/docsifyjs/docsify/issues/190
It say´s we should remove the serviceworker because it´s not needed anymore.

christophs78:
... and some minor cleanup

Browser online:
I modified the ServiceWorker by delaying the load (of the perhaps outdated version) from Cache Storage by 500ms. So the fetch of die up-to-date-version of the docu has higher priority.

Browser offline:
We only look into the Cache Storage.

Notes:
- We have to look how "fast" the browser updates sw.js. Maybe it takes a day or two. (Should not be an issue.)
- Serviceworker seems to fetch content only once within a "session". When a docu-file is modified after Serviceworker fetched it once within the "session" (browser-tab) it does not refetch this file.

### DIFF
--- a/docs/7_0/index.html
+++ b/docs/7_0/index.html
@@ -33,9 +33,6 @@
       plugins: [
         function(hook, vm) {
           hook.ready(function () {
-            console.log("Mein Docsify-Plugin; Init");
-
-            //let el = document.querySelector("h1.app-name");
             let h1Elt = document.querySelector("main aside h1.app-name");
             let versionElt = document.createElement("div");
             versionElt.setAttribute("class", "PF-version")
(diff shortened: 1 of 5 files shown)


melloware (approved):

melloware:
Looks good to me let me know if you want me to merge it so we can start testing it out in the real world.

christophs78:
IMO ready to be merged. Curious how this works in real world. :-)

PR #5686, opened 2020-03-12
Fix #5679: TextEditor attribute disabled + widget methods

### LINKED ISSUE #5679: TextEditor: disabled feature
## 1) Environment
- PrimeFaces version: 7.0
- Not present in GitHub 9.0-SNAPSHOT

## 2) Expected behavior
It would be important to have "disabled" attribute for textEditor, because it is the userfriendly way to implement that the component won't send its input to the server. (it is not the "readonly" usecase)

## 3) Actual behavior
"disabled" feature is general within all the other inputs already. In case of textEditor it seems to be missing.

Workaround can be something like this, but its not clean and cumbersome:
$("#CLIENT_ID").attr("disabled", true).addClass('ui-state-disabled');
$("#CLIENT_ID input").attr("disabled", true).addClass('ui-state-disabled');

melloware:


tandraschko:
+1

Issue #5690, opened 2020-03-13
Schedule: renderingMode must be lowercase


OBreidenbach:
## 1) Environment
- PrimeFaces version: 8.0
- Does it work on the newest released PrimeFaces version? Version? No
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) No
- Application server + version: does not matter
- Affected browsers: does not matter

## 3) Actual behavior

The full calendar is expecting the values for 'rendering' in lower case. Right now it is rendered in upper cases. It seems as org.json changed writing of enums from toString() to name() method.
And maybe full calender v4 is pickier than v3, do not now.

## 4) Steps to reproduce
Add an event with rendering mode  BACKGROUND and look how it is rendered. Upper case BACKGROUND is ignored, lower case works fine.

PR will follow.

OBreidenbach:
As workaround you can use:
 .dynamicProperty("rendering", ScheduleRenderingMode.BACKGROUND.getRendering())
on DefaultScheduleEvent.builder()

PR #5691, opened 2020-03-13
#5690 use lower case for value of full calendar event.rendering

OBreidenbach:
#5690 use lower case for value of full calendar event.rendering

### DIFF
--- a/src/main/java/org/primefaces/component/schedule/ScheduleRenderer.java
+++ b/src/main/java/org/primefaces/component/schedule/ScheduleRenderer.java
@@ -41,6 +41,7 @@ import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.util.Map;
+import java.util.Objects;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/primefaces/component/schedule/ScheduleRenderer.java
```diff
@@ -118,7 +119,7 @@ protected void encodeEventsAsJSON(FacesContext context, Schedule schedule, Sched
                 jsonObject.put("className", event.getStyleClass());
                 jsonObject.put("description", event.getDescription());
                 jsonObject.put("url", event.getUrl());
-                jsonObject.put("rendering", event.getRenderingMode());
+                jsonObject.put("rendering", Objects.toString(event.getRenderingMode(), null));
```
melloware, line 122: I see you are doing Objects.toString but not .toLowerCase()?

melloware (changes requested):

src/main/java/org/primefaces/component/schedule/ScheduleRenderer.java
OBreidenbach: That because in PrimeFaces it is an enum following java naming conventions. The ScheduleRenderingMode enum have internally the correct string for the full calendar. 
.toLowerCase() would not work for ScheduleRenderingMode.INVERSE_BACKGROUND because it must resolve to 'inverse-background'.
You must call getRendering() or toString() on the enum. To be null save I use Objects.toString()

The behavior of org.json changed. In the old version it did a .toString() on all untested object types. Now it tests for enums and calls .name() on them. So now longer the internal string is used but the name.

melloware (approved):

PR #5699, opened 2020-03-16
Fix #5698: LinkButton focus styles, enable/disable widget methods

### LINKED ISSUE #5698: LinkButton: Focus styles
When navigating using a keyboard, p:linkButton do not show focus styles (because the focus is bound to link and not wrapper span element).

melloware:


tandraschko:
+1

please just correct the codestyle (init :) to (init:)

melloware:
Fixed all the methods to method: thanks for pointing that out.

PR #5715, opened 2020-03-20
Fix #5713: DatePicker Add showWeek and weekCalculator

### LINKED ISSUE #1052: Calendar: Issue with show calendar weeks
In the attachment are two files with the displayed calendar weeks. One calendar shows the weeks in the locale "de" (German). Everything is ok. Another calendar shows the weeks in the locale "en" (English). The weeks are not correct. I know, there is a small difference in the weeks for different locales. See the good explanation here: http://www.calendar-week.org/ But the weeks in "en" (also "en_US", "en_UK") are shown wrong.
kalenderwochen_de
kalenderwochen_en

### LINKED ISSUE #5713: DatePicker: Add showWeek option like Calendar
## 1) Environment
- PrimeFaces version: 8.0

## 2) Description
`p:calendar` has the options `showWeek` and `calculateWeek`. This will add an additional column infront of each week containing the week number.

This should be supported by the new `p:datePicker`.

PR is on the way

fanste:
**Changes as of 2020-03-23:**
 - only a single default implementation is used for all possible `start of week` (`options.locale.firstDay` / `options.locale.firstDayOfWeek`)
 - additionally this implementation is based on the locale option `firstDayWeekOffset` which will be guessed based on `start of week` if not provided.
 - it will fix #1052

------------

`weekCalculator` internally defaults to one of three different implemetation.

- start of week = monday: calculateWeekNumber_iso8601 (like the jQuery based calendar)
- start of week = sunday: calculateWeekNumber_sundayFirst
- start of week = saturday: calculateWeekNumber_saturdayFirst

jQuery based calendar calculated wrong week numbers for sow=sunday|saturday

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
+++ b/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
@@ -16,13 +16,6 @@
         factory(jQuery);
     }
 }(function ($) {
-
-    function calculateWeekNumber(d) {
-        var yearStart = new Date(Date.UTC(d.getUTCFullYear(),0,1));
-        var weekNo = Math.ceil(( ( (d - yearStart) / 86400000) + 1)/7);
-        return weekNo;
-    }
-
     $.widget("prime.datePicker", {
 
         options: {
@@ -118,19 +111,31 @@
                 $.extend(this.options.locale, this.options.userLocale);
             }
 
-            if(this.options.showWeek && !this.options.weekCalculator) {
+            if (this.options.showWeek && !this.options.weekCalculator) {
+                this.options.weekCalculator = this.calculateWeekNumber.bind(this);
+
+                // initialize the potentially missing firstDayWeekOffset option
+                // based on the firstDay(OfWeek) option:
+                //  - firstDay = saturday => firstDayWeekOffset = 12
+                //  - firstDay = sunday => firstDayWeekOffset = 6
+                //  - firstDay = monday (ISO8601) => firstDayWeekOffset = 4
+                //
+                // those defaults are based on the locales on the library
+                // moment.js and won't be correct for _all_ locales.
                 var sundayIndex = this.getSundayIndex();
-                if(sundayIndex == 0) {
-                    this.options.weekCalculator = this.calculateWeekNumber_sundayFirst;
-                }
-                else if(sundayIndex == 1) {
-                    this.options.weekCalculator = this.calculateWeekNumber_saturdayFirst;
-                }
-                else if(sundayIndex == 6) {
-                    this.options.weekCalculator = this.calculateWeekNumber_iso8601;
-                }
-                else {
-                    this.options.showWeek = false;
+                if (this.options.locale.firstDayWeekOffset === undefined) {
+                    if (sundayIndex == 0) {
+                        this.options.locale.firstDayWeekOffset = 6;
+                    }
+                    else if (sundayIndex == 1) {
+                        this.options.locale.firstDayWeekOffset = 12;
+                    }
+                    else if (sundayIndex == 6) {
+                        this.options.locale.firstDayWeekOffset = 4;
+                    }
+                    else {
+                        this.options.showWeek = false;
+                    }
                 }
             }
 
@@ -1309,10 +1314,10 @@
 
             if(this.options.showWeek) {
                 dayNamesHtml += '<th scope="col">' +
-	                '<span>' +
-	                this.options.locale.weekHeader +
-	                '</span>' +
-	                '</th>';
+                    '<span>' +
+                    this.options.locale.weekHeader +
+                    '</span>' +
+                    '</th>';
             }
 
             for (var i = 0; i < weekDaysMin.length; i++) {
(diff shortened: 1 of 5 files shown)


melloware:
Looks like you have a merge conflict.

@christophs78 can you review?

src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -109,6 +111,22 @@
                 $.extend(this.options.locale, this.options.userLocale);
             }
 
+            if(this.options.showWeek && !this.options.weekCalculator) {
+                var sundayIndex = this.getSundayIndex();
+                if(sundayIndex == 0) {
+                    this.options.weekCalculator = this.calculateWeekNumber_sundayFirst;
+            	}
+                else if(sundayIndex == 1) {
```
melloware: Brace indent issue. on line 118

src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -1292,9 +1319,42 @@
...
+        },
+
+        calculateWeekNumber_saturdayFirst: function (d) {
+        	d = new Date(Date.UTC(d.year, d.month, d.day));
+            d.setUTCDate(d.getUTCDate() - ((d.getUTCDay() + 6) % 7));
+            var yearStart = new Date(Date.UTC(d.getUTCFullYear(),0,1));
+            var weekNo = Math.ceil(( ( (d - yearStart) / 86400000) + 1)/7);
+            return weekNo;
+        },
+
```
melloware, line 1379: Feel like there has to be a way to refactor the above 3 methods as there is a lot of duplicate code expecially around yearStart and WeekNo.

melloware (changes requested):

src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
fanste: Fixed with the next commit

christophs78:
> @christophs78 can you review?

Yes

src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
fanste: I thought about it, but the only solution I could think of seemed somehow wrong. Anyway, I refactored it.

melloware (approved):

src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -17,6 +17,12 @@
     }
 }(function ($) {
 
+    function calculateWeekNumber(d) {
+        var yearStart = new Date(Date.UTC(d.getUTCFullYear(),0,1));
+        var weekNo = Math.ceil(( ( (d - yearStart) / 86400000) + 1)/7);
+        return weekNo;
+    }
```
melloware: Why not put this method right below the 3 methods using it?

christophs78 (changes requested):
Looks good! Thanks for your work.

Two minor issues:
- Please add the new attributes to primefaces-p.taglib.xml
- Please also do a Showcase-PR which add´s this new features to https://www.primefaces.org/showcase/ui/input/datepicker/datePickerJava8.xhtml.

Should we add some CSS? (Or should this be added by @mertsincan?)
PrimeNG (https://primefaces.org/primeng/showcase/#/calendar) styles it like this:
grafik

melloware:
I think we should add what we can to datepicker.css for generic styling and yeah we probably need to leave the rest up to the theme.

christophs78:
We should add this:
grafik

src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -1293,9 +1326,36 @@
...
+        },
+
         renderWeek: function (weekDates) {
             var weekHtml = '';
 
+            if(this.options.showWeek) {
+                var date = weekDates[0],
+                    cellClass = date.otherMonth && !this.options.showOtherMonths ? ' ui-datepicker-other-month-hidden' : '';
+
+                weekHtml += '<td class="ui-datepicker-week-col' + cellClass + '">' +
```
christophs78: please add `ui-state-disabled`

src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
fanste: I don't think that it would be a good idea to invoke the calculation method with the `this` scope. So I can't add the common method to the widget because then I have to invoke it with `this`.

At least that is my understanding, but I have to test it.

melloware: yeah and if you consider it a private method of the widget you can name it with an underscore so `_calculateWeekNumber` that is how other widgets do methods like these as internal helper methods..

src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
fanste: I have implemeted it the same ways a jquery did it. What is your intention adding this CSS class? When I add `ui-state-disabled`, selecting of the week number is still allowed but it is/may be rendered nearly unreadable (due to some disabled styling like opacity)

christophs78: IMO we should use PrimeNG and PrimeReact as reference.
https://primefaces.org/primeng/showcase/#/calendar
https://primefaces.org/primereact/showcase/#/calendar
Both use `ui-state-disabled`. 

As far as i know PrimeTek uses the (almost) same base SASS for all themes independent of technology. So it would make sense to render the same code like PrimeNG/PrimeReact/PrimeVue.
@mertsincan: Do you want to add something?

mertsincan: Hi,

DOM structure can be like; 
```
<td class="ui-datepicker-weeknumber">
     <span class="ui-state-disabled">13</span>
</td>
```

src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
mertsincan: Also, why UTC? could you please remove UTC format?

https://github.com/primefaces/primereact/blob/master/src/components/calendar/Calendar.js#L1586
```
getWeekNumber(date) {
        let checkDate = new Date(date.getTime());
        checkDate.setDate(checkDate.getDate() + 4 - ( checkDate.getDay() || 7 ));
        let time = checkDate.getTime();
        checkDate.setMonth( 0 );
        checkDate.setDate( 1 );
        return Math.floor( Math.round((time - checkDate.getTime()) / 86400000 ) / 7 ) + 1;
    }
```

src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -110,6 +112,22 @@
                 $.extend(this.options.locale, this.options.userLocale);
             }
 
+            if(this.options.showWeek && !this.options.weekCalculator) {
```
mertsincan: I think you can remove this if statement. Users can calculate week numbers using a custom weekCalculator function;
```
function myWeekCalculator(date) {
    ...
    return weekNo;
}

<p:datePicker weekCalculator="myWeekCalculator"
```

src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
fanste: OK, I will change the DOM structure as requested.

src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
fanste: No I can't remove it. This if checks if a custom week calculator is provided or if some internal defaults have to be configured. But only if `showWeek` is enabled. If the user provides a custom week calculator I don't have to check if the internals are correctly configured because it is totally up to the user.

src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
fanste: I used UTC because there are multiple reports on the web (e.g. [SO, edit from 2017](https://stackoverflow.com/a/6117889/1741604)) that not using UTC may result in incorrect weeks. Additionally I refactored the week calculation a bit based on code from moment.js (which also uses UTC) to really fix non-ISO8601 calendar views (and issue #1052).

fanste:
I've updated the PR message to reflect to current implementation

melloware (approved):

christophs78 (approved):

melloware:
@mertsincan can you do a final review on this one I would like to merge it.

Issue #5717, opened 2020-03-22
DatePicker: smart behaviour for showTime and timeOnly-attributes with value-binding to LocalDate/LocalDateTime/LocalTime


christophs78:
Base idea by @tuerker as part of https://github.com/primefaces/primefaces/issues/5638.

- LocalDate --> showTime=false, timeOnly=false
- LocalDateTime --> showTime=true, timeOnly=false
- LocalTime --> showTime=false, timeOnly=true

Explicit set showTime / timeOnly-attributes override the smart behaviour.

kukel:
Very important feature for being able to use the component in a custom datatable filter (I think).  Was just going to look into something similar since I cannot get a time to be processed on the server side (time is passed from the client to the server and ignored there. Always get a `LocalDate` instead of `LocalDateTime`

PR #5718, opened 2020-03-22
#5717: DatePicker: smart behaviour for showTime and timeOnly-attributes with value-binding to LocalDate/LocalDateTime/LocalTime

christophs78:
Still need to do some QA (and maybe refactoring), but already looks good.

Showcase-PR: https://github.com/primefaces/primefaces-showcase/pull/128

### DIFF
--- a/src/main/java/org/primefaces/component/api/UICalendar.java
+++ b/src/main/java/org/primefaces/component/api/UICalendar.java
@@ -23,16 +23,19 @@
  */
 package org.primefaces.component.api;
 
+import java.time.LocalTime;
 import java.time.chrono.IsoChronology;
 import java.time.format.DateTimeFormatterBuilder;
 import java.time.format.FormatStyle;
 import java.time.format.ResolverStyle;
 import java.util.Locale;
 
+import javax.el.ValueExpression;
 import javax.faces.application.FacesMessage;
 import javax.faces.component.html.HtmlInputText;
 import javax.faces.context.FacesContext;
 
+import org.primefaces.component.datepicker.DatePickerBase;
 import org.primefaces.util.CalendarUtils;
 import org.primefaces.util.LocaleUtils;
 import org.primefaces.util.MessageFactory;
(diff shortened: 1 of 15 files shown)


src/main/java/org/primefaces/component/api/UICalendar.java
```diff
@@ -110,6 +112,23 @@ public boolean isTimeOnly() {
         return (Boolean) getStateHelper().eval(PropertyKeys.timeOnly, false);
     }
 
+    public Boolean isTimeOnlySmart(FacesContext context) {
+        Boolean timeOnly = (Boolean) getStateHelper().eval(PropertyKeys.timeOnly);
```
melloware, line 116: Instead of "Smart" maybe isTimeOnlyDetected() is a better name?

src/main/java/org/primefaces/component/datepicker/DatePickerBase.java
```diff
@@ -213,6 +217,23 @@ public boolean isShowTime() {
         return (Boolean) getStateHelper().eval(PropertyKeys.showTime, false);
     }
 
+    public Boolean isShowTimeSmart(FacesContext context) {
+        Boolean showTime = (Boolean) getStateHelper().eval(PropertyKeys.showTime);
```
melloware, line 221: Detected again instead of Smart?

melloware (changes requested):

tandraschko:
1. Im not sure about the "Detect" suffix
2. Shouldn't we pass the Value type to the "detect" methods? (we already extract the value type, which also check the list generic type e.g.)

christophs78:
>     2. Shouldn't we pass the Value type to the "detect" methods? (we already extract the value type, which also check the list generic type e.g.)

For some calls we already know the value type, for other not:
grafik

christophs78:
I´ll try a second approach where we call `DatePickerBase#setShowTime` from `DatePickerRenderer#encodeScript`. (Only when showTime ist not explicit set.) Hope this works and makes it easier. With the current approach i dislike passing FaceContext arround.

PR #5720, opened 2020-03-23
Fix #5579: Spinner prevent bad input, default decimalPlaces to 0 for certain types

### LINKED ISSUE #5579: Spinner: should not allow to enter non-numeric values

## 1) Environment
- PrimeFaces version: 8.0 RC3

## 2) Expected behavior

## 3) Actual behavior

## 4) Steps to reproduce

and enter the above values.

## 5) Sample XHTML

..

## 6) Sample bean

..

melloware:
Addresses both issues listed in the ticket.


tandraschko:
+1

PR #5721, opened 2020-03-23
Fix #5712: CSP wrap register script in document.ready

### LINKED ISSUE #5712: CSP: onclick events fail on Premium Themes
CspResponseWriter should wrap javascript in $(function() {});
## 1) Environment
- PrimeFaces version: 8.0
- Does it work on the newest released PrimeFaces version? Version? No, 8.0
- Application server + version: JBoss EAP 7.1
- Affected browsers: Firefox, Chrome were tested and failed in some instances

## 2) Expected behavior
Events work after implementing CSP

## 3) Actual behavior
Events don't work after implementing CSP

## 4) Steps to reproduce
The left slide out menu in Barcelona doesn't fire events

## 5) Sample XHTML
<pa:tabMenu>
                <pa:tab icon="fa fa-home" title="Home">
                    <pa:menu>
                        <p:menuitem value="Change Something" icon="fa fa-exchange" action="#{bean.change}"
                                    ajax="false" onclick="loadingStart()"/>
                    </pa:menu>
                </pa:tab>
...

## 6) Sample bean
public String change() {
		return Pages.HOME + Pages.REDIRECT_SUFFIX;
	}

The CSP JavaScript should be wrapped in $(function() { ... });  If the JavaScript inside of the CSP block is run again then it starts to work.

melloware:


tandraschko:
can you explain why it happens?
in general it should not be required as it doesn't execute script, it just attaches the dom-handlers like onclick?

melloware:
I cannot explain why it happens.  When I run his test case its definitely broken. I had not seen this use case in ANY of my testing or production apps.  But his example code definitely shows a potential problem.  Any thoughts?

melloware:
OK looking at the HTML using 8.0 here is what happens..

Barca menu renders with document.ready
```javascript
$(function(){PrimeFaces.cw("BarcelonaMenu","widget_j_idt15_j_idt18",{id:"j_idt15:j_idt18"});});
```

Register does not...
```javascript
PrimeFaces.csp.register('a\-5554d1af\-32b0\-4c81\-a921\-37c9a30f5326','onclick',function(event){PF('block').show();PF('loading').getJQ().show();
```
So is it that the createWidget of barca is actually creating some DOM on the fly and since it hasn't run yet there is nothing to attach the onClick to?

tandraschko:
Doesnt make sense for me :D
The CSP ResponseWriter only catches real "onclick" rendering, so the element MUST be there already, otherwise we woulnt generate the script?!
I think this requires bit more debugging to understand what happens.

We should not apply this change as it required jquery (must not be used on every view) and this may add dom handlers to late for some weird cases.

melloware:
Agreed.  In the bug ticket the last ZIP file in the thread is the one to use to debug and see the issue with CSP true and false.

tandraschko:
i dont have enough time currently to work on such issues, so if you can collect further infos, it would be great

melloware:
OK @tandraschko @william-shuman I have figured out what is going on.

**In Order:**
1. Barcelona Menu is wrapped in `document.ready`

2. CSP.register of the `onclick` is generated normally and executes moving the onclick to a Jquery `on` event.

3. Barcelona Menu now runs and executes the following code:
```javascript
this.menulinks.off("click").on("click", function(f) {
```

So the Barca Menu is actually removing the `click` listener the CSP just registered and therefore nothing happens.   That is why MOVE_SCRIPTS_TO_BOTTOM fixes it as everything runs after the DOM and the order is Barca Menu then the CSP.register.

tandraschko:
hmmm, sounds quite weird :D

but thats only releated to barcelona menu, not the PF core menus?

melloware:
Confirmed its in this Barcelona and probably most other premium themes menu's.  None of the core primefaces menu code calls `off('click')` anywhere which is why I would have seen this in my testing.

tandraschko:
so:
1) the MenuRenderer renders the onclick from the user on the link 
2) the layout.js removes the old click and registers a new click

why do we render 1)? :D

melloware:
I am not positive but I think in normal circumstances Barca layout.js removes onclick and sets up its onclick and then custom onclick form the menu gets applied after. So there are two Jquery registered onclick events on 1 menu option. 1 registered by Barca and 1 by the MenuItem itself.  Thus with CSP this ordering is thrown off.

melloware:
if Barca didn't have that `off` statement this would all just work period I think.

tandraschko:
not sure, see:

`
	
        this.menulinks.off('click').on('click', function(e) {
            var link = $(this),
            item = link.parent(),
            submenu = item.children('ul');

            if(item.hasClass('active-menuitem')) {
                if(submenu.length) {
                    $this.removeMenuitem(item.attr('id'));
                    item.removeClass('active-menuitem');
                    submenu.slideUp();
                }
            }

            else {
                $this.addMenuitem(item.attr('id'));
                $this.deactivateItems(item.siblings(), true);
                $this.activate(item);
            }

            setTimeout(function() {
                $this.nano.nanoScroller();
            }, 500);

            if(submenu.length) {
                e.preventDefault();
            }
        });
`

melloware:
Ahh I see.

tandraschko:
I will close this PR for now as its not a solution IMO and we didn't found whats really wrong here.

@melloware would be great if you can dig deeper but lets discuss possible solutions in the ticket then.

Issue #3720, opened 2018-05-23
Messages: duplicate value on summary and detail if using requiredMesssage attribute


mafaul:
I like to display details using the p:messages component to give users additional info when issuing faces messages programmatically. But with the showDetail set to true, the detail gets duplicated from the suummary when using the requiredMessage attribute.

When using <p:inputText required="true" requiredMessage="Value Required" ... /> the detail is the same as the summary.

This is also the case if the value of the inputfield value is binded to a JPA entity field annotated with @NotNull.

Primefaces Version: 6.2
Affects: Messages, Message, Growl

https://stackoverflow.com/questions/13467881/duplicate-text-in-pmessages-summary-and-detail

![image](https://user-images.githubusercontent.com/12007706/40451766-fe79c8e6-5edf-11e8-9a0e-0fe2306aac5f.png)

mafaul:
https://github.com/primefaces/primefaces/pull/3719
https://github.com/primefaces/primefaces/pull/3718
https://github.com/primefaces/primefaces/pull/3717

tandraschko:
I understand your case but thats not the root problem. You coudl still overwrite the detail message.

But we can introduce this behavior with a attribute on all message components (message, messages, growl), which says something like "don't display detail if detail equals title". Just need a good name and a new PR ;)

tandraschko:
Also one PR for all components in enough next time.

mafaul:
https://github.com/primefaces/primefaces/pull/3722

tandraschko:
We need a attribute on the component itself (p:message doThisNewLogic=true) ;) But we also need good ideas for the attribute first. @melloware @cnsgithub @Rapster ideas? :D

melloware:
My thoughts are:

1. hideSummaryMatchingTitle=true
2. excludeSummaryMatchingTitle=true

tandraschko:
called it "skipDetailIfEqualsSummary" for now

tandraschko:
See: https://github.com/primefaces/primefaces/commit/1178ed49c94a85c2f74a02512df0a4c933227ee4

PR #5733, opened 2020-03-26
Fix #5731: AutoCompete race condition call clear event

### LINKED ISSUE #5731: AutoComplete: forceSelection causes clear event to fire even when item selected
## 1) Environment
- PrimeFaces version: 8.0
- Does it work on the newest released PrimeFaces version? Version? No, 8.0
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) unknown
- Application server + version: Wildfly 18
- Affected browsers: Firefox 74

## 2) Expected behavior

With forceSelection=true and when both itemSelect and clear events are defined, enter text that matches part of an auto complete search, click one of the entries in the dropdown, and only the itemSelect event fires.

## 3) Actual behavior

The clear event is firing and then the itemSelect event fires.

## 4) Steps to reproduce

See code below. Enter some text in the p:autoComplete, pick any of the results, the clear will fire and then the select event will fire. This seems to be from #3270, where the clear event was never firing when forceSelection was true, but looking at the code, it's comparing what was typed in the p:autoComplete against what was selected, this feels wrong to me. This would require a user to type out the entire matching text for the clear event not to fire.

from autocomplete.js:

```
    isValid: function(value) {
        if(!this.cfg.forceSelection) {
            return;
        }

        var valid = false;

        for(var i = 0; i < this.currentItems.length; i++) {
            var stripedItem = this.currentItems[i];
            if (stripedItem) {
                stripedItem = stripedItem.replace(/\r?\n/g, '');
            }

            if(stripedItem === value) {
                valid = true;
                break;
            }
        }

        if(!valid) {
            this.input.val('');
            if(!this.cfg.multiple) {
                this.hinput.val('');
            }
            this.fireClearEvent();
        }

        return valid;
    },
```

We're migrating from 6.2.6 so this is new to us, we use forceSelection=true basically anywhere we use autoComplete so either we're doing something wrong and we're misusing forceSelection, or the source is too aggressive when matching. Thanks!

## 5) Sample XHTML

```
			<p:autoComplete
				placeholder="Search"
				value="#{testController.name}"
				forceSelection="true"
				completeMethod="#{testController.search}"
				var="name"
				itemLabel="#{name}"
				itemValue="#{name}">
				<p:ajax
					event="itemSelect"
					listener="#{testController.onSelect}" />
				<p:ajax
					event="clear"
					listener="#{testController.onClear}" />
			</p:autoComplete>
```

## 6) Sample bean

```
@Named
@ViewScoped
public class TestController implements Serializable {
	
	private String name;
	
	public List<String> search(String search) {
		return Arrays.asList("Greg", "Matt", "Aaron");
	}
	
	public void onSelect(SelectEvent<String> event) {
		System.out.println("onCustomerSelect()");
	}
	
	public void onClear(AjaxBehaviorEvent event) {
		System.out.println("onCustomerClear()");
	}
	
	public String getName() {
		return name;
	}
	
		this.name = name;
	}
}
```

melloware:
I regression tested #3720 and all is still working and the new ticket is also fixed.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
+++ b/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
@@ -749,6 +749,10 @@ PrimeFaces.widget.AutoComplete = PrimeFaces.widget.BaseWidget.extend({
     },
 
     invokeItemSelectBehavior: function(event, itemValue) {
+        if (this.clearingTimeout) {
+            this.deleteClearTimeout();
+        }
+ 
         if(this.hasBehavior('itemSelect')) {
             var ext = {
                 params : [
@@ -963,8 +967,17 @@ PrimeFaces.widget.AutoComplete = PrimeFaces.widget.BaseWidget.extend({
         this.timeout = null;
     },
 
+    deleteClearTimeout: function() {
+        clearTimeout(this.clearingTimeout);
+        this.clearingTimeout = null;
+    },
+
     fireClearEvent: function() {
-        this.callBehavior('clear');
+        var $this = this;
+        $this.clearingTimeout = setTimeout(function() {
+            $this.clearingTimeout = null;
+            $this.callBehavior('clear');
+        }, 100);
     },
 
     isValid: function(value) {


tandraschko:
would be great if you could add a comment on line 977

melloware:
Found a better fix without using Timer so closing this PR.

PR #5736, opened 2020-03-30
Fix #5611: SelectManyMenu new metaKeySelection attribute

### LINKED ISSUE #5611: SelectManyMenu: showCheckbox=false requires a meta key to work as a select many component
Hi guys! It's my third issue on this component, and it's an improvement request rather than a bug.

## 1) Environment
- PrimeFaces version: 8.0RC

## 2) Expected behavior

As the description from the showcase goes

> `SelectManyMenu` is used to choose multiple items from a list.

Being a simple user, I click on the item, it's selected, click again on it - it's unselected. I don't know what meta keys are, and how to use them. It's a select MANY component, several items should be selected by simple clicks rather than fancy combinations (RMC + command) unknown to the experienced.  

It would be great to achieve the same behaviour as a component with `showCheckbox="true"` has. For the table widget, you have `rowSelectMode`, and `rowSelectMode="add"` (`"new"` as default) is what I am describing. Similarly, `selectMode` would be helpful here (with `"add"` as default).

## 3) Actual behaviour

It behaves as a select one component unless you know the trick. You can't select more than one element by simply clicking.

## 4) Steps to reproduce

1. Create a widget.
2. Click on an item.
3. Click on another item. The previous item is deselected. The only selected is the one you just clicked.

## 5) Sample XHTML

      <p:selectManyMenu id="basic">
            <f:selectItem itemLabel="Option 1" itemValue="1" />
            <f:selectItem itemLabel="Option 2" itemValue="2" />
            <f:selectItem itemLabel="Option 3" itemValue="3" />
        </p:selectManyMenu>

melloware:


tandraschko:
not sure about the name

melloware:
@Rapster @tandraschko @jepsar Any ideas for a better name for this attribute?

jepsar:
There is a typo in https://github.com/primefaces/primefaces/pull/5736/files#diff-16663cabb068a2b59fcd07a9efdeb08dR48

But if the name changes it should be changed anyway. How about `metaMode`?

melloware:
how about `metaSelection` which will be true by default (current behavior) and false if you want to disable it?

tandraschko:
what about metaKeySelection?

melloware:
I like it. Updating PR now.

tandraschko:
+1

PR #5738, opened 2020-03-30
Fix #5584: Reflow must remove column title span from draggable

### LINKED ISSUE #5584: Datatable: Additional Header text shown in a dragged datatable row when reflow="true" is set
[pf-5584.zip](https://github.com/primefaces/primefaces/files/4152578/pf-5584.zip)
## 1) Environment
- PrimeFaces version: 8.0RC3
- Does it work on the newest released PrimeFaces version? No (Checked against 7.0.10)
- Application server + version: -
- Affected browsers: All

## 2) Expected behavior

The dragged row should not change when reflow="true" is set.

## 3) Actual behavior

The dragged row contains the header text as well as the row content.

## 4) Steps to reproduce

See supplied reproducer. Try to drag a row.

## 5) Sample XHTML

See supplied reproducer. 

## 6) Sample bean

See supplied reproducer.

melloware:
Tested with reflow="true" and reflow="false" and now the draggable row text is properly the row only.


tandraschko:
+1 if well tested

PR #5739, opened 2020-03-30
Fix #5582: Datatable cancel drag/drop events if table updated while dragging

### LINKED ISSUE #5582: DataTable: Reordering duplicates columns after update
## 1) Environment
- PrimeFaces version: 8.0-Snapshot
- Does it work on the newest released PrimeFaces version? No (7.0)
- Application server + version: WildFly 18
- Affected browsers: Firefox

## 2) Actual behavior

On our page we have a `p:dataTable` with `draggableColumns="true"` which is updated via Ajax when there is new data available. If an update happens after the user has started dragging a column and then the user drops the column at the desired position, this column is duplicated.

## 3) Steps to reproduce

Can be reproduced with the showcase 'DataTable - Reorder' by adding an id to the table and a `p:poll` to simulate the update:

```
<p:poll update="reorderTable" autoStart="true"/>
    
    <p:dataTable id="reorderTable" ...
```

Then drag a column, wait 2 seconds until the table is updated, and drop the column at another position afterwards.

melloware:
Fixes it for both draggableRows and draggableColumns

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
+++ b/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
@@ -158,7 +158,7 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
 
         // #5582: destroy any current draggable items
         var dragdrop = $.ui.ddmanager.current;
-        if ( dragdrop ) {
+        if (dragdrop) {
             document.body.style.cursor = 'default';
             dragdrop.cancel();
         }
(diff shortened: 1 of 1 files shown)


tandraschko:
Please care about checkstyle:
if ( dragdrop ) {
->
if (dragdrop) {

otherwise +1

melloware:
@tandraschko Do you have a general Javascript formatting guideline somewhere? I know we have checkstyle for Java but for javascript I would like to get my Eclipse formatter to match what the standards are?

- Spaces not Tabs
- 4 spaces
- colon next to the function like "refresh: function" and not "refresh : function"

I see some if's withthe paren right next to the if like..
```javascript
if(empty)
```
and some with a space like
```javascript
if (empty)
```

Thoughts I think I can even come up with something that we can run "mvn formatter:format" once we have all the rules for format all the JS files at once or before releases.

tandraschko:
we dont have something but we should align it with our java code (space after if)

melloware:
OK let me see if I can put a formatter together for the team to review.

PR #5743, opened 2020-03-31
Fix #1181: RemoteCommand add CSV support

### LINKED ISSUE #1181: RemoteCommand: Add CSV support / validateClient attribute
For RemoteCommand Client-side validation is not implemented (no attribute "validateClient" is available)

melloware:
Added CSV support to remote command.

Also realized the `validateClient` docs were missing from 8.0 and 9.0 for CommandButton and CommandLink.


tandraschko:
+1 
it would be great if you can add a sample to the showcase

melloware:
Showcase example added!

Issue #5746, opened 2020-04-01
Static Analysis: Inneficient iteration of a Map


dvmarcilio:
I found a small issue when experimenting with [Facebook's Infer Static Analyzer](https://fbinfer.com/). I'm providing a short pull request fixing this issue.

[It happens in TreeRender, line 821-823](https://github.com/primefaces/primefaces/blob/cea1b5317da9d54108abcf5df71e5979c604d7e3/src/main/java/org/primefaces/component/tree/TreeRenderer.java#L821). 

This is similar to the SonarQube rule: ["entrySet()" should be iterated when both the key and value are needed](https://rules.sonarsource.com/java/RSPEC-2864), fixed in #4886.
However, SonarQube does not detect this specific issue.

PR #5747, opened 2020-04-01
Inneficient iteration of a Map (#5746)

dvmarcilio:
The fix here is slightly different than https://github.com/primefaces/primefaces/pull/4887, as in this case the key `String type` is only used within the loop to retrieve the value `UiTreeNode node`. 
So, I just iterated directly on `nodes.values()` (the Map's values Collection).


tandraschko:
Thanks!

PR #5754, opened 2020-04-01
Fix #5753: DatePicker incorrect off() events

### LINKED ISSUE #5753: DatePicker: off() not being called for all on() events
Looking at the code while debugging another issues I realized we had `on` events with mismatched or missing `off` events in the DatePicker.  PR to follow.

melloware:


tandraschko:
+1

PR #5768, opened 2020-04-04
More fixes for Inneficient iteration of Maps (#5746)

dvmarcilio:
Sorry for revisiting Issue #5746, I overlooked other 4 violations.

The fixes here are different from https://github.com/primefaces/primefaces/pull/5747.
They introduce a `Map.Entry<K, V>` variable to preserve the original key and values variables.

### DIFF
--- a/src/main/java/org/primefaces/application/DialogNavigationHandler.java
+++ b/src/main/java/org/primefaces/application/DialogNavigationHandler.java
@@ -87,9 +87,10 @@ public class DialogNavigationHandler extends ConfigurableNavigationHandler {
 
             sb.append(",options:{");
             if (options != null && !options.isEmpty()) {
-                for (Iterator<String> it = options.keySet().iterator(); it.hasNext();) {
-                    String optionName = it.next();
-                    Object optionValue = options.get(optionName);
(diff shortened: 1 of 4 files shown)


tandraschko:
doesn't it make sense to switch to a normal for-loop instead iterator?

dvmarcilio:
> 
> 
> doesn't it make sense to switch to a normal for-loop instead iterator?

I think keeping the `Iterator`s is the less disruptive change here.
In all of these cases the loop checks at the end `if (it.hasNext())` to intercalate the values with a `","`.
The diffs in the PR don't show it, we have to expand it. 
Here is one example, more specifically on line 60: https://github.com/primefaces/primefaces/blob/42065c62c75264b7b13417953eac1f14803fc440/src/main/java/org/primefaces/component/chart/renderer/DonutRenderer.java#L53-L65

melloware:
> And why not use stream Api? Pf8 is Java8+ requiered

I agree with @tandraschko I find them less readable and IF an exception is thrown inside the stream the stack trace is almost impossible to decipher.

markusg80:
In my point of view the stream api is very well readable! Maybe its because i use it very often. The point with debbuging is true! This is a real drawback! But the features of the stream api are so fency

PR #5770, opened 2020-04-04
Fix #5769: Colorpicker add close AJAX event

### LINKED ISSUE #5769: ColorPicker: add close behavior
Hi, 

can you please add a "close" event on p:colorPicker.
I need to fire a listener only if the color has been really choosen not on every mouse move (if I hold my mouse key).

See also:
https://forum.primefaces.org/viewtopic.php?f=3&t=61485

melloware:
Showcase PR added too


tandraschko:
+1

PR #5772, opened 2020-04-06
Fix #5150: DataTable checkboxes accessible by voice commands

### LINKED ISSUE #5150: DataTable: Selection - Checkboxes inaccessible for voice commands
## Summary
When using selection for DataTable, checkboxes can't be toggled via voice commands. The root cause is very similar to [ISSUE-4979](https://github.com/primefaces/primefaces/issues/4979) which is fixed by [PULL-4981](https://github.com/primefaces/primefaces/pull/4981).

* The `<div>` that represents the checkbox doesn't have a role attribute
* The `<div>` that represents the checkbox doesn't have aria-checked attribute
* The `<input>` is not required and can be removed if the `<div>` is fixed

## 1) Environment
*  Does not work with 7.0.6
*  Does not work in the showcase [DataTable - Selection](https://www.primefaces.org/showcase/ui/data/datatable/selection.xhtml)

## 2) Expected behavior
Checkbox is listed when calling a command to interact with checkboxes

> Click checkbox

## 3) Actual behavior
UI Elements are not recognized as checkboxes, therefore nothing happens. You are not able to interact with checkboxes

## 4) Steps to reproduce
Try to use voice commands to access the components checkboxes

## 5) Sample Fix
* Please refer to [PULL-4981](https://github.com/primefaces/primefaces/pull/4981)

MarkusChur:
This PR changes DataTable row selection so that checkboxes are accessible by voice commands. This works with native as well as with div-based checkboxes.

It is a bit more complex in comparison to the ColumnToggler changes because of the common scripts for native input and hidden accessible.


melloware (approved):
Excellent change!  @tandraschko this is similar to what @MarkusChur did for the column toggler checkboxes to make them more accessible.

tandraschko:
+1 in general
but please test some datatable cases if everyhing still works
also a short review of @mertsincan would be great

melloware:
@MarkusChur Can you check to make sure you are meeting this requirement with this new fix: https://github.com/primefaces/primefaces/issues/5868

melloware:
I think your PR is still fine that fix touches a line of code i don't think you touched.

MarkusChur:
yep, changes do not interfere.

melloware:
I have tested this pretty well.  Merging.

Rapster:
Hi @MarkusChur 

I know it's been a while but have you tried your fix using lazy feature? From what I can see I don't think it will work since checkbox will get an id like "form:carTable_null_checkbox" (with lazy feature rowKey is defined on `LazyDataModel#getRowKey` and not `DataTable#rowKey` that's something i'm working on to make the use of rowKey more simple/consistent).

Instead of using rowKey, maybe using the rowIndex would solve that problem?

MarkusChur:
Hey @Rapster 
I'm not sure but most likely I've not tested that. Each individual checkbox needs a unique id. Using rowKey and fallback to rowIndex if rowKey is null might work out. At least it's better than having colliding ids.

PR #5778, opened 2020-04-07
Fix #5777: Remove @ResourceDependencies annotation

### LINKED ISSUE #5777: Task: Remove @ResourceDependencies wrapper annotation
Before Java 8 if you needed to use multiple instances of the same annotation, they had to be wrapped in a container annotation. With Java 8, that's no longer necessary, allowing for cleaner, more readable code.

**Noncompliant Code Example**
```java
@ResourceDependencies ({  // Noncompliant
  @ResourceDependency(..a..),
  @ResourceDependency(..b..),
  @ResourceDependency(..c..),
})
public class Component {
  ...
}
```

**Compliant Solution**
```java
@ResourceDependency(..a..)
@ResourceDependency(..b..)
@ResourceDependency(..c..)
public class Component {
  ...
}
```

melloware:
This looks like a bigger change than it is....

### DIFF
--- a/src/main/java/org/primefaces/behavior/ajax/AjaxBehavior.java
+++ b/src/main/java/org/primefaces/behavior/ajax/AjaxBehavior.java
@@ -28,7 +28,6 @@ import java.util.EnumSet;
 import java.util.Set;
 
 import javax.el.MethodExpression;
-import javax.faces.application.ResourceDependencies;
 import javax.faces.application.ResourceDependency;
 import javax.faces.component.behavior.ClientBehaviorHint;
 import javax.faces.event.AjaxBehaviorListener;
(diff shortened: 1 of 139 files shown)


tandraschko:
+1

tandraschko:
@melloware what happens e.g. with this code: https://github.com/apache/myfaces/blob/master/impl/src/main/java/org/apache/myfaces/application/ApplicationImpl.java#L1672

Will it only collect 1 annotation?

tandraschko:
I think it will return the first only, so we need to revert this change here

tandraschko:
... andcheck mojarra and myfaces how they receive the annotations in all cases, to make it compatible for the future

tandraschko:
or will it automatically be wrapped by a ResourceDependencies annotation?

melloware:
@tandraschko I tested on Mojarra and MyFaces and didnt see any issues.  I believe this is JDK8 compiler magic that if it sees `@Repeatable(ResourceDependencies.class)` it automatially compiles it with the `@ResourceDependencies` parent.

PR #5791, opened 2020-04-14
DataTable rowsPerPageTemplate-ShowAll+ filtering throws NumberFormatE…

### LINKED ISSUE #5704: DataTable rowsPerPageTemplate-ShowAll+ filtering throws NumberFormatException
## 1) Environment
- PrimeFaces version: Primesfaces 8.0
- Affected browsers: all

## 2) Expected behavior

When I select "ShowAll" in DataTable's rows per page drop-down list, all the rows should be filtered.

## 3) Actual behavior

When the "ShowAll" option is selected, an NumberFormatException is thrown during filtering

     java.lang.NumberFormatException: For input string: "*"
        at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
        at java.base/java.lang.Integer.parseInt(Integer.java:638)
        at java.base/java.lang.Integer.parseInt(Integer.java:770)
        at deployment.test.war//org.primefaces.component.datatable.feature.FilterFeature.encode(FilterFeature.java:106)
        at deployment.test.war//org.primefaces.component.datatable.DataTableRenderer.encodeEnd(DataTableRenderer.java:88)
        at javax.faces.api@2.3.9.SP01//javax.faces.component.UIComponentBase.encodeEnd(UIComponentBase.java:595)

## 4) Steps to reproduce

Set the "records per page" to "all" and make some filtering

## 5) Sample XHTML

        <h:form>
            <p:dataTable value="#{dataTableController.rows}" var="row" paginator="true"
                         rowsPerPageTemplate="5,10,{ShowAll|'All'}" rows="5">
                <p:column filterBy="#{row}" filterMatchMode="contains">
                    <h:outputText value="#{row}"/>
                </p:column>
            </p:dataTable>
        </h:form>

## 6) Sample bean

    @Named
    @ApplicationScoped
    public class DataTableController implements Serializable {
        private static final int ROW_COUNT = 15;
        private static final List<String> ROWS = new ArrayList<>(ROW_COUNT);

        static {
            for (int i = 0; i < ROW_COUNT; i++) {
                ROWS.add("Row " + i);
            }
        }

            return ROWS;
        }
    }

tomlux:
…xception #5704

### DIFF
--- a/src/main/java/org/primefaces/component/datatable/feature/FilterFeature.java
+++ b/src/main/java/org/primefaces/component/datatable/feature/FilterFeature.java
@@ -102,7 +102,7 @@ public class FilterFeature implements DataTableFeature {
 
         //update rows with rpp value
         String rppValue = params.get(clientId + "_rppDD");
-        if (rppValue != null) {
+        if (rppValue != null && !rppValue.equals("*")) {
             table.setRows(Integer.parseInt(rppValue));
         }
 
--- a/src/main/java/org/primefaces/component/treetable/TreeTableRenderer.java
+++ b/src/main/java/org/primefaces/component/treetable/TreeTableRenderer.java
@@ -180,7 +180,7 @@ public class TreeTableRenderer extends DataRenderer {
 
             //update rows with rpp value
             String rppValue = params.get(clientId + "_rppDD");
-            if (rppValue != null) {
+            if (rppValue != null && !rppValue.equals("*")) {
                 tt.setRows(Integer.parseInt(rppValue));
             }
 


melloware (approved):

tandraschko:
does it fix the behavior or just the exception?

tomlux:
it's fixing the behavior, not only the exception

But we have to trigger a new filtering to get the correct results which is not included in this fix:

`
<p:ajax event="page" oncomplete="PF('tblTest').filter(); return false;"/>
`

tandraschko:
All right, thanks!

PR #5792, opened 2020-04-14
resolve #5787: Use this._super in #init method of widgets

### LINKED ISSUE #5787: Widgets: Missing this._super(cfg) call in init method of a few widgets
While working on the type declarations for the JavaScript API, I noticed a few widgets are missing the call to `this._super(cfg)` in the init method. Is that intentional? The [init method](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/core/core.widget.js#L74) from the BaseWidget doesn't seem to be doing anything harmful

This results in a few properties such as `widgetVar` and `jq` not being set (violating the Liskov substitution principle unless those properties are made optional in `BaseWidget`), and repeats some boilerplate such as `this.jqId = PrimeFaces.escapeClientId(this.id)`.

Following widgets are missing the super call:

- [BlockUI](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js#L7)
- [DefaultCommand](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/forms/forms.defaultcommand.js#L7)
- [Draggable](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/dragdrop/dragdrop.js#L7) (`this.jq` is set to the target, should probably use `this.jqTarget`  for that)
- [Droppable](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/dragdrop/dragdrop.js#L44) (`this.jq` is set to the target, should probably use `this.jqTarget`  for that)
- [Effect](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/effect/effect.js#L7)
- [Growl](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/growl/growl.js#L7)
- [KeyFilter](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/keyfilter/1-keyfilter.js#L9)
- [Paginator](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/paginator/paginator.js#L4) (pretty strange - `this.id` is an array of the IDs of all paginators [such as the top and bottom of a data list] which is incompatible with the `BaseWidget` super type; setting `id` to the first id and adding `this.extraIds` could help)
- [Poll](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/poll/poll.js#L7)
- [Resizable](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/resizable/resizable.js#L7)
- [Tooltip](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/tooltip/tooltip.js#L7)
- [Watermark](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/watermark/watermark.js#L7)

blutorange:
See #5787

* Added `this._super(cfg)` to the init method of the widgets mentioned in the issue above
* Removed the following lines from the individual widgets as it's done by `BaseWidget#init`:
    * `this.removeScriptElement(this.id)`
    * `this.cfg = cfg;`
    * `this.id = cfg.id;`
    * `this.jqId = PrimeFaces.escapeClientId(this.id); `
* Draggable & Resize
    * For consistency, `BaseWidget#jq` should be the element with the id `BaseWidget#jqId`
    * But these two widgets set `this.jqId = PrimeFaces.escapeClientId(this.id);` and `this.jq = $(PrimeFaces.escapeClientId(this.cfg.target));`
    * :point_right: I changed that to `this.jqTarget = $(PrimeFaces.escapeClientId(this.cfg.target));`
 * Paginator: `cfg.id` is an array of IDs
    * :point_right: I modified `BaseWidget#init` so that it works regardless of whether `cfg.id` is a `string` or `string[]`

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
+++ b/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
@@ -4,9 +4,8 @@
 PrimeFaces.widget.BlockUI = PrimeFaces.widget.BaseWidget.extend({
     
     init: function(cfg) {
-        this.cfg = cfg;
-        this.id = this.cfg.id;
-        this.jqId = PrimeFaces.escapeClientId(this.id);
+        this._super(cfg);
+
         this.block = PrimeFaces.expressions.SearchExpressionFacade.resolveComponentsAsSelector(this.cfg.block);
         this.content = $(this.jqId);
         this.cfg.animate = (this.cfg.animate === false)? false : true;
@@ -21,8 +20,6 @@ PrimeFaces.widget.BlockUI = PrimeFaces.widget.BaseWidget.extend({
         if(this.cfg.blocked) {
             this.show();
         }
-
-        this.removeScriptElement(this.id);
     },
             
     refresh: function(cfg) {
--- a/src/main/resources/META-INF/resources/primefaces/core/core.widget.js
+++ b/src/main/resources/META-INF/resources/primefaces/core/core.widget.js
@@ -74,14 +74,21 @@ if (!PrimeFaces.widget) {
         init: function(cfg) {
             this.cfg = cfg;
             this.id = cfg.id;
-            this.jqId = PrimeFaces.escapeClientId(this.id);
+            if ($.isArray(this.id)) {
+                this.jqId = $.map(this.id, function(id) {
+                    return PrimeFaces.escapeClientId(id);
+                }).join(",");
+            }
+            else {
+                this.jqId = PrimeFaces.escapeClientId(this.id);
+            }
             this.jq = $(this.jqId);
             this.widgetVar = cfg.widgetVar;
             this.destroyListeners = [];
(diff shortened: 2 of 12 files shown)


melloware (approved):

tandraschko:
+1
can you retest the showcase @melloware?

blutorange:
jfyi I tested the changes against the showcase, but an independent test would be great :)

melloware:
yep i am checking

melloware:
Confirmed this all looks good. Merging.

PR #5795, opened 2020-04-14
Fix #5794: Replace transparent pixel with base64 image

### LINKED ISSUE #5794: Performance: Replace dot_clear.gif with Base64 URL
See this article: http://png-pixel.com/

This will save an HTTP trip for dot_clear.gif and the browsers will still cache it properly and it will be part of the CSS load and minification.

I kept noticing every page i hit with datatables or spacers was loading dot_clear.gif and found many articles like the one above saying to base64 encode it.  

I tested it thoroughly and it working great. PR Coming.

melloware:


tandraschko:
+1

PR #5820, opened 2020-04-21
Fix #5809: SelectOneMenu NPE in decode method

### LINKED ISSUE #5809: SelectOneMenu: NPE in decode method
setEditable(true) gives an NPE at SelectOneMenuRenderer.decode.
PrimeFaces version: 7.0, 7.0.12, 8.0

melloware:
After reproducible test case this NPE is obvious and should be guarded.


tandraschko:
+1

PR #5826, opened 2020-04-22
fix #5825

### LINKED ISSUE #5825: DataTable: fix sort oncomplete callback
## 1) Environment
- PrimeFaces version: master, latest 8.0, 7.*

## 2) Expected behavior
If sort ajax request is aborted, there is not any error. The problem is in sort function of datatable.js. The oncomplete callback doesn't fail if args is undefined. The second if condition have to check args like the first one.

vonnai:


tandraschko:
looks for me exact like the same IF-statement :D

vonnai:
it looks, but it is not. There is solved problem with undefined args. Are you going to merge it, or do you have another idea how to fix it? Thanks

tandraschko:
Is there really an difference between

```
    if(args && args.totalRecords) {
```

and

```
    if(args) {
      if (args.totalRecords) {
```

vonnai:
and what about if-statement 
`If (!args.validationFailed) {`
?? In original source there was missing check for args.
This is really bad code:
`if(args && args.totalRecords) {`
`if(!args.validationFailed) {`

tandraschko:
ahhh i see! sorry

What does it exactly fix?
it would be great to have a better description in the issue

vonnai:
you can abort sort ajax request without any error, I put this information into issue's description.

tandraschko:
how can you abort that?
we need at least the informations to replicate issues

vonnai:
Idea:
Datatable reads data from database lazily. Asynchronous events filter, sort and paginate can take a long time. How to stop these long-running request on client side? It's possible to find current ajax request in PrimeFaces.ajax.Queue abort it and execute another e.g. sort event. How to stop query execution on server side it's other story.

It's simple to add aborting of ajax request to onstart callback.

```
abortAjaxRequest: function(source, callback) {
        var aborted = false;
        PrimeFaces.ajax.Queue.xhrs.forEach(function(xhr){
            if (source === xhr.pfSettings.source) {
                aborted = true;
                xhr.abort();
                PrimeFaces.ajax.Queue.removeXHR(xhr);
            }
        });

        if (!aborted && callback) {
            callback.call();
        }
    }

<p:ajax process="@this" event="filter" async="true" 
                onstart="abortAjaxRequest('#{p:resolveClientId('@this', component)}', function(){PF('blockui').show();});" 
                oncomplete="if(status!=='abort'){PF('blockui').hide();}"/>
```

I'm sorry, if my comment is not understandable.

tandraschko:
ok thanks!

PR #5828, opened 2020-04-22
Fix #5827 Inverted order of method resolution in AjaxBehaviorListenerImpl

### LINKED ISSUE #5827: Ajax listener not invoked anymore in Payara 5
it seems that MethodNotFound / IllegalArgumentException are wrapped inside ELException

Fixed by: https://github.com/primefaces/primefaces/commit/e880d5ff6f143a1ffec2a5049e5979afb8fef960

See https://github.com/primefaces/primefaces/issues/5711 for details

dmatej:
- first try to call method with parameters we have
- if it fails, try to call method with parameters
  - then EL implementation fills missing parameters with nulls
  - if the method has no parameters, behavior is still correct
- see also OverloadedMethodTest in EL library

### DIFF
--- a/src/main/java/org/primefaces/behavior/ajax/AjaxBehaviorListenerImpl.java
+++ b/src/main/java/org/primefaces/behavior/ajax/AjaxBehaviorListenerImpl.java
@@ -69,22 +69,10 @@ public class AjaxBehaviorListenerImpl implements AjaxBehaviorListener, Serializa
         }
 
         try {
+            processArgListener(context, elContext, event);
+        }
+        catch (MethodNotFoundException | IllegalArgumentException | ArrayIndexOutOfBoundsException e) {
             listener.invoke(elContext, new Object[]{});
-        }
-        catch (MethodNotFoundException | IllegalArgumentException e) {
-            processArgListener(context, elContext, event);
-        }
-        // JBoss hack, see #1375
-        catch (ArrayIndexOutOfBoundsException e) {
-            processArgListener(context, elContext, event);
-        }
-        // it's wrapped in a ELException since newer Payara versions
-        catch (ELException e) {
-            if (e.getCause() instanceof MethodNotFoundException || e.getCause() instanceof IllegalArgumentException) {
-                processArgListener(context, elContext, event);
-                return;
-            }
-            throw e;
         }
     }
 


tandraschko:
i'm not sure.
I can clearly see that you and also the Tomcat EL wrapts MNF and IAE inside a ELException? So it makes still sense to me?

tandraschko:
and is your problem still not working the latest PrimeFaces changes?
Is it only replicatable with Payara or also with our primefaces-test project? (it uses jetty with Tomcat EL)

tandraschko:
It would be really great if you could create a forked primefaces-test to demonstrate your problem
you can also switch the EL impls via MF/Mojarra context param

then we can really compare

dmatej:
The reproducer is here - with this PR in PF it works without any changes (except changing dependency version of PrimeFaces).
https://github.com/payara/Payara/issues/4435

That wrapping was a mistake - the try block starts after buildParameters, so they are wrapped only if they are thrown by the called target method - that is required by the specification and I think you should not catch these exceptions.

The buildParameters in current EL does not throw exceptions and is much safer, but in old version did - that's why you have those catch block in your code, it respected the state and reacted.

dmatej:
I don't have much time, but I can probably create a test if you can lead me to a place where it belongs to ;-)

tandraschko:
Yes, please
i need a test, i'm completely confused

You can fork
https://github.com/primefaces/primefaces-test
and create a bean+xhtml and we can replicate with profiles in nearly all JSF versions

it uses per default jetty + Tomcat EL impl

So we can test your case and then switch the EL impl (please provide a maven coordinate) and see what happens

dmatej:
### **Well, that wasn't a nice trap.** 

The Jetty uses Apache's EL and even Jetty 9 is compliant only with Java EE 7 Web Profile, so the problem is not reproducible on Jetty. I spent whole afternoon with tries to force Jetty to use the el-ri dependency, but it is not possible. No surprises.

At least I know that the problem is not present on Jetty 9 with Mojarra and MyFaces - it's not their bussiness, the problem is in how interacts PrimeFaces with the EL implementation behind Faces impl, compliant with Java EE8 and later.

So ... I created simple test here: https://github.com/primefaces/primefaces-test/pull/26 - it does not fail, because I did not implement any error processing, but the NPE is visible in the log.

tandraschko:
you should have ask earlier :D i said that you can switch the EL impl easily in JSF

you can just set that context params:
com.sun.faces.expressionFactory
org.apache.myfaces.EXPRESSION_FACTORY

so its enough if you add the EL-impl by your choich as a WAR-dependency and set the contextparam

then we can easily start the jetty with the other EL impl and replicate it

dmatej:
It did not work, because jetty simply did not allow to put el-ri on classpath.

tandraschko:
hmmm, i see :/

then we need to try with tomcat-maven-plugin
i really would like to debug different EL impls and versions

dmatej:
Or you can simply use some of arquillian container. Jetty has also it's own.
Or TestContainers.
Everything has pros and cons, but all environments are debuggable with some configuration.
I'm not sure if and how Tomcat does it, but most Java EE implementations protect their "core dependencies".

tandraschko:
@dmatej 
yep i see
i would like to use tomcat, everything else requires more work on my side to dig into testContainers - i'm also doing this in my sparetime here

the sample xhtml+bean inside your whole testproject should be enough for me to replicate

could you point me to the el-impl maven coordinates + version?

dmatej:
These are current Payara dependencies, the second will be now 3.0.3.payara-p1 with the ugly workaround for PrimeFaces.
```
            <dependency>
                <groupId>jakarta.el</groupId>
                <artifactId>jakarta.el-api</artifactId>
                <version>3.0.3</version>
            </dependency>
            <dependency>
                <groupId>org.glassfish</groupId>
                <artifactId>jakarta.el</artifactId>
                <version>3.0.3</version>
            </dependency>
```
And the com.sun.el.util.ReflectionUtil.invokeMethod + buildParameters are two methods causing the problem, because older versions did not support neither varargs nor missing parameters and in such cases it throwed an exception.

tandraschko:
i added
```
                <dependency>
                    <groupId>org.glassfish</groupId>
                    <artifactId>jakarta.el</artifactId>
                    <version>3.0.3</version>
                </dependency>
```

and set

```
    <context-param>
        <param-name>org.apache.myfaces.EXPRESSION_FACTORY</param-name>
        <param-value>com.sun.el.ExpressionFactoryImpl</param-value>
    </context-param>
```

start with "mvn clean package jetty:run-exploded -Pmyfaces23next" and i can debug that com.sun.el.* stuff is used and correclty a MethodNotFound is thrown when clicking on a item in index.xhtml, after that the fallbacks are called and finally the right method
(NOTE that the profile myfaces23next is new in primefaces-test master, i think Mojarra and older MyFaces will not change anything here)

i will now check the second autoCompleteTest.xhtml view

tandraschko:
autoCompleteTest.xhtml really fails now

@dmatej could you please explain what is the difference? index.xhtml works exactly like all other containers and older/other EL impls.
autoCompleteTest.xhtml fails with the new glassfish EL, which feels weird as the same mechanism works fine in index.xhtml

tandraschko:
i personally dont see any reason why the glassfish-EL should call the method when we explicitly lookup the MethodExpression with no-args
the glassfish-EL correctly throws a MethodNotFound in the index.xhtml but not in autoCompleteTest.xhtml

dmatej:
Those files came from contributor, I only debugged and reviewed the workaround of my colleague (and originally refused until I found that there's no other way than to fix PrimeFaces).
The index file uses different mechanism and my test does not cover it's use case. The second file uses composite component; probably they go through different classes ...

> the glassfish-EL correctly throws a MethodNotFound

And does it? Or it directly calls the right method? EDIT: yes, it throws one of catched exceptions in index.xhtml case, which uses p:ajax element.

The fixed PF 9.0-SNAPSHOT on Payara 5.202-SNAPSHOT (now not sure if fixed or not, but that stacktrace should be same, because PF call directly the correct method with arguments):
image

The original PF9.0-SNAPSHOT on Payara 5.201
image

The stacktrace - I simply added one line as the first line to the reproducer's method:
`new RuntimeException().prinstStacktrace();`

tandraschko:
My last comments are about the current PrimeFaces behavior without your PR, as it worked since years.

As i said, it really confuses me why your EL-Impl correctly throws a MethodNotFound on index.xhtml but not on autoCompleteTest.xhtml.

The behavior is completely the same on our side for both cases.
In both cases we first invoke the MethodExpression which was created with a empty-params-array:
`listener.getMethodExpression(ctx, Void.class, new Class<?>[]{})`

In both cases, the method has a parameter and therfore a MethodNotFound should be thrown.
- index.xhtml throws the MethodNotFoundException
- autoCompleteText.xhtml invokes with the method with a null parameter

dmatej:
I think that driving business logic by exceptions especially without logging exceptions is know antipattern, one of the worst. I see it both in PrimeFaces and in EL.

I don't see any sense of calling nonexisting method without parameter you have (btw if it would succeed, the parameter would be worthless?) and after the call fails you try to call the right method (hurray!), but maybe I am missing something. Before I created that payara reproducer I tried the fix with your jetty and it worked too.

You say it worked for years, but I tried even PF 5.2 and it ended with the same exception.

One problem I see, is that it is not covered by TCK tests, only by EL-RI unit test. I am not sure if it is some consequence of the specification requirement or if it is some "evil feature".

Sources of EL are a chapter for itself, you can create an issue to Eclipse and also PR if you have an idea how to change it. As I said, driving logic by exceptions is a horrible practice. This issue is nearly perfect example of it.

I created the PR for you in my free time, because I think it is good for all. But you are the authority who makes decisions, so now it is your choice, if you will accept or deny it (or change it).

tandraschko:
> I don't see any sense of calling nonexisting method without parameter you have

In EL you dont know if the method is existing or not before calling it. Therefore its important that the ELImpl correctly throws a MethodNotFoundException, which doesn't happen in the autoCompleteTest.xhtml case.

> You say it worked for years, but I tried even PF 5.2 and it ended with the same exception.

i dont mean that our code worked/works fine with your new ELImpl version - i mean that our code works fine in all other containers since years.

> I think that driving business logic by exceptions especially without logging exceptions is know antipattern, one of the worst. 

yep, not nice but thats the way how EL works. See my first point here.



As long as you don't bring any reason why index.xhtml works and autoCompleteTest.xhtml not, i think it's a bug in the ELImpl as it should throw a MethodNotFound in both cases.
If we change the order in PrimeFaces, it would be a workaround for this issue but IMO there would be still something wrong in the ELImpl. index.xhtml and autoCompleteTest.xhtml should behave the same.

tandraschko:
willc lose the PR for now. Please lets discuss further in the issue, not in the PR.

PR #5834, opened 2020-04-23
Fix #5833: Inplace should remain editing on validation failures

### LINKED ISSUE #5833: Inplace: switches to output mode even if validation fails when facets are used
## 1) Environment
- PrimeFaces version: 6.2
- Does it work on the newest released PrimeFaces version? NO
- Application server + version: Tomcat 8.0.39
- Affected browsers: Firefox 75

## 2) Expected behavior
When I have a `p:inplace` with `editor="true"` and nested `f:facet`s for input and output, if I press on the confirm button but validation fails, the component does not switch to output mode and just updates its input facet contents.

## 3) Actual behavior
When I have a `p:inplace` with `editor="true"` and nested `f:facet`s for input and output, if I press on the confirm button but validation fails, the component switches to output mode, without applying changes. Switching again to input mode, shows validation error consequences.

## 4) Steps to reproduce
- create a `<p:inplace editor="true">` with nested `<f:facet name="output">` and `<f:facet name="input">`
- in the `<f:facet name="input">` put an input component with some validation rules and a `<p:message>`
- click on the `<p:inplace>`
- change the input value to an invalid value
- click on the confirm button

## 5) Sample XHTML

```xml
<p:inplace editor="true">
  <f:facet name="output">
    <h:outputText value="#{bean.foo}">		
      <f:convertNumber />
    </h:outputText>
  </f:facet>
  <f:facet name="input">
    <p:inputText id="fooInput" value="#{bean.foo}" required="true">
      <f:convertNumber />
      <f:validateDoubleRange minimum="0" maximum="100" />
    </p:inputText>
    <p:message id="fooMessage" for="fooInput" />
  </f:facet>
  <p:ajax event="cancel" global="false" />
  <p:ajax event="save" global="false" />
</p:inplace>
```

## 6) Sample bean

```java
public class Bean {
  private double foo;
  // getters and setters omitted
}
```

See the difference if you change the facelet fragment to:

```xml
<p:inplace editor="true" label="#{bean.foo}">
  <p:inputText id="fooInput" value="#{bean.foo}" required="true">
    <f:convertNumber />
    <f:validateDoubleRange minimum="0" maximum="100" />
  </p:inputText>
  <p:message id="fooMessage" for="fooInput" />
  <p:ajax event="cancel" global="false" />
  <p:ajax event="save" global="false" />
</p:inplace>
```

In the latter case, the behaviour is as expected: validation errors are immediately shown and the inplace component does not switch to output mode until you fix your changes or hit the cancel button.

melloware:


melloware:
Investigating a better way to fix this.

melloware:
This is the correct way to fix it. All the code was alreayd there to leave the editor open the problem is when looping over the facets it wasn't also looping over the children of those facets.

PR #5835, opened 2020-04-23
Fix #5830: DatePicker dateSelect event in range mode

### LINKED ISSUE #5830: DatePicker: selectEvent with selectionMode="range" and AJAX Event dateChange not work
## 1) Environment
- 8.0
- Application Tomcat 9.0.33
https://forum.primefaces.org/viewtopic.php?f=3&t=61731
## 2) Expected behavior
call dateChange event after set correct start - end
## 3) Actual behavior
By the first click of a day. my ArrayList value is empty. By the second click my value is now null
Post in PF forum: 
## 4) Steps to reproduce
see 5)
## 5) Sample XHTML
```xml
<p:datePicker id="calendarRange" converter="#{value}" validator="#{value.validate}" selectionMode="range" value="#{value.currentValue}"
showIcon="true" size="18"
yearRange="#{value.yearRange}" locale="#{localeProvider.language}"
styleClass="#{value.styleClass}" showButtonBar="true" showOnFocus="false" numberOfMonths="3">
<p:ajax event="dateSelect" process="@this" update="mainForm:filterPanel" />
<f:ajax event="change" execute="@this" render="mainForm:filterPanel"/>
</p:datePicker>
```

melloware:
This only fires the `dateSelect` event when the full range is selected.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
+++ b/src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
@@ -213,13 +213,12 @@ PrimeFaces.widget.DatePicker = PrimeFaces.widget.BaseWidget.extend({
      * @private
      */
     fireDateSelectEvent: function() {
-        if(this.cfg.behaviors) {
-            var dateSelectBehavior = this.cfg.behaviors['dateSelect'];
-
-            if(dateSelectBehavior) {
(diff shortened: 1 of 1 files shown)


christophs78:
Look´s good! :-)

And DatePicker allows to do the following if we want to select a single-day-range:
grafik

PR #5837, opened 2020-04-24
Fix #5748: DatePicker: timeOnly causes infinite loop

### LINKED ISSUE #5748: DatePicker: timeOnly causes infinite loop
## 1) Environment
- PrimeFaces version: 8
- Does it work on the newest released PrimeFaces version?  Non
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source): Non
- Application server + version: Payara 5.201
- Affected browsers: Firefox+ Chrome

## 2) Expected behavior

The timer only is not starting an infinite loop and user can easily set the time

## 3) Actual behavior

After few clicks on the timer an infinite loop is started and the screen become unresponsive

## 4) Steps to reproduce

When trying to set up a datepicker with time only 

## 5) Sample XHTML
```xml
        <ui:decorate template="/fragments/edit.xhtml">
          <p:datePicker
            id="reservationEarliestTime"
            pattern="HH:mm"
            timeOnly="true"
            value="#{instrumentSetting.reservationEarliestTime}">
            <p:ajax
               event="dateSelect"
               listener="#{instrumentSettingManager.reservationEarliestTimeChanged}"
               update="edit"
               process="@this"
               global="false" />
          </p:datePicker>
        </ui:decorate>
```

## 6) Sample bean

```java
    public void reservationEarliestTimeChanged(SelectEvent<LocalTime> event) {
        getInstrumentSetting().setReservationEarliestTime(event.getObject());
        if (getInstrumentSetting().isReservationEarliestLatestTimeInvalid()) {
            getFacesMessagesManager().printError(Messages.get("errorLabel"));
        }
    }
```

christophs78:
@melloware : I need your help!

I tried to fix this issue by triggering dateSelect at mouseUp instead after each increment.

But callBehavior fails because at this point there´s some other "this" object:
grafik

One which does not have the dateSelect-Behavior.

This seems to be a trivial issue. Do you have a tip how to work arround this issue.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
+++ b/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
@@ -1523,8 +1523,10 @@
                 .on('mouseup.datePicker-time', timeSelector, null, function (event) {
                     $this.onTimePickerElementMouseUp(event);
                 })
-                .on('mouseleave.datePicker-time', timeSelector, null, function () {
-                    $this.clearTimePickerTimer();
+                .on('mouseleave.datePicker-time', timeSelector, null, function (event) {
+                    if (this.timePickerTimer) {
+                        $this.onTimePickerElementMouseUp(event);
+                    }
                 })
                 .on('click.datePicker-ampm', ampmSelector, null, function (event) {
                     $this.toggleAmPm(event);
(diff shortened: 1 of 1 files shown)


melloware:
Hmmm from your screenshot "this" looks like its the widget because i can see the "cfg" object?  

This is really strange?  Its almost like somewhere you are missing a `var $this = this` and its getting confused?

tandraschko:
(and please dont commit console.log :D)

christophs78:
Everything fine. :-) Just had tomatoes on my eyes and screwed my XHTML.
So only had to remove my (debug) console.log.

christophs78:
Showcase-PR: https://github.com/primefaces/primefaces-showcase/pull/142

melloware (approved):

PR #5844, opened 2020-04-29
Fix #3818: DataTable ability to unsort column

### LINKED ISSUE #3818: DataTable: add ability to unsort column
At the moment, 6.2.5, a user cannot 'unsort' a column, only asc and desc are offered.

This in combination with the nice multiViewState feature forces a logout to undo sorting.

eduarddrenth:
- SortEvent should reflect all three sort states
- wrong message keys
- cycle through asc/desc/unsort

Sorry, merge show line ending changes, cannot get that right


src/main/java/org/primefaces/event/data/SortEvent.java
```diff
@@ -23,31 +23,44 @@
...
 
-import javax.faces.component.UIComponent;
-import javax.faces.component.behavior.Behavior;
 import org.primefaces.component.api.UIColumn;
 import org.primefaces.event.AbstractAjaxBehaviorEvent;
 import org.primefaces.model.SortOrder;
 
+import javax.faces.component.UIComponent;
+import javax.faces.component.behavior.Behavior;
+
```
melloware, line 32: Please put these imports back they way they were it looks like you just formatted the file.  javax. should always come before org. primefaces

melloware (changes requested):
I think you need to start over with a new PR.  You formatted the files of the Java and JS files.    Your PR should contain the bare minimum changes so we can review.  And never format files.

eduarddrenth:
let's try...

Issue #5438, opened 2019-12-12
DataTable: cleanup sorting


tandraschko:
Filters previous had 3 similar attributes:
Filters
FilterBy
FilterMeta

we should check the same for sorting.

All properties related to sort:
* sortMeta
* sortField
* sortBy
* sortOrder
* sortFunction
* sortMode (single | multiple)
* allowUnsorting

Some facts:
1) it seems like sortBy and sortField is similar. One points to a fieldname of the entity, one is a VE? @Rapster Maybe you have a better overview here?
2) if sortMode = multi, sortMeta is the source of all sort informations
3) if sortMode = single, sortBy+sortOrder+sortFunction is used
4) in general, sortMeta is just a collection of (sortBy+sortOrder+sortFunction) for mulitple columns. So multiSort is actually the same thing as singleSort, just that sortMeta could only contain 1 value instead of multiple.

Goals (Reduce attributes to):
* sortBy (a map of SortMeta, multisort will contain mulitple values, singlesort only 1 value)
* initial sorting can by defined by sortOrder and sortPriority on p:column
* allowMultiSort (sortMode before, default = true?)
* allowUnsorting
* remove old attributes from PropertyKeys+taglib and throw a exception in getter/setter

tandraschko:
All properties related to sort:
* sortMeta
* sortField
* sortBy
* sortOrder
* sortFunction
* sortMode (single | multiple)
* allowUnsorting

Some facts:
1) it seems like sortBy and sortField is similar. One points to a fieldname of the entity, one is a VE? @Rapster Maybe you have a better overview here?
2) if sortMode = multi, sortMeta is the source of all sort informations
3) if sortMode = single, sortBy+sortOrder+sortFunction is used
4) in general, sortMeta is just a collection of (sortBy+sortOrder+sortFunction) for mulitple columns. So multiSort is actually the same thing as singleSort, just that sortMeta could only contain 1 value instead of multiple.

Goals (Reduce attributes to):
* sortBy (a map of SortMeta, multisort will contain mulitple values, singlesort only 1 value)
* sortMode (single | multiple)
* allowUnsorting

i think we should even change sortMode to a bool attribute like allowMultiSort, so that allowUnsorting and sortMode are similar. Should it be true per default?

tandraschko:
any thoughts @melloware @christophs78 ?

melloware:
I like the idea.  

[...]

PR #5845, opened 2020-04-29
Fix #3818: Datatable unsort support

### LINKED ISSUE #3818: DataTable: add ability to unsort column

eduarddrenth:
- SortEvent supports all SortOrder
- unsort support in datatable.js
- wrong aria message keys
- docs


melloware (approved):

tandraschko:
WDYT @mertsincan?

eduarddrenth:
Note that this PR may cause work-arounds (like the one I had myself) to fail.

tandraschko:
@eduarddrenth i have the GO from PrimeTek.
See https://primefaces.org/primereact/showcase/#/datatable/sort

Should we also add a removableSort flag?

eduarddrenth:
When removableSort==false don't unsort? For me it is pretty intuitive without it, anyone else?

tandraschko:
in older PrimeFaces it was not easily possible to unsort. So maybe users like to restore the old behavior which dont allow unsort?

eduarddrenth:
Perhaps, to be backwards compatible the default should than be false. Such an option requires a bit of coding and review. Break open the PR? New issue? More discussion?

melloware:
No we should make the new one unsort by default but give the user the ability to turn it off like @tandraschko said.  The new default is a good one.

tandraschko:
yep
i would call the attribute "allowUnsorting"

src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
```diff
@@ -530,6 +531,7 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
 
             var columnHeader = $(this),
             sortOrderData = columnHeader.data('sortorder'),
+                // TODO here we should exclude unsorted based on cfg
```
melloware: Is the TODO meaning you still need to do it?

eduarddrenth: Yes, still working on it in my fork

melloware (approved):

eduarddrenth:
Added below to Primefaces.d.ts for cfg, did not test if everything works:
```
    export interface DataTableCfg extends PrimeFaces.widget.DeferredWidgetCfg {
        /**
         * Default is true: include "unsort" when repeatedly clicking on column header.
         */
        allowUnsorting: boolean;
```

eduarddrenth:
Testing...I get a console error, fixing datatable.js

eduarddrenth:
tested, working, option is working too

src/main/resources/META-INF/primefaces-p.taglib.xml
```diff
@@ -7300,6 +7300,14 @@
             <required>false</required>
             <type>java.lang.String</type>
         </attribute>
+        <attribute>
+            <description>
+                <![CDATA[Defines wether columns are allowUnsorting.]]>
```
melloware, line 7305: Defines whether columns are are allowed to be unsorted. Default is true.

docs/9_0/components/datatable.md
```diff
@@ -106,6 +106,7 @@ DataTable displays data in tabular format.
 | tabindex                  | null               | String           | Position of the element in the tabbing order.
 | tableStyle                | null               | String           | Inline style of the table element.
 | tableStyleClass           | null               | String           | Style class of the table element.
+| allowUnsorting                | true               | Boolean          | Include "unsort" when repeatedly clicking on column header.
```
melloware: Defines whether columns are are allowed to be unsorted. Default is true.

melloware (changes requested):

eduarddrenth:
changed text

melloware (approved):

tandraschko:
thanks, its a great contribution!

eduarddrenth:
on to the next.... "retain order of sort columns with multi column sort and multiviewstate"

Rapster:
Are you sure you want to make allowUnSorting to true by default? As I'm using sorting a lot lately, I must say it's confusing whenever i click on column's header the result does not change while I expect it to be sorted asc or desc. I quite agree with Thomas statement:
> in older PrimeFaces it was not easily possible to unsort. So maybe users like to restore the old behavior which dont allow unsort?

Because it was not possible before, and it's more intuitive to have this transition: `UNSORTED > ASC > DESC > ASC > DESC`, I would put this property to false by default, WDYT?

eduarddrenth:
Changes can be confusing to people yes, but for me uns->asc->desc->uns is more intuitive.

Anyway, when allowUnSorting is false by default there will be no (expected) complaints and the complaint
"I cannot unsort" can be dealt with.

Rapster:
> Changes can be confusing to people yes, but for me uns->asc->desc->uns is more intuitive.

I'm not talking changes in general, but this one specifically. I think most users expect the scenario I mentionned, you can see few examples [here](https://mdbootstrap.com/docs/jquery/tables/sort/) and [there](https://datatables.net/), and excel does almost the same

> Anyway, when allowUnSorting is false by default there will be no (expected) complaints and the complaint
"I cannot unsort" can be dealt with.

Exactly, and the idea is to have as less people complaining as possible (right? 😅) so allowUnSorting to false by default makes more sense no?

stolp:
It definitely makes sense not to change the default behavior. As a framework consumer I hate surprising changes without my consent.

melloware:
So are we fixing this? I agree with @stolp and @Rapster

Rapster:
It's in my todo list for #5438 😉

eduarddrenth:
way to go

PR #5855, opened 2020-05-01
Fix #5846: aria-hidden and aria-label for inputs

### LINKED ISSUE #5846: Accessibility: Hidden Inputs should have aria-hidden="true"
If you run Google Lighthouse on some components like SelectCheckBoxMenu.  its hidden inputs are being found by ScreenReaders when they should be `aria-hidden="true"` or should use `aria-label` if the label is known.

image

Components:
- [x] AutoComplete
- [x] SelectCheckBoxMenu
- [x] SelectOneMenu (Editor)
- [x] SelectBooleanButton
- [x] SelectOneButton
- [x] SelectOneListBox
- [x] Picklist
- [x] SelectManyButton
- [x] SelectManyMenu
- [x] SplitButton
- [x] Tree

melloware:


tandraschko:
i think mert should review it
our hidden inputs are mostly used for getting a correct focus behavior when tabbing

can you also try if it doesn't destroy it? Not sure if a normal browser just ignores the aria-hidden

tandraschko:
can you fix the conflicts?

melloware:
Resolved!  Thanks for pointing that out.

PR #5858, opened 2020-05-03
Fix #5847, register native event listener for view change

### LINKED ISSUE #5847: Schedule: Issues with "customButtons" when set via the extender
## Environment
Tried it with PF 8.0, but looking at the source, most likely also present in the current master.

## Issue
I was trying to add [custom buttons](https://fullcalendar.io/docs/customButtons) to the schedule / fullcalendar header, by using the `extender` option of the schedule widget:

```xhtml
<p:schedule extender="function(){this.cfg.customButtons={myButton:{text:"foobar"}}}" centerHeaderTemplate="myButton"/>
```

Now that works and adds the button to the center of the schedule header. There is one problem, however, that occurs once an AJAX update is performed. The schedule disappears and the following error message is logged:

```
Uncaught Error: The FullCalendar view "myButton" does not exist. Make sure your plugins are loaded correctly.
```

## Cause

Looking around in the source code a bit, the problem is that when the widget is refreshed, it attempts to pass `myButton` as the value of the [defaultView](https://fullcalendar.io/docs/defaultView) option, which is not allowed.

The schedule components saves the currently selected view in a hidden input field, so that the view can be restored after an AJAX update, see [1-schedule.js#241-263](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/schedule/1-schedule.js#L241). The code there in `bindViewChangeListener`

* iterates over all buttons
* excludes some predefined buttons
* parses the style classes of the buttons to figure out the view of that button
* saves that view via $.fn.data
* listens for a click event on the button
* and finally reads the view via $.fn.data from the button and saves it

Since custom buttons are, well, custom, they are not excluded in step 2, so `myButton` is saved as the view and later set as the `defaultView`, which causes the error.

## Possible solution

This process seems a bit convoluted to me - am I missing something? I took a quick look at the full calendar API and found some events such as [viewSkeletonRender](https://fullcalendar.io/docs/viewSkeletonRender). It seems to be fired when the view is changed:

```javascript
PF("schedule").calendar.on("viewSkeletonRender", ev => console.log(ev.view.type))
```

blutorange:
Uses the native `viewSkeletonRender`  event to listen for when the view changes. This way, custom buttons are not wrongly considered as view buttons. Also reduces the amount of code.

See primefaces/primefaces-showcase#144 for an update to the showcase to reproduce the issue.

### DIFF
--- a/docs/9_0/components/inputtextarea.md
+++ b/docs/9_0/components/inputtextarea.md
@@ -59,7 +59,7 @@ size | null | Integer | Number of characters used to determine the width of the 
 style | null | String | Inline style of the input element.
 styleClass | null | String | Style class of the input element.
 tabindex | null | Integer | Position of the input element in the tabbing order.
-title | null | String | Advisory tooltip informaton.
+title | null | String | Advisory tooltip information.
 autoResize | true | Boolean | Specifies auto growing when being typed.
 maxlength | null | Integer | Maximum number of characters that may be entered in this field.
 counter | null | String | Id of the output component to display remaining chars.
--- a/docs/9_0/components/schedule.md
+++ b/docs/9_0/components/schedule.md
@@ -51,7 +51,7 @@ timeZone | null | Object | String or a java.time.ZoneId instance to specify the 
 tooltip | false | Boolean | Displays description of events on a tooltip.
 clientTimeZone | local | String | Defines how to interpret the dates at browser. Valid values are "local", "UTC" and ids like "America/Chicago". (see https://fullcalendar.io/docs/timeZone)
 showWeekNumbers | false | Boolean | Display week numbers in schedule.
-extender | null | String | Name of javascript function to extend the options of the underlying fullcalendar plugin.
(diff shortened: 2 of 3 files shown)


melloware (approved):

melloware:
@christophs78 mind reviewing this PR?

christophs78:
Looks good! And great examples as part of your ShowCase-PR.
Can´t find regressions. :-)

Really happy to look at your PR´s and the "new blood" you bring into this project. (Hope i brought some "new blood" into this project during the last year.) When i look at your work, especially your JsDoc-work this is absolute awesome!

blutorange:
@christophs78 I'm glad if I can contribute a bit to this project. People from different backgrounds working together can be helpful. Well, depending on the which projects I'm working on, I might have more or less time for this project, but I hope I can contribute at least a bit from time to time : )

tandraschko:
Thanks!

PR #5860, opened 2020-05-03
Implement #5859: PrimeFaces.ajax.Request.handle returns promise-like

### LINKED ISSUE #5859: Core / Ajax: possibility for remote commands to return a promise
Remote commands are useful in many cases, including when you have some custom JavaScript for client-side interaction and want to call a bean method. Running a remote command works fine - we just have to call the function created in the global window scope.

One thing I found lacking, however, is the ability to register a callback for when the remote command completes. Yes, I can add callback via the `onsuccess` or `oncomplete` attributes, but the problem here is that this allows only for one specific callback to be set. Sometimes I need to register a callback dynamically from the calling JavaScript code, or when binding symbols via a closure. For the time being, I made a quick hack by overriding some of the `PrimeFaces.ajax` methods, but a proper solution would be great.

Implementing doesn't seem to be particularly hard, I made a demonstration here: https://github.com/blutorange/primefaces/commit/b665c40fea17ff7bf0cadaa1e58a0fbfdc6b6dbd (Works also when client side validation is on. Creates a `$.Deferred` object, saves it in the `cfg` object, and resolves or rejects it when the AJAX request is done.)

image

Any opinions on this? I could make a PR if there aren't any objections.

blutorange:
This implements #5859.

* I changed the method `PrimeFaces.ajax.Request.handle`. It now create a `$.Deferred` object and saves it in the `cfg` object, and returns it. Later, when the AJAX request is done, that promise is resolved, or rejected when the request failed.
* I also modified the `RemoteCommandRenderer` so that calling a remote command from the client now also returns that promise.
* I also updated the docs, see `remoteCommand.md` for an example.
* See also primefaces/primefaces-showcase#145

I also made sure this works when client-side validation is enabled (`PrimeFaces.vb` instead of `PrimeFaces.ab`). Currently, when csv is enabled, the remote command returns `false` when validation fails. In order not to break things, I left that as is and return the Promise only when validation succeeded.  Or do you think that we can change this - and return a rejected promise when validation fails? That is, change

```javascript
rc = function() {
        if (PrimeFaces.vb({
            s: this,
            a: true,
            p: 'j_idt720:name j_idt720:number',
            u: 'j_idt720:grid'
        })) {
            return pf.ab({
                s: "j_idt720:j_idt759",
                f: "j_idt720",
                p: "j_idt720:name j_idt720:number",
                u: "j_idt720:grid",
                pa: arguments[0]
            });
        } else {
            return false;
        }
    }
```

to this:

```javascript
rc = function() {
        if (PrimeFaces.vb({
            s: this,
            a: true,
            p: 'j_idt720:name j_idt720:number',
            u: 'j_idt720:grid'
        })) {
            return pf.ab({
                s: "j_idt720:j_idt759",
                f: "j_idt720",
                p: "j_idt720:name j_idt720:number",
                u: "j_idt720:grid",
                pa: arguments[0]
            });
        } else {
            return $.Deferred().reject({textStatus: 'error', errorThrown: 'Client-side validation failed.'});
        }
    }
```

### DIFF
--- a/docs/9_0/components/remotecommand.md
+++ b/docs/9_0/components/remotecommand.md
@@ -57,14 +57,74 @@ RemoteCommand is used by invoking the command from your javascript code.
     }
 </script>
 ```
+
 That’s it whenever you execute your custom javascript function(eg customfunction()), a remote call
 will be made, actionListener is processed and output text is updated. Note that remoteCommand
 must be nested inside a form.
 
 
-## Passing Parameters
+## Passing parameters
+
 Remote command can send dynamic parameters in the following way;
 
 ```js
 increment([{name:'x', value:10}, {name:'y', value:20}]);
 ```
+
+To access these parameters in a bean method:
+
+```java
+    public void execute() {
+        String param1 = FacesContext.getCurrentInstance().getExternalContext().getRequestParameterMap().get("x");
+        String param2 = FacesContext.getCurrentInstance().getExternalContext().getRequestParameterMap().get("y");
+        // Do something with the parameters
+        FacesContext.getCurrentInstance().addMessage(null, new FacesMessage(FacesMessage.SEVERITY_INFO, "Executed", "x: " + param1 + ", y: " + param2));
+    }
+```
+
+## Receiving parameters
+
+You can receive parameters from the server in the following way.
+
+First, in your bean method, add a callback parameter:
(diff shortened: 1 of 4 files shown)


tandraschko:
Thanks!

PR #5862, opened 2020-05-04
Fix #5861: ImageCropper file path issues on Unix

### LINKED ISSUE #5861: ImageCropper: Error on Unix systems
## 1) Environment
- PrimeFaces version: 8.0
- Does it work on the newest released PrimeFaces version? Version? NO
- Application server + version: Any UNIX OS
- Affected browsers: ALL

## 2) Expected behavior

Cropping should work on UNIX systems.

## 3) Actual behavior

Based on forum post: https://forum.primefaces.org/viewtopic.php?f=3&t=61847

The problem I am having is with the imageCropper component, in version 6.2 it worked correctly for me but after updating it started to fail. The server log shows me the following warning:

> 17: 19: 55,183 WARNING [javax.enterprise.resource.webcontainer.jsf.lifecycle] (default task-13) Path traversal attempt - absolute path not allowed .: javax.faces.FacesException: Path traversal attempt - absolute path not allowed.

```
java.io.FileNotFoundException: /home/user/Documents/Primefaces/pf-3286/target/primefaces-test-1.0-SNAPSHOTresources/images/nature6.jpg (No such file or directory)
```

## 4) Steps to reproduce

Showcase: http://localhost:8080/showcase/ui/multimedia/cropper.xhtml

melloware:
Checks if the last character of webroot is a file separator "/" and if not appends it.

### DIFF
--- a/src/main/java/org/primefaces/component/imagecropper/ImageCropperRenderer.java
+++ b/src/main/java/org/primefaces/component/imagecropper/ImageCropperRenderer.java
@@ -207,7 +207,8 @@ public class ImageCropperRenderer extends CoreRenderer {
 
                     String webRoot = externalContext.getRealPath(Constants.EMPTY_STRING);
                     String fileSeparator = Constants.EMPTY_STRING;
-                    if (webRoot.charAt(webRoot.length() - 1) != File.separatorChar) {
+                    if (webRoot.charAt(webRoot.length() - 1) != File.separatorChar &&
+                                imagePath.charAt(0) != File.separatorChar) {
                         fileSeparator = File.separator;
                     }
 
--- a/src/main/java/org/primefaces/component/imagecropper/ImageCropperRenderer.java
+++ b/src/main/java/org/primefaces/component/imagecropper/ImageCropperRenderer.java
@@ -207,8 +207,9 @@ public class ImageCropperRenderer extends CoreRenderer {
 
                     String webRoot = externalContext.getRealPath(Constants.EMPTY_STRING);
                     String fileSeparator = Constants.EMPTY_STRING;
-                    if (!webRoot.endsWith(File.separator) && !imagePath.startsWith(File.separator)) {
-                        fileSeparator = File.separator;
+                    if (!(webRoot.endsWith("\\") || webRoot.endsWith("/")) &&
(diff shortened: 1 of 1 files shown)


tandraschko:
what is if imagePath starts with a seperator already?

melloware:
good point let me check both sides.

tandraschko:
also startsWith/endsWith might be nicer as indexOf

melloware:
That is a good suggestion it reads much nicer.

tandraschko:
another thing... imagePath - does it contain the seperator from OS or the normal slash http/java seperator?
Otherwise i could fail on windows if you check for the normal fileseperator

melloware:
No this is an OS file path so its `\` on windows and `/` and Unix which is why I use File.separator.

tandraschko:
are you sure?
every java developer likely uses a normal slash for pathes inside the java/JSF app for resource paths.

melloware:
Hmmm OK let me debug on Windows.

melloware:
I just dumped the result on Windows and here is what it does "C:\dev\primefaces\primefaces-showcase\src\main\webapp\/resources/demo/images/nature/nature6.jpg"

And this is without any changes to the code and it works fine.  Notice the slahses switch midway.

After my code change the exact same URL comes out:

"C:\dev\primefaces\primefaces-showcase\src\main\webapp\/resources/demo/images/nature/nature6.jpg"

So I think its safe to use File.separator than to hardcode `/`  I definitely don't mind hardcoding `/` but it seemed wrong.

tandraschko:
yeah
but your if checks now if imagePath starts with a backslash, which is likely always a normal slash

i would check for both slashes actually

melloware:
OK latest versions checks both sides for all slashes and appends the `/` if missing.  Tested on Windows and working fine.

tandraschko:
+1 now

tandraschko:
in theory its enough to check for the OS FileSeperator in webRoot (as it receveived from servletContext#getRealPath) - imagePath could contain both

melloware:
Thanks i think this errors on the side of safety to make sure the path is valid.

tandraschko:
@melloware imagePath needs startsWith, not endsWith?

melloware:
thanks PR submitted.

PR #5870, opened 2020-05-06
Fix #5869: InputText/InputTextArea validate maxlength

### LINKED ISSUE #5869: InputText: user input not correctly validated with respect to maxlength
## 1) Environment
- PrimeFaces version: 8.0

## 2) Expected behavior

p:inputText maxlength should be validated server sided.

## 3) Actual behavior

The validation happens exclusively on the client side. Note that this was already discovered in #4420 but apparently only fixed for p:inputTextarea.
Credits to @cnsgithub for the original discovery and original PoC code.

Note that I cannot serve with a PR as I am too unfamiliar with the project to fix this in a responsible way and do not want to invest the time to get familiar enough.

## 4) Steps to reproduce

Using the example below (after inserting into the primefaces-test project), click the Hack button.

## 5) Sample XHTML

```
<!DOCTYPE html>

  <h:head>
  </h:head>
  <h:body>

    <h:form id="frmTest">
      <h1>#{testView.testString}</h1>
      <p:inputText value="#{testView.testString}" maxlength="10" />
      <p:commandButton type="submit" value="Submit" update="frmTest" />
      <p:commandButton type="submit" value="Hack" update="frmTest" onclick="document.querySelector('input[maxlength=&quot;10&quot;]').value='a'.repeat(100);" />
    </h:form>

  </h:body>
</html>
```

## 6) Sample bean

```
package org.primefaces.test;

import java.io.Serializable;
import javax.inject.Named;

@Named
@ViewScoped

  private String testString;

  @PostConstruct  
  public void init() {
    testString = "Welcome!";
  }

    return testString;
  }

    System.err.println("received input: " + testString + " with length: " + testString.length());
    this.testString = testString;
  }
}
```

melloware:
@tandraschko this PR does two things. one issue while I was fixing the validation issue.

1. Validation: Fixes the input validation to respect the maxlength property similar to InputTextArea.

2. Performance: The maxlength property of inherited  `HtmlInputText` is actually set to Integer.MIN_VALUE. Our Widget code was doing `.attr("maxlength", inputText.getMaxlength(), Integer.MAX_VALUE)` so that means for every single input on the screen it was always sending maxlength like.
```javascript
pf.cw("InputText","widget_j_idt718_j_idt720",{id:"j_idt718:j_idt720",maxlength:-2147483648});
```

So now with this PR the maxlength property is ONLY sent if it is set by the user.
```javascript
pf.cw("InputText","widget_j_idt723_j_idt725",{id:"j_idt718:j_idt720"});
pf.cw("InputText","widget_j_idt723_j_idt727",{id:"j_idt723:j_idt727",maxlength:10});
```

### DIFF
--- a/src/main/java/org/primefaces/component/inputtext/InputTextRenderer.java
+++ b/src/main/java/org/primefaces/component/inputtext/InputTextRenderer.java
@@ -51,6 +51,10 @@ public class InputTextRenderer extends InputRenderer {
         String submittedValue = context.getExternalContext().getRequestParameterMap().get(clientId);
 
         if (submittedValue != null) {
+            int maxlength = inputText.getMaxlength();
+            if (maxlength > 0 && submittedValue.length() > maxlength) {
+                return;
+            }
             inputText.setSubmittedValue(submittedValue);
         }
     }
@@ -68,7 +72,7 @@ public class InputTextRenderer extends InputRenderer {
         String counter = inputText.getCounter();
         WidgetBuilder wb = getWidgetBuilder(context);
         wb.init("InputText", inputText.resolveWidgetVar(context), clientId)
-                .attr("maxlength", inputText.getMaxlength(), Integer.MAX_VALUE);
+                .attr("maxlength", inputText.getMaxlength(), Integer.MIN_VALUE);
 
         if (counter != null) {
             UIComponent counterComponent = SearchExpressionFacade.resolveComponent(context, inputText, counter);
--- a/src/main/java/org/primefaces/component/inputtext/InputTextRenderer.java
+++ b/src/main/java/org/primefaces/component/inputtext/InputTextRenderer.java
@@ -33,6 +33,7 @@ import org.primefaces.expression.SearchExpressionFacade;
 import org.primefaces.renderkit.InputRenderer;
 import org.primefaces.util.ComponentUtils;
 import org.primefaces.util.HTML;
+import org.primefaces.util.LangUtils;
 import org.primefaces.util.WidgetBuilder;
 
 public class InputTextRenderer extends InputRenderer {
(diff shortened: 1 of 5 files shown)


tandraschko:
should we really just "return;" or cut the submmited value?

melloware:
I can cut the submitted value if you would prefer and I can modify InputTextArea to do the same. Right now its just returning with no change.    Or can we throw some kind of validation error there?

tandraschko:
no idea
actually validation happens based on the submittedValue

we can also try to overwrite validateValue or something

tandraschko:
i would just cut it

melloware:
OK sold I will update this and InputTextArea to just cut it.

tandraschko:
why do set a default in the JS now?
Also it would be great if you can do a unittest for the langutils

melloware:
We set a default in the JS because i the user sets a "counter" that is the only thing that uses the maxlength in the JS file to count down the characters as you type.  Previously it had a value so I was being a good citizen and setting the maxlength to Integer.MAX_VALUE if the user did not set it.  I can move this code down into the "counter" code but it is not how we do it in other widgets?

tandraschko:
nvm, its ok then

tandraschko:
@melloware 
AFAICS getMaxlength default is Integer.MIN_VALUE, which is this case:

assertEquals("abcdefg", LangUtils.substring("abcdefg", 0, Integer.MIN_VALUE));

and it fails

melloware:
@tandraschko you are correct the default is Integer.MIN_VALUE but that is why in both Inputtext and InputTextArea i only substring if its a positive number. a negative maxlength is ignored.

```java
if (maxlength > 0 && submittedValue.length() > maxlength) {
       submittedValue = LangUtils.substring(submittedValue, 0, maxlength);
}
```

tandraschko:
ahh yep :D

tandraschko:
Thanks
just ignored the first IF part...

mertsincan:
> why do set a default in the JS now?

I agree with @tandraschko. I think we shouldn't set the maxlength's value. Because this value can vary from browser to browser. For example, the link below says that the default value is 524288. I'm not sure :D So, we can tell users to set the maxlength attribute if they use this feature. Exp; a warning message or if-statement check in widget can be used.

https://www.geeksforgeeks.org/html-input-maxlength-attribute/

melloware:
@mertsincan no problem would you like me to submit another PR with this fix?

melloware:
@mertsincan new PR submitted here: https://github.com/primefaces/primefaces/pull/5895

PR #5881, opened 2020-05-09
Fix #5880: getConvertedValue now takes Object instead of String

### LINKED ISSUE #5880: ComponentUtils: Improve getConvertedValue()
ComponentUtils.getConverted value right now expects a String value.  Which leads most components to cast it as String before passing it to the method.

```java
String submittedValueString = (String) submittedValue;
Object convertedValue = ComponentUtils.getConvertedValue(context, component, submittedValueString);
```

Instead of just..
```java
Object convertedValue = ComponentUtils.getConvertedValue(context, component, submittedValue);
```

And letting the method handle the String casting.  Its a cleaner and more flexible method.

melloware:
@Rapster please review

### DIFF
--- a/src/main/java/org/primefaces/util/ComponentUtils.java
+++ b/src/main/java/org/primefaces/util/ComponentUtils.java
@@ -172,9 +172,9 @@ public class ComponentUtils {
     }
 
     public static Object getConvertedValue(FacesContext context, UIComponent component, Object value) {
-        String submittedValue = null;
-        if (value != null) {
-            submittedValue = value.toString();
+        String submittedValue = Objects.toString(value, null);
+        if (LangUtils.isValueBlank(submittedValue)) {
+            submittedValue = null;
         }
 
         Converter<?> converter = getConverter(context, component);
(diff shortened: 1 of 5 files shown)


src/main/java/org/primefaces/util/ComponentUtils.java
```diff
@@ -171,10 +171,15 @@ public static Converter getConverter(FacesContext context, UIComponent component
         return context.getApplication().createConverter(converterType);
     }
 
-    public static Object getConvertedValue(FacesContext context, UIComponent component, String value) {
-        Converter converter = getConverter(context, component);
+    public static Object getConvertedValue(FacesContext context, UIComponent component, Object value) {
+        String submittedValue = null;
```
Rapster: You could use `String submittedValue = Objects.toString(value, null);` and also, make the change for every caller? (e.g EditorRenderer, SelectCheckboxMenu, etc.)

melloware:
Excellent suggestion @Rapster I did do one more thing though with a LangUtils.isValueBlank() check that InputNumber was doing as well.  Please review again.

PR #5885, opened 2020-05-11
Fix #5882: ThemeSwitcher not respecting all SelectOne attributes

### LINKED ISSUE #5882: ThemeSwitcher: Not respecting SelectOne attributes
http://primefaces.org/showcase/ui/misc/themeSwitcher.xhtml

If you open the first selectOneMenu, the second selectOneMenu is displayed OVER the overlay of the first.

Can someone check it?

melloware:
So I studied ThemeSwitcher and SelectOne and realized there were lots of problems.  If the ThemeSwitcher was at the bottom of the screen the dropdown panel is supposed to render upwards because it detects the bottom of the screen.  That also wasn't happening along with the other issues.

This fix fixes ALL those issues.


tandraschko:
+1 thanks!

PR #5888, opened 2020-05-12
Fix #5883: Theme documentation

### LINKED ISSUE #5883: Add themes to documentation
AFAICS we dont have a list of build-it themes and small description in our docs

We should also link the old legacy themes projects with all of the old themes

Would be great if someone of the community can do it

melloware:


tandraschko:
+1
perfect, thanks

PR #5894, opened 2020-05-12
Fix #5893: InputRenderer createStyleClass methods

### LINKED ISSUE #5893: InputRenderer: new createStyleClass method
We have a lot of code in Renderer's doing very similar things that looks almost exactly like this.

```java
protected String createStyleClass(InputText inputText) {
        String defaultClass = InputText.STYLE_CLASS;
        defaultClass = inputText.isValid() ? defaultClass : defaultClass + " ui-state-error";
        defaultClass = !inputText.isDisabled() ? defaultClass : defaultClass + " ui-state-disabled";

        String styleClass = inputText.getStyleClass();
        styleClass = styleClass == null ? defaultClass : defaultClass + " " + styleClass;

        return styleClass;
    }
```

To me this is a perfect candidate for moving up to `InputRenderer`.  PR to follow.

melloware:
- Optimized using a StringBuilder vs string concatenation
- Code much easier to read
- Deletes a lot of boilerplate code in Input components

### DIFF
--- a/src/main/java/org/primefaces/component/autocomplete/AutoCompleteRenderer.java
+++ b/src/main/java/org/primefaces/component/autocomplete/AutoCompleteRenderer.java
@@ -190,7 +190,7 @@ public class AutoCompleteRenderer extends InputRenderer {
         String itemLabel;
         String inputStyle = ac.getInputStyle();
         String defaultStyleClass = ac.isDropdown() ? AutoComplete.INPUT_WITH_DROPDOWN_CLASS : AutoComplete.INPUT_CLASS;
-        String inputStyleClass = createStyleClass(ac, "inputStyleClass", defaultStyleClass);
+        String inputStyleClass = createStyleClass(ac, AutoComplete.PropertyKeys.inputStyleClass.name(), defaultStyleClass);
         String autocompleteProp = (ac.getAutocomplete() != null) ? ac.getAutocomplete() : "off";
 
         writer.startElement("input", null);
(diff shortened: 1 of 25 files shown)


tandraschko:
Instead of "inputStyleClass" we should pass PropertyKeys.inputStyleClass.name()

Also, can you try it with EL in inputStyleClass and/or mixed with static strings?
I can't remember in which case getAttributes/createValueExpression/getStateHelper does (not) work.

Also please try MyFaces Mojarra

melloware:
I switched to use Enum.name() for the inputStyleClass ones.  I read the Javadoc of getAttributes and even though its an Object key of the Map it says...
"Return a mutable Map representing the attributes(and properties, see below) associated with this UIComponent,keyed by attribute name **(which must be a String)**. "

So it has to be a String property so `.name()` is the correct solution.

tandraschko:
What happens if you have:
inputStyleClass="#{mybean.prop}"
will it be resolved correctly when you access it via attributes?

melloware:
Tested with this...
```xml
<p:inputText styleClass="#{false ? 'true-style' : 'false-style'}" required="true" />
```
And it rendered `ui-inputfield ui-inputtext ui-widget ui-state-default ui-corner-all false-style`

And then with..
```xml
<p:inputText styleClass="#{true ? 'true-style' : 'false-style'}" required="true" />
```
And it rendered `ui-inputfield ui-inputtext ui-widget ui-state-default ui-corner-all true-style`

tandraschko:
then +1
but please switch to SharedStringBuilder

melloware:
Updated to SharedStringBuilder. Good suggestion!

tandraschko:
+1 if well tested

PR #5910, opened 2020-05-17
Fix #5614: Datatable sorting taglib and docs

### LINKED ISSUE #5438: DataTable: cleanup sorting

### LINKED ISSUE #5614: DataTable: Initial multi-column sorting using SortMeta doesn't highlight column headers
There's a forum conversation for this issue: https://forum.primefaces.org/viewtopic.php?f=3&t=61170

There's also a fork to a PF test case detailing the problem at https://github.com/kawoolutions/primefaces-test

Note the `multisort` dirs under Java and webapp dirs, for beans + XHTML respectively.

## 1) Environment
- PrimeFaces version: 8.0.RC3
- Does it work on the newest released PrimeFaces version? Version? No. 8.0.RC3
- Application server + version: Any
- Affected browsers: Firefox (others untested)

## 2) Expected behavior

Usually the given SortMeta columns for a 

```
<p:dataTable ...
                     sortBy="#{multiSortView.getInitialSortMetas()}"
                     sortMode="multiple">
```
shall be highlighted with `ui-state-active`, which doesn't seem to happen.

## 3) Actual behavior

No highlighting is present.

## 4) Steps to reproduce

Just call the page multisort.xhtml as in the test case, the initial table view suffices to see the effects.

## 5) Sample XHTML

https://github.com/kawoolutions/primefaces-test/tree/master/src/main/webapp/multisort

## 6) Sample bean

See 3 files here: https://github.com/kawoolutions/primefaces-test/tree/master/src/main/java/org/primefaces/test/multisort

melloware:
- add `sortMeta` to taglib
- updated all sort fields docs

### DIFF
--- a/docs/8_0/components/datatable.md
+++ b/docs/8_0/components/datatable.md
@@ -2,6 +2,8 @@
 
 DataTable displays data in tabular format.
 
+[See this widget in the JavaScript API Docs.](../jsdocs/classes/primefaces.widget.datatable.html)
+
 ## Info
 
 | Name | Value |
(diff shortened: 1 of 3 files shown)


melloware:
@tandraschko do you mind reviewing this one its just taglib and docs changes.  I don't think we needed to do anything else?

tandraschko:
thats ok for me currently
we need to refactor the datatable to remove sortMeta in the future and resuse sortBy for both cases

melloware:
ahh OK let me leave this open then. I wasn't clear what the goal was but reusing SortBy does seem like a good idea.

melloware:
OK for now this clears up the confusion on #5614 so merging and closing that ticket.

Issue #5913, opened 2020-05-18
Calendar: java.lang.Object not supported


ohlsen3000:
**Describe the defect**
Evaluating PF8 I started to remove legacy LocalDateConverter for all Calendars, since it is not required anymore. It works so far. But if the the LocalDate value resides in a generic structure it causes the error mentioned above. 

**Environment:**
 - PF Version: _8.0_
 - JSF + version: _MyFaces 2.2.10_
 - Affected browsers: _ALL_

**To Reproduce**
1. Use a calendar that is connected to a LocalDate in a generic class
2. Do not use a custom converter

**Expected behavior**
LocalDate is written to the bean without an error or failed validation

**Example XHTML**
```xml
<p:calendar value="#{myView.rangeDate.min}"/>
```

**Example Bean**
```java
@Named
@WindowScoped
public class MyView  implements  Serializable {

    Range<LocalDate> rangeDate = new Range<>();

    public Range<LocalDate> getRangeDate() {
        return rangeDate;
    }

    public void setRangeDate(final Range<LocalDate> rangeDate) {
        this.rangeDate = rangeDate;
    }
}


public class Range<T> implements Serializable {

    private static final long serialVersionUID = 1L;

    private transient T min;
    private transient T max;

    public T getMin() {
        return min;
    }

    public void setMin(final T min) {
        this.min = min;
    }

    public T getMax() {
        return max;
    }

    public void setMax(final T max) {
        this.max = max;
    }
}
```

tandraschko:
Please ALWAYS try a trunk version before creating a issue. 
I remember that we already fixed something regarding generics.

ohlsen3000:
Just checked current master (9.0-SNAPSHOT), same problem there.

tandraschko:
@christophs78 would you like to check this?
i think we cant fix it easily. We would need to get the parent field etc etc.
Mabye its better to introcude a attribute, to pass the type. like valueType="....LocalDate"

christophs78:
I will look at this within the next week or so.
During our iterations between 7 and 8 we had something like the valueType for some time. But we removed it because we thought/hoped we dont need it. I will look how easy/hard it´s with our current codebase to re-introduce such an attribute.

christophs78:
@ohlsen3000 : Can you work around this this by using `f:convertDateTime`?

```
<p:calendar value="#{myView.rangeDate.min}"/>
    <f:convertDateTime type="localDate" />
</p:calendar>
```

[...]

PR #5925, opened 2020-05-20
Fix #5923: Prevent native date converter errors

### LINKED ISSUE #5923: Calendar: Calendar using java.util.Date and jsf native converter crashes on conversion error
**Describe the defect**
`p:calendar ` component will crash when updating page after converter exception due to internal conversion failure of `LocalDateTime.now()`
In `CalendarRenderer`/`DatePickerRenderer `line [90 ](https://github.com/primefaces/primefaces/blob/076b68be327546432b1c033c18ddfea789f8defa/src/main/java/org/primefaces/component/calendar/CalendarRenderer.java#L90) it tries to convert `defaultDate ` using `CalendarUtils `which in turn throws `ConversionException ` on its own being unable to convert `LocalDateTime` into string since the `CalendarUtils` delegates the conversion to `DateTimeConverter` which fails to convert new java date objects.

Not sure where the bug is but at very least it should catch the potential default date conversion in Renderer as it has no control over the outcome.

**Reproducer**

**Environment:**
 - PF Version: _8.0_
 - JSF + version: _e.g. Mojarra 2.3.9 

**To Reproduce**
use `p:calendar ` component in combination with native jsf dateconverter and invoke conversion error on page. There is javascript error in console while updating the page.

**Expected behavior**
At very least - not get the error in page, resulting in page not being rendered correctly. 

**Example XHTML**
```xml
<p:calendar value="#{regularJavaUtilDateObj}">
    <f:convertDateTime />
</p:calendar>
```
**Example Java**
```java
if (calendar.isConversionFailed()) {
   defaultDate = CalendarUtils.getValueAsString(context, calendar, LocalDateTime.now());
}
```
**Workaround**
Managed to get around it by defining custom local _Renderer_ overrides for `DatePickerRenderer/CalendarRenderer ` in _faces-config.xml_ but this requires duplicating the entire `encodeScript` method. 
```java
if (calendar.isConversionFailed()) {
   defaultDate = CalendarUtils.getValueAsString(context, calendar, new Date());
}
```
The latter works in my specific situation but it might not work for everyone.

melloware:
@christophs78 Not sure if this is the best way to fix it but.

1. Used our new `CalendarUtils.now(cal)` method instead of LocalDateTime.now() so it calculates NOW in the calendar's timezone.

2. I basically eat the conversion exception and log it and fall back to built in conversion.

### DIFF
--- a/src/main/java/org/primefaces/util/CalendarUtils.java
+++ b/src/main/java/org/primefaces/util/CalendarUtils.java
@@ -236,7 +236,7 @@ public class CalendarUtils {
     }
 
     public static final String getValue(FacesContext context, UICalendar calendar, Object value, String pattern) {
-      //first ask the converter, if it fails fall back to built-in conversion
+        //first ask the converter
         if (calendar.getConverter() != null) {
             return calendar.getConverter().getAsString(context, calendar, value);
         }
@@ -530,7 +530,11 @@ public class CalendarUtils {
      */
     public static Temporal now(UICalendar uicalendar) {
         boolean hasTime = uicalendar.hasTime();
+        boolean timeOnly = uicalendar.isTimeOnly();
         ZoneId zone = calculateZoneId(uicalendar.getTimeZone());
+        if (hasTime && timeOnly) {
+            return LocalTime.now(zone);
+        }
         return hasTime ? LocalDateTime.now(zone) : LocalDate.now(zone);
     }
 
@@ -547,6 +551,9 @@ public class CalendarUtils {
             java.util.Date date;
             if (now instanceof LocalDate) {
                 date = java.util.Date.from(((LocalDate) now).atStartOfDay(zone).toInstant());
+            }
+            else if (now instanceof LocalTime) {
+                date = java.util.Date.from(((LocalTime) now).atDate(LocalDate.now(zone)).atZone(zone).toInstant());
             }
             else {
                 date = java.util.Date.from(((LocalDateTime) now).atZone(zone).toInstant());
--- a/src/main/java/org/primefaces/util/CalendarUtils.java
+++ b/src/main/java/org/primefaces/util/CalendarUtils.java
@@ -44,9 +44,12 @@ import java.util.Date;
 import java.util.List;
 import java.util.Locale;
 import java.util.TimeZone;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 public class CalendarUtils {
 
+    private static final Logger LOGGER = Logger.getLogger(CalendarUtils.class.getName());
     private static final String[] TIME_CHARS = {"H", "K", "h", "k", "m", "s"};
 
     private static final PatternConverter[] PATTERN_CONVERTERS =
@@ -236,10 +239,18 @@ public class CalendarUtils {
 
     public static final String getValue(FacesContext context, UICalendar calendar, Object value, String pattern) {
         //first ask the converter
-        if (calendar.getConverter() != null) {
-            return calendar.getConverter().getAsString(context, calendar, value);
+        try {
+            if (calendar.getConverter() != null) {
+                return calendar.getConverter().getAsString(context, calendar, value);
+            }
         }
-        else if (value instanceof String) {
+        catch (Exception e) {
+            if (LOGGER.isLoggable(Level.FINE)) {
+                LOGGER.fine(String.format("Could not convert date value...defaulting to built-in converter"));
+            }
+        }
+
+        if (value instanceof String) {
             return (String) value;
         }
         //Use built-in converter
--- a/src/main/java/org/primefaces/util/CalendarUtils.java
+++ b/src/main/java/org/primefaces/util/CalendarUtils.java
@@ -529,12 +529,12 @@ public class CalendarUtils {
      * @return a Temporal representing either a Date or DateTime
      */
     public static Temporal now(UICalendar uicalendar) {
-        boolean hasTime = uicalendar.hasTime();
         boolean timeOnly = uicalendar.isTimeOnly();
         ZoneId zone = calculateZoneId(uicalendar.getTimeZone());
-        if (hasTime && timeOnly) {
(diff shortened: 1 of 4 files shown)


christophs78:
I don´t think we should do it this way.
`CalendarUtils.java#getValueAsString` is used many times. And we risk to introduce some other issues or change behaviour by eating the exception.

At least it´s not an ideal candidate for a quick fix.

The root issue IMO is that we pass LocalDateTime as Fallback for isConversionFailed. And yes, there may be a custom converter which always expects java.util.Date.
@arnis87: So it may be some kind of work-around to extend the `(Custom)Converter#getAsString` to also handle LocalDateTime correct. I´ll hope this would be easy to implement.

@tandraschko pointed me a few days ago to https://github.com/primefaces/primefaces/issues/5913. As a long term solution we may solve this by introducing a attribute like valueType to explizit define the valuetype. (For whenever someone is not happy with our default-behaviour.)

Let us sleep a night or two and then get back to this issue.

melloware:
The problem i have with that is this code is in 30+ of my production apps.

```xml
<p:calendar value="#{regularJavaUtilDateObj}">
    <f:convertDateTime />
</p:calendar>
```

This has been PF standard and JSF standard for 10+ years.  Making the user do this in 1000's of places..
```xml
<p:calendar value="#{regularJavaUtilDateObj}" type="java.util.Date">
    <f:convertDateTime />
</p:calendar>
```
... is not acceptable to me.  I think PF needs to support the above out of the box.  Even if it means switching `LocalDate.now()` back to `new Date();`

arnis87:
Hi,
I agree with @melloware on this. I think this should be fixed in the library instead of having workarounds. I also have this problem now in multiple applications that are not using newer java date/time classes and probably never will. 
Implementing custom converter as a workaround for `LocalDateTime` handling in code that does not deal with `LocalDateTime` anywhere else is odd at best. 

As I mentioned I do have working workaround for this for now, but it is very fragile since it relies heavily on the original source code and required extra effort to keep in sync. For me there is no big rush but I would like to see this fixed someway in next few elite releases.

christophs78:
Totally agree. (I missed the important point of `<f:convertDateTime />`.)

What´s about using BaseCalendarRenderer#resolveDateType?

christophs78:
The weakness of BaseCalendarRenderer#resolveDateType are corner cases with generics...

Another brainstorming-idea may be to look at the converter. Does we have one? Is it f:convertDateTime? Which type-attribute does it have?

melloware:
I am looking at ResolveType for now.

melloware:
OK check out the latest commit.  It puts this code in a much better state.  It may not be perfect yet but its way better than what is happening right now.

src/main/java/org/primefaces/util/CalendarUtils.java
```diff
@@ -532,4 +533,26 @@ public static Temporal now(UICalendar uicalendar) {
...
+        Temporal now = now(uicalendar);
+        if (dateType.isAssignableFrom(java.util.Date.class)) {
+            ZoneId zone = calculateZoneId(uicalendar.getTimeZone());
+            java.util.Date date;
+            if (now instanceof LocalDate) {
+                date = java.util.Date.from(((LocalDate) now).atStartOfDay(zone).toInstant());
+            }
+            else {
+                date = java.util.Date.from(((LocalDateTime) now).atZone(zone).toInstant());
+            }
```
christophs78: LocalTime?

src/main/java/org/primefaces/util/CalendarUtils.java
```diff
@@ -235,7 +236,7 @@ public static final String getValueAsString(FacesContext context, UICalendar cal
     }
 
     public static final String getValue(FacesContext context, UICalendar calendar, Object value, String pattern) {
-        //first ask the converter
+      //first ask the converter, if it fails fall back to built-in conversion
```
christophs78: we should remove the  "if it fails ...."

christophs78:
Maybe we should add some f:convertDateTime - mutations to datePicker.xhtml and datePickerJava8.xhtml. (Some kind of "Things we should not break because it was used in the past. But you should not do them for new projects.")
Calendar and DatePicker have so many mutations / configurations. It´s realy hard to keep this all under control. UnitTest´s help a bit. But there a so many things we can´t cover with them.

melloware:
Added LocalTime and fixed comment.

src/main/java/org/primefaces/util/CalendarUtils.java
```diff
@@ -530,7 +530,11 @@ public static LocalDateTime toLocalDateTime(ZoneId zoneId, String isoDateString)
      */
     public static Temporal now(UICalendar uicalendar) {
         boolean hasTime = uicalendar.hasTime();
+        boolean timeOnly = uicalendar.isTimeOnly();
         ZoneId zone = calculateZoneId(uicalendar.getTimeZone());
+        if (hasTime && timeOnly) {
```
christophs78: One last thing. :-)
hasTime = date and time
timeOnly = time
(I know this is a bit confusing.)
--> `if (timeOnly) {`

melloware:
Good catch.  Fixed.

melloware:
@christophs78 @tandraschko I just checked in a MUCH more comprehensive fix that checks the following:
1. Is the converter a `javax.faces.convert.DateTimeConverter` ?
2. if YES, then check if we are using JSF 2.3 and get the `type` property which will be "date", "localDate", "localDateTime" etc. an check if it matches the `value` class SimpleName. If it does not fall back to built in conversion.
3. If it does match the class name use the converter.
4. If JSF 2.2 or below always assume its "date" since that is all that converter can do.

christophs78:
Excellent work!
This should hopefully work for all situations. Even these ones we don´t think about yet.

melloware:
Yes it actually helped me catch a bug while I was testing with @Future bean validaiton with a regular Date and `f:convertDateTime`.  The above fix fixed all my issues.

PR #5934, opened 2020-05-21
Fix #5817: JQMIGRATE: jQuery.trim is deprecated; use String.prototype.trim

### LINKED ISSUE #5817: JS: Remediate all JQMIGRATE deprecations

melloware:
Too much to fix in 1 PR so will be creating separate PR's for each thing fixed.


tandraschko:
only difference could be that $.trim also worked for null values`?

melloware:
Yep I was afraid of that.  What do you think we should do?  Create a core.utils `trim`???

tandraschko:
yep, probably

melloware:
I looked at what jQuery trim was doing and make a `PrimeFaces.trim` to match it.

PR #5937, opened 2020-05-21
Fix #5817: JQMIGRATE: Use of number-typed values is deprecated in CSS

### LINKED ISSUE #5817: JS: Remediate all JQMIGRATE deprecations

melloware:
JQMIGRATE: Use of number-typed values is deprecated in jQuery.fn.css
Cause: In past versions, when a number-typed value was passed to `.css()` jQuery converted it to a string and added "px" to the end. As the CSS standard has evolved, an increasingly large set of CSS properties now accept values that are unitless numbers, where this behavior is incorrect. It has become impractical to manage these exceptions in the jQuery.cssNumber object. In addition, some CSS properties like line-height can accept both a bare number 2 or a pixel value 2px. jQuery cannot know the correct way to interpret `$.css("line-height", 2)` and currently treats it as "2px".

**Solution:** Always pass string values to `.css()`, and explicitly add units where required. For example, use `$.css("line-height", "2")` to specify 200% of the current line height or `$.css("line-height", "2px")` to specify pixels. When the numeric value is in a variable, ensure the value is converted to string, e.g. `$.css("line-height", String(height))` and `$.css("line-height", height+"px")`.


tandraschko:
ok, no problem for dimensions like height and width. But really also for zindex?

melloware:
Yep zIndex and Opacity are all on the list.

tandraschko:
should we use `new String` instead `String`? looks more java-like IMO

GedMarc:
On layout I'm hitting this problem where the assignment to string is breaking the calculations used, i could get rid of most of them - but one or two, i simply cannot change from number to string, screws up the math with javascript

I'm not sure how/what they want us to do as a replacement

melloware:
@GedMarc can you give an example?  So for example you can just change `1` to `String(1)` as the old way use to convert to pixels so that one you would have to give it your units like `1px`.

melloware:
> should we use `new String` instead `String`? looks more java-like IMO

@tandraschko You peeked my interest in this so I researched it and I think we are doing the correct thing with `String()` vs `new String()` based on this: https://stackoverflow.com/a/50082325/502366

GedMarc:
@melloware 

So i'm setting css() using an object that has an number as a variable --

```
obj : {
var1: 0
}
```
If i pass the object into the css, you get the warning because of var1.
If you change var 1 to string, obviously math won't work in JavaScript

So what is meant to be the replacement? :)

melloware:
I guess I am confused on what math is being done after it’s passed to .css()???  Once you have handed it off isn’t that the end of it.  Can you give me a more elaborate example?

tandraschko:
where is it in https://jquery.com/upgrade-guide/3.0/ ?

melloware:
Unfortunately @tandraschko I have noticed Jquery has stopped updating there documentation or site examples.  That Upgrade guide is for 3.0 but they never put one out for 3.2, 3.3, 3.4.  

They did create one for 3.5 however though: https://jquery.com/upgrade-guide/3.5/

However the JQMigrate 3.3 plugin has it documented here:
https://github.com/jquery/jquery-migrate/blob/master/warnings.md#jqmigrate-use-of-number-typed-values-is-deprecated-in-jqueryfncss

And if you run the migrate plugin you will see his warnings pop up.

tandraschko:
i see
+1 then

PR #5951, opened 2020-05-23
FileUpload - fix issues with allowTypes, fileLimit, sizeLimit and templateMessage for simple mode

### LINKED ISSUE #5941: FileUpload: default values of `invalidSizeMessage`, `invalidFileMessage` and `fileLimitMessage`
**Description**

In the docs the default value of `invalidSizeMessage`, `invalidFileMessage` and `fileLimitMessage` is null but in 'advanced' mode these attributes have the following default values: 'Invalid file size', 'Invalid file type' and 'Maximum number of files exceeded'.

For simple mode the default value is null. As an enhancement, the default value for simple mode could be the same as the ones in advanced mode.

doleron:
Hi, this PR is intended to fix issue https://github.com/primefaces/primefaces/issues/5941
- fix default invalidSizeMessage for simple mode
- fix no support for `fileLimit` and `allowTypes` validation on simple mode
- fix no support for `templateMessage` on simple mode
Furthermore, docs were updated with proper default value for `invalidSizeMessage`, `invalidFileMessage` and `fileLimitMessage`.


doleron:
The changes were evaluated using a set of automatic junit/selenium test. More details can be found here: https://github.com/doleron/primefaces-interface-test

src/main/java/org/primefaces/component/fileupload/FileUploadRenderer.java
```diff
@@ -95,21 +87,27 @@ protected void encodeScript(FacesContext context, FileUpload fileUpload) throws
...
         else {
             wb.init("SimpleFileUpload", fileUpload.resolveWidgetVar(context), clientId)
-                    .attr("skinSimple", fileUpload.isSkinSimple(), false)
-                    .attr("maxFileSize", fileUpload.getSizeLimit(), Long.MAX_VALUE)
-                    .attr("invalidSizeMessage", EscapeUtils.forJavaScript(fileUpload.getInvalidSizeMessage()), null);
+                    .attr("skinSimple", fileUpload.isSkinSimple(), false);
+        }
+
+        wb.attr("disabled", fileUpload.isDisabled(), false)
+                .attr("invalidSizeMessage", fileUpload.getInvalidSizeMessage(), null)
```
melloware, line 100: Why did you remove the `EscapeUtils` around invalidSize messsage?

melloware (changes requested):

tandraschko:
@doleron thanks for your work!
Could you image to setup some more automatic tests?
We have a selenium framework here: https://github.com/primefaces-extensions/primefaces-selenium

We could automatically create selenium wrappers + test PrimeFaces with the same framework.
That would be my dream to have something running via travis CI each night.

It would be a real big big big benefit for PrimeFaces.

src/main/java/org/primefaces/component/fileupload/FileUploadRenderer.java
doleron: Hi @melloware 

The `EscapeUtils.forJavaScript()` call in `FileUploadRenderer` was removed due the following reasons:

1) It is already called by `WidgetBuilder.attr()`:
```java
public WidgetBuilder attr(String name, String value, String defaultValue) throws IOException {
        if (value != null && !value.equals(defaultValue)) {
            // ...
            rw.write(EscapeUtils.forJavaScript(value));
            // ...
        }
        return this;
    }
```
2) Calling `EscapeUtils.forJavaScript()` before the nullity check results in `"null"` strings instead of `null` references;
```java
assertEquals("null", EscapeUtils.forJavaScript(null)); 
```

I just added some tests for these cases.

melloware (approved):

doleron:
@tandraschko definitively I can!

I'm working on this right now.

melloware:
I have tested this out and its looking good.  I would like to merge it unless anyone else @tandraschko 
 @Rapster @christophs78 wants to do a look over before I do?

christophs78:
Looks good.
As far as i see it does not touch advanced mode. (And so should not break anything for advanced mode.)

Issue #4837, opened 2019-05-25
New feature: combine fileupload with image cropper


djmj:
Image cropper and fileupload are nice plugins but a combination of both would be awesome since the cropping could be done before saving the image in the system.

**Process**

When the user selects an image a crop button / checkbox could be shown next to the preview.

On upload a dialog (master / detaillevel) can be shown where the user can select the crop rectangle and we programmers will automatically receive the cropped image in the bean. 

Maybe the rectangle coordinates can be selected on the local blob? In this case these coordinates can be passed to a custom JSF converter that crops the image, so there is no need to save the images in the bean during cropping?

tandraschko:
very unlike that it will be implemented without sponsoring

djmj:
Understandable but it should also wait until a new base like cropper.js is selected #4136.

melloware:
Now that Cropper has been updated and in 9.0 supports dynamic streamed content this showcase PR demonstrates using FileUpload and Cropper together: https://github.com/primefaces/primefaces-showcase/pull/157

SwiftWizard:
> Now that Cropper has been updated and in 9.0 supports dynamic streamed content this showcase PR demonstrates using FileUpload and Cropper together: https://github.com/primefaces/primefaces-showcase/pull/

Broken link...

melloware:
@SwiftWizard what link are you looking for?

The showcase has the example right here: https://www.primefaces.org/showcase/ui/multimedia/cropper/fileupload.xhtml?jfwid=095fe

PR #5957, opened 2020-05-26
Fix #12: allow Dynamic content support in ImageCropper

### LINKED ISSUE #12: ImageCropper: Dynamic content support
ImageCropper currently only works with static images, enable support for dynamic content via StreamedContent.

doleron:
Hi, this PR is intended to fix issue https://github.com/primefaces/primefaces/issues/12 by allowing `ImageCropper` component to load images dynamically as GraphicImage does.
Example:
dyn-image-cropper
```xhtml
<h:panelGrid columns="1">

    <p:spinner id="iterationSpinner"
        value="#{dynamicCropper.numberOfIterations}"
        min="5" max="50" stepFactor="5"
        valueChangeListener="#{dynamicCropper.updateImage}">
        <p:ajax process="@this" update="cropper" />
    </p:spinner>

    <h:panelGrid columns="2">
        <p:outputPanel id="cropper">
            <p:imageCropper
                value="#{dynamicCropper.croppedImage}"
                id="streamed_cropper" cache="false"
                stream="#{dynamicCropper.image}"
                initialCoords="50,50,150,100"
                minSize="50,50" maxSize="350,350" />
        </p:outputPanel>
        <p:outputPanel id="cropped">
            <p:graphicImage
                rendered="#{not empty dynamicCropper.newImageName}"
                name="demo/images/crop/#{dynamicCropper.newImageName}" />
        </p:outputPanel>
    </h:panelGrid>

    <p:commandButton value="Crop"
        action="#{dynamicCropper.crop}"
        update="cropped msgs" icon="pi pi-clone" />

</h:panelGrid>
```
```java
@Named
@SessionScoped
public class DynamicCropper implements Serializable {
    
    private BufferedImage image;
    private final int width = 400;
    private final int height = 300;
    private int numberOfIterations;
    
    @PostConstruct
    public void init() {
        this.numberOfIterations = 5;
        try {
            this.image = mandelbrotSet(width, height, numberOfIterations);
        } catch (IOException e) {
            e.printStackTrace();
        }
    }
    
    public StreamedContent getImage() {
        
        StreamedContent result = null;
        if (context.getCurrentPhaseId() == PhaseId.RENDER_RESPONSE) {
            result = new DefaultStreamedContent();
        }
        else {
            result = DefaultStreamedContent.builder().contentType("image/png").stream(() -> {
                try {
                    ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
                    ImageIO.write(this.image, "png", outputStream);
                    return new ByteArrayInputStream(outputStream.toByteArray());
                } catch (Exception e) {
                    e.printStackTrace();
                    return null;
                }
            }).build();
        }
        
        return result;
    }
    
    public void updateImage(ValueChangeEvent event) {
        Number newValue = (Number) event.getNewValue();
        try {
            this.image = this.mandelbrotSet(width, height, newValue.intValue());
        } catch (IOException e) {
            e.printStackTrace();
        }
    } 
    // ...
}
```

Note that the `StreamedContent` attribute is bound to (new) `stream` attribute since `ImageCropper` has a `value` attribute already for the destination of cropped image


melloware:
Awesome I have been meaning to get to this one for a while.  Can you add a Showcase PR as well demonstrating use of dynamic content?

src/main/java/org/primefaces/component/imagecropper/ImageCropperRenderer.java
```diff
@@ -180,6 +184,15 @@ public Object getConvertedValue(FacesContext context, UIComponent component, Obj
         Resource resource = getImageResource(context, cropper);
         InputStream inputStream = null;
         String imagePath = cropper.getImage();
+        Object streamObject = cropper.getStream();
+        StreamedContent stream = null;
+        if (streamObject != null && !(streamObject instanceof StreamedContent)) {
```
melloware: I don't think you have check != null because intanceof NULL is always false. you can invert this by checking is streamobject instanceof ELSE throw exception

src/main/java/org/primefaces/component/imagecropper/ImageCropperRenderer.java
```diff
@@ -189,29 +202,38 @@ public Object getConvertedValue(FacesContext context, UIComponent component, Obj
...
-                    String webRoot = externalContext.getRealPath(Constants.EMPTY_STRING);
-                    String fileSeparator = Constants.EMPTY_STRING;
-                    if (!(webRoot.endsWith("\\") || webRoot.endsWith("/")) &&
-                                !(imagePath.startsWith("\\") || imagePath.startsWith("/"))) {
-                        fileSeparator = "/";
-                    }
-
-                    File file = new File(webRoot + fileSeparator + imagePath);
-                    inputStream = new FileInputStream(file);
+                    throw new IllegalArgumentException("Either image and stream attributes are not provided.");
```
melloware: "Either image or stream attribute is not provided"

melloware (changes requested):

docs/9_0/components/imagecropper.md
```diff
@@ -45,6 +45,7 @@ sizeLimit | 10485760 | Long | Maximum number of bytes the image.
 responsive | true | Boolean | Re-render the cropper when resizing the window.
 guides | true | Boolean | Show the dashed lines in the crop box.
 cache | true | Boolean | Controls browser caching mode of the resource. Default is true.
+stream | null | Object | Binary data to stream.
 viewMode | 1 | Integer | Define the view mode of the cropper. If you set viewMode to 0, the crop box can extend outside the canvas, while a value of 1, 2 or 3 will restrict the crop box to the size of the canvas. A viewMode of 2 or 3 will additionally restrict the canvas to the container. Note that if the proportions of the canvas and the container are the same, there is no difference between 2 and 3.
 
 ## Getting started with the ImageCropper
```
melloware: Can you also add this section to the docs...
```
## Dynamic Images
A dynamic image can be used with ImageCropper which requires _org.primefaces.model.StreamedContent_ `stream` as it’s value  
StreamedContent is an interface and PrimeFaces provides a built-in implementation called
_DefaultStreamedContent_. Please see our core documentation about it: [Dynamic Content Streaming / Rendering](/core/dynamiccontent.md)
```

doleron:
Hi @melloware, this is the showcase PR: https://github.com/primefaces/primefaces-showcase/pull/156

tandraschko:
what about making "image" attr a Object and do a instanceOf StreamedContent / String?

src/main/java/org/primefaces/component/imagecropper/ImageCropperRenderer.java
doleron: Fixed!

docs/9_0/components/imagecropper.md
doleron: Done!

melloware:
@tandraschko you might be right `image` might be better than `stream` or even `dynamicImage`??

tandraschko:
we already have `image`, we must just state in the docu that in can be a string or StreamedCOntent

melloware:
right so it will be like graphicImage `value` it can either be a URL or StreamedContent.

src/main/java/org/primefaces/component/imagecropper/ImageCropperRenderer.java
doleron: The check `streamObject != null` is performed because '`cropper.getStream()` null' is a valid condition if '`cropper.getImage()` is not null'. I just included some tests for check these scenarios.

melloware: I think @tandraschko wants you to not introduce a new `stream` property and just re-use `image` so it can either be a URL or StreamedContent like GraphicImage does it.

src/main/resources/META-INF/primefaces-p.taglib.xml
```diff
@@ -11202,6 +11202,14 @@
...
             <type>java.lang.Boolean</type>
         </attribute>
+        <attribute>
+            <description>
+                <![CDATA[Binary data to stream the image dynamically]]>
+            </description>
+            <name>stream</name>
+            <required>false</required>
+            <type>java.lang.Object</type>
+        </attribute>
```
melloware: This needs to be removed right? and `image` needs to be updated to java.lang.Object.

melloware (changes requested):

src/main/resources/META-INF/primefaces-p.taglib.xml
doleron: Done! In addition, `image` is required='true', confirm?

melloware:
Yes I can't imagine you can use this component without `image` !

melloware (approved):

tandraschko:
Another great contribution, thanks!
One small thing, we should use PropertyKeys instead of strings:
```
imageCropper.isCache(), DynamicContentType.STREAMED_CONTENT, true, ImageCropper.PropertyKeys.image.name());
```

melloware:
@doleron i will fix that i will check everywhere DynamicSrc is being used and use property key.

doleron:
@tandraschko @melloware thank you, guys. I'm really glad to collaborate! By the way, regarding this feature request https://github.com/primefaces/primefaces/issues/4837 I'm wondering me if a new showcase entry using Streamed ImageCrop and FileUpload could be enough. What your thoughts?

melloware:
Sure if you think it meets that requirement a showcase PR would be great and then we can close that ticket.

PR #5966, opened 2020-05-28
Fix #5817: JQMIGRATE: jQuery.fn.XXX event shorthand is deprecated

### LINKED ISSUE #5817: JS: Remediate all JQMIGRATE deprecations

melloware:
**Cause:** The .on() and .trigger() methods can set an event handler or generate an event for any event type, and should be used instead of the shortcut methods. This message also applies to the other event shorthands, including: blur, focus, focusin, focusout, resize, scroll, dblclick, mousedown, mouseup, mousemove, mouseover, mouseout, mouseenter, mouseleave, change, select, submit, keydown, keypress, keyup, and contextmenu.

**Solution:** Instead of `.click(fn)` use `.on("click", fn)`. Instead of `.click()` use `.trigger("click")`.


tandraschko:
can you fix the conflicts?

melloware:
Fixed!

Issue #2223, opened 2017-03-31
PhotoCam: improve error handling


marquem:
Hi,

It would be great to improve the photocam's error handling, similar to the discreet error messages provided by the fileUpload component. The problem I have with alerts is, they tend to be ugly, intrusive and aren't very user-friendly on mobile devices.

Thanks,
Michael

## Environment
- PrimeFaces version: 5.2, 6.0, 6.0.17
- Application server + version: JBoss EAP 6.4

tandraschko:
A PR or technical ideas are welcome here ;)

Issue #2152, opened 2017-03-07
PhotoCam: add resolution selection


giorgioprimefaces:
This is a feature request,
It would be great to have resolution selection in future releases,
actually I can only get images 320x240px which is sooo small nowadays!

thanks in advance,
Giorgio

tandraschko:
Does webcamJS support this?

giorgioprimefaces:
if webcamJS is this one
[https://pixlcore.com/](url), it can take hd pictures and you can set many parameters:
Webcam.set({
    width: 320,
    height: 240,
    dest_width: 640,
    dest_height: 480,
    image_format: 'jpeg',
    jpeg_quality: 90,
    force_flash: false
});

tandraschko:
https://github.com/jhuckaby/webcamjs

its this
we use this under the hood. As it's in maintenance mode, we should rewrite the component probably. Feel to give ideas or sponsor it.

giorgioprimefaces:
thanks tandraschko, 
I've just discovered that p:upload used on mobile can use smartphone's camera and its options, so, for what I need it's more than enought.
I will not close this issue because I think it will be useful to improve options in p:photocam for desktop users...

Giorgio

doleron:
Hi, 

Am I missing something or are `width` and `height` what are you looking for? From [docs](https://primefaces.github.io/primefaces/8_0/#/components/photocam): 

| Name | Default | Type | Description | 
| --- | --- | --- | --- |
id | null | String | Unique identifier of the component
width | 320 | Integer | Width of the camera viewport.
height | 240 | Integer | Height of the camera viewport.

They are passed to Webcam.js engine under the wood during the camera initialization:

```javascript
attach: function() {
        // ...
        Webcam.set({
            width: this.cfg.width,
            height: this.cfg.height,
```
In this example it is set up to 960x720:

```xhtml
<p:photoCam width="960" height="720"/>
```

![image](https://user-images.githubusercontent.com/9665358/84003802-f44e6180-a940-11ea-9940-4faa335976cf.png)

melloware:
I think the only thing he mentions differently is 
```
dest_width: 640,
dest_height: 480,
```
Not sure from the webcam docs how the dest is different than the width and height.

doleron:
Well, dest_width and dest_height can be set by photoWidth/photoHeight:

```xhtml
<p:photoCam width="960"height="720"photoWidth="96" photoHeight="72" />
````
[...]

PR #5969, opened 2020-05-31
PrimeFaces AI components

doleron:
Hi, this PR is intended to propose 3 new components for PrimeFaces:

- **SmartCam**: component to apply client-side computer vision detection algorithms in the camera stream. By default, SmarCam performs face detection like a regular smartphone camera does. Optionally face landmarks can be retrieved as well and other custom object detection algorithms can be applied.
- **SpeechRecognition** uses built-in browser Web Speech API to provide multi-language speech recognition.
- **Speaker**: Spearker is a voice synthetizer based on Web Speech API

Examples of usage can be found here: https://github.com/doleron/primefaces-showcase/tree/primefaces-ai 

## SmartCam - face detection + landamark detection

image

```xhtml
<ai:smartCam id="mySmartCamera" widgetVar="mySmartCameraWV"
    autoStart="false"
    width="640" height="480" renderTimeout="50"
    imageHandler="myCameraHandler(video, canvasContext, detections)"
/>
```
## SmartCam - object detection

image
``` xhtml
<ai:smartCam id="mySmartCamera" widgetVar="mySmartCameraWV"
    autoStart="false"
    width="640" height="480" renderTimeout="20"
    imageHandler="myCameraHandler(video, canvasContext, detections)"
    model="return loadModel()"
    predict="return customPredict(model, input)"
/>
```
## Speech Recognition

ezgif-6-9af93bc014e4

```xhtml
<ai:speechRecognition id="mySpeechRecognition" widgetVar="mySpeechRecognitionWV"
    speechHandler="onSpeech(speech)" 
   speechErrorHandler="onError(error)"
    language="en-US"/>
```
## Speaker

image

```xhtml
<ai:speaker id="mySpeaker" widgetVar="mySpeakerWV"
    voice="en-US"/>
```

It is noteworthy that they are in 'proposal stage' yet. Thus, in the current code is not ready for production, there is no documentation and some must-to-have functionalities are not provided but tracked out in the backlog.


christophs78:
Very interesting!

Which browsers do support this?
https://caniuse.com/#search=SpeechRecognition does not work yet with Firefox. What´s the state for the other two features?

https://caniuse.com/#search=speechSynthesis

doleron:
Firefox does support speech recognition, at least the new versions:

image

In my  browser I needed to enable `media.webspeech.recognition.enable` and `media.webspeech.recognition.force_enable`. 
I'm not sure if Edge full support for Speech Recognition is ready tough.

melloware:
Can Smartcam just replace the current Photocam components. The js library is quite old and no longer being updated.  Plus you can’t pick which camera front or back without hacking the js.  So if smartcam is robust to replace pbotocam we should just make it the new Photocam component instead of a new component. Thoughts?

doleron:
My first attempt was embed face recog in PhotoCam but I gave up. Upgrading PhotoCam to HTML5 is really a good thing to do.

melloware:
a few general comments I will do a deeper code review later but.

1. Instead of `language` property use `locale` which is standard for most components and look how TimeLine does it...
```java
wb.attr("locale", timeline.calculateLocale(context).toString());
```
Calculate Locale checks faces-config.xml for locale, your current browser settings or lets you set it specifically.  This is how language is handled in most components.  And no need to set default to "en-US" as `LocaleUtils` will pick an appropriate default for you.

2. I don't think anyone will approve of the `.ai` subpackage or new `ai` taglib.

3. Camera selection "rear or front" should be a first class citizen.  See this forum post: https://forum.primefaces.org/viewtopic.php?f=3&t=46682

4. We should make sure we support these 2 open tickets so we can close them:
https://github.com/primefaces/primefaces/issues/2223
https://github.com/primefaces/primefaces/issues/2152

doleron:
Ok,

I will begin by submitting another PR covering:
1. Upgrading of PhotoCam to HTML5 keeping the current set of features
2. Fixes of https://github.com/primefaces/primefaces/issues/2223 and  https://github.com/primefaces/primefaces/issues/2152
3. Support to device list/selection

tandraschko:
- in general +1 if it works fine in all browsers
- i would not create a second "ai" taglib
- formattings are not correct in some JS files
- all the points from melloware

christophs78:
Is all browsers except IE11 Ok? (I would expect IE11 is (almost) history when PrimeFace 9.0 get´s released.)

src/main/java/org/primefaces/component/ai/smartcam/SmartCamRenderer.java
```diff
@@ -0,0 +1,90 @@
...
+        WidgetBuilder wb = getWidgetBuilder(context);
+        wb.init("SmartCam", component.resolveWidgetVar(context), clientId)
+                .attr("width", component.getWidth(), 320)
+                .attr("height", component.getHeight(), 240)
+                .attr("renderTimeout", component.getRenderTimeout(), 50)
+                .attr("autoStart", component.isAutoStart(), true)
+                .callback("model", "function()", component.getModel())
+                .callback("predict", "function(model, input)", component.getPredict())
+                .callback("imageHandler", "function(video, canvasContext, detections)", component.getImageHandler());
+
```
melloware, line 77: I think all `callbacks` should use `on` like onmodel, onpredict, and onimageHandler to follow PF conventions.

Why not make them `'p:ajax` events as well?

src/main/java/org/primefaces/component/ai/speechrecognition/SpeechRecognitionRenderer.java
```diff
@@ -0,0 +1,87 @@
...
+    protected void encodeScript(FacesContext context, SpeechRecognition component) throws IOException {
+        String clientId = component.getClientId(context);
+
+        WidgetBuilder wb = getWidgetBuilder(context);
+        wb.init("SpeechRecognition", component.resolveWidgetVar(context), clientId)
+                .attr("width", component.getWidth(), 320)
+                .attr("height", component.getHeight(), 240)
+                .attr("language", component.getLanguage(), "en-US")
+                .callback("speechHandler", "function(speech)", component.getSpeechHandler())
+                .callback("speechErrorHandler", "function(error)", component.getSpeechErrorHandler());
```
melloware, line 73: These callbacks should be `onspeech` and `onerror`

melloware (changes requested):

src/main/resources/META-INF/resources/primefaces/ai/smartcam/smartcam.js
```diff
@@ -0,0 +1,122 @@
...
+    	this._super(cfg);
+        if(this.cfg.disabled) {
+            return;
+        }
+        
+        this.cfg.width = this.cfg.width || 320;
+        this.cfg.height = this.cfg.height || 240;
+        this.cfg.renderTimeout = this.cfg.renderTimeout || 50;
+        
+        if (!("autoStart" in this.cfg)) {
```
melloware, line 13: You don't need to initialize this autostart variable at all just say later in the code `if (this.cfg.autoStart)` and its false by default and will only ever be true if you passed in true.  Javascript will assume false even if the variable is undefined.

src/main/resources/META-INF/resources/primefaces/ai/smartcam/smartcam.js
```diff
@@ -0,0 +1,122 @@
...
+        this.loop();
+        
+        if (this.cfg.autoStart) {
+            this.play();
+        }
+
+    },
+    
+	defaultLoad: async function() {
+		return await facemesh.load();
```
melloware, line 41: `await` makes this not work in IE.  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await

src/main/resources/META-INF/resources/primefaces/ai/smartcam/smartcam.js
```diff
@@ -0,0 +1,122 @@
...
+	},
+	
+	defaultPredict: async function(model, input) {
+		return await this.model.estimateFaces(input);
+	},
+    
+    play: function() {
+        var div = document.getElementById(this.cfg.id);
+
```
melloware, line 50: Would rather see this short circuited with `if (!div) return;` rather than all that indented code inside the `if (div)`

src/main/resources/META-INF/resources/primefaces/ai/smartcam/smartcam.js
```diff
@@ -0,0 +1,122 @@
...
+        	this.video = document.createElement("video");
+        	this.video.width = this.cfg.width;
+        	this.video.height = this.cfg.height;
+        	this.canvas = document.createElement("canvas");
+        	this.canvas.width = this.cfg.width;
+        	this.canvas.height = this.cfg.height;
+        	div.appendChild(this.canvas);
+        	this.context = this.canvas.getContext('2d');
+        	
+        	$this = this;
```
melloware, line 61: move the `$this = this` up to the `var` section at the beginning of the method where you declare `var div`

src/main/resources/META-INF/resources/primefaces/ai/smartcam/smartcam.js
```diff
@@ -0,0 +1,122 @@
...
+        
+	        if(navigator.mediaDevices && navigator.mediaDevices.getUserMedia) {
```
melloware, line 63: If these are not true should these be also checked up above where the `if div` check is so we can short circuit this whole method.

src/main/resources/META-INF/resources/primefaces/ai/speaker/speaker.js
```diff
@@ -0,0 +1,77 @@
...
+        }
+        
+        var speech = new SpeechSynthesisUtterance(txt);
+        
+        speech.onend = function (event) {
+            console.log('SpeechSynthesisUtterance.onend');
+        }
+        
+        speech.onerror = function (event) {
+            console.error('SpeechSynthesisUtterance.onerror');
```
melloware, line 37: Shouldn't this be a `callback` hook so users can get the onerror?  Maybe even and AJAX event `callBehavior`

src/main/resources/META-INF/resources/primefaces/ai/speaker/speaker.js
```diff
@@ -0,0 +1,77 @@
...
+        	return;
+        }
+        if (synth.speaking) {
+        	return;
+        }
+        
+        var speech = new SpeechSynthesisUtterance(txt);
+        
+        speech.onend = function (event) {
+            console.log('SpeechSynthesisUtterance.onend');
```
melloware, line 33: Shouldn't this be a `callback` hook so users can get the onerror?  Maybe even and AJAX event `callBehavior`

src/main/resources/META-INF/resources/primefaces/ai/speechrecognition/speechrecognition.js
```diff
@@ -0,0 +1,137 @@
+PrimeFaces.widget.SpeechRecognition = PrimeFaces.widget.BaseWidget.extend({
+	
+	langs:
```
melloware, line 3: See my previous comment about changing to `locale` and none of this will be necessary.

melloware (changes requested):

PR #5973, opened 2020-06-01
Fix #5956 - SelectCheckboxMenu does not show values for which no option exists

### LINKED ISSUE #5956: SelectCheckboxMenu: Behavior when values contain a value for which no SelectItem exists
**Description**

Consider a `<p:selectCheckboxMenu multiple="true"/>` which has a list of values  and some `f:selectItems` set. The behavior when the value list contains a value for which no selectItem exists (which can happen easily when the select items are from a dynamic source) is a bit strange / inconsistent:

* When the page is rendered, values for which no selectItem exists are still rendered.
* But the user cannot delete the offending values, neither by clicking on the "x" icon, nor by deselecting it in the dropdown list.
* But once the form is submitted, the value is removed.

image

First we need to decide how the expected behavior should be. Would probably also be a good idea to check the behavior of other SelectXXX components.

1. I took a look at what happens with the `selectManyCheckbox` from `http://xmlns.jcp.org/jsf/html`. It simply omits any values for which no SelectItem exists. That would be consistent with the current behavior, as the value is removed upon submission anyway. Would also be a simple and small change.
1. Leave everything as it is, but fix the "x" icon so that it actually deletes the value. Still a bit inconsistent, though, as the value is deleted automatically anyway upon submission, irrespective of whether the user deleted it. 
1. Alternatively, we could also add an option for whether values with no SelectItem should be retained. That sounds more complex and also raises other questions such as what the `itemLabel` of such a value should be.

Personally, I'd prefer option 1 - other opinions?

blutorange:
See #5956 

* When multiple is set to true, the component would render all values currently set,
  irrespective of whether a select item existed for that value. This has been changed
  so that values without a matching select item are not rendered.
* This is consistent with both h:selectManyCheckbox and with multiple set to false.
* Once the user submits the form, values without a matching option were removed anyway.

To test this, you can, for example, modify the CheckboxView.java of the showcase and add some defaults selectedCities etc:

image

Before this commit, the options "Tokio" and "313" would have been rendered as chips. Now, they are omitted:

image

### DIFF
--- a/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
+++ b/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
@@ -233,7 +233,6 @@ public class SelectCheckboxMenuRenderer extends SelectManyRenderer {
         Object valuesArray = (submittedValues != null) ? submittedValues : values;
         String listClass = createStyleClass(menu, null, SelectCheckboxMenu.MULTIPLE_CONTAINER_CLASS);
 
-
         writer.startElement("ul", null);
         writer.writeAttribute("label", menu.getLabel(), null);
         writer.writeAttribute("class", listClass, null);
--- a/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
+++ b/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
@@ -233,6 +233,7 @@ public class SelectCheckboxMenuRenderer extends SelectManyRenderer {
         Object valuesArray = (submittedValues != null) ? submittedValues : values;
         String listClass = createStyleClass(menu, null, SelectCheckboxMenu.MULTIPLE_CONTAINER_CLASS);
 
+
         writer.startElement("ul", null);
         writer.writeAttribute("label", menu.getLabel(), null);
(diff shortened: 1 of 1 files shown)


src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
```diff
@@ -233,6 +233,7 @@ protected void encodeMultipleLabel(FacesContext context, SelectCheckboxMenu menu
         Object valuesArray = (submittedValues != null) ? submittedValues : values;
         String listClass = createStyleClass(menu, null, SelectCheckboxMenu.MULTIPLE_CONTAINER_CLASS);
 
+
```
melloware: Looks like you can remove this extra line break that was added?

melloware (changes requested):

tandraschko:
changes in this region needs good testing because it could break some applications
will check it in 1-2 weeks for my applications

src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
blutorange: Sure, removed it.

melloware (approved):
Looks good to me but like @tandraschko said I will let him review and merge it. Thanks for the PR!

PR #5974, opened 2020-06-01
Update JSDoc for the recent changes

blutorange:
I took a look at the changes since my last JSDoc commit and looked for . Looking pretty good, most changes were bug fixes that do not affect the API, for other changes the doc comments were already modified accordingly :) Only a few things I had to do:

* Fixed typings for FileUpload#createXFileId (@christophs78  I assume it was intended for that function to return a string [and not an object]?) and corrected some typos / formatting.
* Fixed a minor bug in the declaration file generator: it now actually reports an error when a tag is missing a description.
* Added a few missing descriptions for @return tags.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/charts/1-charts.js
+++ b/src/main/resources/META-INF/resources/primefaces/charts/1-charts.js
@@ -412,7 +412,7 @@ PrimeFaces.widget.Chart = PrimeFaces.widget.DeferredWidget.extend({
 
     /**
      * Export the current visual state of this chart as an image.
-     * @return {HTMLImageElement}
+     * @return {HTMLImageElement} An HTML image element with an image of the current chart.
      */
     exportAsImage: function() {
         return this.jq.jqplotToImageElem();
--- a/src/main/resources/META-INF/resources/primefaces/core/core.js
+++ b/src/main/resources/META-INF/resources/primefaces/core/core.js
@@ -92,7 +92,7 @@
          * with the given value and the key used as the name. 
          * @param {string} parent The ID of a FORM element.
          * @param {Record<string, string>} params An object with key-value pairs.
-         * @return {typeof PrimeFaces}
+         * @return {typeof PrimeFaces} This object for chaining.
          */
         addSubmitParam : function(parent, params) {
             var form = $(this.escapeClientId(parent));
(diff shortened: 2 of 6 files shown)


melloware:
I meant to ask you.  How on `init` methods does `PartialWidget<Cfg>` work usually IntelliJ complains and I have to use `@param {PrimeFaces.widget.BaseWidget.cfg} cfg` in PrimeFaces Extensions?

blutorange:
@melloware If I understood correctly, you mean it works in PrimeFaces, but not in PrimeFaces Extensions, right? Well, `PartialWidget<Cfg>` basically works as follows:

* `TCfg` is a type parameter (generics) that parametrizes the available properties on the configuration object. This allows for easy subclassing of a component. In Java, this would look like
```java
class Accordion<TCfg extends AccordionCfg> extends BaseWidget<TCfg> {}
```
* `PartialWidget<TCfg>` is a [type alias](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-aliases) (type constructor) that makes all widget config properties optional (it does a little bit more, see [here](https://github.com/primefaces/primefaces/blob/master/src/main/type-definitions/core.d.ts#L257)).

So for `PartialWidget<TCfg>` to work, IntelliJ needs to know about the `PartialWidget` type alias, and it needs to know about the `TCfg` type parameter.

It's a bit hard to tell without the actual error reported by IntelliJ, but I'm assuming it works for PrimeFaces because either

* IntelliJ reads the [core.d.ts](https://github.com/primefaces/primefaces/blob/master/src/main/type-definitions/core.d.ts) file that contains the `PartialWidget` type, and / or
* it reads the [generated type declarations file](https://raw.githubusercontent.com/primefaces/primefaces/master/docs/PrimeFaces.d.ts) which contains the type parameter `TCfg`.

---

That being said, I had been thinking about applying JSDoc to PrimeFaces extensions as well. Though I wanted to wait a bit to see how it works out for this project. I guess we don't have to document everything at once, but we could start by setting up the extensions project so that JavaScript docs are generated as well.

melloware:
OK that makes sense. I think its just that IntelliJ isn't aware of those types I bet its not reading core.d.ts.

melloware:
Duh, its because I am still pointing to PF 8.0 JAR not the 9.0-SNAPSHOT.  Now I feel like dope!

melloware (approved):

PR #5994, opened 2020-06-05
Upgrade Bean Validations 2.0.2

### LINKED ISSUE #5993: General: Bean Validations 2.0
**Description**
I would like to be able to use bean validations 2.x while maintaining the current package hierarchy.

Please use Jakarta validations API < version 3 (jakarta.x namespace)

**Additional context**
There are methods that validations 1.x does not have. The overrides methods for these have been removed

GedMarc:
https://github.com/primefaces/primefaces/issues/5993

Please allow validations 2.x to be used and referenced with PrimeFaces


melloware:
@tandraschko I think we have held off on BV 2.0 because its not supported in EE6.

GedMarc:
Version 9 min jdk 8?????? :D  🗡️ 

I left it with the full package reference name rather than an import for that reason :)

tandraschko:
i dont think we can use ValidateUnwrappedValue because of JavaEE 7 support

melloware:
Confirmed EE7 only uses Validation 1.0.1.  We can't merge this and break support for EE7.

GedMarc:
Hmm - 

So the idea here is to compile with the updated library, but still ensure compatibility all the way back (I went to EE6)

As a provided dependency, it shouldn't cause any conflict on how it is consumed because of how it is defined.
The method will not cause any errors unless it is used, and it will only be used by a version that implements it.

Please try to compile and test this, no errors will come on any version EE6 and up. 
On EE7 I did test on Wildfly 10, and Payara 4

To replicate an error on why I feel this is necessary, simply use the final javax.el version with the current PF, any validation fails with unreadable class missing method and app crash

If there is another solution to this problem - learn me oh great masters!

GedMarc:
i should actually log a bug for that thinking about it, all the PR's I made were in a way linked :")
For this PR, it just cannot under any circumstances be made an import - otherwise i do understand this PR to be perfectly safe

tandraschko:
the code now uses ValidateUnwrappedValue, so if ConstraintDescriptorWrapper is loaded in JavaEE6/7, it should throw a ClassDefNotFoundException

GedMarc:
It really doesn't / won't :) give it a try :)

tandraschko:
It just means that that you dont use primefaces.INTERPOLATE_CLIENT_SIDE_VALIDATION_MESSAGES? ;)

melloware:
@GedMarc can you test your WildFly with primefaces.INTERPOLATE_CLIENT_SIDE_VALIDATION_MESSAGES and using a client side validation like: https://www.primefaces.org/showcase/ui/csv/bean.xhtml

melloware:
OK @tandraschko I can't explain this but it does work fine.  I set the showcase `primefaces.INTERPOLATE_CLIENT_SIDE_VALIDATION_MESSAGES` and backed it up to Hibernate Validator 5 which is BV 1.1.   I confirmed the class `javax.validation.metadata.ValidateUnwrappedValue` is not on the classpath or in the validation-api.jar but this code still works and classloads fine.   Maybe because its not in the `import` statements?

tandraschko:
can you verify that the class is instantiated?

melloware:
Yes i put a debugger on the constructor and it gets instantiated.    Definitely not what i was expecting.

GedMarc:
Yeah told ya :) 

It won't fail until it hits the method reference - which it will never do in an environment that doesn't declare it :) 
The class loader will only attempt to instantiate the referenced return type on calling the method,

I've done tons quite of few of these this way, quite a neat trick on the class loader to enable backwards compatibility :) Just keep it off the imports list

tandraschko:
mkay
why not uncomment //@Override ?

GedMarc:
I was keeping with the rest of the class structure - the other methods that were overriding future methods had it commented out with the version as well :)

melloware:
@GedMarc it works with Override uncommented I tested that too so safe to add the @Overrides in please.

PR #5996, opened 2020-06-05
Fix #5995: Jdk9 module support

### LINKED ISSUE #5995: General: JDK9 module referencing
**Description**
Please include a module-info for primefaces in order to build Java Native Images and to use it JMod, JImage and JPackage

I would like to use moditect as a proven method of forcing module-info definitions

**Additional context**
The file will be placed inside the versions folder for multi-release jars.
Multi-Release jars need to flag set inside of the manifest in order to be jmods or jimages.

GedMarc:
Allows for modular referencing from prime faces


pom.xml
```diff
@@ -1094,6 +1095,30 @@
...
+                        <phase>package</phase>
+                        <goals>
+                            <goal>add-module-info</goal>
+                        </goals>
+                        <configuration>
+                            <!-- Place in META-INF/versions/9/-->
+                            <jvmVersion>11</jvmVersion>
+                            <!--<overwriteExistingFiles>true</overwriteExistingFiles>-->
+                            <module>
+                                <moduleInfoFile>${basedir}/src/moditect/module-info.java</moduleInfoFile>
```
melloware: /src/conf/ is probably a better folder

src/moditect/module-info.java
```diff
@@ -0,0 +1,305 @@
+open module primefaces {
+
```
melloware: /src/conf/ is a better folder

melloware (changes requested):

melloware:
Did you generate this file by hand?  How did you know how to format it?

GedMarc:
Interestingly... already have this in production at a client (didn't want to make it until it was ready)

Yes this is produced by hand, going through many... many **many** iterations.
Operational co-ordinates here : https://mvnrepository.com/artifact/com.guicedee.services/primefaces

Source code for these co-ordinates here (Special consideration to the shading taking place, i'm pulling in itext and the others)
https://github.com/GedMarc/JWebMPDevKit/tree/master/GuicedEE/Services/gee4j/primefaces

GedMarc:
Oh, um, i did the jackson json, org.json, hibernate, and apache commons libraries ones as well, if you need any clarity ask :)

tandraschko:
does this fully work on older application servers with java8?

melloware:
I built this PR with JDK8 and ran the showcase with JDK8 and everything is working fine.  This just seems to put a module-info.class in the root of the JAR and doesn't make any changes other than that.

@GedMarc just double checking that is the correct location of module-info.class for it to be detected in the JAR?  Also it was put in the primefaces.jar but NOT in the primfaces-jakarta JAR so you might need to modify the SHADE plugin slightly to grab this?

melloware (changes requested):
See comments

pom.xml
```diff
@@ -978,7 +978,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-shade-plugin</artifactId>
-                <version>3.2.2</version>
+                <version>3.2.4</version>
```
GedMarc, line 981: 3.2.4 introduces new methods that conflict with the shade/moditect combination

GedMarc:
```
Receiver class io.yupiik.maven.shade.transformer.RelocationTransformer does not define or inherit an implementation of the resolved method 'abstract void processResource(jav
a.lang.String, java.io.InputStream, java.util.List, long)' of interface org.apache.maven.plugins.shade.resource.ResourceTransformer.
```

GedMarc:
this PR requires a few other decisions to be made, as well as unstable in the moditect plugin when referencing multiple shaded jars

future request.

PR #5998, opened 2020-06-05
Fix #5997: Last co-ordinates for the javax. namespace for version 9

### LINKED ISSUE #5997: General : Please use the last javax. maven coordinates for Version 9
**Description**
Please use the final and last jars (Not version 3) for the namespace

Please implement and use the last jars to be released for the javax. namespace

**Additional context**
This must not/should not conflict with any existing source code base in any way

GedMarc:


tandraschko:
why do we need jaxb?

tandraschko:
also we should never add a impl dependency to PF

pom.xml
```diff
@@ -142,10 +151,24 @@
...
+            <artifactId>jakarta.xml.bind-api</artifactId>
+            <version>2.3.3</version>
+            <optional>true</optional>
+        </dependency>
+
+        <dependency>
+            <groupId>org.glassfish.jaxb</groupId>
+            <artifactId>jaxb-runtime</artifactId>
+            <version>2.3.3</version>
+            <optional>true</optional>
```
melloware: Why do we need this runtime?  PF should not need this

pom.xml
```diff
@@ -142,10 +151,24 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>com.sun.xml.bind</groupId>
+            <artifactId>jaxb-impl</artifactId>
+            <version>2.3.3</version>
+            <optional>true</optional>
```
melloware: Why do we need this impl PF should not need this...only the API.

pom.xml
```diff
@@ -97,13 +98,21 @@
...
-            <scope>provided</scope>
+            <groupId>jakarta.el</groupId>
+            <artifactId>jakarta.el-api</artifactId>
+            <version>3.0.3</version>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>org.glassfish</groupId>
+            <artifactId>jakarta.el</artifactId>
+            <version>3.0.3</version>
```
melloware: Impl not needed...

melloware (changes requested):

GedMarc:
Hmm apparently for the tests
Let's see how scoping them into tests will work

melloware:
Well it looks like ValueExpressionAnaylyzer test is failing with an initialization failure.

melloware:
@GedMarc @tandraschko I think the reason he added the JAXB-impl was for the unit tests because in Java 11+ they no longer include JAXB.  Same problem we had with the Showcase we had to update if you want to run JDK11.  It works fine without the impl in Java 8.  But we should probably add it at "TEST" scope so the PF build works in JDK8 or 11.  

Also I wonder if we should have our Travis build be Oracle JDK8 and OpenJDK 11 as the two JDK's we build for?

tandraschko:
its ok for me as test dependency, otherwise -1

+1 to add a java11 travis build

melloware:
OK I will work on adding JDK11 to Travis.

melloware:
PR https://github.com/primefaces/primefaces/pull/6013 supersedes this PR so closing.

PR #6003, opened 2020-06-06
Fix #6002: Remove Aristo theme

### LINKED ISSUE #6002: Task: Remove Aristo Theme
**Description**
Aristo is a legacy ThemeRoller theme and doesn't even support new components like ToggleSwitch.

I think we should remove it for PF 9.0 as PF includes high quality themes already and the legacy ThemeRoller themes are considered deprecated.

@mertsincan WDYT?

melloware:


tandraschko:
JFYI: still no feedback from

But please also update the docu, it says "omega" as default currently which is wrong anyway

melloware:
Good catch updates docs in 2 places

tandraschko:
Please also add it to the migration guide

melloware:
Migration guide updated!

PR #6007, opened 2020-06-07
Implement #5987: TabView: add multiViewState to keep activeIndex across views

### LINKED ISSUE #5987: TabView: add multiViewState to keep activeIndex across views
**Description**
View A has a TabView with Tabs A, B, C and D. (Tab A is selected per default.)  User selects tab C. User navigates to detail-view B. Then he navigates back to view A. Tab A is selected.

Add multiViewState-attribute like DataList, DataTable, ....

christophs78:


tandraschko:
looks good
now we see the real benefit of @Rapster work to make MultiViewState more abstract :)

src/main/resources/META-INF/resources/primefaces/tabview/tabview.js
```diff
@@ -58,6 +58,7 @@
  * @prop {boolean} cfg.scrollable When enabled, tab headers can be scrolled horizontally instead of wrapping.
  * @prop {number} cfg.selected The currently selected tab.
  * @prop {number} cfg.tabindex Position of the element in the tabbing order.
+ * @prop {boolean} cfg.multiViewState Whether to keep TabView state across views.
```
christophs78, line 61: @blutorange : Is it right this way? IntelliJ had not been happy because it thought cfg does not contain multiViewState.

src/main/resources/META-INF/resources/primefaces/tabview/tabview.js
```diff
@@ -459,8 +461,36 @@ PrimeFaces.widget.TabView = PrimeFaces.widget.DeferredWidget.extend({
...
+                        partialSubmitFilter: this.id + '_activeIndex',
+                        process: this.id,
+                        ignoreAutoUpdate: true,
+                        params: [
+                            {name: this.id + '_activeIndex', value: this.getActiveIndex()}
+                        ]
+                    };
+
+                    PrimeFaces.ajax.Request.handle(options);
+                }
```
christophs78: For dynamic="true" we have activeIndex on the server. Also when tabview includes tabChange-Ajax-Event.
This tries to send activeIndex-updates to the server for the other usecases. Is there a other (better) way to do this. (At least this way seems to work.)
And i have to remove the comment afterwards.

src/main/resources/META-INF/resources/primefaces/tabview/tabview.js
blutorange: @christophs78 Looks good to me. The IDE support for the JSDocs is not perfect, we would probably have to switch to ES6 classes and/or TypeScript for that. That said, IntelliJ probably reads the current PrimeFaces.d.ts file, where this property does not exist yet. That file is generated automatically via `mvn clean install -P release` I guess we should run that occasionally and commit the generated files.

Also, once you open a PR, the Travis build check whether the JSDocs are valid. You can also check that locally by running `mvn clean install -P jsdoc` (see also the readme https://github.com/primefaces/primefaces/blob/master/src/main/type-definitions/README.md#build)

src/main/resources/META-INF/resources/primefaces/tabview/tabview.js
melloware: To me this seems perfectly reasonable to do.

melloware (approved):

Issue #4094, opened 2018-09-28
PhotoCam: Not working in IE 11


AlexKolesnik:
## 1) Environment
- PrimeFaces version: 6.2.9, Official Primefaces Showcase 
- Affected browsers: Internet Explorer 11
- WebCam: Logitech HD Webcam C525

## 2) Expected behavior
When using Internet Explorer 11, the page https://www.primefaces.org/showcase/ui/multimedia/photoCam.xhtml should show a video of a connected webcam. In the absense of supporting browser api, the PhotoCam component should fall back to using flash.
I found a working integration of the camera using flash on https://www.onlinemictest.com/webcam-test/

## 3) Actual behavior
Although a flash file is loaded, no camera picture is shown. 

## 4) Steps to reproduce
Open https://www.primefaces.org/showcase/ui/multimedia/photoCam.xhtml in IE 11.  This problem probably has somethong to do also with the specific model of the webcam. So maybe you are not affected.

melloware:
Most likely related to this issue with the underlying PhotoCam code: 
https://github.com/jhuckaby/webcamjs/issues/151
https://github.com/jhuckaby/webcamjs/issues/67

This patch may work...
https://github.com/jhuckaby/webcamjs/pull/70

AlexKolesnik:
> 
> 
> Most likely related to this issue with the underlying PhotoCam code:
> [jhuckaby/webcamjs#151](https://github.com/jhuckaby/webcamjs/issues/151)
> [jhuckaby/webcamjs#67](https://github.com/jhuckaby/webcamjs/issues/67)
> 
> This patch may work...
> [jhuckaby/webcamjs#70](https://github.com/jhuckaby/webcamjs/pull/70)

Inside the primefaces archive I can only find META-INF\resources\primefaces\photocam\webcam.swf.

When I replace this file with your version https://github.com/jhuckaby/webcamjs/raw/b93a0d4fc1be3d07482dedab86ca0e924ac925ef/webcam.swf , then I can indeed see a camera picture. The picture is a bit scaled up though and when I take a snapshot the size is different again.

To conclude.. your patch almost works.

But thank you very much, we are on the right track.

melloware:
OK so yeah it looks like that solution was to fix the IE11 stuff.  it might not be perfect but at least it now works in IE11.

PR #6009, opened 2020-06-08
Fix #2223 and #6008: PhotoCam error handling improvements and device selection

### LINKED ISSUE #2223: PhotoCam: improve error handling

### LINKED ISSUE #6008: PhotoCam: allow choosing the video input device
**Description**
PhotoCam uses the default video device to grab images. There is no way in PhotoCam API to easily change this behavior.

Allow to declare the target device in a tag level as follows:

```xhtml
<p:photoCam device="#{photoCamView.device}"
 ...
/>
```
The device can be define in the client-side as well:
```javascript
PF('photoCamWidget').device = 'deviceId';
```

In addition, the widget should provide device list and support mobile devices semantics like `"user"` and `"environment"` to refer to front and rear cameras respectively.

See also: https://forum.primefaces.org/viewtopic.php?f=3&t=46682

doleron:
Hi,

this PR is intended to fix https://github.com/primefaces/primefaces/issues/2223 and https://github.com/primefaces/primefaces/issues/6008 . 

In a nutshell:
1. `p:photocam` has a new attribute `device` to set up the desired device id. In mobiles, `"user"` stands for the front camera and `"environment"` stands for the rear camera. This terminology comes from browser media API which can be consulted here: https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia

Some prints below (Safari/iOS 13.2) from my tests:

WhatsApp Image 2020-06-08 at 01 58 17 (1)
WhatsApp Image 2020-06-08 at 01 58 17

In my experiments, this feature also worked on Chrome/FF (android 9)

2. The PhotoCam widget has a new `devices` attribute. When the page loads, as soon as possible, browser populates `devices` attribute with native `InputDeviceInfo` objects. The specific device id for a suposing USB camera attached to the computer can be obtained by this collection. The example below shows PhotoCam using a logitech USB camera.

Capturar

3. It is introduced an `onCameraError` fallback which can be used to override the default alert-based error handler mechanism, as asked in https://github.com/primefaces/primefaces/issues/2223 . 

I would like to add some notes for future research:

1. By my tests, PF 8 PhotoCam doesn't work as expected in Chrome/Firefox on iOS devices although this same component runs fine in Safari on IOs. I found that it is not a specific issue of PF PhotoCam but something up to Apple as discussed here: https://stackoverflow.com/questions/29160819/how-to-use-getusermedia-in-chrome-for-ios/45454817 . 

2. Chrome denies usage of camera devices without HTTPS if the request is not to/from localhost. This can be a little problematic to test the component on mobile browsers or even a remote PC. Even tough a tricy chrome configuration can loosen this restriction, for development purposes a self-signed https certificate should do the job.

3. I'm inclined to think that upgrading PhotoCam to HTML5 only after Apple finishes the getUserMedia support would be the best thing to do.

4. Neither PF 8 PhotoCam nor the code in this PR run on Internet Explorer 11.


docs/9_0/components/photocam.md
```diff
@@ -44,6 +44,8 @@ photoHeight | 240 | Integer | Height of the captured photo, defaults to height.
 format | jpeg | Boolean | Format of the image, valid values are "jpeg" default and png.
 jpegQuality | 90 | Integer | Quality of the image between 0 and 100 when the format is jpeg, default value is 90.
 forceFlash | false | Boolean | Enables always using flash fallback even in an HTML5 environment.
+device | false | String | Suggests a video input device.
```
melloware: Suggests a video input devices such as 'user' (aka font camera) or 'environment' (aka rear camera).

src/main/resources/META-INF/primefaces-p.taglib.xml
```diff
@@ -18720,6 +18720,22 @@
             <required>false</required>
             <type>java.lang.Boolean</type>
         </attribute>
+        <attribute>
+            <description>
+                <![CDATA[An device to grab images]]>
```
melloware, line 18725: Suggests a video input devices such as 'user' (aka font camera) or 'environment' (aka rear camera).

src/main/resources/META-INF/primefaces-p.taglib.xml
```diff
@@ -18720,6 +18720,22 @@
...
+            </description>
+            <name>device</name>
+            <required>false</required>
+            <type>java.lang.String</type>
+        </attribute>
+        <attribute>
+            <description>
+                <![CDATA[Webcam engine error fallback]]>
```
melloware, line 18733: Client side callback executed if the camera has an error.

melloware (changes requested):

melloware:
Also looks like there is some JSDOC issues still:
```
[INFO] Error: at [rule:capitalized-comments] (/home/travis/build/primefaces/primefaces/src/main/type-definitions/src/PrimeFaces.d.ts:29751:9)
[INFO] Comments should not begin with a lowercase character.
[INFO] 
[INFO] Error: at [rule:capitalized-comments] (/home/travis/build/primefaces/primefaces/src/main/type-definitions/src/PrimeFaces.d.ts:29757:9)
[INFO] Comments should not begin with a lowercase character.
```

melloware:
@blutorange can you review this? I can't tell why his JSDOC is failing?

docs/9_0/components/photocam.md
```diff
@@ -73,4 +75,65 @@ public class PhotoCamBean {
...
+
+## Browser Compatibility 
+
+This component is strongly dependent of browser / operational system faculties and constraints. Some of known restrictions are listed below:
+
+- Navigators like Google Chrome (version 47 and later) require secure origins (HTTPS or LOCALHOST) to activate media devices functionality. Thus, unless you are accessing the page from LOCALHOST, PhotoCam is likely to do not work in a plain HTTP call. More details can be found here: https://developers.google.com/web/updates/2015/10/chrome-47-webrtc
+- Third-party browsers on iOS devices (Iphone, Ipad, etc) do not have same access to media devices as the Apple native browser Safari. Thus, on an iOS device, PhotoCam should works fine on Safari but not on Chrome or Firefox (or other no-Safari navigators). See http://www.openradar.me/33571214 for any updates on this. Note that this restriction is applied only for iOS devices. In general, third-party browsers running on OSx systems (mackbooks, iMacs and so on) don't have this same restriction.
+- PhotoCam does not support Microsoft Internet Explorer 11 or below. Other eldery browsers are likely to do not work as well.
```
melloware: PhotoCam does not support Microsoft Internet Explorer 11 or below. Other legacy browsers are also not likely to be functional.

melloware (changes requested):

doleron:
@melloware the issue with JSDOC was fixed. (indeed a mis-capitalization of mine on `reload` JS doc description).

docs/9_0/components/photocam.md
doleron: fixed!

docs/9_0/components/photocam.md
doleron: Done!

melloware: Just this one last wording change above...

melloware (changes requested):

melloware (approved):

docs/9_0/components/photocam.md
doleron: Ops! Fixed in last commit.

melloware:
Hmm why does photocam no longer work in IE11?  I had hacked it so it would work before see this ticket: https://github.com/primefaces/primefaces/issues/4094

is there some new specific feature that doesn't work?

doleron:
Well, in every one of three different windows computers I checked here, https://www.primefaces.org/showcase/ui/multimedia/photoCam.xhtml doesn't work on Internet Explorer 11. So, since we know that PhotoCam worked before on IE11 but no longer works, my first hypothesis is that it is due to something changed in IE11 itself or, probably, adobe Flash dependency. Inspecting code execution on IE suggests that everything runs fine until WebcamJS calls webcam.swf. After this point, the camera is not running in despite of no error or warning is thrown. Hence, I'm blind to know what happens between browser, adobe flash, operational system or other players (like anti-virus, windows defender ...).

melloware:
Ahhh for IE11 its most likely Flash related.  So no worries

doleron:
Ok, @melloware I'm thinking in a change to replace the `devices` attribute by a `getAvailableDevices()` method in order to provide a more idiomatic implementation.

melloware:
Good idea.

src/main/resources/META-INF/resources/primefaces/photocam/photocam.js
```diff
@@ -1137,11 +1158,31 @@ PrimeFaces.widget.PhotoCam = PrimeFaces.widget.BaseWidget.extend({
...
+            Webcam.on("error", this.onCameraError);
 
             Webcam.attach(this.id);
             this.attached = true;
         }
     },
+    
+    /**
+     * Default error handler for webcam events
+     * @param {() => void} errorObj Error object containing message, stacktrace and so on.
```
blutorange: Just a quick note regarding the types: As far as I can tell from the name and how it's used, the type of `errorObj` should probably be `Error` instead of `() => void`. Cross-checking with the source code of PhotoCam, `WebcamError` and `FlashError` inherit from the builtin `Error` class.

Also, you can use `@private` to mark a method as private. Since this is an internally used callback, it qualifies in my opinion.

src/main/resources/META-INF/resources/primefaces/photocam/photocam.js
```diff
@@ -1163,6 +1202,33 @@ PrimeFaces.widget.PhotoCam = PrimeFaces.widget.BaseWidget.extend({
...
+    },
+    
+    /**
+     *  Retrieves the available video input device list 
+     *  @return {Promise<InputDeviceInfo[]>} Returns a promise to resolve the enumeration or null if the browser doesn't support media devices enumeration
+     */
+    getAvailableDevices: function() {
+        var result = null;
+        if (navigator.mediaDevices && navigator.mediaDevices.enumerateDevices) {
+            result = navigator.mediaDevices.enumerateDevices().then(devices => devices.filter(
```
blutorange: Lambdas are nice and easier to write, but unfortunately not supported by all target browsers yet, and not used anywhere else in the codebase afaik. Can you change that to `function(devices) { ... }`

melloware: @blutorange is correct. Only ECMASCRIPT5 features are used and supported in PF.

src/main/resources/META-INF/resources/primefaces/photocam/photocam.js
```diff
@@ -1163,6 +1202,33 @@ PrimeFaces.widget.PhotoCam = PrimeFaces.widget.BaseWidget.extend({
         } else {
             PrimeFaces.error('Capture error: AdvancedPhotoCam not attached to the camera');
         }
+    },
+    
+    /**
+     *  Retrieves the available video input device list 
+     *  @return {Promise<InputDeviceInfo[]>} Returns a promise to resolve the enumeration or null if the browser doesn't support media devices enumeration
```
blutorange: This is a little bit pedantic, but TypeScript types allow us to be explicit about the nullness of values (personally I prefer to use `undefined` in favor of `null`, but I don't think there's any consensus on that, so it's fine as it is):

```suggestion
     *  @return {Promise<InputDeviceInfo[]> | null} Returns a promise to resolve the enumeration or `null` if the browser does not support media devices enumeration.
```

But other than that, the JavaScript docs look pretty good : ) I'm glad it's being maintained by everybody now.

src/main/resources/META-INF/resources/primefaces/photocam/photocam.js
```diff
@@ -1092,12 +1103,19 @@ PrimeFaces.widget.PhotoCam = PrimeFaces.widget.BaseWidget.extend({
         if (!("autoStart" in this.cfg)) {
             this.cfg.autoStart = true;
         }
+
```
blutorange, line 1113: Could you add the new properties of the widget configuration to the JavaScript docs around line 1073? Like this:

```
 * @prop {string} cfg.device Suggests a video input device such as `user` (font camera) or `environment` (rear camera).
 * @prop {string} cfg.onCameraError Client side callback executed if the camera has an error. This string must be valid
 * JavaScript code. It is executed within the context of a function with a single parameter `errorObj` of type `Error`.
```

melloware (changes requested):
Once you make @blutorange recommended changes I think we will be good to merge.

doleron:
Sure.

melloware (approved):

PR #6012, opened 2020-06-08
@SafeVarArgs for renderPassThruAttributes

melloware:
@tandraschko after your Sonar cleanup my IDE was warning me to do the above PR.

Can you please review

### DIFF
--- a/src/main/java/org/primefaces/renderkit/CoreRenderer.java
+++ b/src/main/java/org/primefaces/renderkit/CoreRenderer.java
@@ -113,6 +113,7 @@ public abstract class CoreRenderer extends Renderer {
     @SafeVarargs
     protected final void renderPassThruAttributes(FacesContext context, UIComponent component, List<String>... attrs) throws IOException {
         if (attrs == null) {
+            renderDynamicPassThruAttributes(context, component);
             return;
         }
 
--- a/src/main/java/org/primefaces/renderkit/CoreRenderer.java
+++ b/src/main/java/org/primefaces/renderkit/CoreRenderer.java
@@ -23,6 +23,23 @@
  */
 package org.primefaces.renderkit;
 
+import java.io.IOException;
+import java.util.*;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
+import javax.el.PropertyNotFoundException;
+import javax.faces.component.*;
+import javax.faces.component.behavior.ClientBehavior;
+import javax.faces.component.behavior.ClientBehaviorContext;
+import javax.faces.component.behavior.ClientBehaviorHolder;
+import javax.faces.context.FacesContext;
+import javax.faces.context.ResponseWriter;
+import javax.faces.convert.Converter;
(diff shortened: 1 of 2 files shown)


src/main/java/org/primefaces/util/CalendarUtils.java
```diff
@@ -245,7 +245,6 @@ public static final String getValue(FacesContext context, UICalendar calendar, O
                     javax.faces.convert.DateTimeConverter nativeConverter = (javax.faces.convert.DateTimeConverter) converter;
                     if (PrimeApplicationContext.getCurrentInstance(context).getEnvironment().isAtLeastJsf23()) {
                         Field field = javax.faces.convert.DateTimeConverter.class.getDeclaredField("type");
-                        field.setAccessible(true);
```
melloware: Sonar recommended removing this accessible set and i verified it was having no effect.

src/main/java/org/primefaces/renderkit/CoreRenderer.java
```diff
@@ -113,8 +110,10 @@ protected void renderPassThruAttributes(FacesContext context, UIComponent compon
         renderDynamicPassThruAttributes(context, component);
     }
 
-    protected void renderPassThruAttributes(FacesContext context, UIComponent component, List<String>... attrs) throws IOException {
+    @SafeVarargs
+    protected final void renderPassThruAttributes(FacesContext context, UIComponent component, List<String>... attrs) throws IOException {
         if (attrs == null) {
+            renderDynamicPassThruAttributes(context, component);
```
melloware, line 116: If varargs is null the old code still called renderDynamicPassthrough so this puts that back.  Looks at ColorPickerRenderer as an example

tandraschko:
Please dont remove setAccessible. I also wonder why reflection is used over #getType?

melloware:
I use reflection for getType because type is new for JSF 2.3 so to make it backwards compatible reflection must be used to not cause classloader issues.

tandraschko:
Can you recheck this?
https://myfaces.apache.org/core20/myfaces-api/apidocs/javax/faces/convert/DateTimeConverter.html
MF has it since 2.0

melloware:
OK it does look like its always been there but I think the Javadoc is wrong for 2.3.
https://myfaces.apache.org/core23/myfaces-api/apidocs/javax/faces/convert/DateTimeConverter.html#getType--

it says "_Specifies whether the date, time, or both should be parsed/formatted. Values include: date, time, and both. Default based on setting of timeStyle and dateStyle_."

But now the values for 2.3 according to Mojarra:

"_Set the type of value to be formatted or parsed. Valid values are both, date, time localDate, localDateTime, localTime, offsetTime, offsetDateTime, or zonedDateTime. The values starting with "local", "offset" and "zoned" correspond to Java SE 8 Date Time API classes in package java.time with the name derived by upper casing the first letter. For example, java.time.LocalDate for the value "localDate". An invalid value will cause a ConverterException when getAsObject() or getAsString() is called._"

I am going to submit another PR against the original ticket number where I fixed this.

PR #6021, opened 2020-06-10
#3324 MultiViewState optionally per ClientWindow instead of session

### LINKED ISSUE #3324: DataTable: Allow to store TableState for multiViewState in another scopes
## 1) Environment
- PrimeFaces version: 6.2 RC2

## 2) Expected behavior
I would like to have a possibility to use another scope than session scope for storing TableStates.  Would it be possible to allow to define own inplementation of let's say "storage provider" with methods?

org.primefaces.PrimeFaces.**clearTableStates()**
org.primefaces.PrimeFaces.**clearTableState(String key)**
org.primefaces.component.datatable.DataTable.**getTableState(boolean create)**

Using session scope, consumer has to handle clearing of the map in many cases. In our app we use other scopes like Spring's Conversation or Flow scope which we could use to store data for this functionality and they would be cleared automatically when not needed.
Another issue is multi-window usage, where the same table shares it's state with another browser window.

christophs78:
I need to add at least some documentation.

Showcase-PR: https://github.com/primefaces/primefaces-showcase/pull/164

Is there a need to add some cleanup-code for clientWindowMap and remove "old" clientWindows?
IMO the amout of data is quite small so it should not be necessary.

### DIFF
--- a/docs/9_0/core/multiviewstate.md
+++ b/docs/9_0/core/multiviewstate.md
@@ -1,0 +1,33 @@
+# MultiViewState
+
+MultiViewState feature enables components to maintain their state across pages by setting multiViewState attribute to true. 
+
+Following components support MultiViewState:
+
+- [DataTable](/components/datatable.md)
+- [DataList](/components/datalist.md)
+- [DataGrid](/components/datagrid.md)
+- [DataView](/components/dataview.md)
+- [TabView](/components/tabview.md)
+
+Per default MultiViewState is stored per session. As an alternative it can bis stored per client window to support multi-window usage.
+
+This can be configured via
+```xml
+<context-param>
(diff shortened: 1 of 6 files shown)


tandraschko:
1) i would use ConcurrentHashMaps
2) use LangUtils#isBlank
3) update documentation
4) the context-param should have 2 options: session, client-window (all our options are lower case)
5) rename getMVSSessionMap to getMVSStorate or similar

otherwise + 1 :)

christophs78:
Good points. Done.
Switched to ConcurrentHashMaps for both Maps. Ok?
Adding the new option to configuration.md enough or should we add an explicit multiviewstate.md?

tandraschko:
the docu still has `client_window` isntead of `client-window`

it would be great if you can make a global docu about multiviewstate

tandraschko:
looks good, thanks!

Issue #109, opened 2015-04-19
Immediate form evaluation for queued requests with p:ajax


frederikd:
The behaviour of f:ajax (Mojarra) and p:ajax is currently quite different for queued requests.
In Mojarra the values that are supposed to be submitted as part of the ajax request are evaluated at the time the request is added to the queue, i.e. whatever the value was at the time the request is triggered is what is going to be submitted, irrespective of any changes to the value in the meantime.
In Primefaces the values are evaluated at the time the request is dequeued/processed, so any modifications to the form that happened after the request was triggred are reflected in the submission.

Both options seem reasonable to me as you might want to have different behaviour in different circumstances. E.g. if only "@this" is being submitted on an ajax request then you probably want the value at the time the event was triggerd, i.e. whatever the user entered, however if a field is being submitted as the result of another field being changed then you might want to have the latest value.

Our scenario is as follows:
We have pages with lots of fields (p:calendar, p:selectOneMenu, pe:inputNumber, h:inputText etc.) which all trigger ajax requests on change and re-render the parent panel. If the user tabs through and makes lots of changes quickly so that the ajax requests don't respond before the next change is made then some of the user input is lost, because the previous ajax request finishes late, potentially updates the next field and then p:ajax only evaluates the form input value when the request is dequeued. f:ajax on the other hand evaluates the the form input at the time of the request being added to the queue, so whatever the user entered will be sent, not whatever the components value is at the time the request is processed.

Ideally we would like to have p:ajax work like f:ajax, i.e. immediate evaluation rather than delayed evaluation, but it might make sense to add an option (either globally or on the p:ajax element as an attribute) to toggle this behaviour.

This issue is logged as a result of a discussion on issue #23

cagataycivici:
@tandraschko @mertsincan What do you guys think?

tandraschko:
As i already stated, i like the idea and this will probably fix an old bug of all my apps.
I could take care of it but the change is to heavy for a elite release IMO.

cagataycivici:
Let's discuss a bit more after I review.

frederikd:
If there is any additional details you need from us or if we can help in any other way to make this feature happen then please let me know. We are very keen on making full use of p:ajax and would be very happy to contribute where we can. :)

tandraschko:
@cagataycivici I will try if it will fix my problems and create a patch. So you can review both.

cagataycivici:
[...]

PR #6023, opened 2020-06-11
Context Param docu updates

melloware:
@tandraschko The new param in @christophs78 PR made me realize i needed to update our Context-Param docs.  I have done that with this PR but there is one property we are missing that I stared at the code and did Google Searches to see if there was an old PrimeTek Blog Post about but I can't find anything. It is...

`primefaces.EARLY_POST_PARAM_EVALUATION`

The JS code in core.ajax.js doesn't make it any clearer what its actually for???
```javascript
/**
             * Performs the early collection of post parameters (form element values) if the request is configured that
             * way.
             * @param {Partial<PrimeFaces.ajax.Configuration>} cfg Configuration for the AJAX request to send, such as
             * the HTTP method, the URL, and the content of the request.
             * @return {PrimeFaces.ajax.ServerCallbackParameter[]} The collected form element values to be sent with the request.
             */
            collectEarlyPostParams: function(cfg) {

                var earlyPostParams;

                var sourceElement;
                if (typeof(cfg.source) === 'string') {
                    sourceElement = $(PrimeFaces.escapeClientId(cfg.source));
                }
                else {
                    sourceElement = $(cfg.source);
                }
                if (sourceElement.is(':input') && sourceElement.is(':not(:button)')) {
                    earlyPostParams = [];

                    if (sourceElement.is(':checkbox')) {
                        var checkboxPostParams = $("input[name='" + $.escapeSelector(sourceElement.attr('name')) + "']")
                                .filter(':checked').serializeArray();
                        $.merge(earlyPostParams, checkboxPostParams);
                    }
                    else {
                        earlyPostParams.push({
                            name: sourceElement.attr('name'),
                            value: sourceElement.val()
                        });
                    }
                }
                else {
                    earlyPostParams = sourceElement.serializeArray();
                }

                return earlyPostParams;
            },
```

### DIFF
--- a/docs/7_0/gettingstarted/configuration.md
+++ b/docs/7_0/gettingstarted/configuration.md
@@ -13,14 +13,15 @@ pattern of JavaEE. Here is the list of all configuration options defined with a 
 
 | Name | Default | Description |
 | --- | --- | --- |
-| THEME | omega | Theme of the application. |
+| THEME | aristo | Theme of the application. |
 | SUBMIT | full | Defines ajax submit mode, full or partial. |
 | DIR | ltr | Defines orientation, ltr or rtl. |
 | RESET_VALUES | false | When enabled, ajax updated inputs are reset. |
-| CLIENT_SIDE_VALIDATION | false | Controls client side validatation. |
+| CLIENT_SIDE_VALIDATION | false | Controls client side validation. |
 | UPLOADER | auto | Defines uploader mode; auto , native or commons. |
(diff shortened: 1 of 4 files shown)


tandraschko:
See https://github.com/primefaces/primefaces/issues/109

melloware:
Thanks for that hint!  Updated the docu and added the GitHub ticket as a reference for future people like me when trying to figure out what it does.

For 9.0 I also sorted them alphabetically to make it easier to read and find.

PR #6035, opened 2020-06-16
Ajax uploads are not supported in simple upload.

TownCube:
The doco states "Ajax uploads are not supported in simple upload" and all examples of simple file upload require the Ajax attribute on the command button to be disabled.

This commit makes that consistent with a new example added to the doco in PF 8.


tandraschko:
Thanks!

PR #6041, opened 2020-06-17
Missing Secure Attribute in Encrypted Session (SSL) Cookie 

### LINKED ISSUE #6040: DataExporter  : Missing Secure Attribute in Encrypted Session (SSL) Cookie
**Environment:**
 - PF Version: _7.0.14
 - JSF + version: Mojarra 2.2.13
 - Affected browsers: ALL

**To Reproduce**
1. Go to https://www.primefaces.org/showcase/ui/data/dataexporter/basic.xhtml
2. Click on chrome developer tools
image

3. Secure property is not set for primefaces.download when using with HTTPS

**Expected behavior**
When using with HTTPS the cookie.setSecure(request.isSecure()) is not for primefaces.download

https://stackoverflow.com/questions/62418321/ibm-appscan-missing-secure-attribute-in-encrypted-session-ssl-cookie/62418437#62418437

ravihariharan:
PR to fix set secure attribute for primefaces.download

https://github.com/primefaces/primefaces/issues/6040

### DIFF
--- a/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
+++ b/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
@@ -396,8 +396,9 @@ public abstract class DataTableExporter implements Exporter<DataTable> {
         final boolean secure = PrimeRequestContext.isSecure(externalContext);
         Map<String, Object> map = null;
         if (secure) {
-            map = new HashMap<String, Object>();
+            map = new HashMap<String, Object>(2);
             map.put("secure", secure);
+            map.put("sameSite", "Strict");
             externalContext.addResponseCookie(Constants.DOWNLOAD_COOKIE, "true", map);
         }
         else {
--- a/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
+++ b/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
@@ -28,6 +28,7 @@ import org.primefaces.component.datatable.DataTable;
 import org.primefaces.component.export.ExportConfiguration;
 import org.primefaces.component.export.Exporter;
 import org.primefaces.component.overlaypanel.OverlayPanel;
+import org.primefaces.context.PrimeRequestContext;
 import org.primefaces.model.LazyDataModel;
 import org.primefaces.util.ComponentUtils;
 import org.primefaces.util.Constants;
@@ -43,7 +44,6 @@ import javax.faces.component.visit.VisitResult;
 import javax.faces.context.ExternalContext;
 import javax.faces.context.FacesContext;
 import javax.faces.convert.Converter;
-import javax.servlet.http.HttpServletRequest;
 
 import java.io.IOException;
 import java.lang.reflect.Array;
@@ -393,15 +393,17 @@ public abstract class DataTableExporter implements Exporter<DataTable> {
     }
 
     protected void addResponseCookie(ExternalContext externalContext) {
-        final boolean secure = ((HttpServletRequest) externalContext.getRequest()).isSecure();
+        final boolean secure = PrimeRequestContext.isSecure(externalContext);
+        Map<String, Object> map = null;
         if (secure) {
-            Map<String, Object> map = new HashMap<String, Object>();
+            map = new HashMap<String, Object>();
             map.put("secure", secure);
             externalContext.addResponseCookie(Constants.DOWNLOAD_COOKIE, "true", map);
         }
         else {
-            externalContext.addResponseCookie(Constants.DOWNLOAD_COOKIE, "true", Collections.<String, Object>emptyMap());
+            map = Collections.<String, Object>emptyMap();
         }
+        externalContext.addResponseCookie(Constants.DOWNLOAD_COOKIE, "true", map);
     }
 
 }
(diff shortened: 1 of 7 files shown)


tandraschko:
1) you must also take care of PortletRequest. We already have this check somewhere and reuseable.
2) AFAIR the Cookie is also created/removed on clientside

src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
```diff
@@ -379,4 +384,24 @@ public int invoke(FacesContext context) {
...
+        externalContext.setResponseHeader("Expires", "0");
+        externalContext.setResponseHeader("Cache-Control", "must-revalidate, post-check=0, pre-check=0");
+        externalContext.setResponseHeader("Pragma", "public");
+        externalContext.setResponseHeader("Content-disposition", contentDisposition);
+    }
+
+    protected void addResponseCookie(ExternalContext externalContext) {
+        final boolean secure = ((HttpServletRequest) externalContext.getRequest()).isSecure();
+        if (secure) {
+            Map<String, Object> map = new HashMap<String, Object>();
```
melloware, line 398: Initialize the HashMap size to 1 since you are only adding one value `new HashMap<>(1);`  You can just use the diamond operator as we are on Java 8

src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
```diff
@@ -379,4 +384,24 @@ public int invoke(FacesContext context) {
...
+            map.put("secure", secure);
+            externalContext.addResponseCookie(Constants.DOWNLOAD_COOKIE, "true", map);
+        }
+        else {
+            externalContext.addResponseCookie(Constants.DOWNLOAD_COOKIE, "true", Collections.<String, Object>emptyMap());
+        }
+    }
+
```
melloware, line 409: I think a better way to write this method is:
```java
        protected void addResponseCookie(ExternalContext externalContext) {
            final boolean secure = ((HttpServletRequest) externalContext.getRequest()).isSecure();
            Map<String, Object> map;
            if (secure) {
                map = new HashMap<>(2);
                map.put("secure", secure);
                map.put("sameSite", "Strict");
            }
            else {
                map = Collections.emptyMap();
            }
            externalContext.addResponseCookie(Constants.DOWNLOAD_COOKIE, "true", map);
        }
```

I also future proofed it by adding `sameSite=Strict` even though JSF and Servlet spec does not support it yet it hopefully will in the future as this is becoming a requirement of most browsers.

melloware (changes requested):

melloware:
Hopefully the SameSite attribute will be included in Servlet 5.0: https://github.com/eclipse-ee4j/servlet-api/pull/271

src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
ravihariharan: > 1. you must also take care of PortletRequest. We already have this check somewhere and reuseable.

Had to make a minor refactor to support passing an ExternalContext

> 2. AFAIR the Cookie is also created/removed on clientside

src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
```diff
@@ -379,4 +384,27 @@ public int invoke(FacesContext context) {
...
+    protected void addResponseCookie(ExternalContext externalContext) {
+        final boolean secure = PrimeRequestContext.isSecure(externalContext);
+        Map<String, Object> map = null;
+        if (secure) {
+            map = new HashMap<String, Object>(2);
+            map.put("secure", secure);
+            map.put("sameSite", "Strict");
+            externalContext.addResponseCookie(Constants.DOWNLOAD_COOKIE, "true", map);
```
melloware: You can remove this line you are adding it below

melloware (changes requested):

melloware (approved):

melloware:
I just created this ticket now that i realized we are creating client side cookies.
https://github.com/primefaces/primefaces/issues/6045

tandraschko:
i think it's better to just use:
`PrimeRequestContext.getCurrentInstance(context).isSecure()`

and cache the result in `PrimeRequestContext#isSecure`

ravihariharan:
> i think it's better to just use:
> `PrimeRequestContext.getCurrentInstance(context).isSecure()`
> 
> and cache the result in `PrimeRequestContext#isSecure`

Refactored the method to pass FacesContext in place of ExternalContext.

melloware (approved):
@tandraschko You have this comment in `isSecure` of PrimeRequestContext
```java
// currently called once per request - later we might cache the result per request
// and even the method lookup
```

src/main/java/org/primefaces/context/PrimeRequestContext.java
```diff
@@ -207,22 +208,25 @@ public void release() {
      * @return if secure or not.
      */
     public boolean isSecure() {
-        // currently called once per request - later we might cache the result per request
-        // and even the method lookup
-        Object request = context.getExternalContext().getRequest();
+        // currently called once per request
```
melloware: This comment is no longer true now you are caching it after the first result.  @tandraschko I think this is OK because i doubt any site has a mix of HTTP and HTTPS.  At least they should not!

melloware (changes requested):

melloware (approved):

PR #6046, opened 2020-06-17
Fix #6045: Cookies add security if over HTTPS

### LINKED ISSUE #6045: Cookies: Add security to client side cookies
**Description**
Any cookies created on the client side over HTTPS should have `SameSite=Strict` and `secure=true`.

See: https://web.dev/samesite-cookies-explained/

All client side generated cookies over HTTPS should have `SameSite=Strict` and `secure=true`.

melloware:
- removed deprecated jquery.cookie.js plugin
- upgraded to js.cookie which is the pure Javascript successor library to jquery.cookie and supports SameSite
- Add missing `stateful` documentation of Carousel
- Fix JsDocof browser plugin
- Tested in IE11, Edge, Firefox and Chrome

@blutorange please review my JSDoc changes I think I did it right
@ravihariharan this goes along with your other fix


src/main/resources/META-INF/resources/primefaces/core/js.cookie.d.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * Namespace for the cookie Javascript Cookie, available as `Cookies.set` and `Cookies.get`.
+ * 
+ * Contains some additional types and interfaces required for the typings.
+ * https://github.com/js-cookie/js-cookie
+ */
+declare namespace CookieJs {
```
blutorange: @melloware  The JSDocs basically look good to me. If it compiles with the `-P jsdoc` profile, it should be fine.

One minor thing though: `js-cookie` has semi-official type declarations [on DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/js-cookie). Currently, these are for version 2.2 (as js-cookie has only a release candidate of version 3 so far). Eventually we should just use that (installed via npm during the build process).

In preparation for that, it would be good to rename the namespace and interface so that it matches what's on DefinitelyTyped:

```typescript
declare namespace Cookies {
	/**
	 * Represents the properties of a cookie, other that its name and value.
	 */
	export interface CookieAttributes {
        }
}
```

(and change the parts in core.js accordingly)

melloware:
OK I think I fixed it the way you wanted it.

blutorange:
Yeah, seems right to me.

melloware:
@blutorange It looks like js-cookie def even though it says 2.2 is actually up to date docs for 3.0 and has the signature we want.  https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/js-cookie/index.d.ts

How do I switch to use js-cookie in package.json instead of having this placeholder .ts file for Cookies?

blutorange:
@melloware Well, it's definitely not up to date for 3.0: e.g. it contains a `getJSON` method even though [that was removed](https://github.com/js-cookie/js-cookie/releases/tag/v3.0.0-beta.0), lacks the [new builder API](https://github.com/js-cookie/js-cookie/releases/tag/v3.0.0-beta.1) etc. I might take a look at update the declarations on DefinitelyTyped though, shouldn't be too hard as it's not a large library.

melloware:
Yep its not up to date but i am much more comfortable referencing it that having our own docs.  Assuming someone updates it someday and then an "npm update" should fix us up to the latest version of the type.  I really appreciate all your help on this stuff I am finally starting to get it. 👍

PR #6048, opened 2020-06-17
Fix #6047: DatePicker round stepMinute correctly

### LINKED ISSUE #6047: DatePicker: stepMinute=15 incorrect behavior
**Description**
I have a field which is DatePicker with `stepMinute = 15`. The problem is the minutes start not from zero but from some random number (probably from the current time). So increasing or decreasing minutes make those steps like `11, 26, 41, 56`. But the way I'd like and how I think it should work is to start from `00` and then increase to `15, 30, 45`. I think the initial value must be somewhere to the closest interval (eg. if now is `14:35` then it must be `14:30`).

StepMinute should round to the closest minute interval.

Similar issue was reported to PrimeVue: https://github.com/primefaces/primevue/issues/115

**Reproducer:**
[pf-stepMinute.zip](https://github.com/primefaces/primefaces/files/4794268/pf-stepMinute.zip)

**Example XHTML**
```xml
<h:form id="frmTest">
    <p:datePicker id="dpStep" 
				value="#{testView.timestamp}" 
				showTime="true" 
				hourFormat="12"
				stepMinute="15" 
				pattern="MM/dd/yyyy HH:mm" 
				required="true">
	</p:datePicker>
</h:form>
```

**Example Bean**
```java
import java.io.Serializable;
import javax.inject.Named;
import java.time.*;

@Named
@ViewScoped
    
    private LocalDateTime timestamp;
    
    @PostConstruct  
    public void init() {
       
    }

    public LocalDateTime getTimestamp() {
        return timestamp;
    }

    public void setTimestamp(LocalDateTime timestamp) {
        this.timestamp = timestamp;
    }    
}
```

melloware:


tandraschko:
+1 if well tested

melloware:
We discovered this because we had a client requirement and we are already using this fix in the app so I have tested it pretty good.

I did just have to make a fix because our Tester realized you could still type in the Input Text an invalid minute like 17 so this fixes that as well.

melloware:
Our team is still testing today.

Issue #5755, opened 2020-04-01
Tree: Tree loads the tree model from bean again for each TreeNode


yennor:
## 1) Environment
- PrimeFaces version: 8

## 2) Expected behavior
Tree component should load the tree model from bean only once for each update (render) of the tree component.

## 3) Actual behavior
It seems to load it as many times as there are treenodes in the treemodel.
This can lead to significant higher load times, especially if in some calculations are made in the getter (e.g. Because the data shown in the tree component depend on an other component)

## 4) Steps to reproduce
see
https://github.com/yennor/primefaces-test/tree/tree-value

tandraschko:
please give a bit better description; What is loaded as many times? Which method exactly?

yennor:
sorry, I'm tired. anyway:
`<tree value="{#view.treeRoot}"`

In getTreeRoot() you get the root node with all the other nodes attached.
For every node attached getTreeRoot() gets called again.
In the git repository referenced you can see it nicely, even with counter.

melloware:
@yennor yes this is how JSF works.

Which is why if you look at this page on performance: https://primefaces.github.io/primefaces/8_0/#/core/performance

> Try to avoid logic in getters because they can be called multiple times - especially for the rendered attribute!

You should create your treeRoot model once and then in the getter just return the instance of it don't perform any logic or do it like this...

```java
public TreeNode getTreeRoot() {
    if (treeRoot == null) {
        treeRoot = initiializeTreeRoot();
   }
   return treeRoot;
}
```

it should not be impacting your performance at all if there is no logic in the getter.  You can call a getter 1000 times in a loop if it only returns and instance and Java will do that in milliseconds.

yennor:
ok. But it still doesn't make any sense.
I've just tried it out with the datatable. Even when the datatable has got 50 rows. It doesn't call the getter for the row list 50 times. Only two times.
When the tree component got 50 nodes, it calls the getter for the root node 50 times...
Which just doesn't make any sense. It already got all of the nodes with the first call, it doesn't need to call for each subnode again.

Yes, I can change the logic and create the nodes in the setter of the data the tree component depends on or implement an extra listener. But you won't change my opinion in that that was implemented in a sane way ;-).

melloware:
Well whether it’s implemented incorrectly or not is neither here nor there. You violated one of Joshua Bloch’s cardinal rules of never put logic in setters or getters. Use them as accessors only.  

[...]

PR #6051, opened 2020-06-17
Fix #6050 - TreeTable: ajax events not fired from column facets

### LINKED ISSUE #6050: TreeTable: ajax events not fired from column facets
**Describe the defect**
Ajax events are not fired from column facets, as a consequence listeners are not called.

**Environment:**
 - PF Version: 9.0-SNAPSHOT
 - JSF + version: Mojarra 2.3.2
 - Affected browsers:  ALL

**Expected behavior**
Listeners should be called

**Example XHTML**
```html
<h:form id="frmTest">
   <p:treeTable value="#{ttBasicView.root}" var="document">
      <p:column headerText="Name" filterBy="#{document.name}">
         <f:facet name="filter">
            <p:selectOneMenu value="#{selectOneMenuView.option}" >
               <p:ajax event="itemSelect" listener="#{ttBasicView.dummyAction}" />
               <f:selectItem itemLabel="Select One" itemValue="" />
               <f:selectItems value="#{selectOneMenuView.options}" />
            </p:selectOneMenu>
         </f:facet>
         <h:outputText value="#{document.name}" />
      </p:column>
   </p:treeTable>
</h:form>
```

**Example Bean**
```java
@Named("ttBasicView")
@ViewScoped
public class BasicView implements Serializable {

    // showcase bean

    public void dummyAction() {
        System.out.println("I was called");
    }
}
```

Rapster:
Let me know what you think. I could have reused `visitFacets` but the impl is slightly different from what I read in DataTable (e.g setRowKey call). By the way, do we need to call setRowKey while it's been previously called in UITree#visitTree l.812? This method is very consuming at it saves/restore state of the component (recursive call) so if we could get rid of that call that would be a good improvement

### DIFF
--- a/src/main/java/org/primefaces/component/api/UITree.java
+++ b/src/main/java/org/primefaces/component/api/UITree.java
@@ -39,6 +39,8 @@ import javax.faces.component.visit.VisitResult;
 import javax.faces.context.FacesContext;
 import javax.faces.event.*;
 
+import org.primefaces.component.column.Column;
+import org.primefaces.component.columngroup.ColumnGroup;
 import org.primefaces.component.columns.Columns;
 import org.primefaces.component.tree.UITreeNode;
 import org.primefaces.model.CheckboxTreeNode;
@@ -1007,6 +1009,9 @@ public abstract class UITree extends UIComponentBase implements NamingContainer 
                         }
                     }
                 }
+                else if (child instanceof ColumnGroup) {
+                    visitColumnGroup(context, callback, (ColumnGroup) child);
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


melloware:
I know @tandraschko did some performance improvements to Tree maybe you can do something similar here to reduce the calls: https://github.com/primefaces/primefaces/issues/5755

tandraschko:
will try to review next week

Rapster:
Quick question: on a general note, when state shoud be saved and restored?

melloware:
> Quick question: on a general note, when state shoud be saved and restored?

I don't know the answer to this one...

PR #6054, opened 2020-06-18
Fix #6053: Switch client cookies to HTML Local Storage

### LINKED ISSUE #6053: Storage: Switch from cookies to HTML5 local storage
**Description**
Based on a conversation with @GedMarc cookies were used for storing client side state because legacy browers like IE7 could only support cookies.  

Now that PF supports IE11 and higher we should switch client cookie usage for HTML5 LocalStorage which has the same lifespan as a cookie does but allows for more than 4KB of values and is the proper way to handle this type of behavior today.

**Additional context**
https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage

melloware:
- Tested and works perfectly
- Solves any future cookie security issues by not using cookies

@blutorange @ravihariharan


GedMarc:
datatables :)

melloware:
@GedMarc I found every reference to these PrimeFaces.getCookie() etc methods and only PlainMenu, PanelMenu, and Carousel came up in my search.  Does Datatable also use cookies?

tandraschko:
What about PrimeFaces.download Cookie? It was used on the clientside as well

melloware:
Yep let me look at the download cookie i see it now.

melloware:
Had to restore cookie methods to support file download.  As far as I can tell its the only place in all of PF now using cookies.

tandraschko:
TBH i would remove our setStore etc. wrappers
we would need to call them setLocalStore as we soon need sessionStorage for ClientWindow.

melloware:
I like that. let me make that change.

GedMarc:
ahh not for data tables 1.10 and up - i'm out of date 
https://datatables.net/forums/discussion/26654/prevent-datatables-from-creating-cookies

tandraschko:
we dont use "datatables" - we have own script

@melloware 
we probably need to add the contextPath somehow
the storage shares between domains

melloware:
@tandraschko I am confused i thought local storage was domain based already: https://jcubic.wordpress.com/2014/06/20/cross-domain-localstorage/

So are we saying we do NOT want it domain based?

GedMarc:
oo, ok so i understand as 

* localstorage is a perm store, cookies can be a perm store
* cookies can have expiry, localstorage requires manual expiry 

localstorage is for state (client side), cookies are for server-side identifiers, usually only identity

localstorage domains are separated by sub domain

it's about what data is required on the server and what isn't :) sessionstorage comes into play when doing 2-phase authentication, perm logins, long lasting tokens etc that you want killed on tabclose, client side ViewScope data as it were

change of behavior will be experienced when the cookie default expiry would clear the data, now the data will persist until cleared

tandraschko:
@melloware domains are ok
but look how stateKeys are are generated
`this.stateKey = 'carousel-' + this.id;`

this misses the viewId + contextPath

currently the state could unintentionally shared between views and even between webapps inside the same domain

melloware:
AHH ok so right you are saying if someone had a carousel on 2 different pages and both were called `id="carouselTest"` it would share that config.  So by adding viewId and context it makes it unique per page per site.  I understand now.  Let me work on that.

melloware:
OK I am seeing something really strange even on the showcase and I tested my production apps as well.  I see this code in HeadRenderer:
```java
        writer.write("if(window.PrimeFaces){");
        writer.write("PrimeFaces.settings.locale='" + LocaleUtils.getCurrentLocale(context) + "';");
        writer.write("PrimeFaces.settings.viewId='" + context.getViewRoot().getViewId() + "';");
        writer.write("PrimeFaces.settings.contextPath='" + context.getExternalContext().getRequestContextPath() + "';");
```

However in my UI when I look at page source the values for viewId and context are never there...only the locale is coming through?
```javascript
--
if(pf){pf.settings.locale='en';pf.settings.validateEmptyFields=true;pf.settings.considerEmptyStringNull=true;};
```

melloware:
OK its MOVE_SCRIPTS_TO_BOTTOM. When I turn it off I get the full values.
```javascript
if(window.PrimeFaces){PrimeFaces.settings.locale='en';PrimeFaces.settings.viewId='/index.xhtml';PrimeFaces.settings.contextPath='/showcase';PrimeFaces.settings.validateEmptyFields=true;PrimeFaces.settings.considerEmptyStringNull=true;}
```

tandraschko:
weird
would be great if you can check it.

melloware:
@tandraschko I am an idiot. You just added those in 9.0 and I am testing with 8.0 and showcase is on 8.0 so of course those props are not there!  I had just assumed they had been there forever.  OK now I can fix this ticket.

melloware:
OK updated and now it generates a unique key like this:
```
carousel-form:j_idt729-ui-data-carousel.xhtml-showcase
```
Component + ID + View + ContextPath

tandraschko:
i would change it:
context - view - component - id

melloware:
OK i switched it to your order.

tandraschko:
i thought about it and i think we should switch a bit:
context - view - id - key (instead of component; i think some components could store different stuff under different keys)

melloware:
@tandraschko Good idea I will fix...

tandraschko:
perfect
last cosmetics: could we switch the id and key method param? id is important, key probably optional

tandraschko:
Great PR, thank you!

tandraschko:
do you know whats wrong with the travis build?

melloware:
Hmm looking at it is just saying it failed the JSDOC step parsing core.js but it doesn't say what is failing.

melloware:
nevermind I see it.  Will fix.

schaeferm:
Hi there. I really like the idea of this fix. However I noticed that this is a big change from the stateful cookie based implementation. With the old implementation one could create menus with shared state across multiple pages. This is no longer possible and now possibly messes with many main menus. Can you please implement a parameter that allows for the old global state saving instead of the new view based one?

tandraschko:
sounds weird
localStorage is shared between all pages in the same application?

schaeferm:
But the generated keys are per page and so the menu now saves the state per page.

melloware:
@schaeferm is correct we use the view based ID so each page gets a unique local storage.  We would need a new option for Menu to be global which would generate its key not using the ViewId.

melloware:
@schaeferm can you please open a new ticket.

PR #6062, opened 2020-06-22
Fix #3240, #3732, #1892, #324: InputMask upgrade JS library

### LINKED ISSUE #324: Calendar: input using mask can't be cleared
Hello,
we're using Primefaces 5.2.3. Our component making problems is a calendar with a mask set. If it is filled with a value and we remove this and click out of the calendar, the value is automatically entered again. So we currently can't delete the values from the calender if it should be empty... The only workaround I found is using PF(...).setDate(null). But this is just ugly and doesn't work as expected. Is there any other solution?

### LINKED ISSUE #1892: InputMask: incomplete submit value when mask contains '?' and a character at the end
## 1) Environment
- PrimeFaces version: 5.3 or 6.0
- Application wildfly-9.0.0
- Affected browsers: Firefox 49.0.2, Chrome 54.0.2840.71 m, Edge 38.14393.0.0

## 2) Expected behavior
<p:inputMask id="nss" value="#{myBean.number}" mask="9.99.999.?(99)">

with input '12345678' i expecting '1.23.456.78)'

## 3) Actual behavior
result is '1.23.456.(78', missing the **last character**

any combination with '?' and a character on the end of the mask. The last character will not be submited.

### LINKED ISSUE #3240: InputMask: Upgrade JS library
As stated [here](https://github.com/digitalBush/jquery.maskedinput), jquery.maskedinput is no longer maintained. The authors recommend using [cleave.js](https://nosir.github.io/cleave.js/). Maybe PrimeFaces should follow this recommendation in 6.3?

> The time has come to officially call it quits. The web has changed(A LOT) and there are better things out there like Cleave.js. I'll leave this repo up for posterity in an archived state. Thank you to everyone who contributed to or used this plugin over the years.

### LINKED ISSUE #3732: Calendar: Today-button not working if mask="true"
## 1) Environment
- PrimeFaces version: 6.2

- Affected browsers: Firefox, Chromium

## 2) Expected behavior
Current date should be set.

## 3) Actual behavior
Date is showing in the component but is overwritten with the previous value instantly.

## 4) Steps to reproduce

.. 

## 5) Sample XHTML

<p:calendar mask="true" showButtonPanel="true" />

## 6) Sample bean

..

melloware:
The only breaking change which I need to add to the WIKI is that the optional no longer starts with `?` you have to wrap the optional section in `[` and `]` brackets.

However I have confirmed this new library fixes all of the issue listed in the commit message.


melloware:
@blutorange I would like to remove our custom jqueyr.maskedinput.ts and replace with the new one from DefinitelyTyped: https://www.npmjs.com/package/@types/inputmask

How do I do that?

melloware:
OK I think I figured it out and now I am starting to get it!

blutorange:
@melloware Yeah, I think you got it right: add it to the `src/main/type-definitions/package.json`. And since the travis build seems to have been successful, you should be good.

melloware:
InputMask is a large enough lib that I decided to move it out of Jquery Plugins and make it a normal widget in `inputmask/inputmask.js`

PR #6069, opened 2020-06-23
Fix #6068: Validate minDate < maxDate

### LINKED ISSUE #6068: DatePicker\Calendar: Validate minDate is before maxDate
**Describe the defect**
When you invoke the datepicker with the attributes maxdate and mindate while maxdate < mindate holds true, then the datepicker looks corrupted. The month/year is not shown correct.

**Environment:**
 - PF Version: _8.0_
 - JSF + version: _e.g. Mojarra 2.2.8 or MyFaces 2.3.6_

**To Reproduce**
When you invoke the datepicker with the attributes maxdate and mindate wihle maxdate < mindate holds true, then the datepicker looks corrupted. the month/year is not shown correct.

**Expected behavior**
Actually in such a case, when maxdate < mindate  holds true, a warning should be shown that maxdate should not be before mindate. Or you even do not show the datepicker at all since you cannot use anyhow.

melloware:


tandraschko:
I think we should atlest add the clientId to the exception

melloware:
Message is now either:
```
DatePicker : "form:minmax" minimum date must be less than maximum date.
```
or
```
Calendar : "form:minmax" minimum date must be less than maximum date.
```

melloware:
I wonder if I should also do `< 0` instead of `== -1`.  Because if the min and max are equal that doesn't make much sense either does it?

GedMarc:
min+max = 1 day selectable?

melloware:
Yeah you are right if you are using whole days and not time you could theoretically make the calendar have 1 day selectable. I will leave as is.

PR #6073, opened 2020-06-23
Fix #6030: Refactor cookie handling

### LINKED ISSUE #6030: Provide native ClientWindow implementation
In the past MyFaces Orchestra/CODI/DeltaSpike had a Window/TabScope. This was later promoted to JSF 2.2.
So JSF itself provides a good abstraction and API but the default implementation is very basic and has some limitations.

The DeltaSpike LAZY (*1) implementation is, compared to CLIENTWINDOW (*2), very easy to implement and only has a small downside:
When you have a active tab with e.g. windowId=1, and open a link in this tab with e.g. windowId=2, we need to refresh the current link to windowId=1 again.

Whats the difference to the default url mode in the JSF impl?
The windowId will be remembered in the sessionStorage/window.name and therefore we can validate it.
Also a initial redirect is performanced to add the windowId always to the url. Also ExternalContext#redirect is overwritten to add the windowId.

1) https://deltaspike.apache.org/documentation/jsf.html#LAZY
2) https://deltaspike.apache.org/documentation/jsf.html#CLIENTWINDOW

melloware:
- Client Window code refactored to use `PrimeFaces.setCookie`
- Server code refactored so all cookie calls go through one method.

@tandraschko @ravihariharan @christophs78

### DIFF
--- a/src/main/java/org/primefaces/clientwindow/PrimeClientWindowUtils.java
+++ b/src/main/java/org/primefaces/clientwindow/PrimeClientWindowUtils.java
@@ -26,8 +26,11 @@ package org.primefaces.clientwindow;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
+
 import javax.faces.context.FacesContext;
 import javax.servlet.http.HttpServletRequest;
+
+import org.primefaces.util.ResourceUtils;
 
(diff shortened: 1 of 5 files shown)


tandraschko:
+1

Issue #6077, opened 2020-06-25
ImageCropper v9 - File upload null pointer exception


GedMarc:
Uploaded file fails to read if locally referenced - 

ImageCropperRenderer - Null pointer exception 
```
ExternalContext externalContext = context.getExternalContext();
                        // GitHub #3268 OWASP Path Traversal
                        imagePath = FileUploadUtils.checkPathTraversal(imagePath);

                        String webRoot = externalContext.getRealPath(Constants.EMPTY_STRING);
                        String fileSeparator = Constants.EMPTY_STRING;
                        if (!(webRoot.endsWith("\\") || webRoot.endsWith("/")) &&
                                    !(imagePath.startsWith("\\") || imagePath.startsWith("/"))) {
                            fileSeparator = "/";
                        }
```

I think Line 223+ should read - this returns correctly
```
String webRoot = externalContext.getRealPath(imagePath);
String fileSeparator = System.getProperty("file.separator");
```

GedMarc:
```
ERROR org.primefaces.application.exceptionhandler.PrimeExceptionHandler - null
java.lang.NullPointerException: null
	at org.primefaces.component.imagecropper.ImageCropperRenderer.getConvertedValue(ImageCropperRenderer.java:225) ~[primefaces-0.0.0_0-SNAPSHOT.jar:?]
	at javax.faces.component.UIInput.getConvertedValue(UIInput.java:1105) ~[javax.faces-0.0.0_0-SNAPSHOT.jar:?]
	at javax.faces.component.UIInput.validate(UIInput.java:1006) ~[javax.faces-0.0.0_0-SNAPSHOT.jar:?]
	at javax.faces.component.UIInput.executeValidate(UIInput.java:1317) ~[javax.faces-0.0.0_0-SNAPSHOT.jar:?]
	at javax.faces.component.UIInput.processValidators(UIInput.java:733) ~[javax.faces-0.0.0_0-SNAPSHOT.jar:?]
	at javax.faces.component.UIComponentBase.processValidators(UIComponentBase.java:921) ~[javax.faces-0.0.0_0-SNAPSHOT.jar:?]
	at javax.faces.component.UIForm.processValidators(UIForm.java:229) ~[javax.faces-0.0.0_0-SNAPSHOT.jar:?]
	at javax.faces.component.UIComponentBase.processValidators(UIComponentBase.java:921) ~[javax.faces-0.0.0_0-SNAPSHOT.jar:?]
[...]

Issue #3268, opened 2018-02-01
ImageCropper: potential path traversal SONATYPE-2018-0608


cnsgithub:
## 1) Environment
- PrimeFaces version: primefaces-6.2.RC2-SNAPSHOT
- Application server + version: WildFly 11
- Affected browsers: all

## 2) Expected behavior
In `p:imageCropper` images should be read just from external HTTP URLs or resources relative to the web application's context root.

## 3) Actual behavior
If the value of `p:imageCropper image` attribute contains user-supplied input, images from arbitrary file system locations could be read and disclosed since `new File `is used without taking care of path traversal stuff. The restriction to `ExternalContext.getRealPath` already done in `ImageCropperRenderer ` is insufficient.

ImageCroperRenderer.java
```java
                boolean isExternal = imagePath.startsWith("http");

                if (isExternal) {
                    URL url = new URL(imagePath);
                    URLConnection urlConnection = url.openConnection();
                    inputStream = urlConnection.getInputStream();
                    contentType = urlConnection.getContentType();
                }
                else {
                    ExternalContext externalContext = context.getExternalContext();
                    File file = new File(externalContext.getRealPath("") + imagePath); // insufficient
                    inputStream = new FileInputStream(file);
                }
```

## 4) Steps to reproduce
Assuming that your exploded web archive is located here D:\Wildfly11\deploy\webapp.war (webapp.war is a folder), put a sample image file here: d:\private.png.

Navigate to cropper.xhtml in your browser. open console, enter `$("input[id$=coords]").val("0_0_100_100");` and press the button.

The sample image from d:\private.png will be shown after server response.

## 5) Sample XHTML
```xml
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:h="http://java.sun.com/jsf/html" xmlns:p="http://primefaces.org/ui">  
	<h:head>
		<title>imageCropper test</title>
	</h:head>
	<h:form>
	    <h:panelGrid columns="2">
	        <p:imageCropper value="#{cropperView.croppedImage}" image="#{cropperView.userSuppliedInput}" initialCoords="0,0,100,100" />
            <img src="cropped.png" />
	    </h:panelGrid>

	    <p:commandButton value="Crop" action="#{cropperView.crop}" icon="ui-icon-scissors"/>
	</h:form>
</html> 
``` 

## 6) Sample bean
```java
package de.test.primefaces;

import java.io.File;
import java.io.Serializable;
import javax.enterprise.context.SessionScoped;
[...]

PR #6078, opened 2020-06-25
Imagecropper

GedMarc:
#6077 

Use JDK 8 native Path for referencing, important for JDK 11 and up particularly with modules enabled


melloware:
@GedMarc looks like you have 3 checkstyle violations.
```
[ERROR] /home/travis/build/primefaces/primefaces/src/main/java/org/primefaces/component/imagecropper/ImageCropperRenderer.java:228: 'else' child has incorrect indentation level 28, expected level should be 24. [Indentation]
[ERROR] /home/travis/build/primefaces/primefaces/src/main/java/org/primefaces/component/imagecropper/ImageCropperRenderer.java:228:1: File contains tab characters (this is the first instance). [FileTabCharacter]
[ERROR] /home/travis/build/primefaces/primefaces/src/main/java/org/primefaces/component/imagecropper/ImageCropperRenderer.java:229: 'else' child has incorrect indentation level 28, expected level should be 24. [Indentation]
```

src/main/java/org/primefaces/component/imagecropper/ImageCropperRenderer.java
```diff
@@ -220,15 +224,9 @@ else if (imageObject instanceof StreamedContent) {
...
-                        if (!(webRoot.endsWith("\\") || webRoot.endsWith("/")) &&
-                                    !(imagePath.startsWith("\\") || imagePath.startsWith("/"))) {
-                            fileSeparator = "/";
-                        }
-
-                        File file = new File(webRoot + fileSeparator + imagePath);
-                        inputStream = new FileInputStream(file);
+                        String webRoot = externalContext.getRealPath(imagePath);
+	                    Path locationPath  = Paths.get(webRoot);
+	                    inputStream = Files.newInputStream(locationPath, StandardOpenOption.READ);
```
melloware, line 229: CheckStyle

melloware (changes requested):

GedMarc:
Hmm so I'm trying to isolate here, this looks to be a 2 parts here so I'm going to need a bit of guidance - 

This fix is working for me right now, i'll try clean it all up and see what repercussions  i have caused myself

The Paths should take care of everything linux/mac/ and updated. The NIO library has huge updates relating to file management, e.g. using FileStream here may be replaceable with FileChannel (MacOS gets usually a 60% file read increase, i'll find the link somewhere), 

For me in particular, JLink native imaging uses the JRT filesystem... which is readonly, so i have another problem where the uploaded files are trying to be stored in the classpath on a read only module native image xD i can route to a static location easy enough but can't seem to specify where to save the files

* How do I add a custom location to be specified for the uploaded files, in the showcase when i push the button it goes to /demo/images/....

:D :D

melloware:
I think it has to be relative to the web application root, and that is why we had to add the OWASP Path Traversal check: 
https://github.com/primefaces/primefaces/issues/3268

So there is no way to specify some other file location i think it has to be under the web app.  But I did not design this component  it has been like this for years.

GedMarc:
Oh dear.... 

We do either a static location on a server more designed for static content and reverse proxy a location into there
or save the data in a table to be served by a servlet at another location. 
Saving or updating the actual web application with file data, hmmm...

Can you lead me to the widget that is implemented? Is this something that we are open to? Retrieving the data in the same manner as file upload into a stream rather than saving it down to the file system?

For the showcase can always display the crop image in a base 64 data image saving on some disk space for the showcase server?

On the wildfly bug for 6.2, another solution may have been to go this route, and specify ${jboss.temp.directory} as the target directory to make sure it is retrievable on the same context url. It is possible to do the same behavior specifying file saving locations?

melloware:
So if you have 9.0-SNAPSHOT running locally and grab 9.0-SNAPSHOT Showcase from master you can see recently these 2 ticket were implemented: 
https://github.com/primefaces/primefaces/issues/12
https://github.com/primefaces/primefaces/issues/4837

And 2 new Showcase examples were added for ImageCropper `Dynamic` and `Combined With FileUpload`.

image

But yes the showcase is always saving the cropped images in \primefaces-showcase\src\main\webapp\resources\demo\images\crop

PR #6084, opened 2020-06-25
Fix #6083: Add version to theme.css URL

### LINKED ISSUE #6083: Theme: CSS is loaded without version
**Description**
PrimeFaces theme.css is loaded in `HeadRenderer` but does not append the version so its causing the browser to not refresh the cache on PF updates.

```xml
<link type="text/css" rel="stylesheet" 
         href="/primefaces-test/javax.faces.resource/theme.css.xhtml?ln=primefaces-omega" />
```

The version to be appended to the theme.css resource.

```xml
<link type="text/css" rel="stylesheet" 
         href="/primefaces-test/javax.faces.resource/theme.css.xhtml?ln=primefaces-omega&amp;v=8.0"" />
```

**Additional context**
The bug is in `PrimeResource` only looking for ln=primefaces exactly and not "startsWith" primefaces because the themes are `ln=primefaces-omega` etc.

melloware:
I left the original equalsIgnoreCase but added a startsWith before it to catch themes but also will catch 99% of other resources.  

I can be talked into removing the equalsIgnoreCase?

### DIFF
--- a/src/main/java/org/primefaces/application/resource/PrimeResourceHandler.java
+++ b/src/main/java/org/primefaces/application/resource/PrimeResourceHandler.java
@@ -25,6 +25,7 @@ package org.primefaces.application.resource;
 
 import java.io.IOException;
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -102,7 +103,7 @@ public class PrimeResourceHandler extends ResourceHandlerWrapper {
 
     private Resource wrapResource(Resource resource, String libraryName) {
         if (resource != null && libraryName != null
-                    && (libraryName.startsWith(Constants.LIBRARY) || libraryName.equalsIgnoreCase(Constants.LIBRARY))) {
+                    && (libraryName.toLowerCase(Locale.getDefault()).startsWith(Constants.LIBRARY))) {
             return new PrimeResource(resource);
         }
         else {
--- a/src/main/java/org/primefaces/application/resource/PrimeResourceHandler.java
+++ b/src/main/java/org/primefaces/application/resource/PrimeResourceHandler.java
@@ -28,10 +28,12 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+
 import javax.faces.application.Resource;
 import javax.faces.application.ResourceHandler;
 import javax.faces.application.ResourceHandlerWrapper;
 import javax.faces.context.FacesContext;
+
 import org.primefaces.application.resource.barcode.BarcodeHandler;
 import org.primefaces.util.Constants;
 import org.primefaces.util.LangUtils;
@@ -67,25 +69,13 @@ public class PrimeResourceHandler extends ResourceHandlerWrapper {
     @Override
     public Resource createResource(String resourceName, String libraryName) {
         Resource resource = super.createResource(resourceName, libraryName);
-
-        if (resource != null && libraryName != null && libraryName.equalsIgnoreCase(Constants.LIBRARY)) {
-            return new PrimeResource(resource);
-        }
-        else {
-            return resource;
-        }
+        return wrapResource(resource, libraryName);
     }
 
     @Override
     public Resource createResource(String resourceName, String libraryName, String contentType) {
         Resource resource = super.createResource(resourceName, libraryName, contentType);
-
-        if (resource != null && libraryName != null && libraryName.equalsIgnoreCase(Constants.LIBRARY)) {
-            return new PrimeResource(resource);
-        }
-        else {
-            return resource;
-        }
+        return wrapResource(resource, libraryName);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


tandraschko:
not sure as it will also catch premium layouts now?

melloware:
But don't we want it to ?  Here is the issue "pimrfaces-omega" and @mertsincan has made fixes between 7.0 and 8,0 but users upgrading to PF 8.0 won't get those fixes because Chrome has cached it.

We just got burned by this at my client because we completely upgraded our custom in house theme and when we deploy users are still seeing the old theme because its cached.

tandraschko:
i talk about layouts, not themes in PF

melloware:
Oh my bad the Layouts.  Yeah good question i submitted this through PRO account for @mertsincan to review.

tandraschko:
ok, should work
its called "babylon-layout" e.g.

melloware:
I decided to remove the `equalsIgnoreCase` and just to `libraryName.toLowerCase().startsWith(` which will safely catch what we want.

mertsincan:
Thanks a lot for the PR, @melloware ;)

mertsincan:
For now, I have reverted the changes. After discussing this with Çağatay, I will return. We have some special Js codes to change the theme in the configurator menu of all our layouts, and all of them may require a change. Exp;
```
changeComponentsTheme: function(theme, themeParam) {
        var library = 'primefaces-pandora';
        var linkElement = $('link[href*="theme.css"]');
        var href = linkElement.attr('href');
        var index = href.indexOf(library) + 1;
        var currentTheme = href.substring(index + library.length);

        if (currentTheme !== theme) {
            this.replaceLink(linkElement, href.replace(currentTheme, theme), themeParam);
        }
    },
```

melloware:
OK thanks for investigating.  We have a large rollout to thousands of customers and I can't really tell them all to press CTRL+F5 in the browser. :-)

tandraschko:
yeah
its a problem as you also have "primefaces-pandora", which is not part of PF core and would share the same version number then

melloware:
But wouldn't pandora if it has its own resource handler just do this...
```
<link type="text/css" rel="stylesheet" 
         href="/primefaces-test/javax.faces.resource/theme.css.xhtml?ln=primefaces-omega&amp;v=8.0.2&amp;v=3.1"" />
```
Where it would have both PF Number and Pandora number in the cache buster part of the URL.  That would be harmless I would think?

tandraschko:
yeah, i think it would be harmless

cagataycivici:
Yes, we should add version also changeTheme should better be in core so we can reuse it at templates and also users may as well if needed, the method in core should check for library, version, create a new url and load it, that way we don't need to maintain the code at templates one by one.

tandraschko:
@mertsincan @melloware @cagataycivici we have it in core. Check core.js

Can someone compare and check if we need changes in PF core, so that layouts can reuse it?

melloware:
I just tested changeTheme from Core.js and here is the before and after.
```
/showcase/javax.faces.resource/theme.css.xhtml?ln=primefaces-nova-light&v=9.0-SNAPSHOT
```
Calling `changeTheme('babylon');` results in:
```
/showcase/javax.faces.resource/theme.css.xhtml?ln=primefaces-babylon&v=9.0-SNAPSHOT
```

However looking at the Pandora source code it goes through an awful lot of trouble to replace the current theme link than the current core.js does...

```javascript
replaceLink: function(linkElement, href, themeParam) {
        PrimeFaces.ajax.RESOURCE = 'javax.faces.Resource';
        
        var isIE = this.isIE();

        if (isIE) {
            linkElement.attr('href', href);
        }
        else {
            this.disableThemeSelection();

            var $this = this,
            cloneLinkElement = linkElement.clone(false);

            cloneLinkElement.attr('href', href);
            linkElement.after(cloneLinkElement);
            
            cloneLinkElement.off('load').on('load', function() {
                linkElement.remove();
                
                if (themeParam) {
                    themeParam.callback(themeParam.type);
                }
                else {
                    $this.enableThemeSelection();
                }
            });
        }
    }
```

tandraschko:
but the result is the same? So i think we merge you PR and primetek can use Core changeTheme

melloware:
I mean it looks like both of their goals is to replace the theme.css I am just not sure why in Pandora it goes to all the trouble to delete the old link and add the new link.  I assume they are doing it on purpose for some reason I also think the "IE" specific check is strange.

i think @mertsincan or @cagataycivici would have to explain why they are going to all that trouble is it to solve a specific problem?

melloware:
Also I have verified from looking at Pandora that this change would break the Pandora `changeComponentsTheme` because they are not accounting for the `&` like we are they need to add 1 line...
```javascript
var plainURL = href.split('&')[0];
```

So new method in Pandora would be:
```javascript
changeComponentsTheme: function(theme, themeParam) {
        var library = 'primefaces-pandora';
        var linkElement = $('link[href*="theme.css"]');
        var href = linkElement.attr('href');
        var plainURL = href.split('&')[0];
        var index = plainURL.indexOf(library) + 1;
        var currentTheme = plainURL.substring(index + library.length);

        if (currentTheme !== theme) {
            this.replaceLink(linkElement, href.replace(currentTheme, theme), themeParam);
        }
    }
```

PR #6096, opened 2020-07-01
Fix #5808: accept array as value of AutoComplete

### LINKED ISSUE #5808: AutoComplete: accept array as a value (in addition to List)
## 1) Environment
- PrimeFaces version: 8.0

## 2) Expected behavior
- Accept `String[]` as a valid type for a value of the component

## 3) Actual behavior
throws `java.lang.ClassCastException: class java.util.ArrayList cannot be cast to class [Ljava.lang.String; (java.util.ArrayList and [Ljava.lang.String; are in module java.base of loader 'bootstrap')`

Right now the following components accept an array as a valid type:
- SelectCheckboxMenu
- SelectManyCheckbox

But `AutoComplete` accept only `List`.

It has something about converting `List` to `array` in `AutoCompleteRenderer:encodeMultipleMarkup`, but I don't know how to use it.
Also, the result is stored as an `array`  in `decodeMultiple` but then converted to `List` in `validate` method.

To solve the issue I extended `AutoComplete` class with the following methods:
```java
private boolean isInputValueArray = false;

@Override
public Object getValue()
{
    Object value = super.getValue();

    if (value != null && value.getClass().isArray())
    {
        isInputValueArray = true;
        Object[] array = (Object[]) value;
        return Arrays.asList(array);
    }

    return value;
}

@Override
public void setValue(Object value)
{
    if (isInputValueArray && value instanceof List)
    {
        value = ((List<String>) value).toArray(new String[0]);
    }

    super.setValue(value);
}
```

I can create a PR, but I think there is a better way to do that. If you can help me, I will implement it.

astappiev:
Solves #5808

Should be tested before merge.

To answer a question asked in original commit: the `autocomplete` property added to `AutoCompleteBase` is used in `AutoCompleteRenderer`, previously was inherited from HtmlInputText.

### DIFF
--- a/src/main/java/org/primefaces/component/autocomplete/AutoCompleteBase.java
+++ b/src/main/java/org/primefaces/component/autocomplete/AutoCompleteBase.java
@@ -23,13 +23,14 @@
  */
 package org.primefaces.component.autocomplete;
 
-import javax.faces.component.html.HtmlInputText;
+import javax.faces.component.html.HtmlSelectManyCheckbox;
+import javax.faces.context.FacesContext;
 
 import org.primefaces.component.api.InputHolder;
 import org.primefaces.component.api.MixedClientBehaviorHolder;
 import org.primefaces.component.api.Widget;
 
-public abstract class AutoCompleteBase extends HtmlInputText implements Widget, InputHolder, MixedClientBehaviorHolder {
+public abstract class AutoCompleteBase extends HtmlSelectManyCheckbox implements Widget, InputHolder, MixedClientBehaviorHolder {
 
     public static final String COMPONENT_FAMILY = "org.primefaces.component";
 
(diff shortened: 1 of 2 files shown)


melloware (approved):
This looks good to me.  Will let others get some eyes on it.

tandraschko:
-0.25 to make such a big inheritance change for such a improvement.
It would be ok for me If we would have around 5-10 Integration tests First.

melloware:
@tandraschko This now has conflicts and I agree this is a large inheritance change for little value.  Plus we are now extending our new AbstractPrimeHtmlInputText.  I vote for closing this ticket and PR.  Thoughts @Rapster @tandraschko

astappiev:
Ok. We have a working workaround for now, if I whould like to remove it again, I will open a new PR with probably better solution.
Thank you both for participating.

melloware:
OK excellent yes please feel free to provide a new PR in the future!

PR #6099, opened 2020-07-02
Fix #156: Clock allow custom time value

### LINKED ISSUE #156: Clock: Value binding custom time
I like <p:clock> to bind value to server side variable, not necessarily system clock. The clock time can be defined by application. The use case is application demo, the clock can be sometime in the future, for example, 02/15/2050.

melloware:


tandraschko:
+1 if well tested

PR #6101, opened 2020-07-02
Fix #1326: GMap add double click events

### LINKED ISSUE #1326: GMap: Add Double Click Ajax Listener
Please add a support for double click event listener on a polygon on gmap

melloware:


tandraschko:
+1 if well tested

PR #6105, opened 2020-07-03
Fix #5794: Spacer use 'src' instead of CSS

### LINKED ISSUE #5794: Performance: Replace dot_clear.gif with Base64 URL

melloware:


garcia-jj (commented):
Great. It works. Thank you @melloware ;)

Issue #376, opened 2015-05-29
Schedule: server side logic should update event model for eventMove & eventResize


mores:
( Original https://code.google.com/p/primefaces/issues/detail?id=6358 )

Some events in a schedule should be read only - users should be prevented from removing or modifying certain events.

cagataycivici:
There should better be an attribute like updateModelOnEvents="true|false". Where default is true and in your case I think it should be false. Can you confirm?

mores:
There is an attribute called "editable" which prevents an entry to be modified.
In my case I need a little more control, one of the business rules is that an event must be between 9 am and 5 pm. In this case a user should be allowed to move/resize between those values - if they try to resize beyond - it should be denied.

melloware:
See: https://github.com/primefaces/primefaces/pull/6107#issuecomment-654354234

PR #6107, opened 2020-07-04
Fix #6117, #5776, #74: Update FullCalendar to v5.1.0

### LINKED ISSUE #74: Schedule: Determine if an event was moved into allDay
I ran into a problem using the Schedule compoenent. I use a editable schedule and created a listener for the eventMove ajax-event of the schedule.
The problem is that I cannot determine if an event was moved into the allDay-Slot. Looking into the PrimeFaces sources I found out that the ScheduleEntryMoveEvent isn't filled with this information at all.

So, as far as I can see there's no way to find out if an event was moved into the allDay-Slot or vice-versa. 

I'm using PrimeFaces 3.3.1 on Mojarra 2.1.7

Here is a short example of what my schedule looks like:

``` xml
<p:schedule id="scheduleWidget"
  initialDate="#{scheduleController.selectedDate}"
  value="#{scheduleController.model}"
  draggable="true"
  resizable="true"
  slotMinutes="15"/>
  <p:ajax
    event="eventMove"
    listener="#{scheduleController.onEventMove}" />
</p:schedule>
```

In the onEventMove function I cannot determine if the event was moved into allDay.

### LINKED ISSUE #5776: Schedule: Upgrade FullCalendar (4.x --> 5.x)
Something we may look into later this year. (Target PrimeFaces 9.0)

FullCalendar 5.x is currently at beta-state. Should go final in may 2020:
https://fullcalendar.io/blog/2020/04/v5-beta-released

### LINKED ISSUE #6117: Schedule : resizable attribute not being honored
**Describe the defect**
The resizable attribute for p:schedule is not being honored in primefaces-8.0.jar

**Reproducer**
use the community download version of PF 8.0 from https://www.primefaces.org/downloads 

**Environment:**
 - PF Version: _8.0_
 - Affected browsers: Chrome

**To Reproduce**

The resizable attribute for p:schedule is not being honored in primefaces-8.0.jar 

 <p:schedule id="myschedule" resizable="false" ... > .... </p:schedule> will still allow me to resize individual entry items.

**Expected behavior**
The resizable attribute for p:schedule should be honored in primefaces-8.0.jar

blutorange:
This updates FullCalendar to version 5.1.0, see #5776 

## For the release notes:

* FullCalendar uses a different DOM structure with potentially new or changed style classes, check your custom CSS if you have any.
* If you are using a custom `extender` function, check your JavaScript code as many FullCalendar settings have been renamed or changed (see the FullCalendar upgrade guide: https://fullcalendar.io/docs/upgrading-from-v4)
* The event interface contains two new methods `ScheduleEvent#isResizable` and `ScheduleEvent#isDraggable` that should be used instead of `ScheduleEvent#isEditable`. Likewise, the builder has `DefaultScheduleEvent.Builder#resizable` and `DefaultScheduleEvent.Builder#draggable`.

## Remarks

* The `Schedule` component options `resizable` and `draggable` never had any effect because they were always overwritten by the `isEditable()` property on the `ScheduleEvent`. Also, `editable` is just a shortcut for `resizable` and `draggable` - but why can you set these individually on the component but not on the event? I changed it so that there is a `isResizable()` and `isDraggable()` on `ScheduleEvent` that is of type `Boolean` (null meaning that the default of the component is used). `isEditable()` is now deprecated. I also added two method `resizable` and `draggable` to the `DefaultScheduleEvent.Builder`.

* Some themes include custom CSS for FullCalendar. For the most part, everything still looks well. There's one minor issue - the colors in the list view are wrong e.g. for the nova theme. This repo only contains the minified CSS for the themes, not sure where the original sources are. If anybody could update the themes, that would be great.

image

(^ The background color of the rows should be grey - also, on hover, the background turns grey and is almost unrecognizable.)


blutorange:
The Travis build fails because of the dependency-check plugin, which attempts to retrieve the component report from the Sonatype OSS Index:

```sh
curl https://ossindex.sonatype.org/api/v3/component-report \
  -X POST \
  -H "Content-Type: application/vnd.ossindex.component-report-request.v1+json" \
  -H "Accept: application/vnd.ossindex.component-report.v1+json" \
  -d '{"coordinates":["pkg:maven/com.sun.istack/istack-commons-runtime@3.0.7","pkg:maven/net.glxn/qrgen@1.4","pkg:maven/jakarta.activation/jakarta.activation-api@1.2.2","pkg:maven/com.google.guava/listenablefuture@9999.0-empty-to-avoid-conflict-with-guava","pkg:maven/com.google.errorprone/error_prone_annotations@2.2.0","pkg:maven/org.glassfish/jakarta.faces@2.3.14","pkg:maven/jakarta.el/jakarta.el-api@3.0.3","pkg:maven/org.codehaus.mojo/animal-sniffer-annotations@1.17","pkg:maven/org.ehcache.modules/ehcache-xml@3.7.1","pkg:maven/org.ehcache/sizeof@0.3.0","pkg:maven/com.google.code.findbugs/jsr305@3.0.2","pkg:maven/net.sf.ehcache/sizeof-agent@1.0.1","pkg:maven/org.apache.poi/poi-ooxml-schemas@4.1.1","pkg:maven/jakarta.validation/jakarta.validation-api@2.0.2","pkg:maven/commons-io/commons-io@2.7","pkg:maven/org.snakeyaml/snakeyaml-engine@1.0","pkg:maven/com.github.albfernandez/juniversalchardet@2.3.0","pkg:maven/org.apache.xmlbeans/xmlbeans@3.1.0","pkg:maven/javax.javaee.api/javaee.api-compact@3.0-SNAPSHOT","pkg:maven/com.google.zxing/core@3.0.0","pkg:maven/org.checkerframework/checker-qual@2.5.2","pkg:maven/com.google.j2objc/j2objc-annotations@1.1","pkg:maven/com.sun.xml.fastinfoset/FastInfoset@1.2.15","pkg:maven/com.hazelcast/hazelcast@4.0.1","pkg:maven/org.jdom/jdom2@2.0.6","pkg:maven/com.googlecode.owasp-java-html-sanitizer/owasp-java-html-sanitizer@20200615.1","pkg:maven/org.jvnet.staxex/stax-ex@1.8","pkg:maven/com.rometools/rome@1.13.0","pkg:maven/org.apache.commons/commons-compress@1.19","pkg:maven/org.slf4j/slf4j-api@1.7.16","pkg:maven/org.glassfish.jaxb/txw2@2.3.1","pkg:maven/javax.activation/javax.activation-api@1.2.0","pkg:maven/javax.xml.bind/jaxb-api@2.3.1","pkg:maven/jakarta.xml.bind/jakarta.xml.bind-api@2.3.3","pkg:maven/com.google.zxing/javase@3.0.0","pkg:maven/org.ehcache.modules/ehcache-api@3.7.1","pkg:maven/org.terracotta/statistics@2.1","pkg:maven/org.owasp.encoder/encoder@1.2.1","pkg:maven/com.fasterxml.jackson.core/jackson-core@2.9.7","pkg:maven/javax/javaee-api@6.0","pkg:maven/commons-fileupload/commons-fileupload@1.4","pkg:maven/org.glassfish/javax.servlet@3.0","pkg:maven/org.apache.commons/commons-math3@3.6.1","pkg:maven/com.github.virtuald/curvesapi@1.06","pkg:maven/org.ehcache.modules/ehcache-impl@3.7.1","pkg:maven/org.terracotta/offheap-store@2.4.3","pkg:maven/org.apache.poi/poi-ooxml@4.1.1","pkg:maven/org.apache.commons/commons-collections4@4.4","pkg:maven/org.ehcache.modules/ehcache-core@3.7.1","pkg:maven/com.github.librepdf/openpdf@1.3.14","pkg:maven/org.json/json@20180813","pkg:maven/com.google.guava/failureaccess@1.0.1","pkg:maven/com.rometools/rome-utils@1.13.0","pkg:maven/net.sf.ehcache/ehcache@2.7.4","pkg:maven/org.apache.poi/poi@4.1.1","pkg:maven/org.glassfish.jaxb/jaxb-runtime@2.3.1","pkg:maven/org.ehcache/ehcache@3.7.1","pkg:maven/net.sf.barcode4j/barcode4j-light@2.0","pkg:maven/org.ehcache.modules/ehcache-107@3.7.1","pkg:maven/com.google.guava/guava@27.1-jre","pkg:maven/commons-codec/commons-codec@1.13"]}'
  ```

This request results in a 500 error, which I can reproduce here... https://ossindex.sonatype.org/rest#/Component%20vulnerability%20reports/post

image

File: src/main/java/org/primefaces/model/DefaultScheduleEvent.java [hunk 1]
```diff
@@ -183,8 +189,21 @@ public void setOverlapAllowed(boolean overlapAllowed) {
...
+     * {@link #setResizable(Boolean)}.
+     */
     public void setEditable(boolean editable) {
-        this.editable = editable;
+        this.draggable = editable;
+        this.resizable = editable;
+    }
+
+    public void setDraggable(Boolean draggable) {
+        this.draggable = draggable;
```
melloware, line 208: Should these be lower "b" booleans?

docs/9_0/components/schedule.md
```diff
@@ -54,8 +54,8 @@ showWeekNumbers | false | Boolean | Display week numbers in schedule.
...
-weekNumberCalculator | null | String | Client side function to use in custom weekNumberCalculation.
-nextDayThreshold | 09:00:00 | String | When an event's end time spans into another day, the minimum time it must be in order for it to render as if it were on that day. Default is 09:00:00.
+weekNumberCalculator | null | String | Client side function to use in custom `weekNumberCalculation`. This must be a JavaScript expression that results in an integer week number. You can use the local variable `date`, which points to the `Date` object for which the week number is to be determined. To illustrate, if you were to specify `date.getFullYear()`, the year would be shown instead of the actual week number.  
+nextDayThreshold | 09:00:00 | String | When an event's end time spans into another day, the minimum time it must be in order for it to render as if it were on that day. Default is `09:00:00`.
 slotEventOverlap | true | Boolean | If true contemporary events will be rendered one overlapping the other, else they will be rendered side by side.
```
melloware, line 59: Don't forget to update /docs/9_0/core/thirdpartylibs.md

melloware (changes requested):

melloware:
@mertsincan controls the Nova Light theme but I think I can help figure out what changes need to be made and send them.

src/main/java/org/primefaces/model/DefaultScheduleEvent.java
blutorange: I am actually using Boolean objects on purpose here, see the docs on DefaultScheduleEvent.Builder#draggable: There are actually three possible scenarios: (a) draggable = false, which makes the event not draggable,overriding the default that was set on the Schedule component globally, (b) draggable = true, which makes the event always draggable, and (c) draggable = null (the default), which uses the default that was set on the schedule component.

I added some JavaDocs to DefaultScheduleEvent.Builder#draggable, but I guess I could add docs to this method, too.

melloware: Cool as long as it was on purpose just mark this issue resolved.

docs/9_0/components/schedule.md
blutorange: Oh yeah, thanks for remining me. I updated that file with the current 5.1.0 version.

melloware:
I wonder if you update in `pom.xml` the OWASP Dependency Check plugin to the latest 5.3.2???

blutorange:
Yeah, I saw it wasn't the latest version. Though since I'm getting a 500 server error even via the official web interface for oss-index, that doesn't seem to be the cause. When I build locally, I keep getting the same error even with 5.3.2. May just be a temporary issue with oss-index.

Anyways, I updated the dependency so we're on the latest version.

melloware (approved):

christophs78:
We should add the editable vs draggable- und resizeable-thing to the migration guide.
And maybe update FC to the latest 5.x some time before PF 9 goes into RC.
Overall excellent work!

blutorange:
Added it to the migration guide: https://github.com/primefaces/primefaces/wiki/Migration-Guide/_compare/c4d552a87563cf42cc019652e485be00a4d506fb...c04263ed303fa12d360242d1e07b7d2b7286180a

And yes, since they seem to be busy fixing lots of minor bugs, we should definitely use the latest available FullCalendar version.

blutorange:
The oss-index seems to be up and running again, if you restart the Travis build it should succeed now.

melloware:
I don’t have permission to do it but you can with...

`git commit --allow-empty -m "Trigger"`

This empty commit will get squashed anyway with squash and merge

melloware:
@blutorange want to look at this Schedule one too while you have this PR open?  https://github.com/primefaces/primefaces/issues/74  I am hoping its relatively easy to fix.

melloware:
@blutorange Sorry again, but i think your latest PR might fix this issue: https://github.com/primefaces/primefaces/issues/376 ?? Can you confirm?

blutorange:
Regarding #376, as far as I can tell, there's a way to do what they want...
* Original issue description: It's kinda unclear what their issue is exactly... if an event cannot be edited, no data is sent to the server. What didn't work before is that setting a default (on the schedule widget for all events) didn't work - as it was always overwritten by whether the individual event's editable status.  
* [Third comment](https://github.com/primefaces/primefaces/issues/376#issuecomment-107181010) : Fine-tuned control over whether an event can be dragged to a certain time slot is possible via the extender option (https://fullcalendar.io/docs/eventAllow, there should also be an example in the [showcase for this](https://github.com/primefaces/primefaces-showcase/blob/master/src/main/webapp/ui/data/schedule/extender.xhtml))
* I suppose you could also change an event back to its original time on the server in response to the `eventMove` event - but that sounds like bad UI as it would make the event "jump" back when the AJAX request is done...

melloware:
Awesome I just linked your comment to #376 and closed it.  Its been 5 years so that should be good enough.

blutorange:
Regarding #74:

* I just checked the showcase, FullCalendar sends us the info of whether the event is now "allDay" or not, but we aren't sending that info to the server.
* Adding the following to [1-schedule.js#L149](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/schedule/1-schedule.js#L149) and updating the server code in `Schedule.java` should do the trick, it should then be possible to check the event's all day status via `ScheduleEntryMoveEvent#getScheduleEvent#isAllDay`:

```javascript
                        {name: $this.id + '_allDay', value: (eventDropInfo.event.allDay)}
```

I can go ahead and submit a PR, though I guess merging this first would help.

melloware:
@blutorange Just add that fix to this PR and I can then link this PR with #74.

blutorange:
Should work like this now:

image

image

```java
	public void onEventMove(ScheduleEntryMoveEvent event) {
		FacesMessage message = new FacesMessage(FacesMessage.SEVERITY_INFO, "Event moved", "Delta:" + event.getDeltaAsDuration());
		FacesMessage message2 = new FacesMessage(FacesMessage.SEVERITY_INFO, "Event moved", "All day:" + event.getScheduleEvent().isAllDay());
	
		addMessage(message);
		addMessage(message2);
	}
```

PR #6110, opened 2020-07-04
Docu: Prime Client Window

christophs78:
It´s very basic at this point in time. We may add more over time.

### DIFF
--- a/docs/9_0/core/clientwindow.md
+++ b/docs/9_0/core/clientwindow.md
@@ -1,0 +1,42 @@
+# Client Window
+
+Prime Client Window provides a improved version of JSF´s (version 2.2 and up ) CLIENT_WINDOW_MODE.
+This allows for separating state not only per session but also on a per client-window-base.
+(Or call it per-browsesr-tab-base.)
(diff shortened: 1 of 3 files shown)


tandraschko:
i would change the docu in multiviewstate.md a bit.
Currently it reads like its only compatible with PF ClientWindow
but its also compatible with the default JSF one or DeltaSpike.

christophs78:
> 
> 
> i would change the docu in multiviewstate.md a bit.
> Currently it reads like its only compatible with PF ClientWindow
> but its also compatible with the default JSF one or DeltaSpike.

Good point!

tandraschko:
I think we should also point out (at least a bit) whats the difference compared to the plain JSF one
- JSF doesnt append the url param on initial redirect, so f5 creates a new windowId
- JSF doesnt validate if the tab was initialy openend with another windowId
- it uses sessionStorage for save and validate the windowId
- the worklfow is the same as deltaspike lazy mode

tandraschko:
Thanks :)

PR #6113, opened 2020-07-06
Fix #118: Password add show/hide client side events

### LINKED ISSUE #118: Password: add event attribute
Event defines which client side event to display the password feedback, default is focus.

melloware:


tandraschko:
+1 if well tested

melloware:
Modeled it exactly after OverlayPanel and ToolTip do it.  Tested locally in IE, Firefox, and Chrome.

PR #6114, opened 2020-07-06
Fix #5912: DatePicker incorrect behavior with keyboard entry

### LINKED ISSUE #5912: DatePicker: with pattern has confusing behaviour when editing
p:datePicker in PrimeFaces-8.0 (and earlier) seem to have a confusing behavior when editing the input field. 

In PrimeFaces Showcase, p:datePicker with id="german" (that has pattern="dd.MM.yyy" and locale="de") : enter value 01.01.2020, click between 0 and 1 either on day or month and press Backspace (like user wants to change 01 to 10). -> the cursor jumps to the end of input value and nothing gets deleted. This is at least confusing and can cause invalid values to be entered if user does not check carefully what happens on page. 

This p:datePicker behavior could probably be categorized as a bug because p:calendar behaves more logically with same input sequence. 

This might be related to handling of locale as kukeltje on PrimeFaces forum indicated that it seems to work correctly with some locales.

melloware:
Only try and correct the keyboard input if the value is the correct length it is supposed to be.

@christophs78 please review

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
+++ b/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
@@ -1727,9 +1727,11 @@
             var rawValue = event.target.value;
 
             try {
-                var value = this.parseValueFromString(rawValue);
-                this.updateModel(event, value);
-                this.updateViewDate(event, value.length ? value[0] : value);
+                if (rawValue.length === this.options.defaultDate.length) {
+                    var value = this.parseValueFromString(rawValue);
+                    this.updateModel(event, value);
+                    this.updateViewDate(event, value.length ? value[0] : value);
+                }
             }
             catch (err) {
                 if (!this.options.mask) {


christophs78:
I´ll try to play around with your fix on weekend. Currently busy on my day jobs and kids back at home due to closed schools and kindergartens due to increased covid-cases-count.

christophs78:
I took some evening-time for a first look....

This PR breaks synchronisation with the DatePicker-widget for not full-length inputs.
image

Time-ressources aside IMO it would make sense to fix the root issue instead of implementing a quick-fix with other issues. Maybe someone of us can find time within the next weeks or month to look in-depth at it.

melloware:
Yes it does cause that but by synchronizing that is the exact issue.  Because the date is not parsed into a valid date so to me the value should NOT be synchronized until its a correct date no?   I think that is why PrimeNG does not have this issue they aren't as aggressively synchronizing as PF is...

christophs78:
Maybe there is some way to split widget-synchronisation and reformating. Or....

christophs78:
PrimeNG does the widget-synchronisation without reformating. (PrimeNG only reformats when the input loses focus.)

PR #6116, opened 2020-07-06
Fix #1597: jqPlot date I8LN

### LINKED ISSUE #1597: JQPlot Charts: Problem with IL8N
Charts with xAxis as dateAxis does't localize days and months. E.g. the date Jan 5, 2016 is always showed in English.

In Primefaces 5.3.11 charts i18 was fine.

Thanks
Luca

melloware:
Based on this Stack overflow post: https://stackoverflow.com/questions/11611324/jqplot-date-axis-renderer-internationalization

Now when I set locale="it" I see the correct dates in Italian:
image

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/charts/1-charts.js
+++ b/src/main/resources/META-INF/resources/primefaces/charts/1-charts.js
@@ -408,7 +408,7 @@ PrimeFaces.widget.Chart = PrimeFaces.widget.DeferredWidget.extend({
         }
 
         // #1597 must set locale on HTML element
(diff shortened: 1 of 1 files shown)


melloware:
@tandraschko I am wondering if it core.js when we get the settings.locale on init if we should update `<html lang=XX` like i have to do in this PR just to be more correct.
```javascript
 document.documentElement.setAttribute('lang', PrimeFaces.settings.locale);
```

For example the showcase has `<html lang="en"` even if you have locale set it `it` in your faces-config.xml.  Thoughts?

PR #6124, opened 2020-07-07
Fix #6123: MoveScriptsToBottom string perf

### LINKED ISSUE #6123: Perf: MoveScriptsToBottom
**Description**
Currently MoveScriptsToBottom uses string concatenation and `String.replace` function calls to shrink the output script.  However each of these replace calls results in String concat and garbarge collection.

Perform the multiple replaces without String concat and garbage collection by using an efficient method using a String Builder and the least number of passes over the string to do replacement.  Perfect candidate is Commons Lang `StringUtils.replaceEachRepeatedly`.

**Additional context**
The performance of this method say over 1000 characters looking for 10 string replacements takes `0 seconds, 7 millis`

melloware:


tandraschko:
REPLACEMENT_MAP doesnt need a serialVersionUID, its never serialized

tandraschko:
AFAIR we also have a MapBuilder?

melloware:
Good suggestion switching to MapBuilder.

PR #6128, opened 2020-07-08
Fix #2434: Width corrections

### LINKED ISSUE #2434: CSS - Style Width should always be the total Width of a component
See this image + code:

widths

Especially CSS like sometimes is problematic. 

.ui-selectonemenu
       padding-right: 2em;

I think the better approach would be to calculate all the margins, paddings and width so that the component itself never grows bigger than the defined width. Else you'll end up to have to set different widths for each component.

melloware:
Screenshot after PR:
image

Total width is now 300px where before it was 300p + the dropdown arrow 16px.

Input Text had too much padding which pushed the width out past 300px.


tandraschko:
+1 if well tested (maybe also with a premium layout?)

melloware:
I will also ask @mertsincan to review

tandraschko:
can't we do it with pure css?

melloware:
That I am not an expert in it may be a @mertsincan question.

tandraschko:
why do you select the state ? .ui-selectonemenu-trigger.ui-state-default.ui-....

melloware:
You are right i just grabbed the CSS style when the page was rendered to make it exact since this only executes on init?  Let me try removing the state.

melloware:
Removed ui-state-default and its still working great. Thanks for the suggestion.

tandraschko:
same thing about the ui-corner-right - there should only be one trigger?

melloware:
You are correct. Removed that too and still working fine as you are right there will only ever be 1 trigger.

tandraschko:
@melloware couldnt we just solve this completely via CSS instead of JS code?

melloware:
So I think its possible to be done but feels really dangerous and here is why.  So when setting a SelectOne to 300px here is the HTML output
```xml
<div class="ui-selectonemenu ui-widget ui-state-default ui-corner-all" style="width:300px;">
        <input type="text" autocomplete="off" aria-autocomplete="list" aria-disabled="false">
        </div>
            <select tabindex="-1" autocomplete="off" aria-hidden="true"></select>
        </div>
        <label class="ui-selectonemenu-label ui-inputfield ui-corner-all">&nbsp;</label>
        <div class="ui-selectonemenu-trigger ui-state-default ui-corner-right">
            <span class="ui-icon ui-icon-triangle-1-s ui-c"></span>
        </div>
    </div>
</div>
```

The outermost `div` gets `width:300px` which is good.  The problem is the `label` has CSS width 100% (for other use cases of SelectOne and makes it 300PX.  Now the DropDownArrow is rendered for 16 more px.  Even if I set `max-width:300px` on the parent `div` it makes no difference.  And this is only when you delcare an actual width and not using the Widget property `autoWidth=true`.

So is it possible? probably.  Have I been able to figure it out? No.  Maybe a CSS expert could but this code was already doing it with JS so I just "fixed" the JS.

mertsincan:
Thanks a lot for the PR, @melloware. But please see my comment;
https://github.com/primefaces/primefaces/issues/2434#issuecomment-691460557

It works fine for me with the new themes. Could you please check it? Thanks a lot!

melloware:
Yep I responded th problem is with the built in Omega theme and it looks like all older JQuery ThemeRoller Themes.  So if you fix the built in Omega theme and we have already considered those themeroller themes as deprecated then I think we can close the ticktet and the PR.

PR #6130, opened 2020-07-08
Fix #3567: Trigger client row click on ENTER/SPACE

### LINKED ISSUE #3567: DataTable: onRowClick isn't called when row click is triggered via keyboard
DataTable with JS code for attribute "onRowClick": if a row is clicked via mouse the JS code is executed. If a row is selected via keyboard (using Up/Down keys to navigate to a row and pressingn SPACE or ENTER to select that row) the "onRowClick" code isn't executed.

Reason: In datatable.js function bindRowClick() treats a mouse click only and "...on('keydown', ..." in bindSelectionKeyEvents() doesn't treat $this.cfg.onRowClick.

Furthermore the PF user guide doesn't contain description of the attribute "onRowClick".

## 1) Environment
- PrimeFaces version: 6.2.2
- Does it work on the newest released PrimeFaces version? Version? Yes
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) Yes
- Application server + version: irrelevant (Tomcat 8)
- Affected browsers: irrelevant (Firefox and Chrome)

## 2) Expected behavior
the JS code specified for attribute "onRowClick" should be executed if user selects a row via keyboard (SPACE or ENTER)

## 3) Actual behavior
the JS code is only executed when user selects a row via a mouse click

## 4) Steps to reproduce
see above 

## 5) Sample XHTML
<p:dataTable ... onRowClick="myFunction(row)">
...

## 6) Sample bean
irrelevant

melloware:


tandraschko:
Do we have a correct docu for "Furthermore the PF user guide doesn't contain description of the attribute "onRowClick"." now?

melloware:
I am seeing in 8.0 and 9.0 docs the onRowClick is documented? https://primefaces.github.io/primefaces/8_0/#/components/datatable?id=datatable

I think this ticket is so old they were referring to the old PDF docs.

tandraschko:
ok thanks!

PR #6134, opened 2020-07-09
Fix #3060: Wizard effect/effectDuration attributes for step transition

### LINKED ISSUE #3060: Wizard: Support animation for changing steps
Currently Wizard component simply [changes the content html when response is retrieved](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/wizard/wizard.js#L83):
```js
    widget: $this,
    handle: function(content) {
        this.content.html(content);
    }
});
```

While this is working fine, the change is immediate, but sometimes UX requires smooth transitions between the wizard steps. For example, simple fade-in animation would greatly improve user experience.
```js
  widget: $this,
  handle: function (content) {
    this.content.fadeTo(0, 0);
    this.content.html(content);
    this.content.fadeIn();
  }
});
``` 
Height could be easily added into the animation to make transition even smoother.

It would be great if PF supported the animation for content change in Wizard. This could be even completely delegated to users if they need it:
```js
  widget: $this,
  handle: function (content) {
    if ($this.cfg.contentAnimation) {
      $this.cfg.contentAnimation(this.content, content);
    } else {
      this.content.html(content);
    }
  }
});

```

melloware:
Default is false so no change in migration guide.

### DIFF
--- a/docs/9_0/components/wizard.md
+++ b/docs/9_0/components/wizard.md
@@ -35,6 +35,7 @@ nextLabel | null | String | Label of next navigation button.
 backLabel | null | String | Label of back navigation button.
 widgetVar | null | String | Name of the client side widget
 updateModelOnPrev | false | Boolean | If yes, the model will be updated when the "Back" button is clicked. Default is false.
+animate | false | Boolean | When enabled an animation is used to transition wizard steps. Default if false.
 
 ## Ajax Behavior Events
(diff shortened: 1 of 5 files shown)


tandraschko:
+1 if well tested

tandraschko:
should we add a way to change animations?

melloware:
Good idea I will add an '`effect` attribute like other components have to pick the animation.

PR #6137, opened 2020-07-10
Fix #5638: DatePicker detect showTime/showSeconds from pattern

### LINKED ISSUE #5638: DatePicker: ShowTime/ShowSeconds based on pattern
There are some further issues/bugs with the p:datePicker component.

Issue 1) It should not require to explicitly set the attribute showTime in order to be usually for LocalDateTime. Without this attribute the validation does not pass. Event showTime="false" does not work.

Issue 2) When the attribute pattern is NOT fully applied in case of LocalDateTime. For instance, pattern="yyyy-MM-dd" still show the time although I do not want to show it as expressed by the pattern. In general, other patterns like pattern="yyyy-MM-dd HH" are "ignored".

Issue 3) It is not possible to set the seconds (and smaller time units) in case LocalDateTime.

## 1) Environment
- PrimeFaces version: 8.0 RC3
- Affected browsers: All

## 2) Expected behavior

Issue 1) When p:datePicker is used for a LocalDateTime it should not be required to set showTime=true". The pattern should define how the date(Time)Picker should look like.

Issue 2) When p:datePicker is used for a LocalDateTime with the "yyyy-MM-dd" it should be accepted and the time part set implicitly to 00:00:00.000. Accordingly, When p:datePicker is used for a LocalDateTime with the "yyyy-MM-dd HH" it should be accepted and the time part set implicitly to HH:00:00.000. of course, in this case, the time selector should only show the hours part.

Issue 3)  When p:datePicker is used for a LocalDateTime with the "yyyy-MM-dd HH:mm:ss", the time selector should have a seconds selector.

melloware:
@christophs78 @tuerker this builds off the other smart PR for detecting LocalDateTime or LocalTime this add the checking of pattern for `showTime` and `showSeconds`.  

I ran into this issue myself recently switching from Calendar to DatePicker.

### DIFF
--- a/src/main/java/org/primefaces/component/datepicker/DatePickerBase.java
+++ b/src/main/java/org/primefaces/component/datepicker/DatePickerBase.java
@@ -236,6 +236,10 @@ public abstract class DatePickerBase extends UICalendar implements Widget, Input
         return (Boolean) getStateHelper().eval(PropertyKeys.showSeconds, false);
     }
 
+    public Boolean isShowSecondsWithoutDefault() {
+        return (Boolean) getStateHelper().eval(PropertyKeys.showSeconds);
+    }
+
     public void setShowSeconds(boolean showSeconds) {
         getStateHelper().put(PropertyKeys.showSeconds, showSeconds);
     }
--- a/src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java
+++ b/src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java
@@ -55,20 +55,10 @@ public class DatePickerRenderer extends BaseCalendarRenderer {
             Class<?> type = datepicker.getTypeFromValueByValueExpression(context);
 
             if (type != null) {
-                if (LocalDateTime.class.isAssignableFrom(type)) {
-                    datepicker.setShowTime(true);
-                }
-                else {
-                    datepicker.setShowTime(false);
-                }
+                datepicker.setShowTime(LocalDateTime.class.isAssignableFrom(type));
             }
             else {
-                if (CalendarUtils.hasTime(pattern)) {
-                    datepicker.setShowTime(true);
-                }
-                else {
-                    datepicker.setShowTime(false);
-                }
+                datepicker.setShowTime(CalendarUtils.hasTime(pattern));
             }
         }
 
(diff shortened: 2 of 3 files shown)


christophs78:
IMO we should detect showSeconds only automatically when the user does not provide this attribute explicit.

Do something like
`if (datepicker.isShowSecondsWithoutDefault() == null) {`

Why? DatePicker was designed by PrimeTek in a way where pattern should only contain the date part. Everything related to time is not set via pattern, but via explizit attributes like showTime and showSeconds. (I personally don´t think this was a good decision.)

So it was fine to do something like
```
<p:outputLabel for="time5" value="Time (with seconds):" />
<p:datePicker id="time5" value="#{calendarJava8View.dateTime5}" showTime="true" showSeconds="true"/>
```
until now. We should not break this.

christophs78:
Maybe code should look like this:
```
@Override
protected void encodeMarkup(FacesContext context, UICalendar uicalendar, String value) throws IOException {
	DatePicker datepicker = (DatePicker) uicalendar;
	String pattern = datepicker.calculatePattern();

	if (datepicker.isShowTimeWithoutDefault() == null) {
		Class<?> type = datepicker.getTypeFromValueByValueExpression(context);

		if (type != null) {
			if (LocalDateTime.class.isAssignableFrom(type)) {
				datepicker.setShowTime(true);
			}
			else {
				datepicker.setShowTime(false);
			}
		}
		else {
			if (CalendarUtils.hasTime(pattern)) {
				datepicker.setShowTime(true);
			}
			else {
				datepicker.setShowTime(false);
			}
		}
	}

	if (datepicker.isTimeOnlyWithoutDefault() == null) {
		Class<?> type = datepicker.getTypeFromValueByValueExpression(context);

		if (type != null) {
			if (LocalTime.class.isAssignableFrom(type)) {
				datepicker.setTimeOnly(true);
			}
			else {
				datepicker.setTimeOnly(false);
			}
		}
	}

	if (datepicker.isShowSecondsWithoutDefault() == null) {
		if (pattern.contains("s")) {
			datepicker.setShowSeconds(true);
		}
		else {
			datepicker.setShowSeconds(false);
		}
	}
```

melloware:
Great suggestion.  PR updated.

melloware:
Also I simplified all the if else logic as well.

PR #6139, opened 2020-07-10
Fix #5912: DatePicker incorrect behavior with keyboard entry

### LINKED ISSUE #5912: DatePicker: with pattern has confusing behaviour when editing

christophs78:
- [x] My first steps to decouple the update of the input-field from the update of the model and DatePicker-widget. I think this already improves input-behaviour.
- [x] I´ll try to add some code which does the reformating on focus-lost as next step and add this to this PR.
- [x] And maybe we can add something to accept 2-digits-year-inpuz also for 4-digit-patterns.

As already said before people choose libraries like PrimeFaces because they expect this things to work. And now we are as a community on a point where we need to fix dozens of issues for DatePicker. PrimeTek released DatePicker in sub-par quality with 7.0. And now we as community need to do this work.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
+++ b/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
@@ -1704,6 +1704,9 @@
                 this.options.onBlur.call(this, event);
             }
 
+            //reformat input - is always OK? or only if it was modified?
+            this.inputfield.val(this.getValueToRender());
+
             this.inputfield.removeClass('ui-state-focus');
             this.container.removeClass('ui-inputwrapper-focus');
         },
--- a/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
+++ b/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
@@ -1,4 +1,4 @@
-/**
+or only if it was modified?/**
  * Prime DatePicker Widget
  */
 (function (factory) {
@@ -1709,7 +1709,6 @@
                 this.options.onBlur.call(this, event);
             }
 
-            //reformat input - is always OK? or only if it was modified?
             this.inputfield.val(this.getValueToRender());
 
             this.inputfield.removeClass('ui-state-focus');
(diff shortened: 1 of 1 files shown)


melloware:
I totally agree with you that we need to do whatever it takes to make this component work like a user expects it to work.

This is an excellent PR and improves the component behavior quite a bit.

src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -1704,6 +1709,9 @@
                 this.options.onBlur.call(this, event);
             }
 
+            //reformat input - is always OK? or only if it was modified?
```
melloware: I think its safe to remove this comment I think we should always format unless we find it to be a problem.

melloware (approved):

src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -2488,9 +2495,11 @@
             this._setInitOptionValues();
         },
 
-        updateModel: function (event, value) {
+        updateModel: function (event, value, updateInput) {
             this.value = (value === '' ? null : value);
-            this.inputfield.val(this.getValueToRender());
+            if (updateInput != false) {
```
melloware, line 2500: I think you can just say `if (updateInput)` and its truthiness will only enter the block if the value is `true` rather than checking for !false

melloware (changes requested):

src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
christophs78: No, not that easy. updateInput is only passed when i want to disable input-update. All other calls to this function are without this parameter.
So we may change this. But this requires all other calls to this function to pass true for the third parameter.

melloware (approved):

PR #6141, opened 2020-07-11
Fix #92: Tooltip my and at position control

### LINKED ISSUE #92: Tooltip: Make location customizable
Add _my_ and _at_ attributes to make tooltip location customizable.

melloware:


tandraschko:
+1 if well tested

PR #6143, opened 2020-07-12
Fix #6142: CSP issue with InputNumber events

### LINKED ISSUE #6142: InputNumber: valueChange event not fired if CSP context-param is set
When using p:inputNumber, the p:ajax default (valueChange) event does not fire on change of value when the CSP context-param is set... unless the MOVE_SCRIPTS_TO_BOTTOM context-param is _also_ set, in which case it works normally.

**Reproducer**
See attached. Ths sampe code is based on the InputNumber showcase example.
[tester_inputNumber_src.zip](https://github.com/primefaces/primefaces/files/4907592/tester_inputNumber_src.zip)

**Environment:**
 - PF Version: _8.0_
 - Mojarra 2.3
 - Affected browsers: _ALL_

**To Reproduce**
See reproducer. Set web.xml to include CSP and exclude Move_Scripts_to_Bottom. 
The sample app. is based on Showcase.

**Expected behavior**
The p:ajax events should fire normally regardless of whether CSP is set or not.

**Workaround**
Either don't use CSP context-param, or if you need to, also use Move_Scripts_to_Bottom context-param.

melloware:


tandraschko:
how does the namespace fix CSP?

melloware:
Good question.  So what is happening is this...
CSP:OFF MOVE_SCRIPTS:OFF = Works
CSP:ON MOVE_SCRIPTS:ON = Works
CSP:ON MOVE_SCRIPTS:OFF = Fails.

It is an ordering thing.  When the `PrimeFaces.csp.register` runs and registers the `change` event this line of code in the InputNumber init immediately wipes out ALL change events.
```javascript
this.input.prop('onchange', null).off('change').
```

So by namespacing it to `off('change.inputnumber')` it only removes that specific change event and not ALL change events.

melloware:
@tandraschko I think we actually have a more fundamental issue with CSP + InputNumber.  In the `wrapEvents` method it does this..
```javascript
        var originalOnchange = this.input.prop('onchange');
        this.input.prop('onchange', null).off('change').on('change', function (e) {

            var oldValue = $this.copyValueToHiddenInput();
            if (originalOnchange && originalOnchange.call(this, e) === false) {
                $this.setValueToHiddenInput(oldValue);
                return false;
            }
        });
```

However because CSP stripped out the `onchange` event and make it a `csp.register` call this client side JS doesn't ever find the original events `originalOnchange = this.input.prop('onchange');`.

So I need some advice on how you recommend I fix this so the code doesn't rely on the client side "on" events being in the DOM to wrap them?

melloware:
@tandraschko I was able to fix this and make sure it works with CSP and without CSP.  Basically if not using CSP look for the normal DOM events like 'onchange' and use those as the "original event".  However if using CSP we need to look at the Jquery "change" event already attached to the input and store it off and call its handler.  But we must call `off('change') to remove the original handler and call it manually.

Its all working now in all 3 scenarios!

PR #6172, opened 2020-07-22
Fix #6170: Locale default pattern should be 4 digit year

### LINKED ISSUE #6170: DatePicker: YearRange and default pattern
**Describe the defect**
Forum Post: https://forum.primefaces.org/viewtopic.php?f=3&t=62741

**Reproducer**
[pf-datepicker.zip](https://github.com/primefaces/primefaces/files/4959959/pf-datepicker.zip)

**Environment:**
 - PF Version: _8.0_
 - JSF + version: _ALL_
 - Affected browsers: _ALL_

**To Reproduce**
1. Try and pick a year before 2020 and it won't work.

**Expected behavior**
I should be able to pick a year between 1920 and 2020.  However if you manually set `pattern="dd/MM/yyyy"` it works.

**Example XHTML**
```html
<h:form id="frmTest">
            <p:datePicker id="dob" 
                          value="#{testView.dob}" 
                          inline="true" 
                          yearNavigator="true" 
                          monthNavigator="true" 
                          yearRange="1920:2020" 
                          required="true" 
                          requiredMessage="Required" 
                          maxdate="#{testView.maxDate}" 
                          mindate="#{testView.minDate}" />
</h:form>
```

**Example Bean**
```java
@Named
@ViewScoped
    
	@Past
    private LocalDate dob;
    private LocalDate minDate;
    private LocalDate maxDate;
    
    @PostConstruct  
    public void init() {
       minDate = LocalDate.of(1920, 1, 1);
	   maxDate = LocalDate.now();
    }

    public LocalDate getDob() {
        return dob;
    }

    public void setDob(LocalDate dob) {
        this.dob = dob;
    }

    public LocalDate getMinDate() {
        return minDate;
    }

    public void setMinDate(LocalDate minDate) {
        this.minDate = minDate;
    }

    public LocalDate getMaxDate() {
        return maxDate;
    }

    public void setMaxDate(LocalDate maxDate) {
        this.maxDate = maxDate;
    }    
}
```

melloware:
Converts `dd/MM/yy` into `dd/MM/yyyy` or wherever a 2 digit year is found in the locale default should be a 4 digit year.  This will prevent many problems including the one described in this ticket as a "smart default" if `pattern` is not specified.

### DIFF
--- a/src/main/java/org/primefaces/component/api/UICalendar.java
+++ b/src/main/java/org/primefaces/component/api/UICalendar.java
@@ -36,6 +36,7 @@ import javax.faces.component.html.HtmlInputText;
 import javax.faces.context.FacesContext;
 
 import org.primefaces.util.CalendarUtils;
+import org.primefaces.util.LangUtils;
 import org.primefaces.util.LocaleUtils;
 import org.primefaces.util.MessageFactory;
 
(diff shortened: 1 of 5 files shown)


melloware:
@christophs78 I know you are busy but would love your eyes on this PR.

christophs78:
IMO this makes sense.
(In my country (austria) we are used to use 4 digits year. And i think this applies to most of europe. Not totaly sure what users in other areas of the world expect, but when you as an american also prefer 4 digits year it´s a good sign.)

PR #6174, opened 2020-07-27
Fix #6173 - TreeTable: Filter CustomTreeNodes

### LINKED ISSUE #6173: TreeTable: Filter CustomTreeNodes
**Description**
Type of TreeNodes will be lost after filtering TreeTable. The problem is probably caused by 

`TreeTableRenderer.createNewNode(TreeNode node, TreeNode parent); `

This method will always return DefaultTreeNode or CheckboxTreeNode, even if the Type of parameter _node_ is CustomTreeNode. Because this method is called due filtering (TreeTableRenderer.filter(...)), the type of all nodes will be lost. 

**Environment:**
 - PF Version: 8.0
 - JSF:  Mojarra 2.3.9

**Expected behavior**
createNewNode should return a node with the same type as node in parameter, so that nodes in TreeTable are the same type after filtering as before. 

```java
CustomTreeNode node = new CustomTreeNode(...);
TreeNode newNode = createNewNode(node, null) ;
if(newNode instanceof CustomTreeNode){
   //correct
}else{
   //incorrect
}
```

**Possible solutions**
* handle filtering without creating new nodes
* copy node via reflection

Rapster:


melloware (approved):

melloware:
Looks like you have 1 check-style error:
```
[ERROR] src/main/java/org/primefaces/component/treetable/TreeTableRenderer.java:[1312,9] (blocks) RightCurly: '}' at column 9 should be alone on a line.
```

tandraschko:
this breaks all of our applications with TreeTables currently, need to have a deeper look here

Rapster:
Once you have more detail, feel free to create an issue

tandraschko:
puh, quite weird code here.
lets first talk about when and why createNewNode is called?
Currently its called in 2 places:
1) to "clone" the root node and set it to the value? (where is the original value gone then?)
2) in createFilteredNodeFromRowKeys to collect getFilteredRowKeys, which is unused and i dont know why we need a clone for this?

Rapster:
Do you see the same problem on showcase?

tandraschko:
Nope, it just happens when you use custom nodes with wrong constructor. But it worked fine before so i like to check why we have this copy/cloning

tandraschko:
i think i finally know how it works, we just need some better #createNode method, i will work on it

PR #6179, opened 2020-07-30
Fix #6178: DatePicker inline mode add ui-state-error style

### LINKED ISSUE #6178: DatePicker: inline=true add style class 'ui-state-error' to when validation has failed
This will allowing users to create a CSS rule for `body .ui-datepicker-inline.ui-state-error`, or even adding it to themes.

melloware:


jepsar (approved):
Your PR is looking good to me. Thank you, @melloware

PR #6185, opened 2020-07-31
Fix #5776: FullCalendar 5.3.0

### LINKED ISSUE #5776: Schedule: Upgrade FullCalendar (4.x --> 5.x)

melloware:
Tested with Showcase and everything is working properly.


melloware:
Made this a draft so I can keep updating FullCalendar versions up until we are read to go with 9.0

tandraschko:
I understand your point but i prefer to have only a few open PRs :D

melloware:
OK I marked it ready for review. Safe to merge?

tandraschko:
i'm not a user of the fullcalendar ;D better ask to review other schedule contributors

melloware:
@blutorange can you eyeball this before I merge.  I tested all the showcase examples with it and seems straightforward,

blutorange:
@melloware Looks good for the most part. The full calendar changelog for 6.3 only lists bug fixes, for 6.2 also mostly bug fixes and no breaking changes. I also checked out the showcase examples. A few minor things I noticed:

* showcase/ui/data/schedule/configureable.xhtml There are two empty buttons at the left header - which seems to be cause by the comma in the `leftHeaderTemplate` value `prev, next, today` (also the default according to the docs). When I remove the comma it looks good.

image
 
* The tooltip when hovering over an event seems to be a missing a background color?

image

*  showcase/ui/data/schedule/configureable.xhtml The option `displayEventEnd` either needs a fix or better documentation. When I set it to `false`, it shows the start and end time. According to the docs, it can also be set to something like `month:false,basicWeek:true`, but either that's not working or I don't know which keys to use.

image

melloware:
OK let me investigate those two issue.

melloware:
Fixed LeftheaderTemplate issue with https://github.com/primefaces/primefaces-showcase/pull/187

melloware:
As for displayEventEnd I agree with you. We are doing the right things acccording to the docs so possibly its a bug in FullCalendar.
https://fullcalendar.io/docs/displayEventEnd

PR #6195, opened 2020-08-04
Fix #6192 - SplitButton: Both action button and action item are executed

### LINKED ISSUE #6192: SplitButton: Both action button and action item are executed
**Describe the defect**
The core refactoring of @Rapster (#5009) introduced a bug to `p:splitbutton` and maybe to other components (did not check it).
Before the code changes the logic of [`super.decode()`](https://github.com/primefaces/primefaces/blob/master/src/main/java/org/primefaces/component/splitbutton/SplitButtonRenderer.java#L61) was inside an if block and the code below was inside an else-if block ([see here](https://github.com/primefaces/primefaces/blob/b4b33923cac78331d6a372f72784bb5b826b5bf7/src/main/java/org/primefaces/component/splitbutton/SplitButtonRenderer.java#L83)).

That means, that both action listeners will be executed. The one of the menu item in `super.decode()` and the one from the standalone if block. But only the one of the menu item should be executed if a menu item was clicked.

This seems to happen only to a splitbutton that uses the `model` attribute.

**Reproducer**
https://github.com/fanste/primefaces-test/tree/test-6192

**Environment:**
 - PF Version: 8.0
 - JSF + version: Mojarra 2.1.19
 - Affected browsers: ALL

**To Reproduce**
1. Go to 'index.jsf'
2. Click on 'ALL'
3. Only the ALL action listener is executed
4. Click on any menu item of the split button
3. Action listener of ALL and menu item is executed

**Expected behavior**
_Already explained in the defect description_

Rapster:


melloware (approved):
Looks good to me

PR #6197, opened 2020-08-06
Fix #5528: DatePicker time separator from Locale

### LINKED ISSUE #5528: DatePicker : Localization of Time
## 1) Environment
- PrimeFaces version: 8.0.RC2
- Does it work on the newest released PrimeFaces version? Not working on 7.0 either

## 2) Expected behavior

When a datepicker has the showTime option set to true, or the timeOnly option set to true, the time should be localised, according to the locale.

This is possible with the calendar component by setting a pattern, although for datepicker the pattern is for date only according to the documentation.

## 3) Actual behavior

The time is never localised, will always be displayed as HH:mm or HH:mm:ss if showSeconds is true.

## 4) Steps to reproduce

Create a datepicker, with localization options set.

## 5) Sample XHTML

<p:datePicker id="dp" locale="de" timeZone="#{localeBean.timeZone}" showTime="true" pattern="dd/MM/yyyy" />

## 6) Sample bean

import java.io.Serializable;
import java.util.TimeZone;
import javax.inject.Named;

@Named
@SessionScoped
public class LocaleBean implements Serializable {
	
	private TimeZone timeZone = TimeZone.getTimeZone("Europe/Copenhagen");
	
	public TimeZone getTimeZone() {
		return timeZone;
	}

	public void setTimeZone(TimeZone timeZone) {
		this.timeZone = timeZone;
	}
	
}

melloware:
Uses the Locale default time format to determine whether to use `:` or `.` as the time separator.

### DIFF
--- a/src/main/java/org/primefaces/component/datepicker/DatePickerBase.java
+++ b/src/main/java/org/primefaces/component/datepicker/DatePickerBase.java
@@ -23,7 +23,11 @@
  */
 package org.primefaces.component.datepicker;
 
+import java.time.chrono.IsoChronology;
+import java.time.format.DateTimeFormatterBuilder;
+import java.time.format.FormatStyle;
 import java.util.List;
+import java.util.Locale;
 
 import org.primefaces.component.api.InputHolder;
 import org.primefaces.component.api.MixedClientBehaviorHolder;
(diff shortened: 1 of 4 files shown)


christophs78:
Your PR looks good and i like the way you try to fix this issue.

But we should cross-check DatePickerTest.
[ERROR] Failures: 
[ERROR]   DatePickerTest.calculatePatternWithAmPm:1011 expected: <M/d/yyyy hh:mm a> but was: <M/d/yyyy hhnullmm a>
[ERROR]   DatePickerTest.calculatePatternWithSeconds:994 expected: <M/d/yyyy HH:mm:ss> but was: <M/d/yyyy HHnullmmnullss>
[ERROR]   DatePickerTest.calculatePatternWithSecondsAndAmPm:1003 expected: <M/d/yyyy hh:mm:ss a> but was: <M/d/yyyy hhnullmmnullss a>
[ERROR]   DatePickerTest.calculatePatternWithTime:986 expected: <M/d/yyyy HH:mm> but was: <M/d/yyyy HHnullmm>
[ERROR]   DatePickerTest.calculatePatternWithTimeRemove:1020 expected: <yyyy-MM-dd HH:mm> but was: <yyyy-MM-dd HHnullmm>
[ERROR] Errors: 
[ERROR]   DatePickerTest.convertToJava8DateTimeAPI_LocalDateTime:310 » IllegalArgument U...

But this may be only a mocking-issue. Maybe it helps to ensure getTimeSeparator calls the real method from test or something like this.

melloware:
OK fixed unit tests.  Letting it rebuild now.

PR #6199, opened 2020-08-06
Fix #1287: Terminal add AJAX support

### LINKED ISSUE #1287: Terminal - add async support
As a quick fix we can enable async by default for Terminal... 
Can't really think of an example where this is a bad idea. 

I will try to create a pull request these days if you guys think it's a welcomed improvement.

melloware:


tandraschko:
maybe just call the event "command"?

melloware:
Switched `processCommand` to just `command`

tandraschko:
+1

PR #6205, opened 2020-08-07
Fix #5817: Dialog CSS issues with height width

### LINKED ISSUE #5817: JS: Remediate all JQMIGRATE deprecations

melloware:


tandraschko:
can you explain it? Must we also fix other places again?

melloware:
Dialog has Height and Width as Java String instead of Integer.

So you can do "100" or "100px" or "100vh" or "100%".  I was adding 'px' on the end of `this.cfg.height` and width when it simply just needs to ensure its a string with `String(this.cfg.height)`

melloware:
I am going to do more testing today with dialogs of various flavors.

erickdeoliveiraleal:
yes it fixed the original issue I was having, but now my dialog which was in center is in left top position.

melloware:
Yep just fixed that please pull and try again.  I "over" fixed it.

PR #6213, opened 2020-08-10
Fix #6212: ChartJs animation options

### LINKED ISSUE #6212: ChartJs: Animation attributes definition
LineChartModel doesn't present a way to enable/disable chart animation. Animation comes by default, and I could not find a way to disable it.

melloware:
new global Animation option: https://www.chartjs.org/docs/latest/configuration/animations.html


coldhack (approved):
Very objective solution to allow ChartJs animation options.

PR #6214, opened 2020-08-10
#6211 Dialog: add cache-attribute

### LINKED ISSUE #6211: Dialog: Add "cache" attribute
TabView has the attributes dynamic and cache. With the combination of dynamic=true and cache=false you can ensure the content of a tab is reloaded when you switch to the tab.

Dialog currenty only has dynamic-attribute. When you look at https://www.primefaces.org/showcase/ui/data/datatable/selection.xhtml (first example) you see this currently requires a work-around to update the content of the dialog.

christophs78:


melloware (approved):
Well that was pretty easy. :)

PR #6220, opened 2020-08-13
Fix #6219: FilterMeta/SortMeta equals() and hashcode()

### LINKED ISSUE #6219: DataTable: FilterMeta/SortMeta equals() and hashcode()
**Describe the defect**
In our unit tests, we like to compare the some of the data returned by the datatable.getFilterBy(), which returns a Map of <String, FilterMeta>. For instance we setup some initial filters, do some test work, and then verify the filters match what we expect after the work is complete. In PrimeFaces 7.0, would do something like:
```

Map<String, Object> testData = new HashMap<>();
testData,put("key", "value");
assertEquals(testData, datatable.getFilters());
```

The FilterMeta class in Primefaces 8.0 does not implement the equals method. So, the map comparison will fail even if all of the values in the FilterMeta objects match.

**Reproducer**
```
		Map<String, FilterMeta> filters1 = new HashMap<>();
		filters1.put("title", new FilterMeta("title", "title001"));
		
		Map<String, FilterMeta> filters2 = new HashMap<>();
		filters2.put("title", new FilterMeta("title", "title001"));
		
		System.out.println("Equals: " + filters1.equals(filters2));
```
Result is false

**Environment:**
 - PF Version: 8.0.3

**Expected behavior**
Adding a sensible equals method to the FilterMeta class allow for successful comparison of Maps containing the objects of the FilterMeta type.

melloware:
Because these are added to maps they should implement proper equals and hashcode methods

### DIFF
--- a/src/main/java/org/primefaces/model/FilterMeta.java
+++ b/src/main/java/org/primefaces/model/FilterMeta.java
@@ -24,7 +24,10 @@
 package org.primefaces.model;
 
 import java.io.Serializable;
+import java.util.Objects;
+
 import javax.el.ValueExpression;
+
 import org.primefaces.component.api.UIColumn;
 
 public class FilterMeta implements Serializable {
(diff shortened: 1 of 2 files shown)


tandraschko:
I dont think that filterValue should be used?!

melloware:
I an remove it but theoretically if you have someone trying to add the same column and same key the differentiator would be the value as the tie break winner since its the 3rd item checked right?

melloware:
Or your thought is it should replace it if someone tries to add the same column and key.  Thinking about that it does make sense.

PR #6223, opened 2020-08-14
Fix #6219: FilterMeta/SortMeta equals() and hashcode()

### LINKED ISSUE #6219: DataTable: FilterMeta/SortMeta equals() and hashcode()

melloware:


melloware:
@tandraschko I removed sort function but should i also remove sortOrder.  Should it just be column and key for both?

tandraschko:
Exactly, it should replace it
I am even not sure about sortfield, column should be enough?

melloware:
Yep that makes sense.  Make it just columnkey.

melloware:
maybe just columnkey for both Filter and Sort?  Does that make it unique?

melloware:
No I think we need filterField because there is a constructor that does not include columnKey.
```java
public FilterMeta(String filterField, Object filterValue) {
        this.filterField = filterField;
        this.filterValue = filterValue;
    }
```

koan00:
I am curious as to why filterValue was removed from the FilterMeta.equals check ? It seems it would be an important part of checking equality?

melloware:
I think @tandraschko point would be what makes a filter unique should be the columnKeyfilterfield it represents.     If you were to add another of the same columnKey to the map it should replace the other one.  If the filterValue is used it would cause 2 FilterMeta to be added to a map that are for the same Column but have different values no?

koan00:
FilterMeta is on the value side of the Map<String, FilterMeta> of DataTable.getFilterBy(), not the key side. Putting another (key, value) into the map with the same columnKey would replace the older previous FilterMeta regardless of FilterMeta equals implementation. Unless I am mistaken on where else this object is used or the Map your are referring to?

melloware:
No it won't affect our code @tandraschko is saying for completeness.  If for some reason in your code you kept a `Set<FilterMeta>`.

```java
Set set = new HashSet<>;
set.add(new FilterMeta("Col1", "Val2"));
set.add(new FilterMeta("Col1", "Val3));
```

The set above should only have 1 value and it should be Val3 replacing Val2.

koan00:
Yes I understand that hypothetical use case. It just doenst seem to represent what I would think equality means when comparing these types of objects. It also seems odd we would code for a corner case which doesn't actually exist, while not resolving the actual Issue associated to the PR. In our use case for #6219, we test for equality and we would get a false positive in our unit test with the equals implementation that does not include the filterValue.

```
		Map<String, FilterMeta> filters1 = new HashMap<>();
		filters1.put("title", new FilterMeta("title", "title001"));
		
		Map<String, FilterMeta> filters2 = new HashMap<>();
		filters2.put("title", new FilterMeta("title", "title002"));
		
		System.out.println("Equals: " + filters1.equals(filters2));
```
In the case above, the filter value has changed and is different, but it would return true, they are equal. From a logic perspective, we want to test if the filters both exist and have the same value. Without the filterValue in the equals method, we can only determine they exist.

melloware:
Interesting.  @tandraschko ???

tandraschko:
TBH: I think both are actually corner cases :D

Currently in PF its a map yes but we also had a list in the past. And for a DataTable, only one FilterMeta must exist for a column.
Therefore i think its better from PF perspective to implement equals/hashcode for this case.

tandraschko:
shouldnt sort also use sortField like filter?

melloware:
Added sortField back in.

melloware:
@tandraschko think this one is safe to merge now?  I think its definitely accurate now.

jansohn:
> TBH: I think both are actually corner cases :D
> 
> Currently in PF its a map yes but we also had a list in the past. And for a DataTable, only one FilterMeta must exist for a column. Therefore i think its better from PF perspective to implement equals/hashcode for this case.

Sorry for digging up this old comment. `filterBy` is a map since PrimeFaces 8 now and I don't see any indications that we will move back to a list, will we? Can we therefore please add the `filterValue` to the `equals/hashCode` methods as that is the only variable which actually changes during filtering? Duplicates are no issue in my opinion as we already use the `field` name as the map key.

Please reconsider this @tandraschko @melloware @Rapster: three independent developers have now expressed their concerns with the current implementation and have shown real-life examples where this leads to issues (https://github.com/primefaces/primefaces/issues/13013) and unnecessary hacks are needed to make it work.

melloware:
I voted +1 on this @jansohn but @Rapster and @tandraschko were -1.

PR #6225, opened 2020-08-16
Fix #6224: Password ignoreLastPass attribute

### LINKED ISSUE #6224: Password: Ignore LastPass attribute
**Description**
We just got burned by a `p:password` in a dialog form that was hidden and LastPass was filling it out automatically.  You can tell LastPass to ignore fields by adding `data-lpignore="true"` on the element.  I know this can be done with JSF 2.2 PassThrough `pt:data-lpignore="true"` but i think it would be a nice feature of password so users don't have to go look for this special flag and variable and it would work on JSF2.0/2.1 for them.

I was thinking a new `ignoreLastPass` attribute and wanted anyone's opinion @Rapster @tandraschko @christophs78 etc.

see: https://kylenunery.com/lastpass-autofill-rules/

Edit: forgot to mention LastPass is the most popular password manager on the Internet.

melloware:


tandraschko:
+1

PR #6229, opened 2020-08-18
Fix #6187: Row editor prevent multiple focus calls

### LINKED ISSUE #6187: DataTable: Onfocus event fired three times with cellEdit/rowEdit
**Describe the defect**
onfocus fired three times on inputtext with cellEdit DataTable

**Environment:**

- PF Version: 8.0
- JSF + version: 2.2.12-jbossorg-2
- Affected browsers: ALL

**To Reproduce**
2. Click on cell Year
3. Show browser console 
4. See same log three times for focus

**Example XHTML**
```html
        <p:column headerText="Year">
            <p:cellEditor>
                <f:facet name="output"><h:outputText value="#{car.year}" /></f:facet>
                <f:facet name="input"><p:inputText value="#{car.year}" style="width:96%" label="Year" onfocus="console.log('focus row'+#{index})"/></f:facet>
            </p:cellEditor>
        </p:column>
```

melloware:
Use triggerHandler instead of `trigger('focus')` because that propogates the focus to all registered events and bubbles up.  TriggerHandler See: https://stackoverflow.com/questions/3772537/triggerhandler-vs-trigger-in-jquery

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
+++ b/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
@@ -3092,7 +3092,7 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
 
         var inputs=row.find(':input:enabled');
         if (inputs.length > 0) {
-            inputs.first().trigger('focus');
+            inputs.first().triggerHandler('focus');
         }
     },
 


tandraschko:
never used triggerHandler but +1 if well tested

melloware:
I tested but waiting to hear back from the original reporters on whether it fixes all their issues.

melloware:
For lack of response of OP and the fact I have tested this myself with their use case and can see the 3 focus calls going to 1 focus call I am  going to merge this.

PR #6235, opened 2020-08-19
Fix #6226: Added closestWidgetVar EL function

### LINKED ISSUE #6226: EL function closestWidgetVar
Currently you have an EL method `resolveWidgetVar(String expression, UIComponent source)` to find a widget var. I would like to have a method to find the widget var relative to the current component, so go up to the root until a component which implements `Widget` is found.

I created this method in my project:
```
  public static String closestWidgetVar() {
    return Stream.iterate(Components.getCurrentComponent().getParent(),
                          c -> Objects.nonNull(c.getParent()),
                          UIComponent::getParent)
            .filter(Widget.class::isInstance)
            .findFirst()
            .map(Widget.class::cast)
            .map(Widget::resolveWidgetVar)
            .map(w -> String.format("PF('%s')", w))
            .orElse(null);
  }
```
It uses Java 9 and OmniFaces, so it needs a bit of rewriting. If you want to I can create a PR for this.

jepsar:
Fixes #6226

### DIFF
--- a/src/main/java/org/primefaces/expression/SearchExpressionUtils.java
+++ b/src/main/java/org/primefaces/expression/SearchExpressionUtils.java
@@ -87,7 +87,7 @@ public class SearchExpressionUtils {
                 expression);
 
         if (resolvedComponent instanceof Widget) {
-            return javascriptWidgetVar((Widget) resolvedComponent, context);
+            return ((Widget) resolvedComponent).resolveWidgetVar(context);
         }
         else {
             throw new FacesException("Component with clientId " + resolvedComponent.getClientId() + " is not a Widget");
@@ -103,14 +103,10 @@ public class SearchExpressionUtils {
     public static String closestWidgetVarFor(UIComponent component) {
         Widget widget = ComponentTraversalUtils.closest(Widget.class, component);
         if (widget != null) {
-            return javascriptWidgetVar(widget, FacesContext.getCurrentInstance());
+            return widget.resolveWidgetVar(FacesContext.getCurrentInstance());
         }
         else {
             throw new FaceletException("Component with clientId " + component.getClientId() + " has no Widget as parent");
         }
-    }
-
-    private static String javascriptWidgetVar(Widget widget, FacesContext context) {
-        return "PF('" + ((Widget) widget).resolveWidgetVar(context) + "')";
     }
 }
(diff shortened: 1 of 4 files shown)


src/main/java/org/primefaces/expression/SearchExpressionUtils.java
```diff
@@ -85,10 +87,30 @@ public static String resolveWidgetVar(String expression, UIComponent component)
...
-            return "PF('" + ((Widget) resolvedComponent).resolveWidgetVar(context) + "')";
+            return javascriptWidgetVar((Widget) resolvedComponent, context);
         }
         else {
             throw new FacesException("Component with clientId " + resolvedComponent.getClientId() + " is not a Widget");
         }
     }
+
+    // used by p:closestWidgetVar
+    public static String closestWidgetVar() {
```
jepsar: I wasn't sure where to put these methods... is it okay to have them here?

melloware (approved):
Looks good to me but will let Thomas review

tandraschko:
i dont like your change that you automatically add PF('') now
this breaks backward compatibility and doesnt allow to just resolve the widgetVar anymore, without having PF('') around

jepsar:
> i dont like your change that you automatically add PF('') now
> this breaks backward compatibility and doesnt allow to just resolve the widgetVar anymore, without having PF('') around

I didn't change that, I just moved it to a method to DRY things up and fixed the documentation.

https://github.com/primefaces/primefaces/blob/de9a16e875e66c7233aa30fb714b3be7409c2c02/src/main/java/org/primefaces/expression/SearchExpressionUtils.java#L88

.. but I agree it's a bit odd. I was surprised it was working like that. Shall I remove it?

tandraschko:
oha, i see
please, it should just return the varName, not PF('')

tandraschko:
and IMO we should not create 2 methods for it with different names, we had it in the past and its dirty

i think we have 2 solutions:
- test overloading with all current containers (it didnt work in the past)
or
- always pass the UIComponent like we did for resolveClientId. So you always need: #{p:closestWidgetVar(component)}

#{component} like like UIComponet.getCurrentComponent

jepsar:
@tandraschko done

tandraschko:
Please also adjust the migration guide that people must add PF('') now

jepsar:
> Please also adjust the migration guide that people must add PF('') now

Done as well

Issue #2893, opened 2017-11-01
SelectOneMenu: doesn't support string selectItem with combined customContent anymore


wsaca:
Hi, I'm testing the new COLLECT_SCRIPTS and I found the next problem.

## 1) Environment
- PrimeFaces version: 6.2-SNAPSHOT
- Does it work on the newest released PrimeFaces version?
Yes, in version 6.1
- Application server + version: WildFly 10
- Affected browsers: Any browser

## 2) Expected behavior
**p:selectOneMenu** should work if **f:selectItem** is used and the attribute **var** is set.

## 3) Actual behavior
**p:selectOneMenu** throws this error:
javax.el.PropertyNotFoundException: The class 'java.lang.String' does not have the property 'xxxxx'.

## 4) Steps to reproduce
The next example taken of the showcase should reproduce it.

```
<p:selectOneMenu value="#{selectOneMenuView.theme}">
	<f:selectItem itemLabel="Select one..." itemValue="" />
	<f:selectItems value="#{selectOneMenuView.themes}" var="theme" itemLabel="#{theme.displayName}" itemValue="#{theme}" />
	<p:column>
		<h:outputText styleClass="ui-theme ui-theme-#{t.name}" />
	</p:column>
	<p:column>
		<h:outputText value="#{t.displayName}" />
	</p:column>
</p:selectOneMenu>
```

I think this problem was not introduced intentionally, the commit related is: https://github.com/primefaces/primefaces/commit/49c95a753328691ce07695315600ccb24bff4f8a

The lines (363 to 369) were removed:
https://github.com/primefaces/primefaces/blob/f83db0b51eb152f81ea10670689659f49f5be41e/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java#L362-L382

melloware:
I think this commit is the one that lost it:  https://github.com/primefaces/primefaces/commit/49c95a753328691ce07695315600ccb24bff4f8a#diff-e4d3a342df3de892c8fd79acabe27e6b

melloware:
I am running the showcase locally with the latest 6.2-SNAPSHOT and the advanced custom content example is working fine for me...

![image](https://user-images.githubusercontent.com/4399574/32296396-8a91cc42-bf22-11e7-914e-97e6b9cda2f6.png)

wsaca:
The showcase works fine, you should add this line to test it:
<f:selectItem itemLabel="Select one..." itemValue="" />

[...]

PR #6239, opened 2020-08-21
Fix #6236: Apply converter to submitted string values

### LINKED ISSUE #6236: SelectCheckboxMenu: Calls Converter#getAsString with serialized string, compares string with object
Unless I'm doing something horribly wrong, it seems that `SelectCheckboxMenuRenderer#encodeMultipleLabel` mixes the converted string value with the actual object in at least two places.

**Reproducer**

* Clone the current showcase https://github.com/primefaces/primefaces-showcase (9.0-SNAPSHOT)
* Edit `src/main/webapp/ui/input/checkboxMenu.xhtml` and add the following xhtml
* Create a `AtLeast2ValidatorClass` with the code below
* Start the showcase, go to `/showcase/ui/input/checkboxMenu.xhtml`, select 1 theme, and press the `Update` button.
* You get an exception with the stack trace as shown below

```xml
                <h:outputLabel for="themes" value="Themes:" />
                <p:selectCheckboxMenu id="themes" value="#{selectManyView.selectedThemes}" label="Themes"
                				      multiple="true" showCheckbox="true"
               						  converter="#{themeConverter}"
                                      filter="true" filterMatchMode="startsWith" panelStyle="width:250px">
                    <f:validator validatorId="AtLeast2Validator"/>
                    <f:selectItems value="#{selectManyView.themes}" var="theme" itemLabel="#{theme.displayName}" itemValue="#{theme}" />
                </p:selectCheckboxMenu>
                <p:commandButton id="update" value="Update" process="themes" update="themes" partialSubmit="true"/>
```

```java
package org.primefaces.showcase.convert;

import java.util.Collection;

import javax.faces.component.UIComponent;
import javax.faces.validator.FacesValidator;
import javax.faces.validator.Validator;
import javax.faces.validator.ValidatorException;

@FacesValidator(value = "AtLeast2Validator")
public class AtLeast2Validator implements Validator {

	@Override
	public void validate(FacesContext context, UIComponent component, Object value) throws ValidatorException {
		if (value instanceof Collection<?> && ((Collection) value).size() < 2) {
			throw new ValidatorException(new FacesMessage(FacesMessage.SEVERITY_ERROR, "Invalid", "At least 2 items required."));
		}
	}
}
```

```
Aug 21, 2020 11:00:50 AM com.sun.faces.application.view.FaceletViewHandlingStrategy handleRenderException
SEVERE: Error Rendering View[/ui/input/checkboxMenu.xhtml]
java.lang.ClassCastException: class java.lang.String cannot be cast to class org.primefaces.showcase.domain.Theme (java.lang.String is in module java.base of loader 'bootstrap'; org.primefaces.showcase.domain.Theme is in unnamed module of loader org.eclipse.jetty.webapp.WebAppClassLoader @2d36a2a6)
	at org.primefaces.showcase.convert.ThemeConverter.getAsString(ThemeConverter.java:54)
	at org.primefaces.renderkit.SelectRenderer.getOptionAsString(SelectRenderer.java:206)
	at org.primefaces.component.selectcheckboxmenu.SelectCheckboxMenuRenderer.encodeMultipleLabel(SelectCheckboxMenuRenderer.java:245)
	at org.primefaces.component.selectcheckboxmenu.SelectCheckboxMenuRenderer.encodeMarkup(SelectCheckboxMenuRenderer.java:93)
	at org.primefaces.component.selectcheckboxmenu.SelectCheckboxMenuRenderer.encodeEnd(SelectCheckboxMenuRenderer.java:64)
```

**Cause**

* [SelectCheckboxMenuRendere#L235](https://github.com/primefaces/primefaces/blob/master/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java#L235) uses the `submittedValues` of the UIInput if available, which is an array of strings
* The converter in the repro above is necessary only because otherwise the `submittedValues` are reset to `null` when there are no validation errors
* So when submittedValues is not null, the Converter#getAsString is called with the submitted string value instead of the object in [SelectCheckboxMenuRenderer#245](https://github.com/primefaces/primefaces/blob/master/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java#L245)
* Similarly, in [SelectCheckboxMenuRenderer#L252](https://github.com/primefaces/primefaces/blob/master/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java#L252) and L258, the string value is compared against the object from the `SelectItem`. This doesn't work and results in no item being selected upon redisplay.

When I replace `p:selectCheckboxMenu` in the XHTML snippet above with `p:selectManyMenu`, it works as expected. so this is probably a bug that should be fixed.

blutorange:
* SelectCheckboxMenu uses the submitted string values when available, but did not always convert  them via the converter
* Extract comparison code from SelectRenderer#isSelected into new method  #isSelectValueEqual
* Use that method in #encodeMultipleLabel, this takes care of applying the converter
* Also in #encodeMultipleLabel, when converting the value back to a string, don't apply the converter if the value is already a string.
* Also fixed a minor issue in forms.selectmanymenu.js, where the ui-state-active style class was not set during the #init method when using custom content (The checkbox is now inside a TD element, but the CSS selector did not account for that.)

### DIFF
--- a/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
+++ b/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
@@ -26,7 +26,6 @@ package org.primefaces.component.selectcheckboxmenu;
 import java.io.IOException;
 import java.lang.reflect.Array;
 import java.util.List;
-import java.util.Objects;
 
 import javax.faces.component.UIComponent;
 import javax.faces.component.UINamingContainer;
@@ -242,20 +241,19 @@ public class SelectCheckboxMenuRenderer extends SelectManyRenderer {
             int length = Array.getLength(valuesArray);
             for (int i = 0; i < length; i++) {
                 Object value = Array.get(valuesArray, i);
-                String itemValueAsString = getOptionAsString(context, menu, converter, value);
 
                 SelectItem selectedItem = null;
                 for (SelectItem item : selectItems) {
                     if (item instanceof SelectItemGroup) {
                         SelectItemGroup group = (SelectItemGroup) item;
                         for (SelectItem groupItem : group.getSelectItems()) {
-                            if (value.equals(groupItem.getValue())) {
+                            if (isSelectValueEqual(context, menu, value, groupItem.getValue(), converter)) {
                                 selectedItem = groupItem;
                                 break;
                             }
                         }
                     }
-                    else if (Objects.equals(value, item.getValue())) {
+                    else if (isSelectValueEqual(context, menu, value, item.getValue(), converter)) {
                         selectedItem = item;
                         break;
                     }
@@ -263,6 +261,14 @@ public class SelectCheckboxMenuRenderer extends SelectManyRenderer {
 
                 // #5956 Do not render a chip for the value if no matching option exists
                 if (selectedItem != null) {
+                    String itemValueAsString;
+                    if (value == null || value instanceof String) {
+                        itemValueAsString = (String) value;
+                    }
+                    else {
+                        itemValueAsString = getOptionAsString(context, menu, converter, value);
+                    }
+
                     writer.startElement("li", null);
                     writer.writeAttribute("class", SelectCheckboxMenu.TOKEN_DISPLAY_CLASS, null);
                     writer.writeAttribute("data-item-value", itemValueAsString, null);
--- a/src/main/java/org/primefaces/renderkit/SelectRenderer.java
+++ b/src/main/java/org/primefaces/renderkit/SelectRenderer.java
@@ -43,6 +43,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.RandomAccess;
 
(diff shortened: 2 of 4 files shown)


melloware (approved):
excellent PR

tandraschko:
in general it looks really great but it would be really great if you could do some tests for it
in the past we often had trouble with this stuff like when is a converter required;
at least please check tickets like https://github.com/primefaces/primefaces/issues/2893, if we dont break such cases ;D

blutorange:
I tested the examples from the showcase, but I'll look into writing a test for this.

blutorange:
Alright, I 
* [added a test](https://github.com/primefaces/primefaces/pull/6239/commits/509be91dcdc4b6bc0653277040bef6e9c0f61acd) for the parts of the SelectRenderer I changed to make sure components other than SelectCheckboxMenu are not affected. Ran the test once without my changes and once with my changes, and passed both times.
* checked #2893 for SelectOneMenu and it still works, same setup for SelectCheckboxMenu works too (although a "Select One" option does not really make sense, but it works)

tandraschko:
thanks!

PR #6242, opened 2020-08-23

### LINKED ISSUE #6169: DataTable: Lazy getRowData Repeat the query twice
**1) Environment**

PrimeFaces version: 8.0

**2) Expected behavior**

**3) Actual behavior**

**AbstractLazyModel**

```

    private T[] selects;
    private T select;

        if (lazyDataModel == null) {
                @Override
                }

                @Override
                }

                @Override
                    return getKey(t);
                }
            };
        }
        return lazyDataModel;
    }

    protected void refresh() {
        this.setSelect(null);
        search();
    }

    public void search() {
        lazyDataModel = null;
    }
```

**Example Bean**

```
@Slf4j
@Data
@Component
@ViewScoped

    @Autowired


    @Autowired

    @Autowired

    private String codeName;


        this.refresh();
    }

    }




    @PreDestroy
    public void destroy() {
    }


    @Override
    }


    @Override
        return erpSysCodeTypeService;
    }

    @Override
    }

}

```

**Example Xhtml**

```
  <p:dataTable id="dataTableId"
                                         paginatorPosition="bottom"
                                         rowIndexVar="rowIdx"
                                         rows="10"
                                         lazy="true"
                                         selectionMode="single"
                                         rowKey="#{_sysCodeType.id}"
                                         var="_sysCodeType"
                                         emptyMessage="暂无数据."
                                         paginator="true"

                                <p:column width="50">
                                </p:column>

                                <p:column headerText="字典代码">
                                </p:column>

                                <p:column headerText="字典名称">
                                </p:column>

                                <p:column headerText="启用">
                                </p:column>

                                <p:column headerText="创建时间">
                                    </h:outputText>
                                </p:column>

                            </p:dataTable>

```

Rapster:
This PR should fix the corresponding issue, and bring few enhancements also:
- `DataTable#rowKey` should be the single source of truth, today either you use `DataTable#rowKey` or `LazyDataModel#getRowKey` depending your use case, which is a bit confusing.
- in terms of perf:
  - Selection of one element: today it's O(n) which is okay (not great but okay...)
  - Select All: today it would be O(N²) which is pretty bad.
  - Encoding one page: O(N²), `DataTable#getRowData` is called for every row

The solution is pretty simple to make it O(1): using a cache. `DefaultSelectableDataModel` wraps any data model, and build a cache every time `DefaultSelectableDataModel#setWrappedData` is called.

Also, `DataTable#rowKey` becomes optional, by default `System#identityHashCode` is used to create a rowKey of an item.

Let me know what you think, there is only one breaking change: LazyDataModel no longer implements SelectableDataModel, but as I said, since there is a default value for rowKey, the migration should be pretty quick

N.B: this PR is a draft, due to JSF impl few things are not as good as I hoped...


melloware (approved):
i actually like this change a lot and it solves the problem of multiple get RowData calls

tandraschko:
puh, thats not a easy PR which we can just merge

we need at least tests
when i see code like "row key cache", i fear that we can break a lot of stuff

also i would like to see another PR with just the isValueLazyDataModel change, sp we can easier look at the other changes
in the PR i would to see something like:
getLazyDataModel()
which returns null if it's not a lazyDataModel, so we reduce the calls to getDataModel, which likely calls a EL each time

Rapster:
I'm not worried about that internal cache since datamodel is constantly reseted (set to null)... I'm trying to figure out a way to improve the `DefaultSelectableModel#isDirty` method. If I can't, I guess we'll have to live with that, although I think few modifications of this PR should go into PF core (like getting rid of LazyDataModel implementing SelectableModel)

Rapster:
I'm gonna close this PR as I don't see a nice and simple way to avoid building cache every time datamodel is reseted.

Issue #5309, opened 2019-11-06
PickList: Multiple Entries


e-reznik:
## 1) Environment
- PrimeFaces version: **7.0**
- Application server + version: **Payara 5**
- Affected browsers: **Firefox**

## 2) Expected behavior
I want to have some entries multiple times in my right list. So the entries shouldn't disappear from the left list.
![grafik](https://user-images.githubusercontent.com/55981254/68288093-15399600-0084-11ea-8f86-186a0d6a0670.png)

## 3) Actual behavior
When you add (i.e. transfer) one element from left to right, it appears at the right list, but disappears from the left list
![grafik](https://user-images.githubusercontent.com/55981254/68288152-2c788380-0084-11ea-8cf1-eda08cd1fb74.png)

## Possible Solution

PickList has a method [onTransfer](https://www.primefaces.org/docs/vdl/7.0/primefaces-p/pickList.html):

> Client side callback to execute when an item is transferred from one list to another.

So I could check, which element was transferred (deleted from the left list) and put it there again. But maybe there is a more elegant way.

melloware:
Isn't this a duplicate of this: https://github.com/primefaces/primefaces/issues/2858

tandraschko:
I think #2858 can be a part of this ticket.
But this ticket also allows duplicate of the same items in the target and source.

I think we can close #2858 in favor of this.

melloware:
Replaced by https://github.com/primefaces/primefaces/issues/11249

PR #6244, opened 2020-08-24
New Feature #5309: Multiple entries in pickList

e-reznik:
I'm trying to implement the new feature request of [allowing multiple entries in the PickList](https://github.com/primefaces/primefaces/issues/5309).
So far, I've just added the new attribute _allowMultiple_, without any functionality yet.

As this is my first attempt to contribute to such a huge project, every guidance is appreciated :relaxed:

### DIFF
--- a/src/main/java/org/primefaces/component/picklist/PickListBase.java
+++ b/src/main/java/org/primefaces/component/picklist/PickListBase.java
@@ -100,11 +100,11 @@ public abstract class PickListBase extends UIInput implements Widget, ClientBeha
         getStateHelper().put(PropertyKeys.var, var);
     }
 
-    public String getAllowDuplicates() {
-        return (String) getStateHelper().eval(PropertyKeys.allowDuplicates, null);
+    public boolean isAllowDuplicates() {
+        return (Boolean) getStateHelper().eval(PropertyKeys.allowDuplicates, false);
     }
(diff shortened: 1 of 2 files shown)


melloware:
Good start. How about `allowDuplicates` as more accurate?

e-reznik:
@melloware: Thanks, I changed the name.

src/main/java/org/primefaces/component/picklist/PickListBase.java
```diff
@@ -99,6 +100,14 @@ public void setVar(String var) {
         getStateHelper().put(PropertyKeys.var, var);
     }
 
+    public String getAllowDuplicates() {
```
melloware: This should return `boolean` and be `isAllowDuplicates`

melloware (changes requested):

src/main/java/org/primefaces/component/picklist/PickListBase.java
e-reznik: My bad. Fixed.

melloware:
Looks good to continue

tandraschko:
I will close it for now for clarity of our PR list.
Feel free to create a PR again when you finished the implementation :) 

Thanks!

PR #6252, opened 2020-08-26
Fix #4374 menuButton: Add support for submenus

### LINKED ISSUE #4374: menuButton: add support for submenus
PrimeFaces' `<p:menubar>` component supports nested `<p:submenu>`s to group menu-items into sub-menus. This grouping is currently not supported for `<p:menuButton>`. `<p:submenu>`s inside a menuButton is silently ignored. It would be nice to also have sub-menu support in menu-buttons.

## 1) Environment

PrimeFaces 6.2

maruTA-bis5:
This PR add submenus support to menuButton component. Fixes #4374

Example (source: https://github.com/maruTA-bis5/primefaces-test/tree/menubutton-submenu)
image

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/menu/menu.menubutton.js
+++ b/src/main/resources/META-INF/resources/primefaces/menu/menu.menubutton.js
@@ -40,9 +40,7 @@ PrimeFaces.widget.MenuButton = PrimeFaces.widget.TieredMenu.extend({
      * @param {JQuery} submenu 
      */
     showSubmenu: function(menuitem, submenu) {
-        var pos = null;
-
-        pos = {
+        var pos = {
             my: 'left top',
             at: 'right top',
             of: menuitem,
--- a/src/main/resources/META-INF/resources/primefaces/menu/menu.menubutton.js
+++ b/src/main/resources/META-INF/resources/primefaces/menu/menu.menubutton.js
@@ -18,7 +18,7 @@
  * e.g., `flip`, `fit`, `fit flip`, `fit none`.
  * @prop {boolean} cfg.disabled Whether this menu button is initially disabled.
  */
-PrimeFaces.widget.MenuButton = PrimeFaces.widget.BaseWidget.extend({
+PrimeFaces.widget.MenuButton = PrimeFaces.widget.TieredMenu.extend({
 
     /**
      * @override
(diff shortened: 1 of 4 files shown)


melloware:
I am not sure how others will feel extending TieredMenu but I like this change.

src/main/resources/META-INF/resources/primefaces/menu/menu.menubutton.js
```diff
@@ -54,6 +75,14 @@ PrimeFaces.widget.MenuButton = PrimeFaces.widget.BaseWidget.extend({
      * @private
      */
     bindEvents: function() {
+        this.menuId = this.jqId + '_menu';
+        this.button = this.jq.children('button');
+        this.menu = this.jq.children('.ui-menu');
+        this.menuitems = this.jq.find('.ui-menuitem');
+        this.cfg.disabled = this.button.is(':disabled');
```
melloware: Not sure why you moved this out of the `init` method?  Usually initialization of values like this is done in the `init` method.

melloware (changes requested):

tandraschko:
in general it looks ok, also i wonder about mellowares statement
and please dont do things like
        var pos = null;
and asign `pos` in the next statement; just do `var pos = {}`

src/main/resources/META-INF/resources/primefaces/menu/menu.menubutton.js
maruTA-bis5: If initialization is in `init` method, `bindEvents` called from `TieredMenu#init` before initialization and raised script error 
caused by undefined reference.

But I found another solution. I'll fix this.
I'm sorry for my lack of experience with JavaScript.

src/main/resources/META-INF/resources/primefaces/menu/menu.menubutton.js
```diff
@@ -35,11 +35,37 @@ PrimeFaces.widget.MenuButton = PrimeFaces.widget.BaseWidget.extend({
...
+
+        this.timeoutId = setTimeout(function () {
+           submenu.css('z-index', PrimeFaces.nextZindex())
+                  .show()
+                  .position(pos)
+        }, this.cfg.delay);
+    },
+
     /**
      * @override
```
melloware, line 70: You can remove the @override and @inheritdoc here as now that you have renamed the method its no longer overriding the parent.

melloware (changes requested):

src/main/resources/META-INF/resources/primefaces/menu/menu.menubutton.js
maruTA-bis5: `bindButtonEvents` method does not have `@override` and `@inheritdoc`. This line is for `refresh` method.

melloware: sorry the way the review looks the code folding confused me. I see it in the raw file now.

melloware (approved):

tandraschko:
I will merge it now, would be great if someone coud retest in the showcase

tandraschko:
Thanks for the PR!

tandraschko:
the inhertiance changes breaks jsdoc generation. Can someone check it?

melloware:
I can check this.

melloware:
OK I fixed the JSDocs and I just tested MenuButton in the Showcase and everything is looking good.

PR #6254, opened 2020-08-26
Fix #6250,#1493: HTML5 events on Input components

### LINKED ISSUE #1493: ClientValidator should support "oninput" event and probably many other
I wrote a fix for that. The issue comes from `HtmlInputText.getEventNames()` which does not include `input` event name, it does not support much actually... I had to override `InputText.getEventNames` to fix this issue. I really like this feature as it's really reactive and know client asking for instant validation.

Here is the list of event names supported: http://www.w3schools.com/tags/ref_eventattributes.asp Not all of them are for input, but some :)

As I think it's a problem from mojarra and not Primefaces, I don't know what to do :/

### LINKED ISSUE #6250: InputText: Support HTML5 events
**Description**
If text is pasted via the browser menu, the change in the inputtext field goes unnoticed on the server side. Even if the component is registered for 'change' Ajax events, this change will not be propagated immediately. This is especially nasty if you rely on detecting the dirty state of a form on the server side.

A working solution has already be found in the following forum post: [p:inputText start listener if a value was pasted?](https://forum.primefaces.org/posting.php?mode=edit&f=3&p=184227).

A listener should be added to the 'paste' event of the inputfield, which triggers the Ajax change event.

The solution described in the post will only work on a single component, adding this to the component itself is more desirable and add consistent behavior for the different text-based input components.

For consistency, this change should also be added to InputTextarea component as well. TextEditor seems to already cover this nicely out of the box.

melloware:
**Components updated:**
- AutoComplete
- Calendar
- Chips
- DatePicker
- InputMask
- InputNumber
- InputText
- InputTextArea
- Keyboard
- Password
- Spinner

Because of inheritance I created two new classes `AbstractPrimeHtmlInputText` and `AbstractPrimeHtmlInputTextArea` to accomplish this. .  I then updated the HTML.COMMON_EVENTS and all of the Markdown files and TagLib XML to include these new events on all components.

I have done testing and it all works great!  Adding this showcase example for ìnput` so cut, copy and paste can be shown.
```xml
<p:inputText id="lastname" value="#{userView.lastname}">
       <p:ajax event="input" update="out3" />
</p:inputText>
```


tandraschko:
maybe we should use names like: AbstractPrimeHtmlInputTextarea

tandraschko:
otherwise cool PR!

melloware:
OK will rename them to AbstractPrime ...

tandraschko:
+1

PR #6259, opened 2020-08-28
Fix #6258: Improve modal support

### LINKED ISSUE #6258: Modal: Focus Trap
**Description**
We have 3 separate issues that I tried to solve using `pointer-events: none` CSS to block keyboard and mouse input to the modal area.  None of them quite work to solve all use cases:

BlockUI: https://github.com/primefaces/primefaces/issues/1325
Dialog: https://github.com/primefaces/primefaces/issues/1991
SideBar: https://github.com/primefaces/primefaces/issues/6245

All have slightly different use cases but similar problems.

I think we need to use a Focus Trap to solve this issue.  Everything I am reading on Stack Overflow points to that solution and there is a nice little Vanilla JS library that does exactly this: https://github.com/focus-trap/focus-trap

See: https://stackoverflow.com/questions/50178419/how-can-restrict-the-tab-key-press-only-within-the-modal-popup-when-its-open

From their site describes our exact problem:
_Trap focus within a DOM node. There may come a time when you find it important to trap focus within a DOM node — so that when a user hits Tab or Shift+Tab or clicks around, she can't escape a certain cycle of focusable elements. You will definitely face this challenge when you are trying to build accessible modals._

I think we nee to remove `pointer-events` etc and use this library.  I know that seems like a larger change but to me MODAL should truly block all keyboard and mouse input on the modal area and right now that is not happening.

I would like @christophs78 @tandraschko @Rapster or anyone else's thoughts on this?

melloware:
OK so i found out a lot today and I tested this changes and they are working. But not perfect.

So based on ARIA specs modal's should get `aria-modal:true` and `role=dialog` that tells both the browser and screenreaders that this is the only thing that should have focus and be tabbable which is great.
https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html

So I have refactored our `addModal` and `removeModal` method to add these aria attributes and it works great.  HOWEVER there is a cavaeat. For the browser to truly do it right like the modal example above you need a Main area and the dialogs to live outside of it which we can't do..

```xml
<main aria-hidden="true"><!-- main content here--></main>
<dialog role="dialog" aria-modal="true">Your dialog here</dialog>
```

because our dialogs can be anywhere in the DOM is why this doesn't work all the way.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/blockui/blockui.css
+++ b/src/main/resources/META-INF/resources/primefaces/blockui/blockui.css
@@ -1,14 +1,10 @@
 .ui-blockui {
     position: absolute;
-    top: 0;
-    left: 0;
     text-align: center;
-    width: 100%;
-    height: 100%;
-    pointer-events: none;
-    -webkit-user-select: none;
-    -moz-user-select: none;
-    -ms-user-select: none;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
     user-select: none;
 }
 
--- a/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
+++ b/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
@@ -80,15 +80,24 @@ if (!PrimeFaces.utils) {
 
         /**
          * Creates a new (empty) container for a modal overlay. A modal overlay is an overlay that blocks the content
-         * belown it. To remove the modal overlay, use `PrimeFaces.utils.removeModal`.
+         * below it. To remove the modal overlay, use `PrimeFaces.utils.removeModal`.
          * @param {PrimeFaces.widget.BaseWidget} widget An overlay widget instance.
-         * @param {number} zIndex The z-index to set on the modal overlay.
+         * @param {JQuery} overlay The modal overlay element should be a DIV.
          * @param {() => JQuery} tabbablesCallback A supplier function that return a list of tabbable elements. A
          * tabbable element is an element to which the user can navigate to via the tab key.
          * @return {JQuery} The DOM element for the newly added modal overlay container.
          */
-        addModal: function(widget, zIndex, tabbablesCallback) {
-            var id = widget.id;
+        addModal: function(widget, overlay, tabbablesCallback) {
+            var id = widget.id,
+                zIndex = overlay.css('z-index') - 1;
+            
+            overlay.attr({
+                'role': 'dialog'
+                ,'aria-hidden': false
+                ,'aria-modal': true
+                ,'aria-live': 'polite'
+            });
+            
             PrimeFaces.utils.preventTabbing(id, zIndex, tabbablesCallback);
 
             if (widget.cfg.blockScroll) {
@@ -96,7 +105,6 @@ if (!PrimeFaces.utils) {
             }
 
             var modalId = id + '_modal';
-
             var modalOverlay = $('<div id="' + modalId + '" class="ui-widget-overlay ui-dialog-mask"></div>');
             modalOverlay.appendTo($(document.body));
             modalOverlay.css('z-index' , String(zIndex));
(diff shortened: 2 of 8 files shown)


tandraschko:
so we cant apply it actually because we cant assume that everone has main element?

melloware:
Right we need to ensure that dialogs live outside the main element and that there is a main element marked with `aria-hidden=true` when the dialog is displayed.

melloware:
I guess I could look for a MAIN element and apply if it it exists?

melloware:
Working example here: https://github.com/primefaces/primefaces/issues/6258#issuecomment-683317472

tandraschko:
wait wait wait - please dont just merge such core PRs without reviews.
what happens now if you dont have a main? is it completly broken compared to 8.0?

melloware:
I didn't do anything with Main I just posted an example XHTML with a main if you really want to fix this.  But PF can't fix this at its core.  What I did was clean up the modal handling for screen reader's and browsers to be correct.

Please read: https://github.com/primefaces/primefaces/issues/6258#issuecomment-683317472

melloware:
But as to functionality nothing was changed and I actually reverted the CSS from this PR: https://github.com/primefaces/primefaces/pull/5948

As they actually were not fixing the situation.

christophs78:
Should we add some of your findings to dialog-docu? And maybe add two JS-methods to PF-dialog which may be called from onshow and onhide instead of the custom JS code.

melloware:
Great idea i have added it to the Dialog documentation with a new PR.     For Now I left the custom JS code in the example rather than add to PF in case people want to customize it further or not use a MAIN element and maybe just use a DIV etc.
Check it out please: https://github.com/primefaces/primefaces/pull/6266

christophs78:
Looks good. Thank you!

stolp:
As you have added this both to the 8.0 and 9.0 documentation I assume this approach will work with PrimeFaces 8 as well?

If yes, this would be a great generic solution for resolving [(https://forum.primefaces.org/viewtopic.php?f=3&t=61198)]!

Luckily I already kept all my dialogs in a separate section.

melloware:
Yep don't see any reason why this solution would not work in 8.0 as the dialog was already adding aria-modal=true to the dialog. Let me know if it works!

stolp:
I just tried this, but sadly it did not work (registered keyboard events are still triggered in the main section).

There is a slight difference in my application as both main and the div containing the dialogs share the same form. Is this of relevance?

Also the dialog activation is not triggered via PF() but conditionally from the server side. Is this relevant here?

stolp:
Please ignore the above. It is working great.

melloware:
Awesome!

stolp:
My deepest thanks for this awesome patch. This came just in time.

melloware:
Glad I could help.

stolp:
I did some more testing and sadly discovered problems when the dialog itself contains Hotkey elements.

In this case, the action of a CommandButton works, but when the same action is invoked from the corresponding Hotkey it does not work.

I think we need to investigate this further.

melloware:
If you can use my example zip runner to provide an example.  But my guess is anything contained in the dialog is still active and it has not been deactivated.

melloware:
You can use this as a starting point: 
[modal-dialog.zip](https://github.com/primefaces/primefaces/files/5149950/modal-dialog.zip)

Just run `mvn clean jetty:run` and navigate to http://localhost:8080/primefaces-test/test.xhtml

stolp:
Thanks - will look into this later.

stolp:
I tried to reproduce this with a small adaptions to more closely resemble my code but sadly to no avail.

The situation in my application is a bit more complex than the reproducer and there are other factors which may have an influence here.

Due to time constraints I need to stop further investigation for the next days, sorry. I would love to continue later with this.

melloware:
No problem if you come back to it down the road I will be here to try and help figure it out

PR #6265, opened 2020-08-29
Calendar / DatePicker: Fix pattern conversion #6257

### LINKED ISSUE #6257: Calendar/DatePicker: Error with SimpleDatePattern with 1-letter year (eg. dd.MM.y)
**Describe the defect**

So unless I understand the docs wrongly, the `pattern` option on <p:calendar> and <p:datePicker> allows for a [SimpleDateFormat](https://docs.oracle.com/javase/10/docs/api/java/text/SimpleDateFormat.html) pattern.

When the year portion of the pattern is  `yyyy` (full year) or `yy` (2 digits), everything works. However, when it is only `y`, it won't work (see below).

This is because the [SimpleDateFormat](https://docs.oracle.com/javase/10/docs/api/java/text/SimpleDateFormat.html) interprets `y` as:

> For formatting, if the number of pattern letters is 2, the year is truncated to 2 digits; **otherwise it is interpreted as a number**.

While the [JQueryUI datepicker](https://api.jqueryui.com/datepicker/#utility-formatDate) (and thus p:calendar and p:datepicker which uses the same algorithm) interprets `y` as 

> y - year (two digit)

So the Date gets formatted as `20 May 2020` by the server, but the client expects `20 May 20`.

Now I suppose most people would just use `yyyy`, but `y` is essentially the full year with a fall back for 2-digit year numbers.

How I even found out about this issue is because I had been using the default pattern from SimpleDataFormat.LONG. Observe how the following program gives different outputs for different Java versions:

```java
import java.text.SimpleDateFormat;
import java.util.Locale;

public class Test {
  public static void main(String[] args) {
     String s =((SimpleDateFormat)SimpleDateFormat.getDateTimeInstance(SimpleDateFormat.LONG, SimpleDateFormat.SHORT, Locale.GERMAN)).toPattern();
     System.out.println(s);
  }
}
```

With Java 14:

```bash
user@mypc:/tmp/javatest$ java -version
openjdk version "14.0.1" 2020-04-14
OpenJDK Runtime Environment (build 14.0.1+7-Ubuntu-1ubuntu1)
OpenJDK 64-Bit Server VM (build 14.0.1+7-Ubuntu-1ubuntu1, mixed mode, sharing)
user@mypc:/tmp/javatest$ javac Test.java && java Test
d. MMMM y, HH:mm
user@mypc:/tmp/javatest$ javac Test.java && java -Djava.locale.providers=COMPAT,CLDR Test
d. MMMM yyyy HH:mm
```

With Java 8:

```bash
user@mypc:/tmp/javatest$ java -version
openjdk version "1.8.0_265"
OpenJDK Runtime Environment (build 1.8.0_265-8u265-b01-0ubuntu2~20.04-b01)
OpenJDK 64-Bit Server VM (build 25.265-b01, mixed mode)
user@mypc:/tmp/javatest$ javac Test.java && java Test
d. MMMM yyyy HH:mm
```

It seems like the one letter year format `y` is the default since Java 9, although that can be changed with system properties https://stackoverflow.com/a/57050058

**Reproducer**

[reproducer.zip](https://github.com/primefaces/primefaces/files/5141453/reproducer.zip)

**Environment:**
* PF Version: _9.0-SNAPSHOT_

**To Reproduce**

Click in the date input field. A date picker should show up, but instead an error message is logged in the console.

```xhtml
<p:datePicker id="timeDe" value="#{testView.testDate}" showTime="false" pattern="dd.MM.y"/>
```

image

blutorange:
This commit only affects the method CalendarUtils#convertPattern.

Instead of replacing regexp matches, I convert the the pattern the proper way by (a) first parsing it into a token stream, then (b)  converting each token to the closest jquery date/time picker approximation. This also allows patterns with quoted sections to be handled correctly.

Quick performance check shows that this performs similar to the old implementation, possibly faster (since no regular expressions are involved anymore), certainly not significantly slower.

I added a unit test [DateTimePatternConverterTest](https://github.com/primefaces/primefaces/pull/6265/files#diff-1b8150f76c9b784a1cc9a8d930d2f381) that shows which patterns convert the same way as before, and which now convert differently.

I also tested the showcase examples (`calendar/calendar.xhtml`,  `calendar/calendarJava8.xhtml`, `datepicker/datePicker.xhtml`, `datepicker/datePickerJava8.xhtml`) and they work.

For reference:
* JQuery UI datepicker format: https://api.jqueryui.com/datepicker/#utility-parseDate
* JQuery timepicker format: https://trentrichardson.com/examples/timepicker/
* SimpleDataFormat: https://docs.oracle.com/javase/10/docs/api/java/text/SimpleDateFormat.html
* DateTimeFormatter: https://docs.oracle.com/javase/10/docs/api/java/time/format/DateTimeFormatter.html

The following patterns convert differently:

* Literals and patterns with quotes now convert correctly:
    * Pattern `@` converts to `'@'` (previously `@`)
    * Pattern `'bar'` converts to `bar` (previously `'bTTr'`)
    * Pattern `'yyyy'` convert to `'yyyy'` (previously `'yy'`)
* Pattern **E** (*Day name in week*)
    * Previously converted to **E** (not a valid jquery datepicker pattern)
    * Now converted to **D** (*day name short*)
* Pattern **S** (*Millisecond*)
    * Previously converted to **S** (not a valid jquery timepicker pattern)
    * Now converted to **l** (*Milliseconds*)
* Pattern **X** (*zone-offset*)
    * Previously converted to **X** (not a valid jquery timepicker pattern)
    * Now converted to **z** (*Timezone as defined by timezoneList*)
* Pattern **u** (*year*)
    * Previously converted to **u** (not a valid jquery datepicker pattern)
    * Now converted to **yy** (*year (four digit)*)
    * Relates to #5563 : It was reasoned there that a single `u` might be intended as the `SimpleDateFormat` **u** (*Day number of week*) - but I don't see how this could ever work out seeing that `u` is not a valid jquery datepicker pattern. 
* Pattern **y** (*year*)
    * Previously converted to **y** (year (two digit))
    * Now converted to **yy** (*year (four digit)*)
    * Fixes #6257
* Pattern **EE** (*Day name in week*)
    * Previously converted to **EE** (not a valid jquery datepicker pattern)
    * Now converted to **D** (*day name short*)
* Pattern **SS** (*Millisecond*)
    * Previously converted to **SS** (not a valid jquery timepicker pattern)
    * Now converted to **l** (*Milliseconds*)
* Pattern **zz** (time zone name)
    * Previously converted to **zz** (not a valid jquery timepicker pattern)
    * Now converted to **z** (*Timezone as defined by timezoneList*)
* Pattern **ddd** (*day-of-month*)
    * Previously converted to **ddd** (not a valid jquery timepicker pattern)
    * Now converted to **dd** (*day of month (two digit)*)
* Pattern **zzz** (time zone name)
    * Previously converted to **zzz** (not a valid jquery timepicker pattern)
    * Now converted to **z** (*Timezone as defined by timezoneList*)
* Pattern **dddd** (*day-of-month*)
    * Previously converted to **dddd** (not a valid jquery timepicker pattern)
    * Now converted to **dd** (*day of month (two digit)*)
* Pattern **uuu** (*year*)
    * Previously converted to **yu** (not a valid jquery datepicker pattern)
    * Now converted to **yy** (*year (four digit)*)
* Pattern **zzzz** (time zone name)
    * Previously converted to **zzzz** (not a valid jquery timepicker pattern)
    * Now converted to **z** (*Timezone as defined by timezoneList*)
* Pattern **ddddd** (*day-of-month*)
    * Previously converted to **ddddd** (not a valid jquery timepicker pattern)
    * Now converted to **dd** (*day of month (two digit)*)
* Pattern **uuuuuu** (*year*)
    * Previously converted to **yyu** (not a valid jquery datepicker pattern)
    * Now converted to **yy** (*year (four digit)*)
* Pattern **zzzzz** (time zone name)
    * Previously converted to **zzzzz** (not a valid jquery timepicker pattern)
    * Now converted to **z** (*Timezone as defined by timezoneList*)

Would be great if somebody could look over this to make sure we don't break existing things.


melloware:
Thanks for including Excellent unit tests as well!

blutorange:
Yeah, unit tests are pretty much required for this kind of algorithm - I found a few bugs while writing the unit tests.

For the long term, if we want to support java.time.DateTimeFormatter more fully, we probably need to write a custom parse/format JavaScript implementation for p:datepicker based on the DateTimeFormatter algorithm. For now, this should fix the most glaring issues for p:calendar and p:datepicker

melloware:
Totally agreed it just seems that this is doing a great job, and trying to get java and javascript date formats to line up perfectly is like trying to fit a square peg in a round hole sometimes!  I think this fix will cover most reasonable use cases.

christophs78:
IMO this is an excellent PR. Do we need to wait for something until we merge it into master?

melloware:
I really like this PR was just waiting for @tandraschko to review since its a larger change.

tandraschko:
@blutorange couldnt we make the Tokens (e.g. YearToken) singeltons and pass repetitions to buildJqueryUiPattern? This would save objects and i currently dont see why we need to make instances each time?

also i would name it "JQuery" instead "Jquery"

tandraschko:
and whats with the old Old*Converter? Shouldnt we remove it?

blutorange:
@tandraschko 
* Yes, I should probably rename it to "JQuery", considering that's what I named another class.
* I moved the "OldDatePatternConverter" to the src/test folder (so it's not in the actual jar file) and it's there only for a test to ensure the new converter works the same way as the old one.
* A usual pattern has around 2~5 tokens, and these are lightweight Java objects. Also, I did some quick checks regarding performance and it was, if anything, a little bit better than the old one, and in any case pretty fast. I'm sure there's potential for optimizations, but that seemed like micro optimizations to me that wouldn't be worth it.

christophs78:
We should get this into master before we start with Release Candidates for 9.

melloware:
Agreed @blutorange do you need any help with this PR making the change @tandraschko requested?  I am sure one of us can help out if you are busy.

blutorange:
@melloware Yeah, I'm a bit busy right now, but I'll be having some spare time during the weekend, so I'll take a look at it, shouldn't take too long.

blutorange:
@tandraschko Should be good now. I changed the conversion to a visitor builder to avoid having to create token objects and renamed Jquery to JQuery. I left the Old*Converter classes for testing purposes, they are in src/test and not included in the final JAR.

melloware:
Nice refactor!

melloware (approved):

tandraschko:
Thanks!

melloware:
@blutorange can you check the new Code Smells introduced by this code?

See: https://sonarcloud.io/project/issues?id=org.primefaces%3Aprimefaces&resolved=false&sinceLeakPeriod=true&types=CODE_SMELL

I can update if you don't have time but there is one in particular with `&` vs `&&` that looks like a bug.

blutorange:
@melloware Yes, that `&` is indeed wrong. Looks like it should not take to long to update, I'll take a look

melloware:
I just submitted a PR if you want to review.

PR #6267, opened 2020-08-30
Fix #5978: FileDownload: add Ajax support

### LINKED ISSUE #5978: FileDownload: should not require ajax="false"
**Description**
FileDownload currently requires to set ajax to false for the commandButton or commandLink.
1. This is something you need to know. Or read in the documentation.
2. This causes issues when you use multiple forms within one XHTML. (Essentially you loose all input to other forms.)

Maybe there are way´s to solve this which had not been present 10 year ago, when PrimeFaces was built. IMO single page applications (Angular, React, Vue, ...) must have a way to do this without sending an attachment via Content-Disposition-header.

Found eg http://danml.com/download.html and https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/downloads/download

It may require some research how to combine this with the already exisiting code for FileDownload. Or how to integration this into the JSF/PrimeFaces-infrastructure.

FileDownload should work with ajax="true".

**Additional context**
see https://www.primefaces.org/showcase/ui/file/download.xhtml

jepsar:
Fix #5978

Please note that documentation (and showcase) is still to do. I'd like to wait on feedback first and then add the documentation.


melloware (commented):
This looks impressive

tandraschko:
Please dont change from index loop to for loop, thats by design for performance

jepsar:
> Please dont change from index loop to for loop, thats by design for performance

👍 Reverted.

christophs78:
(very) minor point: please add a entry to thirdpartylibs.md

tandraschko:
What are the exact changes in DynmicSourceContentBuilder?

tandraschko:
if its just the "component != null"; please dont refactor the whole class.

tandraschko:
also added the first performance hint (the indexloop stuff) here: https://github.com/primefaces/primefaces/wiki/Contributing-to-PrimeFaces

jepsar:
> What are the exact changes in DynmicSourceContentBuilder?
> if its just the "component != null"; please dont refactor the whole class.

@tandraschko I noticed that what we needed to download a file is very similar to the GraphicImage component: https://primefaces.github.io/primefaces/8_0/#/components/graphicimage?id=how-it-works

So I checked the renderer and found that I could almost use the `DynamicContentSrcBuilder` to create a URI to load in the JavaScript. The only thing we don't have in the `FileDownloadActionListener` was a component.

The component in the `DynamicContentSrcBuilder` is used to get the `ValueExpression` (we already have that) and to add parameters (we don't really need that). So I decided to add a constructor taking a `ValueExpression` and a component which can be null. That was the least drastic change I could think of.

I doubted to move the parameters part to a separate method to reduce indentation.

tandraschko:
i see
then please double check your code - you decode a null "src" in worst case because you "return" in the if statements

src/main/java/org/primefaces/util/DynamicContentSrcBuilder.java
```diff
@@ -65,36 +63,43 @@ public static String build(FacesContext context, Object value, UIComponent compo
...
-                    dynamicResourcesMapping = new LimitedSizeHashMap<>(200);
-                    session.put(Constants.DYNAMIC_RESOURCES_MAPPING, dynamicResourcesMapping);
-                }
-
-                ValueExpression expression = ValueExpressionAnalyzer.getExpression(
+            ValueExpression expression = ValueExpressionAnalyzer.getExpression(
                         context.getELContext(), component.getValueExpression(attributeName));
+            return build(context, streamedContent, component, cache, type, stream, expression);
+        }
+        return encodeResourceURL(context, src, cache);
```
jepsar: @tandraschko are you referring to this line? If so, src could be null originally:

https://github.com/primefaces/primefaces/blob/d05387b93bd5fc68d825bc5d325cc49472110cf5/src/main/java/org/primefaces/util/DynamicContentSrcBuilder.java#L127

Shall I change that into throwing a FacesException?

tandraschko:
i would remove src var and return null at the end of the method

jepsar:
> (very) minor point: please add a entry to thirdpartylibs.md

@christophs78 any idea what the first column should contain? Most (if not all) links are broken. I'll add the library without a link in the first column.

melloware:
@jepsar those links get generated by our JSDocs process so for your new code there will be nothing there yet.  So leave it "link less" like I do for FontAwesome

docs/9_0/core/thirdpartylibs.md
```diff
@@ -42,4 +42,4 @@ are many other third party libraries and plugins used to support features.
 | [Timeline](../jsdocs/classes/timeline.html) | 7.3.7 | https://github.com/visjs/vis-timeline |
 | [TouchSwipe](../jsdocs/modules/jquery.html#swipe) | 1.6.18 | https://github.com/mattbryson/TouchSwipe-Jquery-Plugin |
 | [WebcamJS](../jsdocs/modules/webcam.html) | 1.0.26 | http://github.com/jhuckaby |
-
+| download.js | 4.2 | http://danml.com/download.html |
```
melloware: I usually put these in alphabatical order to make them easier to find.

melloware (changes requested):

tandraschko:
did some small refactorings
thanks for the good PR!

a showcase example would be great!

jepsar:
> a showcase example would be great!

Sure, I was going to. I'll add documentation as well. 😃

PR #6268, opened 2020-08-30
Fix #6238: DatePicker: built in conversion for LocalDateTime does not support timeZone-attribute

### LINKED ISSUE #6238: DatePicker: built in conversion for LocalDateTime does not support timeZone-attribute
**Describe the defect**
When you bind DatePicker (or maybe Calendar) to LocalDateTime and set timeZone-Attribute this does not have any impact. This is because LocalDateTime is without a timezone. (Seems like i forgot this on some point in time.) IMO there´s no easy solution for this. The work-around / solution should be to use `<f:convertDateTime type="zonedDateTime" />`
(I run into this during looking into https://github.com/primefaces-extensions/primefaces-selenium/pull/22.)

**Reproducer**
http://primefaces.org/showcase/ui/input/datepicker/datePickerJava8.xhtml
Input a value into "Time (Timezone GMT+8)" and push submit.

**Thoughts**

- Should we add a note to the documentation of DatePicker and Calendar that timeZone-attribute won´t have an effect on LocalDateTime?
- Should we add a built-in converter for ZonedDateTime?
- ???

(On my day-job this is not an issue because our applications are only used within the default timezone.)

christophs78:


src/main/java/org/primefaces/component/calendar/BaseCalendarRenderer.java
```diff
@@ -157,6 +157,12 @@ public Object getConvertedValue(FacesContext context, UIComponent component, Obj
 
             // Java 8 Date/Time API
             if (Temporal.class.isAssignableFrom(type)) {
+                if (calendar.getTimeZone() != null) {
+                    FacesMessage message = new FacesMessage(FacesMessage.SEVERITY_ERROR,
+                            "timeZone-attribute is not supported for " + type.getName() + ". Use a explicit converter instead of the built-in.",
```
melloware: Slight change..
```
"'timeZone' attribute is not supported for " + type.getName() + ". Use an explicit converter instead of the built-in."
```

melloware (changes requested):

melloware (approved):

PR #6279, opened 2020-09-02
Fix #6274: DatePicker multiple months navigators

### LINKED ISSUE #6274: DatePicker: numberOfMonths > 1 should only show month + year navigator above the first month
It could be a personal thing, but I was a bit confused by the multiple navigators, and it seems that they all affect the first month only.

I would have expected navigation above the first month, and the other months just to have heading texts for month and year.

Screen Shot 2020-09-01 at 13 51 35

```
<p:datePicker monthNavigator="true"
              yearNavigator="true"
              numberOfMonths="3">
```

melloware:
Not too bad:
image


jepsar (approved):
Looking good to me! Thanks!

PR #6293, opened 2020-09-05
Fix #6277: DatePicker: added updateDisabledDates to widget

### LINKED ISSUE #6277: DatePicker: offer updating disabled dates and days through widgetVar
I'm using disabled dates based on a database query. I've chosen to only fetch the available dates based on the currently visible month. I was planning to use the `viewChange` listener to fetch available dates again when the view changes. The listener is called, but the component does not (re)read the disabled dates list. I tried updating the component, but that will cause the popup to close.

I would be nice if we could use something similar to the `LazyScheduleModel` and keep the number of requests limited.

jepsar:
Fix #6277

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
+++ b/src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
@@ -373,7 +373,13 @@ PrimeFaces.widget.DatePicker = PrimeFaces.widget.BaseWidget.extend({
      * @param {string[] | Date[]} disabledDates The dates to disable.
      */
     setDisabledDates: function(disabledDates) {
-        this.jq.data().primeDatePicker.options.disabledDates = disabledDates;
+        var pdp = this.jq.data().primeDatePicker;
+        pdp.options.disabledDates = disabledDates;
+        if (pdp.options.disabledDates) {
+            for (var i = 0; i < pdp.options.disabledDates.length; i++) {
+                pdp.options.disabledDates[i] = pdp.parseOptionValue(pdp.options.disabledDates[i]);
+            }
+        }
         this.updatePanel();
     },
 
@@ -391,11 +397,6 @@ PrimeFaces.widget.DatePicker = PrimeFaces.widget.BaseWidget.extend({
      */
     updatePanel: function() {
         var pdp = this.jq.data().primeDatePicker;
-        if (pdp.options.disabledDates) {
-            for (var i = 0; i < pdp.options.disabledDates.length; i++) {
-                pdp.options.disabledDates[i] = pdp.parseOptionValue(pdp.options.disabledDates[i]);
-            }
-        }
         pdp.panel.get(0).innerHTML = pdp.renderPanelElements();
     },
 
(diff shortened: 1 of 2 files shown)


src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
```diff
@@ -144,6 +144,17 @@ PrimeFaces.widget.DatePicker = PrimeFaces.widget.BaseWidget.extend({
         this.input.data(PrimeFaces.CLIENT_ID_DATA, this.id);
     },
 
+    updateDisabledDates: function(disabledDates) {
```
jepsar: .. or should we simply call this `setDisabledDates`?

melloware: I like `setDisabledDates`, can you also create a duplicate method called `setDisabledDays` that I think takes an array like [0,1] to turn off all Saturday's and Sundays if you look in the showcase and then refactor if they can share code?

melloware (changes requested):

src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
melloware: Also don't forget to JSDoc the method since its public and poeple will want to know what `disabledDates` parameter is.

melloware (changes requested):

melloware (approved):

src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
```diff
@@ -368,4 +368,35 @@ PrimeFaces.widget.DatePicker = PrimeFaces.widget.BaseWidget.extend({
...
+    /**
+     * Update panel.
+     */
+    updatePanel: function() {
+        var pdp = this.jq.data().primeDatePicker;
+        if (pdp.options.disabledDates) {
+            for (var i = 0; i < pdp.options.disabledDates.length; i++) {
+                pdp.options.disabledDates[i] = pdp.parseOptionValue(pdp.options.disabledDates[i]);
+            }
+        }
```
melloware: I tink this pdo.options for disabled dates has to get moved up to `setDisabledDates` method no?  It won't be used in the setDisabledDays() method.

melloware (changes requested):

src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
```diff
@@ -368,4 +368,35 @@ PrimeFaces.widget.DatePicker = PrimeFaces.widget.BaseWidget.extend({
...
+     */
+    setDisabledDays: function(disabledDays) {
+        this.jq.data().primeDatePicker.options.disabledDays = disabledDays;
+        this.updatePanel();
+    },
+
+    /**
+     * Update panel.
+     */
+    updatePanel: function() {
```
melloware, line 399: I wonder if this should be marked @private in the JSDocs?

melloware (changes requested):

melloware (approved):

PR #6295, opened 2020-09-06
Fix #5481: DataTable: removal of processEvent method

### LINKED ISSUE #5481: DataTable: removal of processEvent method
follow-up-issue to https://github.com/primefaces/primefaces/issues/5463

## 1) Environment
- Affected browsers: all

Only relevant for DataTable with Paging but without LazyDataModel.

## 2) Expected behavior
filterBy should be the "single source of truth" for filtering.

## 3) Actual behavior
DataTable materializes filteredValue as java.util.List<?>.
This list is used from `DataTable#processEvent`.
```
@Override
public void processEvent(ComponentSystemEvent event) throws AbortProcessingException {
	super.processEvent(event);
	if (!isLazy() && event instanceof PostRestoreStateEvent && (this == event.getComponent())) {
		Object filteredValue = getFilteredValue();
		if (filteredValue != null) {
			updateValue(filteredValue);
		}
	}
}
```
Sometimes filterBy and filteredValue get´s out of sync. (eg DataTable is bound to a SessionScoped bean) processEvent only seems to be called at Paging-events of DataTable.
Calling `updateValue(filteredValue);` does not seem to be necessary.

## 4) Steps to reproduce
https://www.primefaces.org/showcase/ui/data/datatable/tableState.xhtml
(multiViewState can be set to false, as it´s not relevant for this issue.)

- Select "Renault" as Brand
- Refresh
- Everything looks good
- Go on page 5: no result (NOK)
- Go back on page 1: filter Renault is back (NOK)

## 5) Sample XHTML
Showcase: ui/data/datatable/tableState.xhtml

## 6) Sample bean
Showcase: `org.primefaces.showcase.view.data.datatable.TableStateView` (must be `@SessionScoped` to reproduce this issue)

christophs78:
Removes code that was introduced with https://github.com/primefaces/primefaces/commit/20baf62631dd3662c978abb6cb57669940ca967d#diff-ae92f04fa2a452febc87c6833e245f56.

Currently can´t find a regression due to removing this code.

### DIFF
--- a/src/main/java/org/primefaces/component/datatable/DataTable.java
+++ b/src/main/java/org/primefaces/component/datatable/DataTable.java
@@ -1374,11 +1374,6 @@ public class DataTable extends DataTableBase {
         return iterableChildren;
     }
 
-    @Override
-    public void processEvent(ComponentSystemEvent event) throws AbortProcessingException {
-        super.processEvent(event);
-    }
-
     public void updateFilteredValue(FacesContext context, List<?> value) {
         ValueExpression ve = getValueExpression(PropertyKeys.filteredValue.toString());
 
--- a/src/main/java/org/primefaces/component/datatable/DataTable.java
+++ b/src/main/java/org/primefaces/component/datatable/DataTable.java
@@ -1377,12 +1377,6 @@ public class DataTable extends DataTableBase {
     @Override
     public void processEvent(ComponentSystemEvent event) throws AbortProcessingException {
         super.processEvent(event);
-        if (!isLazy() && event instanceof PostRestoreStateEvent && (this == event.getComponent())) {
-            Object filteredValue = getFilteredValue();
-            if (filteredValue != null) {
-                updateValue(filteredValue);
-            }
-        }
     }
 
     public void updateFilteredValue(FacesContext context, List<?> value) {


Rapster:
You can delete the whole method since it's just doing a super call. Also, I would have a look about this `SelectableDataModelWrapper` I'm not sure what it fixes either, it might be related to something with lazy model and filtering, see commit a92e580ac1d4579661653f95001690d154e20851

christophs78:
> You can delete the whole method since it's just doing a super call. Also, I would have a look about this `SelectableDataModelWrapper` I'm not sure what it fixes either, it might be related to something with lazy model and filtering, see commit [a92e580](https://github.com/primefaces/primefaces/commit/a92e580ac1d4579661653f95001690d154e20851)

This is something you already tried in https://github.com/primefaces/primefaces/pull/6242? You think we should check to remove `SelectableDataModelWrapper` because it may not be necessary?

Maybe i should add some more integration-tests which test LazyDataModel together with filtering, sorting, and selection before.

melloware (approved):

tandraschko:
-0.5 until we find at least a reason why cagatay introduced it

melloware:
Possibly: 
https://code.google.com/archive/p/primefaces/issues/7122
https://code.google.com/archive/p/primefaces/issues/7132

christophs78:
https://code.google.com/archive/p/primefaces/issues/7122 seems to be fixed 
https://code.google.com/archive/p/primefaces/issues/7132 still seems to be there (or was reintroduced sometime between 2014 and 2020)

look at https://github.com/primefaces-extensions/primefaces-integration-tests/pull/34

christophs78:
IMO we should write as much integration tests as necessary. After this we decide what we do with this PR and https://github.com/primefaces/primefaces/pull/6297.

IMO this is the only way to make fact based decisions.

melloware:
I agree your regression tests have already come in handy finding regression issues

christophs78:
Which scenario´s should we test which are not yet covered by our integrations tests?
When we look at the code it must be scenarios where DataTable is bound to a List (not a LazyDataModel) and Filter is used.

tandraschko:
whats with this PR @christophs78 @Rapster ?

Rapster:
I managed to find at least one problem by removing this method. Simply update datatable after filtering, the filtered list gets lost.
```xml
<p:commandButton process="@none" value="Update" update="@next" />
<p:dataTable var="car"
             value="#{dtFilterView.cars1}"
             widgetVar="carsTable1"
             emptyMessage="No cars found with given criteria"
             filteredValue="#{dtFilterView.filteredCars1}">
   ...
</p:dataTable>

```

Though, I managed to make it work by removing processEvent method and removing the `isFilteringEnabled` condition in `DataTable#saveState` method.

christophs78:
I´ll try to add a integration-test for your usecase later this week.

Issue #1442, opened 2016-05-20
DataTable: Rowedit filtered column , updates wrong row


anunayak81:
Hi,

I have 2 applications, one of which uses PF 3.2 and another uses PF 5.1. In both applications, I am facing an issue with rowedit of a datatable with filters.

Several screens in my applications, have Editable data-tables with filter on every column. Whenever, user filters by a particular column and tries to edit the same column in the matching rows, it actually updates some other rows. 

So for example, let us assume my datatable (5 rows per page) has a column Employee ID with ids A001, A002, A003, A004, A005, B006, B007, B008, B009, B010. It has another column for employee name. So as per the pagination , it shows all employee IDS starting with A in first page and the Bs in page 2.  Now if I filter by employee ID with filter value "B0", it shows all employee Ids starting with B. Now if I edit the employee ids starting with B to start with C (So B006 becomes C006 and so on) and then save each of them, you can see that it actually updates the A record (A001) as the underlying id is actually that of page1 though the filter has been effective. 

The way this works seems to be quite strange and is causing a lot of issues in our application. Please advice.

Thanks,
Anu

dkellenb:
We face the same problem with PF 6.0. Changing values on a filtered or sorted data table, will update the model using the index in the filterer / sorted model. There are SO questions on this with no answer:
- http://stackoverflow.com/questions/33527471/primefaces-datatable-rendering-list-wrong-after-listener-sort-bug
- http://stackoverflow.com/questions/29544230/selection-goes-wrong-after-default-sorting-in-primefaces-datatable
- http://stackoverflow.com/questions/24754118/after-sorting-it-deletes-the-wrong-line-of-primefaces-datatable

MaurerThomas:
I am also facing the same problem on PF 6.0. Any updates regarding this issue?

melloware:
Integration test added:
https://github.com/primefaces-extensions/primefaces-integration-tests/blob/b9de2d127c975c4bfcf5efe547a62aaf4dac352f/src/test/java/org/primefaces/extensions/integrationtests/datatable/DataTable007Test.java#L95-L136

melloware:
as of PF 8.0 this is no longer an issue.  Not sure where along the lines it got fixed but the Integation test above proves its working.

PR #6297, opened 2020-09-06
#6296: DataTable: evaluate removal of SelectableDataModelWrapper

### LINKED ISSUE #6296: DataTable: evaluate removal of SelectableDataModelWrapper
follow-up to https://github.com/primefaces/primefaces/pull/6295

originally introduced with https://github.com/primefaces/primefaces/commit/a92e580ac1d4579661653f95001690d154e20851

christophs78:
@rapster: With our current integration-tests i can´t find a regression. But IMO we should add some more integration-tests before we merge this one.


melloware:
So if I am reading this right the table has to be "selectable" and "filterable" at the same time?

christophs78:
https://github.com/primefaces-extensions/primefaces-integration-tests/pull/34

melloware:
Do you think this will fix this also: https://github.com/primefaces/primefaces/issues/1442

christophs78:
> Do you think this will fix this also: #1442

Sound´s like i should add integration-tests for the combination of filter (and maybe sort) combined with rowedit. I am not sure without checking. 
We all know DataTable is quite complicated and it´s realy hard to tell sometimes why which code was introduced. Does some code still makes sense...

tandraschko:
whats with this PR @Rapster @christophs78 ?

PR #6299, opened 2020-09-07
added new function to obtain widgets by type

matinh:
For details see forum topic at https://forum.primefaces.org/viewtopic.php?f=3&t=47792


tandraschko:
please also add it to the documentation and create a issue for tracking

matinh:
I created JavaDoc style documentation. Where else should I add documentation?

tandraschko:
the userguide (https://primefaces.github.io/primefaces/9_0/#/), its also on the sources

matinh:
Ok, I found I have to update https://github.com/primefaces/primefaces/blob/master/docs/9_0/core/javascriptAPI.md.
Do I also have to update `*.html` files in `docs/jsdocs` or are they generated somehow?
And what about the `export` in `PrimeFaces.d.ts`?

I didn't find this info on https://github.com/primefaces/primefaces/wiki/Contributing-to-PrimeFaces. Maybe it should be added to this wiki page?

Thanks for your help.

tandraschko:
they are generated automatically via our travis task

tandraschko:
thanks!

PR #6305, opened 2020-09-08
Fix #6301: DataTable encode footer last

### LINKED ISSUE #6301: DataTable: body should be rendered before footer
https://github.com/primefaces/primefaces/blob/cb796a4e7e0be29e715110e7fcfa6a128eddf2a8/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java#L444-L446

The main issue with this is when you have focusable elements in the body and the footer, the tab order is wrong.

I'm aware that the datatable is a complex component, so probably it will take more then just swapping the lines above. I'll have some time later this week to swap the lines and check with the showcase if all the examples still work.

melloware:


jepsar (approved):

tandraschko:
+1 if well tested

melloware:
I have tested the Showcase examples. I am going to merge this and run our PF Integration Tests which we have 15 datatable tests so far and will verify nothing is broken.

melloware:
All integration tests passing after latest commits!  https://travis-ci.com/github/primefaces-extensions/primefaces-integration-tests/builds/184671056

PR #6311, opened 2020-09-10
Fix #5326,#1499,#1661: Datatable rowReorder AJAX update

### LINKED ISSUE #1499: DataTable: saving edited row after reorder (drag n' drop) updates wrong row
When `p:dataTable` is used with attributes `draggableRows="true"` and `editable="true"` and user tries to reorder row by dragging it when row is in editing mode - after saving, row in old position gets changed, and not the dragged one.

Strangely `event.getObject()` on `rowEdit` event returns correct Object (dragged one), which is unedited, because values are overwritten to Object in old position.

This editing could work correctly or dragging could be disabled when rows are edited.

Primefaces 6.0

### LINKED ISSUE #1661: DraggableRow with Datatable CellEdit Bug - Unable to display the correct data in output mode
**The output mode always displays the old data after triggering the draggable row feature**

Example code:

**draggableTable.xhtml**

``` xhtml
    xmlns:fn="http://java.sun.com/jsp/jstl/functions">

<h:head>

</h:head>

<h:body>
    <h:form id="form">
        <p:growl id="msgs" showDetail="true" />

        <p:dataTable value="#{editableTableBean.persons}" var="person" draggableRows="true"
            editable="true" editMode="cell">
            <f:facet name="header">
                Draggable Row with Celledit Bug Report
            </f:facet>

            <p:column headerText="Name">

                <p:cellEditor>
                    <f:facet name="output">
                        <p:outputLabel value="#{person.name}"></p:outputLabel>
                    </f:facet>

                    <f:facet name="input">
                        <p:inputText value="#{person.name}"></p:inputText>
                    </f:facet>
                </p:cellEditor>
            </p:column>

            <p:column headerText="Date of Birth">
                <p:cellEditor>
                    <f:facet name="output">
                        <p:outputLabel value="#{person.dateOfBirth}">
                            <f:convertDateTime pattern="dd MMM, yyyy"/>
                        </p:outputLabel>
                    </f:facet>
                    <f:facet name="input">
                        <p:calendar value="#{person.dateOfBirth}" 
                                    pattern="dd MMM, yyyy"
                                    navigator="true">
                        </p:calendar>
                    </f:facet>
                </p:cellEditor>
            </p:column>
        </p:dataTable>
    </h:form>
</h:body>
</html>
```

**EditableTableBean.java**

``` java
package org.primefaces.showcase.edit;

import java.io.Serializable;
import java.text.ParseException;
import java.text.SimpleDateFormat;
import java.util.ArrayList;
import java.util.List;

@ManagedBean(name="editableTableBean")
@ViewScoped
public class EditableTableBean implements Serializable {

    private static final long serialVersionUID = -4217194079931366514L;
    private List<Person> persons = new ArrayList<Person>();

    @PostConstruct
    public void init() {
        SimpleDateFormat sdf = new SimpleDateFormat("dd/MM/yyyy");

        Person person1;
        Person person2;
        try {
            person1 = new Person("John", sdf.parse("01/01/1979"));
            person2 = new Person("Rose", sdf.parse("01/01/1969"));

            this.getPersons().add(person1);
            this.getPersons().add(person2);
        } catch (ParseException e) {
            e.printStackTrace();
        }
    }

    public List<Person> getPersons() {
        return persons;
    }

    public void setPersons(List<Person> persons) {
        this.persons = persons;
    }


}

```

**Person.java**

``` java
package org.primefaces.showcase.edit;

import java.io.Serializable;
import java.util.Date;

public class Person implements Serializable {


    private static final long serialVersionUID = 7247747818315769598L;
    private String name;
    private Date dateOfBirth;

    public Person(String name, Date dateOfBirth) {
        super();
        this.name = name;
        this.dateOfBirth = dateOfBirth;
    }

    public String getName() {
        return name;
    }

        this.name = name;
    }

    public Date getDateOfBirth() {
        return dateOfBirth;
    }

    public void setDateOfBirth(Date dateOfBirth) {
        this.dateOfBirth = dateOfBirth;
    }

}

```

Let's run the exmaple code:

snapshot2

Before dragging the row, I was able to set the data and to display the result correctly.

For example, if I want to set Rose's birthday to Jan 1st, 1970, the result is correct.

snapshot2

snapshot2

However, if I drag the row first before editing the row, the result is incorrect.

snapshot2

snapshot2

**The output data is different from the input data**

Tested on Primefaces v5.3 / Primefaces v5.3.13 / Primefaces v6.1 SNAPSHOT (with the latest commit of : 719e5584d1a60b9359972d3979cc2b3b85be464d )
Browser: latest Chrome and Firefox
OS: Linux
No errors at all on Web Server's console and Browser's console

### LINKED ISSUE #5326: DataTable: draggableRows -> wrong value sent to view after reordering
## 1) Environment
- PrimeFaces version: 7.0.9
- Application server + version: WildFly 10.1.0
- Affected browsers: All

## 2) Expected behavior
After dragging and reordering rows in a DataTable the currently displayed value is sent to the view by an action.

## 3) Actual behavior
After dragging and reordering rows in a DataTable the value previously at this position is sent to the view by an action.

image

## 4) Steps to reproduce
Click on the commandButtons to confirm that "1" displays the growl message for TestValue with id 1 and so on. Reorder a row. Click on the commandButtons to confirm that the growl message displays the TestValues still in the 1,2,3 order and not in the order displayed after reordering the rows.

## 5) Sample XHTML
```
<!DOCTYPE html>

<h:head>
</h:head>
<h:body>

    <h1>#{testView.testString}</h1>

    <h:form>

        <p:dataTable value="#{testView.testValues}"
                     draggableRows="true"
                     var="testValue">

            <p:column>
                <h:outputText value="#{testValue.id}: "/>
                <p:commandButton action="#{testView.testAction(testValue)}"/>
            </p:column>

        </p:dataTable>

        <p:growl life="8000">
            <p:autoUpdate/>
        </p:growl>

    </h:form>

</h:body>
</html>
```

## 6) Sample bean
```
package org.primefaces.test;

import javax.inject.Named;
import java.io.Serializable;
import java.util.ArrayList;
import java.util.List;

@Named
@ViewScoped

    private String testString;
    private List<TestValue> testValues;

    @PostConstruct
    public void init() {
        testString = "Welcome to PrimeFaces!!!";

        testValues = new ArrayList<>();
        testValues.add(new TestValue(1));
        testValues.add(new TestValue(2));
        testValues.add(new TestValue(3));
    }

    public void testAction(TestValue testValue) {
        FacesContext.getCurrentInstance().addMessage("id", new FacesMessage(testValue.toString()));
    }

        return testString;
    }

        this.testString = testString;
    }

    public List<TestValue> getTestValues() {
        return testValues;
    }

    public void setTestValues(List<TestValue> testValues) {
        this.testValues = testValues;
    }

    public class TestValue {
        private int id;

        public TestValue(int id) {
            this.id = id;
        }

        public int getId() {
            return id;
        }

        @Override
        public String toString() {
            return "TestValue{" +
                    "id=" + id +
                    '}';
        }
    }
}
```

Please see the attached test project, cloned from the primefaces-test.git.
[datatable-draggablerows-test.zip](https://github.com/primefaces/primefaces/files/3839991/datatable-draggablerows-test.zip)

melloware:


tandraschko:
will this update the whole datatable? wouldnt be a great fix

melloware:
It does and if you look almost all the other Datatable AJAX events update=this. Events such as `addRow` `liveScroll` `virtualScroll` `page` `filter` all have update=@this.  But its missing on RowReorder and you can see from the reproducer it clearly causes the issues.

tandraschko:
but whats the real issue here?

The reorder is done via JS and finished on the client actually. The post is just required to synchronize it to the server side.
So just adding a update fixes it but its not like we should do it.

melloware:
So what happens in all 3 tickets is when you rowReoder the state of the table rows is clearly wrong.  So for example
Row1
Row2
Row3

if you click button in Row1 it says "Row 1 clicked".  Now if you reorder those rows...
Row3
Row2
Row1

So Row1 is on the bottom and click Row1 button it says "Row 2 selected".   To me this is bad and unexpected behavior.  and its been reported 3 times over the years and again this morning by @rider87 on the forums.

tandraschko:
ok but why exactly? does the row have a wrong rowkey data attribute then? 
The fix may just be a small JS statement instead a fat datatable update.

melloware:
Well it could be the entire table rowKeys are off?  But from this ticket https://github.com/primefaces/primefaces/issues/1661 if you are using CellEditing all of the CellEditors are showing incorrectly when you click on Edit Row after re-ordering..  So to me it feels like its more than just RowKey is off but I could be wrong?

tandraschko:
maybe, yes
we should check it in detail before merge this

melloware:
yep and @rider87 is provider another reproducer of the issue he is having with rowReordering and I want to review that once he provides it.

melloware:
Another thing by making this the default behavior it solves more problems than it creates and users can always go back to the original behavior with...
```xml
<p:ajax event="rowReorder" update="@none"/>
```

christophs78:
Just had a call with a colleague which also run into https://github.com/primefaces/primefaces/issues/1499.
(rowReorder combined with rowEditor)
Currently you can drag around rows which are switched to edit-mode. But after dragging them around rowEdit-save updates the wrong row.

With your update we probably will simply loose the edit-state and all input´s in the edited row.

IMO one option would be to throw a Exception when rowReorder and rowEditor are enabled together and tell the user that this simply does not work. Or we fix the root cause.
(Our old (and a bit sad story) off all the PF-features which not allways can be combinded together.)

Maybe another option would be to temporarily disable rowReorder when at least one row is in edit-mode.

melloware:
I feel like if the table is in edit mode we should disable re-ordering and once editing is done reinstate row ordering.  That to me see like reasonable behavior.

melloware:
OK I just updated the PR to disable row reordering while you are editing and re-enable it after you cancel or save. It works great!

christophs78:
Multiple rows may be in edit mode. We can re-enable reorder only if no row is in edit mode.

melloware:
OK updated to check it will only re-enable row reorder if not more rows are currently being edited.

christophs78:
There seems to be a timing issue.

Just added some console.log
```
// #1499 enable rowReorder when done editing
console.log("rows in editmode: " + $('tr.ui-row-editing').length);
if (this.cfg.draggableRows && $('tr.ui-row-editing').length === 0) {
	console.log("enable rowReorder");
	this.tbody.sortable("enable");
}
```

Output on console is:
image

melloware:
Try now I moved it inside the success event of the AJAX request.

christophs78:
Following seems to be the right position:
```javascript
onsuccess: function(responseXML, status, xhr) {
			widget: $this,
			handle: function(content) {
				if(expanded) {
					this.collapseRow(row);
				}

				this.updateRow(row, content);

				// #1499 enable rowReorder when done editing
				if (this.cfg.draggableRows && $('tr.ui-row-editing').length === 0) {
					this.tbody.sortable("enable");
				}
			}
		});

	return true;
},
```
Maybe some parts run async or something like this.

melloware:
Perfect fixed

christophs78:
Thanks for your work!
(I don´t want to think how this all works combined with paging, sorting, filtering ... Somewhere we need to stop. At least for now.)

melloware:
Yeah but I think this is a good improvement.  You should not be able to reorder rows while editing so I feel like this definitely makes more sense from a usability perspective!  Thanks for your help!

tandraschko:
for me its ok to disable it but why do we still need the update? if the update is required, we have a another bugs somewhere.

melloware:
I reported this to my PRO account so @mertsincan is reviewing.

tandraschko:
if you remove the update=@this for now, +1 from my side to apply it

melloware:
After discussing with @mertsincan I have removed the update per @tandraschko request.

Issue #4573, opened 2019-02-22
Add resetRows method to dataTable to reset rows attribute


mertsincan:
Reported By PRO User;
> Use case:
The number of rows the datatable should display may change on Ajax request, because we have a value expression bound to the rows property. So it is necessary that the bound EL expression gets recalculated. However, the value of the rows property is stored in the StateHelper. Once a value has been set into the StateHelper, it will be returned from there. Only if it is null, then the bound EL expression will be re-evaluated, see javax.faces.component.ComponentStateHelper.eval(Serializable, Object):
```
public Object eval(Serializable key, Object defaultValue) {
        Object retVal = get(key);
        if (retVal == null) {
            ValueExpression ve = component.getValueExpression(key.toString());
            if (ve != null) {
                retVal = ve.getValue(component.getFacesContext().getELContext());
            }

        }

        return ((retVal != null) ? retVal : defaultValue);
    }
```
> So, we intended to reset it to null so the EL expression gets recalculated.

PR #6312, opened 2020-09-10
Fix #5465: DataTable: reset-method does not reset rows-(per-page-)attribute

### LINKED ISSUE #5465: DataTable: reset-method does not reset rows-(per-page-)attribute
Think reset should call resetRows which was added with https://github.com/primefaces/primefaces/issues/4573

christophs78:


src/main/java/org/primefaces/component/api/UIData.java
```diff
@@ -259,7 +267,10 @@ public void setRowStatePreserved(boolean _paginator) {
     }
 
     public void resetRows() {
-        getStateHelper().remove(PropertyKeys.rows);
+        Object rows = getStateHelper().eval(PropertyKeys.internal_rowsInitialValue);
+        if (rows != null) {
+            setRows((int) rows);
+        }
```
christophs78: Need some more time to think about this, because we should not break https://github.com/primefaces/primefaces/issues/4573.

christophs78: IMO resetRows as it is up to this PR does only work when rows is bound to a value expression.
It does not work when the attribute-value is written direct into the xhtml. (rows="5")

christophs78:
Hope, i finally figured out whats right for `rows="5"` and for `rows="#{someBean.someProperty}"`.

melloware:
Looks good to me but i will let @tandraschko sine he was just in here fixing it.

tandraschko:
WDYT about creating some like InternalPropertyKeys (with protected modifier) for it?

christophs78:
> WDYT about creating some like InternalPropertyKeys (with protected modifier) for it?

Good idea! Done.

melloware (approved):

PR #6314, opened 2020-09-11
Fix #6313: Schedule fire viewChange for prev,next,today

### LINKED ISSUE #6313: Schedule: ajax behavior event for prev, next, today buttons
With each click on the previous, next, or today i want to re-render the p:schedule component because in my use case for instance the minTime/maxTime may change over the different days.

I have seen that currently the following five ajax behavior event are supported:

dateSelect | org.primefaces.event.SelectEvent | When a date is selected.
eventSelect | org.primefaces.event.SelectEvent | When an event is selected.
eventMove | org.primefaces.event.ScheduleEntryMoveEvent | When an event is moved.
eventResize | org.primefaces.event.ScheduleEntryResizeEvent | When an event is resized.
viewChange | org.primefaces.event.SelectEvent | When a view is changed.

It would be extremely needed to add at least one more event, which is triggered whenever prev, next, or today is clicked.

Or is there already a possibility implemented to realize this behavior?

Many thanks in advance.

melloware:


tandraschko:
looks good from my side but someone of the schedule contributors should review it

melloware:
@christophs78 @blutorange do you guys mind reviewing this?  I actually discovered this solution on stack overflow and have tested it and it works great.

blutorange:
@melloware Would be great if FullCalendar has a built-in event for this, but there doesn't seem to be one. Well, there's [viewRender](https://fullcalendar.io/docs/v3/viewRender), but it doesn't contain info on whether prev/next was clicked. So this solution seems like the best option to me for now.

One thing that I might change is to use a delegated event listener instead of registering a listener on the buttons directly. FullCalendar uses React, and I wouldn't count on the buttons never getting re-rendered. Something like:

```javascript
this.jq.on("click", '.fc-prev-button, .fc-next-button, .fc-today-button', function() { .. });
```

melloware:
Agreed there is no native support for these events so this is how people on StackOverflow were doing it and it works.

I just tried your delegated event model above and it works for the prev/next buttons but the Today button no longer fires the event. I think it might be related to this answer: https://stackoverflow.com/a/30662011/502366

So for now I will leave as is as it appears to work and continues to work after updating.

PR #6320, opened 2020-09-15
Fix #6319: Slider allow negative sign if min<0

### LINKED ISSUE #6319: Slider: prevents enter "-" in associated <p:spinner> component
1) Environment

PrimeFaces version: 8.0

2) Expected behavior

Allow enter of "-" in associated input if `minValue` is less then 0

3) Actual behavior

Slider prevents enter "-" in associated input.

probably skipped by this line:

https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/slider/slider.js#L127

4) Example XHTML

```xml
  <p:spinner value="#{bean.value}"
        min="-10"
        max="10"
  />
  <p:slider for="@previous" 
        minValue="-10"
        maxValue="10"
  />
```

melloware:


tandraschko:
+1

PR #6326, opened 2020-09-17
Fix #4446: Only process loaded (clientside) row expansions

### LINKED ISSUE #4446: SelectOneMenu: RowExpansion value lost
This issue is similar to one I already reported and fixed but now for select one menu: https://github.com/primefaces/primefaces/issues/4138

If the SelectOneMenu is inside a RowExpension and the row is closed the submitted value will be lost on ajax process.

## 1) Environment
- PrimeFaces version: 6.2
- Application server + version: Wildfly-11
- Affected browsers: ALL

## 2) Expected behavior
The value should be submitted.

## 3) Actual behavior
The submitted value will be changed to default ("")

## 4) Steps to reproduce
- Select something from an SelectOneMenu which is inside a RowExpension.
- Close the RowExpension.
- Process the form.
- Open the RowExpension.
- The SelectOneMenu will be empty.

fanste:


melloware:
As mentioned in the ticket we should test with sorting, filtering, and paging to make sure its still working properly.

src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
```diff
@@ -4561,6 +4563,32 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
...
+    writeRowExpansions: function() {
+    	this.expansionHolder.val(this.loadedExpansionRows.join(','));
+    },
+
+    rowExpansionLoaded: function(rowIndex) {
+    	if (!PrimeFaces.inArray(this.loadedExpansionRows, rowIndex)) {
+    		this.loadedExpansionRows.push(rowIndex);
+    		this.writeRowExpansions();
+    	}
+    },
```
melloware, line 4594: Looks like your formatting spacing here in JavaScript is a little off compared to rest of the file?

melloware (changes requested):

melloware (approved):

tandraschko:
the code looks ok

melloware:
@tandraschko Its been tested by @fanste @rider87 and myself.  Think this is safe to merge??

PR #6339, opened 2020-09-23
implement #6337: DataGrid: add rowStyle and rowStyleClass-attributes

### LINKED ISSUE #6337: DataGrid: add rowStyle and rowStyleClass-attributes
This would eg allow to switch the amount of columns in grid-view depending on the screen-width.

similar to https://github.com/primefaces/primefaces/issues/6336

christophs78:
Showcase-PR: https://github.com/primefaces/primefaces-showcase/pull/193

### DIFF
--- a/src/main/java/org/primefaces/component/datagrid/DataGridRenderer.java
+++ b/src/main/java/org/primefaces/component/datagrid/DataGridRenderer.java
@@ -171,8 +171,12 @@ public class DataGridRenderer extends DataRenderer {
         int itemsToRender = rows != 0 ? rows : grid.getRowCount();
         int numberOfRowsToRender = (itemsToRender + columns - 1) / columns;
         int displayedItemsToRender = rowIndex + itemsToRender;
-        String columnClass = DataGrid.COLUMN_CLASS + " " + GridLayoutUtils.getColumnClass(columns) + " " + grid.getRowStyleClass();
+        String columnClass = DataGrid.COLUMN_CLASS + " " + GridLayoutUtils.getColumnClass(columns);
         String columnInlineStyle = grid.getRowStyle();
+
+        if (!LangUtils.isValueBlank(grid.getRowStyleClass())) {
+            columnClass += " " + grid.getRowStyleClass();
+        }
 
         writer.startElement("div", null);
         writer.writeAttribute("class", DataGrid.GRID_ROW_CLASS, null);
--- a/src/main/java/org/primefaces/component/datagrid/DataGridRenderer.java
+++ b/src/main/java/org/primefaces/component/datagrid/DataGridRenderer.java
@@ -33,6 +33,7 @@ import javax.faces.context.ResponseWriter;
 import org.primefaces.renderkit.DataRenderer;
 import org.primefaces.util.ComponentUtils;
 import org.primefaces.util.GridLayoutUtils;
+import org.primefaces.util.LangUtils;
 import org.primefaces.util.WidgetBuilder;
 
 public class DataGridRenderer extends DataRenderer {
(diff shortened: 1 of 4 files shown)


src/main/java/org/primefaces/component/datagrid/DataGridRenderer.java
```diff
@@ -170,20 +171,24 @@ protected void encodeGrid(FacesContext context, DataGrid grid) throws IOExceptio
         int itemsToRender = rows != 0 ? rows : grid.getRowCount();
         int numberOfRowsToRender = (itemsToRender + columns - 1) / columns;
         int displayedItemsToRender = rowIndex + itemsToRender;
-        String columnClass = DataGrid.COLUMN_CLASS + " " + GridLayoutUtils.getColumnClass(columns);
+        String columnClass = DataGrid.COLUMN_CLASS + " " + GridLayoutUtils.getColumnClass(columns) + " " + grid.getRowStyleClass();
+        String columnInlineStyle = grid.getRowStyle();
+
+        writer.startElement("div", null);
+        writer.writeAttribute("class", DataGrid.GRID_ROW_CLASS, null);
```
melloware, line 182: By moving this outside the loop doesn't that mean only 1 row will ever be created?  this looks like it was meant to create a DIV per row?

melloware (changes requested):

src/main/java/org/primefaces/component/datagrid/DataGridRenderer.java
christophs78: This IMO was necessary for a very old grid system before PrimeFaces 6.0 introduced Grid CSS.
IMO DataGrid was adapted to Grid CSS at some point, but this was never cleaned up.

christophs78: ... and i had to modify this because with ui-g over each row there was no way to add additional responsiveness.
Check this against the showcase. :-)

melloware: perfect

melloware (approved):

src/main/java/org/primefaces/component/datagrid/DataGridRenderer.java
```diff
@@ -170,20 +171,24 @@ protected void encodeGrid(FacesContext context, DataGrid grid) throws IOExceptio
         int itemsToRender = rows != 0 ? rows : grid.getRowCount();
         int numberOfRowsToRender = (itemsToRender + columns - 1) / columns;
         int displayedItemsToRender = rowIndex + itemsToRender;
-        String columnClass = DataGrid.COLUMN_CLASS + " " + GridLayoutUtils.getColumnClass(columns);
+        String columnClass = DataGrid.COLUMN_CLASS + " " + GridLayoutUtils.getColumnClass(columns) + " " + grid.getRowStyleClass();
```
melloware: sorry one more thing will `+ grid.getRowStyleClass` append the word "null" if its not set?  typically we check for `!LangUtils.isvalueblank` before appending an optional style class?

melloware (changes requested):

src/main/java/org/primefaces/component/datagrid/DataGridRenderer.java
christophs78: Thanks for pointing me on this!

src/main/java/org/primefaces/component/datagrid/DataGridRenderer.java
```diff
@@ -171,19 +172,27 @@ protected void encodeGrid(FacesContext context, DataGrid grid) throws IOExceptio
         int numberOfRowsToRender = (itemsToRender + columns - 1) / columns;
         int displayedItemsToRender = rowIndex + itemsToRender;
         String columnClass = DataGrid.COLUMN_CLASS + " " + GridLayoutUtils.getColumnClass(columns);
+        String columnInlineStyle = grid.getRowStyle();
+
+        if (!LangUtils.isValueBlank(grid.getRowStyleClass())) {
+            columnClass += " " + grid.getRowStyleClass();
+        }
```
melloware, line 179: Maybe I am not seeing it but I don't see where you are using the columnClass variable?  I don't see you setting it anywhere?

melloware (changes requested):

melloware (approved):
Nevermind answered my own question.

PR #6347, opened 2020-09-25
Fix #4928: BreadCrumb: accessibility

### LINKED ISSUE #4928: BreadCrumb: Improve accessibility
## 1) Environment
- Application server + version: **-**
- Affected browsers: **-**

## 2) Expected behavior

According to the ARIA best practices site (https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html) a breadcrumb should be:
- rendered as ordered list
- use no additional aria roles
- hide seperators
- set `aria-current="page"` for the last link

## 3) Actual behavior

The breadcrumbs are an unordered list using `menu` aria roles.

---
I fixed that for my project using a custom renderer and can provide a pull request if you think that this should be changed.

jepsar:
Fix #4928: BreadCrumb: accessibility


melloware:
@jepsar Can you do me a favor and just test this Nova Light theme out locally by dropping this CSS into PF dir.

[nova-light.zip](https://github.com/primefaces/primefaces/files/5283598/nova-light.zip)


I want to see if I made it forward and backward compatible. I basically changed the SCSS from..
```css
.ui-breadcrumb {
        ul {
```

to

```css
.ui-breadcrumb {
        ul, ol {
```

melloware (approved):

jepsar:
I'll have a look tomorrow.

The separator (chevron) also needs to be added in the style sheet.

melloware:
OK here is the full SCSS for breacrumb...
```css
    .ui-breadcrumb {
        background-color: $panelContentBgColor;
        border: $panelContentBorder;
        padding: $panelContentPadding;
        
        ul, ol {
            li {
                float: none;
                display: inline-block;
                vertical-align: middle;
                
                .ui-menuitem-link {
                    color: $textColor;
                    margin: 0;
                    float: none;
                    padding: $menuitemPadding;

                    &:focus {
                        background-color: $menuitemHoverBgColor;
                        outline: 0 none;
                    }
                    
                    > span {
                        display: inline;
                        margin-right: $inlineSpacing / 2;
                    }
                }
                
                &.ui-breadcrumb-chevron {
                    @include icon_override("\e901");
                    font-size: $fontSize + 2;
                    margin-top: 2px;
                    color: $textSecondaryColor;
                }
                
                &:first-child {
                    a.ui-icon-home {
                        @include icon_override("\e925");
                        color: $textSecondaryColor;
                        margin: 0 $inlineSpacing 0 0;
                        font-size: $fontSize + 4;
                        vertical-align: top;
                        
                        span {
                            display: none;
                        }
                    }
                }
            }
        }
    }
```

tandraschko:
We should talk with Primetek before

melloware:
Agreed @mertsincan would have to make the theme changes but i just want @jepsar to verify my CSS changes are correct before I send them to him.  He would have to update the themes correctly.  But this change by default keeps things the way they are and you have to turn it on with accesible="true" so this way its backwards compatible.

jepsar:
@melloware I don't think your CSS fixes the chevron as it is no longer an element in the HTML. I added some CSS to a draft PR https://github.com/primefaces/primefaces-showcase/pull/198

tandraschko:
pinged mert last month about it, lets see

tandraschko:
IMO we should not add a new attribute but always render the new markup if the theme was adjusted. Mert should decide here, i just dont think that 2 different markups are a good way to go.

mertsincan:
Yes, I agree with Thomas. It can always be accessible. That's why we can use the ol tag. Also if you create a new Github issue with the "theme" label and 9.0 milestone I can add new styles to the designer API for breadcrumb. Thank you very much for this nice work.

Also, could @jepsar please remove "Note: currently this mode is not styled by default yet!" text? ;)

jepsar:
@mertsincan I've removed the switch attribute. For a demo, see https://github.com/primefaces/primefaces-showcase/pull/198/files

Added new issues for styling: #6757

src/main/java/org/primefaces/component/breadcrumb/BreadCrumbRenderer.java
```diff
@@ -28,6 +28,7 @@
 import javax.faces.component.UIComponent;
 import javax.faces.context.FacesContext;
 import javax.faces.context.ResponseWriter;
+import org.apache.commons.collections4.keyvalue.DefaultMapEntry;
```
melloware: We don't use Commons Collections right in PF? I don't think you can use this?

melloware (changes requested):

src/main/java/org/primefaces/component/breadcrumb/BreadCrumbRenderer.java
tandraschko: nope. how can it compile?

melloware: It must be part of a provided transitive dependency?

tandraschko: yeah from POI. can you please create a issue in general to hide this dependencies in PF?

melloware: Yep will create an issue.

jepsar: Sorry.. should have been `Map.entry`.. will fix this now

melloware (approved):

mertsincan:
Hi @jepsar ,

Thanks a lot for your contribution ;)

PR #6348, opened 2020-09-25
Implement #6335: DataView: support PrimeFlex

### LINKED ISSUE #6335: DataView: support PrimeFlex
similar to https://github.com/primefaces/primefaces/issues/6334

DataView currently always uses the old Grid-CSS for it´s grid-layout. We may offer an option (eg via new gridLayout-attribute) for PrimeFlex-Grid.

christophs78:
Showcase-PR: https://github.com/primefaces/primefaces-showcase/pull/196


src/main/java/org/primefaces/component/dataview/DataViewBase.java
```diff
@@ -44,7 +44,8 @@
         listIcon,
         multiViewState,
         gridRowStyle,
-        gridRowStyleClass
+        gridRowStyleClass,
+        usePrimeFlex
```
melloware: I wonder if this attribute should be less specific and make it like datatable layout="grid" or layout="flex"

melloware: That way we can make a consistent property in all the components?

melloware (changes requested):

src/main/java/org/primefaces/component/dataview/DataViewBase.java
christophs78: I had the same thoughts but no good answer yet. DataView comes with list and grid - view. You can toggle between both. Layout-attribute defines which one is shown per default.
Now we have two technical ways how we render grid - view. 
I am not happy with my attribute-naming. But i am pretty sure we can not add this to the existing layout-attribute.

melloware: So PanelGrid has layout https://primefaces.github.io/primefaces/8_0/#/components/panelgrid?id=responsive

_Displays data in a 'tabular' layout, 'grid' layout or 'flex' layout. The grid and flex layout are responsive layouts. Default value is 'tabular'._

Why can't you just add the option "flex" to the layout instead of this new property?

christophs78: Yep. I was the one who added layout="flex" to PanelGrid.
PanelGrid looks almost same in all three layout-modes. The difference is its html-rendering. html-table vs. div with grid-css-classes vs div with primeflex-css-class.

DataView is IMO a different thing.
http://primefaces.org/showcase/ui/data/dataview/paginator.xhtml

It has two layouts which look fundamental different. List and Grid. 
You have a toggle switch.
image

In XHTML you do something like
```xhtml
<p:dataView>
   <p:dataViewGridItem>
     </p:dataViewGridItem>
     <p:dataViewListItem>
     </p:dataViewListItem>
</p:dataView>
```

Now we have two technical "flavours" of grid. But from an enduser-perspective it´s still the same grid.

From a somehow isolated perspective i´d like to deliver PrimeFlex as part of PrimeFaces. (Like Grid CSS comes with PrimeFaces) And just use internal PrimeFlex because it "does the right thing". And no need to offer Grid CSS and Prime Flex - modes . And bring PrimeFlex on your own into the project, ...
But i understand the arguments why this is a bad idea.

Maybe it´s better to introduce a new Context-Param which tells PrimeFaces to use PrimeFlex whereever it can instead of Grid CSS.

Or...

melloware: OK I see the issue now.

melloware: We could make it like `RTLAware` and `ComponentUtils.isRTL` have a `FlexAware` and `ComponentUtils.isFlex` and that way you can set it with Context Param or with the individual component with a `flex=true` property.

melloware (approved):
FlexAware changes look great!

src/main/java/org/primefaces/component/api/FlexAware.java
```diff
@@ -0,0 +1,45 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.component.api;
+
+/**
+ * Use PrimeFlex instead Grid CSS
+ */
+public interface FlexAware {
```
Rapster, line 29: I would rename it PrimeFlexAware (to make it more understandable and to avoid any confusion with flex css display for instance)

src/main/java/org/primefaces/component/api/FlexAware.java
```diff
@@ -0,0 +1,45 @@
...
+     * @return true to use PrimeFlex instead of Grid CSS.
+     */
+    boolean isFlex();
+
+    /**
+     * Use PrimeFlex (instead Grid CSS) for this component?
+     *
+     * @param flex true to use PrimeFlex instead of Grid CSS.
+     */
+    void setFlex(boolean flex);
```
Rapster: setter is not mandatory to implement to make it work: can be removed

src/main/java/org/primefaces/component/api/FlexAware.java
melloware: The only reason I like FlexAware is its concise like our others RTLAware, TouchAware, etc...

Rapster: Yeah I agree I think PrimeFlexAware is still a concise name but it's not a big deal anyway, if you think it's better you can keep it that way then 😉

src/main/java/org/primefaces/component/api/FlexAware.java
christophs78: Done. :-)

src/main/java/org/primefaces/component/dataview/DataViewRenderer.java
```diff
@@ -235,21 +235,32 @@ protected void encodeGridLayout(FacesContext context, DataView dataview) throws
...
             ResponseWriter writer = context.getResponseWriter();
-
             int columns = grid.getColumns();
             int rowIndex = dataview.getFirst();
             int rows = dataview.getRows();
             int itemsToRender = rows != 0 ? rows : dataview.getRowCount();
             int numberOfRowsToRender = (itemsToRender + columns - 1) / columns;
-            String columnClass = DataView.GRID_LAYOUT_COLUMN_CLASS + " " + GridLayoutUtils.getColumnClass(columns);
+            String columnClass = DataView.GRID_LAYOUT_COLUMN_CLASS + " ";
+            if (ComponentUtils.isFlex(context, dataview)) {
```
melloware: might want to store off the result of `boolean isFlex = ComponentUtils.isFlex(context, dataview)` to reuse

melloware (changes requested):

src/main/java/org/primefaces/component/dataview/DataViewRenderer.java
christophs78: done

melloware (approved):

tandraschko:
looks good in general
i would just do a short return in ComponentUtils#isFlex when the component.isFlex(); is true; this avoid is get getInstance call in probably many cases

melloware:
@tandraschko I wonder if we should do that for `isRTL` and `isTouch` as well in `Componentutils` to allow the Component to override the global setting?

tandraschko:
yep, sounds goods

melloware:
So...
```java
public static boolean isFlex(FacesContext context, FlexAware component) {
        return component.isFlex() || PrimeRequestContext.getCurrentInstance(context).isFlex();;
 }
```

melloware:
I am going to merge this so you can work on DataGrid next

PR #6353, opened 2020-09-26
Fix #5905: AutoComplete: emptyMessageSelect listener

### LINKED ISSUE #5905: Autocomplete: Custom empty result action
## 1) Environment
- PrimeFaces version: 8.0

## 2) Expected behavior

Possibility to customize, container for no results found, so that it can for example add a button asking the user if he wants to create a new record, EX:

image

## 3) Actual behavior

..

## 4) Steps to reproduce

.. 

## 5) Sample XHTML

..

## 6) Sample bean

..

jepsar:
Fix https://github.com/primefaces/primefaces/issues/5905

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
+++ b/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
@@ -969,14 +969,14 @@ PrimeFaces.widget.AutoComplete = PrimeFaces.widget.BaseWidget.extend({
      * @private
      */
     invokeMoreTextBehavior: function() {
-        if(this.hasBehavior('moreText')) {
+        if(this.hasBehavior('moreTextSelect')) {
             var ext = {
                 params : [
-                    {name: this.id + '_moreText', value: true}
+                    {name: this.id + '_moreTextSelect', value: true}
                 ]
             };
 
-            this.callBehavior('moreText', ext);
+            this.callBehavior('moreTextSelect', ext);
         }
     },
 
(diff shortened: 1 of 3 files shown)


melloware (approved):

src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
```diff
@@ -587,6 +587,18 @@ PrimeFaces.widget.AutoComplete = PrimeFaces.widget.BaseWidget.extend({
             $this.checkMatchedItem = false;
         });
 
+        this.panel.on('click', function() {
```
melloware: make it `click.emptyMessage` I like to add namespaces so when debugging we can easily tell where the onClick came from

melloware (changes requested):

tandraschko:
IMO "emptyMessage" as eventname is quite unclear when its triggered

jepsar:
> IMO "emptyMessage" as eventname is quite unclear when its triggered

Agreed. I tried to keep it in line with `moreText`. I think both should be suffixed with "Click" or "Select".

jepsar:
@tandraschko @melloware Shall I rename the events to `emptyMessageSelect` and `moreTextSelect` (and update the breaking changes section in the migration guide)?

tandraschko:
i think so
WDYT @melloware ?

melloware:
Yes I think we should its more specific name to what is happening.  I like "Select"

jepsar:
Done (including migration guide).

melloware (approved):

PR #6363, opened 2020-09-28
Fix #6359: Dynamic downloads should not be cached

### LINKED ISSUE #6359: FileDownload: configure Cache-Control
**Description**
Currently the fileDownload component uses Cache-Control: public and Pragma: public as the default response header.  Sometimes users have the possibility to download sensitive data. These files should not be cached anywhere. (especially considering the GDPR)

Adding a new attribute "cache-handling" or possibly a generic "response-headers", where the cache-control level can be configured.

melloware:
Consolidated cache control into ResourceUtils.

3 spots are caching and were all doing it slightly differently.

### DIFF
--- a/src/main/java/org/primefaces/application/resource/BaseDynamicContentHandler.java
+++ b/src/main/java/org/primefaces/application/resource/BaseDynamicContentHandler.java
@@ -28,7 +28,10 @@ import java.text.SimpleDateFormat;
 import java.util.Calendar;
 import java.util.Locale;
 import java.util.TimeZone;
+
 import javax.faces.context.ExternalContext;
+
+import org.primefaces.util.ResourceUtils;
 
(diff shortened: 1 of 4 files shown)


tandraschko:
1) if (PrimeRequestContext.getCurrentInstance(context).isSecure()) { - can you check why we only add it on https? doesnt make sense for me
2) addCacheControl is to common, maybe addNoneCacheControl

melloware:
OK I called it "addNoCacheControl" as that is a better name thanks!  

I thought that secure only was strange too let me look when that was added.

melloware:
OK the `isSecure` check  was from Cagatay in 2013 with the commit message "SSL fix for IE8". So I think it is safe to remove.

tandraschko:
+1, sounds good

PR #6371, opened 2020-09-29
Fix #1039: DefaultMenuModel allow manually set Ids

### LINKED ISSUE #1039: DefaultMenuItem: Id gets overwritten
in BaseMenuRenderer encodeEnd a UniqueId is being generated for every MenuElement of the MenuModel.
If an ID was set before, it gets overwritten.

### LINKED ISSUE #6367: DefaultMenuItem and DefaultSubMenu setting Id does not work
DefaultMenuItem and DefaultSubMenu  setting id does not work, For exaple, I am using `defaultMenuItem.setId("testMenuItem") `
Normaly, it should be `<li id="menuform:testMenuItem" />` but it render like that `<li id="menuform:j_idt51_3" />`

melloware:
I think its this simple just use the ID if it was already set else generate a custom one.

### DIFF
--- a/src/main/java/org/primefaces/model/menu/BaseMenuModel.java
+++ b/src/main/java/org/primefaces/model/menu/BaseMenuModel.java
@@ -27,6 +27,8 @@ import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.primefaces.util.LangUtils;
+
 /**
  * Base implementation for model of a programmatic menu
  */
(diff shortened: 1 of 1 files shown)


Rapster:
looks good

stolp:
This patch is extremely helpful for people doing test automation on primefaces applications, so thank you very much for this!

melloware:
Apparently this has been a big problem for anyone who uses Premium Themes too which most use a Custom Menu object for the main menu: https://forum.primefaces.org/viewtopic.php?f=40&t=53843

tandraschko:
i just thinking about if we should always increase the counter, not sure

melloware:
I can definitely increase the counter no matter what. I don't think it matters either way but let me know if you do want me to make that change.

PR #6374, opened 2020-09-30
Fix #3812,#929: Datatable selection fix

### LINKED ISSUE #929: Using datatable within ui:repeat breaks selection
I am using DataTable inside ui:repeat. Datatable is with multiple selectionMode. The selection is not working and only the selection from last datatable is considered. 

The same issue is present if datatable is used inside p:rowExpansion or p:AccordianPanel (with model)

Version Tested: 5.2, 5.3
Forum discussion and proposed solution : http://forum.primefaces.org/viewtopic.php?f=3&t=36494&p=137360#p137360

### LINKED ISSUE #3812: Datatable: more than one dataTable within p:repeat (or ui:repeat) breaks selection
## 1) Environment
- PrimeFaces version: 6.2.5
- Application server + version: jetty from primefaces-test (not important)
- Affected browsers: all (server side bug)
- **Related issue: #929**

## 2) Expected behavior
In a scenario having more than one dataTable inside p:repeat (or ui:repeat), selected itens on dataTable should work for all selected records on all tables.

## 3) Actual behavior
Only the values from the last table are processed

## 4) Steps to reproduce
Use this reproducer: https://github.com/lagar84/primefaces-test
In the reproducer: 
    Check all records on both tables (click on all checkboxes)
    Click on Submit ALL button
    BUG: Only the last table values are shown

## 5) Sample XHTML
```
<!DOCTYPE html>

    <h:head>
    </h:head>
    <h:body>

        <h3>Steps to reproduce the problem:</h3>
        <ol>
        	<li>Check all records on both tables (click on all checkboxes)</li>
        	<li>Click on Submit ALL button</li>
        	<li>BUG: Only the last table values are shown</li>        
        </ol>
        
        <h:form>
        	<p:repeat var="par" value="#{testView.parents}">
	        	<p:dataTable id="parTable" value="#{par.children}" var="rec" rowKey="#{rec.id}"
			 		 selection="#{par.childrenSelected}" rowSelectMode="checkbox">
					 
					 <p:column selectionMode="multiple" style="width: 50px; text-align:center;" />
				     
				     <p:column headerText="ID" style="width: 100px;">
				     	<h:outputText value="#{rec.id}" />
				     </p:column>
				     <p:column headerText="Name" >
				     	<h:outputText value="#{rec.name}" />
				     </p:column>
					 
				</p:dataTable>
				<br/>
			</p:repeat>
			<br/>
			<br/>
			
			<p:commandButton value="Submit ALL" process="@form" update="@form"/>
			
			<br/>
			<br/>
			<h3>You selected:</h3>
			<p:repeat var="parSel" value="#{testView.parents}">
				<p:repeat var="chSel" value="#{parSel.childrenSelected}">
				
					<h:outputText value="#{chSel.id}" />
					<h:outputText value=" - " />
					<h:outputText value="#{chSel.name}" />
					<br/>
				
				</p:repeat>
			</p:repeat>
        
        </h:form>

    </h:body>
</html>
```

## 6) Sample bean
```
package org.primefaces.test;

import java.io.Serializable;
import java.util.ArrayList;
import java.util.List;

@ViewScoped
       
    private List<RecordParent> parents = new ArrayList<>();
    
    @PostConstruct  
    public void init() {
                
        List<Record> children1 = new ArrayList<>();
				     children1.add(new Record(1, "Mojarra"));
				     children1.add(new Record(2, "Myfaces"));
        
		List<Record> children2 = new ArrayList<>();
			         children2.add(new Record(3, "Eclipselink"));
			         children2.add(new Record(4, "Hibernate"));
			         
		RecordParent parent1 = new RecordParent(children1);
		RecordParent parent2 = new RecordParent(children2);
		
		parents.add(parent1);
		parents.add(parent2);
        
    }

	public List<RecordParent> getParents() {
		return parents;
	}

	public void setParents(List<RecordParent> parents) {
		this.parents = parents;
	}

   
	
}
```

```
package org.primefaces.test;

import java.util.ArrayList;
import java.util.List;

public class RecordParent {

	private List<Record> children = new ArrayList<>();
	private List<Record> childrenSelected;
	
	public RecordParent(List<Record> children) {
		this.children = children;
	}

	public List<Record> getChildren() {
		return children;
	}

	public void setChildren(List<Record> children) {
		this.children = children;
	}

	public List<Record> getChildrenSelected() {
		return childrenSelected;
	}

	public void setChildrenSelected(List<Record> childrenSelected) {
		this.childrenSelected = childrenSelected;
	}
}
```

```
package org.primefaces.test;

public class Record {

	private Integer id;
	private String name;
	
	public Record() {
	}

	public Record(Integer id, String name) {
		super();
		this.id = id;
		this.name = name;
	}
	

	public Integer getId() {
		return id;
	}

		this.id = id;
	}

	public String getName() {
		return name;
	}

		this.name = name;
	}
	
	
}

```

datatable_selection_repeat_bug_reproducer

melloware:
I have confirmed it is fixed using this test case and also regression tested all the Selection examples in the Showcase and everything is working.

Also of note this doesn't just fix tables in a `p:repeat` it also fixes them for AccordionPanel/TabView etc when using a dynamic `model`.  The Datatable has the same issue.

### DIFF
--- a/src/main/java/org/primefaces/component/datatable/DataTable.java
+++ b/src/main/java/org/primefaces/component/datatable/DataTable.java
@@ -757,10 +757,6 @@ public class DataTable extends DataTableBase {
         return null;
     }
 
-    public Object getLocalSelection() {
-        return getStateHelper().get(PropertyKeys.selection);
-    }
-
     @Override
     public Map<String, Class<? extends BehaviorEvent>> getBehaviorEventMapping() {
         return BEHAVIOR_EVENT_MAPPING;
--- a/src/main/java/org/primefaces/component/datatable/DataTableBase.java
+++ b/src/main/java/org/primefaces/component/datatable/DataTableBase.java
@@ -23,14 +23,15 @@
  */
 package org.primefaces.component.datatable;
 
-import org.primefaces.component.api.*;
-import org.primefaces.model.FilterMeta;
-import org.primefaces.model.SortMeta;
+import java.util.Collections;
+import java.util.Map;
 
 import javax.el.MethodExpression;
 import javax.faces.component.behavior.ClientBehaviorHolder;
-import java.util.Collections;
-import java.util.Map;
+
+import org.primefaces.component.api.*;
+import org.primefaces.model.FilterMeta;
+import org.primefaces.model.SortMeta;
 
 public abstract class DataTableBase extends UIData
         implements Widget, RTLAware, ClientBehaviorHolder, PrimeClientBehaviorHolder, Pageable, MultiViewStateAware<DataTableState> {
@@ -167,11 +168,21 @@ public abstract class DataTableBase extends UIData
     }
 
     public Object getSelection() {
-        return getStateHelper().eval(PropertyKeys.selection, null);
+        Object o = getLocalSelection();
+        if (o == null) {
+            return getStateHelper().eval(PropertyKeys.selection, null);
+        }
+        else {
+            return o;
+        }
     }
 
     public void setSelection(Object selection) {
-        getStateHelper().put(PropertyKeys.selection, selection);
+        getStateHelper().put(PropertyKeys.selection + getClientId(), selection);
+    }
+
+    public Object getLocalSelection() {
+        return getStateHelper().get(PropertyKeys.selection + getClientId());
     }
 
     public String getEmptyMessage() {


tandraschko:
"PropertyKeys.selection + getClientId()" ready pretty weird and should not be required. Other components in a repeat are also working correctly.

melloware:
I agree but this definitely fixes the problem we are having.  @Rapster suggested its something in UIDATA but I can verify that if not the Selection always just uses the last Datatable in the p:repeat.

tandraschko:
yeah but we should also try to fix the root of the issue, not hack around it. So -1 to fix it this way

melloware:
Understood.  I will need some help on this one

tandraschko:
I would try to debug with ui:repeat and MyFaces 2.3-next and check what happens in restore/saveState

melloware:
OK I can see the problem is here in `DataTable.processUpdates`.

https://github.com/primefaces/primefaces/blob/8149e51621d3f6bdb71fdb918bdb4f0e959615e4/src/main/java/org/primefaces/component/datatable/DataTable.java#L302-L307

The value in `getLocaleSelection` is clearly wrong.
```java
selectionVE.setValue(context.getELContext(), getLocalSelection());
```

melloware:
When I run with MyFaces 2.3-next the reproducer does not even render the tables. It works with the other 4 MyFaces and Mojarra implementations.

tandraschko:
did you try M4? 
If not, please create a issue and maybe a PR :P

melloware:
Let me try M4.

melloware:
M4 has the same issue. It is strange no errors reported but either using ui:repeat or p:repeat it does not render the tables.

tandraschko:
does this also happen on the showcase or only in this special example?

melloware:
The Repeat example works on the showcase with M4 just not in this example where its repeating Datatables.

tandraschko:
ok. so please create a MF issue and, if you like, a PR.

melloware:
Thanks to your help I have tested MyFaces M4 with `ui:repeat` and `p:repeat` and they both exhibit the same behavior.

The value of `getLocaleSelection` in `processUpdates` is always getting the value for LAST datatable in the repeat.

tandraschko:
It could be that UIRepeat doesnt push the "par" entry/varStatus when calling processUpdates. Maybe you can debug MF UIRepeat a bit what happens.
In general p:repeat should not be used with MF, it's a stupid clone of the Mojarra ui:repeat.

melloware:
OK debuggin MF ui:repeat and I can see in the loop its calling `processUpdates` with the correct child Datatable in each loop.
```java
child.processUpdates(faces);
```

I can see it looping over both child tables properly:
```
DataTable [id=j_id_7:j_id_8:0:parTable]
DataTable [id=j_id_7:j_id_8:1:parTable]
```

However on loop 1 `datatable.getLocalSelection`is the last datatable and on loop 2 it evaluates to NULL because when it called  `setSelection(null);` in the first loop it nulled it out.

So it looks like its the proper datatable, and `UIRepeat` is looping over the tables correctly.  But there is something I am not seeing.

tandraschko:
i think there could be 2 cases:
- getting the selection is not evaluated inside the correct row
- setting the selection is not within the current row state

i think you have to debug a bit to check that the varStatus has to be correclty for the correct row for both cases

melloware:
OK i was able to take UI Repeat out of the equation and reproduce the problem with dynamic tabview. Both examples are included in this reproducer:
[pf-3812.zip](https://github.com/primefaces/primefaces/files/5340473/pf-3812.zip)

UIRepeat: http://localhost:8080/primefaces-test/repeat.xhtml

TabView: http://localhost:8080/primefaces-test/tabview.xhtml

Show the same behavior so its not related to UI Repeat the bug is definitely in the Datatable Selection.  You might be right about the row var by why then does it work normally with a datatable.  The Selection is correct when not using multiple datatables which explains why the "hack" in this PR works it makes the selection unique to each datatable.  But the WHY this is happening is still a mystery to me.  

I am going to report through my PRO account.

tandraschko:
oh, mert will have fun :D
the bug might also be in our UIData/UITabPanel stuff. I will just not accept such a workaround with clientId in the state attr, this is wrong and only a hack.

melloware:
@tandraschko Agreed. I am using this workaround locally in an overridden faces-config.xml to get past our production issue but I 100% agree this is a "hack".

tandraschko:
i will close this PR
as i already said, it would be just a hack

tandraschko:
@melloware please attach your sample to the issue - or create a new issue with the easier example and close the old ones

PR #6378, opened 2020-10-01
Fix #6355: DatePicker lazy disabled

### LINKED ISSUE #6355: DatePicker: lazy load disabled dates

jepsar:
Fix #6355

For showcase see https://github.com/primefaces/primefaces-showcase/pull/201

### DIFF
--- a/src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java
+++ b/src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java
@@ -31,7 +31,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
-import java.util.Objects;
+import java.util.Map.Entry;
 import javax.faces.component.UIComponent;
 import javax.faces.context.FacesContext;
 import javax.faces.context.ResponseWriter;
@@ -41,7 +41,7 @@ import org.primefaces.component.api.UICalendar;
 import org.primefaces.component.calendar.BaseCalendarRenderer;
 import org.primefaces.expression.SearchExpressionFacade;
 import org.primefaces.expression.SearchExpressionUtils;
-import org.primefaces.model.ScheduleEvent;
+import org.primefaces.model.datepicker.DateMetaData;
 import org.primefaces.model.datepicker.DateMetaDataModel;
 import org.primefaces.model.datepicker.LazyDateMetaDataModel;
 import org.primefaces.util.CalendarUtils;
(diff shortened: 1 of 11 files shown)


src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java
```diff
@@ -27,25 +27,83 @@
...
+        }
+
+        encodeDateMetaDataAsJSON(context, datePicker, model);
+    }
+
+    protected void encodeDateMetaDataAsJSON(FacesContext context, DatePicker datePicker, DateMetaDataModel model) throws IOException {
+        JSONObject jsonDateMetaData = new JSONObject();
+
+        if (model != null) {
+            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("M/d/yyyy");
```
jepsar: I thought the JS needed this format. Needs to be replaced.

src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java
```diff
@@ -30,22 +30,78 @@
...
+            encodeDateMetaData(context, datePicker);
+        }
+        else {
+            super.encodeEnd(context, component);
+        }
+    }
+
+    protected void encodeDateMetaData(FacesContext context, DatePicker datePicker) throws IOException {
+        String clientId = datePicker.getClientId(context);
+        DateMetaDataModel model = datePicker.getModel();
```
melloware, line 66: Why not immediately exit here if `model == null`?

src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java
```diff
@@ -30,22 +30,78 @@
...
+            LazyDateMetaDataModel lazyModel = ((LazyDateMetaDataModel) model);
+            lazyModel.clear();
+            lazyModel.loadDateMetaData(startDate, endDate);
+        }
+
+        encodeDateMetaDataAsJSON(context, datePicker, model);
+    }
+
+    protected void encodeDateMetaDataAsJSON(FacesContext context, DatePicker datePicker, DateMetaDataModel model) throws IOException {
+        JSONObject jsonDateMetaData = new JSONObject();
```
melloware, line 87: Rather than checking `model != null` 2 lines down why not make the first line of this method `if (model == null) return;` to short circuit this whole method?

src/main/java/org/primefaces/model/datepicker/DateMetaData.java
```diff
@@ -0,0 +1,32 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.model.datepicker;
+
+public interface DateMetaData {
+
+    public boolean isDisabled();
+
+    // TODO add styleClass in new PR
```
melloware: TODO?

src/main/java/org/primefaces/model/datepicker/DefaultDateMetaDataModel.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+package org.primefaces.model.datepicker;
+
+import java.io.Serializable;
+import java.time.LocalDate;
+import java.util.HashMap;
+import java.util.Map;
+
+public class DefaultDateMetaDataModel implements DateMetaDataModel, Serializable {
+
```
melloware, line 32: Missing `private static final long serialVersionUID = 1L;`

src/main/java/org/primefaces/model/datepicker/DefaultDateMetaData.java
```diff
@@ -0,0 +1,92 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.model.datepicker;
+
+import java.io.Serializable;
+
+public class DefaultDateMetaData implements DateMetaData, Serializable {
+
```
melloware, line 29: Missing `private static final long serialVersionUID = 1L;`

melloware (changes requested):

src/main/java/org/primefaces/model/datepicker/DateMetaData.java
```diff
@@ -0,0 +1,30 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.model.datepicker;
+
+public interface DateMetaData {
+
+    public boolean isDisabled();
```
melloware: dont think you need public on this interface method.

melloware (changes requested):

melloware (approved):

melloware (approved):

tandraschko:
can you review @christophs78 ?

jepsar:
Please note that the whole model seems overkill for just setting disabled, but I'm planning to use the same model to add properties like style class and title.

christophs78:
From a bird's eye view this PR looks good. Same for https://github.com/primefaces/primefaces/pull/6380.
I´ll try to take some time later this week to checkout both PR´s and do some manual regression-tests.

christophs78:
Approved.

@jepsar : May i ask you for adding an integration-test? (something like eg https://github.com/primefaces-extensions/primefaces-integration-tests/blob/master/src/test/java/org/primefaces/extensions/integrationtests/datepicker/DatePicker003Test.java) Just to avoid breaking this at some point in the future.

jepsar:
@christophs78 I'll have a go at it

melloware:
I am going to merge this since its passed all review and then @jepsar can add integration tests using PF-9.0-SNAPSHOT.

PR #6380, opened 2020-10-02
Fix #6379: DatePicker navigator button disabling

### LINKED ISSUE #6379: DatePicker: month navigation buttons get not disabled when min-/maxDate is reached
**Describe the defect**
In the popup of the Calendar component the forward and backward navigation buttons are disabled if the min- or maxDate are reached.
This was already part of #5967. The fix was to ignore clicking the navigation buttons. But with this solution there ist no distinction between active and disabled navigation button. 
In my opinion the behavior should be as by the calendar. The navigation buttons of the datePicker should rendered but disabled, as it is for the Calendar.

**Reproducer**
Look at the showcase for Calendar - Restricted vs. DatePicker - MinMax,
or #5967

**Environment:**
 - PF Version: PrimeFaces-8.0.2-SNAPSHOT (showcase)
 - JSF + version: Mojarra-2.3.13
 - Affected browsers: does not matter

Possible Solution:
Add style class in renderBackwardNavigator and renderForwardNavigator if link sould be disabled to indicate that.

melloware:
Prev and Next buttons are now disabled if they cannot be advanced.

image

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
+++ b/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
@@ -1106,6 +1106,9 @@
                     }
                 });
             }
+
+            // #6379 set state of navigator buttons
+            this.setNavigationState(this.viewDate);
         },
 
         renderTriggerButton: function () {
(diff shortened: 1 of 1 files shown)


tandraschko:
can you review @christophs78 ?

christophs78:
Checked some corner-cases:

- combined with monthNavigator and yearNavigator --> OK
- combinded with (eg) numberOfMonths="3" --> OK, but not perfect 
grafik

Overall it's a clear improvement compared to our status-quo.

PR #6388, opened 2020-10-04
Fix #5438 - DataTable: cleanup sorting

### LINKED ISSUE #5438: DataTable: cleanup sorting

Rapster:


src/main/java/org/primefaces/component/datatable/DataTableBase.java
```diff
@@ -114,7 +107,8 @@
         escapeText,
         rowEditMode,
         stickyTopAt,
-        globalFilterFunction
+        globalFilterFunction,
+        sortPriority
```
melloware, line 111: Why is sortPriority here on the Datatable looks like you added it to the column?

Rapster: Wrong PR I created a new one, see #6390

Issue #2110, opened 2017-02-22
DataTable decoding does not consider dynamically added/removed columns correctly


jessepeng:
## 1) Environment
- PrimeFaces version: 6.0
- Application server + version: Tomcat 8.5.8 with Mojarra 2.2.14
- Affected browsers: all

## 2) Expected behavior

When adding/removing columns dynamically (e.g. using `getChildren().add()`), attached AJAX events to these columns or components inside these columns should get fired on every request. In detail, the `processChildren()` method should iterate over all actual children at every point in time.

## 3) Actual behavior

The `processChildren()` method iterates over the table's columns using `getIterableChildren()`. In `DataTable` this method is overwritten as follows:

```
@Override
    protected List<UIComponent> getIterableChildren() {
        if(iterableChildren == null) {
            iterableChildren = new ArrayList<UIComponent>();
            for(UIComponent child : this.getChildren()) {
                if(!(child instanceof ColumnGroup)) {
                    iterableChildren.add(child);
                }
            }
        }
        
        return iterableChildren;
    }
```

As one can easily see, once `iterableChildren` is set during first rendering, this method does not reflect further changes to the column structure. So if after the first rendering, a column is added using `getChildren().add()`, it will not be considered in further calls to `processChildren()`.

By the way, `getColumns()` suffers from the same assumption that columns won't be changed in between two requests. However, there is a `setColumns(...)` method that can be used to make the table "forget" about it's previous columns.

## 4) Steps to reproduce

* Create a DataTable
* Add column dynamically (using `getChildren().add()`) and add AJAX behavior to this column (e.g. by adding a CommandButton)
* Render table
* Notice that AJAX event will get fired
* Do something with column structure (e.g. add/remove a column) with AJAX behavior
* Restore view and render table again (e.g. using a postback request)
* Notice that AJAX event will not get fired

## 5) Possible fixes for this problem

Easy fix 1: Just as there is a `setColumns(List<UIColumn> columns)` method, there should be a `setIterableChildren()` method to reset `iterableChildren` so that the table notices that there are different children.

Easy fix 2: Do not save the iterable columns in an instance variable, but do the iteration every time the method is called (probably bad idea, since it's called during every request phase and changes in between the request phases should not happen).

More complex but better fix (not tested, just an idea): using for example a `PostAddToViewEvent`, columns could notify their parent table that they have been added and that the lists `columns` and `iterableColumns` should be invalidated.

tandraschko:
[...]

PR #6390, opened 2020-10-05
Fix #5438 - DataTable: cleanup sorting

### LINKED ISSUE #2601: DataTable: with dynamic column looses sorting on refresh
## 1) Environment
- PrimeFaces version: 6.1.3

## 2) Expected behavior
When you rerender a sorted datatable with dynamic columns the sorted column should keep the style class "ui-state-active" 

## 3) Actual behavior
All columns are now "active"

## 4) Steps to reproduce
See also example zip for prepared test [primefaces-datatable-column-sort.zip](https://github.com/primefaces/primefaces/files/1197195/primefaces-datatable-column-sort.zip)
            
## 5) Sample XHTML

   ```
 <p:commandButton update="cars" value="refresh"  icon="ui-icon-refresh"/>
            <p:dataTable id="cars" var="car" value="#{dtColumnsView.cars}" widgetVar="carsTable">
                <p:columns value="#{dtColumnsView.columns}" var="column" columnIndexVar="colIndex"
                           sortBy="#{car[column.property]}">
                    <f:facet name="header">
                        <h:outputText value="#{column.header}"/>
                    </f:facet>
                    <h:outputText value="#{car[column.property]}"/>
                </p:columns>
            </p:dataTable>
```

## 6) Analysis

When a table is sorted the current sorting is saved by SortFeature.java in the table sortBy property

```
ValueExpression sortByVE = sortColumn.getValueExpression(Column.PropertyKeys.sortBy.toString());
table.setValueExpression(DataTable.PropertyKeys.sortBy.toString(), sortByVE);
```

**Notice that for dynamic columns this is always the same string.** In the example it will be always "#{car[column.property]}" 

When the table is refreshed each column checks if it is currently sorted by comaparing against this string. But because all strings are the same all columns calculate that they are active. org.primefaces.component.datatable.DataTableRenderer#resolveDefaultSortIcon:

```
if((sortField != null && field != null && sortField.equals(field)) || (tableSortByExpression != null && 
  tableSortByExpression.equals(columnSortByExpression))) {
      if(sortOrder.equalsIgnoreCase("ASCENDING"))
           sortIcon = DataTable.SORTABLE_COLUMN_ASCENDING_ICON_CLASS;
       else if(sortOrder.equalsIgnoreCase("DESCENDING"))
           sortIcon = DataTable.SORTABLE_COLUMN_DESCENDING_ICON_CLASS;
        }
```

## 7.) Workaround
For now it seems to work when I override resolveDefaultSortIcon in a simpler custom renderer:

 ```
  protected String resolveDefaultSortIcon(DataTable table, UIColumn column, String sortOrder) {
     String sortIcon = null;
     if(column.equals(table.getSortColumn())) {
       if(sortOrder.equalsIgnoreCase("ASCENDING"))
         sortIcon = DataTable.SORTABLE_COLUMN_ASCENDING_ICON_CLASS;
       else if(sortOrder.equalsIgnoreCase("DESCENDING"))
         sortIcon = DataTable.SORTABLE_COLUMN_DESCENDING_ICON_CLASS;
     }
     return sortIcon;
   }
```

### LINKED ISSUE #3347: DataTable: Pass nulls to the sortFunction or make the nullSortOrder column-specific
## 1) Environment
- PrimeFaces version: 6.1
- Affected browsers: Firefox 57

## 2) Problem description
When sorting a column in a `<p:dataTable />` there is the possibility to define a custom `sortFunction`. However, within this sort function the user has no access to `null ` values as they are already dealt with by the `BeanPropertyComparator#compare` method.

While it is possible to define the `nullSortOrder` on the table (which decides whether nulls should be first or last) this approach has two problems:
1. nulls can never be "in the middle"
2. All columns in the table have to follow the same strategy. It is impossible to have column A follow the "nulls first" strategy while column B follows the "nulls last" strategy.

## 3) Suggested solutions
1. Add the attribute `nullSortOrder` to the column. This would allow different columns to implement different strategies how they can deal with null values. If there is the `nullSortOrder` attribute set on the column, it should overwrite the value of the table.
2. Let the implementer of the sort function deal with nulls. I think when you already offer to make a custom implementation it is not to much to ask of the implementer to deal with null values. Most Java programmers should be used to it anyway when they implement custom Comparators or Comparable classes.

### LINKED ISSUE #5438: DataTable: cleanup sorting

### LINKED ISSUE #6372: DataTable: headerRow refactoring
**Description**
Based on https://www.primefaces.org/showcase/ui/data/datatable/rowGroup.xhtml

As I started to have a look at #5438, `DataTable#sortBy` is used for grouping feature. And since the idea is to remove `DataTable#sortBy` I wonder how to make row grouping working... The way I see it, we should be able to simply configure row grouping like this:
```xml
<p:dataTable var="car" value="#{dtRowGroupView.cars}">

    <p:headerRow groupBy="#{car.brand}" expandable="true" />

    <p:column headerText="Year">
    </p:column>
    <p:column headerText="Color">
    </p:column>
    <p:column headerText="Id">
    </p:column>
</p:dataTable>
```

**Additional context**
My idea is to give slightly less responsability to DataTable which already contains a lot of attributes. Also, I like the idea that headerRow component is the determinist element _(carrying all sorts of attributes)_ that will activate the feature, just like `Column#groupRow` does.

As of today, headerRow has to contain at least one column. I don't have any scenario where several columns would be needed but I guess it can be solved with a facet if user wants to do so.

Anyway, let me know what you think?

### LINKED ISSUE #6468: Datatable: multi-sort loses order with lazy
**Describe the defect**
I've a table with columns [a,b,c,d,e] and lazy initialization. When I provide an intilial list of sortmeta (c, a) it sorts data correctly. But, If I change page, It loses sortmeta order (a, c). For example:

After apply sort columns:
- sortMeta: c DESC, a DESC

After change page (a and c got swapped):
- sortMeta: a DESC, c DESC 

**Environment:**
 - PF Version: 8.0
 - JSF 2.3: Jakarta faces 2.3.14
 - Affected browsers: ALL

**Expected behavior**
Maintain sortmeta order with pagination. 

**Example XHTML**
```html
<p:dataTable scrollable="true" scrollHeight="auto" rows="#{exampleBean.session.nReg}" first="#{exampleBean.session.nPage}" id="dtExample" paginator="true" paginatorTemplate="{CurrentPageReport} {FirstPageLink} {PreviousPageLink} {PageLinks} {NextPageLink} {LastPageLink} {RowsPerPageDropdown}" rowsPerPageTemplate="20,50,100" emptyMessage="Empty table" value="#{exampleBean.dataSource}" var="rec" paginatorAlwaysVisible="true" currentPageReportTemplate="{totalRecords}" selection="#{exampleBean.selectedRecords}" rowKey="#{rec.uid}" sortOrder="descending" style="font-size: 0.9em !important;" lazy="true" sortMode="multiple" styleClass="myTooltip" rowIndexVar="rowIndex" rowStyleClass="#{exampleBean.styleClazz(rec)">
        <p:column selectionMode="multiple" style="width:16px;text-align:center" exportable="false"/>
	<p:column headerText="A" sortBy="#{rec.a}">#{rec.a}</p:column>
        <p:column headerText="B" sortBy="#{rec.b}">#{rec.b}</p:column>
        <p:column headerText="C" sortBy="#{rec.c}">#{rec.c}</p:column>
        <p:column headerText="D" sortBy="#{rec.d}">#{rec.d}</p:column>
	<p:column headerText="E" sortBy="#{rec.e}">#{rec.e}</p:column>
</p:dataTable>
```

Rapster:
**Remove**:
- DataTable#sortMeta _(replaced by DataTable#sortBy)_
- DataTable#sortField _(replaced by UIColumn#field)_
- DataTable#sortOrder _(replaced by UIColumn#sortOrder)_
- DataTable#sortFunction _(replaced by UIColumn#sortFunction)_
- DataTable#expandableRowGroups _(replaced by HeaderRow#expandable)_
- DataTable#nullSortOrder _(replaced by UIColumn#nullSortOrder)_
- DataTable#caseSensitiveSort _(replaced by UIColumn#caseSensitiveSort)_

**Update**
- DataTable#sortMode: default value "multiple"
- DataTable#allowUnsorting" default value "false"
- DataTable#sortBy: expects a single or a collection of SortMeta _(default sorting can either be configured through markup or programmatically)_

**Add**
- Column#sortPriority: defines in which order columns sorter should apply initially
- Column#sortOrder
- Column#nullSortOrder
- Column#caseSensitiveSort
- #6372

**Breaking changes**
- DataTable#sortBy: no longer expect a Map, but a single or a collection of SortMeta _(use SortMeta#builder)_ `UIColumn#field` or `UIColumn#sortBy` must be set to enable default sorting.
- As it is possible from now on to configure default sorted columns on a single sort mode datatable. Users will be expected to implement only one method `LazyDataModel#load(int first, int pageSize, Map<String, SortMeta> sortBy, Map<String, FilterMeta> filterBy)`


tandraschko:
in general it looks quite good, just some points from my first rough review:
1) why not store the sortMeta as sortBy? So we are similar to filterBy and the user could theoretically build it like he want via a Bean.
2) you init sortMeta as component var via PostAddToView; i'm not sure if this works correctly for all cases like MyFaces ViewPooling. Also in Mojarra its sometimes called 2 times in the past. We should in general avoid to store "state" in variables - use the stateHelper here like stated in 1) - 2 birds 1 stone.

Rapster:
Thanks Thomas for your feedbacks.
 
1. The reason why I don't store sortMeta just like sortBy used to do, it's simply because sortMeta is a map and sortBy shouldn't be a map but a list. Users shouldn't bothered figuring out what client id their column would be, on top of that, SortMeta is not really API friendly since it expects columnKey, clientId, value/method expressions... The backing code is quite horrible if you have to configure default sorting... But! I'm working on reintroducing `DataTable#sortBy` in this PR. I'm surely will be able to do something about it. Ideally, it should look like this:

```xml
<p:dataTable var="car" value="#{dtSortView.cars1}" sortBy="#{dtSortView.sorters}">
    <p:column headerText="Id" field="id" />
    <p:column headerText="Year" field="year" />
    <p:column headerText="Brand" field="brand" />
    <p:column headerText="Color" field="color" />
</p:dataTable>
```
`dtSortView.sorters` being a list of SortMeta introducing a new constructor:
```java
    public class SortMeta {

        // attributes

        public SortMeta(String sortField, SortOrder sortOrder, ValueExpression sortBy, MethodExpression sortFunction, Integer priority) {
            // init attributes
        }
    }
```

Finally, you'll be able to configure default sorting through markup or programmatically.

2.  Yes that something I recently realized, few attributes are stored in statehelper and some are not which creates a lot of confusion and issues. I'm gonna try to figure something out so sortMeta is stored in statehelper. I quite like the PostAddToView initilization, that's something we do already for `p:columns`. I don't see how to initialize sortMeta without it...

tandraschko:
Can't we make sortMeta just a list and a add a constructor, like you mentioned, for easy enduser construction? :D (+ rename to sortBy)
We should just align with with filters, we may also think about filterBy/filterMeta then.

Rapster:
Well, having a map makes things so much easier (and actually faster). Even we were turning sortMeta into a list, you still need to sync internal sortmeta and sortBy (defined by the user)  in order to set the column, columnkey _(you can't make it one unique property, I tried, it will create other issues)_. 

I managed to put the sort metas into the statehelper. So far I've been using PostAddToView event to trigger that sync, which looks like the most appropriate place to do so. I don't know much about view pooling but sounds like a particular feature of MyFaces, but in both impl PostAddToView event should be either called during restore view or render response phase. I guess you're referring to #2110?

I'll rename the attribute sortMeta into sortBy later, and yes filters should work the same way (I'm planning to work on it once I'm done with this PR)

tandraschko:
thats the problem, PostAddToView is called only once per cached view AFAIR. Just dont rely it on, always get the Map via StateHelper and update

Rapster:
> thats the problem, PostAddToView is called only once per cached view AFAIR.

You mean during an ajax request for example?

This is something I'm doing now, but I still need to execute the initilization of sortMetas, where should i do it if it's not in PostAddToView?

tandraschko:
Views are reused for multiple requests and its only fired in one request probably.
Just lazy init?

Rapster:
>Views are reused for multiple requests and its only fired in one request probably.

Ok, from what I can see PostAddToView is called for every request I made on showcase (navigation, refresh, ajax)

> Just lazy init?

This is something I tried the very first time, as a quick & dirty approach... Lazy initialization is something that I hate, basically you lose control of your attributes, for example I can't tell you when exactly "sort by metas" are initialized... But right now, I finally managed to make it work 😄

src/main/java/org/primefaces/component/datatable/DataTable.java
```diff
@@ -25,11 +25,13 @@
 
 import java.lang.reflect.Array;
 import java.util.*;
+import java.util.List;
```
melloware: Not sure why it imported List when java.util.* was above it?

melloware (changes requested):
This looks good to me but i assume there are some HeaderRow and Datatable 9.0 MD docs that need updating as well?

Rapster:
I'll deal with the docs once PR reviewed

christophs78:
I´ll do a testdrive later this week. And i´ll update / extend our integration-tests.

melloware (approved):

christophs78:
When i build your PR one unittest fails:
`[ERROR]   DataTableTest.testAllowUnsorting:44 expected: <true> but was: <false>`

christophs78:
Just made a draft-PR for out integration-tests: https://github.com/primefaces-extensions/primefaces-integration-tests/pull/54/files
@Rapster: Did i adapt them as intended by you?

christophs78:
Additional to documentation and integration-tests we need to update showcase after this PR is done.
I´ll try to add some more integration-tests on top of our existing ones.

Rapster:
Hi @christophs78 

Thank you for taking time to test that PR. I'll have to start to have a look to these integration-tests, sounds quite interesting. Regarding the error, yes `DataTable#allowUnsorting` default value has been modified to false. Any plan to move these ITs into primefaces repo? Since you're probably the first to have a look to this PR with Thomas, feel free to give any feedbacks :wink:

src/main/java/org/primefaces/component/datatable/DataTable.java
```diff
@@ -1547,50 +1411,134 @@ public void resetMultiViewState() {
...
+
+        // merge internal sortBy with user sortBy
+        if (sortByTmp != null) {
+            Collection<SortMeta> sortBy;
+            if (sortByTmp instanceof SortMeta) {
+                sortBy = Collections.singletonList((SortMeta) sortByTmp);
+            }
+            else if (!(sortByTmp instanceof Collection)) {
+                throw new FacesException("DataTable#sortBy expects a single or a collection of SortMeta");
+            }
```
christophs78, line 1468: Would it be theoretically possible to add there some kind of backward-compatibility? (Convert PF<=8  sortBy into SortMeta and write warning to log.)
I don´t want to say this is a good idea. Just thinking about it.

Rapster: True, on one hand it can easily be done, on the other hand, since clients will have to do a bit of cleanup of their tables it might be the best opportunity for them to change their sortBy than wait till the backward compatibility is removed ... The decision is up to @tandraschko

Rapster:
Any idea when this PR could be reviewed?

tandraschko:
~1-2 weeks, dont have many time to review this big PR and i have a deadline here

tandraschko:
@Rapster @melloware should we move the migration guide to the offical docs here? So rapster could also prepare the migration guide, it would be much easier to get a overview of the changes when we could first check the migration guide changes

Rapster:
You mean adding a subsection "Migration guide" in "Getting Started" section for example? Anyway, yes sounds like a good idea

melloware:
Agreed. That way we could keep it with each set of docs per version.

tandraschko:
yep

melloware:
I can backfill the Migration Guide section in the 7.0 and 8.0 docs if you handle the 9.0 version @Rapster ?

tandraschko:
i will do it for all versions now, @Rapster can just add his new changes :)

Rapster:
Sounds good

tandraschko:
we added the migration guide now, could you add it to your PR?

Rapster:
Not yet, I was waiting for your review before modifying the docs

tandraschko:
it would be much easier for me to get an overview of your changes  :D

Rapster:
I'm working on it, but if that's an overview you want, you can read it from [here](https://github.com/primefaces/primefaces/pull/6390#issue-497711470) 😄

melloware:
Looks like there are conflicts too.

tandraschko:
@Rapster Just to be sure:
1) the sortMeta is built with columnKey as map key 
2) Initial SortMeta must be built with field or sortBy
3) columnKey == clientId
4) initialSortMeta will search the right columnKey via field || sortBy

What is with https://github.com/primefaces/primefaces/pull/5407? The PR doesnt solve/change this behavior currently, so we still have field and sortBy, right?

I wonder if we should do the following:
Allow to set the columnKey on p:column (maybe just a "key" attr) and allow the SortMetaBuilder to use the columnKey to build the initial sortMeta. Maybe better as field || sortBy but requires another attribute.
So maybe we should talk about 5407 first.

tandraschko:
Or even better: columnKey == id attr of the column (not clientId), so we dont need a new attr or rely on field/sortBy in our Builder

Rapster:
> @Rapster Just to be sure:
> 
> 1. the sortMeta is built with columnKey as map key

Yes

> 2. Initial SortMeta must be built with field or sortBy

Yes
> 3. columnKey == clientId

Yes
> 4. initialSortMeta will search the right columnKey via field || sortBy

Yes via field, sortBy is used to deduce field if not defined

> What is with #5407? The PR doesnt solve/change this behavior currently, so we still have field and sortBy, right?

This PR does a better a job in terms of resolving the field from a VE, but it doesn't do anything regarding sorting

> I wonder if we should do the following: Allow to set the columnKey on p:column (maybe just a "key" attr) and allow the SortMetaBuilder to use the columnKey to build the initial sortMeta. Maybe better as field || sortBy but requires another attribute.

My first initiative was actually to get rid of `UIColumn#columnKey`, and replace it with `UIColumn#field`. At least one reason I didn't do so, it turns out people use sometimes same field for different columns, see #6421, don't know why but it happens... 

> So maybe we should talk about 5407 first.

Right now this is not the topic, but once this PR and the other about filtering will be merged, I will merge #5407

> Or even better: columnKey == id attr of the column (not clientId), so we dont need a new attr or rely on field/sortBy in our Builder

Yeah but people still need to know id of the columns, less painful than clientId for sure but remains tricky. I don't think we should bother too much about this, UIColumn#field should be the key and if people need several columns for the same field then they should add more field to their backing beans/view objects.

tandraschko:
ok i see.
Shouldnt we then better remove #columnKey and use #clientId directly? But this wouldnt work for p:columns probably?

Rapster:
Yeah but whether it's columnKey or clientId, problem is the same, we'll have to loop over sortMeta to find the right match. If it was UIColumn#field it would be a lot more easy to achieve. But right now, let's keep columnKey...

tandraschko:
ok. Did you test your PR with dynamic columns/p:columns? how is the columnKey constructed there?

Rapster:
This PR has been tested will all showcases, for p:columns a subset of DynamicColumns is created:
```java
            if (child instanceof Columns) {
                Columns uiColumns = (Columns) child;
                String uiColumnsClientId = uiColumns.getClientId(context);

                for (int j = 0; j < uiColumns.getRowCount(); j++) {
                    DynamicColumn dynaColumn = new DynamicColumn(j, uiColumns);
                    dynaColumn.setColumnKey(uiColumnsClientId + separator + j);
                    columns.add(dynaColumn);
                }
            }
```

tandraschko:
ups, unsure if i selected "squashe and merge" :D

Thanks for effort @Rapster , i really enjoy the result.
Please update the showcase by yourself.

Could you also please update filtering in the same level now? (Builder for initial filterBy and that stuff)

melloware:
From what I can tell it looks like you squashed and merged correctly.

Rapster:
Thanks Thomas,

It's okay, I can do the merge next time :wink: but yeah squashing commits is pretty cool to avoid cumbersome/messy history (specially that, sometimes I don't bother too much in my commit messages).

I'll update showcase and prepare filtering PR

melloware:
We will update all the Datatable Integration Tests and let you know if we find any issues.

PR #6393, opened 2020-10-05
 #6391: CI: move from Travis CI to Github Actions 

### LINKED ISSUE #6391: CI: move from Travis CI to Github Actions
We have already done this for https://github.com/primefaces-extensions/primefaces-integration-tests.
We will do this for PF Selenium with https://github.com/primefaces-extensions/primefaces-selenium/issues/35.

After both are done successful i´ll try to provide a PR to do this for PrimeFaces too.

christophs78:
@Rapster, @melloware 
Can you please check the Sonar-integration and my replacement for
`if [[ "$TRAVIS_PULL_REQUEST" = "false" ]]; then mvn clean org.jacoco:jacoco-maven-plugin:prepare-agent install sonar:sonar; fi`
And propably we need to add SONAR_TOKEN and maybe GITHUB_TOKEN to the secrets-store of this Github-project.


melloware:
@Rapster what I had to do for PF Extensions was to generate a new Key in SonarCloud and then we will need @tandraschko or @mertsincan to go into "Settings->Secrets" here on GitHub which I can't see and add SONAR_TOKEN key value with the key from SonarCloud.

tandraschko:
whats the benefits compared to travis?

christophs78:
> whats the benefits compared to travis?

Less resource-constraint (-> faster builds), better integrated into Github.

melloware:
+1 to @christophs78 plus...
1.  it seems to be the trend all of the Apache Commons projects have moved to this from Travis.
2. Allows us to turn on Security Scanning for security flaws in our app using the CodeQL action
3. Allows us to use the Dependabot action to automatically provide PR's to let us know when a JAR/plugin needs to be updated. You can close the PR if you don't want to upgrade to that particular version.

Rapster:
Do you need me for anything? I think @tandraschko has permissions to generate sonarcloud tokens

melloware:
OK @tandraschko then we just need "Settings ->Secrets" here on this GitHub project and add SONAR_TOKEN with the generated Sonar key.

tandraschko:
dont think i have enough permissions. @mertsincan can you give me more permissions or can you do it?

.github/workflows/build.yml
```diff
@@ -0,0 +1,35 @@
...
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-m2
+      - name: Log github.ref
+        run: echo "${{ github.ref }}"
+      - name: Build with Maven
+        run: mvn clean install -Psecurity-checks -Pjsdoc -Dmaven.javadoc.skip=true -Djsdoc.skip.typedoc=true --batch-mode --show-version
+      - name: SonarCloud
+        if: ${{ github.ref == 'refs/heads/master' }}
+        run: mvn clean org.jacoco:jacoco-maven-plugin:prepare-agent install sonar:sonar; fi -Dsonar.organization=primefaces -Dsonar.host.url=https://sonarcloud.io -Dsonar.login=${{secrets.SONAR_TOKEN}}
```
melloware: I don't think the `fi` work here.  I am trying it in PFE and it complains about `fi` I would think the IF alone is enough to run this step?

melloware (changes requested):

Rapster:
Regarding sonarcloud, why don't you reuse the token that has been used for travis?

melloware:
> Regarding sonarcloud, why don't you reuse the token that has been used for travis?

When I tried it in PFE it didn't work. But as soon as I generated a new key and put that in it worked.  No idea why unless the Actions are somehow different than from Travis?

Rapster:
Ah did you use the private token _(not the hashed one)_? Anyway, I didn't keep the private token 😅 @tandraschko I gave you admin permissions for sonarcloud, you should be able to ~end the world~ generate a new token

.github/workflows/build.yml
christophs78: https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions
I also think the if alone should do it´s job. Not totally sure whether i compare refs github.ref correct.

melloware: The refs look correct to me in the PFE build I can see it comparing them correctly.

melloware: This PR in PFE shows it skipped the Sonar task:
image

melloware:
@Rapster that was it you can't used the Hash key and hey don't give you a way at looking at an old key so I just generated a new one and that worked. :)

PR #6398, opened 2020-10-07
Fix #6397: InputNumber+Slider combination syncing

### LINKED ISSUE #6397: Slider: InputNumber two way sync is not working
From Stack Overflow: https://stackoverflow.com/questions/64240342/how-to-update-a-pslider-from-the-pinputnumber-it-is-for

1) Environment

PrimeFaces version: 8.0

2) Expected behavior

Two way syncing of InputNumber and Slider should work.

3) Actual behavior

You can see its broken on the Showcase currently: https://www.primefaces.org/showcase/ui/input/slider.xhtml in the "InputNumber Slider" example type 50 into the box the slider doesn't move.

4) Example XHTML

```xml
<p:inputNumber  id="decimal" value="#{sliderView.number2}" decimalSeparator="," />
<p:slider for="decimal" minValue="0.2" maxValue="7.1" step="0.1" />
```

melloware:
The real issue is the Slider was subscribed to a `change` event on the Hidden Input that was never being fired.

You can see its broken on the Showcase currently: https://www.primefaces.org/showcase/ui/input/slider.xhtml in the "InputNumber Slider" example type 50 into the box the slider doesn't move.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js
+++ b/src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js
@@ -160,6 +160,7 @@ PrimeFaces.widget.InputNumber = PrimeFaces.widget.BaseWidget.extend({
      */
     setValueToHiddenInput: function(value) {
         this.hiddenInput.val(value);
+        this.hiddenInput.trigger('change');
     },
 
     /**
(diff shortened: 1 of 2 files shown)


tandraschko:
could you please verify that there will only be one onchange event + AJAX request call with this change?

melloware:
I have confirmed only 1 change event is fired.
```xml
<p:inputNumber id="nmb1" value="#{sliderView.number3}" symbol=" €" valueChangeListener="#{sliderView.onInputChanged}">
         <p:ajax event="change" process="@this" update="growl" />
</p:inputNumber>
<p:slider for="nmb1" minValue="0.2" maxValue="7.1" step="0.1">
	 <p:ajax  process="@this" update="growl" />
</p:slider>
```

tandraschko:
also when you manually type something like a normal user does (without slider)?

melloware:
Correct.

PR #6406, opened 2020-10-12
Fix #6355: DatePicker Sonar Fixes

### LINKED ISSUE #6355: DatePicker: lazy load disabled dates

melloware:
@jepsar pleaese review.  

Reported here: https://sonarcloud.io/project/issues?id=org.primefaces%3Aprimefaces&resolved=false&sinceLeakPeriod=true&types=CODE_SMELL


tandraschko:
I will also rename DateMetaData - "Metadata" is one word

PR #6408, opened 2020-10-12
#6404: DataExporter: should not require ajax="false"

### LINKED ISSUE #6404: DataExporter: should not require ajax="false"
DataExporter requires to set `ajax="false"` for `p:commandButton`.

We already solved this for `FileDownload`. (https://github.com/primefaces/primefaces/issues/5978)
I´ll hope we can do the same for `DataExporter`.

christophs78:
Showcase-PR: https://github.com/primefaces/primefaces-showcase/pull/209


src/main/java/org/primefaces/component/export/DataExporter.java
```diff
@@ -64,11 +66,13 @@
     private ValueExpression exporter;
 
     public DataExporter() {
+        ResourceUtils.addComponentResource(FacesContext.getCurrentInstance(), "filedownload/filedownload.js");
```
melloware, line 76: Didn't know if there was a better way to include this?  I guess there is no renderer to attach to so I guess this is OK.

src/main/java/org/primefaces/util/DynamicContentSrcBuilder.java
```diff
@@ -65,8 +65,12 @@ else if (value instanceof String) {
         }
         else if (value instanceof StreamedContent) {
             StreamedContent streamedContent = (StreamedContent) value;
+            ValueExpression ve = null;
+            if (attributeName != null && attributeName.length() > 0) {
```
melloware: Could this be `!LangUtils.isValueBlank(attributeName)` ?

src/main/java/org/primefaces/component/datatable/export/DataTableExcelExporter.java
```diff
@@ -227,6 +243,35 @@ protected Sheet createSheet(Workbook wb, String sheetName, ExcelOptions options)
         return wb.createSheet(sheetName);
     }
 
+
+    protected String getMonitorKeyCookieName(FacesContext context, ELContext elContext) {
+        String monitorKeyCookieName = Constants.DOWNLOAD_COOKIE + context.getViewRoot().getViewId().replace('/', '_');
```
melloware: I wonder if this could be moved to ResourceUtils as a static method and reused between FileDownload and this?

melloware (changes requested):

src/main/java/org/primefaces/util/DynamicContentSrcBuilder.java
christophs78: Done. I always search for StringUtils and can´t remember LangUtils...

src/main/java/org/primefaces/component/datatable/export/DataTableExcelExporter.java
christophs78: done

src/main/java/org/primefaces/component/export/DataExporter.java
christophs78: No renderer. Don´t know a better place. It´s the same FileDownloadActionListener does.

melloware (approved):

tandraschko:
hmm. can't we move whole if(ajax)else{} logic to a base class/interface? looks weird to have it multiple times

christophs78:
This would essentially mean to pass byte[] (PDF, xls(x)) or String(builder) (CSV, XML) to a generic method. Someone may argue, this is a performance regression for the non-ajax-usecase. But it would allow to cleanup duplicated code.

tandraschko:
why is it a performance regression ? (didnt check the source in detail :P)

christophs78:
Currently non-ajax Excel-export does something like:
```java
protected void writeExcelToResponse(FacesContext context, Workbook generatedExcel, String filename) throws IOException {
	ExternalContext externalContext = context.getExternalContext();
	externalContext.setResponseContentType(getContentType());
	setResponseHeader(externalContext, getContentDisposition(filename));
	addResponseCookie(context);

	OutputStream out = externalContext.getResponseOutputStream();
	generatedExcel.write(out);
}
```

Afterwards it would be something like
```java
ByteArrayOutputStream out = new ByteArrayOutputStream();
generatedExcel.write(out);
```
and pass `out.toByteArray()` to the generic method.

IMO it´s worth to refactor this to cleanup code.

melloware:
+1

christophs78:
Refactored this for XLS(X) and PDF. Will look into CSV und XML tomorrow.

christophs78:
Refactoring done

src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
```diff
@@ -337,6 +338,66 @@ public void export(FacesContext context, List<DataTable> tables, ExportConfigura
...
+            setResponseHeader(externalContext, ComponentUtils.createContentDisposition("attachment", filenameWithExtension));
+
+            addResponseCookie(context);
+
+            OutputStream os = externalContext.getResponseOutputStream();
+            OutputStreamWriter osw = new OutputStreamWriter(os, encodingType);
+            PrintWriter writer = new PrintWriter(osw);
+            writer.write(content.toString());
+            writer.flush();
+            writer.close();
```
melloware: I don't see your close the outputStreamWriter.  A better way may be try with resources.
```java
        try (OutputStreamWriter osw = new OutputStreamWriter(os, encodingType);
            writer.write(content.toString());
            writer.flush();
        }
```

melloware (changes requested):

src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
christophs78: Good point. Fixed. (Just migrated old and not so perfect code before.)

melloware (approved):

tandraschko:
hmm, still unhappy and unsure why we need to buffer it in a BAOS. couldnt we make it without?

christophs78:
> hmm, still unhappy and unsure why we need to buffer it in a BAOS. couldnt we make it without?

PDF-export already worked with BAOS before. For Excel-export i don´t have a (good) idea. IMO we don´t want to pass `org.apache.poi.ss.usermodel.Workbook` to a method in `abstract class DataTableExporter`. `org.apache.poi.ss.usermodel.Workbook` can only write to some kind of OutputStream.

tandraschko:
AFAICS you render the ajax download as base64 url? cant we stream it somehow and move the excel-export to the 2. request?

christophs78:
> AFAICS you render the ajax download as base64 url? cant we stream it somehow and move the excel-export to the 2. request?

Yep. I´m not totally happy with this approach too. But what´s the alternative? Store the content somwhere on the server. (Kind of session?) But what happens if the client does not fetch the download because something fails. We may generate a memory leak or need some kind of cleanup-process. 
What´s your opinion?

tandraschko:
therefore we have streamedContent, which stores the EL-Expression / path to the exporter component in the session.

christophs78:
IMO there is no EL-Expression to store in this usecase. I would need your help when you have a concrete plan how StreamedContent would work with DataExporter.

tandraschko:
yeah, not easy
we could >maybe< store the the clientId and viewStateId instead the EL-Expression in this case, restore the view in the download request and call the exporter. 
Not sure. Maybe base64 is ok for now but we should not make it default via AJAX as its not perfect from memory/bandwidth/streaming perspective.
But we should then not use the whole DynamicContentSrcBuilder, just the util method to make a base64 string from the BAOS

tandraschko:
IMO we should do a big refactoring here and pass the OutputStream to Expotert#export

So the DataExporter ActionListener handles the AJAX/NON-AJAX case and response-cookie stuff

christophs78:
> IMO we should do a big refactoring here and pass the OutputStream to Expotert#export
> 
> So the DataExporter ActionListener handles the AJAX/NON-AJAX case and response-cookie stuff

Yes. Makes sense. Always more to do when we dig into old code. :-)

> Maybe base64 is ok for now but we should not make it default via AJAX as its not perfect from memory/bandwidth/streaming perspective.

For existing usage nothing changes. Dev´s until now had to add ajax="false" to eg p:CommandButton that it worked at all. p:DataExport can not set default´s for Ajax or not Ajax itself. It´s decided by the surrounding component.

christophs78:
I´ll try to play arround with this ideas later this week.
What´s our current release - schedule? Just to keep an eye on getting things done before we want to go into RC-phase.

tandraschko:
Actually october but no idea about the current plan. @mertsincan ?

christophs78:
> But we should then not use the whole DynamicContentSrcBuilder, just the util method to make a base64 string from the BAOS

Done.

src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
```diff
@@ -337,6 +337,62 @@ public void export(FacesContext context, List<DataTable> tables, ExportConfigura
...
+        ajaxDownload(filenameWithExtension, content, null, context);
+    }
+
+    private void ajaxDownload(String filenameWithExtension, byte[] content, String encodingType, FacesContext context) {
+        String contentType = getContentType();
+        if (!LangUtils.isValueBlank(encodingType)) {
+            contentType += "; charset=" + encodingType;
+        }
+
+        String base64 = DatatypeConverter.printBase64Binary(content);
```
melloware: This is the only line that gives me pause as it required JAXB to be on the classpath to work. What about `java.util.Base64` and doing
```java
String encoded = Base64.getEncoder().withoutPadding().encodeToString(someByteArray);
```

melloware (changes requested):

src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
christophs78: Good Point. I also fixed this for DynamicContentSrcBuilder.

melloware: Awesome! I think that is better

christophs78:
> IMO we should do a big refactoring here and pass the OutputStream to Expotert#export
> 
> So the DataExporter ActionListener handles the AJAX/NON-AJAX case and response-cookie stuff

Did some refactoring´s in this direction. It´s a bit more complicated because we have doExport and postExport-methods. I´m not totally happy with my results.

christophs78:
@tandraschko , @melloware : From my side this PR is finalized. There may be an option to solve this with a second request instead of base64. But IMO that´s far from easy to implement. We may do this in a second push if someone of us has time und a good plan how to get this done with a second request.

tandraschko:
not sure if i like it but i didnt check it in detail.
IMO the nicest way would be to pass a OutputStream to the DataExporter and from the outside when can decide if we pass the ResponseWriter Stream or a BAOS to collect data for base64.

christophs78:
Just to discuss a few pro´s and con´s before i invest further time:

- Passing OutputStream to #postExport would decrease memory-usage for non-ajax Excel-Export
- There is no difference for PDF-Export (it already uses BOAS internal). We may further iterate and pass OutputStream already to #preExport to optimize this.
- CSV- and XML-Export are a different thing because they output text via StringBuilder. We would move String(Builder) to OutputStream-conversion from DataTableExporter#sendExport2Client 

IMO there are some trade-offs. Maybe Excel-Export ist our number-one-optimization-target because it´s by far most often used. And so it may make sense to make some compromises to eg CSV- and XML-Export.

tandraschko:
I think CSV should also directly write to a stream, not via StringBuilder

christophs78:
@tandraschko : Hope my last refactoring reflects your ideas.

melloware (approved):

tandraschko:
I would even move the the main logic outside of the DataTableExporter#export to DataExporter#processAction

christophs78:
> I would even move the the main logic outside of the DataTableExporter#export to DataExporter#processAction

Also did this.

There´s always more to refactor. But at some point we have to stop at least for now.

tandraschko:
like it much more now

@Rapster @melloware could you also do a detailed review here?

src/main/java/org/primefaces/component/export/Exporter.java
```diff
@@ -26,9 +26,17 @@
...
-    void export(FacesContext facesContext, List<T> component, ExportConfiguration config) throws IOException;
-}
\ No newline at end of file
+    void export(FacesContext facesContext, List<T> component, OutputStream outputStream) throws IOException;
+
+    String getContentType();
+
+    String getFileExtension();
+
+    void setExportConfiguration(ExportConfiguration exportConfiguration);
```
Rapster: 1. I would keep ExportConfiguration  in export method, having a setter makes your API tricky to use since it must be called before export method (and there is nothing to guarantee that, also exporters should be as much stateless as possible)
2. getContentType: should only return content type without charset, do the concatenation in `DataExporter#processAction`
3. getFileExtension: I don't know if a file extension includes a dot or not...

melloware:
I have done a deep review and this PR looks good to go to me.  I have been reviewing and watching the progress the whole time. So +1 from me.

Rapster (changes requested):

melloware:
Also when we merge this we should definitely Squash and Merge so it appears as only 1 commit.

src/main/java/org/primefaces/component/export/Exporter.java
christophs78: 1) done
2) done
3) added JavaDoc

PR #6413, opened 2020-10-14
Fix #6334: DataGrid flex support

### LINKED ISSUE #6334: DataGrid: support PrimeFlex
DataGrid currently supports "tabular" and "grid" for layout-attribute. We may add "flex" as an alternative to "grid".
Use PrimeFlex-Grid for layout="flex" instead of the old Grid-CSS.
This should allow for more stable layouts.

melloware:
@christophs78 please review


christophs78:
Looks good! (I´ll do a additional test-run after this PR is merged.)

PR #6415, opened 2020-10-16
Add note on scope for deeper graph navigation

cd2357:
To navigate beyond depth 1 in the graph, the ManagedBean needs to be SessionScoped.

See https://stackoverflow.com/a/18880589

### DIFF
--- a/docs/9_0/components/mindmap.md
+++ b/docs/9_0/components/mindmap.md
@@ -94,5 +94,5 @@ parent | null | MindmapNode | Parent node instance.
 **Tips**
 
 - IE 7 and IE 8 are not supported due to technical limitations, IE 9 is supported.
-- The `ManagedBean` backing this component must be `@SessionScoped` to support navigating to nodes deeper than level 1 below the root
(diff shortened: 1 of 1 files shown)


tandraschko:
Actually all longer-living-scopes are supported, just @RequestScoped is to short for such features. There are many others like conversation, window, ...
Im a bit unsure because many features requires longer-living scopes, thats something more general.

cd2357:
Yes. The reason for the PR was just to document that, because it was not clear why it wasn't working with the default setup (using java and xhtml snippets from the PF showcase example).

Ideally, the component would throw an exception if its used with a `@RequestScoped` bean, but not sure how to do that.

Alternatively, the [showcase java snippet](https://www.primefaces.org/showcase/ui/data/mindmap.xhtml) could be changed to use `@ViewScoped` (currently using `@RequestScoped`).

melloware:
Fixed showcase PR: https://github.com/primefaces/primefaces-showcase/pull/208

tandraschko:
Throwing an exception isnt that easy and we would have to do it for many many components.
e.g. many DataTable features are only working with longer-living scopes. I would not document in this way and would like to close it.

maybe we can document it more general on what scope types should be used for N example cases

tandraschko:
will close it for now as we fixed the sample and as long we dont have a better idea to document it

PR #6417, opened 2020-10-17
#6355: DatePicker: merged Ajax requests

### LINKED ISSUE #6355: DatePicker: lazy load disabled dates

jepsar:
Fix https://github.com/primefaces/primefaces/issues/6355

See also https://github.com/primefaces/primefaces-showcase/pull/210

### DIFF
--- a/src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java
+++ b/src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java
@@ -41,12 +41,12 @@ import org.primefaces.component.api.UICalendar;
 import org.primefaces.component.calendar.BaseCalendarRenderer;
 import org.primefaces.expression.SearchExpressionFacade;
 import org.primefaces.expression.SearchExpressionUtils;
+import org.primefaces.model.datepicker.DateMetadata;
+import org.primefaces.model.datepicker.DateMetadataModel;
 import org.primefaces.model.datepicker.LazyDateMetadataModel;
 import org.primefaces.util.CalendarUtils;
 import org.primefaces.util.ComponentUtils;
 import org.primefaces.util.WidgetBuilder;
-import org.primefaces.model.datepicker.DateMetadata;
-import org.primefaces.model.datepicker.DateMetadataModel;
 
 public class DatePickerRenderer extends BaseCalendarRenderer {
 
@@ -54,8 +54,12 @@ public class DatePickerRenderer extends BaseCalendarRenderer {
     public void encodeEnd(FacesContext context, UIComponent component) throws IOException {
         DatePicker datePicker = (DatePicker) component;
 
-        if (context.getExternalContext().getRequestParameterMap().containsKey(datePicker.getClientId(context))) {
-            encodeDateMetadata(context, datePicker);
+        if (ComponentUtils.isRequestSource(datePicker, context)) {
+            Map<String, String> params = context.getExternalContext().getRequestParameterMap();
+            String clientId = datePicker.getClientId(context);
+            if (params.get(clientId + "_year") != null && params.get(clientId + "_month") != null) {
+                encodeDateMetadata(context, datePicker);
+            }
         }
         else {
             super.encodeEnd(context, component);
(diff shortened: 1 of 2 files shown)


src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
```diff
@@ -278,64 +278,58 @@ PrimeFaces.widget.DatePicker = PrimeFaces.widget.BaseWidget.extend({
...
      */
-    fireViewChangeEvent: function(year, month) {
-        if(this.cfg.behaviors) {
+    fireViewChangeEvent: function(date) {
+        var _self = this, fired = false;
+        var options = {
+            params: [
+                {name: this.id + '_year', value: date.getFullYear()},
+                {name: this.id + '_month', value: date.getMonth()}
+            ]
```
melloware, line 297: I think it might be becaue you are missing:
```javascript
source: _self.id,
                    process: _self.id,
                    update: _self.id,
                    formId: _self.cfg.formId,
```

melloware:
I think it might not be calling `encodeEnd` because your component is not being re-rendered because there is no update="".

src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
jepsar: Yes, update did the job!

src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
```diff
@@ -318,6 +318,7 @@ PrimeFaces.widget.DatePicker = PrimeFaces.widget.BaseWidget.extend({
             var viewChangeBehavior = this.cfg.behaviors['viewChange'];
 
             if(viewChangeBehavior) {
+                options.update = (options.update||'') +' '+ this.id;
```
melloware: Not sure if you need to add `process` too?

jepsar: I'm not sure. Did some minor improvements and tested with the showcase PR I submitted (I've added events) and it's all working for me.

jepsar:
@tandraschko @christophs78 can you also have a look please?

src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
```diff
@@ -278,63 +278,55 @@ PrimeFaces.widget.DatePicker = PrimeFaces.widget.BaseWidget.extend({
...
+                    }
+                });
+                return true;
+            };
+        }
 
-                viewChangeBehavior.call(this, ext);
+        if (this.hasBehavior('viewChange')) {
+            if (lazy) {
+                options.update = (options.update||'') +' '+ this.id;
```
tandraschko: Please add spaces before/after || and +

src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java
```diff
@@ -41,21 +41,25 @@
...
 
 public class DatePickerRenderer extends BaseCalendarRenderer {
 
     @Override
     public void encodeEnd(FacesContext context, UIComponent component) throws IOException {
         DatePicker datePicker = (DatePicker) component;
 
-        if (context.getExternalContext().getRequestParameterMap().containsKey(datePicker.getClientId(context))) {
-            encodeDateMetadata(context, datePicker);
+        if (ComponentUtils.isRequestSource(datePicker, context)) {
```
tandraschko: isRequestSource && eventName == viewChange && lazy=true

then you also dont need to the _year+_month null check next

tandraschko:
@jepsar added comments

jepsar:
I think it's okay now. Thanks for all the feedback. I've learned a lot from working on this issue.

tandraschko:
Super.encodeEnd must also be called for other events

tandraschko:
thanks. would be great if someone could retest

tandraschko:
@jepsar also did a small refactoring, please see my last commit and review

PR #6419, opened 2020-10-17
#6418: SelectOneMenu: performance-optimization

### LINKED ISSUE #6418: SelectOneMenu: performance-optimization
Especially for SelectOneMenu´s with a huge amount of SelectItems. (hundreds to thousands)

some ideas:

- [x] SelectOneMenuRenderer#encodeSelectItems; improve selected-check
- [x] SelectOneMenuRenderer renders currently encodeHiddenSelect and encodePanelContent; instead render only encodeHiddenSelect on the server; use this to render the panel on the client with JS; with custom content keep it as-is

christophs78:
1. optimizing SelectOneMenuRenderer#encodeSelectItems especially helps for dynamic="true"
2. Only doing encodeHiddenSelect on server and move encodePanelContent to the client results in way smaller HTML we need to send from server to client. And on client we have less DOM until the users enters SelectOneMenu.

### DIFF
--- a/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
+++ b/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
@@ -370,12 +370,15 @@ public class SelectOneMenuRenderer extends SelectOneRenderer {
             writer.endElement("table");
         }
         else {
+            // TODO: do with JS
+            /*
             writer.startElement("ul", null);
             writer.writeAttribute("id", menu.getClientId(context) + "_items", null);
             writer.writeAttribute("class", SelectOneMenu.LIST_CLASS, null);
             writer.writeAttribute("role", "listbox", null);
             encodeOptionsAsList(context, menu, selectItems);
             writer.endElement("ul");
+             */
         }
     }
 
@@ -623,10 +626,16 @@ public class SelectOneMenuRenderer extends SelectOneRenderer {
         ResponseWriter writer = context.getResponseWriter();
 
         if (option instanceof SelectItemGroup) {
+            //TODO: https://developer.mozilla.org/de/docs/Web/HTML/Element/optgroup
+
             SelectItemGroup group = (SelectItemGroup) option;
+
+            writer.startElement("optgroup", null);
+            writer.writeAttribute("label", group.getDescription(), null);
             for (SelectItem groupItem : group.getSelectItems()) {
                 encodeOption(context, menu, groupItem, selectedOption, values, submittedValues, converter, itemIndex);
             }
+            writer.endElement("optgroup");
         }
         else {
             boolean disabled = option.isDisabled();
--- a/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
+++ b/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
@@ -633,7 +633,10 @@ public class SelectOneMenuRenderer extends SelectOneRenderer {
             SelectItemGroup group = (SelectItemGroup) option;
 
             writer.startElement("optgroup", null);
-            writer.writeAttribute("label", group.getDescription(), null);
+            writer.writeAttribute("label", group.getLabel(), null);
+            if (group.isDisabled()) {
+                writer.writeAttribute("disabled", "disabled", null);
+            }
             for (SelectItem groupItem : group.getSelectItems()) {
                 encodeOption(context, menu, groupItem, selectedOption, values, submittedValues, converter, itemIndex);
             }
--- a/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
+++ b/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
@@ -24,7 +24,6 @@
 package org.primefaces.component.selectonemenu;
 
 import java.io.IOException;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.logging.Level;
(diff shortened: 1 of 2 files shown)


christophs78:
Think/hope it´s almost done. I have to remove some `LOGGER.log` from `SelectOneMenuRenderer`, but otherwise i´m almost happy with the current state of things.
I´ll like to hear some feedback. :-)

melloware:
Not sure if you have done any kind of performance Testing with say 4000 random items to determine performance gain?

christophs78:
Generated HTML-code is reduced somewhere between 50 und 70%. And we save server-CPU and reduce GC caused by SelectOneMenuRenderer.
I´ll provide more numbers.

tandraschko:
Browser + Server times would be interessting

melloware:
I am always a big fan of making the browser do more work than server but just want to make sure we aren't lagging on browser side.

christophs78:
Usecase:
image

Before:
1.0MB, ~150ms (after multiple runs)
image

After:
299KB, ~20ms (after multiple runs)
image

Performance also depends heavily on converter usage. eg https://showcase.omnifaces.org/converters/SelectItemsIndexConverter has absolute worse performance when used with p:SelectOneMenu and a huge amount of items. (CPU-time grows exponential.)

Client-numbers follow.

christophs78:
Client numbers: Edge 86

Duration when opening SelectOneMenu first time.

Before:
image

After:
image

When i analyze this right we add ~230ms for client-side rendering (renderPanelContentFromHiddenSelect)
image
Maybe i´m able to further optimize renderPanelContentFromHiddenSelect.

christophs78:
Initial-Rendering:

Before:
image

After:
image

christophs78:
Some initial thoughts?
IMO we see some pros and some cons. I´ll dig further into the client side.

tandraschko:
So Initial page load is faster both Server and Client, Just opening the overlay is slower?

christophs78:
> So Initial page load is faster both Server and Client, Just opening the overlay is slower?

Yes

tandraschko:
then +1, great work
just some integrationtests before would be nice

src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
```diff
@@ -143,10 +152,17 @@ protected void encodeMarkup(FacesContext context, SelectOneMenu menu) throws IOE
...
         encodeInput(context, menu, clientId, selectItems, values, submittedValues, converter);
+        long t2 = System.nanoTime();
         encodeLabel(context, menu, selectItems);
         encodeMenuIcon(context, menu, valid);
+        long t3 = System.nanoTime();
         encodePanel(context, menu, selectItems);
+        long t4 = System.nanoTime();
+
+        LOGGER.log(Level.INFO, "SelectOneMenuRenderer#encodeInput " + ((t2 - t1) / 1000000));
+        LOGGER.log(Level.INFO, "SelectOneMenuRenderer#encodePanel " + ((t4 - t3) / 1000000));
```
melloware: I think its safe to get rid of these loggers and system.nanotime calls?

src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
```diff
@@ -105,8 +108,14 @@ public void encodeEnd(FacesContext context, UIComponent component) throws IOExce
...
             Object submittedValues = getSubmittedValues(menu);
 
+            long t1 = System.nanoTime();
             encodeHiddenSelect(context, menu, clientId, selectItems, values, submittedValues, converter);
+            long t2 = System.nanoTime();
             encodePanelContent(context, menu, selectItems);
+            long t3 = System.nanoTime();
+
+            LOGGER.log(Level.INFO, "SelectOneMenuRenderer#encodeHiddenSelect " + ((t2 - t1) / 1000000));
+            LOGGER.log(Level.INFO, "SelectOneMenuRenderer#encodePanelContent " + ((t3 - t2) / 1000000));
```
melloware: Remove LOGGER and System.nanoTime calls?

melloware (changes requested):

src/main/resources/META-INF/resources/primefaces/forms/forms.selectonemenu.js
```diff
@@ -94,16 +95,17 @@ PrimeFaces.widget.SelectOneMenu = PrimeFaces.widget.DeferredWidget.extend({
...
         this.itemsWrapper = this.panel.children('.ui-selectonemenu-items-wrapper');
-        this.options = this.input.children('option');
+        this.options = this.input.find('option');
         this.cfg.effect = this.cfg.effect||'fade';
 
         this.cfg.effectSpeed = this.cfg.effectSpeed||'normal';
         this.cfg.autoWidth = this.cfg.autoWidth === false ? false : true;
         this.cfg.dynamic = this.cfg.dynamic === true ? true : false;
         this.cfg.appendTo = this.getAppendTo();
+        this.cfg.renderPanelContentOnClient = this.cfg.renderPanelContentOnClient === true ? true : false;
```
melloware: can't you just do:
```javascript
 this.cfg.renderPanelContentOnClient = this.cfg.renderPanelContentOnClient === true;
```

src/main/resources/META-INF/resources/primefaces/forms/forms.selectonemenu.js
```diff
@@ -1298,10 +1303,105 @@ PrimeFaces.widget.SelectOneMenu = PrimeFaces.widget.DeferredWidget.extend({
...
+     * @private
+     * @param {JQuery} parentItem An parentItem (select, optgroup) for which to render HTML-code.
+     * @return {string} Rendered HTML-code.
+     */
+    renderSelectItems: function(parentItem) {
+        var content = "";
+
+        var opts = parentItem.children("option, optgroup");
+
+        opts.each((i, opt) => {
```
melloware: This is ES6 syntax is it not?  I think you will have to write it as traditional ES5 javascript.

melloware (changes requested):

src/main/resources/META-INF/resources/primefaces/forms/forms.selectonemenu.js
christophs78: done

src/main/resources/META-INF/resources/primefaces/forms/forms.selectonemenu.js
christophs78: Done. Maybe we should rethink after IE11 is gone whether we relay want to stick to ES5 forever.

christophs78:
@tandraschko : Integrationtest-PR and propably PF Selenium - PR follows later this week. I´ll realy wish some other people also write integrationtest. Currently it look like all other skip this due to a lack of time or whatever.

melloware (approved):

PR #6429, opened 2020-10-20
Fix #6420: ScheduleRenderer encodeEnd cleanup

### LINKED ISSUE #6420: Schedule: fix encodeEnd behavior
See https://github.com/primefaces/primefaces/pull/6417

Can someone take this?

melloware:
- Fixed Line Ending from DatePicker commit.
- @tandraschko I only check for LazyModel and `isRequestSource` without the event name as it seems multiple events can trigger this redraw in Schedule?

When I look at the Form Data for Lazy schedule when pressing say the Previous or Next Day buttons.
```
javax.faces.partial.ajax: true
javax.faces.source: j_idt729:j_idt731
javax.faces.partial.execute: j_idt729:j_idt731
javax.faces.partial.render: j_idt729:j_idt731
j_idt729:j_idt731: j_idt729:j_idt731
j_idt729:j_idt731_start: 2020-11-01T04:00:00.000Z
j_idt729:j_idt731_end: 2020-12-13T05:00:00.000Z
j_idt729: j_idt729
j_idt729:j_idt731_view: dayGridMonth
javax.faces.ViewState: -6890984407278275305:9217892359940988958
```
There is no `javax.faces.behavior.event` for many of these Schedule calls.


src/main/java/org/primefaces/component/schedule/ScheduleRenderer.java
```diff
@@ -66,7 +66,7 @@ public void decode(FacesContext context, UIComponent component) {
     public void encodeEnd(FacesContext context, UIComponent component) throws IOException {
         Schedule schedule = (Schedule) component;
 
-        if (context.getExternalContext().getRequestParameterMap().containsKey(schedule.getClientId(context))) {
+        if (schedule.getValue() instanceof LazyScheduleModel && ComponentUtils.isRequestSource(schedule, context)) {
             encodeEvents(context, schedule);
```
melloware, line 70: This is the real fix the other two files are just EOL fixes.

tandraschko:
we need a identifier (some components also just use a special request param, often like clientId_someCase), to only encodeEvents when the JS client expects that.

melloware:
OK let me look at that.

melloware:
OK added an `_event` type param since these are all Schedule Event based changes and check for that.

tandraschko:
I would still do ComponentUtils.isRequestSource && schedule.isEventRequest(context)

and remove the lazy check. If you check encodeEvents, it doesnt require the LazyModel

melloware:
Done when I looked through the code it looks like it was always checking for Lazy in the encodeEvents code and just returns an empty JSON array if not lazy so i thought it was safe but you are right this way is better.

PR #6432, opened 2020-10-20
Fix #6430: DataExporter column alignment

### LINKED ISSUE #6430: DataExporter: Align cells based on column alignment
**Description**
Another feature that PFE Exporter has that needs to be migrated is the ability to right, center, and left align cells in PDF and Excel exports.

See: https://www.primefaces.org/showcase-ext/sections/exporter/columnAlignments.jsf

Right now I don't like the way PFE does it assuming that your column is `<h:outputText>` and checking its styles.  It should check the `<p:column>` itself.

**Example XHTML**

If the column has a `style` containing the word "right" or a `styleClass` containing the word "right" then right align the cell.

Using Style...
```xml
<p:column style="text-align:right"
```

Using styleClass...
```xml
<p:column styleClass="column-right"
```

melloware:


tandraschko:
migration guide is missing + conflicts

melloware:
Conflicts resolved and migration guide updated.

PR #6438, opened 2020-10-22
Fix #4840: Datatable attribute 'renderEmptyFacets'

### LINKED ISSUE #4840: DataTable: Cannot update footer only as initial footer is not rendered anymore
## 1) Environment
- PrimeFaces version: `7.0`
- Application server + version: `Tomcat 8.5.37`
- Affected browsers: tested with `Firefox` and `Edge`

## 2) Expected behavior
`<h:panelGroup id="buttonPanel">` component in `<f:facet name="footer">` should be updateable (without updating the whole datatable) even if in the initial state all `<p:commandButton>` components are not rendered.

## 3) Actual behavior
I need to update the whole datatable to have an updated footer with buttons based on the selected row.

## 4) Steps to reproduce
Extend showcase `Single with Row Click` (https://www.primefaces.org/showcase/ui/data/datatable/selection.xhtml) by showing the button only if a row is selected.
We therefore currently add a named container, e.g. `h:panelGroup` which can be separately updated and accordingly add row select and unselect events.

## 5) Sample XHTML
```
<p:dataTable id="singleDT" var="car" value="#{dtSelectionView.cars2}" selectionMode="single" selection="#{dtSelectionView.selectedCar}" rowKey="#{car.id}">
    <f:facet name="header">
        Single with Row Click
    </f:facet>
    <p:ajax event="rowSelect" update="buttonPanel" />
    <p:ajax event="rowUnselect" update="buttonPanel" />
    <p:column headerText="Id">
    </p:column>
    <p:column headerText="Year">
    </p:column>
    <p:column headerText="Brand">
    </p:column>
    <p:column headerText="Color">
    </p:column>
    <f:facet name="footer">
        <h:panelGroup id="buttonPanel">
            <p:commandButton process="singleDT" update=":form:carDetail" icon="pi pi-search" value="View" oncomplete="PF('carDialog').show()" rendered="#{dtSelectionView.selectedCar != null}" />
        </h:panelGroup>
    </f:facet>
</p:dataTable>
```

## 6) Sample bean
unchanged to showcase.

This is probably caused by https://github.com/primefaces/primefaces/issues/2825. Our current solution still worked with PrimeFaces 6.1.

### LINKED ISSUE #6456: Performance: Use getChildCount() instead of size/empty
**Description**
Do not use `size()` or `isEmpty()` for any child counting using `getChildCount()` instead.

melloware:
I went with 'renderEmptyFacets` and I overloaded the ComponentUtils method so we could reuse this on other future components if necessary.

### DIFF
--- a/src/main/java/org/primefaces/util/ComponentUtils.java
+++ b/src/main/java/org/primefaces/util/ComponentUtils.java
@@ -470,11 +470,12 @@ public class ComponentUtils {
     }
 
     /**
-     * Checks if the facet and one of the first level child's is rendered.
+     * Checks if the facet and one of the first level children is rendered.
      * @param facet The Facet component to check
-     * @return true when facet and one of the first level child's is rendered.
+     * @param ignoreChildren flag to ignore children and only check the facet itself
+     * @return true if the facet should be rendered, false if not
      */
-    public static boolean shouldRenderFacet(UIComponent facet) {
+    public static boolean shouldRenderFacet(UIComponent facet, boolean ignoreChildren) {
         if (facet == null || !facet.isRendered()) {
             // For any future version of JSF where the f:facet gets a rendered attribute (https://github.com/javaserverfaces/mojarra/issues/4299)
             // or there is only 1 child.
@@ -482,11 +483,20 @@ public class ComponentUtils {
         }
 
         // Facet has no child but is rendered
-        if (facet.getChildren().isEmpty()) {
+        if (facet.getChildren().isEmpty() || ignoreChildren) {
             return true;
         }
 
         return shouldRenderChildren(facet);
+    }
+
+    /**
+     * Checks if the facet and one of the first level children is rendered.
+     * @param facet The Facet component to check
+     * @return true when facet and one of the first level children is rendered.
+     */
+    public static boolean shouldRenderFacet(UIComponent facet) {
+        return shouldRenderFacet(facet, false);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/main/java/org/primefaces/util/ComponentUtils.java
```diff
@@ -470,25 +470,35 @@ public static boolean isRequestSource(UIComponent component, FacesContext contex
...
+    public static boolean shouldRenderFacet(UIComponent facet, boolean ignoreChildren) {
         if (facet == null || !facet.isRendered()) {
             // For any future version of JSF where the f:facet gets a rendered attribute (https://github.com/javaserverfaces/mojarra/issues/4299)
             // or there is only 1 child.
             return false;
         }
 
         // Facet has no child but is rendered
-        if (facet.getChildren().isEmpty()) {
+        if (facet.getChildren().isEmpty() || ignoreChildren) {
```
tandraschko: switch the statements

tandraschko: also always use facet.getChildCount() instead of facet.getChildren().size()/isEmpty()

melloware:
Good catch fixed in 2 places.

buunsal:
If facet is really empty or for example <h: outputText rendered = "false" /> used in facet, facet will not render even if 'renderEmptyFacet' attribute is true. It needs to have a component inside facet and rendered = "false" to work properly. I added isRenderEmptyFacet to the initial if case to work properly in the panel component: https://github.com/primefaces/primefaces/pull/7207

Example Code:
```
<p:dataTable renderEmptyFacets="true">
    <f:facet name="header">
        <h:outputText value="Header" rendered="false"/>
    </f:facet>

</p:dataTable>
```

Can we do the same within the datatable component? WDYT @melloware @tandraschko

melloware:
I am confused why it wouldn't render?  Because it checks if it should check the children or not and if renderemptyfacets=true the above case should still render the facet?

PR #6444, opened 2020-10-22
Fix #6442: InputNumber 'inputMode' attribute for touch devices

### LINKED ISSUE #6442: InputNumber: Open numerical keyboard default on touch devices
**Description**
The inputNumber component default opens an alphabetic keyboard on touch devices even though it only permits numerical input.

The inputNumber component should default open a numerical keyboard

See: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode

melloware:
InputMode is supported by most major browses now: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode

This sets inputMode to 'numeric' if DecimalPlaces==0 and 'decimal' if DecimalPlaces>0 or allows the user to override to 'none' if no behavior is desired.


tandraschko:
Why not just passthrough?

melloware:
It could definitely be a pass through but I think this is better default behavior and expected behavior of the component on mobile devices. And since we know their decimal places can choose the right keyboard for them.

PR #6448, opened 2020-10-23
Fix #6447: InputNumber browser autofill

### LINKED ISSUE #6447: InputNumber: Browser AutoFill not working
**Describe the defect**
If you input a value like "55,55" into an element and submit the form, the browser displays a dropdown select the next time you visit the page and double click the same element.
When you select the previously entered value "55,55" The value is pasted, but as soon as the focus is lost on the input element, autoNumeric deletes the value.

See AutoNumeric defect: https://github.com/autoNumeric/autoNumeric/issues/536

**Expected behavior**
If you input a value like "55,55" into an element and submit the form, the browser displays a dropdown select the next time you visit the page and double click the same element.
When you select the previously entered value "55,55" it is pasted into the element.

**NOTE:** One of our production clients reported this defect.

melloware:
We have tested this fix in our production application and verified it resolves the issue based on the AutoNumeric fix described in the issue ticket.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js
+++ b/src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js
@@ -50,6 +50,7 @@ PrimeFaces.widget.InputNumber = PrimeFaces.widget.BaseWidget.extend({
         PrimeFaces.skinInput(this.input);
 
         this.wrapEvents();
+        this.bindInputEvents();
 
         this.autonumeric = new AutoNumeric(this.jqId + '_input', this.cfg);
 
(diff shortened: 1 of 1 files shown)


melloware:
Still investigating.

melloware:
Discovered during testing that the old `input` event we were using for MouseWheel and Paste is no longer needed as its built into the newer AutoNumeric already.   So we could remove that for a single `blur` event to address  the browser autofill issue.

tandraschko:
+1 if well tested

melloware:
Our QA team discovered this in PRD and have thoroughly tested this fix in a large application.  Thanks for reviewing!

PR #6450, opened 2020-10-24
#6449: DatePicker: Metadata add styleClass

### LINKED ISSUE #6449: DatePicker: Metadata: add styleClass

jepsar:
Fix #6449


docs/9_0/components/datepicker.md
```diff
@@ -293,11 +293,12 @@ private DateMetadataModel lazyModel;
 @PostConstruct
 public void init() {
     DefaultDateMetadata metadataDisabled = DefaultDateMetadata.builder().disabled(true).build();
+    DefaultDateMetadata metaDataDeadline = DefaultDateMetadata.builder().styleClass("deadline").build();
```
tandraschko: Metadata

src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
```diff
@@ -302,12 +302,19 @@ PrimeFaces.widget.DatePicker = PrimeFaces.widget.BaseWidget.extend({
...
                         var disabledDates = [];
+                        var dateStyleClasses = {};
                         for (date in dateMetadata) {
+                            var parsedDate = pdp.parseOptionValue(date);
                             if (dateMetadata[date].disabled) {
-                                disabledDates.push(date);
+                                disabledDates.push(parsedDate);
+                            }
+                            if (dateMetadata[date].styleClass) {
+                              dateStyleClasses[pdp.toISO(parsedDate)] = dateMetadata[date].styleClass;
```
melloware: Check your spacing here it looks like this line needs 2 more space indent.

melloware (changes requested):

src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
jepsar: I already fixed my setting this weekend. Will fix this one.

src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -1602,6 +1610,10 @@
             return _classes;
         },
 
+        toISO: function (date) {
```
melloware: JSDoc this method what it returns and why you are only returning first 10 chars.

melloware (changes requested):

src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
melloware: Maybe call it `toISOString(date)`

jepsar: > Maybe call it `toISOString(date)`

I'll rename it to `toISODateString` and add documentation. Basically we just care about the date part here and not about the time, so I'm taking only the date part.

melloware (approved):

src/main/java/org/primefaces/model/datepicker/DefaultDateMetadata.java
```diff
@@ -62,12 +75,15 @@ public boolean equals(Object obj) {
             return false;
         }
         final DefaultDateMetadata other = (DefaultDateMetadata) obj;
-        return this.disabled == other.disabled;
+        if (this.disabled != other.disabled) {
+            return false;
+        }
+        return Objects.equals(this.styleClass, other.styleClass);
     }
```
melloware, line 76: Why not just make your equals and hashcode methods...
```java
@Override
    public int hashCode() {
        return Objects.hash(disabled, styleClass);
    }

    @Override
    public boolean equals(Object obj) {
        if (this == obj) {
            return true;
        }
        if (obj == null) {
            return false;
        }
        if (!(obj instanceof DefaultDateMetadata)) {
            return false;
        }
        DefaultDateMetadata other = (DefaultDateMetadata) obj;
        return disabled == other.disabled && Objects.equals(styleClass, other.styleClass);
    }
```

melloware (changes requested):

src/main/java/org/primefaces/model/datepicker/DefaultDateMetadata.java
jepsar: I did not bother writing them. It's how they were generated by NetBeans.

melloware: Ahh mine above were generated by Eclipse.  Interesting!

jepsar: .. doesn't matter, I'll change them

melloware: OK cool yeah its juts that `Objects.hash()` already does all the magic inside of it so its strange Netbeans generates like that.

docs/9_0/components/datepicker.md
jepsar: Ah, that slipped in here by copy pasting it from the showcase.. let me fix it there as well

melloware (approved):

tandraschko:
Am i blind or isnt loadInitialLazyMetadata called?

jepsar:
> Am i blind or isnt loadInitialLazyMetadata called?

It is called:

https://github.com/primefaces/primefaces/pull/6450/files?diff=unified&w=1#diff-6d9ed21fdb2d6506a99c1130235119aa694d1bb61118c0a0e715a93bcc1bd411R209

PR #6452, opened 2020-10-26
Fix #6451: Schedule add 'height' attribute

### LINKED ISSUE #6451: Schedule: height, width, aspectRatio related problem
I am trying to migrate to the latest PF snapshot which works with the fullcalendar version 5. 
However, i am not able to to get it right. I am not sure whether this is a PF or fullcalendar issue?

I have the following use case:

```
        <p:schedule
          id="schedule"
          widgetVar="scheduleWidgetVar"
          value="#{instrumentReservationManager.lazyEventModel}"
          initialDate="#{instrumentReservationManager.initialDate}"
          showWeekends="#{instrumentReservationManager.instrumentReservationSetting.weekends}"
          minTime="#{instrumentReservationManager.instrumentReservationSetting.minTimeFormat}"
          maxTime="#{instrumentReservationManager.instrumentReservationSetting.maxTimeFormat}"
          slotLabelInterval="#{instrumentReservationManager.instrumentReservationSetting.scheduleSlotLabelInterval}"
          slotDuration="#{instrumentReservationManager.instrumentReservationSetting.scheduleSlotDuration}"
          slotEventOverlap="false"
          slotLabelFormat="['HH:mm']"
          columnHeaderFormat="week:'dddd DD MMM', month:'dddd'"
          rightHeaderTemplate="timeGridDay timeGridWeek dayGridMonth"
          leftHeaderTemplate="prev next today"
          tooltip="true"
          showWeekNumbers="false"
          allDaySlot="false"
          nextDayThreshold="00:00"
          styleClass=""
          draggable="true"
          resizable="true"
          view="timeGridWeek"
          extender="initSchedule"/>
        <h:outputScript>
          function initSchedule() {
            // Configure fullCalendar
            this.cfg.buttonText = {today: 'Today'};
            this.cfg.views = {
              week: {
                displayEventTime: false,

              },
              day: {
                displayEventTime: false,
              }
            };
        </h:outputScript>
```
and following problem depicted in the screenshot below: the height of the calendar is not computed right. it always tries to set height according to the aspectRatio. however, i want that the natural height and no scrollbars are used. this should be the case when contentHeight is auto. Howevre, this has no effect.

Screenshot from 2020-10-26 16-05-37

the view actually should loook like as follows (as it looks like with the older PF snapshot where the migration to fullcalendar v5 was not not done.

Screenshot from 2020-10-26 16-34-03

Any hint/fix is highly appreciated. Thanks in advance.

melloware:
Doc: https://fullcalendar.io/docs/height

> Sets the height of the entire calendar, including header and footer.
> 
> Integer, "auto", a CSS value like "100%"
> 
> By default, this option is unset and the calendar’s height is calculated by aspectRatio.
> 
> If an integer is specified, the height of the calendar will be guaranteed to be that exact pixel height. If the contents will not fit within the height, scrollbars will appear.
> 
> If "auto" is specified, the view’s contents will assume a natural height and no scrollbars will be used.
> 
> If "100%" is specified, the height of the calendar will match the height of its parent container element. See an example. Any other valid CSS value is accepted as well.

### DIFF
--- a/docs/9_0/components/schedule.md
+++ b/docs/9_0/components/schedule.md
@@ -26,40 +26,41 @@ widgetVar | null | String | Name of the client side widget.
 value | null | Object | An org.primefaces.model.ScheduleModel instance representing the backed model
 locale | null | Object | Locale for localization, can be String or a java.util.Locale instance
 dir | ltr | String | Defines direction of schedule. Valid values are "ltr" (default) and "rtl".
-aspectRatio | null | Float | Ratio of calendar width to height, higher the value shorter the height is
-view | dayGridMonth | String | The view type to use, possible values are 'dayGridMonth', 'dayGridWeek', 'dayGridDay', 'timeGridWeek', 'timeGridDay', 'listYear' , 'listMonth', 'listWeek', 'listDay'.
-initialDate | null | java.time.LocalDate | The initial date that is used when schedule loads. If ommitted, the schedule starts on the current date
-showWeekends | true | Boolean | Specifies inclusion Saturday/Sunday columns in any of the views
-style | null | String | Style of the main container element of schedule
-styleClass | null | String | Style class of the main container element of schedule
-draggable | true | Boolean | When true, events are draggable.
-resizable | true | Boolean | When true, events are resizable.
-showHeader | true | Boolean | Specifies visibility of header content.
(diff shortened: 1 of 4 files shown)


melloware:
Showcase PR: https://github.com/primefaces/primefaces-showcase/pull/214

PR #6457, opened 2020-10-27
Fix #6456: Perf getChildCount()

### LINKED ISSUE #6456: Performance: Use getChildCount() instead of size/empty

melloware:


src/main/java/org/primefaces/component/organigram/OrganigramHelper.java
```diff
@@ -96,8 +96,8 @@ public static OrganigramNode findTreeNode(OrganigramNode searchRoot, OrganigramN
      * @param node The root node.
      */
     public static void buildRowKeys(OrganigramNode node) {
-        if (node.getChildren() != null && !node.getChildren().isEmpty()) {
-            for (int i = 0; i < node.getChildren().size(); i++) {
+        if (node.getChildren() != null && node.getChildCount() > 0) {
```
tandraschko, line 99: switch those statements

src/main/java/org/primefaces/component/organigram/OrganigramRenderer.java
```diff
@@ -211,7 +211,7 @@ protected void renderNode(FacesContext context, ResponseWriter writer, Map<Strin
         requestMap.remove(organigram.getVar());
 
         // render child nodes
-        if (node.getChildren() != null && !node.getChildren().isEmpty()) {
+        if (node.getChildren() != null && node.getChildCount() > 0) {
```
tandraschko, line 214: switch the statements

src/main/java/org/primefaces/component/organigram/OrganigramHelper.java
melloware: actually do i need to check for NULL children at all of childCount>0 they can't be NULL right?  I think I can just remove the NULL check.

tandraschko: unsure, as its a model? :D i wouldnt remove it, just to be sure

melloware:
I put it back the way it was because Chidlren can be NULL and the getChildCount of that model calls "children.size()" so it could throw an NPE.  For these two Organigram I just left the way it is to prevent NPE.

PR #6463, opened 2020-10-28
Fix #6458: DatePicker defaults before super.encodeEnd

### LINKED ISSUE #6458: DatePicker: showTime not detected automatically for LocalDateTime
In the migration document (8->9), it is stated that the attribute showTime is detected automatically based on value-binding to LocalDateTime. I cannot nor verify this for the latest PF9-snapshot. Could somebody please check and fix this. Many thanks in advance.

melloware:
Basically what was happening was in `BaseCalendarRenderer.encodeEnd` decisions were being made with configs before calling `DatePicker.encodeMarkup`.

So we had to move this default setting to the `encodeEnd` before calling `super.encodeEnd` so the proper values for `showTime`, `showSeconds` etc are calculated before it tries to build the default markup date.

@christophs78 @jepsar 

I have tested both the Showcase and the provided test case from user.  We also have our Primefaces Integration Tests that will run.

I will add an Integration Test case there.

### DIFF
--- a/src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java
+++ b/src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java
@@ -54,12 +54,38 @@ public class DatePickerRenderer extends BaseCalendarRenderer {
 
     @Override
     public void encodeEnd(FacesContext context, UIComponent component) throws IOException {
-        DatePicker datePicker = (DatePicker) component;
+        DatePicker datepicker = (DatePicker) component;
 
-        if (datePicker.getModel() instanceof LazyDateMetadataModel && ComponentUtils.isRequestSource(datePicker, context, "viewChange")) {
-            encodeDateMetadata(context, datePicker);
+        if (datepicker.getModel() instanceof LazyDateMetadataModel && ComponentUtils.isRequestSource(datepicker, context, "viewChange")) {
+            encodeDateMetadata(context, datepicker);
             return;
         }
+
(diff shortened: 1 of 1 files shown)


src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java
```diff
@@ -54,12 +54,38 @@
 
     @Override
     public void encodeEnd(FacesContext context, UIComponent component) throws IOException {
-        DatePicker datePicker = (DatePicker) component;
+        DatePicker datepicker = (DatePicker) component;
```
tandraschko: why did you rename? its actually datePicker

melloware: Its mixed throughout the file with datepicker and datePicker i will fix it every where to be datePicker.

melloware:
Integration Test updated: https://github.com/primefaces-extensions/primefaces-integration-tests/commit/d014a7ec8d0553e95f4b14fb5b856f50b616e1af

melloware:
OK i renamed lowercase `datepicker` to proper `datePicker` throughout the whole file.

christophs78:
Makes sense. Thank´s for analyzing!

melloware:
New Integrations tests as now passing after this fix!

jepsar:
Merged to https://github.com/primefaces/primefaces/pull/6450

PR #6466, opened 2020-10-29
Migration Guide Updates

melloware:
- Moved migration guides outside of versions like we have for JSDOCS
- Referenced the migration guide as links to outside docs so can be shared across versions
- Added 6.0 and 6.1 Migration Guides
- Fixed Datatable issue with 7.0 Migration guide


tandraschko:
thats a nice idea!

PR #6471, opened 2020-11-02
Fix #6469: InputMask optional character fix

### LINKED ISSUE #6469: InputMask: optional part does now work correctly
@melloware with the change from the april to latest snapshot, p:inputMask is not working anymore correctly at least for mask with optional cases.

example:

            <p:inputMask
              value="#{user.phoneNumber.areaCode}"
              mask="9[999]"
              size="4" />

previously it was working well as follows:

            <p:inputMask
              value="#{user.phoneNumber.areaCode}"
              mask="9?999"
              size="4" />

can you please check and fix this.

when i try to save, it clears the inputMask fields (see screenshot)

Screenshot from 2020-11-02 13-33-45

melloware:
When using optional characters the value submitted to the server side contains the remaining mask which must be removed.

So when submitting "12__" it must be "12"

Integration Test: https://github.com/primefaces-extensions/primefaces-integration-tests/commit/147341e414c61e0db84b70ad9c3af578735d0688

### DIFF
--- a/docs/9_0/components/inputmask.md
+++ b/docs/9_0/components/inputmask.md
@@ -76,8 +76,8 @@ style | null | String | Inline style of the input element.
 styleClass | null | String | Style class of the input element.
 tabindex | null | Integer | Position of the input element in the tabbing order.
 title | null | String | Advisory tooltip information.
-validateMask | false | Boolean | Defines whether mask pattern would be validated or not on the server side.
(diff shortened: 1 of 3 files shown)


melloware:
Integration test properly failed as expected now going to merge this fix and re-run the integration test proving it fixes the issue

PR #6472, opened 2020-11-02
Sonar Fixes

melloware:
@blutorange please review these changes as they touch your code.

### DIFF
--- a/src/main/java/org/primefaces/component/datatable/export/DataTableCSVExporter.java
+++ b/src/main/java/org/primefaces/component/datatable/export/DataTableCSVExporter.java
@@ -31,7 +31,6 @@ import java.util.List;
 
 import javax.faces.FacesException;
 import javax.faces.component.UIComponent;
-import javax.faces.context.ExternalContext;
 import javax.faces.context.FacesContext;
 
 import org.primefaces.component.api.DynamicColumn;
@@ -65,8 +64,6 @@ public class DataTableCSVExporter extends DataTableExporter {
     public void doExport(FacesContext context, DataTable table, ExportConfiguration exportConfiguration, int index) throws IOException {
         try (OutputStreamWriter osw = new OutputStreamWriter(getOutputStream(), exportConfiguration.getEncodingType());
             PrintWriter writer = new PrintWriter(osw);) {
-
-            ExternalContext externalContext = context.getExternalContext();
 
             if (exportConfiguration.getPreProcessor() != null) {
                 // PF 9 - attention: breaking change to PreProcessor (PrintWriter instead of writer)
--- a/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
+++ b/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
@@ -23,9 +23,13 @@
  */
 package org.primefaces.component.datatable.export;
 
-import java.io.*;
+import java.io.IOException;
+import java.io.OutputStream;
 import java.lang.reflect.Array;
-import java.util.*;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
 import java.util.stream.Collectors;
 
 import javax.el.MethodExpression;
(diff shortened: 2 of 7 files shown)


src/main/java/org/primefaces/component/datatable/export/DataTableCSVExporter.java
```diff
@@ -66,8 +65,6 @@ public void doExport(FacesContext context, DataTable table, ExportConfiguration
         try (OutputStreamWriter osw = new OutputStreamWriter(getOutputStream(), exportConfiguration.getEncodingType());
             PrintWriter writer = new PrintWriter(osw);) {
 
-            ExternalContext externalContext = context.getExternalContext();
-
```
melloware, line 70: Dead variable according to Sonar.

src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
```diff
@@ -337,8 +342,6 @@ public void export(FacesContext context, List<DataTable> tables, OutputStream ou
      */
     protected abstract void doExport(FacesContext facesContext, DataTable table, ExportConfiguration exportConfiguration, int index) throws IOException;
 
-    public abstract String getFileExtension();
-
```
melloware, line 341: Already declared in Interface.

src/main/java/org/primefaces/convert/CombinedDateTimePatternTokenVisitor.java
```diff
@@ -99,8 +99,6 @@ public void visitTokenLetter(int letter, int repetitions) {
 
             // time zone
             case 'z':
-                visitZonezToken(repetitions, builder);
-                break;
             case 'Z':
```
melloware, line 102: Made this a fall through since both 'z' and 'Z' do the same thing.

src/main/java/org/primefaces/convert/CombinedDateTimePatternTokenVisitor.java
```diff
@@ -175,13 +175,12 @@ private static void visitDayOfWeekToken(int repetitions, JQueryDateTimePatternBu
...
-            default: // 2+
-                builder.appendDayOfYearWithLeadingZeros();
-                break;
+        if (repetitions == 1) {
+            builder.appendDayOfYearNoLeadingZeros();
+        }
+        else {
+            // 2+
+            builder.appendDayOfYearWithLeadingZeros();
         }
```
melloware, line 184: Sonar recommended removing the switch for if else for readability

src/main/java/org/primefaces/component/datatable/export/DataTableCSVExporter.java
blutorange: Pretty sure I haven't worked on the DataTableCSVExporter, but yes, the variable `externalContext` is not used anywhere.

src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
blutorange: Again haven't worked on DataTableExporter, but also yes, that declaration can be removed.

src/main/java/org/primefaces/convert/CombinedDateTimePatternTokenVisitor.java
blutorange: I made a separate method for each case, and for 'z' and 'Z' the JQueryUI pattern happens to be the same for both. Makes the code a tiny bit easier to read imo, but I've got no problem with this change.

src/main/java/org/primefaces/convert/CombinedDateTimePatternTokenVisitor.java
blutorange: I used the switch-pattern in other methods too so always using that keeps it consistent, but yeah, this is fine as well.

src/main/java/org/primefaces/component/datatable/export/DataTableCSVExporter.java
christophs78: Yes, should be removed.

src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
christophs78: Yes, should be removed.

blutorange:
@melloware  Thanks for fixing these issues, looks good to me : )

PR #6476, opened 2020-11-03
#6404: DataExporter: update migration guide

christophs78:


docs/migrationguide/9_0.md
```diff
@@ -65,7 +65,11 @@
 ## Cache
   * HazelCast now requires version 4.X see: (https://github.com/primefaces/primefaces/issues/5989)
 
-## Exporter
+## (Data)Exporter
```
melloware: I think its safe to remove the parethesis since DataExporter is the name of the component.

melloware (changes requested):

docs/migrationguide/9_0.md
christophs78: Done

melloware (approved):

PR #6477, opened 2020-11-03
Fix #6474: Timeline added subgroup attributes to model

### LINKED ISSUE #6474: Timeline: subgroups
**Description**

It would be cool if the widget supported the `subgroups` functionality. You optionally provide a subgroup id

subgroup | String or Number | none | The id of a subgroup. Groups all items within a group per subgroup, and positions them on the same height instead of staking them on top of each other. can be ordered by specifying the option subgroupOrder of a group.
-- | -- | -- | --

and able to configure the following options:

subgroupOrder | String or Function | none | Order the subgroups by a field name or custom sort function. By default, groups are ordered by first-come, first-show.
-- | -- | -- | --
subgroupStack | Object or Boolean | none | Enables stacking within individual subgroups. Example: {'subgroup0': true, 'subgroup1': false, 'subgroup2': true} For each subgroup where stacking is enabled, items will be stacked on top of each other within that subgroup such that they do no overlap. If set to true all subgroups will be stacked. If a value was specified for the order parameter in the options, that ordering will be used when stacking the items.
subgroupVisibility | Object | none | Ability to hide/show specific subgroups. Example: {'hiddenSubgroup0': false, 'subgroup1': true, 'subgroup2': true} If a subgroup is missing from the object, it will default as true (visible).

(the tables were taken from https://visjs.github.io/vis-timeline/docs/timeline/)

melloware:


tandraschko:
why do you add it in the migration guide? thats nothing that breaks/influces existing applications?

melloware:
Got it so Should it go in What's New?

tandraschko:
unsure... 
i would only post "big tickets" there and link to github issue tracker maybe

melloware:
OK removed migration guide change

ssshammi:
Hi I am using version   "primereact": "7.1.0", do you think I will be able to use subgroups or i need to use the latest version. I dont want to lose my work by upgrading. Just this component is that i need.

melloware:
> Hi I am using version "primereact": "7.1.0", do you think I will be able to use subgroups or i need to use the latest version. I dont want to lose my work by upgrading. Just this component is that i need.

@ssshammi this is PrimeFaces JSF you want the PrimeReact issues page.

PR #6479, opened 2020-11-04
OWASP Dependency Check 6.0.3

melloware:


melloware:
Looks like OWASP is having trouble reaching Sonatype right now.  Will try to kick off this build again later in the day.

PR #6491, opened 2020-11-08
#6490: SelectRenderer: add fast path to doValidateSubmittedValues

### LINKED ISSUE #6490: SelectRenderer: add fast path to doValidateSubmittedValues
`SelectRenderer#doValidateSubmittedValues` eats up a good amount of CPU-cycles. Especially with explicit converter.
We can stop evaluating when `submittedValues.length == validSubmittedValues.size()`.

christophs78:


tandraschko:
can you explain the fix?

christophs78:
> can you explain the fix?

After we have validated value for each submitted value there is no need to further loop through the other available SelectItems.

tandraschko:
couldnt it be that the size could be different and still valid?

christophs78:
When eg 3 values have bean submitted we IMO can have a maximum of 3 valid values.
After we found 3 valid values we can stop iterating SelectItems.
When we find eg only 2 valid values everything is like before. We return valid values after we iterated over all SelectItems.

christophs78:
Hm. It´s more complicated when we have oldValues which are now disabled. I have to take this into account.

tandraschko:
i will close it for now, please come back if you have a second idea/PR.

PR #6495, opened 2020-11-09
Fix #3212: Widget lifecycle events

### LINKED ISSUE #3212: Widget Core: lifecycle events
This is a feature request for an upcoming PrimeFaces version (maybe already 6.2?). 

Currently you have to come up with an own logic on how to invoke some javascript function when a component is rendered / initialized / updated / destroyed / etc. This could be tricky if it is a DeferredWidget.

It would be a nice feature if all widget implement some kind of lifecycle callbacks so you only have to add the desired javascript function as an attribute to the widget. Possible callbacks could be: onInit, onRender, onUpdate, onRemove. Maybe the callbacks could even be split into _before_ and _after_ calls (onBeforeInit => at the start of init(); onAfterInit => at the end of init()).

melloware:


docs/9_0/core/ajaxRendering.md
```diff
@@ -148,4 +148,17 @@ This can be easily archived by attaching `p:autoUpdate` to the component:
 
 ### Skipping auto update
 
-You can also skip all auto-updates by setting _ignoreAutoUpdate_ to `true` on the trigger component (like `p:ajax`)
\ No newline at end of file
+You can also skip all auto-updates by setting _ignoreAutoUpdate_ to `true` on the trigger component (like `p:ajax`)
```
tandraschko, line 151: i wouldnt add this in "ajax rendering", thats not related to AJAX directly
of course refresh and destroy is only called on AJAX but postConstruct is always called

tandraschko:
the codechanges looks good from a rough fast short review

docs/9_0/core/ajaxRendering.md
melloware: Any preference on where to put it in the docs?  I couldn't quite figure out the best place.

melloware: I am actually thinking under core we need a whole new page title "Widgets".  I think there is a lot of info about Widgets we can add there.

tandraschko:
is it SPI? Maybe a "global integration points" page or something? not sure

melloware:
OK for now I created a new `widgets.md` under Core and added some basic Widget information including this new section on LifeCycle events.

tandraschko:
Thanks, perfect!

PR #6498, opened 2020-11-10
Fix #6083: Add version to theme.css URL

### LINKED ISSUE #6083: Theme: CSS is loaded without version

melloware:
Resubmitting this PR

### DIFF
--- a/src/main/java/org/primefaces/application/resource/PrimeResourceHandler.java
+++ b/src/main/java/org/primefaces/application/resource/PrimeResourceHandler.java
@@ -28,10 +28,12 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+
 import javax.faces.application.Resource;
 import javax.faces.application.ResourceHandler;
 import javax.faces.application.ResourceHandlerWrapper;
 import javax.faces.context.FacesContext;
+
 import org.primefaces.application.resource.barcode.BarcodeHandler;
 import org.primefaces.util.Constants;
 import org.primefaces.util.LangUtils;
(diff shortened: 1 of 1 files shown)


tandraschko:
i think this is not correct if you use a premium theme, right?
Themes in core.jar can append the PF version but not external ones as the release lifecycle is different

melloware:
Correct this would only apply to a theme library that starts with "primefaces". So all of the themes included with PF like "primefaces-nova-light" and "primfaces-omega" etc.

At our production client we have two themes and we kept that pattern of "primefaces-client-dark" and "primefaces-client-light". 

However for some premium themes like Babylon the library name is "babylon-layout" so this would not pick it up.  I wasn't trying to solve the premium theme issue i was trying to solve the themes included with PF problem.

tandraschko:
what about pfe?

melloware:
You mean would this interfere with PFE?  I have tested and it does not interfere at all.

tandraschko:
i think would append the PF version to PFE?

melloware:
It does so now in PFE 9.0 I also changed it version from `v=` to `e=`.

So for PF/PFE 9.0 URL's look like this...for example if PF was 9.0 community and PFE was 9.0.1 release the URL would look like this

`/primeext-showcase/javax.faces.resource/fa/font-awesome.css.jsf?ln=primefaces&v=9.0&e=9.0.1`

So now we get cache busting whenever you update your PF or PFE version which to me is ideal.

tandraschko:
I would not pass the locale, it's the default behavior of "toLowerCase" without params anyway

melloware:
Removed `Locale` from toLowercase.

PR #6504, opened 2020-11-12
#6418: SelectOneMenu: performance-optimization; refactor dynamic-optimization

### LINKED ISSUE #6418: SelectOneMenu: performance-optimization

christophs78:
The backside of my optimization regarding dynamic=true last month had been more converter-calls.
So i rolled back part of my code-changes and tried to take a different optimization-route.

I also added an additional integration-test to do my best to ensure nothing breaks. :-)
https://github.com/primefaces-extensions/primefaces-integration-tests/pull/63

### DIFF
--- a/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
+++ b/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
@@ -24,7 +24,6 @@
 package org.primefaces.component.selectonemenu;
 
 import java.io.IOException;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


tandraschko:
would be great if you can explain your change

christophs78:
One of my original notes was SelectOneMenu with dynamic=true uses almost the same amount of CPU like SelectOneMenu with dynamic=false for the initial request. Because SelectOneMenu still iterates over all SelectItems, call´s the Converter for each SelectItem, ... (Background is dynamic=true need´s to render its invisible HTML-select with one HTML-option-child for the currently selected SelectItem.)
The only difference are less bytes which need to be transfered to the client for dynamic=true during the initial request.
So i tried to find a way not to iterate over all SelectItems for the initial dynamic=true - request. My first attempt was to figure out the currently selected SelectItem in an extra step. This PR´s rolls this attempt back. Instead encodeSelectItems now - for the initial dynamic-true-rendering -  stops iterating over SelectItems after we found the one selected SelectItem.

PR #6512, opened 2020-11-15
#6511: AutoComplete: optionally allow to consume REST-endpoint for suggestions

### LINKED ISSUE #6511: AutoComplete: optionally allow to consume REST-endpoint for suggestions
Idea would be to optionally use a simple REST-endpoint to provide AutoComplete-suggestions.
This would avoid having the JSF-lifecycle-overhead including RestoreViewPhase for just providing a few AutoComplete-suggestions.
Once a value is selected "everthing" (eg converters) would happen JSF-"style".

I´ll try to work on this within the next weeks.
Just started with a simple Showcase-Draft-PR: https://github.com/primefaces/primefaces-showcase/pull/221

christophs78:
Showcase-PR: primefaces/primefaces-showcase#221

### DIFF
--- a/src/main/java/org/primefaces/model/rest/AutoCompleteSuggestion.java
+++ b/src/main/java/org/primefaces/model/rest/AutoCompleteSuggestion.java
@@ -23,11 +23,17 @@
  */
 package org.primefaces.model.rest;
 
+import java.io.Serializable;
+import java.util.Objects;
+
 /**
  * AutoComplete-REST-Endpoint (indicated via completeEndpoint-attribute) has to return a list of suggestions.
  * This class may be used to represent one suggestion.
  */
-public class AutoCompleteSuggestion {
+public class AutoCompleteSuggestion implements Serializable {
+
+    private static final long serialVersionUID = 7487414702117430146L;
+
     private String value;
     private String label;
 
@@ -50,5 +56,19 @@ public class AutoCompleteSuggestion {
 
     public void setLabel(String label) {
         this.label = label;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        AutoCompleteSuggestion that = (AutoCompleteSuggestion) o;
+        return value.equals(that.value) &&
+                label.equals(that.label);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(value, label);
     }
 }
--- a/src/main/java/org/primefaces/model/rest/AutoCompleteSuggestion.java
+++ b/src/main/java/org/primefaces/model/rest/AutoCompleteSuggestion.java
@@ -1,0 +1,54 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2020 PrimeTek
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.model.rest;
+
+/**
+ * AutoComplete-REST-Endpoint (indicated via completeEndpoint-attribute) has to return a list of suggestions.
+ * This class may be used to represent one suggestion.
+ */
+public class AutoCompleteSuggestion {
+    private String value;
+    private String label;
+
+    public AutoCompleteSuggestion(String value, String label) {
+        this.value = value;
+        this.label = label;
+    }
+
+    public String getValue() {
+        return value;
+    }
+
+    public void setValue(String value) {
+        this.value = value;
+    }
+
+    public String getLabel() {
+        return label;
+    }
+
+    public void setLabel(String label) {
+        this.label = label;
+    }
+}
--- a/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
+++ b/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
@@ -73,6 +73,7 @@
  * @prop {number} cfg.selectLimit Limits the number of simultaneously selected items. Default is unlimited.
  * @prop {number} cfg.scrollHeight Height of the container with the suggestion items.
  * @prop {boolean} cfg.unique Ensures uniqueness of the selected items.
+ * @prop {string} cfg.restEndpoint REST-Endpoint for fetching autocomplete-suggestions. (instead of completeMethod)
  */
 PrimeFaces.widget.AutoComplete = PrimeFaces.widget.BaseWidget.extend({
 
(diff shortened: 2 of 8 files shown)


christophs78:
image

tandraschko:
yep, thats expected
in think it can be a great addition to components to provide REST integration points in future

tandraschko:
We should think about maiking it a clientbehavior or at least call it "completeEndpoint"

tandraschko:
And we should think about adding the Rest Models to PF

christophs78:
> And we should think about adding the Rest Models to PF

which package-name?

tandraschko:
looks good to me
Could we maybe think about other cases/components and collect some possible endspoints?

if we have many, we should maybe think about something like:

<p:endpoint event="complete" value="/myurl" />

instead of

<p:ajax event="complete" ... />

src/main/java/org/primefaces/model/rest/AutoCompleteSuggestion.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+package org.primefaces.model.rest;
+
+/**
+ * AutoComplete-REST-Endpoint (indicated via completeEndpoint-attribute) has to return a list of suggestions.
+ * This class may be used to represent one suggestion.
+ */
+public class AutoCompleteSuggestion {
+    private String value;
+    private String label;
```
melloware, line 38: Should be serializable and add equals(), hashcode(), and toString() methods.

melloware (changes requested):

melloware:
How about instead of "completeEndpoint" we call it "restCallback" or something more generic. I can also see this being useful on any component that supports `lazy/dynamic` and `f:selectItems`.  And for the custom model object it could be `RestSelectItem` or possibly even just use `SelectItem` as the callback POJOs??  Just some ideas.

tandraschko:
i dont think "restCallback" is good. each feature might need an own endpoint.

melloware:
OK how about just generically "endpoint" then?

tandraschko:
no :D
each feature needs a unique name like "feature1Endpoint" or "feature2Endpoint"
therefore a clientbehavior could be a good way

christophs78:
One thing we need to have an eye on is the amount of rendering-logic we duplicate. (Have it in our classic server-based Java-code and have a second JavaScript-based implementation.)
When we think - as an extreme example - about DataTable it IMO makes limited sense to duplicate this. 
It´s IMO kind of trade-off between what we gain on a per-component/feature-base from adding REST-support and which complexity do we add with this. And how well do REST-endpoints play togehter with this component/feature and it´s JSF-serverside logic.
Do we have some conrete ideas based on our brainstorming where REST-integration would make sense? (on a per componenten/feature-base)
Personally i would invest maybe a few hours per month on this topic. (Including integration tests.) But i personally won´t start a hundreds of hours side-project: ;-)

tandraschko:
i personally think we only should use it for smaller features like this one here.
maybe lazy-load of selectItems of different other components like selectOneMenu would be a good improvement, too. Or search in selectOneMenu.
probably anything related to a "input autocomplete"

i cant think of other usecases now. As you already said, DataTable would be to much effort to make it work for all cases.

tandraschko:
I think we should leave it as it for now (no behavior etc), lets see if we implement other cases in the future.

I think a small but good improvement would be to have a real "root-model" for each possible rest endpoint. WDYT?
Somelike like:
```
class AutoCompleteSuggestionResponse {
     private List<AutoCompleteSuggestion> suggestions;
}
```

and add mellowares suggestions (hashcode, equals, serializable)

christophs78:
> ```
> class AutoCompleteSuggestionResponse {
>      private List<AutoCompleteSuggestion> suggestions;
> }
> ```

I´ll look into this. I have to check how Jackson handles this OOTB.

> and add mellowares suggestions (hashcode, equals, serializable, tostring)

Do you have some prefered implementation for hashcode and equals? (Maybe an existing class which does it in your prefered way i may look into.) Personally i think serializeable is not direct fit to kind of DTO which is returned as JSON by a REST-service. But it should not to any harm. To some degree same for hashcode and equals.

christophs78:
> And for the custom model object it could be `RestSelectItem` or possibly even just use `SelectItem` as the callback POJOs?? Just some ideas.

SelectItems has some features, AutoComplete does not support. AutoComplete is just value + label. No disabled, no description, no explicit escape, no noSelectionOption.

christophs78:
Overall its a REST-API. So `AutoCompleteSuggestion` and maybe `AutoCompleteSuggestionResponse` are just convenience-classes. Theoretically AutoComplete may consume a PHP, NodeJS, C# or whatever REST-service.

melloware:
OK Agreed all around.

christophs78:
`List<AutoCompleteSuggestion>` results in following example JSON:
```javascript
[{"value":"3","label":"Luna-Blue"},{"value":"4","label":"Luna-Amber"},{"value":"5","label":"Luna-Green"},{"value":"6","label":"Luna-Pink"}]
```

Something like `AutoCompleteSuggestionResponse` results in:
```javascript
{"suggestions":[{"value":"3","label":"Luna-Blue"},{"value":"4","label":"Luna-Amber"},{"value":"5","label":"Luna-Green"},{"value":"6","label":"Luna-Pink"}]}
```

IMO this does not make sense. We may compensate this by adding Jackson-Annotations to `AutoCompleteSuggestionResponse` but i don´t think we want to add Jackson (compile-time) dependency. And maybe there are some people who don´t use Jackson but something else. (for whatever reason)

tandraschko:
nope, we shouldnt add jackson.
i just thought its easier to have a single response-root model for each possible endpoint in the future.
I dont care if we have a "suggestions" subnode in json or not TBH :)

christophs78:
I have the following class on my dev-machine:
```java
public class AutoCompleteSuggestionResponse {

    private List<AutoCompleteSuggestion> suggestions = new ArrayList<>();

    public AutoCompleteSuggestionResponse(List<AutoCompleteSuggestion> suggestions) {
        this.suggestions = suggestions;
    }

    public void addSuggestion(AutoCompleteSuggestion suggestion) {
        suggestions.add(suggestion);
    }

    public List<AutoCompleteSuggestion> getSuggestions() {
        return suggestions;
    }

    public void setSuggestions(List<AutoCompleteSuggestion> suggestions) {
        this.suggestions = suggestions;
    }
}
```

But i personally like it that much.

Look into the showcase-PR - IMO following is very straight forward:
```java
@GET
@Path("/autocomplete")
@Produces({ MediaType.APPLICATION_JSON })
public List<AutoCompleteSuggestion> autocomplete(@QueryParam("query") String query) {
	String queryLowerCase = query.toLowerCase();
	List<Theme> allThemes = service.getThemes();
	return allThemes.stream()
			.filter(t -> t.getName().toLowerCase().contains(queryLowerCase))
			.map(t -> new AutoCompleteSuggestion(Integer.toString(t.getId()), t.getDisplayName()))
			.collect(Collectors.toList());
}
```

IMO it makes good use of streams. Yes we can put this into AutoCompleteSuggestionResponse. But i don´t thinks this improves the REST-service-code.

melloware:
I say lets leave it out for now then.

tandraschko:
in the docus i would do:
`completeEndpoint="#{request.contextPath}/rest/theme/autocomplete"`
instead of
`completeEndpoint="../../rest/theme/autocomplete"`

tandraschko:
after this change and the equals/hashcode thing, i would merge it

christophs78:
We integrated this today in our real-world-project. And it turn´s out after all there are reasons to introduce `AutoCompleteSuggestionResponse`. The REST-endpoint may limit the amount of results it send´s to the client. So we need a flag to indicate there may be more results. This flag has to be located in `AutoCompleteSuggestionResponse`.

And i have to look that emptyMessage work´s with REST-endpoint to.
Update: emptyMessage work with completeEndpoint OOTB.

I'll get in touch after that's done.

src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
```diff
@@ -909,14 +910,18 @@ PrimeFaces.widget.AutoComplete = PrimeFaces.widget.BaseWidget.extend({
...
+                        suggestions.suggestions.forEach(function(suggestion) {
                             var labelEncoded = $("<div>").text(suggestion.label).html();
                             var itemValue = labelEncoded;
                             if (!!suggestion.value) {
                                 itemValue = $("<div>").text(suggestion.value).html();
                             }
                             html += '<li class="ui-autocomplete-item ui-autocomplete-list-item ui-corner-all" data-item-value="' + itemValue + '" data-item-label="' + labelEncoded + '" role="option">' + labelEncoded + '</li>';
                         });
+                        if (suggestions.moreAvailable == true && $this.cfg.moreText) {
+                            var moreTextEncoded = $("<div>").text($this.cfg.moreText).html();;
```
melloware: Double ;; on this line

christophs78: Thanks for pointing me on this!

christophs78:
1. Docu is updated.
2. `AutoCompleteSuggestionResponse` is introduced and supports `moreText` via `moreAvailable`-property.

equals/hashcode thing IMO makes as i mentioned before limited sense for DTO´s we send to the client. If we realy want to add equals/hashcode for this please point me to some kind of blueprint-class.

tandraschko:
i would just auto-generate it for the suggestion class, also implement serializable for both models
so you auto-avoid duplicate items in a list e.g.

christophs78:
> i would just auto-generate it for the suggestion class, also implement serializable for both models
> so you auto-avoid duplicate items in a list e.g.

done :-)

tandraschko:
looks good for me now
maybe we should add a hint in the docs "the json looks like this ... but we have a model class for it" 
or something similar?

tandraschko:
Thanks for the great PR!

PR #6516, opened 2020-11-16
Fix #4919: SelectOneMenu alwaysDisplayLabel attribute

### LINKED ISSUE #4919: SelectOneMenu: label attribute not working
## 1) Environment
- PrimeFaces version: 7.0.0
- Application server + version: JBoss EAP 7.1.1
- Affected browsers: IE11, Chrome 75

## 2) Expected behavior

The label of the selectonemenu is "Hello World"

## 3) Actual behavior

The label is the itemLabel of the selected entry.

## 4) Steps to reproduce

Just run the xhtml.

## 5) Sample XHTML

```
    <h:head>
        <title></title>
        
    </h:head>
    <h:body>
		<p:selectOneMenu label="Hello World">
			<f:selectItem itemValue="YELLOW" itemLabel="Yellow" />
			<f:selectItem itemValue="BLUE" itemLabel="Blue" />
			<f:selectItem itemValue="RED" itemLabel="Red" />
		</p:selectOneMenu>
    </h:body>
</html>
```

## 6) Sample bean

None

melloware:
Reimplements this change with the added `alwaysDisplayLabel` attribute and tested both uses cases from the ticket.

### DIFF
--- a/docs/9_0/components/selectonemenu.md
+++ b/docs/9_0/components/selectonemenu.md
@@ -35,7 +35,9 @@ widgetVar | null | String | Name of the client side widget.
 effect | blind | String | Name of the toggle animation.
 effectSpeed | normal | String | Duration of toggle animation, valid values are "slow", "normal" and "fast".
 disabled | false | Boolean | Disables the component.
-label | null | String | User presentable name.
+alwaysDisplayLabel | false | Boolean | Always display the `label` value instead of the selected item label.
+label | null | String | User presentable name used in conjuction with `alwaysDisplayLabel` to display instead of selected item.
+labelTemplate | null | String | Displays label of the element in a custom template. Valid placeholder is {0}.
 onchange | null | String | Client side callback to execute on value change.
 onkeyup | null | String | Client side callback to execute on keyup.
 onkeydown | null | String | Client side callback to execute on keydown.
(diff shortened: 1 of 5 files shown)


tandraschko:
TBH i dont know if this makes sense as you never see the selected item anymore, if you dont open the overlay?

melloware:
I think all we are doing is giving consistent option like SelectCheckBoxMenu has in its "Basic" example: http://localhost:8080/showcase/ui/input/checkboxMenu.xhtml

Where if a user like @Eehol always wants to display a certain value for the SelectOne no matter what value is selected.

if we aren't then we should remove the `label` attribute and close this ticket as it looks like `label` is never actually used.

Eehol:
Hello, just for clarifying ; by using that attribute we could override the default label with a custom method that will handle a label that would be too long.
For example, we have a business case where a SelectItem label is 200 chars long. I would like to be able to override the selected label to use a method that will give a text that displays only the 50 first characters followed by "..."

melloware:
Yep so you would do this..
```xml
 <p:selectOneMenu id="labeled" value="#{view.value}" label="#{view.label}" alwaysDisplayLabel="true">
```

Where `#{view.label}` returned `StringUtils.abbreviate(value, 50)`

Eehol:
Yup and that's why that feature makes sense to me (to respond to @tandraschko comment) 

> TBH i dont know if this makes sense as you never see the selected item anymore, if you dont open the overlay?

PR #6522, opened 2020-11-17
Fix #5579: Spinner do not allow non numeric characters

### LINKED ISSUE #5579: Spinner: should not allow to enter non-numeric values

melloware:
Failing Integration Tests proving the bugs.
https://github.com/primefaces-extensions/primefaces-integration-tests/commit/2028db3f4cf8670825892ce6eb0698ca72885860

```
Error:  Failures: 
Error:    Spinner001Test.testDecimalSeparator:97 expected: <34> but was: <3>
Error:    Spinner001Test.testInvalidCharacter:80 expected: <0> but was: <abc0>
Error:    Spinner001Test.testThousandSeparator:114 expected: <1,456> but was: <14,560>
Error:    Spinner002Test.testInvalidCharacter:83 expected: <> but was: <abc>
[INFO] 
Error:  Tests run: 114, Failures: 4, Errors: 0, Skipped: 1
```

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/spinner/spinner.js
+++ b/src/main/resources/META-INF/resources/primefaces/spinner/spinner.js
@@ -17,7 +17,7 @@
  * configuration is usually meant to be read-only and should not be modified.
  * @extends {PrimeFaces.widget.BaseWidgetCfg} cfg
  * 
- * @prop {number} cfg.decimalPlaces Number of decimal places.
+ * @prop {string} cfg.decimalPlaces Number of decimal places.
  * @prop {string} cfg.decimalSeparator The character separating the integral and fractional parts of the number.
  * @prop {number} cfg.max Minimum allowed value for this spinner.
  * @prop {number} cfg.maxlength Maximum number of characters that may be entered in this field.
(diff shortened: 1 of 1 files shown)


melloware:
Passing Integration Tests after merging this fix.

https://github.com/primefaces-extensions/primefaces-integration-tests/actions/runs/368763582

tuerker:
is this fix already merged into the master? i used the latest PF snapshot. there it was still not working. you can still enter non-numeric characters for spinner on a numeric field. 

Screenshot from 2020-11-19 12-11-31

or is there still something to be fixed?

melloware:
Yep its merged in and you can see I have a Integration Test proving the fix works.
```java
 @Test
    @Order(3)
    @DisplayName("Spinner: GitHub #5579 Test integer invalid characters should be ignored")
    public void testInvalidCharacter(Page page) {
        // Arrange
        Spinner spinner = page.spinner;
        Assertions.assertEquals("0", spinner.getValue());

        // Act
        ComponentUtils.sendKeys(spinner.getInput(), "abc");
        page.button.click();

        // Assert
        Assertions.assertEquals("0", spinner.getValue());
        assertConfiguration(spinner.getWidgetConfiguration());
    }
```

tuerker:
thanks. good to know. then, there must be another bug (as you can see from screenshot above) which appears under certain conditions. hope to find that and provide you a reproducer.

tuerker:
could be good if the delete, backspace, left and right-arrow buttons would still work. useful in case of longer numbers. i know than one could discuss whether the spinner is the right widget in such a case.

tandraschko:
why isnt spinner type=number?

melloware:
> why isnt spinner type=number?

Good question it looks like it has always been this way.   Want me to submit another ticket changing it to `type=number`???

tandraschko:
you have to check all features if thats possible
some ppl would like to see it for AutoNumeric buts thats not possible because of prefix/suffix strings

melloware:
I can try it. I have quite a few Integration tests testing the Spinner now so we would see if any fail.  That is the nice thing of adding these Integration Tests!

melloware:
confirmed its type=text because of `prefix` and `suffix`.  The `type=number` breaks that feature.

PR #6529, opened 2020-11-19
Timeline: added nested group functionality

### LINKED ISSUE #6530: Timeline: Nested groups feature
https://visjs.github.io/vis-timeline/examples/timeline/groups/nestedGroups.html

buunsal:
https://visjs.github.io/vis-timeline/examples/timeline/groups/nestedThreeLevels.html

Adds new feature to primefaces timeline component for nested groups such as in the visjs link above.

I will also make a pull request for the showcase with an example like in the picture.
image

### DIFF
--- a/src/main/java/org/primefaces/component/timeline/TimelineRenderer.java
+++ b/src/main/java/org/primefaces/component/timeline/TimelineRenderer.java
@@ -30,10 +30,7 @@ import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
 
 import javax.faces.component.UIComponent;
 import javax.faces.context.FacesContext;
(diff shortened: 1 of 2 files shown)


melloware:
Please create an Issue ticket so we can link this PR to an issue.

src/main/java/org/primefaces/model/timeline/TimelineGroup.java
```diff
@@ -44,6 +45,17 @@
...
+    /**
+     * root level of the nested groups.
+     */
+    private Integer treeLevel;
+
+    /**
+     *
+     * it is a set of ids of nested groups that a group or nested group contains.
+     */
+    private HashSet<String> nestedGroups;
```
melloware: This should just be Set and not HashSet.  That way someone could use a TreeSet or any other kind of Set they want.

src/main/java/org/primefaces/component/timeline/TimelineRenderer.java
```diff
@@ -316,6 +316,18 @@ else if (data != null) {
             fsw.write(", content:\"" + groupsContent.get(group.getId()) + "\"");
         }
 
+        if (group.getTreeLevel() != null) {
+            fsw.write(", treeLevel: \"" + group.getTreeLevel() + "\"");
+
+            if (group.getNestedGroups() != null) {
```
melloware:
```java
Set<String> nestedGroups = group.getNestedGroup();
if (nestedGroups  != null && !nestedGroups.isEmpty()) {
```

melloware (changes requested):

melloware (approved):

PR #6533, opened 2020-11-19
Fix #5579: Spinner allow navigation keys

### LINKED ISSUE #5579: Spinner: should not allow to enter non-numeric values

melloware:
@tuerker added navigation keys

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/spinner/spinner.js
+++ b/src/main/resources/META-INF/resources/primefaces/spinner/spinner.js
@@ -142,6 +142,13 @@ PrimeFaces.widget.Spinner = PrimeFaces.widget.BaseWidget.extend({
                     $this.format();
                 break;
 
+                case keyCode.BACKSPACE:
+                case keyCode.DELETE:
+                case keyCode.LEFT:
+                case keyCode.RIGHT:
+                    return;
+                break;
+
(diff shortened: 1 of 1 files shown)


tuerker:
strange. although i have the latest PF snapshot (see your changes in the screenshot), the four buttons do not work. attached is my reproducer.

Screenshot from 2020-11-19 17-42-27

[p-spinner.zip](https://github.com/primefaces/primefaces/files/5568346/p-spinner.zip)

melloware:
Let me try it

melloware:
Works fine for me in latest Firefox and Chrome.  All 4 keys work fine arrows and backspace and delete

melloware:
I even put a breakpoint in the JS and I can see it hitting the `return` statement when I press one of those 4 keys.

tuerker:
I tried it on my machine at home. There, the same reproducer is working. Sorry, for the false alarm.

And as usual, many thanks for your great job.

melloware:
No problem thanks for reporting back!

PR #6536, opened 2020-11-19
Fix #6535 - MessageFactory: use varargs

### LINKED ISSUE #6535: MessageFactory: use varargs
**Description**
Use varargs instead of Object[]

Rapster:


tandraschko:
the change in general looks good, however the code in MessageFactory is by design that i didnt use a shared method for both String and FacesMessage return values, as this now creates a necessary array.

melloware (approved):

Rapster:
I can always revert it even though I'm not sure what differences it makes...

tandraschko:
less memory, less GC, thats all

Rapster:
Mmmh... If we were working on embedded systems with very limited memory I would understand, but in such scenario not sure... Anyway, I'll revert it

tandraschko:
yeah, i can understood you
but we have to microoptimize the core-functions of a framework
just think about hundrets of concurrent requests
if we only save like 10 objects in such a function each request, it isn't that few

same for for vs foreach on component childlists

tandraschko:
thanks!

PR #6537, opened 2020-11-20
Fix #6501 - ColumnToggler: added resize columns function to columntoggler

### LINKED ISSUE #6501: ColumnToggler: Toggling on columns after resizing others breaks table width
**Describe the defect**
The issue happens when you use both resizableColumns="true" and a p:columnToggler in a dataTable.
As long as you only resize while all columns are visible, you can toggle columns and they will always be fitted to the table width. But when you toggle some columns off, resize some of the remaining ones, and show again the hidden ones, the columns are now wider than the table and a scrollbar appears.

**Environment:**
 - PF Version: _PrimeFaces 8.0-SNAPSHOT_
 - JSF + version: _Mojarra 2.3.13_
 - Affected browsers: _tested on Chrome and Firefox_

**To Reproduce**
I used the DataTable - Resize example from the showcase and added a columnToggler like shown in the example xhtml.
1. Use the column toggler to hide one or more columns.
2. Resize the remaining columns.
3. Use the column toggler to show all columns again.
4. See columns getting too wide and scrollbar appear.

**Expected behavior**
The column widths should be fit to the table width as the default resizeMode of the dataTable suggests.

**Example XHTML**
```html
<f:facet name="header">
    Basic Resize with toggler
    <p:commandButton id="toggler" type="button" value="Columns" icon="pi pi-align-justify"/>
    <p:columnToggler datasource="products1" trigger="toggler"/>
</f:facet>
```

buunsal:


src/main/resources/META-INF/resources/primefaces/columntoggler/columntoggler.js
```diff
@@ -397,6 +398,38 @@ PrimeFaces.widget.ColumnToggler = PrimeFaces.widget.DeferredWidget.extend({
         this.changeTogglerState(column, false);
         this.fireToggleEvent(false, (index - 1));
         this.updateColspan();
+        this.resizeColumns(false);
+    },
+
+    /**
+     * Resize columns after toggle
+     * @param {boolean} isChecked `true` when the number of column will increase
```
melloware: number of column(s) will increase

src/main/resources/META-INF/resources/primefaces/columntoggler/columntoggler.js
```diff
@@ -397,6 +398,38 @@ PrimeFaces.widget.ColumnToggler = PrimeFaces.widget.DeferredWidget.extend({
...
+
+        var columns = this.table.find('> .ui-datatable-tablewrapper > table > thead > tr > th');
+
+        if(columns.length) {
+            var visibleColumnLength = columns.filter(':visible').length;
+
+            if (isChecked && visibleColumnLength > 1) {
+                columns.each(function() {
+                    var col = $(this),
+                        colStyle = col[0].style,
```
melloware, line 419: I haven't checked by can `colStyle` be NULL or undefined if a column has no style?  Then the next line will throw an error when its checks its width.

melloware (changes requested):

src/main/resources/META-INF/resources/primefaces/columntoggler/columntoggler.js
buunsal: Columns always have style from ui-datatable or themes. In addition, the same lines are used exactly in 5 different places in the datatable.js file.

melloware (approved):

mertsincan:
With this PR, the problem still exists. When we toggle a few columns, the table shows a horizontal scroll. Unfortunately, reverted.

PR #6551, opened 2020-11-24
AutoUpdate doc pub/sub

melloware:


tandraschko:
thanks, looks good in general :)
I just wonder if we should name it event/observer instead publish/subscribe, which is more aligned with CDI

melloware:
Yep good call. I will fix

melloware:
OK updated.

PR #6560, opened 2020-11-25
Fix #5438: Improved error messages

### LINKED ISSUE #5438: DataTable: cleanup sorting

melloware:
I think good error messages will be key for users upgrading their PF and getting errors on their datatable.  I made these two issues more clear as to what is wrong.


Rapster:
Thank you for reporting @melloware but I slightly change your fix

PR #6562, opened 2020-11-26
New Feature #5309: Multiple entries in pickList

e-reznik:
This is my second attempt of the implementation of feature request #5309, as can be seen in pull request #6244

So far, I've added the new attribute `allowDuplicates` and changed the method declaration.
The next step would be to identify the remaining required method(s).

As this is my first attempt to contribute to such a huge project, every guidance is appreciated :-)


melloware (commented):
So far so good from my point of view.

tandraschko:
I will close this draft now. 2 weeks without changes and i would like to clenup our issue tracker abit.
Mabye you can try to get something working first and then comeup with a draft PR. We are also on gitter if you need some guidance.

Issue #1921, opened 2016-11-24
DataTable: lazy loading pagination error


sergi-mm:
I've detected an issue related to pagination when using lazy loading in datatable. The problem is the following: starting by the example of showcase: http://www.primefaces.org/showcase/ui/data/datatable/lazy.xhtml, if we add a button to delete the records one by one, if you are for instance in page 2, when all record of this page have been deleted, the pagination "brokes" and displays page 1 with empty data.

Assume that the button "Delete" calls a method called "deleteCar(Car car)" which deletes the car object from the list (or directly from database), and then update the datatable, firing again the load() method from LazyDatamodel.

```html
<p:dataTable var="car" value="#{dtLazyView.lazyModel}" paginator="true" rows="10"
                 paginatorTemplate="{RowsPerPageDropdown} {FirstPageLink} {PreviousPageLink} {CurrentPageReport} {NextPageLink} {LastPageLink}"
                 rowsPerPageTemplate="5,10,15" selectionMode="single" selection="#{dtLazyView.selectedCar}" id="carTable" lazy="true">
        <p:ajax event="rowSelect" listener="#{dtLazyView.onRowSelect}" update=":form:carDetail" oncomplete="PF('carDialog').show()" />
        <p:column headerText="Id" sortBy="#{car.id}" filterBy="#{car.id}">
            <h:outputText value="#{car.id}" />
        </p:column>
        <p:column headerText="Year" sortBy="#{car.year}" filterBy="#{car.year}">
            <h:outputText value="#{car.year}" />
        </p:column>
        <p:column headerText="Brand" sortBy="#{car.brand}" filterBy="#{car.brand}">
            <h:outputText value="#{car.brand}" />
        </p:column>
        <p:column headerText="Color" sortBy="#{car.color}" filterBy="#{car.color}">
            <h:outputText value="#{car.color}" />
        </p:column>
<p:column headerText="Color" sortBy="#{car.color}" filterBy="#{car.color}">
            <p:commandButton value="Delete" action="#{dtLazyView.deleteCar(car)}" update="carTable">
</p:commandButton>
        </p:column>
    </p:dataTable>
```

In the example, an IndexOutOfBoundsException will be also thrown, when deleting the last record of page 2, variable "first" will pass a value of 10 (assuming rows per page is 10) and a subList from 10 to 10-1...

FastainCS09:
[...]

Issue #1184, opened 2016-02-25
Ability to sort inputs in datatable


cagataycivici:
Add support to sort inputs in datatable and if validation fails sorting should not happen.

Issue #405, opened 2015-06-05
commandLink returns incorrect item on dataTable with liveScroll="true" lazy="true" 


darthstark1138:
When using a commandLink (or commandButton) on a liveScroll, lazy datatable, the first items (before the first lazy load) report the correct item. After loading the second batch of items, items in the first batch correspond to items in the second batch.

It seems that JSF reuses the indexes when loading more items using ajax.

This has been reproduced on 4.0 and 5.2.

The problem have been reported previously as [Issue 6879: DataTable with lazy loading, livescroll and commandbutton not working](https://code.google.com/p/primefaces/issues/detail?id=6879) on Google Code. It has been marked as WontMoveToGithub.

kukel:
> _"It seems that JSF reuses the indexes when loading more items using ajax."_

Did you check this in the generated html? Or with other debugging?

And a full example that demonstrates the problem would be nice. (The original ticket requested more details)

cagataycivici:
A simple test case would help.

hink084:
This is still a problem in PrimeFaces v6.0 using Glassfish 4.1 / Mojarra 2.2.13.

The following example is crude but to the point.
A similar problem exists with the RowEditor (See modified example below).

**Recreation Steps:**
1. Cut and paste the following code.
2. Click the first few links in the DataTable and notice the output in the console.
3. Scroll down to initiate the LiveScroll feature.
4. Clicking on Link 0-14 Incorrect
5. Clicking on Link > 14 Broke

**_TestCase.java_**

```
import java.io.Serializable;
import java.util.ArrayList;
import java.util.List;
import java.util.Map;
import javax.faces.event.ActionEvent;
import javax.faces.view.ViewScoped;
import javax.inject.Named;
import org.primefaces.model.LazyDataModel;
import org.primefaces.model.SortOrder;

@Named(value = "testCase")
@ViewScoped
public class TestCase implements Serializable{
    private DataModel dataModel;

    public TestCase(){
        this.dataModel = new DataModel();
    }

    public DataModel getDataModel(){
        return dataModel;
    }

    public void setDataModel(DataModel dataModel){
        this.dataModel = dataModel;
    }

    public void MyButton_OnClick(ActionEvent event){
        MyObject myObject = (MyObject) event.getComponent().getAttributes().get("MyData");
        System.out.println("MyObject Id: " + myObject.getId() + " MyObject Name: " + myObject.getName());
    }

    public final static class MyObject implements Serializable{
        private int id;
[...]

PR #6570, opened 2020-11-27
Fix #1390 - DataTable: Filter getFilteredValue out of sync

### LINKED ISSUE #1390: DataTable: Filter getFilteredValue out of sync
I'm in the the process of updating an old 3.1 Primefaces app.  In this app we make use of the datatable filter ajax event and call a listener to update some backingbean properties.  I've noticed a change of behaviour in this listener.  Previously when this listener ran, either the filieredValue defined list would be updated or the filteredValue from the FilterEvent>getSource would be current.  It would appear now that these are still the old values.  These are effectively 1 step behind the displayed values. 

So we had this nested in a standard filterable datatable

`<p:ajax event="filter" listener="#{MyBackingBean.onFilter}" process="@this" update="@this"/>`

This calls the following method:

```
public void onFilter(FilterEvent filterEvent) {
        List filteredList = ((DataTable)filterEvent.getSource()).getFilteredValue();`
    }
```

On first filter this list will be null.  On second filter it will show the results of the 1st filter and so on.  I have tried this on 5.2._, 5.3._ and 6.0RC3. 

I've also noticed that elements included in the filter update which are placed before the datatable have their getters called prior to the FilterFeature Encode call.  Elements after the datatable have the getters called after the encode.  This doesn't seem correct to me.  I would assume all getters be called after the filter has completed.

Rapster:
The reason why is out of sync is pretty simple. Events are processed during APPLY_REQUEST_VALUES phases and filtered values are decoded in the next phase PROCESS_VALIDATIONS (`see FilterFeature#encode`): listener is called before filtered values get decoded.

First thing to fix: filtered values have to be decoded in APPLY_REQUEST_VALUES phase. So far, `FilterFeature#shouldDecode` returns false while it actually should return true. And filtering should be run in that same phase so `FilterEvent#filteredData` can be set

Second thing, when custom filters are used, values from filters component are retrieved from `ValueHolder#getLocalValue` but local value is set during PROCESS_VALIDATIONS phase, which is already too late... Instead I'm using `UIInput#getSubmittedValue` set during `APPLY_REQUEST_VALUES` phase.

Being said, that would explain why `FilterFeature#decode` was called from `DataTable#processValidators`, to filter after local value has been set. But at this point, it's all about fixing consequences and not the origin of the problem. I removed few things like `DataTable#processEvent` which I think tries to fix a consequence and again , not the root cause.

I realized `FilterFeature#decode` and `SortFeature#decode` does similar stuff, just like `DataTable#lazyLoadData` and `DataTable#loadLazyScrollData`, I refactor these methods to avoid duplication


Rapster:
The reason why is out of sync is pretty simple. Events are processed during APPLY_REQUEST_VALUES phases and filtered values are decoded in the next phase PROCESS_VALIDATIONS (`see FilterFeature#encode`): listener is called before filtered values get decoded.

First thing to fix: filtered values have to be decoded in APPLY_REQUEST_VALUES phase. So far, `FilterFeature#shouldDecode` returns false while it actually should return true. And filtering should be run in that same phase so `FilterEvent#filteredData` can be set

Second thing, when custom filters are used, values from filters component are retrieved from `ValueHolder#getLocalValue` but local value is set during PROCESS_VALIDATIONS phase, which is already too late... Instead I'm using `UIInput#getSubmittedValue` set during `APPLY_REQUEST_VALUES` phase.

Being said, that would explain why `FilterFeature#decode` was called from `DataTable#processValidators`, to filter after local value has been set. But at this point, it's all about fixing consequences and not the origin of the problem. I removed few things like `DataTable#processEvent` which I think tries to fix a consequence and again , not the root cause.

I realized `FilterFeature#decode` and `SortFeature#decode` does similar stuff, just like `DataTable#lazyLoadData` and `DataTable#loadLazyScrollData`, I refactor these methods to avoid duplication

tandraschko:
i like this PR a lot as it moves the code where it belongs to.
Just one thing:
`filterValue = ((UIInput) filterFacet).getSubmittedValue()`

this should be the RAW value and localValue should be the converted one. Not sure if it breakes something.
Also can filter trigger validation errors? What could happen in worst case? filtering with invalid value?

please test test test

melloware:
Just thinking out loud here after reviewing this fix you don't think this will fix or is a similar issue to these two issues do you?
https://github.com/primefaces/primefaces/issues/1921
https://github.com/primefaces/primefaces/issues/405

They are both Lazy model issues with it calculating something incorrectly upon scrolling.

tandraschko:
@Rapster i would like to finish this in 1-2 days as i need to refactor treetable to have MVS support

i like this change but i'm not sure if submittedValue is correct. If we cant use getValue/getLocalValue, then it might be to early to access it. If a validation error occurs, we would overwrite our old/correct filterValue with the invalid new submittedValue.

Rapster:
@melloware i'm having a look to these issues, so far can't reproduce #1921 

@tandraschko I don't think we have other choices than use submittedValue, and so far I managed to make it work (without dirty trick, except I had to introduce `ComponentUtils#getConvertedValue2` 😅 ). Regarding validation errors, what sorting does so far is that, it prevents sorting, see https://github.com/primefaces/primefaces/issues/1184

tandraschko:
why not?
do we need the filterValue in that phase?
we could make filterValue evaluation lazy to move it to later phases; or do it in processValidators

Rapster:
> do we need the filterValue in that phase?

Yes you do, events (like filter) are queued in APPLY_REQUEST_VALUES phase, SortEvent is built expecting the new filtered value

tandraschko:
why are they queued so early?
I dont think this events should be fired before validation is complete?

Rapster:
True... A simple use case that convinced me is that, let's say i'm using datepicker range, and for some reason range is badly set I shouldn't expect filter event to be triggered...

Another thing, I realized `FilterEvent#getData` is deprecated since 2014 it doesn't say why but I must say I do agree with this, if you need to know the new filtered value, simply check your `SelectionView#filteredValue` bean. SortEvent/FilterEvent should carry sort/filter meta and nothing else.

It doesn't change the problem, filters still need to be decoed at the right phase.

> we could make filterValue evaluation lazy to move it to later phases; 

Any existing example in code base so far?

tandraschko:
About the phase for the event... Sometimes we set the right phase e.g. in the WizardRenderer:
event.setPhaseId(PhaseId.INVOKE_APPLICATION); 

I think we should think about events, which makes only sense after validation and set the phaseId correctly. But that doesnt fix this problem that we access the localValue/Value to early. 

> simply check your SelectionView#filteredValue bean

yep. i think the event + filtereredValue will always be in the same controller, so there is no need to pass it in the event itself. we can remove it IMO.

> Any existing example in code base so far?

dont think so.
I think we have to look at the whole thing first.
UIComponent / Renderer #decode will be called in APPLY_REQUEST_VALUES. In case of UIInput, it will set the submittedValue, which will be coverted and validated in later phases.

So in our case, the FilterFeature#decodeFilterValue is correct for the most cases as it gets values from the request. Just the case where we access UIInput is wrong, we should do this in a later phase.
Probably we can do it like in the #processValidators method before.

In general the question is: why dont we move the sortBy and filterBy decoding completely to after #processValidator, like it was before with FilterFeature?

Rapster:
I pushed a first draft of what we recently discussed, i'll see if I can figure something out more clean...

tandraschko:
for me its quite ok, its similar to other components 
also i dont like the customEvents things, it's quite a pattern in PF

tandraschko:
In other components #validate is used instead of #processValidators, i think we should do the same here

Rapster:
There is no `DataTable#validate`, this method comes with `UIInput` class. Right now, I guess it's okay, not great but okay... 2 questions:
- Just like I did for FilterFeature & SortFeature, should I do the same for other feature?
- SortEvent expects a bunch of parameters while we could simply pass the map of SortMeta, I think I should change that

melloware:
@Rapster you have some merge conflicts to resolve.

Rapster:
New PR #6604

PR #6571, opened 2020-11-28

### LINKED ISSUE #6169: DataTable: Lazy getRowData Repeat the query twice

Rapster:


melloware:
Can you explain your solution?  Is this just moving all the code around so the method doesn't get called from the datatable and the feature?

Rapster:
Fair point! I'm going to start with this: there is no way to prevent calling `DataTable#getRowData`. My first attempt was to use a cache to prevent this and have better perf, but as the model is reseted all the time (by design I'm guessing) my PR didn't bring a lot to the table.

Now, my idea is to avoid scenarios like expensive call to differents endpoints (e.g database, REST etc.), turns out the result sounds like a lot like SelectItemsConverter from Omnifaces does, except it's not based on toString() representation but either `DataTable#rowKey` or `System#identityHashCode`.

To sum up:
- users no longer need to implement `LazyDataModel#getRowKey` and `LazyDataModel#getRowData`. As of today, either you need to set `DataTable#rowKey` or `LazyDataModel#getRowKey` if lazy mode, from now on one entry point: `DataTable#rowKey` and yet, this is optional.

- Whenever selection has to be decoded, `LazyDataModel#getRowData` will under the hood, loop over Datatable rows till it finds a matching result by using rowKey strategy that has been set with `DataTable#rowKey`. You might wonder what is that cache in DefaultSelectableModel, let's say i select all elements in reverse order (from bottom to top) time complexity will remain O(N), today this scenario is O(N²) _(because it'll keep iterating over elements even though some have been visited)_

Back to the issue, no query will be executed at all

melloware:
Very nice.

src/main/java/org/primefaces/component/datatable/DataTableBase.java
```diff
@@ -229,11 +229,11 @@ public void setScrollRows(int scrollRows) {
         getStateHelper().put(PropertyKeys.scrollRows, scrollRows);
     }
 
-    public Object getRowKey() {
-        return getStateHelper().eval(PropertyKeys.rowKey, null);
+    public String getRowKey() {
```
melloware, line 232: Does this mean RowKey can't be a Long anymore or a composite key?

Rapster: It can, it will be coerce into String. What do you mean by composite key?

melloware: Composite key meaning if it was a Hibernate Entity it can have a PK class that is more than one field to make up its uniquness.
```java
public class MyEntity {

     @Id
     private MyPk myPrimaryKey;
```

Rapster: Then, it will work same as today your PK will be coerced into a String (and hopefully you overrided `PK#toString()`

melloware: OK!

tandraschko:
JFYI: im not sure if i will merge further refactorings now for 9.0
9.0-RC1 is actually planned for december and we should target stabilization and bugfixes now

Rapster:
Sounds good to me :wink:

melloware:
Merge conflicts here too

Rapster:
Let's merge #6604 first

Rapster:
New PR #6607

PR #6584, opened 2020-12-01
Fix #6583/#4541: Overlays scrolling in dialog and window

### LINKED ISSUE #4541: Autocomplete: wrong fixed position autocomplete-panel on page with scroll and p:dialog
## 1) Environment
- PrimeFaces version: 7.0.RC2
- Application server + version: Tomcat 7 + 9
- Affected browsers: Chrome, Firefox, Edge...

## 2) Expected behavior
Autocomplete panel on dialog will not leave when scroll on mouse.

## 3) Actual behavior
Autocomplete panel on dialog leaves when scroll on mouse.

## 4) Steps to reproduce
On the vertical scrollbar page, I see p:dialog with autocomplete dropdown.
Unpack the autocomplete panel by clicking on the dropdown.
Scrolling the mouse on the main page and the autocomplete panel leave position.

## 5) Sample XHTML

```

<h:head>
</h:head>

<h:body style="height: 1500px;">
	<p:commandButton value="Modal" type="button" onclick="PF('dlg2').show();" />

	<p:dialog header="Autocomplete" widgetVar="dlg2" modal="true" position="400,200" width="300" height="150">
		<h:form>
			<p:autoComplete id="dd" dropdown="true" scrollHeight="250"
				value="#{testController.autocompleteValue}" 
				completeMethod="#{testController.completeText}" >				
			</p:autoComplete>
		</h:form>
	</p:dialog>
</h:body>
</html>
```

image

## 6) Sample bean
```
public List<String> completeText(String query) {
	List<String> results = new ArrayList<String>();
	for (int i = 0; i < 100; i++) {
		results.add(String.format(query + " - %#04x", i));
	}
	return results;
}
```

### LINKED ISSUE #6583: Improvement in behavior of components that have overlay panels in scrollable containers
**Describe the defect**
There have been multiple defects reported about overlay panel of components not sticking with the component when scrolling in a dialog or in a window.

**Reproducer**
[overlay-scroll.zip](https://github.com/primefaces/primefaces/files/5625120/overlay-scroll.zip)

**Components:**
- [x] AutoComplete https://github.com/primefaces/primefaces/issues/4541
- [x] SelectOneMenu https://github.com/primefaces/primefaces/issues/4609
- [x] SelectCheckboxMenu
- [x] ColorPicker
- [x] DatePicker
- [x] Calendar

**Environment:**
 - PF Version: _8.0_
 - JSF + version: _ALL_
 - Affected browsers: _ALL_

**To Reproduce**
There are two types of problems.  

1. When a component is opened in a dialog that can be scrolled scrolling in the dialog separates from the component like:
image

2. Scrolling in the Document Window can separate the component like:
image

melloware:
The reproducer in the ticket tests all these scenarios in pages test.xhtml, and test3.xhtml.   I also retested on the showcase to make sure nothing was broken and all is working perfectly.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
+++ b/src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
@@ -63,6 +63,7 @@ PrimeFaces.widget.DatePicker = PrimeFaces.widget.BaseWidget.extend({
         this.configureLocale();
 
         //events
+        this.bindPanelCreationListener();
         this.bindDateSelectListener();
         this.bindClearButtonListener();
         this.bindViewChangeListener();
@@ -112,7 +113,6 @@ PrimeFaces.widget.DatePicker = PrimeFaces.widget.BaseWidget.extend({
         //Initialize datepicker
         this.cfg.panelStyleClass = (this.cfg.panelStyleClass || '') + ' p-datepicker-panel';
         this.cfg.viewDate = this.viewDateOption;
-        this.cfg.appendTo = this.cfg.appendTo ? PrimeFaces.expressions.SearchExpressionFacade.resolveComponentsAsSelector(this.cfg.appendTo) : null;
         this.cfg.rangeSeparator = this.cfg.rangeSeparator||'-';
         this.cfg.timeSeparator = this.cfg.timeSeparator||':';
 
@@ -137,7 +137,7 @@ PrimeFaces.widget.DatePicker = PrimeFaces.widget.BaseWidget.extend({
             PrimeFaces.skinButton(triggerButton);
         }
 
-        //mark target and descandants of target as a trigger for a primefaces overlay
+        //mark target and descendants of target as a trigger for a PrimeFaces overlay
         if(!this.cfg.inline) {
             this.jq.data('primefaces-overlay-target', this.id).find('*').data('primefaces-overlay-target', this.id);
         }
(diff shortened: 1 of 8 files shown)


src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
```diff
@@ -215,6 +215,23 @@ PrimeFaces.widget.DatePicker = PrimeFaces.widget.BaseWidget.extend({
         }
     },
 
+    /**
+     * Callback for after the overlay panel is created.
+     * @private
+     */
+    bindPanelCreationListener: function() {
+        var _self = this;
```
melloware, line 223: I have a future PR where I will convert all _self to $this in the 1-datepicker.js as it doesn't do things like our other components and it should be consistent.

melloware:
Explaining this PR in detail:

- Standardize all of these components to have an `alignPanel` method.  Most did but some did not
- SelectOne had solved some of these problems so refactored its `appendTo` method to a util `resolveAppendTo` so it can be reused.
- `registerScrollHandler` fixes the problem of scrolling the window/document itself
- `resolveAppendTo` fixes the problem of scrolling inside a fixed dialog

mertsincan:
Hi @melloware,
Thanks a lot for your effort! I think we can hide overlays while scrolling and resizing https://github.com/primefaces/primereact/issues/1589

melloware:
OK I can upate this so instead of `alignPanel` call `hide` Let me update this and get back to you.

mertsincan:
This will be better in performance. Thank you ;)

melloware:
OK I updated them all to `hide` instead of `alignPanel` and I optimized hide by checking for `panel.is(':visible')` before trying to do any code.

mertsincan:
Looks good!!! Thanks @melloware ;) You can merge it.

PR #6594, opened 2020-12-02
Fix #6593 - UIPageableData: new class

### LINKED ISSUE #6593: UIPageableData: new class
**Description**
UIData is an enchanced version of `javax.faces.component.UIData`. Actually, it's not quite accurate. UIData "supports" also pageable  feature, and fill few holes starting from JSF 2.0 (bug & enchancements). As of today, it's hard to make a distinction between PF feature and JSF impl. fixes.

Another thing, is that a lot of classes inherits from UIData PF while they don't use any features (like `p:columns`). Basically, I would like to keep UIData from PF as small as possible to clearly see the gap between UIData from JSF and PF.

Rapster:


Rapster:
@christophs78 I was wondering if I should move methods like setRows/setFirst/resetRows in UIPageableData? I know you recently modified that for #5465. Those modifications were made when pageable datatable is used right?

christophs78:
> @christophs78 I was wondering if I should move methods like setRows/setFirst/resetRows in UIPageableData? I know you recently modified that for #5465. Those modifications were made when pageable datatable is used right?

Yes, makes sense.

melloware (approved):

Rapster:
I think this one is good to go @christophs78 @tandraschko ?

tandraschko:
A comment on UIPageableData  would be nice, otherwise it looks good

PR #6601, opened 2020-12-03
Fix #6600: Add CSS class ui-selectonemenu-item-group-children

### LINKED ISSUE #6600: SelectOneMenu: add CSS class to children of SelectItemGroup
**Description**
`SelectItem`s could be grouped in `SelectItemGroup`s. But `SelectOneMenu` renders everything in a flat list instead of a nested list. Therefore it is not possible to tell which select item belongs to a group and which select item is not part of a group.

As it would be really cumbersome to change the markup and to adjust all the javascript to handle the new markup, I suggest to simply add a CSS class to the select items that belong to a group - e.g. `ui-selectonemenu-item-group-children`.
That class could be used to indent the children accordingly.

We do not need to know the indention level, as optgroup may not be nested anyway (the standard HTML select of the browser renders them on the same level)

**Additional context**

grafik

On the left side you cannot recognize, that H is not part of the group G. Both items are on the same level.
On the right side, G1 (the one and only child of G) is indented and so you can easily recognize which item belongs to which group.

fanste:
to optgroup children

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/forms/forms.selectonemenu.js
+++ b/src/main/resources/META-INF/resources/primefaces/forms/forms.selectonemenu.js
@@ -1347,17 +1347,17 @@ PrimeFaces.widget.SelectOneMenu = PrimeFaces.widget.DeferredWidget.extend({
      * Renders Panel-HTML-code for SelectItems.
      * @private
      * @param {JQuery} parentItem An parentItem (select, optgroup) for which to render HTML-code.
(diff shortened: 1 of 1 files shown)


melloware (changes requested):
How about instead of `isIndented` something else.  Because even though your intent is to style it indented what if someone else chose to just make them italics or some other font.  `isGrouped` maybe?

fanste:
Renamed the argument as suggested.

melloware (approved):

Issue #223, opened 2015-05-03
DataTable: discuss default rowKey mechanism / use JSF converter for rowKey and rowData


djmj:
What about enhancing this approach using a JSF converter since it follows the same conversion idea of JSF and its components and provides a clean reusable interface?

In my generic DAO's my overriden methods simply call the converter method.
A converter could also be passed to datatable attribute in markup as in any other component.

```
public class MyJsfGenericDAO <T extends Entity> extends LazyDataModel<T>
{
    private final Converter converter;

    public MyJsfGenericDAO(...)
    {
        this.converter = new MyEntityConverter(entities);
    }


   @Override
   public Object getRowKey(final T t)
   {
      return this.converter.getAsString(FacesContext.getCurrentInstance(), null, t);
   }

   @Override
   public T getRowData(final String rowKey)
   {
      return (T) this.converter.getAsObject(FacesContext.getCurrentInstance(), null, rowKey);
   }
}
```

http://forum.primefaces.org/viewtopic.php?f=3&t=41556&p=132065

Another idea is to add a default rowKey mechanism via Entity#hashCode.

Probably both makes sense.

melloware:
No upvotes in 5 years and I don't see any real advantage of changing the code to support this.

djmj:
I also dont see much advantage to it or it could even make smaller projects / tables more complicated then just using attribute _rowKey_ directly. 

The only positive thing i see is the JSF consistency across data conversions with an attribute like _converter_.

Rapster:
Agreed, I'll coment later on this

tandraschko:
we dont need to do it for 9.0 but i like the approach
often you already have a converter available for a Entity e.g. and this can just be reused (@FacesConverter(for = MyEntity.class))

Rapster:
I'll assign it to you Thomas, as I don't see the added-value of this. I see why `Converter` comes in mind as it looks very similar to what we do with selection, but turns out, with multi-selection `LazyDataModel#getRowData` is not optimized at all, and in normal mode, `DataTable#rowKey` is good enough,

tandraschko:
I dont plan to work on it for now

tandraschko:
This can easily implemented via:

```
public abstract class LazyDataModel<T> extends ListDataModel<T> implements SelectableDataModel<T>, Serializable {

    private Converter converter;

    public LazyDataModel() {

    }

    public LazyDataModel(Class<T> modelClass) {
[...]

PR #6607, opened 2020-12-05

### LINKED ISSUE #6169: DataTable: Lazy getRowData Repeat the query twice

Rapster:
New PR replacing #6571

>@melloware Can you explain your solution? Is this just moving all the code around so the method doesn't get called from the datatable and the feature?

> @Rapster  Fair point! I'm going to start with this: there is no way to prevent calling `DataTable#getRowData`. My first attempt was to use a cache to prevent this and have better perf, but as the model is reseted all the time (by design I'm guessing) my PR didn't bring a lot to the table.
> 
> Now, my idea is to avoid scenarios like expensive call to differents endpoints (e.g database, REST etc.), turns out the result sounds like a lot like SelectItemsConverter from Omnifaces does, except it's not based on toString() representation but either `DataTable#rowKey` or `System#identityHashCode`.
> 
> To sum up:
> - users no longer need to implement `LazyDataModel#getRowKey` and `LazyDataModel#getRowData`. As of today, either you need to set `DataTable#rowKey` or `LazyDataModel#getRowKey` if lazy mode, from now on one entry point: `DataTable#rowKey` and yet, this is optional.
> 
> - Whenever selection has to be decoded, `LazyDataModel#getRowData` will under the hood, loop over Datatable rows till it finds a matching result by using rowKey strategy that has been set with `DataTable#rowKey`. You might wonder what is that cache in DefaultSelectableModel, let's say i select all elements in reverse order (from bottom to top) time complexity will remain O(N), today this scenario is O(N²) _(because it'll keep iterating over elements even though some have been visited)_
> 
> Back to the issue, no query will be executed at all

### DIFF
--- a/docs/9_0/components/datatable.md
+++ b/docs/9_0/components/datatable.md
@@ -522,7 +522,7 @@ keeps previous selections same as selecting a row with mouse click when metakey 
 RowKey must be a unique identifier from your data model and used by datatable to find the selected
 rows. By default, and if selection enabled, rowKey is set to his [identity hash code](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#identityHashCode-java.lang.Object-).
 You can either define your own key by using the rowKey attribute or by binding a data model
-which implements _org.primefaces.model.SelectableDataModel_.
+which implements _org.primefaces.model.RowKeyAble_.
 
 ## Dynamic Columns
 Dynamic columns is handy in case you can’t know how many columns to render. Columns
--- a/docs/9_0/components/datatable.md
+++ b/docs/9_0/components/datatable.md
@@ -520,9 +520,10 @@ keeps previous selections same as selecting a row with mouse click when metakey 
 
 ## RowKey
 RowKey must be a unique identifier from your data model and used by datatable to find the selected
-rows. By default, and if selection enabled, rowKey is set to his [identity hash code](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#identityHashCode-java.lang.Object-).
+rows. By default, and if selection enabled, rowKey is set to bean's [hash code](https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html#hashCode--) _(so better to have it implemented)_.
 You can either define your own key by using the rowKey attribute or by binding a data model which implements _org.primefaces.model.RowKeyMapper_.
-For multiviewstate datatables, it's better to define a deterministic rowKey if you want to be able to retrieve your previous selection.
+
+> In some scenarios, like multiviewstate or inline cell editing, it's better to define a deterministic rowKey _(e.g entity's id)_, to ensure you'll be able to retrieve your previous selection even though bean changed.
 
 ## Dynamic Columns
 Dynamic columns is handy in case you can’t know how many columns to render. Columns
--- a/docs/9_0/components/datatable.md
+++ b/docs/9_0/components/datatable.md
@@ -519,8 +519,9 @@ keeps previous selections same as selecting a row with mouse click when metakey 
 "checkbox" mode allows row selection with checkboxes only.
 
 ## RowKey
-RowKey should a unique identifier from your data model and used by datatable to find the selected
-rows. You can either define this key by using the rowKey attribute or by binding a data model
+RowKey must be a unique identifier from your data model and used by datatable to find the selected
+rows. By default, and if selection enabled, rowKey is set to his [identity hash code](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#identityHashCode-java.lang.Object-).
+You can either define your own key by using the rowKey attribute or by binding a data model
 which implements _org.primefaces.model.SelectableDataModel_.
 
 ## Dynamic Columns
--- a/docs/9_0/components/datatable.md
+++ b/docs/9_0/components/datatable.md
@@ -522,7 +522,7 @@ keeps previous selections same as selecting a row with mouse click when metakey 
 RowKey must be a unique identifier from your data model and used by datatable to find the selected
 rows. By default, and if selection enabled, rowKey is set to his [identity hash code](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#identityHashCode-java.lang.Object-).
 You can either define your own key by using the rowKey attribute or by binding a data model which implements _org.primefaces.model.RowKeyAble_.
-For multiviewstate datatables, it's better to define a deterministic rowKey if you want to able to retrieve your previous selection.
+For multiviewstate datatables, it's better to define a deterministic rowKey if you want be to able to retrieve your previous selection.
 
 ## Dynamic Columns
 Dynamic columns is handy in case you can’t know how many columns to render. Columns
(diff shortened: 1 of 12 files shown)


melloware:
`RowKeyAware` seems to be a name that is consistent with our other interfaces like `TouchAware` `FlexAware` etc?

Rapster:
I thought of that too, but most of these `xxxAware` are used for components while it's not the case here so I didn't want create any confusion. Being said, we could have a `RowKeyAware` extending `RowKeyAble` which adds `getRowKey()` method. But so far, we only have datatable like this, Tree/TreeTable handle selection differently, so I don't think it worth it for now...

melloware:
No I agree its not worth introducing another Interface was just trying to come up with a better name than `RowKeyAble`

Rapster:
What about  `RowKeyMapper`?

melloware:
I like `RowKeyMapper` better than `Able`!

tandraschko:
JFYI: I likely wont review + apply it for 9.0, we are already in bugfix mode here and holiday time is near.

Rapster:
I think it would be nice to have this for next release, as it's a good addition for the component, and users will be happy to no longer implements getRowKey and getRowData. If @tandraschko you don't have time, maybe someone could have a try in case I missed something?

tandraschko:
Im unsure if we should support identityHashCode as it doesnt work most of the times.
It only works when your model is "longer scoped" then requestscoped and no LazyDataModel is used. IMO its better to force a rowKey.

Rapster:
As long your `DataTable#value` doesn't change every request, _(and I don't know about how other users do it)_, but in my case whatever holds data is ViewScoped, so it will work all the times. But even if it's RequestScoped, it's just a default rowkey to ease your devs, nothing forces you to use that key, you can always set your own rowKey in such scenario (and in some cases like MVS, it's actually the thing to do, see doc)

tandraschko:
it just will only work for one single case:
`List<Model>` which is in View or SessionScoped

everything else should be broken and `List<Model>` in a longer scope is probably the worst implementation a user could do.

Rapster:
You mean `List<Model>` in a ViewScoped bean is the worst implementation a user could do?

tandraschko:
in most of the times, yes
the source of a DataTable is mostly from a Database and therefore LazyDataModel is the way to go.
Otherwise you have "all pages" and rows saved in the view and/or session, which bad for memory and often replicated to other nodes.

Rapster:
I could use `Object#hashCode` as default actually, it's more deterministic _(if Object#hashCode implemented of course)_ Obviously, it might not work for cell editing, but just like MVS, `DataTable#rowKey` needs to be defined and as long users are aware of that it's good enough

tandraschko:
IMO we should align the documentation that rowKey can be implemented to use hashCode in the applicaitno-code for this basic usecase and we should still force rowKey to be implemented.

hashCode is of course a better here but, as you already said, it only works if implemented. And when we still force rowKey to be implemented in the application, the users are aware that hashCode/or custom rowKey needs to be implemented correctly.

tandraschko:
@Rapster i will revert this for now. We didnt finish the discussion yet, so please just dont merge the PR next time.

tandraschko:
We could also think about making rowkeymapper a converter

Rapster:
Ok, what's wrong?

tandraschko:
i only reviewed the hashCode part of this whole PR and we didnt even agree on this single part yet ;D

tandraschko:
some general thoughts:

1) it would be great if you could write more text on a PR, its hard to follow a bigger refactoring just by reading code
2) RowKeyMapper can be replaced by Converter, AFAIR we had a issue about it
3) LazyDataModel#getRowData really belongs to LazyDataModel as it it lazy-loads a single Entity
4) RowKeyMapperDataModel, even not sure why you introduced it, has a cache which can hold wrong data therefore we had 3)

Rapster:
> it would be great if you could write more text on a PR, its hard to follow a bigger refactoring just by reading code

Sorry I didn't realize it was that big... If I'm not mistaken, I explained the whole solution in the original PR

> RowKeyMapper can be replaced by Converter, AFAIR we had a issue about it

I think you're refering to [this](https://github.com/primefaces/primefaces/issues/223). I thought about it, indeed it has both methods similar to getRowKey and getRowData but since `SelectableDataModel` was already there, I followed the same idea. I can tweak something using `Converter`

> LazyDataModel#getRowData really belongs to `LazyDataModel` as it it lazy-loads a single Entity

Not really, `LazyDataModel` is all about fetching data depending certain criterias and nothing more. Making the correspondence between an `Object` and a `String` representation _(used for selection for example)_ is the responsability of something else like a Converter/RowKeyMapper. That's why I came up with `RowKeyMapperDataModel`, there is a clean separation between who's responsible for handling selection, and another for loading data. But also, there is two entry point for setting rowKey, `DataTable#rowKey` and `LazyDataModel#getRowKey` which make it more confusing to use the feature. `DataTable#getRowKey` should be the only entry point. In other words, whether it's lazy or not, it goes through the same code: `RowKeyMapperDataModel`

> RowKeyMapperDataModel, even not sure why you introduced it, has a cache which can hold wrong data therefore we had 3)

Explained in the orignal PR, in two words, it's for performance reasons _(I'm going to rename `cache` to avoid any confusion)_. One thing, I could improve on that front is as to when I should clear it up, I reverted that commit... But if you have a closer look, you'll realize `UIData#setDataModel(null)` is constantly called, so that's why I didn't really bothered as to when I should clear it up. So holding wrong data is very unlikely to happen, specially that it's only used for decoding selection (String to Object)

P.S: I attached the explanation of the solution from the original PR in here (which I should've probably done in the first place 😅 )

melloware:
@Rapster yes that is the original issue if you do support converter let's reopen that ticket and link the PR.

Rapster:
@melloware Basically what I have in mind so far, is just to make `RowKeyMapperDataModel` implements `Converter`, and nothing more _(RowKeyMapper becomes useless, therefore removed)_. I don't intend to support converters for `DataTable` for this use case. @djmj You might have an opinion also on this PR.

melloware:
Yep that is fine with me!

PR #6618, opened 2020-12-08
Fix #6612 - FileUpload agnostic to FileTypeDetector providers

### LINKED ISSUE #6602: Is it possible to use primefaces with thymeleaf?
I am using thymeleaf a the rendering engine of my springboot project.

Is it possible to use primefaces? how?

### LINKED ISSUE #6612: Overview Mime-Types over Tika?
It only has ~60kb instead of 7mb TIKA. https://github.com/overview/mime-types

Could some check this? Does it cover all cases like TIKA?

Rapster:


tandraschko:
I like the very much.
I just think we should do a ClassForName if the ServiceLoader doesnt contain our configured FILE_TYPE_DECTECTOR

tandraschko:
and we should not remove TIKA from the optional dependencies, we should document it somewhere that TIKA and mime_types can be used to make it more secure

Rapster:
I thought about that, but I think it's a better practice to register detector with SPI like it is mentionned in the javadoc. Regarding your last statement, I can put it back even though they're not optional dependencies, in fileupload doc it's mentionned that using Tika is recommended ( i can update with mime types)

tandraschko:
if we only rely on the SPI, than we dont need the context param.
a user can then create his own services file and create his own FileTypeDetector with ordering etc.

Rapster:
Even though you create your own FileTypeDector, there is no guarantee it'll be the first to be executed. For instance, you have 3 SPI including custom FileTypeDector, your custom might never be called because Tika has been able to resolve mimetype before (based on Files#probeContentType impl)

tandraschko:
the first loaded services file should be this from the webapp actually
if we cant rely on that, then we should also not rely that the ServiceLoader contains the configured one from web.xml context path and add a ClassForName as fallback (and we added this TIKA hack only for this case which occured in a environment)

Rapster:
Alright I did a little test, you can control ordering through an SPI file! :-D For example, user might want to set this specific order:
```
org.apache.tika.filetypedetector.TikaFileTypeDetector
org.primefaces.showcase.Fake1
org.primefaces.showcase.Fake2
```

Others will be appended/discovered later. So yes, i'll get rid of the context-param (which is even better)

tandraschko:
looks good to me but we should also add the hint with the ordering to the docs

melloware (approved):

PR #6623, opened 2020-12-08
Fix #6612: Detect Tika for magic bytes check

### LINKED ISSUE #6612: Overview Mime-Types over Tika?

melloware:


melloware:
Our other choice is to just be consistent and always use file extension and just let Tika do what it does?

Rapster:
I'm thinking more the other way around, always remove the extension and mime-types do something about not based his result on extension, and if NIO returns null or application-octetstream that's good: you shouldn't accept a file that a content analyzer is not be able to figure out

melloware:
Actually I think Mime Types does work without the extension the test I was trying was this:
```
//FIXME PNG not recognized by Apache Tika?
//        Assertions.assertTrue(FileUploadUtils.isValidType(fileUpload, "test.Png", png));
```

PNG apparently is not detected by Magic Bytes in Tika or Mime Types.

PR #6630, opened 2020-12-09
Fix #6629: InputMode attribute to relevant components

### LINKED ISSUE #6629: Mobile: HTML5 InputMode attribute
**Description**
Add `inputmode` to `AbstractPrimeHtmlInputText`.

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode

The `inputmode` attribute tells the browser on devices with dynamic keyboards which keyboard to display.  This is useful hint on mobile devices.

I already added it to `InputNumber` so it can support either numeric or decimal keyboard only but I would like to move the prop up and allow you to do something like:

```xml
<p:inputText id="expiry" placeholder="MM/YY" inputMode="numeric"/>
```

melloware:


tandraschko:
+1

PR #6632, opened 2020-12-09
Fix #6631: add vertical and horizontal scroll to timeline component

### LINKED ISSUE #6631: Timeline: Vertical and horizontal scroll
https://visjs.github.io/vis-timeline/examples/timeline/groups/visibleGroups.html

buunsal:


src/main/java/org/primefaces/component/timeline/TimelineBase.java
```diff
@@ -212,6 +214,22 @@ public void setMaxHeight(Integer maxHeight) {
...
 
+    public Boolean getHorizontalScroll() {
+        return (Boolean) getStateHelper().eval(PropertyKeys.horizontalScroll, false);
+    }
+
+    public void setHorizontalScroll(Boolean horizontalScroll) {
+        getStateHelper().put(PropertyKeys.horizontalScroll, horizontalScroll);
+    }
+
+    public Boolean getVerticalScroll() {
```
melloware: `isVerticalScroll` and `isHorizontalScroll` instead of `get`

melloware (changes requested):

src/main/java/org/primefaces/component/timeline/TimelineBase.java
tandraschko: If "boolean" is used, "is" prefix should be used. If "Boolean" is used, "get" prefix should be used.
So its correct actually

Rapster: Use `boolean`, not the wrapper 😉

melloware: Yes i would convert to lower b boolean and if false do not send the widgetVar since that is the default rather that checking for null.  then the "is" makes it clear this is a boolean flag.  Just my two cents.

buunsal: I replaced `Boolean`s with `boolean`s.

src/main/java/org/primefaces/component/timeline/TimelineRenderer.java
```diff
@@ -207,12 +207,8 @@ protected void encodeScript(FacesContext context, Timeline timeline) throws IOEx
...
             wb.attr("maxHeight", timeline.getMaxHeight());
         }
-        if (timeline.getHorizontalScroll() != null) {
-            wb.attr("horizontalScroll", timeline.getHorizontalScroll());
-        }
-        if (timeline.getVerticalScroll() != null) {
-            wb.attr("verticalScroll", timeline.getVerticalScroll());
-        }
+        wb.attr("horizontalScroll", timeline.isHorizontalScroll());
+        wb.attr("verticalScroll", timeline.isVerticalScroll());
```
melloware:
```java
 wb.attr("verticalScroll", timeline.isHorizontalScroll(), false);
 wb.attr("verticalScroll", timeline.isVerticalScroll(), false);
```
This won't send the JSON through network traffic if the value is false which is the default.

melloware (changes requested):

src/main/java/org/primefaces/component/timeline/TimelineRenderer.java
buunsal: Thanks for your advice. I made the requested change.

melloware (approved):

melloware (changes requested):
Don't forget to update /docs/9_0/components/timeline.md to add these two new properties.

melloware (approved):

Rapster:
Hi @mertsincan 

We're trying to keep commits as consistent as possible and keep the commit history clean, next time please use Squash & Merge feature:

image

PR #6635, opened 2020-12-09
Fix #6633: Improve drag and drop cancel

### LINKED ISSUE #6633: PickList: Transferred item rolled back if transfer event refreshes a DataTable
**Describe the defect**
A `p:pickList` with a `transfer`-event that updates a `p:dataTable` is totally broken due to a change that was implemented for issue #5582. The comment on the changed code is
```
// #5582: destroy any current draggable items
```
And that is the root cause of my issue. It really destroys ANY currently active drag'n'drop.

**Reproducer**
If really required, I will try to setup a small testcase.

**Environment:**
 - PF Version: 9.0-SNAPSHOT
 - Affected browsers: ALL

**To Reproduce**
- Item is dragged from source list to target list
- On drop, the transfer-event kicks in and updates a datatable (that is required, because the datatable provides additionaly input field based on the selected items from the picklist)
- The datatable is updated, the create widget function of primefaces is invoked, which invokes the refresh function of the datatable, which again invokes unbindEvents that contains the code from #5582.

This is really hard to debug and therefore I made some images:
This image is taken right after the drop happend. It was captured using firefox' dev-tool feature to halt on DOM changes. The items was really moved and contained in the target list. It even appears on the server side in the listener-method if the event.
grafik

After the ajax request returns, the dropped item will be removed again from the target list
grafik

Right after that `f.appendChild(o)` the item is back at the old position of the source list.

fanste:
It compares the parent of the currently dragged item with the `jq` of the datatable to determine if this drag and drop affects the datatable.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
+++ b/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
@@ -385,7 +385,7 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
         // #5582: destroy any current draggable items
         if (this.cfg.draggableColumns || this.cfg.draggableRows) {
             var dragdrop = $.ui.ddmanager.current;
-            if (dragdrop) {
+            if (dragdrop && ($.ui.ddmanager.current.currentItem.closest('.ui-datatable')[0] == this.jq[0])) {
                 document.body.style.cursor = 'default';
                 dragdrop.cancel();
             }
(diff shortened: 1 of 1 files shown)


melloware:
I am testing this.

fanste:
`currentItem` is not always set. Sometimes it is `element`. New version is currently tested  (EDIT: internally).

melloware:
I didn't see you commit anything?  Yes for column headers dragging `currentItem` is not set.

melloware:
I made some minor enhancements.  I tried it out with both column re-odering and row reordering and it works great.

fanste:
Yes... these changes make perfect sense. Thanks for testing and improving the fix.

PR #6639, opened 2020-12-10
Fix #6637 - DataTable: multi-sort does not always respect priority

### LINKED ISSUE #6637: DataTable: mutli-sort does not always respect priority
The whole refractoring regarding #5438 reverted the functionality of #4557.

1. `UITable.getSortMetaAsString` now returns ALL sort meta instead of only the active sort meta.
2. `UITable.getSortMetaAsString` returns the sort meta at a random order, not according the priority.
3. `LazyDataModel.load#multiSortMeta` is a `HashMap` and a user has to sort the meta according the priority before it could be applied to e.g. the DB-query
4. It seems to work (untested, only based on code) for internal sorting via `SortFeature.sort` as it converts the sort meta map to a `LinkedHashMap` sorted using `SortMeta.compare`

In the most cases (if not all), the sort meta has to be applied according to the priority or the result is not the desired one. Multiple possibilities: 
- Why not always provide a sorted map (`LinkedHashMap`)? => sort it after the initial map was built and after sort feature was decoded
- Provide at least a utility to sort the map according to the priority so not every user has to implement the priority-sorting.

Regarding error 1 & 2: The returned string must always contain the sort meta sorted by priority. Additionally it should only contain the active sort meta or the `datatable.js` has to be fixed as it only expects active sort meta.

Rapster:


melloware:
I think this is safe to merge so @fanste can test and I can re-run the integration tests.

Rapster:
ITs should pass, I took the habit the run them 😉

tandraschko:
@Rapster could you also adjust the same part in the treetable?

Rapster:
Not sure what I'm supposed to do, I think it's safer if you do it 😅

tandraschko:
@Rapster i did it for you now. For the next PRs regarding filter/sort, people must adjust both TreeTable and DataTable

PR #6651, opened 2020-12-11
Fix #6650: LinkButton duplicates onclick event

### LINKED ISSUE #6650: LinkButton: onclick attribute gets duplicated
**Describe the defect**

onclick attribute gets duplicated on `<span />` and `<a />` rendered tags.

**Environment:**
 - PF Version: _8.0_
 - JSF + version: _Mojarra 2.2.8_
 - Affected browsers: _Chrome_

**To Reproduce**

Copy & paste example code.

**Expected behavior**

onclick attribute should be rendered only once.

**Example XHTML**
```html
<?xml version='1.0' encoding='UTF-8' ?>
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:h="http://xmlns.jcp.org/jsf/html" xmlns:p="http://primefaces.org/ui">
<h:body>
	<p:linkButton id="print" onclick="window.print();" value="Print" />
</h:body>
</html>
```

**Rendered HTML**
```html
<body>
    <span id="print" class="ui-linkbutton ui-button ui-widget ui-state-default ui-corner-all ui-button-text-only" onclick="window.print();">
        <a href="/htmc/test.xhtml" onclick="window.print();">
            <span class="ui-button-text ui-c">Print</span>
        </a>
	</span>
	<script id="print_s" type="text/javascript">
	$(function() {
		PrimeFaces.cw("LinkButton", "widget_print", {
			id: "print"
		});
	});
	</script>
</body>
```

melloware:


tandraschko:
unsure, why not render click on the `a`? i think it belongs a bit more there

melloware:
OK i will switch it. I didn't know which to put the click but the 'A' is good with me.

melloware:
switched and removed from SPAN and left onclick on A link.

PR #6656, opened 2020-12-13
Fix #6575 - Datatable : Clean up selection

### LINKED ISSUE #3917: DataTable: preselection does not take account "disableSelection"
## 1) Environment
- PrimeFaces version: 6.2
- Does it work on the newest sources in GitHub? (Build by source -> No https://github.com/primefaces/primefaces/wiki/Building-From-Source)
- Application server + version: Tomcat 8
- Affected browsers: All

## 2) Expected behavior
 Scenario is as such:

- DataTable containing a bunch of data, each one of them can be added to a cart
- Select an element and add it to the cart
- From there, it shouldn't be possible to select this item (as it is in a cart: this is our requirement). We're using `disabledSelection` to avoid that...
- But after refreshing the datatable, the previous item should no longer be selected

## 3) Actual behavior
After refreshing the datatable, the previous item is still selected. 

## 4) Steps to reproduce
I used this expression to fake the cart feature: `disabledSelection="#{car.id eq dtSelectionView.selectedCar.id}"` in https://www.primefaces.org/showcase/ui/data/datatable/selection.xhtml (See RadioButton sample)

## 5) Sample XHTML
See 4

## 6) Explanations
There is at least an easy explanation, the managed bean being ViewScoped, selected items lives as long we're using the same view: this is why the last selection is kept. But preselection in `DataTable` should consider `disabledSelection`.

### LINKED ISSUE #5868: Datatable: disabledSelection with initial selection removes active ui-state-active-class to the checkboxes
Reported By PRO User;
> We have encountered a difference between versions 6.2.x and 7.0.x that cause a graphic bug for the checkbox element.
In the 6.2.x version, when we check the checkbox and then we disable it, the checkbox keeps both the “ui-state-active” and “ui-state-disabled” CSS classes.
It doesn’t show the same behavior in the 7.0.x version: if a checkbox is checked it has the “ui-state-active” CSS class, but when we disable it, that class is overridden by the “ui-state-disabled” one.
We expect to have both CSS classes in the checkbox element.
We have encountered this bug in the “dataTable” component by defining the attribute “disabledSelection” and we have found this difference by testing both PrimeFaces (JSF) versions.

### LINKED ISSUE #6575: DataTable: clean up selection

### LINKED ISSUE #6682: DataTable: force selection while disabled checkbox
**Describe the defect**
Even though check are disabled, you can still force selection by tweaking selection holder in HTML:

1. Go here: http://localhost:8080/showcase/ui/data/datatable/selection.xhtml
2. Disable selection with whatever condition you like (mine was `disabledSelection="#{car.id eq car.id}"` 😅 )
3. Modify `form:checkboxDT_selection` input value with a few ids
4. Click view: elements are selected

Rapster:
- Centralize selection code into SelectionFeature
- Keep selection state in view scope, and no longer request scope

No breaking changes


src/main/java/org/primefaces/component/datatable/DataTable.java
```diff
@@ -294,18 +297,6 @@ public void processValidators(FacesContext context) {
         }
     }
 
-    @Override
```
Rapster, line 297: @tandraschko @melloware I remember you guys faced an issue with this code recently, but I can't recall what it was... Can you remind me what was the issue number?

src/main/java/org/primefaces/component/datatable/DataTable.java
```diff
@@ -294,18 +297,6 @@ public void processValidators(FacesContext context) {
...
-        super.processUpdates(context);
-
-        ValueExpression selectionVE = getValueExpression(PropertyKeys.selection.toString());
-
-        if (selectionVE != null) {
-            selectionVE.setValue(context.getELContext(), getLocalSelection());
-            setSelection(null);
-        }
-    }
-
```
Rapster, line 307: @tandraschko @melloware I remember you guys faced an issue with this code recently, but I can't recall what it was... Can you remind me what was the issue number?

tandraschko: can only remember that the selection attr was broken in ui:repeat.

Rapster: Got it #3812

tandraschko:
cant we get rid of selectedRowKeysAsString? It was easy to remove it for resizable/visible columns at least

Rapster:
I'll check and keep you posted

Rapster:
I think I'm done with this one. Ready for review

tandraschko:
did you run the integration tests? Would be great if you could add some more integration tests, so we can be sure that nothing breaks

BTW @melloware / @christophs78 big kudos for the many integration tests! pulled the project today for fixing sorting issues and wow, we really have many of them now.

melloware:
The integration tests have come in really handy!  I think we have about 150 tests running nightly now!

melloware:
@Rapster we have "single" and "multiple" selection examples already in Integration Tests like DataTable002Test and DataTable010Test so you could probably modify one of those or copy it.

tandraschko:
NVM; is meant saveState :D

tandraschko:
the PR looks good; we should just add some integration tests before merge; some that proof selection works like before and 1-2 new ones, which proofs that update works.

tandraschko:
if RC1 is released, its too late for such refactorings; i would really like to see this merged before.

tandraschko:
and as addRow and editRow was changes, 1-2 simpel integration tests for this feature would be great too.

melloware:
@tandraschko We already had an Edit Row test so I just added an Add Row test as well.  

They are in `DataTable007Test` here: https://github.com/primefaces-extensions/primefaces-integration-tests/blob/6d4c8dfb8e7914c77f6b86501f7985c62506284e/src/test/java/org/primefaces/extensions/integrationtests/datatable/DataTable007Test.java

melloware:
I just checked out this PR and built it and ran all the Integration tests including our Single Selection and Multiple Selection, Add Row an Edit Row tests and everything is passing!!!

Rapster:
Thank you very much @melloware

src/main/java/org/primefaces/component/datatable/feature/ScrollFeature.java
```diff
@@ -82,7 +78,7 @@ public void encode(FacesContext context, DataTableRenderer renderer, DataTable t
 
     @Override
     public boolean shouldDecode(FacesContext context, DataTable table) {
-        return false;
+        return table.isScrollingRequest(context) && table.isSelectionEnabled();
```
tandraschko: Why did you change all this features?

In the past we always decodde SelectionFeature, just if selection is enabled and indepdent if the current request is a selection request.
So in each request the SelectionFeature was called and decoded if the _selection param is available

Rapster: > Why did you change all this features?

I don't think I change anything, I just adapt according to my changes and move decode stuff in decode method instead of encode method. Arf now that I see it, loading data is also done in encode method... I'll have to change that

> In the past we always decodde SelectionFeature, just if selection is enabled and indepdent if the current request is a selection request.

Maybe, but as of today you have a few features just like this one that requires decoding selection (ex-findSelectedRowKeys). Right now, I just adapt according to my changes, tomorrow everything should probably be done in SelectionFeature

tandraschko: Hmmm thats weird actually.
SelectionFeature#decode runs before all encode, so why does ScrollFeature#encode needs a findSelectedRowKeys call?
Maybe we can remove this calls with your refactoring.

Rapster: I think those calls are to cover default selection, where items has been added right away to the `DataTable#selection` value. For example, you add a new row, and this row should be selected by default, that's what findSelectedRowKeys used to do (and I'm still doing) it's setting new rowKeys through `DataTable#selection` value. But I feel it would be a lot more simple if a new attribute `selectedRow` was added to DataTable, to allow default selection. Item would be then added to `DataTable#selection` once component processed, doing so it would prevent reading `DataTable#selection` at all and figure out what are the rowKeys

tandraschko: That something for V11. I think its better to add this code to DataTableRenderer#preRender or similar isntead of in multiple XXXFeature which are not related to selection.

Rapster: Not that simple, I'd rather come up with a clean solution in V11 than fiddling something now.

tandraschko:
your PR now contains the changes of the other PR which refactored rowKey stuff; im not ok with this for now in this PR.

tandraschko:
Its not good to mix such big changes in general as a default rowKey is not related to "cleanup selection" / "save selection in viewstate"
Updates this issue for it: https://github.com/primefaces/primefaces/issues/223

Rapster:
From the start it's always been more than just saving selection in view state. This PR is part of the "Clean up series", so yes it involves more changes than usual but, if you look closer it's not that big. SelectionFeature is the one which got a big refactoring.

Regarding the rowKey stuff, that's still part of selection so obviously I modified it a tiny bit (and not as much as my other PR), and again it's a nice to have feature, completely transparent internally and for users. Recently I just aligned the behavior in lazy mode,  as I introduced default hashCode in my very first commits. `LazyDataModel#getRowData` gets a default impl, since ii'm guessing  most of the users (80%?) does the same thing.

#223 Default rowKey and support converter are two different things, so not sure this is the best place to talk about it

If default rowKey really bothers you, I can revert it... and I'll simply introduce it later

tandraschko:
`default rowKey` and `converter` are different things but targets the same problem, to default impl this attr/methods.
I think we should avoid the hashCode and the LazyDataModel change or we also work on the Converter thing, that would take your improvement to the next level and is more failsafe as hashCode, if a converter is registered for that object.
HashCode might not be implemented, the Converter IS implemented correctly if available.

That order would be possible:
- use rowKey method/attr if defined
- use converter if available
- fallback to hashCode

Also think the change is quite easy and the converter can be cached as request scoped / field in DataTable.

Rapster:
I'm gonna revert default rowKey so this PR gets merged and we can move on. I modify LazyDataModel so it gets cleaner, let me know if you want me to revert it as well

tandraschko:
ok, looks good to me now

I assume you tested lazy,non-lazy,multi&single-selection, mojarra, myfaces?
then +1

melloware:
Our Integration tests test all of the above but only on MyFaces since Tomee only runs Myfaces. So I cant speak for Mojarra.

Rapster:
I did all my tests using showcase and mojarra

melloware:
OK then lets squash and merge this and I will re-run the Integration Tests on GitHub Actions.

PR #6667, opened 2020-12-15
Fix #6648: Excel option to disable strong type of cells

### LINKED ISSUE #6648: Exporter: Excel export use numeric data type for numeric data
**Description**
This was reported by a customer of mine.  When using the datatable export for Excel it exports all cells as Strings.  This causes Excel 2016 to report that a column  "Number stored as text" error because it knows that cell should actually be numeric.

I would like to update the Excel exporter to handle this and make it a number.

**Current:** (year is treated as a String)
image

**Fixed:** (year is stored as a number)
image

melloware:
@jxmai I added an `ExcelOption` that is `true` by default to enable or disable strong typing of cells.  My thought was call it `StronglyTypedCells` because I can see us also adding `Date` type cells in the future as well.

### DIFF
--- a/docs/9_0/components/dataexporter.md
+++ b/docs/9_0/components/dataexporter.md
@@ -336,6 +336,7 @@ public class CustomizedDocumentsView implements Serializable {
         excelOpt.setFacetFontStyle("BOLD");
         excelOpt.setCellFontColor("#00ff00");
         excelOpt.setCellFontSize("8");
+        excelOpt.setStronglyTypedCells(true);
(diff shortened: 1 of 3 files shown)


jasonex7:
Hi @melloware, could this option be set per column?

melloware:
No right now its at the whole exporter level.  It would have to be new property on each column for you to be able to do that...

jxmai:
missed the notification previously. Thanks @melloware 👍

PR #6670, opened 2020-12-15
Fix #6669: Datatable/Treetable clearFilters clear facets

### LINKED ISSUE #6669: Datatable: clearFilter widget function should support custom filters
Below you will find a repoducer which demontrates that a datatable filter is not cleared when it is created using a facet-filer.

[p-facet-filter-datatable.zip](https://github.com/primefaces/primefaces/files/5697235/p-facet-filter-datatable.zip)

To reproduce the problem:

1) select true or false from the selectonemenu of colum 1 and enter "00" into the filter of the second column
2) now use the radio-buttons on/off which are simply invoking "clearFilters()"

The filter of column 1 is not clear while the filter of column 2 is cleared.

Many thanks in advance.

melloware:


Rapster:
Do you think we could do in such a way that users don't need to put `.ui-customfilter`? We mostly know what users might use as filters

melloware:
They don't put in customfilter it looks like PF does...
```xml
<f:facet name="filter">
          <h:selectOneMenu onchange="PF('entityVar').filter()">
            <f:selectItem itemLabel="#{messages.noFilterLabel}" />
            <f:selectItem itemValue="true" />
            <f:selectItem itemValue="false" />
          </h:selectOneMenu>
        </f:facet>
```

That is all that is in the XHTML and PF has this..

```java
public static final String COLUMN_CUSTOM_FILTER_CLASS = "ui-column-customfilter";
```

Rapster:
Oh ok cool, I got confused with [showcase](https://www.primefaces.org/showcase/ui/data/datatable/filter.xhtml)

melloware:
This seems to work for basic components like `h:selectOne` and `datePicker` etc but not for advanced components like `SelectOne` or `SelectManyCheckbox`.

melloware:
Also @Rapster while testing showcase on your new Filter demo if in the Year range filter you type "2" and press enter the page blows up.

Rapster:
Indeed, looks new... I'm having a look

Rapster:
Oh yeah I see, no it's not new... I never tried putting something else than a date range...

Rapster:
A way to fix this would probably be to add css a class to those input hidden to recognize `ValueHolder#getValidatableInputClientId` that can be cleared?

melloware:
Actually I think my :input is finding All of them including the hidden. I need to test it again.

melloware:
@Rapster I think we should turn this to return false instead of throw an exception what do you think in RangeFilterConstraint? 
```java
if (!(filter instanceof List) || ((List<?>) filter).size() != 2) {
            throw new IllegalArgumentException("Filter expects a " + List.class.getName() + " with a size equals to 2");
        }
```

Rapster:
Yeah true, I'll fix it. Thanks for reminding me

Issue #6665, opened 2020-12-14
Propsal: further refactoring on DataTable/TreeTable column related state


tandraschko:
Currently we have the following state, which is also stored in the MVS:

```
    private Map<String, SortMeta> sortBy;
    private Map<String, FilterMeta> filterBy;
    private String orderedColumnsAsString;
    private Map<String, Boolean> visibleColumns;
    private Map<String, String> resizableColumns;
```
and orderedColumns is still not refactored.

TBH i dont like further refactorings for 100% as our current code is quite nice and stable compared to 8.0 BUT the next version will be 10.0 and i dont like to refactor everything again. And we have at least 3-4 weeks before a RC1.

we could merge and unify many many code if we would create a
```
private Map<String, ColumnState> columnState;
```

which would contain all those info on sorting and filtering, width, ordering, resize etc. etc.
we could also unify the SortBuilder and FilterBuidler and attributes, we would only have something like:

```
<p:dataTable columnsConfigs="..." />
```

with
```
ColumnStateBuilder......sortOrder(...).filterValue(....).width(200).visible(false).build();
```

maybe the same on p:column for a single entry.

@Rapster @christophs78 @melloware WDYT?

tandraschko:
i know there are many other things to refactor but this is probably the most core thing currently we could do.
I would really like to reduce the code here, especially in UITable.

I would also like to re-use DataTableFeatures in TreeTable but this are internal refactorings, this one here is a big public refactoring.

melloware:
I think this would be a good refactoring to help get the two Tables more manageable. I know @Rapster has been in there way more than me though.

tandraschko:
more managable and more unified columnSettings instead of X builder, just 1 for all

Rapster:
I don't really have an opinion on this since, as I already mentionned, I think TreeTable is simply "wrong" as it is. My project integrate this component with which we're not really happy about... And I'll probably start working on a new version of it based on DataTable in the next weeks/months (that's also one of the reasons why I'm trying to clean up stuff as much as I can to have a better start).

But since I didn't start anything yet, and don't know when I'll be able to finalize something. I guess, for now removing duplicates and make code more consistent is always a good thing 😉

tandraschko:
I have some customers and they are quite happy with it actually ;D

nevertheless, this change propsed here does a major refactoring independent of TreeTable, it should just merge column related state to one place and its builder/attributes.

any ideas about the name?
[...]

PR #6679, opened 2020-12-17
#6665 proposal

tandraschko:


src/main/java/org/primefaces/model/TableState.java
```diff
@@ -0,0 +1,55 @@
...
+ * THE SOFTWARE.
+ */
+package org.primefaces.model;
+
+import java.util.List;
+import java.util.Map;
+import javax.el.MethodExpression;
+import org.primefaces.model.filter.FilterConstraint;
+
+public class TableState {
```
melloware, line 31: Serializable?

src/main/java/org/primefaces/model/TableState.java
```diff
@@ -0,0 +1,55 @@
...
+    
+    private int first;
+
+    private int rows;
+    
+    private Map<String, ColumnState> columns;
+    
+    private Object globalFilterValue;
```
melloware, line 39: transient?

melloware (changes requested):
If we were using Lombok you could just slap `@Builder` and `@Data` on these two classes :)

tandraschko:
please dont review in this detail, that is just to show what data i would like to store in which form :D Far away from anything real.

tandraschko:
i really like the structure in general, im just unsure how a user should define default sort/filter via code now...

PR #6683, opened 2020-12-18
Further Table refactorings

tandraschko:
Whats changed?
- Introduce UITableFeatures, which can be used by TreeTable and DataTable; This is not for 100% complete now but it shows how toggle,resize are completely reused and the decoding part of sort+filter is also reused
- move all code where it belongs to from point of lifecycle; initSortBy and initFilterBy is only called once and initialy created with user,global and MVS data. After that only sort/filterFeature updates it from request data. Before MVS restore was called on each request, which is wrong actually in all MVS components currently. 
- it merges display order, resize width and visibility into ColumnDisplayState (not sure about the name yet) instead of 3 maps
- ColumnDisplayState will be reused soon to maintain the column displayOrder


tandraschko:
sorry, it now also contains the Exporter changes from melloware; im not that familiar with PRs :P
How can i resolve that?

tandraschko:
Maybe you can check the changes now here: https://github.com/tandraschko/primefaces/commit/230f8cac5244bd58d3476ad6e72212ef48a4fc32
dont have much time anymore today

some observations i made:
- Sort and Filter both load data via LazyDataModel, even FilterFeature calls SortFeature... i think this all should be moved to preRender or something, therefore i also made a new UITableFeature which doesnt contain that stuff...

tandraschko:
@Rapster please review
i think this a good way to the correct way when to restore and store viewstate, multiviewState

Rapster:
> sorry, it now also contains the Exporter changes from melloware; im not that familiar with PRs :P
How can i resolve that?

What I would do:

- Do a reset mixed
- Stash your changes
- Create a new branch
- Apply your stash, and commit & push (create a new PR)

It'll get more messy if you don't do it now

And rebase everytime from primefaces/master upstream, see https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/configuring-a-remote-for-a-fork

Regarding your changes, I'll have to fork your branch to have a clear idea what the changes are all about.

Considering upcoming release is in a few weeks, and vacation is here. I would keep those changes for next release (11?)

tandraschko:
You can simply check my linked single commit.

Currently i dont see any commits from PrimeTek about the new components, so we have enough time.
This code might not fix existing Bugs but it fixes the lifecycle of how this maps are managed. its also to base to make displayPriority possible and reduce code.
So its quite Important for me.

I will also make some cleanup Tasks for the lazy loading stuff mentioned above for v11

christophs78:
> sorry, it now also contains the Exporter changes from melloware; im not that familiar with PRs :P
> How can i resolve that?

My way :-)
- create a new branch based on master
- cherry-pick your relevant commit(s) from the existing branch (quite easy - at least with IntelliJ)

@melloware´s approach sound good too.

tandraschko:
Nvm
I think i mixed to many "Baustellen" (German Word)
I will create some smaller issues, would be great If you could take one @Rapster , i will assign it later to you

PR #6685, opened 2020-12-18
Sonar Fixes: map.computeIfAbsent()

melloware:


tandraschko:
I wont make a diff here but in general computeIfAbsent is slower

melloware:
Interesting under the covers I just assumed it was doing `containsKey` etc so it would be the same just more concise.  i will have to read up on this.

tandraschko:
Also its misused here, its already computed

melloware:
OK i will kill this PR.

PR #6687, opened 2020-12-18
Fix #2061: Datatable allow click and double click together

### LINKED ISSUE #2061: DataTable: rowSelect & rowDblselect doesnt't work together
This is not a new issue in 6.0 but it does exist in 5.3 as well. but i was hoping that it will be addressed in the latest 6.0 release.

If we need to have both  single click and doubleClick event in data table only one works. is there any work around to this? in my application we need to achieve two separate task on single click vs double click. Thanks.

melloware:
Similar fix to what I did for Schedule. I heavily tested this and updated showcase example.

Today with both events together you get 3 AJAX events `RowSelect`, `RowSelect`, `RowDblSelect` which is not desirable.  You want to distinguish RowSelect from a Double Select event.

1. `RowSelect` without `RowDblSelect` works the same as it does right now with no delay
2. `RowDblSelect` without `RowSelect` works the same as it does today.
3. Both together adds a delay while single clicking to see if you are going to DoubleClick before it fires the Single click event. If you do double click then only the DoubleClick event is fired.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
+++ b/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
@@ -801,7 +801,8 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
      * @private
      */
     bindRowEvents: function() {
-        var $this = this;
+        var $this = this,
+            hasDoubleClickBehavior = this.hasBehavior('rowDblselect');
 
         this.bindRowHover(this.rowSelector);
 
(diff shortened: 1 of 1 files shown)


tandraschko:
thats something for @mertsincan

melloware:
OK I assigned him to review.

For reference it seems to be the way most people solve it with JQuery using a Timeout:
https://stackoverflow.com/questions/5471291/javascript-with-jquery-click-and-double-click-on-same-element-different-effect

mertsincan:
Hi,
I think the browser should decide when the click and double click events are invoked. We shouldn't interfere with this with Timeout. This is my personal opinion :D 

Please check the following behavior;
```
<!DOCTYPE html>
<html>
<body>
     <button ondblclick="console.log('doubleclick')" onclick="console.log('click')">Test</button>
</body>
</html>
```

melloware:
@mertsincan Yep I agree that is what the browser does if fires 2 click events and 1 double click event.

image

But to me when trying to use both that doesn't make sense. I think most users expect it to work how Windows or MacOS works.  where the two events are separate and click is not triggered when you double click.

This only adds the timer if you are actually using DoubleClick event else it remains the same.    I am fine with closing this PR but we should close the main Issue ticket with WontFix then.

PR #6694, opened 2020-12-20

### LINKED ISSUE #6169: DataTable: Lazy getRowData Repeat the query twice

Rapster:
New PR replacing #6607


tandraschko:
1) RowKeyMapper mentioned in the docs, i assume it should mention how to use a converter now? Do we need a rowConverter attr or is it enough to implement it in the model?
2) #getDataModel now wraps in each call, thats ok but we should do it like everywhere in JSF -> FacesWrapper, DataModelWrapper, unwrapping, and so on
3) RowKey changed to String (ok) but the signature still is `public Object getRowKey(Object object) {`
4) Does the rowKey attr still work fine after the changes for a simple case (List of models bound to a bean, without lazyModel)? i hope that EL doesnt conflict with it in some case (getRowKey(obj) vs getRowKey())

tandraschko:
and yes, i think using the Converter interface is a good approach here and also fixes #223 (reopened it)
we should just think about if we could support a auto-lookup of the converter (@FacesConverter(for=MyEntity.java)), so a user can either define the rowKey attr, use a global converter or the DataModel can implement Converter

Rapster:
I'm gonna close this for now, and focus on #6575 which seems to me more important

PR #6704, opened 2020-12-21
Fix #1895/#6691: Chips allow pasting of delimited list

### LINKED ISSUE #1895: Chips: Paste delimited list to add multiple
Autocomplete is one of the best input component and i have a new idea to increase its usability.

**Feature**

Allow copy and pasting of a delimiter separated list into the textbox to quickly add multiple elements. It is an feature for advanced users to save much input time. 

**A)**

1. Activate _advanced mode_ clicking on an icon (similar to source mode of text-editor). 
2. A normal text input with a delimiter separated list is shown.
3. User can simply edit this list by hand.
4. User leaves _advanced mode_.

user can also reorder elements in this mode

**B)** 

https://select2.github.io/examples.html#tokenizer

**Example**

I have a string in my clipboard like _"test1, test2, test3"_ and an autocomplete with attributes _value="#{bean.list}" multiple="true" delimiter=","_. 

When i paste this string into the textbox the string gets split resulting in three elements.

_"test1"_
_"test2"_
_"test3"_

**Result**

Three new elements are in the final arraylist

**Workaround**

Use a dialog with an input field bound to the same value to allow mass input.

### LINKED ISSUE #6691: Chips: make items editable / toggle source
Chips currently has the disadvantage, that its not possible to edit each item invididually or toggle displaying the _source_ string to do so. 

The only option is to remove the item and add it again at the end but this changes the order of items.

Making each item editable or adding a _toggle source_ button at the end of the component to show it as a delimiter separated list could be a solution.

melloware:
- New attribute `separator` similar to PrimeNG but different in this is the separator like ',' for pasting a list of items into the Chips.
- New attribute `addOnPaste`  if you want to immediately process the items while pasting. Default is false so you can paste and then edit the list.
- New widget method `edit` allowing you to convert the Chips into a separator list for mass editing so you can keep the order of items.

So you could have a chips like this.
image

Call Edit with...
```xml
<p:commandButton value="Edit" onclick="PF('chips').edit(); return false;" />
```

And turn into this..
image

@djmj

### DIFF
--- a/docs/9_0/components/chips.md
+++ b/docs/9_0/components/chips.md
@@ -29,6 +29,7 @@ Chips is used to enter multiple values on an inputfield.
 | inputmode | null | String | Hint at the type of data this control has for touch devices to display appropriate virtual keyboard.
 | max | null | Integer | Maximum number of entries allowed.
 | addOnBlur | false | Boolean | Whether to add an item when the input loses focus.
+| addOnPaste | false | Boolean | Whether to add the items immediately when pasting into the input. Default false.
 | onblur | null | String | Client side callback to execute when input element loses focus.
 | onchange | null | String | Client side callback to execute when input element loses focus and its value has been modified since gaining focus.
 | onclick | null | String | Client side callback to execute when input element is clicked.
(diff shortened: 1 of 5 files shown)


djmj:
Sounds great. Can you also toggle back after editing? Like klicking "edit" button twice.

melloware:
> Sounds great. Can you also toggle back after editing? Like klicking "edit" button twice.

I will change it to `toggleEditor()`.

djmj:
Great thanks and once we updated to newer version we will use this definetly.

melloware:
Integration Tests: https://github.com/primefaces-extensions/primefaces-integration-tests/commit/231e4e7d08873f33f2d0ca54d16b46228a267843

Issue #5455, opened 2019-12-18
CSP: Timeline AJAX Missing CSP nonce


melloware:
Getting "Missing CSP nonce" error when making an AJAX call.

**Steps to reproduce:**
1. Enable CSP in Showcase
```xml
    <context-param>
        <param-name>primefaces.CSP</param-name>
        <param-value>true</param-value>
    </context-param>
```

2. Navigate to http://localhost:8080/showcase/ui/data/timeline/allEvents.xhtml 

3. When the RangeChanged Ajax event is fired you will get the below stacktrace.

**StackTrace:**
```
SEVERE: Error Rendering View[/ui/data/timeline/allEvents.xhtml]
javax.faces.FacesException: Missing CSP nonce
        at org.primefaces.csp.CspState.validate(CspState.java:75)
        at org.primefaces.csp.CspState.getNonce(CspState.java:57)
        at org.primefaces.csp.CspResponseWriter.listenOnEndAttribute(CspResponseWriter.java:185)
        at org.primefaces.csp.CspResponseWriter.write(CspResponseWriter.java:167)
        at org.primefaces.csp.CspPartialResponseWriter.write(CspPartialResponseWriter.java:167)
        at java.io.Writer.write(Writer.java:192)
        at java.io.Writer.write(Writer.java:157)
        at org.primefaces.util.WidgetBuilder.init(WidgetBuilder.java:59)
        at org.primefaces.util.WidgetBuilder.init(WidgetBuilder.java:78)
        at org.primefaces.component.growl.GrowlRenderer.encodeEnd(GrowlRenderer.java:67)
        at javax.faces.component.UIComponentBase.encodeEnd(UIComponentBase.java:595)
        at javax.faces.component.UIComponent.encodeAll(UIComponent.java:1654)
        at com.sun.faces.context.PartialViewContextImpl$PhaseAwareVisitCallback.visit(PartialViewContextImpl.java:615)
        at com.sun.faces.component.visit.PartialVisitContext.invokeVisitCallback(PartialVisitContext.java:159)
        at javax.faces.component.UIComponent.visitTree(UIComponent.java:1456)
        at javax.faces.component.UIComponent.visitTree(UIComponent.java:1468)
        at javax.faces.component.UIComponent.visitTree(UIComponent.java:1468)
        at com.sun.faces.context.PartialViewContextImpl.processComponents(PartialViewContextImpl.java:400)
[...]

PR #6723, opened 2020-12-26
#6722 CSP: provide infrastructure to add CSP-support to PrimeFaces-Selenium

### LINKED ISSUE #6722: CSP: provide infrastructure to add CSP-support to PrimeFaces-Selenium
see https://github.com/primefaces-extensions/primefaces-integration-tests/pull/78

christophs78:


src/main/resources/META-INF/resources/primefaces/core/core.csp.js
```diff
@@ -63,7 +69,29 @@ if (!PrimeFaces.csp) {
                 };
 
                 $(element).off(jqEvent).on(jqEvent, jsWrapper);
+
+                //TODO: Depending on some (debug-)switch?
```
melloware: Yes add `if (PrimeFaces.isDevelopmentProjectStage()) {` so this only runs when JSF stage is development.

src/main/resources/META-INF/resources/primefaces/core/core.csp.js
```diff
@@ -63,7 +69,29 @@ if (!PrimeFaces.csp) {
...
+         */
+        getRegisteredEvent: function(id, event) {
+            if (this.registeredEvents.has(id)) {
+                var shortenedEvent = event.substring(2, event.length),
+                    jqEvent = shortenedEvent + '.' + id;
+
+                return this.registeredEvents.get(id).get(jqEvent);
             }
+            return "";
         },
```
melloware, line 103: what about another helper than returns a boolean like `hasRegisteredEvent` which we can use in Selenium to determine whether this should be AJAX or not?

tandraschko:
Im a bit unsure about this topic. Will think about ir

src/main/resources/META-INF/resources/primefaces/core/core.csp.js
christophs78: Are we sure integration-tests based on PrimeFaces Selenium are only executed in JSF development-stage?

src/main/resources/META-INF/resources/primefaces/core/core.csp.js
christophs78: good idea - done

christophs78:
What do you think about replacing
```javascript
/**
 * Map containing registered event listeners per element.
 * @type {Map<string, Map<string, string>>}
 */
registeredEvents: new Map(),
```
with
```javascript
/**
 * Map containing registered event listeners per element.
 * @type {Map<string, Map<string, boolean>>}
 */
hasRegisteredEvents: new Map(),
```
This would reduce our memory-footprint to an absolute minimum.

tandraschko:
Sounds better but PFSelenium needs to check the script content (isAjaxScript)?!

melloware (approved):

melloware:
@tandraschko He is doing that in this PrimeSelenium PR: https://github.com/primefaces-extensions/primefaces-selenium/pull/63

tandraschko:
i dont get how its working if you just have the info "there is a onclick script available" - you also need to search trough the event-script-string actually?

christophs78:
Do we add non-ajax scripts via `PrimeFaces.csp.register`?
When yes we may add some logic like to following to core.csp.js:
`return script.contains("PrimeFaces.ab(") || script.contains("pf.ab(") || script.contains("mojarra.ab(") || script.contains("jsf.ajax.request");`
That´s the logic currently placed in `org.primefaces.extensions.selenium.component.base.ComponentUtils#isAjaxScript`.

christophs78:
Or do we prefer to store the whole script inside the map like we had before?

tandraschko:
AFAIR we do it for every script 
We could do it but -1 for production

christophs78:
When we do it only for JSF dev-stage this means, you must turn your application into dev-stage to use PrimeFaces Selenium together with CSP. We would need to add this to the documentation.

Our own integration tests currently run in production-stage.
```xml
<context-param>
    <param-name>javax.faces.PROJECT_STAGE</param-name>
    <param-value>Production</param-value>
</context-param>
```

To be honest i´m currently not sure what the right trade-offs are.
I understand it´s not ideal to store CSP-JS in a map for production stage.

christophs78:
Updated this PR to track only ajaxified events inside our csp-map.

melloware:
Personally I would do the `Primefaces.isDevelopmentStage()` because in my experience Selenium test sould not be run against a Production system. If you don't know you have a bug before it gets to PRD that is not the time to run Selenium.  Just my two cents.

tandraschko:
I think they should run in stage systemtest

melloware:
OK so we should add new helpers for `PrimeFaces.isNotProductionStage()`

christophs78:
There may be stages like QS or integration which are not covered by JSF stage. In my organization we use JSF development stage only during local development.
SystemTest makes sense. I‘ll do this and we add this to PF Selenium Docu.

melloware (approved):

tandraschko:
Another Idea: why not overwrite the method in PFSelenium?

christophs78:
> Another Idea: why not overwrite the method in PFSelenium?

Good idea! I tried this and currently tun into a timing issue. We have to overwrite PrimeFaces.csp.register after PF-core.js is loaded but before this function is called the first time.
I tried this as part of PF Selenium onload.js. Looks like this comes to late. Some idea´s how to fix this?

```javascript
console.log("check PrimeFaces.csp...");
if (PrimeFaces.csp) {
    console.log("PrimeFaces.csp is active");

    PrimeFaces.csp.REGISTERED_AJAXIFIED_EVENT_LISTENERS = new Map();

    var cspRegisterOrig = PrimeFaces.csp.register;
    PrimeFaces.csp.register = function(id, event, js){
        console.log("PF Selenium - PrimeFaces.csp.register");
        debugger;

        if (event) {
            cspRegisterOrig(id, event, js);

            //Collect some basic information about registered ajaxified event listeners
            if (!PrimeFaces.csp.REGISTERED_AJAXIFIED_EVENT_LISTENERS.has(id)) {
                PrimeFaces.csp.REGISTERED_AJAXIFIED_EVENT_LISTENERS.set(id, new Map())
            }

            var jsAsString = js.toString();
            var ajaxified = (jsAsString.indexOf("PrimeFaces.ab(") >= 0) || (jsAsString.indexOf("pf.ab(") >= 0) || (jsAsString.indexOf("mojarra.ab(") >= 0) || (jsAsString.indexOf("jsf.ajax.request") >= 0);
            PrimeFaces.csp.REGISTERED_AJAXIFIED_EVENT_LISTENERS.get(id).set(jqEvent, ajaxified);
        }
    };

    PrimeFaces.csp.hasRegisteredAjaxifiedEvent = function(id, event) {
        console.log("PF Selenium - PrimeFaces.csp.hasRegisteredAjaxifiedEvent");
        debugger;

        if (PrimeFaces.csp.REGISTERED_AJAXIFIED_EVENT_LISTENERS.has(id)) {
            var shortenedEvent = event.substring(2, event.length),
                jqEvent = shortenedEvent + '.' + id;

            return PrimeFaces.csp.REGISTERED_AJAXIFIED_EVENT_LISTENERS.get(id).get(jqEvent);
        }
        return false;
    };
}
```

Console-debug.output:
grafik

christophs78:
Please have a look at https://github.com/primefaces-extensions/primefaces-selenium/pull/63/files.
Specially at `PrimefacesSeleniumPhaseListener `and `pfselenium.core.csp.js`.

There is only one issue left:
`org.primefaces.csp.CspPartialResponseWriter` renders `sb.append("PrimeFaces.csp.register('");`

`pfselenium.core.csp.js` does following check:
```javascript
var ajaxified = (jsAsString.indexOf("PrimeFaces.ab(") >= 0) || (jsAsString.indexOf("pf.ab(") >= 0) || (jsAsString.indexOf("mojarra.ab(") >= 0) || (jsAsString.indexOf("jsf.ajax.request") >= 0);
```

But `org.primefaces.application.resource.MoveScriptsToBottomResponseWriter#mergeAndMinimizeInlineScripts` does `.replace("PrimeFaces.ab", "pf.ab")`.

So PF sends following JS to the client:
```javascript
var ajaxified = (jsAsString.indexOf("pf.ab(") >= 0) || (jsAsString.indexOf("pf.ab(") >= 0) || (jsAsString.indexOf("mojarra.ab(") >= 0) || (jsAsString.indexOf("jsf.ajax.request") >= 0);
```

And this check does not detect `PrimeFaces.csp.register` written by `org.primefaces.csp.CspPartialResponseWriter`.

We are so close to our target.

tandraschko:
Hmm, i dont understand the "pf.ab" vs "PrimeFaces.ab" problem. Only when MOVE_SCRIPTS_TO_BOTTOM is active, its replaced and then PrimeFaces.ab will never occur?

christophs78:
Integration tests currently set
```xml
<context-param>
	<param-name>primefaces.MOVE_SCRIPTS_TO_BOTTOM</param-name>
</context-param>
```

Our trick should work independent of `primefaces.MOVE_SCRIPTS_TO_BOTTOM`.

Tried it with primefaces.MOVE_SCRIPTS_TO_BOTTOM=false....

This bring´s back our timing-issue:
grafik

grafik
gets executed before
grafik

Maybe because $(function is delayed by jQuery or Browser...

christophs78:
Yep `$(function() { ... });` is a shortcut for `$(document).ready(function() { ... });`
https://stackoverflow.com/questions/7642442/what-does-function-do

melloware:
Yep PF wraps in `$(function() { ... });` when not using Move to Bottom so the scripts are executed after the DOM has finished loading.  With MOVE_TO_BOTTOM the scripts all all moved to the bottom therefore they don't need the function wrapper because by definition they are running after the DOM has loaded.

I had to make this fix a while ago:
https://github.com/primefaces/primefaces/commit/fdfbb292f399968c884e5a9724ae02521203fc2f

To fix: https://github.com/primefaces/primefaces/issues/5455

christophs78:
What a hell of a ride.
Finally got this working without modifiying PrimeFaces. So - from my perspective - we can close this PR.
Had been able to do this in PrimeFaces Selenium. 
Please look at https://github.com/primefaces-extensions/primefaces-selenium/pull/63/files.
I introduced `PrimefacesSeleniumPhaseListener` which adds `PrimeFacesSeleniumSystemEventListener` as `PreRenderViewEvent`. 
`PrimeFacesSeleniumSystemEventListener` adds `PrimeFacesSeleniumDummyComponent` to the component-tree.
And finally `PrimeFacesSeleniumDummyComponent` adds `pfselenium.core.csp.js` after `core.js` was added by PrimeFaces itself.
This works independent of JSF-stage, MOVE_TO_BOTTOM , ...

melloware:
Excellent work I will close this PR and issue!   great job on fixing it all on the PF selenium side. I left you code review comments.

PR #6724, opened 2020-12-26
Fix #6721: B.C. Dates can't use JS constructor with String

### LINKED ISSUE #6721: Timeline: Support B.C. Dates
In the last version, you force that DateTime with timezone should be used for the timeline component. We would like to see the timeline component for past times. In the previous versions, we could see a date which is past like B.C but now we could not. How could we do that? Is there any way to do it?

Feature request:
The timeline component should support the past like B.C.

melloware:
See here why we have to do this: 
 https://stackoverflow.com/questions/25846123/how-to-format-bc-dates-like-700-01-01


tandraschko:
+1 if well tested

melloware:
Integration Test:
https://github.com/primefaces-extensions/primefaces-integration-tests/commit/c4f762aa39b5e947b973e55570ca001386cb8e34

PR #6732, opened 2020-12-29
Fix #6730: Datatable selectionPageOnly attribute

### LINKED ISSUE #6730: Datatable: support true Global Select/Deselect
**Description**
The provided checkbox-based multiple selection model features a global selection checkbox in the table header.

This checkbox only applies to the currently displayed rows of the table.
This will become apparent when paging is activated, so the checkbox example in the Primefaces showcase will not reveal this properly.

This is both counter-intuitive and cumbersome when using with large sets of data.

As an example, imagine having a bulk delete action you want to apply to several hundred rows of data.

A similar problem has also been resolved with the DataExporter component in the past. Here the boolean attribute `pageOnly` lets the user choose between a page or data set oriented approach.

The envisioned solution should likewise add a new boolean attribute to the Datatable component (e.g. `globalSelectPageOnly`). The default should be `true`, to keep backwards compatibility. When set to `false`, all entries of the table should be affected by global selection/deselection.

**Additional context**

It has been recommended in the past, that a solution using a `toggleSelect` event listener can also take care of global selection behavior.

See [https://stackoverflow.com/questions/22965320/select-all-in-datatable-jsf-primefaces](https://stackoverflow.com/questions/22965320/select-all-in-datatable-jsf-primefaces) and [https://stackoverflow.com/questions/35085709/primefaces-datatable-select-all-with-paginator](https://stackoverflow.com/questions/35085709/primefaces-datatable-select-all-with-paginator).

This is basically working, but an irritating glitch is still present.

When individually selecting all visible rows, the global selection checkbox will get activated that despite not all table entries are selected. So to the user it appears that all entries have been selected while only the rows currently displayed page are selected.

This issue has initially been discussed in the PrimeFaces forum here: [https://forum.primefaces.org/viewtopic.php?f=3&t=64788](https://forum.primefaces.org/viewtopic.php?f=3&t=64788)

**Example XHTML**

Taken from the checkbox example in the Primefaces showcase:
```xml
          <p:dataTable id="checkboxDT" var="car" value="#{dtSelectionView.cars6}"
            paginator="true" rows="5" 
            selection="#{dtSelectionView.selectedCars}" rowKey="#{car.id}" style="margin-bottom:0">
              <f:facet name="header">
                  Checkbox
              </f:facet>
              <p:column selectionMode="multiple" style="width:16px;text-align:center"/>
              <p:column headerText="Id">
              </p:column>
              <p:column headerText="Year">
              </p:column>
              <p:column headerText="Brand">
              </p:column>
              <p:column headerText="Color">
              </p:column>
          </p:dataTable>```

melloware:
I sorted the Datatable properties alphabetically because it had so many properties in the enum it was hard to see them.  I feel like its much easier to see now.

### DIFF
--- a/docs/10_0/components/datatable.md
+++ b/docs/10_0/components/datatable.md
@@ -89,6 +89,7 @@ DataTable displays data in tabular format.
 | scrollable                | false              | Boolean          | Makes data scrollable with fixed header.
 | selection                 | null               | Object           | Reference to the selection data.
 | selectionMode             | null               | String           | Enables row selection, valid values are “single" and “multiple".
+| selectionPageOnly         | true               | Boolean          | When using paginator and checkbox selection mode the select all checkbox in header will select the current page if true or all rows if false. Default true.
 | sortMode                  | multiple           | String           | Defines sorting mode, valid values are _single_ and _multiple_.
 | sortBy                    | null               | SortMeta / Collection<SortMeta> | Property to be used for default sorting. Expects a single or a collection of SortMeta.
 | skipChildren              | false              | Boolean          | Ignores processing of children during lifecycle, improves performance if table only has output components.
--- a/src/main/java/org/primefaces/component/datatable/DataTableBase.java
+++ b/src/main/java/org/primefaces/component/datatable/DataTableBase.java
@@ -23,10 +23,10 @@
  */
 package org.primefaces.component.datatable;
 
-import org.primefaces.component.api.*;
-
 import javax.el.MethodExpression;
 import javax.faces.component.behavior.ClientBehaviorHolder;
+
+import org.primefaces.component.api.*;
 
 public abstract class DataTableBase extends UIPageableData implements Widget, RTLAware, ClientBehaviorHolder,
         PrimeClientBehaviorHolder, UITable<DataTableState> {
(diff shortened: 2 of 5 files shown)


melloware:
Tested in IE11, Firefox, Chrome, Edge with both selectionPageOnly="true" and "false.

melloware:
Integration Tests: https://github.com/primefaces-extensions/primefaces-integration-tests/issues/86

stolp:
I just manually checked this with our application and it is working flawlessly with and without lazy mode.

buunsal:
@melloware why you check is paginator true or false? 

`
this.cfg.selectionPageOnly = this.cfg.selectionPageOnly === false ? !this.cfg.paginator : true;
`

I think it should be possible to selectall with livescroll even if the paginator is false. What are you thinking?

melloware:
Good point!  I have never used "LiveScroll" ever so did not know you could do Selection with LiveScroll?

buunsal:
I tried without this paginator if check and working without any issue. I think we can remove this if check

melloware:
OK do you want to remove it or me?

buunsal:
I will remove

PR #6733, opened 2020-12-29
Fix #6478 BreadCrumb: option to only render the last n items

### LINKED ISSUE #6478: BreadCrumb: option to only render the last n items
We have an application where you can select an item from a list, see details of the item, select a child (ManyToOne) and basically recurse until no childs are available anymore. The path is shown in a breadcrumb, which can get very long:

Screen Shot 2020-11-04 at 16 30 12

So I think it would be a good addition to have an attribute to only render the last n items.

And then maybe options to show the root or not and to go up a level, so you can end up with something like

root / .. / item 5 / item 6 / item 7

or

.. / item 5 / item 6 / item 7

or

item 5 / item 6 / item 7

jepsar:
Fix #6478

Showcase: https://github.com/primefaces/primefaces-showcase/pull/255

### DIFF
--- a/src/main/resources/META-INF/primefaces-p.taglib.xml
+++ b/src/main/resources/META-INF/primefaces-p.taglib.xml
@@ -2168,6 +2168,22 @@
             <required>false</required>
             <type>java.lang.Boolean</type>
         </attribute>
+        <attribute>
+            <description>
+                <![CDATA[Number is items to limit the breadcrumb to. Default is null.]]>
+            </description>
+            <name>limit</name>
+            <required>false</required>
+            <type>java.lang.Integer</type>
+        </attribute>
+        <attribute>
+            <description>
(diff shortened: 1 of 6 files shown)


src/main/resources/META-INF/primefaces-p.taglib.xml
```diff
@@ -2168,6 +2168,22 @@
...
+            </description>
+            <name>limit</name>
+            <required>false</required>
+            <type>java.lang.Integer</type>
+        </attribute>
+        <attribute>
+            <description>
+                <![CDATA[Boolean flag indicating whether the root should be permanently visible when a limit is set. Default is true.]]>
+            </description>
+            <name>permaRoot</name>
```
melloware, line 2183: Maybe `alwaysShowRoot` instead of `permaRoot`?

jepsar:
Converted to draft as this needs work after https://github.com/primefaces/primefaces/pull/6347 has been merged

tandraschko:
see Merts comment in the issue

PR #6738, opened 2020-12-30
Fix #6737: SortMeta sortPriority from Integer to primitive

### LINKED ISSUE #6737: DataTable: Single Sort should not have to declare priority
**Describe the defect**
While fixing the Showcase demo for SummaryRow it had a single sort declared but no `sortPriority` declared and it was not sorting the table by default.  But as soon as I added `sortPriority=1` to the column it then sorted by default correctly.

The issue is in SortMeta if you pass `NULL` to `priority(NULL)` it sets it to NULL instead of its default `MIN_VALUE`.

**Failed:**
```xml
<p:column field="category" headerText="Category" sortOrder="asc" />
```

**Works:**
```xml
<p:column field="category" headerText="Category" sortOrder="asc" sortPriorty="1" />
```

**Environment:**
 - PF Version: _9.0_

**Expected behavior**
Sorting should default even if no `sortPriority` is provided.

**Example XHTML**
```html
<p:dataTable var="product" value="#{dtSummaryRowView.products}" sortMode="single">
            <p:column field="code" headerText="Code"/>

            <p:column field="name" headerText="Name"/>

            <p:column field="category" headerText="Category" sortOrder="asc" />

            <p:column field="quantity" headerText="Quantity"/>
            <p:summaryRow>
                <p:column colspan="3" style="text-align:right">
                    <h:outputText value="Total:"/>
                </p:column>
                <p:column>
                    <h:outputText value="#{dtSummaryRowView.randomPrice}">
                        <f:convertNumber type="currency" currencySymbol="$"/>
                    </h:outputText>
                </p:column>
            </p:summaryRow>
        </p:dataTable>
```

PR coming

melloware:


tandraschko:
not sure i like the fix, why isnt it sorted? null priority seems to be ok IMO.
Also unsure we use int MIN_VALUE instead of 0?

@Rapster cc

melloware:
MIN_VALUE actually is INTEGER.MAX_VALUE making it the highest value.  If there is a NULL somehow the sorting is not working. To me a SORT PRIOIRTY should never be NULL nor was that @Rapster intention with defaulting the variable..
```java
 private Integer priority = MIN_PRIORITY;
```
however nothing stops someone from passing in a NULL in which it should return to default.

melloware:
I have found multiple places in the code setting sortPriority but there is no gaurantee the value is not NULL like...
```java
SortMeta tsSortMeta = entry.getValue();
intlSortBy.setPriority(tsSortMeta.getPriority());
intlSortBy.setOrder(tsSortMeta.getOrder());
```

tandraschko:
why should it be never null? null means undefined and wont be serialized to session e.g. where as a "default int" would be serialized.
nevertheless, if it should not be nullable, also no Integer should be used, just int.

Rapster:
What about using an int instead?

melloware:
Well if its NULL when you are doing Streams.sort how do you want NULL treated as the highest value or lowest value?  And yes `int` would make more sense.

I can try switching it to int but then the default will be `0` right if not set at all which will make it the HIGHEST sorted value not the lowest which to me is also wrong?

tandraschko:
whats wrong with 0 as default anyway? everything starts with 0 :D

melloware:
OK let me give it a shot switching it to `int`

Rapster:
No, priority will remain MIN_PRIORITY as it is his default value

tandraschko:
and why? why is min/max switched? thats confusing

melloware:
OK switched it to `int` and confirmed Showcase examples are all working and it fixed the SummaryRow example of a column with no sort priority.

melloware:
The reason MIN_VALUE =212884660707 highest Integer value is if you do this..

```java
sortBy = new ArrayList<>();
        sortBy.add(SortMeta.builder()
                .field("name")
                .order(SortOrder.ASCENDING)
                .build());

        sortBy.add(SortMeta.builder()
                .field("category")
                .order(SortOrder.ASCENDING)
                .priority(1)
                .build());
```

The meta without a declared priority becomes the highest value so in this case it will be higher than priority 1. If we make it 0 then in the example above "name" would be sorted first instead of "category" since they did not declare a priority.

tandraschko:
ok, i see; then we have to clarify whats correct and adjust other prios and docs; lower number/higher prio seems to be more common: https://ux.stackexchange.com/questions/73445/is-a-higher-priority-a-smaller-number
i implemented display-priority the other way around

melloware:
I am open to changing it I think we would just have to change this one line.
```java
private int priority = MIN_PRIORITY;
```
If we were to decide that unassigned priorities become `0` which I am OK with for consistency.

tandraschko:
Im not saying that we should change sort order but maybe display order. both are ok for me but we should be consistent.
maybe we can collect "all prios" in all PF and check how they are implemented and make some tickets.

Rapster:
I can understand it sounds confusing but it's as much confusing the other way around, it all depends how you understand/speak of priorities. I don't think there is a good solution for this. For example, if I say "it's your top 1 priority to fix this issue", you understand that's your first priority is to fix an issue than "it's your top 212884660707 priority to fix this issue" (personnally I understand it as "it can wait, not a biggy")

The right solution is just to make priorities ordering consistent in all PF

melloware:
Agreed. @tandraschko I think we should align displayPriority and sortPriority since these are "new" to PF 10 before we release and then change them on people later?  Responsive Priority has been around for a while so I am hesitant to touch that?

tandraschko:
yep, i searched over the docs about `priority` and we only have `sort`, `responsive` and `display` in column/columns.
`responsive` is implemented like `sort` (lower number, higher prio), so lets adjust `display` to match that

just still a bit unsure about making it not-nullable. we could also implement a custom-sorter which treats `null` as `MIN_PRIORTY`

tandraschko:
wait, display priority is already correct :D i adjusted the docs now.

melloware:
OK I reverted it back to `Integer` and allow NULLS the only thing that really needed to be fixed was the `Comparator` to prevent NPE when comparing the two values!

Rapster:
I think you should keep int type, it's my mistake if I left Integer for sortPriority. If you look everywhere else, it's a int and not Integer

melloware:
HA OK i will put back.

melloware:
OK fixed back to `int`

Rapster:
> the only thing that really needed to be fixed was the Comparator to prevent NPE when comparing the two values!

Plus, from the moment you introduce NULL values, you have to deal with it everywhere not only comparators (and for primitive types it's not worth it)

tandraschko:
totally ok for me

PR #6744, opened 2020-12-31
Primefaces X: add new components

### LINKED ISSUE #6712: New Component: Skeleton from PrimeVue
Hi, 

It would be very awesome, if you can implement also the Skeleton (from PrimeVue) in Primefaces:
https://primefaces.org/primevue/showcase/#/skeleton

I think the requirement should be clear.

buunsal:
New components are: Avatar, Badge, Chip, Divider, ScrollTop, Skeleton, Splitter, Tag

### DIFF
--- a/docs/9_0/components/avatar.md
+++ b/docs/9_0/components/avatar.md
@@ -1,0 +1,103 @@
+# Avatar
+
+Avatar represents people using icons, labels and images.
+
+## Info
+
+| Name | Value |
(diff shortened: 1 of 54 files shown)


melloware:
@squores I know this is a work in progress but I am going to submit a PR for your review that makes some small changes to do things the way we do in other components.  Excellent Job on this!!!

PR #6747, opened 2020-12-31
PrimeX use StyleClassBuilder

melloware:
Discussed with @sqores on Discord this cleans up all the String concat += stuff and makes the code much cleaner

### DIFF
--- a/src/main/java/org/primefaces/component/avatar/AvatarRenderer.java
+++ b/src/main/java/org/primefaces/component/avatar/AvatarRenderer.java
@@ -28,6 +28,7 @@ import org.primefaces.renderkit.CoreRenderer;
 import javax.faces.component.UIComponent;
 import javax.faces.context.FacesContext;
 import javax.faces.context.ResponseWriter;
+
 import java.io.IOException;
 
 public class AvatarRenderer extends CoreRenderer {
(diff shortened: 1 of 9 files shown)


buunsal (approved):

tandraschko:
why Objects.equals(avatar.getShape(), "circle")?
"circle".equals(avater.getShape()) is shorter and widely used as pattern

melloware:
I use ObjectEquals because it prevents NULL checking.  A lot of places if you look in the review there was `if != null and if shape.equals("circle") etc.  This allows the boolean to always evaluate with never having to do NULL checking or assume one side is not null.

but you are right I could do "circle.equals() but for some reason Objects.equals looks better to me.

I can switch them all if you want?

tandraschko:
yeah, `avater.getShape().equals("circle")` would a require a null check but `"circle".equals(avater.getShape())` does not.
in the past there were checkstyle/PMD rules for this: https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/checks/coding/EqualsAvoidNullCheck.html

i would use this pattern everywhere and try to introduce this rule; for other types only Object.equals is possible ofc

melloware:
OK I flipped them all not to use `Objects.equals`

PR #6748, opened 2020-12-31
CheckStyle EqualsAvoidNullCheck

melloware:
294 paces this violation occurred after I turned on the CheckStyle rule

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -490,7 +490,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-checkstyle-plugin</artifactId>
-                <version>3.1.0</version>
+                <version>3.1.1</version>
                 <configuration>
                     <consoleOutput>true</consoleOutput>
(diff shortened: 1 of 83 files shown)


melloware:
If this is too big we can close this PR and I can do it again after 10.0 goes live.

I have also switched my Imports to match @Rapster and @sqores  which is...

org
javax
java
com

any imports with more than 5 get .*.

tandraschko:
It would be much easier to turn off auto-reorder of imports.
It doesnt make any code better and just harder to review.

tandraschko:
We should either
1) add IDE configs for codestyle in git and add checkstyle rules for import order
Or
2) add a readme that committers should disable auto format stuff. 

 
2. Is IMO much better...

melloware:
I like the idea of coming up with what our standard is. Then adding check style rules to enforce it!

tandraschko:
i just can say that i dont like that because to strict checks make it much harder to contribute and disable auto-formatting of the IDEs takes the most pain away.

melloware:
Understood let me close this one and submit another with the bare minimum changes.

PR #6758, opened 2021-01-03
add JSF 3.0 to docu

christophs78:
maybe we should also add some note regarding primefaces-10.0-SNAPSHOT-jakarta.jar


tandraschko:
we should add somewhere that the jakarta dependency classifier is required

melloware:
We could do it like I did in the PF Extensions README.MD?

See: https://github.com/primefaces-extensions/primefaces-extensions

PR #6767, opened 2021-01-04
Fix #6766: Rating widget API validations

### LINKED ISSUE #6766: Rating: Widget method validations
**Describe the defect**
While writing Integration Tests for rating I found some validation issues on the client side widget methods. We currently block being above the min/max on the server side with ticket: https://github.com/primefaces/primefaces/issues/4203

However I found the following scenarios were failing:
- Calling `setValue` on a disabled rating allowed you to set the rating
- Calling `setValue` on a readonly rating allowed you to set the rating
- Calling  `setValue` with a value below 0 took the value like -1 instead of resetting
- Calling  `setValue` with a value above max stars took the value like 14 instead of capping at the max

**Reproducer**
See PrimeFaces Integration Tests

image

melloware:
With these fixes all Integration Tests pass now:

image

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/rating/rating.js
+++ b/src/main/resources/META-INF/resources/primefaces/rating/rating.js
@@ -50,12 +50,12 @@ PrimeFaces.widget.Rating = PrimeFaces.widget.BaseWidget.extend({
      * @private
      */
     bindEvents: function() {
-        var _self = this;
+        var $this = this;
 
         this.stars.on("click", function() {
-            var value = _self.stars.index(this) + 1;   //index starts from zero
+            var value = $this.stars.index(this) + 1;   //index starts from zero
 
-            _self.setValue(value);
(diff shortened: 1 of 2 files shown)


src/main/resources/META-INF/resources/primefaces/rating/rating.js
```diff
@@ -92,18 +92,32 @@ PrimeFaces.widget.Rating = PrimeFaces.widget.BaseWidget.extend({
      * @param {number} value New rating value to set (number of starts selected).
      */
     setValue: function(value) {
+        if(this.jq.hasClass('ui-state-disabled') || this.jqInput.is('[readonly]')) {
```
tandraschko: maybe add a isDisabled and isReadonly method for it?

melloware: Great idea. Added!

PR #6775, opened 2021-01-06
Fix #6774: Localize emptyMessage

### LINKED ISSUE #6774: DataTable: Add emptyMessage localization support
**Description**
I always need to set a `p:datatable`s `emptyMessage` to a localized string. Often times I just forget that though. The text "`No records found.`" seems to be hard coded into 
`src/main/java/org/primefaces/component/datalist/DataListBase.java`,  
`src/main/java/org/primefaces/component/datagrid/DataGridBase.java` and  
`src/main/java/org/primefaces/component/treetable/TreeTableBase.java`.

**Solution**
This is something I would really like to see added to something like the [PrimeFaces Locales Javascript API](https://github.com/primefaces/primefaces/wiki/Locales).

melloware:
Refactored to move emptyMessage to UIPageableData as this reduces some cut and paste in 3 of the components.  TreeTable still extends Tree not UIPageable so for that I just made the change in Tree.


melloware:
I sorted the Messages.properties alphabetically because the file was a mess to look at.

melloware:
I need to update the Localization docs.

tandraschko:
thanks!

PR #6782, opened 2021-01-07
Fix #6776: ContextMenu IOS fixes

### LINKED ISSUE #6776: ContextMenu: Datatable long touch is not working on IOS
Datatable Context Menu is not working on Iphone IOS (Safari and Chrome). On Android, touch & hold opens the Contextmenu, but it is not working on Iphone (both Safari and Chrome).

**Reproducer**
The example shown in PF Showcase is not working on IOS too.
https://www.primefaces.org/showcase/ui/data/datatable/contextMenu.xhtml

**Environment:**
 - PF Version: _8.0.6_
 - JSF 2.2
 - Affected browsers: iOS

**To Reproduce**
1. Go to 'https://www.primefaces.org/showcase/ui/data/datatable/contextMenu.xhtml' with iOS device 
2. Touch & Hold on row to see context menu
3. Text selection mode is started instead of showing context menu

**Expected behavior**
Datatable context menu should be shown with Touch & Hold action on iOS.

**Example XHTML**
All source coding is exist at page => https://www.primefaces.org/showcase/ui/data/datatable/contextMenu.xhtml

melloware:
OK there were multiple issues with IOS.  to repeat everything is working in Android but IOS is much different.

1. Using this CSS fix to disable IOS from bringing up the "Copy/Paste/Share" menu on long tap.
https://stackoverflow.com/a/11237968/502366

2. IOS needed to load the TouchSwipe longtap for Datatable, TreeTable, etc that use Contextmenu. Android does not because it interprets Long tap as a "rght click" so our normal Context menu stuff is all working.

Tested in Chrome Desktop, Chrome Mobile, and IOS 14 mobile.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/menu/menu.contextmenu.js
+++ b/src/main/resources/META-INF/resources/primefaces/menu/menu.contextmenu.js
@@ -100,6 +100,10 @@ PrimeFaces.widget.ContextMenu = PrimeFaces.widget.TieredMenu.extend({
                 if (targetWidget) {
                     if (typeof targetWidget.bindContextMenu === 'function') {
                         targetWidget.bindContextMenu(this, targetWidget, this.jqTargetId, this.cfg);
+                        // GitHub #6776 IOS needs long touch on table/tree but Android does not
+                        if(PrimeFaces.env.ios) {
+                            $this.bindTouchEvents();
+                        }
                         binded = true;
                     }
                 }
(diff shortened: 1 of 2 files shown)


melloware:
Not sure I love the CSS name `ui-touch-selection-disabled` but I am open to suggestions.

src/main/resources/META-INF/resources/primefaces/menu/menu.contextmenu.js
```diff
@@ -100,6 +100,9 @@ PrimeFaces.widget.ContextMenu = PrimeFaces.widget.TieredMenu.extend({
                 if (targetWidget) {
                     if (typeof targetWidget.bindContextMenu === 'function') {
                         targetWidget.bindContextMenu(this, targetWidget, this.jqTargetId, this.cfg);
+                        if(PrimeFaces.env.ios) {
```
tandraschko, line 104: please add a comment here, whiy we only bind on IOS in this case

PR #6780, opened 2021-01-07
Fix #6700: Safari AJAX fix

melloware:
@squores if it is necessary to have the scrollTop fix and the native JS `replaceWith` does not work this is what I envision the Safari specific fix to look like

PR #6783, opened 2021-01-07
Fix #6700: Core: After updating a component using AJAX in Safari, the current scroll position changes

### LINKED ISSUE #6700: Core: After updating a component using AJAX in Safari, the current scroll position changes
Test code;
```
 
<h:head>
    <title>Test</title>
</h:head>
<h:body>
    <h:form>
	    <div style="height: 1000px">
	    	<div style="height:400px"></div>   
	    	<div style="height:1800px"> 
		        <h:inputText id="name" value="#{testBean.test}"></h:inputText>
		        <p:selectOneMenu>
		        	<f:selectItems value="#{testBean.values}" />
				<p:ajax listener="#{testBean.test()}" update="@form" />        
			</p:selectOneMenu>
		</div>
	    </div>
    </h:form>
</h:body>
</html>
```

christophs78:
@sqores, @mertsincan, @tandraschko, @melloware 
My plan "B". We would need to add some if-statement to do this only for Safari and/or browser which support https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/replaceWith.
Not totally sure whether it´s a good idea without side-effects.
And first we should look whether this improves Safari-behaviour.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
+++ b/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
@@ -251,7 +251,9 @@ if (!PrimeFaces.ajax) {
                     PrimeFaces.ajax.Utils.updateHead(content);
                 }
                 else {
-                    $(PrimeFaces.escapeClientId(id)).replaceWith(content);
+                    var element = document.getElementById(id);
+                    var fragment = document.createRange().createContextualFragment(content);
+                    element.replaceWith(fragment);
                 }
             }
         },


melloware:
You also have to wrap the whole thing in `PrimeFaces.env.isIE` and do the normal current code for IE since there is no DOM `replaceWith` for IE.  also we need to check if it removes any registered events right?

tandraschko:
if this is really working, we must only do this for Safari and we need to call jQuery methods probably: https://github.com/jquery/jquery/blob/master/src/manipulation.js#L414

melloware:
Something like my fix here: https://github.com/primefaces/primefaces/pull/6780/files

Except replace the guts inside `isSafari`

PR #6784, opened 2021-01-08
add card component

buunsal:


src/main/java/org/primefaces/component/card/CardRenderer.java
```diff
@@ -0,0 +1,114 @@
...
+        writer.endElement("div");
+
+        //BODY START
+        writer.startElement("div", card);
+        writer.writeAttribute("class", Card.BODY_CLASS, null);
+
+        //title
+        writer.startElement("div", card);
+        writer.writeAttribute("class", Card.TITLE_CLASS, null);
+        if (ComponentUtils.shouldRenderFacet(titleFacet)) {
```
melloware, line 72: Not positive but should we render these divs at all if `shouldRenderFacet` is false.  Looks like we are always rendering empty DIVS for title, header, footer etc?

PR #6797, opened 2021-01-11
Fix #6796: Skip setTimeout if Menubar delay is 0

### LINKED ISSUE #6796: Menubar: Inactive submenu visible
**Describe the defect**
There are times when mousing over the menubar where a submenu will become visible even though another submenu is active and open. The erroneous submenu will be visible, however the submenu heading is not highlighted. The erroneous submenu will remain open until the user interacts with that specific submenu again. The user can interact with other submenus, while the erroneous submenu is open. I've attached a screenshot depicting the issue. Through debugging I've found that it's related to the addition of the delay attribute, and calling setTimeout called in PrimeFaces.widget.Menubar.prototype.showSubmenu. Since setTimeout places a message on the queue for the browser to process once there is no script processing, it's running into a situation where the submenu is no longer the active submenu by the time it 'shows' the submenu.

PrimeFaces - menubar - stuck submenu reproducer - screenshot

**Reproducer**
Reproducer attached
[primefaces-test-menubar-stuck_submenu.zip](https://github.com/primefaces/primefaces/files/5797482/primefaces-test-menubar-stuck_submenu.zip)

**Environment:**
 - PF Version: _8.0_
 - JSF + version: _Mojarra 2.2.20_
 - Affected browsers: _Tested on Chrome, Brave and Firefox_

**To Reproduce**
It can be a little difficult at times to recreate the issue, generally it occurs as a user is mousing quickly across the menubar from right to left and vice versa. Sometimes using a circular type motion from left to right.
1. Mouse over the menubar from left to right and vice versa, and even where the mousing is not necessarily linear.
4. See submenu open that is not 'active' along with other submenus still showing and hiding as expected.

**Expected behavior**
Submenus in the menubar should not display if they aren't the active submenu.

**Example XHTML**
See attached reproducer

**Example Bean**
N/A

JHookDev:
To fix issue 6796 I am proposing that we skip using setTimeout if the Menubar delay is 0. The submenu show will execute immediately rather than being submitted to the browser message queue for processing.


src/main/resources/META-INF/resources/primefaces/menu/menu.menubar.js
```diff
@@ -43,11 +43,18 @@ PrimeFaces.widget.Menubar = PrimeFaces.widget.TieredMenu.extend({
             clearTimeout(this.timeoutId);
         }
 
-        this.timeoutId = setTimeout(function () {
-           submenu.css('z-index', PrimeFaces.nextZindex())
-                  .show()
-                  .position(pos)
-        }, this.cfg.delay);
+        //avoid using timeout if delay is 0
+        if(this.cfg.delay != null && this.cfg.delay > 0) {
```
melloware: You can change this to...
```javascript
if(this.cfg.delay && this.cfg.delay > 0) {
```

This will be false if delay ==0 or delay == null or delay == undefined.

melloware (changes requested):

melloware (approved):

src/main/resources/META-INF/resources/primefaces/menu/menu.menubar.js
JHookDev: Thanks for the feedback. I've updated the change accordingly.

PR #6800, opened 2021-01-12
Datatable - allow selection through different groups

### LINKED ISSUE #6799: Datatable: Grouped keyboard selection
**Describe the defect**
Grouped Data Table key up/down only select the prev/next row within the current group. I'm not sure if it was itentional or not

**Reproducer**
Data Table with group and active selection

**Environment:**
 - PF Version: _8.0_
 - JSF + version: _e.g. Mojarra 2.2.8 or MyFaces 2.3.6_

**To Reproduce**
1. Select a row
2. Press up (or down) on the keyboard
3. If the first (or last) row of a group is selected we cannot navigate further

**Expected behavior**
One should be able to select the row in the prev/next group

crocimi:
navigate to nextUntil or prevUntil using key down/up

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
+++ b/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
@@ -867,7 +867,7 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
                     case keyCode.UP:
                     case keyCode.DOWN:
                         var rowSelector = 'tr.ui-widget-content.ui-datatable-selectable',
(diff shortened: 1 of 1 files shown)


tandraschko:
change looks good to me, can you retest @melloware ?

melloware (approved):

melloware:
Tested and this is working for me.  I like this solution

melloware:
Actually @Mich2424 did you test this in non group scenario.  Won't this go all the way to the 1st row or last row in a non grouped scenario every time?

melloware:
OK I just tested it myself in normal non grouped mode and its working as expected.  I think we should merge this fix.

crocimi:
I have tested it also in a non-group scenario, I think it is ok.  nextAll = get all **following** siblings, eq(0)=take the first one

PR #6801, opened 2021-01-12
Fix #6802: Core: Remove redundant initializer

### LINKED ISSUE #6802: Core: Remove redundant initializer
Java will initialize fields with known default values so any explicit initialization of those same defaults is redundant and results in a larger class file (approximately three additional bytecode instructions per field).

arturobernalg:
* Remove redundant initializer

### DIFF
--- a/docs/10_0/components/organigram.md
+++ b/docs/10_0/components/organigram.md
@@ -41,10 +41,10 @@ node at the backend needs to match the p:organigramNode helper component.
 public class OrganigramView implements Serializable {
     private OrganigramNode rootNode; 
     private OrganigramNode selection;
-    private boolean zoom = false; 
+    private boolean zoom; 
     private String style = "width: 800px";
-    private int leafNodeConnectorHeight = 0; 
-    private boolean autoScrollToSelection = false;
+    private int leafNodeConnectorHeight; 
+    private boolean autoScrollToSelection;
     private String employeeName;
 
     @PostConstruct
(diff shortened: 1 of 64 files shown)


Rapster:
First create an issue, follow template, explain what's on your mind

tandraschko:
I dont think this brings more performance, i hope the compiler is smart enough to create the same code in both ways ;)

but i think the PR is ok for codestyle / more consistent.

Rapster:
I was planning to change the title, as I don't think it's about performance here but more about code consistency. Issue still need to be created to fully understand what's on your mind (so far we're guessing)

arturobernalg:
Hi @Rapster  @tandraschko 

Agree with change the title. Is basically minor improvement.  The  idea it's avoid explicit initialization

melloware (approved):

melloware:
I am OK with this too.

stolp:
Am I mistaken, or does this patch introduce additional whitespace before the semicolons?

    private boolean zoom = false; 
    private boolean zoom  ;
                        ^^

Rapster:
Well spotted, only in md files though. I quite don't mind boolean initiliazed with false, as it shows explicitely the intention, but that's my personal taste , also IntelliJ doesn't seem to complain about it (there is probably a reason to it)

arturobernalg:
> Am I mistaken, or does this patch introduce additional whitespace before the semicolons?
> 
> ```
> private boolean zoom = false; 
> private boolean zoom  ;
>                     ^^
> ```

Mi mistake. fixed

PR #6805, opened 2021-01-12
Fix #6804: Calendar alignPanel increase delay before rendering

### LINKED ISSUE #6804: Calendar: Opening off screen in Premium themes
The Calendar component is opening off screen when positioned close to the right side of the page

Example:
image

Reproducer: https://github.com/thecatloko/primefaces-test

melloware:
1ms was not enough before rendering for the Calendar panel to properly use `flipfit` logic from Jquery.  Increasing this to 50ms provides no noticeable UI delay but does cause the popup panel to do collision detection correctly.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/calendar/calendar.js
+++ b/src/main/resources/META-INF/resources/primefaces/calendar/calendar.js
@@ -168,7 +168,7 @@ PrimeFaces.widget.Calendar = PrimeFaces.widget.BaseWidget.extend({
                     }
                     
                     $this.alignPanel();
-                }, 1);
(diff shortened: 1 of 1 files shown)


tandraschko:
if we use less timeout, it might not work for sure?

melloware:
Positive with his reproducer.  At 1ms the panel is drawing off the edge of the browser.  BY increasing the timeout the reproducer then runs and correctly positions the calendar.  I can try and tweak the 50ms to be lower but honestly 50ms is not noticeable by the naked eye I can't tell at all.

But this definitely fixes the issue.

PR #6806, opened 2021-01-12
Fix #6798: Schedule allow background events

### LINKED ISSUE #6798: Schedule: Background Events
**Description**
To prevent date selection(dateSelect) in a given date interval I've to close the pop-up right after it's opened. This is frustrating because the user has to click in order to notice that reservation at selected date is not possible.

I would like an additional attribute in the scheduler which accepts start and end date/time and the cells in this time interval are visually distinguishable(color as another attribute). FullCalendar apparently does that through [Background Events](https://fullcalendar.io/docs/background-events)

**Additional context**
Ideally opening of calendar should be optional. 

**Example XHTML**
```xml
          <p:schedule
            id="scheduleGrid"
            backgroundEventStartDate="#{bean.bgStartDate}"
            backgroundEventEndDate="#{bean.bgEndDate}"
            backgroundEventColor="#{bean.bgColor}"
            backgroundEventPopUp="true"
            widgetVar="scheduleWidgetVar"
            value="#{bean.lazyEventModel}">
          </p:schedule>
```

Thanks in advance

melloware:
From the FullSchedule docs added a few more properties to allow background events.

Usage:
```java
event = DefaultScheduleEvent.builder()
				.title("")
				.startDate(LocalDateTime.now().minusDays(4))
				.endDate(LocalDateTime.now().minusDays(3))
				.overlapAllowed(true)
				.editable(false)
				.resizable(false)
				.display("background")
				.backgroundColor("lightgreen")
				.build();
eventModel.addEvent(event);
```

Screenshot:
image

### DIFF
--- a/src/main/java/org/primefaces/component/schedule/ScheduleRenderer.java
+++ b/src/main/java/org/primefaces/component/schedule/ScheduleRenderer.java
@@ -122,9 +122,27 @@ public class ScheduleRenderer extends CoreRenderer {
                     jsonObject.put("durationEditable", event.isResizable());
                 }
                 jsonObject.put("overlap", event.isOverlapAllowed());
-                jsonObject.put("classNames", event.getStyleClass());
-                jsonObject.put("description", event.getDescription());
-                jsonObject.put("url", event.getUrl());
+                if (event.getStyleClass() != null) {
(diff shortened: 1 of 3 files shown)


tandraschko:
looks good to me

PR #6825, opened 2021-01-15
Fix #6824: Add chronoline component

### LINKED ISSUE #6824: New Component: Chronoline
Chronoline visualizes a series of chained events.

buunsal:


src/main/java/org/primefaces/component/chronoline/ChronolineRenderer.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    protected void encodeSeparator(ResponseWriter writer, FacesContext context, UIComponent markerFacet, boolean isLastItem) throws IOException {
+        writer.startElement("div", null);
+        writer.writeAttribute("class", Chronoline.EVENT_SEPARATOR_CLASS, null);
+
+        if (ComponentUtils.shouldRenderFacet(markerFacet)) {
+            markerFacet.encodeAll(context);
+        }
+
```
melloware, line 122: Extra spacer line here not needed between if and else

melloware (changes requested):

src/main/java/org/primefaces/component/chronoline/ChronolineRenderer.java
buunsal: I agree with your opinion. Even "if else" is up to me, it should be written like this 😄 

```
if (ComponentUtils.shouldRenderFacet(oppositeFacet)) {
   oppositeFacet.encodeAll(context);
} else {
   writer.write("&nbsp;");
}
```

In fact, my ide do this automatically. However, I have to edit the code and add extra blanks every time to avoid getting a maven checkstyle error 😞

PR #6827, opened 2021-01-15
Fix #6561: Menus fix ARIA support

### LINKED ISSUE #6561: Menubar, Menu, MenuButton: not accessible
**Describe the defect**
p:menubar (p:menu, p:meneButton as well) is not accessable. That is, when tabbing through the menu using a screen reader like JAWS, it does not read any of the menu items, even all the generated elements have titles or aria-lables.

**Reproducer**
See here: https://www.primefaces.org/showcase/ui/menu/menubar.xhtml

**To Reproduce**
Open a screen reader (JAWS for example), tab through "File" "Edit" menus and submenus, the screen reader will read nothing.

**Expected behavior**
It would be nice to have an unified focus and screen reader interaction behaviour, like p:tabMenu, which works perfectly.

**Example XHTML**
It seems that the focus is always landed on the surrounding elements like <div ... ></div>, or <li … />, while the accessable values are inside the inner elements. However, simply adding a title or aria-label to <li … /> does not work either.

melloware:
- Adds `aria-label` to submenu `ul` lists as required by the spec
- Adds `tabindex=0` to the focused menu item all others have `-1` see Roving Tabindex https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_roving_tabindex
- Adds `aria-expanded` property to submenus and toggles it `true` and `false`

This fixes MenuBar, TieredMenu, Menu, and MenuButton.

### DIFF
--- a/src/main/java/org/primefaces/component/menu/BaseMenuRenderer.java
+++ b/src/main/java/org/primefaces/component/menu/BaseMenuRenderer.java
@@ -176,6 +176,7 @@ public abstract class BaseMenuRenderer extends MenuItemAwareRenderer {
         writer.startElement("div", null);
         writer.writeAttribute("tabindex", menu.getTabindex(), null);
         writer.writeAttribute("class", "ui-helper-hidden-accessible", null);
+        writer.writeAttribute(HTML.ARIA_ROLE, HTML.ARIA_ROLE_NONE, null);
         writer.endElement("div");
     }
 }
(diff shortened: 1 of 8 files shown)


melloware:
Working with @serial389 to test this and make sure it actually works all the way.

tandraschko:
what should we do with this PR for now? close?

PR #6842, opened 2021-01-17
Fix #6841 - Core: remove deprecates introduced in 8.0

### LINKED ISSUE #6841: Core: remove deprecates introduced in 8.0
Deprecated but still used:
- ~CalendarUtils#calculateTimeZone~
- `TimeLineModel#getIndex` (not sure what I'm supposed to do with this...)
- `p:chart` untouched as the whole component will be deprecated

Rapster:


melloware (approved):

melloware (approved):

melloware (approved):
This looks good.

Rapster:
Should I merge it then?

melloware:
I feel like we should WDYT @tandraschko

melloware:
Or should we wait until right after RC1 that way anyone using RC1 straight for 6.2 or 7.0 will see the deprecations but they will be removed before 10.0.0 Final?  I don't really care either way but just food for thought...

tandraschko:
I will review tomorrow

tandraschko:
looks good for me, i just cant speak for the timeLine-Id methods.
Thats up to the timeline contributors, but we dont have statements here AFAICS?

Rapster:
Yeah I had a quick try, and it seems fine. If you look at `TimeLineModel#getIndex` (called by `TimeLineModel#update`) all it does is a `events.indexOf` and set the very same index then, so I think it's safe to remove this. For RC1, I would like to have this merged so users spend their time dealing with breaking changes only once, and not on several RC releases

melloware:
I say go ahead and merge it

tandraschko:
I'm really unsure about the index; we didnt deprecate it in the migration guide in the past.

melloware:
We should have. If I recall the index was used for the old Almeda Timeline used before 8.0.  When we switched to the modern vis-timeline version it was deprecated because i am pretty sure its not used.

Rapster:
Original issue: #3245 merged and deprecated in 8.0

tandraschko:
i currently checked my customers applications and its used there. I need to check the issue and docs.

tandraschko:
yep, i see the reason now. Still supporting index would just be a hack.

PR #6846, opened 2021-01-18
Fix #6711: Autocomplete change event with CSP

### LINKED ISSUE #6711: AutoComplete: change event not triggered
**Describe the defect**
Autocomplete does no longer trigger the change ajax event. Though in the same setup inputText does trigger change ajax event. Using Primefaces 8.0, autocomplete triggers change ajax event.

Autocomplete still triggers itemSelect if using a completeMethod, and also triggers blur event. It's just change event that is effected. I guess the problem is caused by https://github.com/primefaces/primefaces/commit/6876cfc31ac18229a228f6f61924d9b852fc4224

The defect only occurs when CSP is enabled, by adding:

`<context-param>
</context-param>`

to web.xml

**Environment:**
 - PF Version: _8.0.5_
 - JSF + version : _Jakarta Faces 2.3.14_
 - Affected browsers: _ALL_

**To Reproduce**
1. Change value in inputText and tab, label is updated
2. Change value in autoComplete and tab, label is not updated

**Expected behavior**
Typing text in an autocomplete and focusing other component should trigger a change event. 

**Example XHTML**
```html
<h:form>

	<p:inputText value="#{autocompleteTestBean.value}">
		<p:ajax event="change" update="val"/>
	</p:inputText>

	<p:autoComplete value="#{autocompleteTestBean.value}">
		<p:ajax event="change" update="val"/>
	</p:autoComplete>
	
	<h:outputLabel id="val" value="#{autocompleteTestBean.value}"/>

</h:form>
```

**Example Bean**
```java
@Component
@Scope(value = "view")
public class AutocompleteTestBean extends ControllerBean implements Serializable {

	@Getter
	@Setter
	private String value;
}
```

melloware:
This is exactly like the fix I made for InputNumber: https://github.com/primefaces/primefaces/issues/6142

Basically CSP has already stripped off the "onchange" event from the Input so we need to inspect Jquery events to see if change is registered when using CSP.

I have scanned the codebase and it looks like only InputNumber and AutoComplete do this so there should be no where else this is happening.

@christophs78 @tandraschko 

I have tested in CSP=true and CSP=false mode with the reproducer and everything is working properly again,

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
+++ b/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
@@ -110,9 +110,6 @@ PrimeFaces.widget.AutoComplete = PrimeFaces.widget.BaseWidget.extend({
         this.touchToDropdownButton = false;
         this.isTabPressed = false;
         this.isDynamicLoaded = false;
-        
-        this.cfg.onChange = this.input.prop('onchange');
-        this.input.prop('onchange', null).off('change');
 
         if(this.cfg.cache) {
             this.initCache();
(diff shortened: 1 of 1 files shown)


christophs78:
Maybe we can add this to our integration-tests.

melloware:
Yep going to add one right now as working on it.

tandraschko:
+1 if well tested, code looks good

PR #6850, opened 2021-01-18
Fix #6849: Autocomplete search immediately if queryDelay=0

### LINKED ISSUE #6849: AutoComplete: Improve queryDelay=0
**Description**
When `queryDelay=0` we should not use `setTimeout` we should run immediately.

Also the docs for queryDelay are missing in the MD file.

Fire immediately when `queryDelay=0`

melloware:
- Updated missing queryDelay in docs
- Sorted 10.0 MD props because there are so many it was impossible to find them
- Don't run in `setTimeout` if queryDelay is 0


tandraschko:
+1

PR #6857, opened 2021-01-19
Fix #6856: Required-indicator not correctly resolved for CC

### LINKED ISSUE #6856: OutputLabel: Label in composite component cannot resolve required attribute of input component in another composite component
**Describe the defect**
I have two composite components. One (CC1) contains the `p:outputLabel` and the other one (CC2) a `p:inputNumber`. Both are connected using the for-attribute.

The `OutputLabelRenderer` tries to resolve the `required` attribute of the `p:inputNumber` which is defined as `required="#{cc.attrs.required}"` (required for CC2 is not hardcoded in the composite component, it is an attribute on the composite component). The problem is, that the renderer does not push the parent (==CC2) of the `p:inputNumber` to the stack and therefore the EL-resolvers use CC1 for the reference to `cc.attrs`.

OmniFaces had the same bug in its `o:tree` and fixed it using the code I will provide in a PR

- https://stackoverflow.com/a/36294437/1741604
- https://github.com/omnifaces/omnifaces/commit/bfa8d00901a6e2079220f80986675e6c3cb89bf3

**Environment:**
 - PF Version: 9.0-SNAPSHOT (But 10.0.0-SNAPSHOT has the same code)

fanste:


tandraschko:
I dont think that fix is good enough. We dont know if there is a parent CC of the outputlabel. I also wonder if JSF shouldnt always pushComponentToEL the parent component when rendering? Otherwise many stuff should be broken actually. 
We evaulate many times e.g. a ValueExpression during rendering, so the #{cc} is already available without pushing it to.

please provide a forked primefaces-test.

tandraschko:
I can just imagine that invokeOnDeepestEditableValueHolder would need to push the current visited deeper level CC?

fanste:
`invokeOnDeepestEditableValueHolder ` is not used as `forComponent` is not a composite component.

For performance reasons I can implement a null-check. Currently `pushComponentToEL` simply pushes itself to the stack if the provided component is null.

The only thing that might be broken is the `#{component}` expression as that now references the composite component and not the label anymore. But then it was already broken in the old code, because the callback is working on the `forComponent` and not on the label and therefore `#{component}` should point to the `forComponent` and not to the label.

Additionally the pushed component is only active during the callback invokation.

So three solutions to improve the PR:
 1. implement NULL-check
 2. Only push the composite component without changing the `#{component}` 
 3. push `forComponent` after the composite to make the stack as correct as it should be

fanste:
Anonther solution would be to simply use `UIViewRoot,invokeOnComponent` which should do the magic but might be a lot slower. It might be possible that this also obsoletes the `invokeOnDeepestEditableValueHolder`. But that is only an assumption.

To stick with standard API the most correct solution would be otherwise to first push the composite component (if it exists) and then the `forComponent`, That would fix `#{cc} => latest composite component before forComponent` and `#{component} => forComponent`.

tandraschko:
yeah, there is a reason that JSF 2.3 SearchExpressionResolver only works with callbacks.
It wont however replace invokeOnDeepestEditableValueHolder.

fanste:
Then which solution is desired?

1. `UIViewRoot.invokeOnComponent`
2. Resolve the parent composite component and use `cc.invokeOnComponent`. If it is `null` directly use `forComponent.invokeOnComponent`.
3. Resolve the parent composite component and only use `cc.invokeOnComponent` if the latest composite component on the stack is not the resolved one. Otherwise directly use `forComponent.invokeOnComponent`

All three solutions could be done with standard API (checked with JSF 2.2).

fanste:
`invokeOnComponent` cannot be used as Mojarra does not push the all components of the traversal to the stack but only the one that matched the clientId. MyFaces pushes everything.
That means for Mojarra that in this case `#{cc}` is never updated with the correct composite component. Only `#{component}` would be correct.

#1633  seems to address the very same issue but I don't think that anything was changed to support that use case. It's quite hard to see what really got merged and is still available.

tandraschko:
lets move the discussion into the issue, this PR just feels like a hack and also not all cases are covered.

PR #6859, opened 2021-01-19
Fix #6858: update vis.js libraries

### LINKED ISSUE #6858: Timeline: dragging the mouse up and down over the groups not moving the rest of the timeline up and down
Dragging the mouse up and down over the groups not moving the rest of the timeline up and down.

buunsal:
Closes #6858


melloware:
I removed the last line of the file `//# sourceMappingURL=vis-timeline-graph2d.min.js.map` only because in the browser console it gives an annoying "source map not found warning".   You also want to update /docs/10_0/core/thirdpartylibraries.md that is where we keep track of versions of these 3rd party libs.

PR #6886, opened 2021-01-23
Fix #6870 BreadCrumb: should not render separators in HTML

### LINKED ISSUE #6870: BreadCrumb: should not render separators in HTML
It does not comply with WAI-ARIA. See https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html

> To prevent screen reader announcement of the visual separators between links, they are added via CSS:
The separators are part of the visual presentation that signifies the breadcrumb trail, which is already semantically represented by the nav element with its label of Breadcrumb. So, using a display technique that is not represented in the accessibility tree used by screen readers prevents redundant and potentially distracting verbosity.

See also https://github.com/primefaces/primefaces/commit/4260e749c00648c86ba0a29ea5dc0e418206ff9a#r46150813

jepsar:
Fix #6870.

I know we should not modify the theme CSS files, but I did just one to indicate what should be changed in all of them.

Screen Shot 2021-01-23 at 15 10 17


mertsincan:
Good PR ;) Thanks a lot ;)

PR #6889, opened 2021-01-23
Migrationguid: Datepicker new features

christophs78:


melloware (changes requested):
Shouldn't this go in whatsnew.md in 10.0.0 and not here?

christophs78:
Yes, i can move it to whats new. At some point we only wanted to have big features on the whats new list.

melloware:
You are right I know changes go in the migration guide i thought new features went there but now you have me questioning myself...

PR #6891, opened 2021-01-24
Fix #6890: DataExporter work with column field

### LINKED ISSUE #6890: DataTable: exports do not work with field
@melloware @Rapster The datatable exporter seems not work with the new column field attribute notation if i have not missed any additionally required adaptations.

Just add the following column to /showcase/ui/data/dataexporter/basic.xhtml to verify:

```
                  <p:column
                    field="code"
                    headerText="CodeNew"/>

```
The exporter delivers empty fields for such a column.

melloware:


Rapster (changes requested):
Looks good. Use isNotBlank instead for field value checking. And also, it's no defect here, but a new feature

melloware:
I switched it to an enhancement

melloware:
OK all changed to LangUtils.isNotBlank.

PR #6892, opened 2021-01-24
[JSDoc] #6885 Update dependencies, fix some typings, regenerate in /doc/jsdocs

### LINKED ISSUE #6885: TypeScript: Updates
**Description**
@blutorange Before 10.0.0 final is released i wouldn't mind some help with a few things. 

1. A review of our current TypeScript stuff.

2. I tried to do `npm update` to update the libs and it did but got all kinds of ESLINT issues I didn't understand.

3. I think we want to update `package.json` to Timeline 7.4.6 and FullCalendar 5.5.0 but was also getting errors I didn't understand!

Much appreciated.

blutorange:
For #6885

 * Update links in docs/10_0_0, new TypeDoc version uses a module
      based structure, which changed the paths / URLs
 * Fix / add some typings and missing properties in widgets
 * Add constructor(cfg) to BaseWidget
 * PrimeFaces.ajax.Queue.xhrs should be an array
 * PrimeFaces.ajax.CallbackOnsuccess etc. set their this context to AjaxSettings (other that onstart)
 * Correct argument of AJAX client side callbacks
* Make keys in PrimeFaces.locale optional and add new keys
* Correct typing of get/setValue for PrimeFaces.widget.InputNumber
* Use JQuery.TriggeredEvent instead of JQuery.Event and update to JQuery 3.5.x typings
* Do not make classes such as BaseTree abstract, this is wrong because JS classes arent abstract, and it complicates generic typing when using Constructor types
* Add global confirmDialog and confirm source typings
* Fix typing of parameter msg of PrimeFaces.confirm / PrimeFaces.dialog.DialogHandler.confirm
* Remove PrimeFaces.widget.ConfirmDialog.ConfirmDialogMessage.onShow and add beforeShow and icon properties
* Add RemoteCommand type for modelling remote commands with generics
* Update TypeScript / TypeDoc and fix tests, adjust code for new TypeDoc version
* Update npm dependencies and npm / node version
 * Update acorn


blutorange:
Note: I also need to update the links in the showcase, but I'll do that later today

PR #6894, opened 2021-01-25
Fix #6873: Use capitalCase field if headerText/facet is null

### LINKED ISSUE #6873: DataTable: default headerText based on field
It would be cool if you could further extend the datatable column field concept such that you can use p:column without the necessity to provide a headerText:

`  <p:column field="name" />
`

instead of 

`  <p:column field="name" headerText="Name" />
`

or

`  <p:column field="name" headerText="#{messages.nameLabel}" />
`

In this default case, when the headerText attribute is missing, the headerText should be the capitalized value of the attribute field. In this case: 

`Name
`

The transformation of the field name should work as follows: 

`axxByyCzz` 

should be transformed to 

`Axx Byy Czz` 

by default.

melloware:
@Rapster @tandraschko I have tested this and its working great.  Some notes and a question:

- Unit test provided for `LangUtils.toCapitalCase()` 
- This will only use `field="productName"` as headerText `"Product Name"` only if the header facet is not used AND the headerText == null.  So this is a last resort fallback when using field.
- I check for headerText==null and not `LangUtils.isNotBlank` because I wanted to give someone the option to do this still `<p:column field="productName" headerText="" />` if they truly did not want a header text.

**Question:** I put this in the `Column(s)Base` class when calling `getHeaderText()` because this allows it to work seamlessly in all parts of the code. If I don't do this I have to litter this `capitalCase` check in 5 places DatatableRenderer, CSV ,PDF, XML, XLS Exporter code.  By placing it in the base it just works in all these places that ask for HeaderText.  I know we don't do this a lot but was wondering if this is OK?  Or would you like this check moved to the 5 places instead?

### DIFF
--- a/src/main/java/org/primefaces/util/LangUtils.java
+++ b/src/main/java/org/primefaces/util/LangUtils.java
@@ -34,6 +34,7 @@ import java.lang.reflect.TypeVariable;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.*;
+import java.util.regex.Pattern;
 
 import javax.faces.FacesException;
 import javax.xml.bind.DatatypeConverter;
@@ -41,6 +42,7 @@ import javax.xml.bind.DatatypeConverter;
 public class LangUtils {
 
     public static final Object[] EMPTY_OBJECT_ARRAY = new Object[0];
+    private static final Pattern CAPITAL_CASE = Pattern.compile("(?<=.)(?=\\p{Lu})");
 
     private LangUtils() {
     }
@@ -413,6 +415,30 @@ public class LangUtils {
         catch (NoSuchAlgorithmException e) {
             throw new FacesException(e);
         }
+    }
+
+    /**
+     * Converts a sting to capital case even counting Unicode characters.
+     * <pre>
+     * thisIsMyString = This Is My String
+     * ThisIsATest = This Is A Test
+     * helloWorld = Hello World
+     * </pre>
+     *
+     * @param value the value to capital case
+     * @return the returned value in capital case or empty string if blank
+     */
+    public static String toCapitalCase(String value) {
+        if (LangUtils.isValueBlank(value)) {
+            return Constants.EMPTY_STRING;
+        }
+
+        // split on unicode uppercase characters
+        String[] values = CAPITAL_CASE.split(value);
+        if (values.length > 0) {
+            values[0] = values[0].substring(0 , 1).toUpperCase() + values[0].substring(1).toLowerCase();
+        }
+        return String.join(" ", values);
     }
 
 
--- a/src/main/java/org/primefaces/component/column/Column.java
+++ b/src/main/java/org/primefaces/component/column/Column.java
@@ -30,6 +30,7 @@ import javax.faces.component.UIComponent;
 import javax.faces.context.FacesContext;
 
 import org.primefaces.component.celleditor.CellEditor;
+import org.primefaces.util.LangUtils;
 
 public class Column extends ColumnBase {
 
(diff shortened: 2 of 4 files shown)


src/main/java/org/primefaces/util/LangUtils.java
```diff
@@ -415,6 +415,30 @@ public static String md5Hex(byte[] bytes) {
...
+     * @param value the value to capital case
+     * @return the returned value in capital case or empty string if blank
+     */
+    public static String toCapitalCase(String value) {
+        if (LangUtils.isValueBlank(value)) {
+            return Constants.EMPTY_STRING;
+        }
+
+        // split on unicode uppercase characters
+        String[] values = value.split("(?<=.)(?=\\p{Lu})");
```
Rapster: Use pre-compiled regex

melloware: I can't believe I missed that!   Thanks its fixed.

Rapster:
>  I put this in the Column(s)Base class when calling getHeaderText() because this allows it to work seamlessly in all parts of the code. If I don't do this I have to litter this capitalCase check in 5 places DatatableRenderer, CSV ,PDF, XML, XLS Exporter code. By placing it in the base it just works in all these places that ask for HeaderText. I know we don't do this a lot but was wondering if this is OK? Or would you like this check moved to the 5 places instead?

Not too sure about this, indeed it's not a common practise, and I like xxxBase classes to stay away from any kind of logic, but what about overriding `Column#getHeaderText` & `Columns#getHeaderText` classes instead?

mertsincan:
Hi @melloware,

In the current version of PF, what is displayed in the column's header if user doesn't set a headerText to any column?

melloware:
@Rapster great advice moved to overrides in Column/Columns.

@mertsincan In PF 8.0 if a user does not user either `headerText` or `<f:fact type="header"` then nothing is displayed in the column header and even after this PR that will continue to be true.  However if they use the `field` type column completely naked like...

```xml
<p:column field="productName" />
```

The header will be "Product Name"  If they do not want that they can easily do...

```xml
<p:column field="productName" headerText="" />
```

melloware:
Example:
```xml
 <p:dataTable var="row" value="#{dtBasicView.products}">
            <p:column field="code" sortable="false" filterable="false" />
            <p:column field="name" headerText="Product Name" />
            <p:column field="category" headerText=""/>
            <p:column field="quantity" />
</p:dataTable>
```

Does this:
image

mertsincan:
@melloware, thanks a lot for info ;) 50% for me :D

Rapster:
I think it's a cool feature for prototypes or mono language app (and yet...). Otherwise I don't see this used really often, since users will certainly use i18n

melloware:
Agreed but in one my my mono language apps I always name my fields properly and I have the boiler plate of repeating myself for the headerText so I do see the use case here.  Especially for internal apps written for the business owners in one language say for an internal HR app.

or for CRUD screens when building viewer editor for CRUD table editing.

melloware:
So right now it sounds like this is...

Myself +1
Mert +0.5
Rapster +0.5

??

tuerker:
dear all, this is a very useful extension, especially for prototyping it is great.

With a little step it will also great for production systems. As @Rapster correctly stated, i18n is typically used.
Here, now the the headerText should be created via the message.properties file such that

<p:column field="name"/>

sets the headerText to the value set in messages.properties

name=XXXXX

@melloware as always many thanks to you.

stolp:
I think it has its uses and it should not hurt anyone.

+1

tandraschko:
as PF is a nice way for rapid prototyping, i think this is quite usefull.

PR #6900, opened 2021-01-26
Fix #3551: Datatable remember range selection rowindex between updates

### LINKED ISSUE #3551: DataTable: Multiselection with shift-click doesn't work after update
## 1) Environment
- PrimeFaces version: **6.2**
- Does it work on the newest released PrimeFaces version? **No - 6.2**
- Application server + version: **WildFly 12.0**
- Affected browsers: **Firefox**

## 2) Expected behavior

After clicking entry a) of the datatable and then shift-clicking a second entry b), all entries from a) to b) should be selected, even if the DataTable has bean updated in the meantime.

## 3) Actual behavior

After first click, entry a) is selected.
After updating the DataTable, still a) is selected, this is OK.
But after shift-clicking b), all entries from the first row to b) are selected.

## 4) Steps to reproduce

With the example, select one entry, wait a second, then shift-select another entry

## 5) Sample XHTML

Just use the "Multiple with Meta and Shift keys" example from:
https://www.primefaces.org/showcase/ui/data/datatable/selection.xhtml

and add the following line to make it update the DataTable, this issue is not just related to the <p:poll> component, it could also be updated with a button or anything else: 
`<p:poll interval="1" update="multipleDT"/>`

melloware:
This fix simply sets the shift range index to NULL on first load but will not forget it if the datatable is updated.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
+++ b/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
@@ -783,8 +783,12 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
         this.selection = (preselection === "") ? [] : preselection.split(',');
 
         //shift key based range selection
-        this.originRowIndex = null;
-        this.cursorIndex = null;
+        if(!this.originRowIndex) {
+            this.originRowIndex = null;
+        }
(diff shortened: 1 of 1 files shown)


melloware:
Closing PR as it does not fix the whole issue.

PR #6901, opened 2021-01-26
Fix #6897 - DataTable: support Converter#forClass for field shorthand

### LINKED ISSUE #6897: DataTable: support Converter#forClass for field shorthand notation
@melloware @Rapster It seems that the field notation ignores existing faces converters. 
I am seeing this for LocalDateTime fields.

```
<p:column field="startDate" headerText="startDate" />
```
delivers field values of the default LocalDateTime pattern, e.g. 2021-01-22T09:02:43.147

while in case of
```
<p:column field="startDate" headerText="startDate">
      <h:outputText value="#{entity.startDate}">
</p:column>
```
the field values are actually converted using the existing converter, e.g. 2021-01-22 09:02.
my converter is defined as follows:

```
@FacesConverter(forClass = LocalDateTime.class)
public class LocalDateTimeConverter implements Converter<Object> {
 ...
    private static String getPattern(UIComponent component) {
        String pattern = (String) component.getAttributes().get("pattern");
        if (pattern == null) {
            pattern = "yyyy-MM-dd HH:mm";
        }
        return pattern;
    }
}

```

I am not sure whether this a bug report  or feature request. In any case, many thanks in advance for having a view on this issue.

Rapster:


src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
```diff
@@ -1342,6 +1338,15 @@ else if (column instanceof DynamicColumn) {
         writer.endElement("td");
     }
 
+    protected void encodeDefaultFieldCell(FacesContext context, DataTable table, UIColumn column, ResponseWriter writer) throws IOException {
+        Object value = UITable.createValueExprFromVarField(context, table.getVar(), column.getField()).getValue(context.getELContext());
```
melloware: I wonder if you should wrap this up in a static `UITable` static helper that returns the value as a String run through the converter.  That way the DataExporter can use it as well.

melloware:
I am thinking something like this..

```java
String value = UITable.getConvertedFieldValue(context, table.getVar(), column);
```

Rapster:
Done

melloware (approved):

melloware:
OK I am going to use this code to fix the DataExporter as well.

tandraschko:
please dont just merge such stuff now after RC1 without another review, its to risky to break something.
Why is it static?

melloware:
Its a static helper so it can be resued across Exporter and DT.  See other PR which was dependent on this fix: https://github.com/primefaces/primefaces/pull/6905

tandraschko:
yeah but there is no reason to make it static
you always work with the DT instance in this places

melloware:
I see what you mean.

tandraschko:
commited the small refactoring

tandraschko:
in theory we should not even pass FacesContext but DataTable is still not in a nice shape
much better as in 8.0 but a lot more todo for 11.0

PR #6903, opened 2021-01-27
Fix #6902: Timeline zoom/move widget methods are by user

### LINKED ISSUE #6902: Timeline: Calling move()/zoom() functions on timeline does not fire rangeChanged event
When using the move/zoom buttons on timeline as per the example here [https://www.primefaces.org/showcase/ui/data/timeline/custom.xhtml?jfwid=fbbab](url) 
the rangeChanged ajax event does not get fired.

melloware:
Because the widget methods do not involved a mouse the byUser flag from timeline is false.  So we must tell the code that this was initiated by a user when calling widget `zoom` or `move`


melloware:
Tested with and without CSP and can very #5455 is still working.

buunsal (approved):

Issue #4536, opened 2019-02-13
Menu: wrong calculated position with overlay="true"


astranCZ:
## 1) Environment
- PrimeFaces version: 7.0.RC2
- Application server + version: Tomcat 7 and 9
- Affected browsers: All

## 2) Expected behavior
Calculated position of tag p:menu (overlay="true") on p:dialog is not dependent on the position of the scroll bar main window. 

## 3) Actual behavior
Calculated position of tag p:menu (overlay="true") on p:dialog is dependent on the position of the scroll bar main window.

## 4) Steps to reproduce
Create page with vertival scroll bar. Open p:dialog with button with p:menu (overlay="true").
p:menu position is dependent on the location of the scroll bar.

## 5) Sample XHTML

```
<html xmlns="http://www.w3.org/1999/xhtml"
	xmlns:h="http://java.sun.com/jsf/html"
	xmlns:f="http://java.sun.com/jsf/core"
	xmlns:p="http://primefaces.org/ui">

<h:head>
</h:head>

<h:body style="height: 1500px;">
	<p:commandButton value="Modal" type="button" onclick="PF('dlg2').show();" />

	<p:dialog header="Modal Dialog" widgetVar="dlg2" modal="true" position="top" height="150">
		<h:form>
			<h:outputText value="This is a Modal Dialog." />
			<h3>Overlay</h3>
			<p:commandButton id="dynaButton" value="Show menu" type="button" />
			<p:menu overlay="true" trigger="dynaButton" my="left top" at="left bottom">
				<p:menuitem value="Save" />
				<p:menuitem value="Update" />
			</p:menu>
		</h:form>
	</p:dialog>
</h:body>
</html>
```

OBreidenbach:
I have the same behavior with a SelectOneMenu inside of a Dialog. If the drop down is visible it stays on its position in the view port during scrolling. So it looses contact to the component.
That did not happen with PF6.2

melloware:
@OBreidenbach do you mean this issue?  https://github.com/primefaces/primefaces/issues/4180

melloware:
@astranCZ Like I said in the other thread this isn't a fix more like a workaround but for now you can...

On 7.0RC2 is to set the p:dialog property to blockScroll="true".  That disables the page scrollbar while a modal dialog is being displayed.  To me this is good default behavior of a modal dialog as you should not need to scroll the outer page if a modal is displayed.

```xml
[...]

Issue #5492, opened 2020-01-02
Upgrade jQuery UI 1.13


tandraschko:
As jQuery UI is not active anymore, we should try to move away from long-term perspecive.

Currently we have following dependencies:
- some widgets like calendar and slider
- animations
- utils
- class names (we just use the same names, we don't need to change it)

tandraschko:
A first step would be to check which components depend on jQuery UI and check for replacement.

tandraschko:
for animations there are many replacements availble like animation.css or velocity.js

melloware:
I follow JqueryUI and it has been getting some love recently from @mgol  They are updating it at the very least to support  the upcoming JQuery 4.0 release.  See recent PR updates here: https://github.com/jquery/jquery-ui/pulls?q=is%3Apr+is%3Aclosed+sort%3Aupdated-desc

melloware:
Here is everything in Jquery UI.  

 widget.js
 position.js
 data.js
 disable-selection.js
 focusable.js
 form-reset-mixin.js
 keycode.js
 labels.js
 scroll-parent.js
 tabbable.js
 unique-id.js
 widgets/draggable.js
 widgets/droppable.js
 widgets/resizable.js
 widgets/selectable.js
 widgets/sortable.js
 widgets/datepicker.js
 widgets/mouse.js
 widgets/slider.js
 effect.js
 effects/effect-blind.js
 effects/effect-bounce.js
 effects/effect-clip.js
 effects/effect-drop.js
 effects/effect-explode.js
 effects/effect-fade.js
 effects/effect-fold.js
 effects/effect-highlight.js
 effects/effect-puff.js
 effects/effect-pulsate.js
 effects/effect-scale.js
 effects/effect-shake.js
 effects/effect-size.js
 effects/effect-slide.js
 effects/effect-transfer.js

melloware:
JqueryUI release a 1.13 alpha which removes all deprecated jquery functions and contains some bug fixes.

https://github.com/jquery/jquery-ui/releases/tag/1.13.0-alpha.1

melloware:
Final Release blog post: https://blog.jqueryui.com/2021/10/jquery-ui-1-13-0-released/

melloware:
Adding Security label as this fixes 3 CVE's:

[CVE-2021-41184](https://github.com/jquery/jquery-ui/security/advisories/GHSA-gpqq-952q-5327#advisory-comment-69059)
[...]

PR #6907, opened 2021-01-28
Fix #6906: Menu fix both dialog and non dialog show

### LINKED ISSUE #5035: Menu: wrong position when using trigger element

### LINKED ISSUE #6906: Menu: wrong calculated position with overlay="true"
Same defect as #4536. Defect was solved in 7.x but comes back with 8.0

## 1) Environment
PrimeFaces version: 8.0.4
Application server + version: TomEE 8.0.4
Affected browsers: All

## 2) Expected behavior
Calculated position of tag p:menu (overlay="true") on p:dialog is not dependent on the position of the scroll bar main window.

## 3) Actual behavior
Calculated position of tag p:menu (overlay="true") on p:dialog is dependent on the position of the scroll bar main window.

## 4) Steps to reproduce
Create page with vertival scroll bar. Open p:dialog with button with p:menu (overlay="true").
p:menu position is dependent on the location of the scroll bar.

## 5) Sample XHTML
``` 

<h:head>
</h:head>

<h:body style="height: 1500px;">
	<p:commandButton value="Modal" type="button" onclick="PF('dlg2').show();" />

	<p:dialog header="Modal Dialog" widgetVar="dlg2" modal="true" position="top" height="150">
		<h:form>
			<h:outputText value="This is a Modal Dialog." />
			<h3>Overlay</h3>
			<p:commandButton id="dynaButton" value="Show menu" type="button" />
			<p:menu overlay="true" trigger="dynaButton" my="left top" at="left bottom">
				<p:menuitem value="Save" />
				<p:menuitem value="Update" />
			</p:menu>
		</h:form>
	</p:dialog>
</h:body>
</html>
```

melloware:
Draft for now which puts back the original fix for #4536.  Waiting to hear back from @astappiev  on his use case for #5035 which broke this.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/menu/menu.base.js
+++ b/src/main/resources/META-INF/resources/primefaces/menu/menu.base.js
@@ -106,9 +106,12 @@ PrimeFaces.widget.Menu = PrimeFaces.widget.BaseWidget.extend({
             }
         });
 
-        //Hide overlay on resize
-        PrimeFaces.utils.registerResizeHandler(this, 'resize.' + this.id + '_align', $this.jq, function() {
-            $this.align();
+        //Hide overlay on resize/scroll
+        PrimeFaces.utils.registerResizeHandler(this, 'resize.' + this.id + '_hide', $this.jq, function() {
+            $this.hide();
+        });
+        PrimeFaces.utils.registerScrollHandler(this, 'scroll.' + this.id + '_hide', function() {
+            $this.hide();
         });
 
         //dialog support
@@ -121,9 +124,11 @@ PrimeFaces.widget.Menu = PrimeFaces.widget.BaseWidget.extend({
      */
     setupDialogSupport: function() {
         var dialog = this.trigger.parents('.ui-dialog:first');
+        this.isDialogContainer = false;
 
         if(dialog.length == 1 && dialog.css('position') === 'fixed') {
             this.jq.css('position', 'fixed');
+            this.isDialogContainer = true;
         }
     },
 
@@ -135,8 +140,18 @@ PrimeFaces.widget.Menu = PrimeFaces.widget.BaseWidget.extend({
             'z-index': PrimeFaces.nextZindex(),
             'visibility': 'hidden'
         });
-        this.align();
-        this.jq.show();
+
+        if(this.isDialogContainer) {
+            // #6096 In dialog must be shown then aligned
+            this.jq.show();
+            this.align();
+        }
+        else {
+            // #5035 Non dialog must be aligned then shown
+            this.align();
+            this.jq.show(); 
+        }
+
         this.jq.css('visibility', '');
     },
 


melloware:
OK fixed it and made sure both test cases work.

#6906: Inside Dialog:
[pf-6906.zip](https://github.com/primefaces/primefaces/files/5889298/pf-6906.zip)

#5035: Outside Dialog:
[pf-5035.zip](https://github.com/primefaces/primefaces/files/5889304/pf-5035.zip)

astappiev:
Looks more like a workaround, ideally, we have to investigate why the shift appears and fix a reason. But the method is deep inside Jquery UI.

Also, if you open a menu in modal and scroll the window, the menu will be fixed on the screen, but a button will be moved with a page.

Don't you have plans to get rid of JQuery and all plugins? I guess something like [Popper.js](https://popper.js.org/) should be better.

melloware:
@astappiev Not in 10,0.0 it doesn't it scrolls with the window.  That was fixed already.

astappiev:
Sure, PF 10 is already RC.

But in general, most of the components which use UI is using Jquery and/or Jquery UI. I guess it will be super hard to replace it.
Meanwhile, Jquery UI looks abandoned and JQuery is rarely used in new projects.

melloware:
Right not arguing the jQuery UI point.    Was simply stating your statement of _"Also, if you open a menu in modal and scroll the window, the menu will be fixed on the screen, but a button will be moved with a page."_ was incorrect.  It works fine in 10.0.0.RC1.

melloware:
You can view the ticket about jQuery UI here: https://github.com/primefaces/primefaces/issues/5492

astappiev:
Thanks for the link to the ticket.

And yes, I gave the wrong description. In my reproducer, if you open user menu and then scroll down, the navbar will be fixed, but the menu will be moved.
image

But I know, the navbar is not PF component and it has a fixed position.

melloware:
I just fixed that issue too.  Refresh this PR.

```javascript
PrimeFaces.utils.registerScrollHandler(this, 'scroll.' + this.id + '_hide', function() {
            $this.hide();
        });
```

it should have been there we do this for other overlays.

astappiev:
Hm, thank you. Even simpler than I expected.

melloware:
we actually do that on all the other overlays somehow i missed this one.

tandraschko:
This is something that must be reviewed by PrimeTek.
But in general i wonder what happens in other overlays like OverlayPanel?

melloware:
Closing this PR as Mert is fixing this with his other issues: #6978 and #6973

PR #6909, opened 2021-01-29
Fix #6908: remove ScheduleRenderingMode and occurences

### LINKED ISSUE #6908: DefaultScheduleEvent: Remove ScheduleRenderingMode
**Description**
Coming from the Migration Guide of FullCalendar (https://fullcalendar.io/docs/upgrading-from-v4) the "rendering" attribute has been replaced with the new "display" attribute. I am working on a project where we previously set the rendering attribute via the Enum "ScheduleRenderingMode" in combination with the builder function "renderingMode".

Since the "rendering" attribute does not exist anymore this function is now obsolete.

I would suggest removing the ScheduleRenderingMode and its occurences/functions in other classes entirely since it is not needed anymore.

fcjm:
I deleted the ScheduleRenderingMode and its occurences since "rendering" is not used anymore in FullCalendar. See issue: https://github.com/primefaces/primefaces/issues/6908.


Rapster:
display should be an enum no, WDYT?

> These settings have been renamed to display and eventDisplay respectively. They accept:
> - 'block' - new. always display as a solid rectangle in daygrid
> - 'list-item' - new. always display with a dot when in daygrid
> - 'auto' - new. the default. in daygrid, will display as 'block' if all-day or multi-day, otherwise will display as 'list-item'
> - 'background'
> - 'inverse-background'
> - 'none' - new. don't display at all

fcjm:
Yes, you are right. An enum would fit the display attribute better. I will change it in a bit.

fcjm:
Okay, I changed display to use an enum. I pretty much just did it like it was done before with the ScheduleRenderingMode, only renaming it to ScheduleDisplayMode. I also tested this solution in our application and it worked.

melloware:
Thanks for doing this. I just added display the other day and left it a String because I wasn't sure if fullcalendar would keep changing what was available in there but I have no problem with the enum.   And thanks for the PR!  I will fix the Showcase when this PR is merged.

melloware (approved):

melloware:
2 more changes to add to this PR.

1. in /docs/migrationguide/10_0_0.md please update the Schedule section to say  _* DefaultScheduleEvent replaces ScheduleRenderingMode used in "rendering" with ScheduleDisplayMode used in "display" with values_

2. in /docs/10_0_0/components/schedule.md there is a section about the Default Event we need to replace the rednering with display

melloware:
I also added these 3 fields to the DefaultSheduleEvent do you mind adding them to schedule.md as well?
```
private String backgroundColor;
    private String borderColor;
    private String textColor;
```

fcjm:
Yes i will do that. Just a question, do you want me to list the different enum values in the /docs/migrationguide/10_0_0.md ?

melloware:
Yep I think you can just list them like "block", "inline" background" etc.

fcjm:
Is that okay?

melloware:
perfect!

tandraschko:
+1

PR #6914, opened 2021-01-30
Fix #6912: FileUpload global attribute

### LINKED ISSUE #6912: FileUpload: Ajax Status is not working while using basic+auto upload
The upload component with basic and auto options is not triggering the ajax status. I'm not sure if this component has been designed to trigger ajax status, but I guess this can be a good feature if has not been planned.

**Reproducer**
https://github.com/garcia-jj/primefaces-issue-upload-ajax

**Environment:**
 - PF Version: _10.RC1_
 - JSF + version: _e.g. Mojarra 2.2.
 - Affected browsers:  ALL_

**To Reproduce**
2. Click on 'Click' button, and you can see that Ajax Status will be updated
3. Upload a file, and you can see that Ajax Status will not be updated while browser is uploading the file.
4. Using Chrome 87.0.4280.141

**Expected behavior**
Ajax status indicator updated while the file is uploading

Thank you, team.

melloware:
- Added `global` attribute to FileUpload
- Since it uses a custom $ajax request had to do the AJAX start, send, complete in this request
- Sorted the MD attributes as there are a lot of them.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/fileupload/3-fileupload.simple.js
+++ b/src/main/resources/META-INF/resources/primefaces/fileupload/3-fileupload.simple.js
@@ -22,6 +22,7 @@
  * @prop {string} cfg.invalidSizeMessage Message to display when size limit exceeds.
  * @prop {number} cfg.maxFileSize Maximum allowed size in bytes for files.
  * @prop {string} cfg.messageTemplate Message template to use when displaying file validation errors.
+ * @prop {string} cfg.global Global AJAX requests are listened by ajaxStatus, false will not trigger ajaxStatus.
  */
 PrimeFaces.widget.SimpleFileUpload = PrimeFaces.widget.BaseWidget.extend({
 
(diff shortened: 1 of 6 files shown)


tandraschko:
our AJAX core always sets global=false, to i think its by design. So our PF global is just a flag for trigger our events, but not pass it to jquery.
i think you also may adjust 2-fileupload.js?

melloware:
You are right added to `2-fileupload.js` as well.

garcia-jj:
Hello @melloware . Thanks for this pull request. I'll take a test tomorrow, and I'll keep you posted. I appreciate your help.

Thank U

src/main/resources/META-INF/resources/primefaces/fileupload/3-fileupload.simple.js
```diff
@@ -234,11 +238,14 @@ PrimeFaces.widget.SimpleFileUpload = PrimeFaces.widget.BaseWidget.extend({
             data: formData,
             processData: false,
             contentType: false,
-            global: false,
+            global: $this.cfg.global,
```
tandraschko: pass false here, like we do in PF ajax core

melloware: Fixed to false.

PR #6924, opened 2021-02-02
Fix #6922: SelectOne highlight correct item when grouped

### LINKED ISSUE #6922: SelectOneMenu: Clicking the same item twice leads to wrong item being highlighted
**Describe the defect**
If you click on the same Item twice in the dropdown menu of the SelectOneMenu component with groups you will see that the wrong item is being selected & highlighted. From my research it seems like the problem lies within the revert() function of the SelectOneMenu widget which unselects the current item and instead tries to select the "Please Select" item.

Called in forms.selectonemenu.js: 

        .on('click.selectonemenu', function() {
            $this.revert();
            $this.selectItem($(this));
            $this.changeAriaValue($(this));
        });

revert function: 

    /**
     * Unselect the selected item, if any, and select the `please select` option.
     */
    revert: function() {
        if(this.cfg.editable && this.customInput) {
            this.setLabel(this.customInputVal);
            this.items.filter('.ui-state-active').removeClass('ui-state-active');
            this.items.eq(0).addClass('ui-state-active');
        }
        else {
            // this gets called
            this.highlightItem(this.items.eq(this.preShowValue.index()));
        }
    },

The function selectItem() will be skipped for the most part, since the currentOption still matches the originally selected item. Therefore sameOption will be true and the changes from the revert() function will remain.

**To Reproduce**
SelectOneMenu Showcase -> Grouped

If you select "Turkey" twice for example the label will change to "Germany"

melloware:
This uses the same code used in other places in the SelectOne to determine the correct index because Groups throw off the index.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/forms/forms.selectonemenu.js
+++ b/src/main/resources/META-INF/resources/primefaces/forms/forms.selectonemenu.js
@@ -399,7 +399,7 @@ PrimeFaces.widget.SelectOneMenu = PrimeFaces.widget.DeferredWidget.extend({
             this.items.eq(0).addClass('ui-state-active');
         }
         else {
-            this.highlightItem(this.items.eq(this.preShowValue.index()));
+            this.highlightItem(this.items.eq(this.options.index(this.preShowValue)));
         }
     },
 


tandraschko:
+1 if you tested some cases

melloware:
Doing some more testing and also waiting to hear from OP about another possible issue related.

PR #6927, opened 2021-02-02
Fix #6923/#6926: Sidebar modal and showCloseIcon attrs

### LINKED ISSUE #6923: Sidebar: add attribute to disable modal masking
The Sidebar always darkens the background when opened. There should be a `modal` attribute to set the masking (like there is for the Dialog component).

### LINKED ISSUE #6926: Sidebar: add attribute to disable close on click
The Sidebar always closes when clicking anywhere on the background.
There should be an option to disable this behaviour.
There could also be a `permanent` attribute to show the sidebar permanently.

melloware:
Following OverlayPanel lead added 2 new attributes `modal` and `showCloseIcon` to follow it.

### DIFF
--- a/src/main/java/org/primefaces/component/sidebar/SidebarRenderer.java
+++ b/src/main/java/org/primefaces/component/sidebar/SidebarRenderer.java
@@ -64,7 +64,9 @@ public class SidebarRenderer extends CoreRenderer {
             writer.writeAttribute("style", style, null);
         }
 
-        encodeCloseIcon(context);
+        if (bar.isShowCloseIcon()) {
+            encodeCloseIcon(context);
+        }
 
         if (!bar.isDynamic()) {
             renderChildren(context, bar);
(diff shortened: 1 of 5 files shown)


src/main/java/org/primefaces/component/sidebar/SidebarRenderer.java
```diff
@@ -91,9 +93,11 @@ private void encodeScript(FacesContext context, Sidebar bar) throws IOException
         WidgetBuilder wb = getWidgetBuilder(context);
         wb.init("Sidebar", bar)
                 .attr("visible", bar.isVisible(), false)
+                .attr("modal", bar.isModal(), false)
```
tandraschko: false? shouldnt it be true?
This flag is actually used to NOT pass the default value to the JS.
the JS then needs to interpret "null" as "true" of course

melloware: Yep I was being lazy let me fix.

melloware:
Updated for proper defaults.

tandraschko:
looks good now
+1 if well tested

PR #6937, opened 2021-02-03
Fix #6355: DatePicker fixed missing formId in lazy metadata Ajax request

### LINKED ISSUE #6355: DatePicker: lazy load disabled dates

jepsar:
originally should have been part of this ticket: https://github.com/primefaces/primefaces/issues/6355

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
+++ b/src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
@@ -25,6 +25,7 @@
  * @param {JQueryPrimeDatePicker.PickerInstance} PrimeFaces.widget.DatePicker.PreShowCallback.datePicker The current
  * date picker instance.
  *
+ * @prop {string} cfg.formId Client ID of the form that is used for AJAX requests.
  * @prop {JQuery} input The DOM element for the hidden input element with the selected date.
  * @prop {JQuery} jqEl The DOM element for the inline picker or the input.
  * @prop {boolean} refocusInput Whether focus should be put on the input again.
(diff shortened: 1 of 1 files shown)


tandraschko:
forms are mostly resolved on the server side and passed to the JS.
Can you check how other places did this?

jepsar:
https://github.com/primefaces/primefaces/blob/47a1f00dc3f2be3a66e37e75330eba65a9e380d8/src/main/resources/META-INF/resources/primefaces/schedule/1-schedule.js#L44

tandraschko:
What about DataTable e.g.?

jepsar:
It is used halve a dozen times in the data table:

https://github.com/primefaces/primefaces/blob/47a1f00dc3f2be3a66e37e75330eba65a9e380d8/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js#L1760

But I cannot find where it is set. It's not in the script nor in the renderer. Also, based on this search:

https://github.com/primefaces/primefaces/search?l=JavaScript&q=formId

it seems to be "standard" to set the `formId` in the widget `init` function.

One use in Java:

https://github.com/primefaces/primefaces/search?l=Java&q=formId

jepsar:
I should at least move it to the init method, so it does not have to be resolved every Ajax request...

melloware:
@tandraschko I agree this is quite strange.  Should we just set formId in the BaseWidget `init` method so we don't have to repeat ourselves and the closest form will always be set in all widgets?

tandraschko:
not every widget requires a form, so i would not change this.
I would just change this PR to make it like other widgets, to init form in init?

jepsar:
I think we need the same fix on data table though. Will update this PR this evening (CET).

tandraschko:
how does it work currently there?

melloware (approved):

jepsar:
By the way.. I'm still not sure if not having a formId will cause issues. It will be a small performance boost to have it, so less form elements need to be handled and send. But it does seem to be missing in data table. See for example:

https://www.primefaces.org/showcase/ui/data/datatable/lazy.xhtml

and in the JS console:

PF('widget_form_customerTable').cfg.formId

tandraschko:
would be great if you can create a issue and play a bit with it.
in general we should add it as the AJAX core will otherwise fallback to the first form found in the DOM normally, which isnt correct.

PR #6945, opened 2021-02-04
Fix #6940: InputNumber firing change event when value not changed

### LINKED ISSUE #6940: InputNumber: ajax change event fired even if value isn't changed
**Describe the defect**
When i use and p:inputNumber with ajax behaviour on change that updates the component container, the event gets fired even if the value isn't changed. Note: the exact same example using p:inputText is working correctly.

**Environment:**
 - PF Version: 8.0.7

**To Reproduce**
1. select the first input number, change it's value
2. use TAB to move to the second (a change event is fired - correct)
3. from the second input number use SHIFT+TAB to go back (a new change event is fired - WRONG)
4. from now on, moving from one input to the other (without changing the value) fires a change ajax event

**Expected behavior**
A change event should be fired only when the input value is actually changed

**Example XHTML**
```html
<h:form id="frmTest">
	<p:outputPanel styleClass="x">
		<p:inputNumber value="#{bean.a}">
			<p:ajax process="@this" update="@(.x)"/>
		</p:inputNumber>
		<p:inputNumber value="#{bean.b}">
			<p:ajax process="@this" update="@(.x)"/>
		</p:inputNumber>
	</p:outputPanel>
</h:form>
```

**Example Bean**
```java
@Named
@ViewScoped
	private double a=3.2;
	private double b=1.2;
	public double getA() {
		return a;
	}
	public void setA(double a) {
		this.a = a;
		return;
	}
	public double getB() {
		return b;
	}
	public void setB(double b) {
		this.b = b;
		return;
	}
}
```

melloware:
This issue is that `set` does not initialize the `rawValueOnFocus` so we simply need to initialize it when we are setting the value the first time the autonumeric is created.

Also I noticed in CopyValueToHiddenInput if comparing "1.20" to "1.2" it was triggering the copy and making no difference. It should be comparing the numerical values not the String values.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js
+++ b/src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js
@@ -57,6 +57,8 @@ PrimeFaces.widget.InputNumber = PrimeFaces.widget.BaseWidget.extend({
         if (this.valueToRender !== "") {
             //set the value to the input the plugin will format it.
             this.autonumeric.set(this.valueToRender);
+            // GitHub #6940 blur firing too many change events
+            this.autonumeric.rawValueOnFocus = this.valueToRender;
         }
 
         this.copyValueToHiddenInput();
@@ -160,7 +162,7 @@ PrimeFaces.widget.InputNumber = PrimeFaces.widget.BaseWidget.extend({
 
         var newVal = this.getValue();
 
-        if (oldVal !== newVal) {
+        if (Number(oldVal) !== Number(newVal)) {
             this.setValueToHiddenInput(newVal);
         }
 


tandraschko:
+1 if well tested
i also had some problems in the past that inputNumber throws
1) 2 AJAX onchange request on real onchange
2) 1 AJAX onchange request when onblur

would be great if you can verify this always when changing somehting about the event handling in inputnumber

melloware:
So far with this fix in my testing it appears the onChange is fired only when you actually change the number and then leave the field.    I think we have it all correct only 1 onchange is being fired when its actually supposed to fire.  at least from my testing.

fanste:
This nice little trick is not that nice ;)

`Number("")` and `Number("0")` returns the same (`0`). Therefore the component is broken for the value `0` if no initial value is set (or the field was cleared).

My suggestion: `((oldVal == '') ^ (newVal == '')) || Number(oldVal) !== Number(newVal)` (`^` = XOR)
Read it as: Either oldVal or newVal is empty or numeric value is not equal

melloware:
Crap you are right.  OK let me make that fix.

melloware:
however what if oldvalue and newvalue are both '' shouldn't that not fire a change event?

fanste:
Than that won't be a change event, I think. Empty is empty.

melloware:
OK I assumed you test this with the monkeypatch?
```javascript
if (PrimeFaces.widget.InputNumber) {

    PrimeFaces.widget.InputNumber.prototype.copyValueToHiddenInput = function() {
        var oldVal = this.hiddenInput.val();
        var newVal = this.getValue();

        if (((oldVal === '') ^ (newVal === '')) || Number(oldVal) !== Number(newVal)) {
            this.setValueToHiddenInput(newVal);
        }

        return oldVal;
    }
};
```

fanste:
I've tested it that way and empty string / zero is working again

PR #6947, opened 2021-02-04
Fix #6946: SelectOne add aria-label to _input

### LINKED ISSUE #6946: SelectOneMenu: WCAG warning for ARIA
**Describe the defect**
See forum post: https://forum.primefaces.org/viewtopic.php?f=3&t=65116&

Result of WCAG validator..

image

melloware:


tsobis (commented):
Fixed the problem on version 8 as well, both achecker and wave are accepting it and the validator.w3.org does not give any problems.
Sorry, I made a mistake, I tested my changes of the code.
`menu.getAttributes().get(HTML.ARIA_LABEL).toString() `
instead of 
`menu.getPlaceholder()`
 Running tests again.

tandraschko:
+1 if well tested, code looks good

tsobis:
Both 
` menu.getLabel()`
and 
`menu.getPlaceholder()`
return null
in my case i fetched the label from the aria-label attribute of the component
`menu.getAttributes().get(HTML.ARIA_LABEL)`
I hope this helps

melloware:
@tsobis Can you check my commits again I just made an improvement where I look for the aria-label and if not found falls back to the original code of label then placeholder.

tsobis:
Yes, this time seems to work fine
selectOneFix

melloware:
Awesome thank you for reporting this issue and testing it!

PR #6952, opened 2021-02-05
#6921: PrimeFaces.Ajax: add ignoreAutoUpdate to PrimeFaces.current().ajax().update()

### LINKED ISSUE #6921: PrimeFaces.Ajax: add ignoreAutoUpdate to PrimeFaces.current().ajax().update()
eg `p:commandButton` and `p:autoUpdate` already support this

Keep existing method-signatures AS-IS. Add another one which allows to pass ignoreAutoUpdate - boolean. Internally call the same implementation.

christophs78:


tandraschko:
Maybe just call it ignoreAutoUpdate in Primefaces.ajax?

melloware (approved):

PR #6955, opened 2021-02-05
Fix #6948: Widget getParentForm lazy initialization

### LINKED ISSUE #6948: DataTable/TreeTable: widget cfg.formId is missing
See for example:

https://www.primefaces.org/showcase/ui/data/datatable/lazy.xhtml

and in the JS console:

PF('widget_form_customerTable').cfg.formId

See also https://github.com/primefaces/primefaces/pull/6937

melloware:
Tested locally and verified all is working fine sorting, filtering, paging etc.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
+++ b/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
@@ -831,7 +831,7 @@ PrimeFaces.widget.AutoComplete = PrimeFaces.widget.BaseWidget.extend({
                 source: this.id,
                 process: this.id,
                 update: this.id,
-                formId: this.cfg.formId,
+                formId: this.getParentFormId(),
                 onsuccess: function (responseXML, status, xhr) {
                     PrimeFaces.ajax.Response.handle(responseXML, status, xhr, {
                         widget: $this,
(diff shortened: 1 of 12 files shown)


Rapster:
What about putting that in BaseWidget? So all widgets gets it?

melloware:
I suggested that too and @tandraschko didn't like it because he said all widgets don't need it but I didn't see the harm in it as even it is sets it to NULL its not a big deal as core.ajax that uses this field will just try again if the field even has a partial AJAX request.

Rapster:
The only issue I see, is that retrieving formId might have a tiny downside in terms of performance but can't we make it lazy?

tandraschko:
Lazy would be ok, something like a "getParentForm" method

melloware:
OK updated this PR to have a lazy `getParentForm` in the base widget.  Changed all places using it.

This is much better as it defers looking up the parent until the first time it is used and then all subsequent calls after that has the form so this is a good optimization

tandraschko:
Please introduce getParentForm and getParentFormId, which reuses the first method

melloware:
Good suggestion. Done.

tandraschko:
i think we should even cache a null formId / if no parent form is present?

tandraschko:
but lets leave it as it is for now

PR #6958, opened 2021-02-06
Fix #6913: Datatable NPE

### LINKED ISSUE #6913: DataTable: Sort on ListTableModel fails with NPE
DataTable Sort on ListDataModel fails with NPE - WrappedData List lost after sort order has been changed.
Using List instead of ListDataModel works.

**Environment:**
 - PF Version: _10.0.0-SNAPSHOT_
 - JSF + version: _Mojarra 2.3.14_
 - Java Runtime: 11.0.9.1

**Example Bean**
```java
@Named
@ViewScoped

    private DataService dataService;
    private DataModel dataModel;

    public DataModel getDataModel() {
         if ( dataModel == null ) {
              dataModel = new ListDataModel<>( dataService.getData() );
         }
         return dataModel;
    }
}
```

**Stacktrace**
```
10:06:24,955 INFO  [javax.enterprise.resource.webcontainer.jsf.context] (default task-7) java.lang.NullPointerException: java.lang.NullPointerException
	at deployment.test-SNAPSHOT.war//org.primefaces.component.datatable.feature.SortFeature.sort(SortFeature.java:154)
	at deployment.test-SNAPSHOT.war//org.primefaces.component.datatable.feature.SortFeature.decode(SortFeature.java:107)
	at deployment.test-SNAPSHOT.war//org.primefaces.component.datatable.DataTableRenderer.decode(DataTableRenderer.java:72)
	at deployment.test-SNAPSHOT.war//javax.faces.component.UIComponentBase.decode(UIComponentBase.java:509)
	at deployment.test-SNAPSHOT.war//org.primefaces.component.api.UIData.processDecodes(UIData.java:117)
	at deployment.test-SNAPSHOT.war//com.sun.faces.context.PartialViewContextImpl$PhaseAwareVisitCallback.visit(PartialViewContextImpl.java:606)
	at deployment.test-SNAPSHOT.war//com.sun.faces.component.visit.PartialVisitContext.invokeVisitCallback(PartialVisitContext.java:159)
	at deployment.test-SNAPSHOT.war//org.primefaces.component.api.UIData.visitTree(UIData.java:683)
	at deployment.test-SNAPSHOT.war//javax.faces.component.UIComponent.visitTree(UIComponent.java:1468)
	at deployment.test-SNAPSHOT.war//javax.faces.component.UIComponent.visitTree(UIComponent.java:1468)
	at deployment.test-SNAPSHOT.war//javax.faces.component.UIComponent.visitTree(UIComponent.java:1468)
	at deployment.test-SNAPSHOT.war//javax.faces.component.UIForm.visitTree(UIForm.java:355)
	at deployment.test-SNAPSHOT.war//javax.faces.component.UIComponent.visitTree(UIComponent.java:1468)
	at deployment.test-SNAPSHOT.war//javax.faces.component.UIComponent.visitTree(UIComponent.java:1468)
	at deployment.test-SNAPSHOT.war//com.sun.faces.context.PartialViewContextImpl.processComponents(PartialViewContextImpl.java:400)
	at deployment.test-SNAPSHOT.war//com.sun.faces.context.PartialViewContextImpl.processPartial(PartialViewContextImpl.java:261)
	at deployment.test-SNAPSHOT.war//org.primefaces.context.PrimePartialViewContext.processPartial(PrimePartialViewContext.java:65)
	at deployment.test-SNAPSHOT.war//javax.faces.component.UIViewRoot.processDecodes(UIViewRoot.java:1008)
	at deployment.test-SNAPSHOT.war//com.sun.faces.lifecycle.ApplyRequestValuesPhase.execute(ApplyRequestValuesPhase.java:55)
	at deployment.test-SNAPSHOT.war//com.sun.faces.lifecycle.Phase.doPhase(Phase.java:76)
	at deployment.test-SNAPSHOT.war//com.sun.faces.lifecycle.LifecycleImpl.execute(LifecycleImpl.java:177)
	at deployment.test-SNAPSHOT.war//javax.faces.webapp.FacesServlet.executeLifecyle(FacesServlet.java:707)
	at deployment.test-SNAPSHOT.war//javax.faces.webapp.FacesServlet.service(FacesServlet.java:451)
```

```XHTML
<p:dataTable id="datatable" value="#{dataTable018.dataModel}" var="lang" paginator="true" rows="3"
                         filteredValue="#{dataTable018.filteredProgLanguages}" filterEvent="enter" rowKey="#{lang.id}" selectionMode="single"
                         rowsPerPageTemplate="2,3,5,10">
    <p:column headerText="ID" sortBy="#{lang.id}">
        <h:outputText value="#{lang.id}"/>
    </p:column>
    <p:column headerText="Name" sortBy="#{lang.name}" filterBy="#{lang.name}" filterMatchMode="contains">
        <h:outputText value="#{lang.name}"/>
    </p:column>
    <p:column headerText="First appeared" sortBy="#{lang.firstAppeared}" filterBy="#{lang.firstAppeared}" filterMatchMode="gte">
        <h:outputText value="#{lang.firstAppeared}"/>
    </p:column>
    <p:column headerText="OS" field="mainOs.nameShort"/>
</p:dataTable>
```

melloware:
@Rapster here is @robtha 's patch applied...

### DIFF
--- a/src/main/java/org/primefaces/component/datatable/DataTable.java
+++ b/src/main/java/org/primefaces/component/datatable/DataTable.java
@@ -141,26 +141,25 @@ public class DataTable extends DataTableBase {
     public static final String SMALL_SIZE_CLASS = "ui-datatable-sm";
     public static final String LARGE_SIZE_CLASS = "ui-datatable-lg";
 
-    public static final Map<DataTableFeatureKey, DataTableFeature> FEATURES = MapBuilder.<DataTableFeatureKey, DataTableFeature>builder( new LinkedHashMap<>() )
-            .put(DataTableFeatureKey.PAGE, new PageFeature())
-            .put(DataTableFeatureKey.SCROLL, new ScrollFeature())
-            .put(DataTableFeatureKey.FILTER, new FilterFeature())
-            .put(DataTableFeatureKey.SORT, new SortFeature())
-            .put(DataTableFeatureKey.SELECT, new SelectionFeature())
+    public static final Map<DataTableFeatureKey, DataTableFeature> FEATURES = MapBuilder.<DataTableFeatureKey, DataTableFeature>builder()
             .put(DataTableFeatureKey.DRAGGABLE_COLUMNS, new DraggableColumnsFeature())
+            .put(DataTableFeatureKey.FILTER, new FilterFeature())
+            .put(DataTableFeatureKey.PAGE, new PageFeature())
+            .put(DataTableFeatureKey.SORT, new SortFeature())
             .put(DataTableFeatureKey.RESIZABLE_COLUMNS, new ResizableColumnsFeature())
+            .put(DataTableFeatureKey.SELECT, new SelectionFeature())
             .put(DataTableFeatureKey.ROW_EDIT, new RowEditFeature())
             .put(DataTableFeatureKey.CELL_EDIT, new CellEditFeature())
             .put(DataTableFeatureKey.ROW_EXPAND, new RowExpandFeature())
+            .put(DataTableFeatureKey.SCROLL, new ScrollFeature())
             .put(DataTableFeatureKey.DRAGGABLE_ROWS, new DraggableRowsFeature())
             .put(DataTableFeatureKey.ADD_ROW, new AddRowFeature())
             .build();
 
-    private static final Map<String, Class<? extends BehaviorEvent>> BEHAVIOR_EVENT_MAPPING =
-        MapBuilder.<String, Class<? extends BehaviorEvent>>builder( new LinkedHashMap<>() )
+    private static final Map<String, Class<? extends BehaviorEvent>> BEHAVIOR_EVENT_MAPPING = MapBuilder.<String, Class<? extends BehaviorEvent>>builder()
             .put("page", PageEvent.class)
-            .put("filter", FilterEvent.class)
             .put("sort", SortEvent.class)
+            .put("filter", FilterEvent.class)
             .put("rowSelect", SelectEvent.class)
             .put("rowUnselect", UnselectEvent.class)
             .put("rowEdit", RowEditEvent.class)
@@ -1074,11 +1073,5 @@ public class DataTable extends DataTableBase {
     @Override
     public void setWidth(String width) {
         getStateHelper().put(InternalPropertyKeys.width, width);
-    }
-
-    public boolean hasActiveFilter() {
-        // Empty Global Filter contains empty string
-        return getFilterByAsMap().values().stream().filter( f -> f.getFilterValue() != null && !LangUtils.isValueEmpty( f.getFilterValue().toString() ) )
-                                 .findFirst().isPresent();
     }
 }
--- a/src/main/java/org/primefaces/component/datatable/DataTable.java
+++ b/src/main/java/org/primefaces/component/datatable/DataTable.java
@@ -141,7 +141,8 @@ public class DataTable extends DataTableBase {
     public static final String SMALL_SIZE_CLASS = "ui-datatable-sm";
     public static final String LARGE_SIZE_CLASS = "ui-datatable-lg";
 
-    public static final Map<DataTableFeatureKey, DataTableFeature> FEATURES = MapBuilder.<DataTableFeatureKey, DataTableFeature>builder()
+    public static final Map<DataTableFeatureKey, DataTableFeature> FEATURES =
+                MapBuilder.<DataTableFeatureKey, DataTableFeature>builder(new LinkedHashMap<DataTableFeatureKey, DataTableFeature>())
             .put(DataTableFeatureKey.DRAGGABLE_COLUMNS, new DraggableColumnsFeature())
             .put(DataTableFeatureKey.FILTER, new FilterFeature())
             .put(DataTableFeatureKey.PAGE, new PageFeature())
@@ -1074,4 +1075,5 @@ public class DataTable extends DataTableBase {
     public void setWidth(String width) {
         getStateHelper().put(InternalPropertyKeys.width, width);
     }
+
 }
--- a/src/main/java/org/primefaces/component/datatable/DataTable.java
+++ b/src/main/java/org/primefaces/component/datatable/DataTable.java
@@ -141,25 +141,26 @@ public class DataTable extends DataTableBase {
     public static final String SMALL_SIZE_CLASS = "ui-datatable-sm";
     public static final String LARGE_SIZE_CLASS = "ui-datatable-lg";
 
-    public static final Map<DataTableFeatureKey, DataTableFeature> FEATURES = MapBuilder.<DataTableFeatureKey, DataTableFeature>builder()
-            .put(DataTableFeatureKey.DRAGGABLE_COLUMNS, new DraggableColumnsFeature())
-            .put(DataTableFeatureKey.FILTER, new FilterFeature())
+    public static final Map<DataTableFeatureKey, DataTableFeature> FEATURES = MapBuilder.<DataTableFeatureKey, DataTableFeature>builder( new LinkedHashMap<>() )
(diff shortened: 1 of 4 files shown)


src/main/java/org/primefaces/component/datatable/DataTable.java
```diff
@@ -141,25 +141,26 @@
...
             .put(DataTableFeatureKey.CELL_EDIT, new CellEditFeature())
             .put(DataTableFeatureKey.ROW_EXPAND, new RowExpandFeature())
-            .put(DataTableFeatureKey.SCROLL, new ScrollFeature())
             .put(DataTableFeatureKey.DRAGGABLE_ROWS, new DraggableRowsFeature())
             .put(DataTableFeatureKey.ADD_ROW, new AddRowFeature())
             .build();
 
-    private static final Map<String, Class<? extends BehaviorEvent>> BEHAVIOR_EVENT_MAPPING = MapBuilder.<String, Class<? extends BehaviorEvent>>builder()
+    private static final Map<String, Class<? extends BehaviorEvent>> BEHAVIOR_EVENT_MAPPING =
+        MapBuilder.<String, Class<? extends BehaviorEvent>>builder( new LinkedHashMap<>() )
```
Rapster: I don't think order matters here, does it?

src/main/java/org/primefaces/component/datatable/DataTable.java
```diff
@@ -1074,4 +1075,10 @@ public String getWidth() {
     public void setWidth(String width) {
         getStateHelper().put(InternalPropertyKeys.width, width);
     }
+
+    public boolean hasActiveFilter() {
```
Rapster: Introduce in UITable instead:
```java
    default boolean isFilteringCurrentlyActive() {
        return getFilterByAsMap().values().stream().anyMatch(FilterMeta::isActive);
    }
```

src/main/java/org/primefaces/component/datatable/feature/FilterFeature.java
```diff
@@ -133,62 +133,67 @@ public void encode(FacesContext context, DataTableRenderer renderer, DataTable t
     }
 
     public void filter(FacesContext context, DataTable table) {
```
Rapster, line 135: Check with @tandraschko and @melloware for this part, it's been recently modified and I didn't follow those changes

src/main/java/org/primefaces/component/datatable/DataTable.java
robtha: actually not necessary for the fix. the idea is to get a predictable order for processing features and to process features that might modify table data first.

src/main/java/org/primefaces/component/datatable/feature/FilterFeature.java
robtha: actually minor change. filtering is processed only if filter values entered.
looks a lot because identation has changed.

src/main/java/org/primefaces/component/datatable/DataTable.java
robtha: used FilterMeta::isActive before but changed because it did not work if global filter has been specified. e.g. showcase sample always has an empty string as filterValue therefore isActive returns true.

melloware:
@Rapster do you wan to hack this PR?  I submitted it so we all could see it.

src/main/java/org/primefaces/component/datatable/DataTable.java
melloware: @robtha I think Rapster is saying the LinkedhashMap is good for the Features but not necessary for BehaviorEvents.

robtha: @melloware I have to admit that I didn't really check if the order of BehaviourEventS has any impact.

melloware:
@robtha maybe you could provide me another patch file with only the bare minimum changes needed to be made and no formatting changes.  that would make this PR also easier to review!

robtha:
@melloware Patch files shows changes even if only indentation on code block has been changed. I am not aware of any option to ignore
whitespace changes from a patch.
To keep changes small I must not ident code properly.
But with incorrect indentation a successfull build will be prevented by checkstyle plugin.

I have applied 2 patch files. 
FilterFeature_noindent.patch.txt: With incorrect indentation so you can see that there are very view changes to the code FilterFeature.patch.txt: With correct indentation that passes checkstyle plugin.

[FilterFeature.patch.txt](https://github.com/primefaces/primefaces/files/5937249/FilterFeature.patch.txt)
[FilterFeature_noindent.patch.txt](https://github.com/primefaces/primefaces/files/5937250/FilterFeature_noindent.patch.txt)

melloware:
OK I applied your minimum patch and tweaked it so its fail fast instead of wrapping that whole block in an IF statement so now its much more readable and I think still accomplishes the same thing.

src/main/java/org/primefaces/component/datatable/DataTable.java
melloware: @Rapster So the question is should the UITable method change to be the one here?

melloware:
@Rapster Rob also recommended returning NULL instead of empty ArrayList() but I am not so sure about that. I know we typically like to return empty lists and not NULL.  Thoughts?

src/main/java/org/primefaces/util/MapBuilder.java
```diff
@@ -45,7 +45,7 @@ private MapBuilder(Map<K, V> map) {
     }
 
     public static <K, V> MapBuilder<K, V> builder() {
-        return new MapBuilder<>(new HashMap<K, V>());
+        return new MapBuilder<>(new LinkedHashMap<K, V>());
```
melloware: @Rapster any harm in making the default MapBuilder always use a LinkedHashMap so the values are always stored in the order they are put in the map?

Rapster: LinkedHashMap will take up more memory, I don't think it should be default. Instead introducing two 2 methods hash and linked would be more practical, and remove builder methods

melloware: OK I created two methods `hashed()` and `linked()` but I left `builder()` just call hashed() for backwards compatibility

src/main/java/org/primefaces/component/datatable/DataTable.java
Rapster: Sorry I'm really confused about all the modifications that has been made on DT for the last couple of months, and also quite surprised about issues users are facing... As of now, I'm just trying to give little help on code review, I mentionned UITable since TreeTable should have same functionnalities

melloware: I guess I am asking that @robtha is saying your currently `isFilteringCurrentlyActive` in UITable is not complete enough and his above detects empty global filters etc.  I am asking should his method replace `isFilterCurrentlyActive` in UITable?

Rapster: Then, the problem is in FilterMeta#isActive then? But that's strange... Empty string should be set null by FilterFeature. For now, isFilteringCurrentlyActive should be introduced with @robtha impl, and it will most probably be reused for issue like #6913

tandraschko:
no idea about the PR in general, it feels a bit hacky. Up to @Rapster.

Just dont like the 2 new method of the MapBuilder. I think we are more flexible with the old way (we also have e.g. a LimitedSizeHashMap), also the methodnames sounds weird (linked must actually be linkedHash then)

melloware:
OK changes made...

1. UITable.isFilteringCurrentlyActive is created and is combination of Rapster and Robtha suggestions. Please review it.
2. MapBuilder put back the way it was and just FEATURES made a LinkedHashMap

Rapster:
I'm having a look, the PR looks hacky indeed as said Thomas. Though, MapBuilder looks good, MapBuilder is not meant to be flexible, it's meant to be convenient and easy to init.

PR #6959, opened 2021-02-07
Fix #6953: Refactor Global Filter

### LINKED ISSUE #6953: DataTable: Global FilterFunction Broken
**Describe the defect**
Working on a showcase example to put back the old GlobalFilterFunction and the functionality has changed.

```xml
                <p:dataTable var="customer" value="#{dtFilterView.customers2}" widgetVar="customersTable2"
                    filteredValue="#{dtFilterView.filteredCustomers2}" filterBy="#{dtFilterView.filterBy}"
                    globalFilterFunction="#{dtFilterView.globalFilterFunction}">

                    <f:facet name="header">
                        <div class="p-text-left">
                            <h:outputText value="Search all fields using globalFilterFunction:" />
                            <p:inputText id="globalFilter" onkeyup="PF('customersTable2').filter()" style="width:150px"
                                placeholder="Enter keyword" />
                        </div>
                    </f:facet>
```

```java
 public boolean globalFilterFunction(Object value, Object filter, Locale locale) {
        String filterText = (filter == null) ? null : filter.toString().trim().toLowerCase();
        if (LangUtils.isValueBlank(filterText)) {
            return true;
        }
        int filterInt = getInteger(filterText);

        Customer customer = (Customer) value;
        return customer.getName().toLowerCase().contains(filterText)
                || customer.getCountry().getName().toLowerCase().contains(filterText)
                || customer.getRepresentative().getName().toLowerCase().contains(filterText)
                || customer.getStatus().name().toLowerCase().contains(filterText)
                || customer.getActivity() < filterInt;
    }
```

its expecting the `value` to be the Customer POJO and right now it is just passing it each individual column value as a String which causes `java.lang.ClassCastException: java.lang.String cannot be cast to org.primefaces.showcase.domain.Customer`.

The old code was passing the table.RowData();
```java
if (hasGlobalFilter && globalFilterFunction != null) {
                globalMatch = (Boolean) globalFilterFunction.invoke(elContext, new Object[]{table.getRowData(), globalFilterValue, filterLocale});
            }
```

But the new code is passing each individual column.

melloware:


tandraschko:
Still dont like the code for 100%. Shouldt we maybe store the function in the meta and access this field instead of get via DT all the time? Otherwise we should remove the method param.
Therefore we need reviews and discussions before merge PRs.

melloware:
You are right. This puts back the original intent and removes `table.getGlobalFilterFunction`. It makes the code a little less readable checking its instanceof but does accomplish your task of no dead code and not using the table.

src/main/java/org/primefaces/model/filter/GlobalFilterConstraint.java
```diff
@@ -23,16 +23,9 @@
...
-
 public class GlobalFilterConstraint extends ContainsFilterConstraint {
 
     public GlobalFilterConstraint() {
     }
 
-    @Override
-    public boolean isMatching(FacesContext ctxt, Object value, Object filter, Locale locale) {
-        return super.isMatching(ctxt, value, filter, locale);
-    }
```
melloware, line 37: Sonar recommended removing this method as its just overriding and calling SUPER.  Removed and verified all integration tests are still working.

PR #6966, opened 2021-02-09
Fix #6964: UIRepeat not removing var from context

### LINKED ISSUE #6964: p:repeat: fix Mojarra ui:repeat does not reset EL var correctly
**Describe the defect**
The "resetValues=true" attribute + UIRepeat behave not as expected (please check the attached application and the code snippet below)

**Reproducer**
primefaces-test application is attached
[primefaces-test-master.zip](https://github.com/primefaces/primefaces/files/5946677/primefaces-test-master.zip)

**Environment:**
 - PF Version: 10.0.0
 - JSF + version: Mojarra (all versions)
 - Affected browsers: ALL_

**To Reproduce**
1. Go to the index page on attached primefaces-test application
2. Click on 'Reproduce bug' link

**Expected behavior**
I expect the variable #{value} stay empty

**Example XHTML**
```html
    <h:form>
        Expected to stay empty: #{value}
    </h:form>
    <h:form>
        <ui:repeat value="#{testView.strings}" var="value">
            <p:outputLabel value="#{value} "/>
        </ui:repeat>
    </h:form>

    <h:form>
        <!-- this link will update the variable value in the first form with value from the UIRepeat -->
        <p:commandLink value="Reproduce bug" process="@this" update="@(form)" resetValues="true"/>
        <!-- and this link works fine -->
        <p:commandLink value="Not reproduce bug" process="@this" update="@(form)" resetValues="false"/>
    </h:form>
```

**Example Bean**
```java
@Data
@Named
@ViewScoped
    
    private final List<String> strings = new ArrayList<>();

    @PostConstruct  
    public void init() {
        strings.addAll(Arrays.asList("Winona Lindsey", "Pandora Peay", "Honey Dean"));
    }

    public List<String> getStrings() {
        return strings;
    }
}
```

melloware:
@tandraschko I believe I fixed this in the correct place in the finally of visitTree.  

The other changes were just trailing dead whitespace that I am surpised CheckStyle was not complaining about.

I tested against the reproducer and the issue is fixed.  If you approve I will then submit a PR to Mojarra.

### DIFF
--- a/src/main/java/org/primefaces/component/repeat/UIRepeat.java
+++ b/src/main/java/org/primefaces/component/repeat/UIRepeat.java
@@ -349,6 +349,18 @@ public class UIRepeat extends UINamingContainer {
         }
     }
 
+    private void clearValue(FacesContext ctx) {
+        if (this.var != null || this.varStatus != null) {
+            Map<String,Object> attrs = ctx.getExternalContext().getRequestMap();
+            if (this.var != null) {
+                attrs.remove(this.var);
+            }
+            if (this.varStatus != null) {
+                attrs.remove(this.varStatus);
+            }
+        }
+    }
+
     private Map<String,SavedState> childState;
 
     private Map<String,SavedState> getChildState() {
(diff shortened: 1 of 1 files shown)


tandraschko:
looks good, +1

PR #6996, opened 2021-02-16
Fix #6834: Datatable checkbox event fixes

### LINKED ISSUE #6834: Checkbox: Regression, custom styling of checkbox active state no longer working
**Describe the defect**
The custom styling for a checkbox actiove state is no longer working.  I am no CSS expert, but I had to add an additional attribute selector to bring it back into working order.   

**Reproducer**
n/a

**Environment:**
 - PF Version: 10.0 Snapshot
 - JSF + version: _e.g. Mojarra 2.3.9

**To Reproduce**

**Expected behavior**

// PF 8.0
```css
.ui-chkbox-box.ui-state-active {
    border-color: #007ad9;
    background-color: #007ad9;
    color: #000;
}
```

// PF 10.0
```css
.ui-chkbox-box.ui-state-active, .ui-chkbox-box[aria-checked='true'] {
    border-color: #007ad9;
    background-color: #007ad9;
    color: #000;
}
```

melloware:
OK so this was broken in a couple of ways...

1. No `click` event on the checkbox itself so it was only adding `active` style AFTER you left the checkbox in blur.
2. Events were not specific enough so when i added off then on `click.dataTable` it remove ALL click datatable events so had to name the checkbox and checkboxALL events specifically so only they would get removed and not all click events on the table.
3. For some odd reason `focus` of the checkbox was removing the `active` style which makes absolutely no sense.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
+++ b/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
@@ -1125,20 +1125,23 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
         }
 
         //keyboard support
-        this.tbody.off('focus.dataTable blur.dataTable change.dataTable', checkboxSelector)
-                    .on('focus.dataTable', checkboxSelector, null, function() {
+        this.tbody.off('focus.dataTableCheck blur.dataTableCheck change.dataTableCheck click.dataTableCheck', checkboxSelector)
+                    .on('focus.dataTableCheck', checkboxSelector, null, function() {
                         var input = $(this);
-
-                        if(input.attr('aria-checked') === "true" || input.prop('checked')) {
-                            input.removeClass('ui-state-active');
-                        }
 
                         input.addClass('ui-state-focus');
 
                         $this.focusedRow = input.closest('.ui-datatable-selectable');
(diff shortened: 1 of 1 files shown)


src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
```diff
@@ -1125,20 +1125,23 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
...
         //keyboard support
-        this.tbody.off('focus.dataTable blur.dataTable change.dataTable', checkboxSelector)
-                    .on('focus.dataTable', checkboxSelector, null, function() {
+        this.tbody.off('focus.dataTableCheck blur.dataTableCheck change.dataTableCheck click.dataTableCheck', checkboxSelector)
+                    .on('focus.dataTableCheck', checkboxSelector, null, function() {
                         var input = $(this);
 
-                        if(input.attr('aria-checked') === "true" || input.prop('checked')) {
-                            input.removeClass('ui-state-active');
-                        }
```
melloware, line 1134: This part above makes absolutely no sense to remove the active state just because the checkbox has focus.

cagataycivici:
We'll just keep ui-state-active.

Issue #6978, opened 2021-02-11
Improve registered events on components


mertsincan:
For example, if a component is using the PrimeFaces.utils.registerScrollHandler to close its panel, the scroll event is called even if the panel is closed. Such events can be registered when the panel is visible.

https://user-images.githubusercontent.com/11868120/107623300-2ea79580-6c6a-11eb-836e-2c2cf7894f90.mov

- [x] AutoComplete
- [x] CascadeSelect
- [x] Datepicker
- [x] Password
- [x] SelectCheckboxMenu
- [x] SelectOneMenu
- [x] ThemeSwitcher
- [x] SplitButton
- [x] OverlayPanel
- [x] Menu
- [x] MenuButton
- [x] SlideMenu
- [x] TieredMenu
- [x] ConfirmPopup
- [x] ContextMenu
- [x] ColorPicker
- [x] Keyboard

tandraschko:
we currently do this in each widget itself, see dialog.js ln784

mertsincan:
Yes, but, I could not find unbind operations in components. The scroll event is always called. I think this will cause performance problems.

tandraschko:
registerScrollHandler should remove itself on widget destroy

EDIT: it already does, see core.utils.js lin 384

tandraschko:
Or do you like to remove/add the scroll handler when open/close a overlay?
its better of course but more "maintainence" in each widget

mertsincan:
@tandraschko
https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/forms/forms.password.js#L137

https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/forms/forms.password.js#L154

tandraschko:
yep, thats cool

melloware:
@mertsincan I always wondered about this since DatePicker registers and unregisters on popup open and close but I noticed we weren't doing that anywhere else in PF.  This is a good improvement!

Issue #6973, opened 2021-02-10
Improve animation performance on overlay panels


mertsincan:
Remove show/hide animation of JqueryUI from the connected overlay panel. They can use CSS animations.

- [x] AutoComplete (remove 'effect' and 'effectDuration' attributes)
- [x] CascadeSelect
- [x] Datepicker
- [x] Password
- [x] SelectCheckboxMenu
- [x] SelectOneMenu (remove 'effect' and 'effectSpeed' attributes)
- [x] ThemeSwitcher (remove 'effect' and 'effectSpeed' attributes)
- [x] SplitButton
- [x] OverlayPanel (remove 'showEffect' and 'hideEffect' attributes)
- [x] Menu (overlay mode)
- [x] MenuButton
- [x] SlideMenu (overlay mode)
- [x] TieredMenu (overlay mode)
- [x] ConfirmPopup (remove 'showEffect' and 'hideEffect' attributes)
- [x] ContextMenu
---

This feature will not be added as the following components use a 3rd party API.
- [ ] ColorPicker
- [ ] Keyboard

christophs78:
IMO there´s a regression with lazy SelectOneMenu. (reproduceable in showcase; ui/input/oneMenu.xhtml)
When you first click on the Lazy one it does not open. Please compare this to http://primefaces.org/showcase/ui/input/oneMenu.xhtml

melloware:
I think DatePicker might have something similar going on.

christophs78:
Overall this is a really good improvement!
We "only" should avoid breaking existing usecases. Together we get this done. :-)

mertsincan:
> IMO there´s a regression with lazy SelectOneMenu. (reproduceable in showcase; ui/input/oneMenu.xhtml)
When you first click on the Lazy one it does not open. Please compare this to http://primefaces.org/showcase/ui/input/oneMenu.xhtml

@christophs78, it works fine for me.  Which browser are you using?

christophs78:
> > IMO there´s a regression with lazy SelectOneMenu. (reproduceable in showcase; ui/input/oneMenu.xhtml)
> > When you first click on the Lazy one it does not open. Please compare this to http://primefaces.org/showcase/ui/input/oneMenu.xhtml
> 
> @christophs78, it works fine for me. Which browser are you using?

Edge 88; i´ll recheck this evening; currently (very) busy on my day-job

christophs78:
[...]

PR #7006, opened 2021-02-18
Fix #7004: Refactor resolveAppendTo

### LINKED ISSUE #7004: DatePicker: JS-Error in destroy and refresh
**Describe the defect**
When the DatePicker widget should be removed during `PrimeFaces.ajax.Response.destroyDetachedWidgets` it will throw a JS error because of an illegal usage of a utililty method.

grafik

```Javascript
// ... code from refresh
PrimeFaces.utils.cleanupDynamicOverlay(this, this.panel, this.id + '_panel', this.cfg.appendTo);

// ... code from destroy
PrimeFaces.utils.removeDynamicOverlay(this, null, this.id + '_panel', this.cfg.appendTo);
```

In both cases `this.cfg.appendTo` is a string - per default `@(body)`. But both utility methods expect a JQuery instance. (see [here](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/core/core.utils.js#L83))

**Environment:**
 - PF Version: 10.0.0-RC1
 - Affected browsers: ALL

melloware:
@tandraschko refactored to pass in panel and fixed the DatePicker cleanup overlay issue

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
+++ b/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
@@ -103,7 +103,7 @@ PrimeFaces.widget.AutoComplete = PrimeFaces.widget.BaseWidget.extend({
         this.cfg.ariaEmptyMessage = this.cfg.emptyMessage||'No search results are available.';
         this.cfg.dropdownMode = this.cfg.dropdownMode||'blank';
         this.cfg.autoHighlight = (this.cfg.autoHighlight === undefined) ? true : this.cfg.autoHighlight;
-        this.cfg.appendTo = PrimeFaces.utils.resolveAppendTo(this);
+        this.cfg.appendTo = PrimeFaces.utils.resolveAppendTo(this, this.panel);
         this.cfg.myPos = this.cfg.myPos||'left top';
         this.cfg.atPos = this.cfg.atPos||'left bottom';
         this.cfg.active = (this.cfg.active === false) ? false : true;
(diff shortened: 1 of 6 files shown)


tandraschko:
thanks!

vonnai:
hi, why should be overlay panel scrolled together with content when dialog fitViewPort = true? I expect that the overlay panel is always fixed to its input component. After your fix, there is not any solution to how to stop scrolling of overlay panel with dialog content together. If I set appendTo e.g. :dialogFormId:dialogContentId, it will be ignored and will be appended to the body. Thanks for you reply.

image

melloware:
@vonnai have you tried 10.0.0RC2?  I believe this is already fixed.

vonnai:
@melloware I have not tried yet, but as I see there are not any changes in selectonemnu.js. The original appendTo method has been only moved to another lib class, haven't been? The code is the same as in 8.0.3 for forms.selectonemu.js. Why is overlay panel appended to the body if it's inside dialog?

```
if(!overlay.parent().is(document.body)) {
                    widget.cfg.appendTo = "@(body)"
                    return widget.cfg.appendTo;
                }
```

melloware:
@vonnai there were 3 major fixes done for 10.0.0.RC2 I am begging you to try it.  Trust me its fixed.

https://github.com/primefaces/primefaces/issues/6978

https://github.com/primefaces/primefaces/issues/6973

vonnai:
@melloware ok thanks, I'll try it.

PR #7010, opened 2021-02-18
Fix for #7008 - DataScroller with SelectBooleanCheckbox: Initial checked value is overwritten for loaded chunks during first decode

### LINKED ISSUE #7008: DataScroller with SelectBooleanCheckbox: Initial checked value is overwritten for loaded chunks during first decode
**Describe the defect**
If I create a datascroller with a list of items, that have a boolean value that is initially true and bound to a p:selectBooleanCheckbox within the datascroller, the checkbox is checked for entries that are initially dispayed (ok) but for additionally loaded elements due to scrolling the checkbox is not checked anymore.

**Reproducer**
[primefaces-test.zip](https://github.com/primefaces/primefaces/files/6003830/primefaces-test.zip)

**Environment:**
 - PF Version: _10.0.0.RC1_
 - Affected browsers: ALL

**To Reproduce**
1. Go to test page in attached primefaces-test
2. Scroll down until additional entries are loaded
4. See error (additional entry index>5 have an unchecked checkbox)

**Expected behavior**
All checkboxes should be checked according to data in Model.

Possible Reason for that:
The problem is related to the fact that in **SelectBooleanCheckboxRenderer.decode** and submittedValue of null is interpreted as "false" and set on the underlying model. For lazy loaded chunk this seems to be executed during the ajax loadChunks call and since the new elements were just yet added their is not value in the requestparameter map for them.
Other Input Forms (e.g inputtext) for fine because their just skip the updatemodel phase if submited value == null.

**Example XHTML**
```html
<h:form id="frmTest">
   	<p:dataScroller var="_item" value="#{testView.model}" chunkSize="5" >
   		<div style="height:300px">
   			<p:inputText value="#{_item.text}" />
   			<p:selectBooleanCheckbox value="#{_item.check}"/> This checkbox should always be checked, but as soon as new chunk is loaded it is not.
   		</div>
   	</p:dataScroller>
</h:form>
```

**Example Bean**
```java
@Data
@Named
@ViewScoped

	private ListDataModel<Item> model = new ListDataModel<>();

	@Data
	public static class Item implements Serializable {

		private String text;
		private boolean check=true;

		public Item(String text) {
			super();
			this.text = text;
		}
	}

	@PostConstruct
	public void init() {
		List<Item> list = new ArrayList<>();
		for (int i =0; i<50; i++) {
			list.add(new Item("Text "+i));
		}
		model = new ListDataModel<>(list);
	}

	public ListDataModel<Item> getModel() {
		return model;
	}

}
```

gako:
processing of children is skipped if the current request is a loadrequest and rowindex is largetr then offset (this is a new child that was added in this loadrequest)
This fix should work for all kind of children that were prone to this problem, since the decode method of the child is never called.

### DIFF
--- a/src/main/java/org/primefaces/component/api/UIData.java
+++ b/src/main/java/org/primefaces/component/api/UIData.java
@@ -178,6 +178,10 @@ public class UIData extends javax.faces.component.UIData {
         }
     }
 
+    protected boolean shouldProcessChild(FacesContext context, int rowIndex, PhaseId phaseId ) {
+        return true;
+    }
+
     protected void processChildren(FacesContext context, PhaseId phaseId) {
         int first = getFirst();
         int rows = getRows();
@@ -190,6 +194,10 @@ public class UIData extends javax.faces.component.UIData {
 
             if (!isRowAvailable()) {
                 break;
+            }
+
+            if (!shouldProcessChild(context, rowIndex, phaseId)) {
+                continue;
             }
 
             if (iterableChildren == null) {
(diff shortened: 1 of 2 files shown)


tandraschko:
looks good for me, thanks!

PR #7013, opened 2021-02-19
Fix #7012: Slider add min max range functionality

### LINKED ISSUE #7012: Slider: min max range styles missing
image
Jquery UI Slider have 4 range mode and we are supporting only 2 of them. Also I think we need "min" range by default.

buunsal:
image
I also commited new demos to showcase. Basic one not acceptable and contrary to what we use in our other ui libraries.
Primeng sample: [Primeng Slider](https://primefaces.org/primeng/showcase/#/slider)

But i need code review.  It could be written better maybe.

### DIFF
--- a/src/main/java/org/primefaces/component/slider/SliderRenderer.java
+++ b/src/main/java/org/primefaces/component/slider/SliderRenderer.java
@@ -112,16 +112,14 @@ public class SliderRenderer extends CoreRenderer {
                 .callback("onSlide", "function(event,ui)", slider.getOnSlide())
                 .callback("onSlideEnd", "function(event,ui)", slider.getOnSlideEnd());
 
-        if ("true".equals(range)) {
-            wb.attr("range", true);
-        }
-
-        else if ("false".equals(range)) {
-            wb.attr("range", false);
-        }
-
-        else {
-            wb.attr("range", range);
+        switch (range) {
+            case "true":
+            case "false":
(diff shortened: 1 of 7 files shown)


src/main/java/org/primefaces/component/slider/SliderRenderer.java
```diff
@@ -106,13 +106,24 @@ protected void encodeScript(FacesContext context, Slider slider) throws IOExcept
...
+        }
+
+        else if ("false".equals(range)) {
+            wb.attr("range", false);
+        }
+
+        else {
+            wb.attr("range", range);
+        }
+
```
melloware, line 124: How about this instead:
```java
                switch (range) {
		case "true":
		case "false":
			 wb.attr("range", Boolean.valueOf(range));
			break;
		default:
			wb.attr("range", range);
			break;
		}
```

melloware (changes requested):

src/main/java/org/primefaces/component/slider/SliderRenderer.java
buunsal: Ok for me. I added your code.

tandraschko:
you also may adjust the taglib xml

melloware:
And \docs\migrationguide\10_0_0.md

buunsal:
added

melloware (approved):

PR #7017, opened 2021-02-20
Fix #10/#7018 Treetable exporter (based on Datatable exporter)

### LINKED ISSUE #10: DataExporter for TreeTable
Currently DataExporter only supports DataTable, add ability to export TreeTable as well.

### LINKED ISSUE #7018: TreeTable: Add Support for new DataExporter
**Description**
At the moment only Datatable is implemented with the _new_ DataExporter Implementation

Also support exporting TreeTables with the new generic DataExporter

See PR for possible Solution, atm its more or less a 1:1 copy of the DataTable Exporter with adjustments for the TreeTable node iteration.
Maybe some common functionality could be merged, but then they would no longer be independent, and I would have needed to change the datatable functionality as well, which I would rather not do. So atm there is a bit of copied/duplicated code.

see Pull Request at: https://github.com/primefaces/primefaces/pull/7017

```

### PR DESCRIPTION
**Pull request by gako**
I also fixed a minor issue with cloning nodes in TreeTable, which was impossible if a custom class extended CheckboxTreeNode or DefaultTreeNode.

### DIFF
--- a/src/main/java/org/primefaces/component/export/DataExporter.java
+++ b/src/main/java/org/primefaces/component/export/DataExporter.java
@@ -41,7 +41,10 @@ import javax.faces.event.ActionEvent;
 import javax.faces.event.ActionListener;
 
 import org.primefaces.PrimeFaces;
+import org.primefaces.component.datatable.DataTable;
 import org.primefaces.component.datatable.export.DataTableExporterFactory;
+import org.primefaces.component.treetable.TreeTable;
+import org.primefaces.component.treetable.export.TreeTableExporterFactory;
 import org.primefaces.expression.SearchExpressionFacade;
 import org.primefaces.util.ComponentUtils;
 import org.primefaces.util.Constants;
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by melloware** [comment]
This also fixes ticket #10

**Comment by tandraschko** [comment]
@gako can you also provide a showcase PR?
@melloware can you retest?

otherwise +1, looks ok

**Comment by melloware** [comment]
@gako I will add the Showcase PR I will write it while I am testing this PR out..

**Comment by tandraschko** [comment]
i also fix the cloning in a own ticket: https://github.com/primefaces/primefaces/issues/7061

**Comment by melloware** [comment]
@gako can you rebase this PR looks like you have some merge conflicts now

**Comment by melloware** [comment]
Tested this and now merging so I can add a Showcase PR.

**Comment by tandraschko** [comment]
@gako why the new Exporter uses invokeOnClosestTreeTableParent instead of invokeOnClosestIteratorParent?

**Comment by tandraschko** [comment]
i think this whole stuff needs a bit refactoring
the DataTableVisitCallBack, parent iterator, counting, exporting looks bit weird

why do we need to visit the parent iterating component? why not for treetable (it looks for the parent treetable)?
maybe someone could have a look

**Comment by melloware** [comment]
Good question I think @ravihariharan did the major refactoring when they broke the exporter up to allow it to be be per component.  Or not sure if it was always that way.

**Comment by tandraschko** [comment]
just look at this loop:

```
        for (TreeTable table : tables) {
            TreeTableVisitCallBack visitCallback = new TreeTableVisitCallBack(table, exportConfiguration, index);
            int nbTables = visitCallback.invoke(context);
            index += nbTables;
        }
```

why does it count tables in context/childen of a table.
just go through the code, TreeTableVisitCallBack will look for parent TreeTables and invoke. DataTable looks for parent ui:repeat/tabPanel or other stuff and invokes it.

Rapster:
I don't know about TreeTable impl, but that loop is there to support ui:repeat and multiple target tables references. Regarding the index, used for setting a new sheet/paragraph to clearly distinguish each tables

tandraschko:
@Rapster can you provide a simple test xhtml for the ui:repeat/multiple target case? Would be great. I would like to refactor this, visitTree should already do this.

Issue #3654, opened 2018-05-10
SelectOneMenu: after refreshing the widget in an AJAX request the popup opens at wrong position


movalz:
If an AJAX request updates a SelectOneMenu widget and user opens the popup of that SelectOneMenu the popup opens at wrong position: top left corner of the body instead of aligned to the widget.

Reason:
In PF 6.1 SelectOneMenu had following function, called from init():
```
appendPanel: function() {
        var container = this.cfg.appendTo ? PrimeFaces.expressions.SearchExpressionFacade.resolveComponentsAsSelector(this.cfg.appendTo): $(document.body);

        if(!container.is(this.jq)) {
            **container.children(this.panelId).remove();**
            this.panel.appendTo(container);
        }
    }
```
This removed the old popup panel ('..._panel') before appending the new panel.

In PF 6.2 this method was removed and instead init() contains following code:
`PrimeFaces.utils.registerDynamicOverlay(this, this.panel, this.id + '_panel');`
The passed argument "this.panel" is a two element list, because `this.panel = $(this.panelId)` finds the old and the new popup panel.  `PrimeFaces.utils.registerDynamicOverlay` doesn't work correctly with such a two element list: it calls `PrimeFaces.utils.appendDynamicOverlay` which does nothing because if `(!elementParent.is(appendTo) && !appendTo.is(overlay))` returns false for the two element `overlay` var.

## 1) Environment
- PrimeFaces version: 6.2.2
- Does it work on the newest released PrimeFaces version? No
- Does it work on the newest sources in GitHub? No
- Application server + version: Tomcat 8 (arbitrary)
- Affected browsers: IE, FF, Chrome (arbitrary)

## 2) Expected behavior
the popup panel opens aligned to the widget (as normal for a select one dropdown widget)

## 3) Actual behavior
the popup panel opens at the top left corner of the HTML body

## 4) Steps to reproduce
perform an AJAX call which updates the SelectOneMenu widget in it's response. Then open the SelectOneMenu's popup panel via clicking the widget. 

## 5) Sample XHTML
```
<p:selectOneMenu id="mySom" value="#{bean.mySom}" effect="none">
  <f:selectItems value="#{bean.selectItemsMySom}" />
</p:selectOneMenu>

<p:commandButton id="updateMySom" ajax="true" value="AJAX Action" action="#{bean.updateMySom}" process="@all" update="mySom" />
```

## 6) Sample bean
straight forward

tandraschko:
[...]

Issue #3178, opened 2018-01-18
Refactor overlay update logic


tandraschko:
Currently each dynamic overlay looks if there is already a element with the same id available under the appendTo target and removes this
since jquery 2 this is a little bit more complex and should be refactored and moved to common place:

e.g. overlaypanel:

        if(this.jq.length > 1) {
            $(document.body).children(this.jqId).remove();
            this.jq = $(this.jqId);
        }
        else {
            // this is required if the id does NOT contain a ':' - See #2485
            $(document.body).children("[id='" + this.id + "']").not(this.jq).remove();
        }

tandraschko:
First prototype commited. Probably the logic to prevent inputs on modality will be moved too.
Not sure about components like selectOne menue, were only a part is a dynamic overlay. Probably this will not extend DynamicOverlayWidget but used the util methods.

tandraschko:
main components refactored, need to adjust now components like selectOneMenu and autoComplete

PR #7024, opened 2021-02-21
Fix #7009: OverlayPanel/ColorPicker use resolveAppendTo

### LINKED ISSUE #7009: Overlay/ColorPicker: Refactor setupDialogSupport
**Description**
Refactor overlaypanel.js #setupDialogSupport to use `Primefaces.core.utils.resolveAppendTo`

only 1 DOM overlay left after update
all DOM overlay removed on widget destroy
both with custom appendTo (like @Form) and default.
That is really a tricky place an everything must work.

also see ColorPicker (may a bit different, no appendTo config support, we may refactor this even in another ticket)

melloware:
Retested OverlayPanel/ColorPicker used inside and outside of a dialog and everything is working.  Retested existing components that were using `resolveAppendTo` and they are also all working properly.

Let me know if you have any other scenarios you want me to test


src/main/resources/META-INF/resources/primefaces/colorpicker/1-colorpicker.js
```diff
@@ -192,13 +192,8 @@
...
      */
     setupDialogSupport: function() {
-        var dialog = this.jqEl[0].closest('.ui-dialog');
-        if (dialog) {
-            var $dialog = $(dialog);
-
-            if($dialog.length == 1 && $dialog.css('position') === 'fixed') {
-                this.overlay.css('position', 'fixed');
-            }
+        if(!this.cfg.appendTo) {
```
tandraschko: sure? so dialog support only works when appendTo is not configured by the user?

tandraschko:
There are in general some important basic we should cover by integration tests.
All are related that we never have duplicate overlays in the DOM, when updating containers, widgets are updated, widgets must handle that only 1 overlay is in the DOM.
Why does that happen?

A) 
1) widget is rendered initially, the container is updated and the widget not rendered again
2) widget must be destroyed and the destroy logic must remove the overlay, which may be e.g. directly inside the body (appendTo=@(body))

B)
1) widget is rendered initially, the container is updated and the widget is also updated
2) the old overlay might by a body child (appendTo=@(body)) and the new one is rendered inside the widget DOM
3) after widget init, the old overlay is removed and the new one is moved

src/main/resources/META-INF/resources/primefaces/colorpicker/1-colorpicker.js
melloware: No in the colorPicker case its only used in popup mode but Mert just refactored this code to bind the events on show and destroy them on hide. I am caching the original AppendTo value so we don't execute this code over and over every time the ColorPicker is show.  It is only needed once to correct the dialog CSS.

melloware:
I will write some Integration tests based on your use cases.

melloware:
I wrote this battery of OverlayPanel tests..
https://github.com/primefaces-extensions/primefaces-integration-tests/commit/7b2cff177f20fd85ca6bae9c84dc0d41c32c617b

OverlayPanel001Test - tests overlay panel not in a dialog and shows/hides/updates/destroys the component
OverlayPanel002Test - tests overlay panel inside a dialog and shows/hides/updates/destroys the component and tries to change the appendTo as well 

I think I have all your test cases covered?

melloware:
OK so my tests fail right now with the current code base:
```
Failures: 
Error:    OverlayPanel002Test.testDialogChangeAppendTo:136->assertConfiguration:144 expected: <@(body)> but was: <form:btnDestroy>
Error:  Errors: 
Error:    OverlayPanel001Test.testAppendToDialog:123 » NoSuchElement no such element: Un...
Error:    OverlayPanel001Test.testDestroy:106 » NoSuchElement no such element: Unable to...
```

I believe the reasonis the refactored code always tries to set to appendToBody if being used inside a dialog no matter what you have set.
```javascript
//append to body if not already appended by user choice
                if(!overlay.parent().is(document.body)) {
                    widget.cfg.appendTo = "@(body)";
                    return widget.cfg.appendTo;
                }
```
Basically it says if the overlayParent is not the document body make it the document body when using inside a dialog.  Which is what I am seeing locally.  But the legacy code was not doing that and I have a feeling that is not right because right now the legacy allows you to set the appendTo when inside a dialog? ` <@(body)> but was: <form:btnDestroy>`

melloware:
OK I got all the unit tests working for current PRD code and making sure panel was destroyed and no duplicates.

src/main/resources/META-INF/resources/primefaces/colorpicker/1-colorpicker.js
melloware: You are right i have to recalculate the color picker every time setupDialogSupport is called.

melloware:
So the one thing I don't understand is why does this component only destroy the dynamic overlay if the user did NOT add a custom `appendTo`?

tandraschko:
in general we dont need to destroy it if no appendTo is set and no default is set to .e.g body.
When the overlay is still in DOM below the "widget root", it will be destroyed automatically.

melloware:
AHHH OK I just learned something new.  Then OK my Integration Tests prove the current functionality and also now prove with my latest fixes that this new code does not change the behavior at all from the current implementation just cleans up the resolving of whether you are in a dialog or not.

src/main/resources/META-INF/resources/primefaces/overlaypanel/overlaypanel.js
```diff
@@ -55,6 +55,7 @@ PrimeFaces.widget.OverlayPanel = PrimeFaces.widget.DynamicOverlayWidget.extend({
         this.cfg.hideEvent = this.cfg.hideEvent||'click.ui-overlaypanel';
         this.cfg.dismissable = (this.cfg.dismissable === false) ? false : true;
         this.cfg.showDelay = this.cfg.showDelay || 0;
+        this.appendTo = this.cfg.appendTo;
```
tandraschko: i would do it like others here:
this.cfg.appendTo = PrimeFaces.utils.resolveAppendTo(...)

and remove the IF around removeDynamicOverlay
but please make sure that removeDynamicOverlay correclty skips "null" appendTo.
so we dont need always this If-statement in the widgets.

melloware: I made the changes you requested and re-ran all the integration tests and things are still passing as expected after this change.

src/main/resources/META-INF/resources/primefaces/overlaypanel/overlaypanel.js
```diff
@@ -105,9 +103,7 @@ PrimeFaces.widget.OverlayPanel = PrimeFaces.widget.DynamicOverlayWidget.extend({
         this._super();
 
         // see #setupDialogSupport
-        if (!this.cfg.appendTo) {
-            PrimeFaces.utils.removeDynamicOverlay(this, this.jq, this.id, $(document.body));
-        }
+        PrimeFaces.utils.removeDynamicOverlay(this, this.jq, this.id, $(document.body));
```
tandraschko: the last param is the "appendTo", which may be configured to e.g. someId.
I think we must use PrimeFaces.utils.resolveDynamicOverlayContainer(this) like in datepicker?

tandraschko: Also unsure if we need this as it extends DynamicOverlayWidget, which uses registerDynamicOverlay and auto-registers removeDynamicOverlay?

melloware: OK I added the methods for `resolve` and `refresh` now calls `cleanupDyanmicOverlay` instead of destroy just like DatePicker.  I think we should extend `DynamicOverlay` because that's exactly what this is right?

melloware: Good point Do we even need this since it does extend DynamicOverlay.  Let me try and pull those calls out and re-run the integration tests.

melloware: OK you are right I removed those calls in `refresh` and `destroy` and all integration tests passing.  So this is all handled in DynamicOverlayWidget we must have just never refactored this guy.

melloware:
Sorry had a caching issue. My Integration tests are now failing.  let me investigate.

melloware:
OK I am now convinced the code in 8.0 was wrong because this is all handled properly in DynamicOVerlayWidget it knows on `refresh` and `destroy` whether its supposed to destroy the overlay or not.    The overlay was basically saying destroy it if the `appendTo` was not set which is just wrong.

tandraschko:
@melloware please resolve conflicts

melloware:
Conflicts resolved and integration tests passing.

src/main/resources/META-INF/resources/primefaces/overlaypanel/overlaypanel.js
```diff
@@ -428,20 +410,10 @@ PrimeFaces.widget.OverlayPanel = PrimeFaces.widget.DynamicOverlayWidget.extend({
...
-                    }
-
-                    //append to body if not already appended by user choice
-                    if(!this.cfg.appendTo) {
-                        this.jq.appendTo(document.body);
-                    }
-                }
+            this.cfg.appendTo = PrimeFaces.utils.resolveAppendTo(this, this.target, this.jq);
+            //append to body if not already appended by user choice
+            if(!this.cfg.appendTo) {
```
tandraschko: i think i understood this now.
if we manually appendTo body now, without setting this.cfg.appendTo=@body, our DynamicOverlayWidget logic will never remove it.
Can you check why your integration tests doesnt fail? Normally if you set appendTo=null and update it, there should be 2 overlays below the body?

melloware: So here is what I can tell is happening and maybe i am confused so forgive me.

if appendTo== null then we automatically append to the body.
DyanmicOverlayWidget always registers `refresh` and `destroy` methods.
Your `refresh` method calls `cleanupDynamicOverlay` which will remove the overlay only if `appendTo` is actually set else it leaves it in the DOM.
If you change the appendTo with an update the refresh method should destroy the old one still only leaving one in the DOM right?

melloware: So previously my `destroy` test was saying the overlay was destroyed when appendTo was null and now its saying its not its still in the body.  So I updated my test thinking this behavior is wrong.

melloware: So here is what happens when appendTo==null
```java
@Test
    @Order(3)
    @DisplayName("OverlayPanel: Destroy overlay")
    public void testDestroy(Page page) {
        // Arrange
        OverlayPanel overlayPanel = page.overlayPanel;
        overlayPanel.show();
        Assertions.assertTrue(overlayPanel.isDisplayed());

        // Act
        page.btnDestroy.click();

        // Assert
        Assertions.assertFalse(overlayPanel.isDisplayed());
        assertConfiguration(overlayPanel.getWidgetConfiguration());
    }
```

Hree is what happens when `appendTo="@(body)"` or if used inside a dialog which sets the `@(body)`
```java
 @Test
    @Order(4)
    @DisplayName("OverlayPanel: Dialog Destroy overlay")
    public void testDialogDestroy(Page page) {
        // Arrange
        OverlayPanel overlayPanel = page.overlayPanel;
        page.btnDialog.click();
        Assertions.assertTrue(page.dialog.isDisplayed());
        overlayPanel.show();
        Assertions.assertTrue(overlayPanel.isDisplayed());

        // Act
        page.btnDestroy.click();

        // Assert
        try {
            overlayPanel.isDisplayed();
            Assertions.fail("OverlayPanel should have been destroyed.");
        }
        catch (NoSuchElementException ex) {
            // overlay panel should be destroyed
        }
    }
```

tandraschko: btnDestroy destroys the overlaypanel with updating a container and set rendered false to overlaypanel?
Then please check in the first test (appendTo==null), that there is now appended overlay left below the body

melloware: yes appendTo==null the Overlay remains and its not destroyed.

appendTo==@Body the overlay is destroyed

melloware: Oh actually let me check if it sets Rendered=false

melloware: OK yes the test calls Destroy and updated the entire `@form` and the overlay `rendered="false"` and its still in the DOM.

tandraschko: yep, we should default then to "@(body)" via stateHelper if that should be the default? and remove the special logic + move the resolveAppendTo in the #init method?

melloware: I think that is what we have been doing everywhere else let me try it.

melloware: OK with latest changes now in the OverlayPanelBase is defaulted to `"@(body)"` like  we do in other components. this also lets me remove the part in setupDialogSupport about appending it to the body if NULL since `DynamicOverlayWidget` is already doing that with `registerDynamicOverlay`.

However just checking in removeDynamicOverlay...
```javascript
removeDynamicOverlay: function(widget, overlay, overlayId, appendTo) {
            appendTo.children("[id='" +  overlayId + "']").not(overlay).remove();
        }
```

This says only remove the over lay if the appenTo has this id but its NOT already the overlay we have.  Is that correct?

tandraschko: i think null is passed on destruction and lightbox just uses it with param for "refresh", which is correct
but maybe lightbox should use cleanupDynamicOverlay on refresh and removeDynamicOverlay on destruction... or just inherit our base widget?

melloware: I have never used lightbox but it does seem like a prime candidate to extend DynamicOverlay.  Let me try it with the showcase.

melloware:
OK @tandraschko I think we can make all our components like SelectOneMenu etc that are `panel` based overlays a million times better.  Look at my last commit and now LightBox is a fully dynamic overlay with all our goods of refresh and destroy listeners etc.   If you approve I can change all the other panel based ones??

melloware:
I did CascaseSelect and Lightbox so you could see what they look like

src/main/resources/META-INF/resources/primefaces/core/core.widget.js
```diff
@@ -622,7 +622,19 @@ if (!PrimeFaces.widget) {
...
             this._super(cfg);
 
-            PrimeFaces.utils.registerDynamicOverlay(this, this.jq, this.id);
+            // do not bind overlay if widget disabled
+            if(cfg.disabled === true) {
+                return;
+            }
+
+            // check if this is a panel based overlay like lightbox or cascadeselect
+            if(this.panel) {
```
tandraschko: any nicer way then this?

melloware: HA I knew you are going to say that let me see if I can do something better.

melloware: OK I overloaded the constructor of DyanmicWidget to take the two extra params.  If htey are not provided it defaults to `this.jq` and `this.id` like it was already doing before.

melloware:
@blutorange this code works fine in JS but something tells me I am not allowed to overload a constructor like this even though it works fine.  TypeScript will not let me fix this.  Any thoughts?

blutorange:
@melloware  Well, `DynamicOverlayWidget` is a subclass of `BaseWidget`. According to the Liskov substitution principle, everywhere you can use `BaseWidget` you should be able to use a `DynamicOverlayWidget`. But the `init` method of `DynamicOverlayWidget` requires additional arguments, which one may not pass when dealing with a `BaseWidget`.

In other words, there is no guarantee that these arguments are actually passed -- you can make that fact explicit by marking these additional arguments as optional, see https://www.typescriptlang.org/play?#code/MYGwhgzhAEBCkFNoG8BQ0PQJYDssBcAKAMywCcJ8AuaSs3AcwEoaA3AeywBMUBfVfqlCQYAZQCuAIwCM0BAA98CHFxjwISNJmx4ipCtVr56OBgBpaCYOxU0c4gLaSEZFtA7cU6TP0HCo0BKSAExyisqqcIhe2rgEJOSUNHSMFhrWKgD8do7OrmycPFo+AkA

For JSDoc, you can make the arguments optional by enclosing them in brackets

```
         * @param {JQuery} [overlay] The DOM element for the overlay.
         * @param {string} [overlayId] The ID of the overlay, usually the widget ID.
         * @param {JQuery} [target] The DOM element that is the target of this overlay
```

melloware:
That is exactly what I needed because those 3 fields are optional.

src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
```diff
@@ -232,7 +232,7 @@ PrimeFaces.widget.DatePicker = PrimeFaces.widget.BaseWidget.extend({
 
         this.cfg.onPanelCreate = function() {
             $this.panel = this.panel;
-            $this.cfg.appendTo = PrimeFaces.utils.resolveAppendTo($this, $this.panel);
+            $this.cfg.appendTo = PrimeFaces.utils.resolveAppendTo($this, $thi.jq, $this.panel);
```
fanste: There is a typo at `$thi.jq`

melloware: Good catch!  Fixed

tandraschko:
@melloware 
i would like to get this merged but it would be great if we have some more tests, im not sure if they cover everything
i will try to explain some cases and here are similar tickets:
https://github.com/primefaces/primefaces/issues/3654
https://github.com/primefaces/primefaces/issues/7004
https://github.com/primefaces/primefaces/issues/5343
https://github.com/primefaces/primefaces/issues/3178


1) component with no appendTo defined
    nothing should be moved
    overlayid should be only once in the dom
    updating on it should do nothing, no appendTo logic triggered (not sure if we can verify, no onRefresh/Destroy listener from the appendTo logic should be registered)
    
    
2) componen with appendTo defined
    init must move the overlay to the configured appendTo
    refresh must remove the old overlay below the appendTo, and move the new overlay, which is a child of the widget, to the appendTo
        NOTE: when updating the component, the overlayId will be twice in the DOM for a short time (there is acutally not test/assert that its only there once)
    destroy must remove the overlay, which is below the appendTo
    
    
    
and there is even the special case for "dialog support"
it would be great to have those tests for 1-2 existing components and for those components which you touch with this PR

melloware:
I definitely don't think the tests cover all these scenarios.  Should we close this PR for now so you can do the project restructuring?  We can always open it again in the future.  This is not critical as its just a refactoring ticket.  Thoughts?

PR #7031, opened 2021-02-22
Fix #7029: LinkButton allow content

### LINKED ISSUE #7029: LinkButton: broken with custom markup
works
```
<p:linkButton href="#{request.contextPath}#{pdf.resourcePath}" target="_blank" value="#{msgs[pdf.msgKey]}">
</p:linkButton>
```

works not:
```
<p:linkButton href="#{request.contextPath}#{pdf.resourcePath}" target="_blank">
    <span>#{msgs[pdf.msgKey]}</span>
</p:linkButton>
```

melloware:
Refactored to use `StyleClassBuilder` and allow child content.

image

```xml
            <p:linkButton outcome="productDetail" icon="pi pi-star"  style="margin-right:20px;">
                <f:param name="productId" value="50"/>
                <span>Content</span>
            </p:linkButton>
```

### DIFF
--- a/src/main/java/org/primefaces/component/linkbutton/LinkButton.java
+++ b/src/main/java/org/primefaces/component/linkbutton/LinkButton.java
@@ -23,14 +23,14 @@
  */
 package org.primefaces.component.linkbutton;
 
-import java.util.List;
-import java.util.Map;
+import org.primefaces.util.ComponentUtils;
 
 import javax.faces.application.ResourceDependency;
+import javax.faces.component.UIComponent;
+import javax.faces.component.UIParameter;
 
-import org.primefaces.util.ComponentUtils;
-import org.primefaces.util.HTML;
-import org.primefaces.util.LangUtils;
+import java.util.List;
+import java.util.Map;
 
 @ResourceDependency(library = "primefaces", name = "components.css")
 @ResourceDependency(library = "primefaces", name = "jquery/jquery.js")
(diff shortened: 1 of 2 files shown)


src/main/java/org/primefaces/component/linkbutton/LinkButton.java
```diff
@@ -45,30 +49,36 @@
...
-        }
-        else if (value == null && !LangUtils.isValueBlank(icon)) {
-            styleClass = HTML.BUTTON_ICON_ONLY_BUTTON_CLASS;
-        }
+        String styleClass = PrimeRequestContext.getCurrentInstance(context).getStyleClassBuilder()
+                    .add("ui-linkbutton")
+                    .add(getStyleClass())
+                    .add(hasValue && !hasIcon, HTML.BUTTON_TEXT_ONLY_BUTTON_CLASS)
+                    .add(!hasValue && hasIcon, HTML.BUTTON_ICON_ONLY_BUTTON_CLASS)
+                    .add(isTextAndIcon && getIconPos().equals("left"), HTML.BUTTON_TEXT_ICON_LEFT_BUTTON_CLASS, HTML.BUTTON_TEXT_ICON_RIGHT_BUTTON_CLASS)
```
Rapster: "left".equals...

src/main/java/org/primefaces/component/linkbutton/LinkButton.java
```diff
@@ -45,30 +49,36 @@
...
-        String userStyleClass = getStyleClass();
-        if (userStyleClass != null) {
-            styleClass = styleClass + " " + userStyleClass;
+    public boolean hasDisplayedChildren() {
+        if (getChildCount() == 0) {
+            return false;
         }
-
-        return "ui-linkbutton " + styleClass;
+        else {
```
Rapster: remove else?

src/main/java/org/primefaces/component/linkbutton/LinkButton.java
```diff
@@ -45,30 +49,36 @@
...
+            for (int i = 0; i < getChildCount(); i++) {
+                UIComponent child = getChildren().get(i);
+                if (child instanceof UIParameter || !child.isRendered()) {
```
Rapster: what about returning true if child not UIParameter? I just feel is more readable and straightforward that way

melloware:
Thanks @Rapster great feedback.

src/main/java/org/primefaces/component/linkbutton/LinkButton.java
```diff
@@ -45,30 +49,34 @@
         return ComponentUtils.getUIParams(this);
     }
 
-    public String resolveStyleClass() {
-        String icon = getIcon();
-        Object value = getValue();
-        String styleClass = "";
+    public String resolveStyleClass(FacesContext context) {
```
Rapster: I know we have other resolveStyleClass in component, but I think it makes more sense to have that in renderers

melloware: I was wondering why this was here.  I agree I should move it to the renderer.

tandraschko:
looks good to me
i just copied it from CommandButton in the past, i'm ok with moving to renderer

src/main/java/org/primefaces/component/linkbutton/LinkButton.java
melloware: moved into renderer

PR #7033, opened 2021-02-24
Fix #7026 - Datatable: SelectOneMenu has wrong value after filtering

### LINKED ISSUE #7026: Datatable: SelectOneMenu has wrong value after filtering
**Describe the defect**
After filtering in a datatable with a SelectOneMenu in a column the value of the SelectOneMenu gets a wrong value. It seems like the selection is still the value from the entry which was in the row before the filtering.

**Reproducer**
https://github.com/felixoi/primefaces-test

**Environment:**
 - PF Version: 10.0.0-RC1
 - Affected browsers: All

**To Reproduce**
1. Take a look at the last row in the datatable. The SelectOneMenu has the value 'THREE'.
2. Filter the name by 'Test 5'.
3. As you can see, 'ONE' is now selected.
4. You can remove the filter, 'ONE' is still selected for 'Test 5'

**Expected behavior**
The value of the SelectOneMenu should not change.

Rapster:


felixoi (approved):

melloware (approved):

melloware:
@Rapster I think this latest change caused DataTable009Test of the Regression tests to fail last night.
```java
public class DataTable009Test extends AbstractDataTableTest {

    @Test
    @Order(1)
    @DisplayName("DataTable: filter - issue 1390 - https://github.com/primefaces/primefaces/issues/1390")
    public void testFilterIssue1390(Page page) {
```

melloware:
These regression tests are really paying for themselves!

Rapster:
I see... I'm starting to think the previous code was good and Mojarra is not doing things right... Though, as I have a logical explanation why it's no longer working, it really makes me wonder how come it works on MyFaces 😅

PR #7047, opened 2021-02-26
Fix #7044/#7045: SelectOne default to value when no label

### LINKED ISSUE #7044: SelectOneMenu: empty value breaks layout
Whenever the item value of a p:selectOneMenu is empty, layout of the menu changes (shrinks). Example:

```
          <p:selectOneMenu value="#{user.title}">
            <f:selectItem itemLabel="" />
            <f:selectItem itemValue="Dr." itemLabel="Dr." />
            <f:selectItem itemValue="Prof. Dr." itemLabel="Prof. Dr." />
          </p:selectOneMenu>

          <p:selectOneMenu value="#{user.salutation}">
            <f:selectItem itemLabel="#{messages.select}" itemDisabled="true" noSelectionOption="true" />
            <f:selectItem itemValue="Mr" itemLabel="Mr" />
            <f:selectItem itemValue="Ms" itemLabel="Ms" />
          </p:selectOneMenu>

```

Screenshot from 2021-02-26 16-06-56

### LINKED ISSUE #7045: SelectOneMenu: item value is not printed when there is no label
p:selectOneMenu delivers empty label when itemLabel is not explicitly defined. Example:

```
        <p:selectOneMenu value="#{user.title}">
            <f:selectItem itemLabel="" />
            <f:selectItem itemValue="Dr." />
            <f:selectItem itemValue="Prof. Dr." />
          </p:selectOneMenu>
```

Screenshot from 2021-02-26 16-12-49

p:selectOneMenu in contrast just prints the item value in such a case:

```
        <h:selectOneMenu value="#{user.title}">
            <f:selectItem itemLabel="" />
            <f:selectItem itemValue="Dr." />
            <f:selectItem itemValue="Prof. Dr." />
          </h:selectOneMenu>
```

melloware:
This makes it act like normal JSF SelectOneMenu.

image

### DIFF
--- a/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
+++ b/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
@@ -466,8 +466,7 @@ public class SelectOneMenuRenderer extends SelectOneRenderer {
         for (int i = 0; i < selectItems.size(); i++) {
             SelectItem selectItem = selectItems.get(i);
             Object itemValue = selectItem.getValue();
-            String itemLabel = selectItem.getLabel();
-            itemLabel = isValueBlank(itemLabel) ? "&nbsp;" : itemLabel;
+            String itemLabel = getOptionLabel(selectItem);
 
             String itemStyleClass = SelectOneMenu.ROW_CLASS;
             if (selectItem.isNoSelectionOption()) {
@@ -624,9 +623,7 @@ public class SelectOneMenuRenderer extends SelectOneRenderer {
                     writer.writeAttribute("data-title", option.getDescription(), null);
                 }
 
-                if (!isValueBlank(option.getLabel())) {
-                    writer.writeText(option.getLabel(), null);
-                }
+                writer.writeText(getOptionLabel(option), null);
 
                 writer.endElement("option");
             }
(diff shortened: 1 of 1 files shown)


melloware:
Fixes both issues...

image

tandraschko:
should we really default to itemValue.toString if no label is defined?

tandraschko:
yep, i think we should just do:
itemLabel =  isValueBlank(itemLabel) ? "&nbsp;" : itemLabel;

Otherwise you leave the label empty to display nothing but it defaults to "MyEntity[userid=1]" or something

melloware:
I am fine with that that is what we do in the mode higher up I think he was just pointing out that is not what JSF does but I am OK with that. Let me fix.

melloware:
Refactored into getOptionLabel method so it can be reused in both places.

fanste:
Seems to be a bit buggy. You replace the blank value with `&nbsp;` but still `writer.setText(...)` is used. This effectively writes `&amp;nbsp;` to the stream. If you use e.g. `widget.getSelectedLabel()` you get `&nbsp;` instead of ` ` (single space)

But this is still some kind of a breaking change and should be mentioned in the migration guide. If you added a blank item in JSF you now get a space. This breaks all JavaScripts if you rely on the blank value in your JavaScript (e.g. using `widget.getSelectedLabel()`)

melloware:
@fanste I believe the `&nbsp;` is checked for in the widget.  So it could just be that `getSelectedLabel()` needs to be adjusted right?

fanste:
Possibly yes, but it simply feels wrong that there is a readable `&nbsp;` instead of the HTML entity, IMHO.

PR #7050, opened 2021-02-28
Fix #7038: Datatable don't display until reflow is complete

### LINKED ISSUE #7038: DataTable: Slow header loading in reflow mode
**Describe the defect**
In a "Datatable" with the attribute reflow = "true" and in Responsive mode, the column headers are displayed lag behind the content.

**Reproducer**
Go to Showcase http://primefaces.org/showcase/ui/data/datatable/responsive.xhtml.

**Environment:**
- PF Version: 8 and 10.0
- java version "1.8.0_265"
- Affected browsers: All

**Example**

prime

melloware:
Hides the visibility of the datatable so it still takes up space and the DOM and allows the reflow to initialize all the columns and layout before actually displaying it to prevent the flicker and redraw


tandraschko:
@mertsincan please review

PR #7064, opened 2021-03-02
Fix #6940: InputNumber firing change event when value not changed

### LINKED ISSUE #6940: InputNumber: ajax change event fired even if value isn't changed

melloware:
@fanste I used === instead of == in your example.  But thanks for the suggestion

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js
+++ b/src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js
@@ -159,10 +159,9 @@ PrimeFaces.widget.InputNumber = PrimeFaces.widget.BaseWidget.extend({
      */
     copyValueToHiddenInput: function() {
         var oldVal = this.hiddenInput.val();
-
         var newVal = this.getValue();
 
-        if (Number(oldVal) !== Number(newVal)) {
+        if (((oldVal === '') ^ (newVal === '')) || Number(oldVal) !== Number(newVal)) {
             this.setValueToHiddenInput(newVal);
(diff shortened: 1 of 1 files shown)


fanste:
As already reported in #6945 it seems to fix the empty string / zero handling.

PR #7069, opened 2021-03-03
Fix #7065: Password align feedback panel and input width

### LINKED ISSUE #7065: Password: feedback width not equals to parent input
**Describe the defect**

the width of the feedback is not equal to the parent input

**Reproducer**

Schermata da 2021-03-03 10-40-50

**Environment:**
 - PF Version: _10.0_
 - JSF + version:  _2.3_
 - Affected browsers: _Chrome, iOS, ALL_

**To Reproduce**
1. Go to [https://www.primefaces.org/showcase/ui/input/password.xhtml?jfwid=88a94](url)
2. Click on Password Feedback example
3. Type 'a'

**Expected behavior**
The feedback should have the same width of the parent input
like the PrimeVue Password component

Schermata da 2021-03-03 10-41-16

**Example XHTML**
```html
<h:form id="frmTest">
  <h5>Feedback</h5>
  <p:password id="feedback" value="#{passwordView.password}" feedback="true"/>
</h:form>
```

**Example Bean**
```java
@Named
@RequestScoped
public class PasswordView implements Serializable {

    private String password;
    public String getPassword()  { return password; }
    public void setPassword(String password1) { this.password = password; }

}
```

melloware:
image

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/forms/forms.password.js
+++ b/src/main/resources/META-INF/resources/primefaces/forms/forms.password.js
@@ -250,6 +250,7 @@ PrimeFaces.widget.Password = PrimeFaces.widget.BaseWidget.extend({
         this.panel.css({
             left:'',
             top:'',
+            width: this.jq.outerWidth(),
             'transform-origin': 'center top'
         })
         .position({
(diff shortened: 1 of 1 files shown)


tandraschko:
setting to 11 as its not a bug

PR #7077, opened 2021-03-05
Fix #7072: Better error message when menu id is invalid

### LINKED ISSUE #7072: ClassCastException with DefaultMenuItem in MenuModel when set an id programmatically
Describe the defect
When setting MenuModel for a menubar programmatically and giving an id to an DefaultMenuItem, their is an ClassCastException on clicking the menu items.

Environment:
PF Version: 10.0.0-RC 1+2

Example working:
```xml
            <p:menubar>
                <p:menuitem value="Default Item" id="default" />
                <p:submenu label="Submenu">
                    <p:menuitem value="Submenu Default Item" />
                </p:submenu>
            </p:menubar>
```

Example not working:
```xml
                <p:menubar model="#{controller.menu}" />
```

```java
	public MenuModel getMenu() {
		final DefaultMenuModel menu = new DefaultMenuModel();

		menu.getElements().add(DefaultMenuItem.builder().value("Default Item").id("default").build());
		final DefaultMenuItem submenu_item = DefaultMenuItem.builder().value("Submenu Default Item").build();
		menu.getElements().add(DefaultSubMenu.builder().label("Submenu").elements(ImmutableList.of(submenu_item)).build());
		menu.generateUniqueIds();
		return menu;
}
```

Exception:
SEVERE [javax.enterprise.resource.webcontainer.jsf.context] (default task-3) java.lang.ClassCastException: org.primefaces.model.menu.DefaultMenuItem cannot be cast to org.primefaces.model.menu.MenuGroup at org.primefaces.renderkit.MenuItemAwareRenderer.findMenuitem(MenuItemAwareRenderer.java:192)

In Primefaces 8, both ways are working.

melloware:
javax.faces.FacesException: Menu ID 'default' is not valid for this component. Remove any manually set IDs from the MenuModel.

### DIFF
--- a/src/main/java/org/primefaces/renderkit/MenuItemAwareRenderer.java
+++ b/src/main/java/org/primefaces/renderkit/MenuItemAwareRenderer.java
@@ -176,7 +176,14 @@ public class MenuItemAwareRenderer extends OutcomeTargetRenderer {
                 return null;
             }
 
-            int childIndex = Integer.parseInt(paths[0]);
+            int childIndex;
+            try {
+                childIndex = Integer.parseInt(paths[0]);
+            }
+            catch (NumberFormatException e) {
+                throw new FacesException("Menu ID '" +  id +  "' is not valid for this component. Remove any manually set IDs from the MenuModel.");
+            }
+
             if (childIndex >= elements.size()) {
                 return null;
             }


tandraschko:
is this really invalid? i didnt follow all changes made.

melloware:
Basically because Menu, MenuButton, TieredMenu, MegaMenu, MenuBar all use `DefaultMenuModel` we made a fix to allow you to use an ID set on the model instead of the generated ID's.

See: https://github.com/primefaces/primefaces/issues/1039

That works for Menu, TieredMenu, MenuButton etc but for MegaMenu and MenuBar they cannot use a default id and must generate those ids like "menu_0_0" or "menu_2_1" which means submenu 2 item 1.  Its all based on the menu coordinates.

All this does is give a good explicit error message for MenuBar/MegaMenu if you try to set your own ID's in the model.

tandraschko:
hmm, then im not sure for 100% if the exception message fits this :D

melloware:
I am open to any message I just think a NumberFormatException right now tells the user nothing about what is wrong.  Any thoughts on a better or more generic message?

tandraschko:
i cant follow exactly. The message is:
`Menu ID '" +  id +  "' is not valid for this component. Remove any manually set IDs from the MenuModel.`

but you said:
`Basically because Menu, MenuButton, TieredMenu, MegaMenu, MenuBar all use DefaultMenuModel we made a fix to allow you to use an ID set on the model instead of the generated ID's.`

melloware:
Oh I apologize.  Basically for MegaMenu and MenuBar the manually setting of ID's is what causes this issue.  For all other types of menus manually setting the ID is OK.   But since they all just use DefaultMenuModel like this user was doing if they manually set an id they will get this NumberFormatException.  So basically manually setting ID in DefaultMenuModel is ok for Menu, MenuButton etc,  BUT NOT allowed for MenuBar and MegaMenu which call the method above.

tandraschko:
hmmm, sounds a bit weird
why dont they work all the same?

melloware:
I don't know the history of it but MegaMenu/MenuBar autogenerate the ID's by coordinates.  So when you click a menu item the AJAX on the client send takes "menu_2_1" and sends "2_1" to this method which then parses it to find SubMenu2->Menu1.  None of the other menu components seems to do this.

tandraschko:
I think we should clarify this before and do the correct changes, to unify all components

PR #7085, opened 2021-03-07
Fix #7079/#7082: CascadeSelect NPE or incorrect label displayed

### LINKED ISSUE #7079: CascadeSelect: NPE when placeholder-attribute is not set
Noted while writing integration-tests for CascadeSelect.

### LINKED ISSUE #7082: CascadeSelect: does not show selected value (eg after submit/update or for pre-selected values)
Integration-test PR is incoming

melloware:
Fixes both issues.  If for example you had a pre-selected value like "Montreal"
```java
private String selection = "Montreal";
```
It was not showing up as the selected label on the UI.

### DIFF
--- a/src/main/java/org/primefaces/component/cascadeselect/CascadeSelectRenderer.java
+++ b/src/main/java/org/primefaces/component/cascadeselect/CascadeSelectRenderer.java
@@ -41,6 +41,7 @@ import org.primefaces.expression.SearchExpressionUtils;
 import org.primefaces.renderkit.SelectOneRenderer;
 import org.primefaces.util.ComponentUtils;
 import org.primefaces.util.HTML;
+import org.primefaces.util.LangUtils;
 import org.primefaces.util.WidgetBuilder;
 
 public class CascadeSelectRenderer extends SelectOneRenderer {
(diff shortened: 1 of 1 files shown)


christophs78 (approved):

src/main/java/org/primefaces/component/cascadeselect/CascadeSelectRenderer.java
```diff
@@ -83,18 +84,18 @@ protected void encodeMarkup(FacesContext context, CascadeSelect cascadeSelect) t
 
         renderARIACombobox(context, cascadeSelect);
 
-        encodeInput(context, cascadeSelect);
-        encodeLabel(context, cascadeSelect);
+        String valueToRender = ComponentUtils.getValueToRender(context, cascadeSelect);
```
christophs78, line 87: It´s only partial solved. Only solved for SelectItems which only have a label.
It does not work when we have value and label. In this case we see after update the value instead of the label.

melloware: Fixing with another PR

PR #7088, opened 2021-03-08
Fix #7082: CascadeSelect use SelectItem label with value fallback

### LINKED ISSUE #7082: CascadeSelect: does not show selected value (eg after submit/update or for pre-selected values)

melloware:
This has to recursively check the SelectItem to find the value and associate the label as discovered by Integration tests.

Order for label is:
1. SelectItem Label
2. ValueToRender if Label could not be determined
3. PlaceHolder if #1 and #2 are not definded.

### DIFF
--- a/src/main/java/org/primefaces/component/cascadeselect/CascadeSelectRenderer.java
+++ b/src/main/java/org/primefaces/component/cascadeselect/CascadeSelectRenderer.java
@@ -24,7 +24,9 @@
 package org.primefaces.component.cascadeselect;
 
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
 
 import javax.faces.component.UIComponent;
 import javax.faces.component.UISelectOne;
@@ -86,7 +88,7 @@ public class CascadeSelectRenderer extends SelectOneRenderer {
 
         String valueToRender = ComponentUtils.getValueToRender(context, cascadeSelect);
         encodeInput(context, cascadeSelect, valueToRender);
-        encodeLabel(context, cascadeSelect, valueToRender);
+        encodeLabel(context, cascadeSelect, selectItems, valueToRender);
         encodeTrigger(context);
         encodePanel(context, cascadeSelect, selectItems);
 
@@ -116,9 +118,17 @@ public class CascadeSelectRenderer extends SelectOneRenderer {
         writer.endElement("div");
     }
 
-    protected void encodeLabel(FacesContext context, CascadeSelect cascadeSelect, String valueToRender) throws IOException {
+    protected void encodeLabel(FacesContext context, CascadeSelect cascadeSelect, List<SelectItem> itemList, String valueToRender) throws IOException {
         ResponseWriter writer = context.getResponseWriter();
-        String placeholder = LangUtils.isNotBlank(valueToRender) ? valueToRender : cascadeSelect.getPlaceholder();
+
+        Converter converter = ComponentUtils.getConverter(context, cascadeSelect);
+        String itemLabel = valueToRender;
+        SelectItem foundItem = findItemByValue(context, cascadeSelect, converter, itemList, valueToRender);
+        if (foundItem != null) {
+            itemLabel = foundItem.getLabel();
+        }
+
+        String placeholder = LangUtils.isNotBlank(itemLabel) ? itemLabel : cascadeSelect.getPlaceholder();
         String styleClass = getStyleClassBuilder(context)
                 .add(placeholder != null, CascadeSelect.LABEL_CLASS)
                 .add(placeholder == null, CascadeSelect.LABEL_EMPTY_CLASS)
@@ -190,9 +200,7 @@ public class CascadeSelectRenderer extends SelectOneRenderer {
                 boolean isGroup = selectItem instanceof SelectItemGroup;
                 Object itemValue = selectItem.getValue();
                 String itemLabel = selectItem.getLabel();
-                String itemValueAsString = converter != null ?
-                    converter.getAsString(context, cascadeSelect, selectItem.getValue())
-                    : String.valueOf(selectItem.getValue());
+                String itemValueAsString = getOptionAsString(context, cascadeSelect, converter, selectItem.getValue());
                 String itemStyleClass = getStyleClassBuilder(context)
                         .add(CascadeSelect.ITEM_CLASS)
                         .add(isGroup, CascadeSelect.ITEM_GROUP_CLASS)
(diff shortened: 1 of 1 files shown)


tandraschko:
i wonder how this is done in other components?

melloware:
SelectoneMenu does the same sort of thing but even more complex because it has an "Editable" mode for the label.  I wonder if I should refactor this method up to `SelectRenderer` since I basically borrowed this code.
```java
/**
     * Recursive method used when editable="true" to detect whether the submitted value
     * matches a value in one of the SelectItems. GitHub #2862 and #6507
     * @param context FacesContext
     * @param selectItems the List of SelectItems
     * @param editorInput the editor input value to search for
     * @return either the value found or NULL if no item found
     */
    protected Object findEditableValue(FacesContext context, List<SelectItem> selectItems, String editorInput) {
        Object foundValue = null;
        for (int i = 0; i < selectItems.size(); i++) {
            SelectItem item = selectItems.get(i);
            if (item instanceof SelectItemGroup) {
                SelectItemGroup selectItemGroup = (SelectItemGroup) item;
                foundValue = findEditableValue(context, Arrays.asList(selectItemGroup.getSelectItems()), editorInput);
                if (foundValue != null) {
                    break;
                }
            }
            else {
                if (item.getLabel().equalsIgnoreCase(editorInput)) {
                    foundValue = item.getValue();
                    break;
                }
            }
        }

        return foundValue;
    }
```

tandraschko:
+1 for refactoring after 10.0

melloware:
OK created the 11.0 ticket to refactor this!

src/main/java/org/primefaces/component/cascadeselect/CascadeSelectRenderer.java
```diff
@@ -116,9 +118,18 @@ protected void encodeInput(FacesContext context, CascadeSelect cascadeSelect, St
         writer.endElement("div");
     }
 
-    protected void encodeLabel(FacesContext context, CascadeSelect cascadeSelect, String valueToRender) throws IOException {
+    protected void encodeLabel(FacesContext context, CascadeSelect cascadeSelect, List<SelectItem> itemList, String valueToRender) throws IOException {
         ResponseWriter writer = context.getResponseWriter();
-        String placeholder = LangUtils.isNotBlank(valueToRender) ? valueToRender : cascadeSelect.getPlaceholder();
+
+        String itemLabel = valueToRender;
+        if (LangUtils.isNotBlank(valueToRender)) {
```
tandraschko: Actually we should not check for empty or null here, one might even have selectItem with value="" or value="#{null}"

melloware: Fixed!

src/main/java/org/primefaces/component/cascadeselect/CascadeSelectRenderer.java
```diff
@@ -116,9 +118,18 @@ protected void encodeInput(FacesContext context, CascadeSelect cascadeSelect, St
         writer.endElement("div");
     }
 
-    protected void encodeLabel(FacesContext context, CascadeSelect cascadeSelect, String valueToRender) throws IOException {
+    protected void encodeLabel(FacesContext context, CascadeSelect cascadeSelect, List<SelectItem> itemList, String valueToRender) throws IOException {
         ResponseWriter writer = context.getResponseWriter();
-        String placeholder = LangUtils.isNotBlank(valueToRender) ? valueToRender : cascadeSelect.getPlaceholder();
+
+        String itemLabel = valueToRender;
+        if (valueToRender != null) {
```
tandraschko: why still the null check?

melloware: You are right this is unnecessary now.  Removed.

src/main/java/org/primefaces/component/cascadeselect/CascadeSelectRenderer.java
```diff
@@ -247,6 +256,36 @@ protected void encodeListItems(FacesContext context, CascadeSelect cascadeSelect
...
+            if (item instanceof SelectItemGroup) {
+                SelectItemGroup selectItemGroup = (SelectItemGroup) item;
+                foundValue = findItemByValue(context, Arrays.asList(selectItemGroup.getSelectItems()), value);
+                if (foundValue != null) {
+                    break;
+                }
+            }
+            else {
+                String itemValue = Objects.toString(item.getValue());
+                if (value.equalsIgnoreCase(itemValue)) {
```
tandraschko: this will lead to a NPE, when value==null
i would just use Object.equals, without ignoreCase

src/main/java/org/primefaces/component/cascadeselect/CascadeSelectRenderer.java
```diff
@@ -247,6 +256,36 @@ protected void encodeListItems(FacesContext context, CascadeSelect cascadeSelect
...
+     * @param value the input value to search for
+     * @return either the SelectItem found or NULL if not found
+     */
+    protected SelectItem findItemByValue(FacesContext context, List<SelectItem> selectItems, String value) {
+        SelectItem foundValue = null;
+        for (int i = 0; i < selectItems.size(); i++) {
+            SelectItem item = selectItems.get(i);
+            if (item instanceof SelectItemGroup) {
+                SelectItemGroup selectItemGroup = (SelectItemGroup) item;
+                foundValue = findItemByValue(context, Arrays.asList(selectItemGroup.getSelectItems()), value);
```
tandraschko: will Arrays.asList work if selectItemGroup.getSelectItems() is null?

melloware: Agreed I need to add a NULL guard here because Arrays.asList would throw an NPE if the array is null.

melloware:
So when switching from coverting a String to `Objects.equals()` it no longer works.
```java
Object itemValue = item.getValue();
if (Objects.equals(value, itemValue)) {
```
The reason is the Value Object is a `Place` and its toString() would equal the label but with `Objects.equals` its comparing `Montreal` String the Place `Place("MRTL", null, "Montreal");` and never finds a match.
image

tandraschko:
then we should even check for converters and so on

melloware:
OK adding converter check.

melloware:
Fixed both places in the code to use the already built `SelectRenderer.getOptionAsString` and added NPE guards as requested.

src/main/java/org/primefaces/component/cascadeselect/CascadeSelectRenderer.java
```diff
@@ -247,6 +253,41 @@ protected void encodeListItems(FacesContext context, CascadeSelect cascadeSelect
...
+                    continue;
+                }
+                foundValue = findItemByValue(context, component,  converter, Arrays.asList(selectItemGroup.getSelectItems()), value);
+                if (foundValue != null) {
+                    break;
+                }
+            }
+            else {
+                String itemValueAsString = getOptionAsString(context, component, converter, item.getValue());
+                if (itemValueAsString.equals(value)) {
```
tandraschko: Can't we use Objects.equals?

melloware: Yep switched to Object.equals() verified its still working properly.

PR #7095, opened 2021-03-09
Fix #7086: CellEdit focus revert #6187

### LINKED ISSUE #7086: DataTable: cellEdit no longer sets focus on InputText content
**Describe the defect**
In PrimeFaces 8 clicking in a table cell will automatically activate/focus the input text, so the user can start typing immediately.

In PrimeFaces 10 this behavior has changed. The user needs to click in the input text component again to enter content.

For keyboard only usage this is even worse. Hitting tab from an inputText cell will activate the next cell, but the keyboard focus is not there. hitting tab again sets the focus and the user may enter content.

I think this is a regression. Same can be found for other components as well (SelectMenu for example).

**Reproducer**
This can be reproduced in the showcase.

**Environment:**
 - PF Version: _8.0_, _10.0-SNAPSHOT_
 - Affected browsers: _ALL_

**To Reproduce**
1. Click on a table cell in a table with editMode=cell
2. Click on the cell again to gain focus
3. Use Tab / Shift Tab to navigate
4. 
**Expected behavior**
Clicking on a table cell should set the focus on the contained input component.

melloware:
Reverts #6187

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
+++ b/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
@@ -3198,7 +3198,7 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
 
         var inputs=row.find(':input:enabled');
         if (inputs.length > 0) {
-            inputs.first().triggerHandler('focus');
+            inputs.first().trigger('focus');
         }
     },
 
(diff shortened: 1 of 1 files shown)


melloware:
I marked 6187 with a workaround.  I don't think we can fix this scenario for everybody so we need to put it back to 8.0 behavior and have the users in 6187 just eat the event themselves.

PR #7110, opened 2021-03-11
Fix #5781: Remove FontAwesome 4.7

### LINKED ISSUE #5781: Task: Remove FontAwesome 4.7
Since FontAwesome 5 can be used and users can keep it up to date with whatever version they choose we should remove FontAwesome 4.7 from core PrimeFaces JAR.

Plus PrimeTek has been going with PrimeIcons so it makes even more sense to remove it.

See: https://primefaces.github.io/primefaces/8_0/#/core/fonticons

Vote up or down here if your like or dislike the idea.

melloware:


tandraschko:
@melloware adjusted the docs, please review

PR #7111, opened 2021-03-11
Fix #7089: SelectOne Refactor findSelectItem

### LINKED ISSUE #7089: SelectOneMenu/CascadeSelect: Refactor findSelectItem
**Description**
`SelectOneMenu.findEditableValue` and `CascadeSelect.findItemByValue` are the same basic method.  This should be refactored in 11.0 up to the parent `SelectRenderer`.

melloware:
We already have Integration Tests checking both scenarios so we can verify this refactor

### DIFF
--- a/src/main/java/org/primefaces/renderkit/SelectOneRenderer.java
+++ b/src/main/java/org/primefaces/renderkit/SelectOneRenderer.java
@@ -23,11 +23,17 @@
  */
 package org.primefaces.renderkit;
 
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+
 import javax.faces.component.UIComponent;
 import javax.faces.component.UISelectOne;
 import javax.faces.context.FacesContext;
+import javax.faces.convert.Converter;
+import javax.faces.model.SelectItem;
+import javax.faces.model.SelectItemGroup;
 
 public abstract class SelectOneRenderer extends SelectRenderer {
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/primefaces/renderkit/SelectOneRenderer.java
```diff
@@ -71,5 +77,48 @@ protected Object getSubmittedValues(UIComponent component) {
...
+                String valueOrLabel, boolean byValue) {
+        SelectItem foundValue = null;
+        for (int i = 0; i < selectItems.size(); i++) {
+            SelectItem item = selectItems.get(i);
+            if (item instanceof SelectItemGroup) {
+                SelectItemGroup selectItemGroup = (SelectItemGroup) item;
+                if (selectItemGroup.getSelectItems() == null) {
+                    continue;
+                }
+                foundValue = findSelectItem(fc, component,  converter, Arrays.asList(selectItemGroup.getSelectItems()), valueOrLabel, byValue);
```
tandraschko, line 128: i like it more to do a short "return" here
we can remove 1 if, 2 breaks, 1 var with that

melloware: Hmmm can you elaborate?  I am not quite sure I follow.

tandraschko: "return findSelectItem(...." but nvm, doesnt work
did a to short review :D

melloware: Got it. Yeah i only want to break the loop if its found.

tandraschko:
i would still prefer 2 protected methods (findSelectItemByLabel, findSelectItemByValue) and they should call the private one. WDYT?

also it would be great to have 2-3 basic unittests

melloware:
OK renamed methods ByLabel and ByValue and created private method called by both.

I already have Integration Tests for both scenarios CascadeSelect tests the `selectItemByValue` and the SelectOne tests `findItemByLabel` so we will know if this is broken or not.

melloware:
All Integration Tests passing after this fix: https://github.com/primefaces-extensions/primefaces-integration-tests/actions/runs/656557637

PR #7113, opened 2021-03-12
Fix #6917: Add StyleBuilder

### LINKED ISSUE #6917: StyleBuilder: create style attribute easily
**Description**
Similar to #6610 for created style classes we need one similar for creating style attribute.

melloware:
Added Builder and I picked SkeletonRenderer to show its capabilities and how much it cleans up IF ELSE's!


tandraschko:
please add some unittests

melloware:
Added unit tests to test all cases

tandraschko:
another test with multiple calls to builder.add would be nice, so we catch all cases

melloware:
Added "complex" use case with multiple different `adds`

PR #7127, opened 2021-03-14
Fix #7126: DatePicker auto detect display mode

### LINKED ISSUE #7126: DatePicker: AutoDetect Display Mode
**Description**
Add a new `autoDetectDisplay` attribute to the DatePicker which will detect if the browser is Mobile it will set `touchUi=true` else it will be false for Desktop browser.

Based on conversation on Discord with @jepsar and @willbigas

melloware:


src/main/resources/META-INF/resources/primefaces/core/core.env.js
```diff
@@ -32,7 +32,7 @@ if (!PrimeFaces.env) {
          */
         init : function() {
             this.browser = $.browser;
-            this.mobile = this.browser.mobile;
+            this.mobile = (this.browser.mobile) ? true : false;
```
melloware, line 35: Always force env to have a true or false value. Right now on desktop its `undefined`.

christophs78: https://api.jquery.com/jquery.browser/
Is this something we still can rely on?

melloware:
I updated that jQuery component a couple of years ago and it still seems to accurately identify all of the browsers. If there are any that are incorrect we can fix that plugin. It's actually pretty simple

tandraschko:
not sure actually
i think it makes sense to set touchUI on mobile (and also make it deactivatable) but not sure if responsive == touchUI

melloware:
Well I thought the definition of Responsive Design was that the application adjust itself to the size of the browser. So in Mobile it acts differently than in Desktop.  

> Responsive web design (RWD) is an approach to web design that makes web pages render well on a variety of devices and window or screen sizes. Recent work also considers the viewer proximity as part of the viewing context as an extension for RWD.[1] Content, design and performance are necessary across all devices to ensure usability and satisfaction.

So that is why I chose the attribute `responsive`.

mertsincan:
I agree with @tandraschko. if responsive == touchUI, I'm not sure too :/ Could this cause confusion? I think it would be better for the user to change it.

melloware:
Yes but the issue on Discord and Forums and I agree with them is I want it to be touchui for Mobile devices and normal mode on Desktop browsers for my 30+ datepickers in my app without having to write any code and do something to flip them?  I think this feature would be a huge convenenience. I don't mind changing the name to something else besides `responsive` ???

melloware:
renamed `responsive` to `autoDetectDisplay`.

PR #7129, opened 2021-03-15
Fix Spanish translation client side

albfernandez:
In the context of a button clicked to clear the data, the correct Spanish word is "Limpiar"


melloware:
Thank you very much!

PR #7132, opened 2021-03-16
run build against JDK 8, 11, 16 (instead of 8, 11, 15)

christophs78:


melloware:
Looks like Java 16 is failing the WebXmlParserTest

christophs78:
> Looks like Java 16 is failing the WebXmlParserTest

Yep, probably due to  https://openjdk.java.net/jeps/396

christophs78:
Think this is something we need to discuss. Xerces is not there OOTB with Java 16.
WebXmlParser seems to be a integral part of PrimeFaces.
We may force our users to bring Xalan + Xerces (or Saxon), but not sure how much sense this makes.
@tandraschko

christophs78:
Think we have a simple solution. :-)

tandraschko:
but our XML parsing code still works on j16?

christophs78:
> but our XML parsing code still works on j16?

As far as i currently understand everything that uses javax.xml.xpath still works.
But you can´t explizit use com.sun.org.apache.xpath.internal.XXX - classes any more.

tandraschko:
why do even use this com.sun stuff hardcoded?

christophs78:
> why do even use this com.sun stuff hardcoded?

"We" used it in WebXmlParserTest. So i modified this test as you can see in the PR-diff. IMO it was a test-only issue. This test had some room for improvements. :-)

melloware (approved):

GedMarc:
yeah jdk 16 locked down a lot, time to look at bringing in the module-info as well - the auto-deny instead of auto-permit change has ripple effects,
The jakarta faces xml also looks needed now, not for jdk 16, but for IDE's to pick up the tag library sigh...

christophs78:
> yeah jdk 16 locked down a lot, time to look at bringing in the module-info as well - the auto-deny instead of auto-permit change has ripple effects,
> The jakarta faces xml also looks needed now, not for jdk 16, but for IDE's to pick up the tag library sigh...

Do you want to do a PR for this?
Had there been some previous discussion about module-info a few months or a year or two before?
IMO a solution which needs less maintenance is preferred.

PR #7150, opened 2021-03-22
Core: i18n message property encoding

jepsar:


jepsar:
Stolp already has the same PR ready

Issue #1715, opened 2016-08-22
DataTable - Filtering doesn't rerender Paginator; therefore the Paginator disappears sometimes with paginatorAlwaysVisible=false


jxmai:
**filterTable.xhtml**

``` xhtml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml"
    xmlns:ui="http://java.sun.com/jsf/facelets"
    xmlns:h="http://java.sun.com/jsf/html"
    xmlns:f="http://java.sun.com/jsf/core"
    xmlns:p="http://primefaces.org/ui"
    xmlns:fn="http://java.sun.com/jsp/jstl/functions">

<h:head>
</h:head>

<h:body>
        <h:form>
            <p:commandButton actionListener="#{tableBean.init()}" update=":tableForm" value="reload" oncomplete="PF('tableWidget').filter();"></p:commandButton>
        </h:form>

        <h:form id="tableForm">
            <p:dataTable value="#{tableBean.persons}" 
                        id="persons"
                        widgetVar="tableWidget"
                        paginator="true"
                        paginatorAlwaysVisible="false"
                        rows="2"
                        paginatorTemplate="{CurrentPageReport} {FirstPageLink} {PreviousPageLink} {PageLinks} {NextPageLink} {LastPageLink} {RowsPerPageDropdown}"
                        rowsPerPageTemplate="2"
                        filteredValue="#{tableBean.filterPersons}"
                        var="person">
                <f:facet name="header">
                        Bug Report 
                </f:facet>

                <p:column headerText="Name" filterBy="#{person.name}" filterMatchMode="contains" footerText="contains">
                    <h:outputText value="#{person.name}" />
                </p:column>
            </p:dataTable>
        </h:form>

</h:body>
</html>
```

**TableBean.java**

``` java

import java.io.Serializable;
import java.util.ArrayList;
import java.util.List;

import javax.annotation.PostConstruct;
import javax.faces.bean.ManagedBean;
import javax.faces.bean.ViewScoped;

@ManagedBean(name = "tableBean")
@ViewScoped
public class TableBean implements Serializable {

    private List<Person> persons = new ArrayList<Person>();

    private List<Person> filterPersons;

[...]

Issue #1178, opened 2016-02-23
DataTable: Header/Footer not updated by sort/filter 


xKopax:
Today when we filter the datatable 
`PF('table').filter()`
the datatable's footer is not updated automatically. However the footer usually holds summarized values of the columns. It'd be nice a easier way to acomplish this task.

Thanks!

sudheerj:
As of now you can update footer values with the help of p:remoteCommand

http://stackoverflow.com/questions/24452834/how-to-ajax-update-an-item-in-the-footer-of-a-primefaces-datatable

kukel:
@sudheerj: as of now? That is a two year old workaround...

jansohn:
+1 for this. Having to add a separate `p:remoteCommand` is cumbersome...

SharpShooter17:
I have the same problem. Is there any good solution?

marcelocaser:
Yes, this is very important for us!

marcelocaser:
> I have the same problem. Is there any good solution?

per hour, did I use this solution? https://stackoverflow.com/questions/24452834/how-to-ajax-update-an-item-in-the-footer-of-a-primefaces-datatable

mertsincan:
Fixed in https://github.com/primefaces/primefaces/issues/7165. Please use partialUpdate="false" for this issue.

PR #7157, opened 2021-03-23
Add partialUpdate attribute to DataTable

mertsincan:
For this issue;
https://github.com/primefaces/primefaces/issues/1178
https://github.com/primefaces/primefaces/issues/1715

| Name | Default | Type | Description
| ---| --- | --- | --- |
| partialUpdate | true | Boolean  | When disabled, it updates the whole table instead of updating a specific field such as body element in the client requests of the dataTable. Default is true.

**Note:** When disabled, it updates the whole table for Sorting, Filtering, and Page events.

**Test codes;**
<details>
<summary>TestModel.java</summary>

```
package org.primefaces.test;

import java.io.Serializable;

public class TestModel implements Serializable {
    
    public int id;
    public String str;
    public int number;

    public TestModel(int id, String str, int number) {
        this.id = id;
        this.str = str;
        this.number = number;
    }

    public int getId() {
        return id;
    }

    public void setId(int id) {
        this.id = id;
    }

    public String getStr() {
        return str;
    }

    public void setStr(String str) {
        this.str = str;
    }

    public int getNumber() {
        return number;
    }

    public void setNumber(int number) {
        this.number = number;
    }

    @Override
    public int hashCode() {
        int hash = 3;
        hash = 19 * hash + this.id;
        return hash;
    }

    @Override
    public boolean equals(Object obj) {
        if (this == obj) {
            return true;
        }
        if (obj == null) {
            return false;
        }
        if (getClass() != obj.getClass()) {
            return false;
        }
        final TestModel other = (TestModel) obj;
        if (this.id != other.id) {
            return false;
        }
        return true;
    }
}
```
</details>

<details>
<summary>TestView.java</summary>

```
package org.primefaces.test;

import java.io.Serializable;
import java.util.ArrayList;
import java.util.List;
import javax.inject.Named;

@Named
@ViewScoped

    private List<TestModel> testModels;
    private List<TestModel> filteredTestModels;


    @PostConstruct
    public void init() {
        testModels = new ArrayList<>();

        testModels.add(new TestModel(1, "a", 1));
        testModels.add(new TestModel(2, "b", 2));
        testModels.add(new TestModel(3, "c", 3));
        testModels.add(new TestModel(4, "d", 4));
        testModels.add(new TestModel(5, "e", 5));
        testModels.add(new TestModel(6, "f", 6));
        testModels.add(new TestModel(7, "g", 7));
        testModels.add(new TestModel(8, "h", 8));
        testModels.add(new TestModel(9, "k", 9));
        testModels.add(new TestModel(10, "l", 10));
        testModels.add(new TestModel(11, "m", 11));
    }

    public List<TestModel> getTestModels() {
        return testModels;
    }

    public void setTestModels(List<TestModel> testModels) {
        this.testModels = testModels;
    }

    public List<TestModel> getFilteredTestModels() {
        return filteredTestModels;
    }

    public void setFilteredTestModels(List<TestModel> filteredTestModels) {
        this.filteredTestModels = filteredTestModels;
    }
}
```
</details>

<details>
<summary>test.xhtml</summary>

```
<!DOCTYPE html>

    <h:head>
    </h:head>
    <h:body>
        <h:form id="frmTest">
            <p:dataTable widgetVar="testDT" var="testModel" value="#{testView.testModels}" reflow="true" rows="5" partialUpdate="false"
                         paginatorTemplate="{NumberOfEntries} {CurrentPageReport} {FirstPageLink} {PreviousPageLink} {PageLinks} {NextPageLink} {LastPageLink} {RowsPerPageDropdown}"
                         id="testDTFilter" filteredValue="#{testView.filteredTestModels}" paginator="true" paginatorPosition="top">
                
                <f:facet name="{NumberOfEntries}" >
                    <h:outputFormat value="Filtered {0} of {1} entries">
                        <f:param value="#{testView.filteredTestModels == null or testView.filteredTestModels.size() == testView.testModels.size() ? 0 : testView.filteredTestModels.size()}" />
                        <f:param value="#{testView.testModels.size()}" />
                    </h:outputFormat>
                </f:facet>
                
                <p:column headerText="Id" sortBy="#{testModel.id}" filterBy="#{testModel.id}">
                    <h:outputText value="#{testModel.id}"/>
                </p:column>
                <p:column headerText="Str" sortBy="#{testModel.str}" filterBy="#{testModel.str}">
                    <h:outputText value="#{testModel.str}"/>
                </p:column>
                <p:column headerText="Number" sortBy="#{testModel.number}" filterBy="#{testModel.number}">
                    <h:outputText value="#{testModel.number}"/>
                </p:column>
                <f:facet name="footer">
                    <div style="text-align: right">
                       Total Number: #{testView.filteredTestModels == null ? testView.testModels.stream().map(item->item.number).sum() : (testView.filteredTestModels.size() > 0 ? testView.filteredTestModels.stream().map(item->item.number).sum() : 0)} 
                    </div>
                </f:facet>
            </p:dataTable>
        </h:form>
    </h:body>
</html>
```
</details>

**Test video;**
https://www.dropbox.com/s/mc09itd4sbderfe/datatable_autoUpdate_test.mov?dl=0

### DIFF
--- a/src/main/java/org/primefaces/component/datatable/DataTable.java
+++ b/src/main/java/org/primefaces/component/datatable/DataTable.java
@@ -211,6 +211,10 @@ public class DataTable extends DataTableBase {
         return context.getExternalContext().getRequestParameterMap().containsKey(getClientId(context) + "_encodeFeature");
     }
 
+    public boolean isAutoUpdateRequest(FacesContext context) {
+        return context.getExternalContext().getRequestParameterMap().containsKey(getClientId(context) + "_autoUpdate");
+    }
+
     public boolean isRowEditRequest(FacesContext context) {
         return context.getExternalContext().getRequestParameterMap().containsKey(getClientId(context) + "_rowEditAction");
     }
--- a/src/main/java/org/primefaces/component/datatable/DataTableBase.java
+++ b/src/main/java/org/primefaces/component/datatable/DataTableBase.java
@@ -107,7 +107,8 @@ public abstract class DataTableBase extends UIPageableData implements Widget, RT
         stripedRows,
         showGridlines,
         size,
-        widgetVar
+        widgetVar,
+        autoUpdate
     }
 
     public DataTableBase() {
@@ -687,5 +688,13 @@ public abstract class DataTableBase extends UIPageableData implements Widget, RT
 
     public void setSelectionPageOnly(boolean selectionPageOnly) {
         getStateHelper().put(PropertyKeys.selectionPageOnly, selectionPageOnly);
+    }
+
+    public boolean isAutoUpdate() {
+        return (Boolean) getStateHelper().eval(PropertyKeys.autoUpdate, false);
+    }
+
+    public void setAutoUpdate(boolean autoUpdate) {
+        getStateHelper().put(PropertyKeys.autoUpdate, autoUpdate);
     }
 }
--- a/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
+++ b/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
@@ -88,15 +88,23 @@ public class DataTableRenderer extends DataRenderer {
                     feature.encode(context, this, table);
                 }
             }
+
+            if (table.isAutoUpdateRequest(context)) {
+                render(context, table);
+            }
         }
         else {
-            preRender(context, table);
-
-            encodeMarkup(context, table);
-            encodeScript(context, table);
+            render(context, table);
         }
 
         context.getApplication().publishEvent(context, PostRenderEvent.class, table);
+    }
+
+    protected void render(FacesContext context, DataTable table) throws IOException {
+        preRender(context, table);
+
+        encodeMarkup(context, table);
+        encodeScript(context, table);
     }
 
     protected void preRender(FacesContext context, DataTable table) {
@@ -246,6 +254,7 @@ public class DataTableRenderer extends DataRenderer {
                 .attr("rowHover", table.isRowHover(), false)
                 .attr("clientCache", table.isClientCache(), false)
                 .attr("multiViewState", table.isMultiViewState(), false)
+                .attr("autoUpdate", table.isAutoUpdate(), false)
                 .nativeAttr("groupColumnIndexes", table.getGroupedColumnIndexes(), null)
                 .callback("onRowClick", "function(row)", table.getOnRowClick());
 
--- a/src/main/java/org/primefaces/component/datatable/feature/FilterFeature.java
+++ b/src/main/java/org/primefaces/component/datatable/feature/FilterFeature.java
@@ -128,7 +128,9 @@ public class FilterFeature implements DataTableFeature {
 
         context.getApplication().publishEvent(context, PostFilterEvent.class, table);
 
-        renderer.encodeTbody(context, table, true);
+        if (!table.isAutoUpdateRequest(context)) {
+            renderer.encodeTbody(context, table, true);
+        }
     }
 
     public void filter(FacesContext context, DataTable table) {
--- a/src/main/java/org/primefaces/component/datatable/feature/PageFeature.java
+++ b/src/main/java/org/primefaces/component/datatable/feature/PageFeature.java
@@ -50,7 +50,7 @@ public class PageFeature implements DataTableFeature {
             table.loadLazyData();
         }
 
-        if (!isPageState) {
+        if (!isPageState && !table.isAutoUpdateRequest(context)) {
             renderer.encodeTbody(context, table, true);
         }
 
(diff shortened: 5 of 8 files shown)


mertsincan:
@tandraschko @melloware @Rapster @sqores WDYT?

tandraschko:
i would like to see another name for this feature as its completely different compared to p:autoUpdate / the old autoUpdate attribute on panelGrid.

melloware:
I like this change too but I agree with Thomas that `autoUpdate` might imply something its not. how about `autoRender` or `autoRerender`??

mertsincan:
@tandraschko +1 ;) When it's ok for everyone, I'll change its name. Do you have any other feedback? :D

Rapster:
Instead of introducing a new attribute, why not making it as a new behavior? Meaning, from now on, it will update the whole table?

melloware (approved):

tandraschko:
wheter autoUpdate nor autoRender/autoReRender are correct.
The new flag means that the whole table is rendered, not just the "data" part (rows/columns)

Does it make sense to make it the default behavior? Not sure here. It may flicker more and its not required in all cases.

I'm also not sure about this line:
table.setValue(null);

please add some comments mert :P

tandraschko:
we may also think about making it a attribute to p:ajax, like we did for skipChildren. 
skipChildren isnt used in all components, neither this new attribute would be used in all components.
But it would be great to have it configurable for each event, not global in DT.

Maybe something like "p:ajax updateDataOnly=true|false"

Rapster:
I agree it doesn't hurt, but giving a lot of options to a component makes it harder to maintain and a more cumbersome... If you look at the corresponding issues, there are defects, I think most of users expect this to be the behavior.

Regarding updateDataOnly, makes only sense for data components don't you think?

Yup please comment code, specially on datatable, not sure I follow why so much java code is modified but I'm probably missing something

mertsincan:
> Instead of introducing a new attribute, why not making it as a new behavior? Meaning, from now on, it will update the whole table?

@Rapster I think it would be better not to make such a change by default for now. As Cagatay said, innovations will be made in the DataTable in future versions. After these innovations, it would be better if we talk about this issue together and decide.
Example: https://primefaces.org/primeng/showcase/#/table/filter :D 

> I'm also not sure about this line:
> table.setValue(null);

@tandraschko it is the most important line of all code :D Since all states are rendered on the server side, when we delete the filter value from input field, a sort operation is performed according to the values that the filter has previously changed. Therefore, I need to reset the value. The filterFeature interferes too much with the value.

mertsincan:
> we may also think about making it a attribute to p:ajax, like we did for skipChildren.
- Yes, but for such issues, Most users don't want to use <p: ajax> events in each operation to update the table. So updateDataOnly="true|false" may be a different option.

Exp;
```
<p:dataTable autoRender="true"> // for filter, page and sort

OR

<p:dataTable>
    <p:ajax event="filter" updateDataOnly="false" /> // for filter
    <p:ajax event="sort" updateDataOnly="false" /> // for sort
    <p:ajax event="page" updateDataOnly="false" /> // for page
</p:dataTable>
```

tandraschko:
please dont call it autoRender, it currently does "autoRender of data". The new flag switches to "autoRender of whole component".
It also doesnt make sense to add it on p:ajax and p:dataTable. Either we decide to make it a global dataTable option or make it configurable for each event. Not both.

> Yes, but for such issues, Most users don't want to use <p: ajax> events in each operation to update the table. So updateDataOnly="true|false" may be a different option.

Nope, its exactly the same option, just global. Thats always our way like process on p:ajax. Default is @this, if someone needs to configure it, you can add p:ajax to overwrite to the default. Its the same with this new config.



We could even call it something like "partial component rendering".
Per default, AJAX (doesnt matter if p or f), always renders the whole component. 
In some components we implemented "partial rendering" of a component. DataTable just renders the "data / rows / columns" and replaces only this parts in the DOM.
We have this already in MANY components. Sidebar loads only the content, accordion or tabView only loads the new tab content and replaces only this tab-content in the DOM.
Its a global feature, we already have it since ages.

mertsincan:
Thank you very much for your feedback. I think "partial component rendering" word sounds good. For now, however, we can add it to the dataTable as an attribute named "autoRerenderComponent". Based on feedback from users and tests, we can make an improvement as suggested by you and remove the "autoRerenderComponent" attribute if necessary. For now, it would be more accurate to detect the effects of this feature on users' applications. Of course, your suggestion is more accurate as it covers all components for such problems.

@tandraschko Can you open a new issue about your suggestion? Thanks a lot!

tandraschko:
@mertsincan Im ok with the attribute for now as we can move the flag later to p:ajax if required.
BUt i would like to propose something like following names ("update" or "render"):
1) partialUpdate (my favorite; with true as default which means the current behavior)
2) updateDataOnly
3) updateWholeComponent

autoRerenderComponent doesnt really fit for 100% at it already does auto-rendering but just for the data, not the whole compnent.

mertsincan:
Nice!! partialUpdate +1 ;)

PR #7169, opened 2021-03-25
Fix #7166: Excel export currency cells

### LINKED ISSUE #7166: DataExporter: Excel export use currency data type for currency data
**Description**
Similar to https://github.com/primefaces/primefaces/issues/6648

When using Currency in a table export it to excel as a currency cell.

image

Reported by @willbigas on Discord

melloware:
I learned more about currency than I ever could have imagined.

Now in a Datatable/TreeTable you can do this..
```xml
                    <p:column styleClass="p-text-right">
                        <f:facet name="header">
                            <h:outputText value="Price"/>
                        </f:facet>
                        <h:outputText value="#{product.price}">
                            <f:convertNumber type="currency" />
                        </h:outputText>
                    </p:column>
```

Which will display like this depending on locale below is US and then DE.
image

image

Then the new CurrencyValidator I borrowed from Commons Validator but had to heavily modify for our needs handles detecting and parsing the currency.  And it also handles telling Excel what your currency format is all based on your Locale!

So now these cells are real currency cells that you can do Math and formulas on.
image

### DIFF
--- a/src/main/java/org/primefaces/component/datatable/export/DataTableExcelExporter.java
+++ b/src/main/java/org/primefaces/component/datatable/export/DataTableExcelExporter.java
@@ -26,6 +26,7 @@ package org.primefaces.component.datatable.export;
 import java.awt.Color;
 import java.io.IOException;
 import java.util.List;
+import java.util.Locale;
 
 import javax.faces.component.UIComponent;
 import javax.faces.component.UIPanel;
@@ -46,10 +47,7 @@ import org.primefaces.component.datatable.DataTable;
 import org.primefaces.component.export.ExcelOptions;
 import org.primefaces.component.export.ExportConfiguration;
 import org.primefaces.component.export.ExporterOptions;
-import org.primefaces.util.ComponentUtils;
-import org.primefaces.util.Constants;
-import org.primefaces.util.LangUtils;
-
+import org.primefaces.util.*;
 
 public class DataTableExcelExporter extends DataTableExporter {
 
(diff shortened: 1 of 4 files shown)


stolp:
This is great.

For the sake of completion: This will also work with explicit `currencySymbol` and `locale` attributes in convertNumber?

melloware:
@stolp I don't think it will.  Because I am using `locale = LocaleUtils.getCurrentLocale(context);` to detect your current locale.  What gets passed to the Exporter is just an already calculated value.

I think we would have to tweak the `Exporter.exportValue()` method which is how it gets the value from the cell.  So right now my method is receving the already parsed string from the `<h:outputText`

stolp:
I did some research and now I'm confused by your reply.

My assumption is that in case that a currency symbol is present in the value (I think it is there), so `CurrencyValidator.parse` will not take the locale attribute in account in its first attempt to parse it. Only if the currency symbol is missing the locale is taking into account.

See: https://commons.apache.org/proper/commons-validator/apidocs/org/apache/commons/validator/routines/CurrencyValidator.html#parse-java.lang.String-java.text.Format-

So I now assume that both `currencySymbol` and `locale` scenarios should work as well.

Or am I missing something?

melloware:
No you are correct the problem is I am generating 1 excel CellStyle for currency for the whole sheet using your global locale. `locale = LocaleUtils.getCurrentLocale(context);`

However if you had 2 columns in your datatable like this...

```xml
                    <p:column styleClass="p-text-right" headerText="US">
                        <h:outputText value="#{product.price}">
                            <f:convertNumber type="currency" currencySymbol="$" locale="US" />
                        </h:outputText>
                    </p:column>
                    <p:column styleClass="p-text-right" headerText="Germany">
                        <h:outputText value="#{product.price}">
                            <f:convertNumber type="currency" currencySymbol="€" locale="DE" />
                        </h:outputText>
                    </p:column>
```

That the output would look at your global locale and only format any that match your currency.  So in the above one column would be Currency for US if my locale was US and the Germany column would be Strings in the Excel Sheet.

stolp:
Thanks, that clears this up!

PR #7173, opened 2021-03-26
Fix #7075: Overlays resize/scroll handling for mobile

### LINKED ISSUE #7075: Overlays: On mobile not working
I'm using a Samsung Galaxy S10, and when a click at https://www.primefaces.org/showcase/ui/input/oneMenu.xhtml -> Advanced, the select items list closes unexpectedly.

melloware:
Because certain mobile devices, and screen size and settings may trigger either a scroll or a resize when an overlay is displayed we must use `align` instead of hiding the panel.  This is how PF 8 and before worked it always called align but in 10 we decided to make it work like PrimeReact/Vue/NG and `hide` the overlay.  I still think hiding is fine for non-mobile.


tandraschko:
please discuss/review with Mert

mertsincan:
Thanks a lot @melloware ;)

PR #7179, opened 2021-03-29
Fix #7167: Dynamic overlay not handling deferred widgets

### LINKED ISSUE #7167: Datatable: liveScroll not working when inside OverlayPanel
Datatable liveScroll not working when inside OverlayPanel

**Reproducer**
Datatable liveScroll (inside OverlayPanel) worked on PF 8, but after upgrading PF10, it's not working.

 - PF Version: 10
 - JSF + version: 2.2
 - Affected browsers: ALL

melloware:
OK this was tricky to debug but what is happening is that Datatable is a deferred widget which sets up its scrolling in the `_render` method.  The `_render` method was never being called and when I debugged its because the `jq.is(':visible')` was returning FALSE and thus the _render method was never being called to setup scrolling.

This was all part of the CSS refactoring for animations and what I had to do was call `show` before processing the AJAX response so the DeferredWidget would test  `jq.is(':visible')` was TRUE and then finishing setting up the datatable.

Attached is reproducer.
[pf-7167.zip](https://github.com/primefaces/primefaces/files/6223602/pf-7167.zip)

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/overlaypanel/overlaypanel.js
+++ b/src/main/resources/META-INF/resources/primefaces/overlaypanel/overlaypanel.js
@@ -475,6 +475,8 @@ PrimeFaces.widget.OverlayPanel = PrimeFaces.widget.DynamicOverlayWidget.extend({
                 {name: this.id + '_contentLoad', value: true}
             ],
             onsuccess: function(responseXML, status, xhr) {
+                // must show the overlay first so deferred wigets can render
+                $this._show(target);
                 PrimeFaces.ajax.Response.handle(responseXML, status, xhr, {
                         widget: $this,
                         handle: function(content) {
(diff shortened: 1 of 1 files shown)


tandraschko:
hmm not sure
the current code is by design to first replace the content, then show the new content. Otherwise it may cause weird flickering

melloware:
Yep that's why I assigned to Mert. This works fine in 8.0 but with the new CSS Transition handler callbacks the `:visible` is false and thus any DeferredWidget will not render correctly in a dynamic overlay.

brivalin (approved):

mertsincan:
Thanks @melloware ;) @sqores could you please check this issue in your local? We can work together on this issue.

buunsal:
Everything looks good. I think we can merge

tandraschko:
im really not sure if we should merge this, it feels wrong and hacky

tandraschko:
i will do some debugging and come back

tandraschko:
fixed it now by making OverlayPanel a real deffered container

PR #7182, opened 2021-03-30
Fix #7176/#6445: DatePicker not handling events properly with mask

### LINKED ISSUE #6445: DatePicker: Race-Condition between input-event and mask support
**Environment:**
 - PF Version: _9.0-SNAPSHOT_
 - Affected browsers: _ALL_

**To Reproduce**
1. Select date or enter it manually (do not press any key, after the component accepted the date)
2. Select the whole date using the mouse (not Ctrl+A or similar)
3. Press backspace / delete to clear the input
4. The datepicker won't update it's model

**Expected behavior**
Model of the datepicker should be updated to empty value.

**Root cause and solution**
The problem is caused by the inputmask library. It caches the backspace key to clear the field. It then triggers the `input` event. The `onUserInput` function of the datepicker won't process that `input` event, because it misses the previous `keydown` event (early out because of an IE11 fix). The `keydown` event will be handled after the `input` event.

The solution is quite simple: Move `this.applyMask()`  to the end of the `init` method or at least after `this.jq.datePicker(this.cfg);`. The event callbacks will be reordered so that the datepicker will process the `keydown` event before the inputmask library will do it.

**Example XHTML**
```html
<h:form id="frmTest">
  <p:datePicker widgetVar="dp" mask="true" pattern="yyyy-MM" view="month" />
</h:form>
```
**Example Javascript (DevTools)**
```js
console.log(PF('dp').getDate());
```

### LINKED ISSUE #7176: Datepicker: Change event is not called when using mask="true"
**Describe the defect**
The ajax "change event" is not called when the date is type istead of being selected.
The behaviour should be the same as the other inputs, the value of the input is changed and the user leaves the field,  the event is called.
Calendar works like this.

**Environment:**
 - PF Version: 10.0
 - JSF + version: _e.g. Mojarra 2.2.14
 - Affected browsers:  ALL_

**To Reproduce**
1. focus datepicker
2. type a date and leave
3. change event is not called

**Expected behavior**
If I change the value of the datepicker typing the date then the ajax method "change" should be called.

**Example XHTML**
```html
<p:datePicker mask="true"
		      value="#{testView.data}"
		      maskSlotChar=" "
		      pattern="dd/MM/yyyy"
		      monthNavigator="true"
		      yearNavigator="true"
		      showIcon="true"
		      showOnFocus="false">
	<p:ajax event="dateSelect" listener="#{testView.onDateSelect()}"/>
	<p:ajax event="change" listener="#{testView.onChange()}"/>		
</p:datePicker>
```

**Example Bean**
```java
@Named
@ViewScoped

	private Date data;

	public void onDateSelect() {
		System.out.println("hi date select");
	}

	public void onChange() {
		System.out.println("hi onChange");
	}

	public Date getData() {
		return data;
	}

	public void setData(Date data) {
		this.data = data;
	}
	
}
```
[primefaces-test-master.zip](https://github.com/primefaces/primefaces/files/6214023/primefaces-test-master.zip)

melloware:
- Wrapped IE specific code in IE so we know to remove it later on when we remove IE.
- Fixes both issues properly
- The onchange event will still not fire in IE but honestly every month that goes by IE is going to drift further and further away from supportability.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
+++ b/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
@@ -1752,7 +1752,10 @@
         },
 
         onInputKeyDown: function (event) {
-            this.isKeydown = true;
+            if (PrimeFaces.env.isIE()) {
+                this.isKeydown = true;
+            }
+
             if (event.keyCode === 27) {
                 //put the focus back to the inputfield
                 this.inputfield.trigger('focus');
@@ -1768,11 +1771,13 @@
         },
 
         onUserInput: function (event) {
-            // IE 11 Workaround for input placeholder
-            if (!this.isKeydown) {
-                return;
+            if (PrimeFaces.env.isIE()) {
+                // IE 11 Workaround for input placeholder
+                if (!this.isKeydown) {
+                   return;
+                }
+                this.isKeydown = false;
             }
-            this.isKeydown = false;
 
             var rawValue = event.target.value;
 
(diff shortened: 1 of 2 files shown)


melloware:
@fanste please review this change if you don't mind.

fanste (approved):

PR #7185, opened 2021-03-30
docs validationg -> validating

jepsar:


melloware:
Nice work. I meant to fix this as I fixed in the the Taglib before 10.0 as it was spelled wrong there too.  But I forgot!

jepsar:
Just noticed it. Was amazed how many instances were wrong.

melloware:
Its because it came from the TagLib when we first generated the 7.0 docs from the taglib.  And it was wrong because we used to generate the taglib.xml from a generator.  So its been propagating for years. :)  But now its all fixed!

PR #7188, opened 2021-04-01
Fix #7186: DatePicker improved range splitting

### LINKED ISSUE #7186: DatePicker: rangeSeparator split can conflict with date part separator
https://github.com/primefaces/primefaces/blob/4b74f0f75a2beb4a7915a0b12298ebc413b90958/src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java#L305-L314

With default range separator and ISO date format you will get the following string value for all days in March 2021 (in this case, Dutch format): `01-03-2021 - 31-03-2021`. Split by `-`:

Screenshot 2021-03-31 at 14 26 37

Splitting like `submittedValue.split(" " + datePicker.getRangeSeparator() + " ");` will be better, but still not bullet proof

Maybe using `–` (long dash) instead of `-` (regular hash) by default as the range separator? But this makes typing values harder.

jepsar:
Fix #7186

* Don't use a reg ex to split (you could inject special characters as separator)
* Split on separator padded with spaces (token)
* If the date pattern contains the token, throw an exception

### DIFF
--- a/src/test/java/org/primefaces/util/CalendarUtilsTest.java
+++ b/src/test/java/org/primefaces/util/CalendarUtilsTest.java
@@ -202,7 +202,7 @@ public class CalendarUtilsTest {
         Date now = (Date) CalendarUtils.now(datePicker, java.util.Date.class);
         assertTrue(new Date().compareTo(now) >= 0);
     }
-    
+
     @Test
     public void convertPattern() {
         assertNull(CalendarUtils.convertPattern(null));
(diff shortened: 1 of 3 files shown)


melloware (approved):

src/test/java/org/primefaces/util/CalendarUtilsTest.java
```diff
@@ -202,12 +202,25 @@ public void now_Time() {
...
+    public void splitRange() {
+        List<String> splitRange = CalendarUtils.splitRange("2021-03-01 - 2021-03-31", "yyyy-MM-dd", "-");
+        assertEquals(2, splitRange.size());
+        assertEquals("2021-03-01", splitRange.get(0));
+        assertEquals("2021-03-31", splitRange.get(1));
+        splitRange = CalendarUtils.splitRange("2021-03-01", "yyyy-MM-dd", "-");
+        assertTrue(splitRange.isEmpty());
+        splitRange = CalendarUtils.splitRange("", "yyyy-MM-dd", "-");
+        assertTrue(splitRange.isEmpty());
+    }
```
melloware, line 230: I don't see a test here where SplitRange throws the FacesExceptionfor having the pattern?

jepsar: Added

Issue #7201, opened 2021-04-06
Menu: Rendering menus throw ClassCastException if child element is XML comment


wal-jan:
**Describe the defect**
XML comments as child elements of _p:MenuButton_ (and presumably other menu elments) cause a ClassCastException.

**Reproducer**
[primefaces-test.zip](https://github.com/primefaces/primefaces/files/6265793/primefaces-test.zip)

**Environment:**
 - PF Version: _8.0_ _10.0_
 - JSF + version: _ALL_
 - Affected browsers: _ALL_

**To Reproduce**
1. mvn clean jetty:run; http://localhost:8080/primefaces-test
2. See error (stacktrace)
```txt
javax.servlet.ServletException: class com.sun.faces.facelets.compiler.UIInstructions cannot be cast to class org.primefaces.model.menu.MenuElement
...
	at org.primefaces.component.tieredmenu.TieredMenuRenderer.encodeElements(TieredMenuRenderer.java:110)
```

**Expected behavior**
A successfully rendered page.

**Example XHTML**
```html
        <h:form id="frmTest">
            <p:menuButton value="Menu">
                <!-- This comment breaks rendering of menu button. -->
            </p:menuButton>
        </h:form>
```

**Example Bean**
N/A

wal-jan:
I've already corrected the child element handling an can provide a PR.

melloware:
Sure please provide we can review!

melloware:
Reproducer: 
[pf-7201.zip](https://github.com/primefaces/primefaces/files/6267393/pf-7201.zip)

PR #7202, opened 2021-04-06
#7201 Corrected rendering menu elements (children)

wal-jan:


tandraschko:
Please explain it
Actually we should just skip non-MenuItem, Not sure why you switched all to Object

wal-jan:
_AbstractMenu#getElements()_ returns either _List&lt;org.primefaces.model.menu.MenuElement&gt;_ in case of menu has a model or _List&lt;javax.faxces.component.UIComponent&gt;_ otherwise. The common parent is List&lt;Object&gt;. Just to avoid class cast exceptions in subsequent foreach loops.

As far I can see the return value of _AbstractMenu#getElements()_ may also be _null_ so I added null checks.

melloware:
I think there is an easier way to do this.  Let me me investigate.

melloware:
See my PR: https://github.com/primefaces/primefaces/pull/7205/files

PR #7207, opened 2021-04-07
Fix #7206: Panel attribute 'renderEmptyFacets'

### LINKED ISSUE #7206: Panel: renders empty facets
As it was added to the DataTable component before, if the facet is empty, facet should not be rendered, or user should decide this by using the "renderEmptyFacets" attribute. https://github.com/primefaces/primefaces/pull/6438

buunsal:


melloware:
OK I just tested this and its working and makes it more correct.
```java
    /**
     * Checks if the facet and one of the first level children is rendered.
     * @param facet The Facet component to check
     * @param alwaysRender flag can be set by components to always render the facet if it exists
     * @return true if the facet should be rendered, false if not
     */
    public static boolean shouldRenderFacet(UIComponent facet, boolean alwaysRender) {
        if (facet == null) {
            // no facet declared at all
            return false;
        }

        // user wants to always render this facet so it can be updated
        if (alwaysRender) {
            return true;
        }

        if (!facet.isRendered()) {
            // For any future version of JSF where the f:facet gets a rendered attribute (https://github.com/javaserverfaces/mojarra/issues/4299)
            // or there is only 1 child.
            return false;
        }

        // Facet has no child but is rendered
        if (facet.getChildCount() == 0) {
            return true;
        }

        return shouldRenderChildren(facet);
    }
```

That says if the Facet is NOT NULL then the customer wants to render it it should always be rendered.

Rapster:
Quick thought, instead of doing this for every component, what about introducing an Interface "ShouldRenderFacet" (feel free to come up with a better name) and then `ComponentUtils#shouldRenderFacet` test the instance type and check whether or not attribute is set to true. Basically all our components (supporting facets) could inherit this, and no need to touch every single renderer. Saying this, and I think I already mentionned in the past, we should have a top interface for PF UI component, I can dig deeper that last point

melloware:
I think I like the Interface Idea.

buunsal:
> OK I just tested this and its working and makes it more correct.
> 
> ```java
>     /**
>      * Checks if the facet and one of the first level children is rendered.
>      * @param facet The Facet component to check
>      * @param alwaysRender flag can be set by components to always render the facet if it exists
>      * @return true if the facet should be rendered, false if not
>      */
>     public static boolean shouldRenderFacet(UIComponent facet, boolean alwaysRender) {
>         if (facet == null) {
>             // no facet declared at all
>             return false;
>         }
> 
>         // user wants to always render this facet so it can be updated
>         if (alwaysRender) {
>             return true;
>         }
> 
>         if (!facet.isRendered()) {
>             // For any future version of JSF where the f:facet gets a rendered attribute (https://github.com/javaserverfaces/mojarra/issues/4299)
>             // or there is only 1 child.
>             return false;
>         }
> 
>         // Facet has no child but is rendered
>         if (facet.getChildCount() == 0) {
>             return true;
>         }
> 
>         return shouldRenderChildren(facet);
>     }
> ```
> 
> That says if the Facet is NOT NULL then the customer wants to render it it should always be rendered.

I added your requested changes. Thanks 👍🏻

buunsal:
> Quick thought, instead of doing this for every component, what about introducing an Interface "ShouldRenderFacet" (feel free to come up with a better name) and then `ComponentUtils#shouldRenderFacet` test the instance type and check whether or not attribute is set to true. Basically all our components (supporting facets) could inherit this, and no need to touch every single renderer. Saying this, and I think I already mentionned in the past, we should have a top interface for PF UI component, I can dig deeper that last point

For Primefaces 11, we can open another issue and talk and decide it. But since this PR will be added to Primefaces 6,7,8,10,11 versions, I want to add it like this right now, if there is no problem.

melloware:
Agreed @sqores I think this is safe for you to patch all the Elite versions with as it fixes the issue.  And then for 11 I will open another ticket and do the Interface way.  @Rapster I will have you review my idea after I submit so you can tell me if I got it right.

melloware (approved):

PR #7209, opened 2021-04-08
Fix #7208 - Refactor PF API

### LINKED ISSUE #7208: PrimeUIComponent: new class
The idea is to give a clear and basic definition of our ui components (and also filling this gap in JSF api missing a UIComponent interface).

Rapster:
Throwing ideas about refactoring PF API. Not sure how far I'll go with this, but I think few things are worth attention... (And yes it's goes beyond the original PrimeUIComponent idea)

To sum up:
- PrimeUIComponent: PF extension/abstraction of UIComponent
- PrimeUICommand: PF extension/abstraction of  UICommand
- PrimeUIOutcomeTarget: PF extension/abstraction of UIOutcomeTarget
- PrimeClientBehaviorHolder: PF extension of ClientBehaviorHolder (more like a workaround of some bug)

Pending ideas (note to myself):
- CoreRenderer: We have a standard way of implementing components, show us the way... Rename into `PrimeRenderer`, and introduce `encodeMarkup`/`encodeScript`
- MixedClientBehaviorHolder
- MenuItemAware/DialogReturnAware: poor/bad design (bad workaround for multiple inheritance)

In general, I understand big refactoring can impact premium themes (as they certainly have custom component) but PF team (e.g members+ PrimeTek) should be able to update them after such refactoring (private git repo would be great)


melloware:
I like what you did here.  I am actually OK with the way you have it now with `Widget` extending the `PrimeUIComponent` interface.

tandraschko:
will review in some weeks
currently im in JSF spec dev, therefore i dont plan to review such big refactorings for now in PF

Rapster:
I'll let you know once PR ready to review, I'm still trying to figure few things out

Rapster:
Would you be interested to see this fixed (at least, first 4 points) ? I would probably start from scratch

melloware:
HA you closed this?  want to start it back up again now that 12.0 is out?

Rapster:
Yeah I guess I didn't have much time and the whole restructuring of modules happened... You think it's better to wait for 13.0?

melloware:
well personally i feel like with all our Integration tests now over 640+ and 12.0.0 and 11.0.0 are really solid i feel like you could start this now that we have Integration Test protection.  Just my 2 cents.

Rapster:
Well, it seems my whole idea went caput... I'll try to sum it up the best I can. My idea, was to create to introduce new interfaces `PrimeUIComponent` and `PrimeUICommand` (basically, asbtraction & extension of `UIComponent` & `UICommand` spec for PF). Primary goal was to remove duplication of methods thanks to default methods of java 8, and do a lot more based on that.

I started tested with `Poll` component, and realized `update` instruction was not working (`PrimeUICommand#update`). Long story short, when a facelet component is built, it builds some `MetadataTarget `which defines how to write/access attributes. According to whether it's Mojarra or MyFaces, `update` is not considered as a property of `Poll` (despite it inherits from PrimeUICommand) because `BeanInfo#getPropertyDescriptors` do not return descriptors from interfaces, see [`MetadataTargetImpl`](https://github.com/eclipse-ee4j/mojarra/blob/master/impl/src/main/java/com/sun/faces/facelets/tag/MetadataTargetImpl.java). You have to add additionnal code for that... As it is not a property, when ajax request is built from `CoreRenderer#preConfiguredAjaxRequestBuilder`, the request doesn't contain update instruction (because `Poll#getUpdate` will return null because it doesn't "see" the `PrimeUICommand#getUpdate` method)

So there it is, can't do whatever I had in mind 😞 . Don't know if there is a way to workaround this, but I don't wanna introduce too much code just for this _(unless there is a smart way to go about it)_. Maybe this could be fixed for Faces 4.0, but even if it gets fixed, it doesn't make much of a difference for PF...

PR #7215, opened 2021-04-09
Fix #7212: AutoComplete: Add option to highlight split query string 

### LINKED ISSUE #7212: AutoComplete: Add option to highlight split query string
Our auto complete methods allow you to search on multiple words, split by space. So searching for "pears apples" will match with "apples and pears". I would like to have this as well, as an option, in the query string highlighting. So, "**apples** and **pears**" will be shown.

https://github.com/primefaces/primefaces/blob/4f34341de3ed915d296541c496fb39033727929e/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js#L763-L773

@melloware you can assign me on this

jepsar:
Fix #7212

I don't think we even need an option for this and simply always do it like this.

Screenshot 2021-04-09 at 14 51 24

Tested with showcase `AutoCompleteView`

```java
    public List<Country> completeCountry(String query) {
        List<Country> countries = countryService.getCountries();
        for (String part : query.toLowerCase().trim().split("\\s+")) {
            countries = countries.stream().filter(t -> t.getName().toLowerCase().contains(part)).collect(Collectors.toList());
        }
        return countries;
    }
```

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
+++ b/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
@@ -762,16 +762,19 @@ PrimeFaces.widget.AutoComplete = PrimeFaces.widget.BaseWidget.extend({
 
             //highlight query string
             if(this.panel.children().is('ul') && query.length > 0) {
-                var queryParts = query.trim().replaceAll(/(\s+)/g, ' ').split(' ');
-                for(var i = 0; i < queryParts.length; i++) {
-                    queryParts[i] = PrimeFaces.escapeRegExp(queryParts[i]);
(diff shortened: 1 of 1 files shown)


melloware (approved):

jepsar:
If anything should be configurable, I think it should be a function to take care of the highlighting.

melloware:
Yeah if you broke the highlighting out into its own function it would make it easier to override and customize.

jepsar:
> Yeah if you broke the highlighting out into its own function it would make it easier to override and customize.

@melloware Yes, but I don't see any other way how one would highlight a query... do you? I think I'll just leave it like this.

melloware:
no prob

PR #7221, opened 2021-04-12
Document MixedClientBehaviorHolder

melloware:


tandraschko:
would be great to add the newer information too (see discord)

melloware:
Updated. Let me know what you think.

PR #7234, opened 2021-04-16
added outputConsumer to StreamedContent

### LINKED ISSUE #7233: FileDownload: Write direct to the Outputstream
**Description**
In many applications some data, e.g. csv file will be generated on the fly. so it would be a memory efficent feature to write directly to the outputstream

Maybe a vistor pattern could be the solution, so you can use the stream as is or define a outputstream consumer.

```
 
			return DefaultStreamedContent.builder().contentType("text/csv").name("my.csv")
					.outputStream((outputstream) -> { 
//do something with the outputstream
						}).build();

```

neXus1987:
this allows us to write directly to the responseOutputStream

### DIFF
--- a/src/main/java/org/primefaces/model/DefaultStreamedContent.java
+++ b/src/main/java/org/primefaces/model/DefaultStreamedContent.java
@@ -27,7 +27,9 @@ import org.primefaces.util.Lazy;
 import org.primefaces.util.SerializableSupplier;
 
 import java.io.InputStream;
+import java.io.OutputStream;
 import java.io.Serializable;
+import java.util.function.Consumer;
 
 /**
  * Default implementation of a StreamedContent
@@ -35,11 +37,13 @@ import java.io.Serializable;
 public class DefaultStreamedContent implements StreamedContent, Serializable {
 
     private Lazy<InputStream> stream;
+    
     private String contentType;
     private String name;
     private String contentEncoding;
     private Integer contentLength;
-
+    private Consumer<OutputStream> outputConsumer;
+   
     public DefaultStreamedContent() {
         // NOOP
     }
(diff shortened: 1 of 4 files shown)


melloware:
Can you do a bare minimum PR without any formatting changes?  I see a lot of code has been reformatted.

neXus1987:
sure this should be better

tandraschko:
i think we may also check this for GraphicImage
also please add this to the docs + add some javadoc in StreamedContent

Rapster:
Is it possible to provide a sample of your use case?

neXus1987:
this is my use case:

```
        DefaultStreamedContent.builder().contentEncoding("text/csv").name("my.csv").outputConsumer((os)->{
            try (PrintWriter p = new PrintWriter(new OutputStreamWriter(os, "UTF-8"))) {
             
                    p.println("test-line");
              
            } catch (Exception e1) {
                e1.printStackTrace();
            }
        }).build();
```

Rapster:
@neXus1987 If you're using omnifaces, you can use `Faces#sendFile` which will do exactly what you want and you don't need `p:fileDownload`

tandraschko:
the change makes totally sense for me as we dont need to buffer
however, not sure about the name

maybe just "writer"?
`Consumer<OutputStream> writer`

Rapster:
I was more thinking something like `outputCallback`, but up to you writer sounds good too. Otherwise, why using a fileDownload component over something like `Faces#sendFile`?

Rapster:
Ah maybe for the new ajax feature?

tandraschko:
maybe yep
it also makes sense for normal graphicImage streaming as we dont need to buffer anymore

Rapster:
Yeah I see how useful it is for graphicImage, but not sure regarding fileDownload

src/main/java/org/primefaces/model/DefaultStreamedContent.java
```diff
@@ -110,6 +119,11 @@ public Builder contentLength(Integer contentLength) {
             return this;
         }
 
+        public Builder outputConsumer(Consumer<OutputStream> outputConsumer) {
+            streamedContent.outputConsumer = outputConsumer;
+            return this;
+        }
+
         public DefaultStreamedContent build() {
             return streamedContent;
```
Rapster, line 128: Throws exception if both stream and consumer are not null

tandraschko:
finished by #7295 and #7233 now

tandraschko:
thanks to @neXus1987 for his pre-work

PR #7239, opened 2021-04-20
Fix #7216: Convert LocalDateTime with showTime="false"

### LINKED ISSUE #7216: Datepicker: LocalDateTime with showTime="false" throws converter error
Datepicker with value=LocalDateTime with showTime="false" throws converter error

 PF Version: 10.0.0
 Mojarra 2.2
 Affected browsers:  ALL

1. Create a Datepicker with a LocalDateTime as a value
2. Set showTime="false"
3. Submit Value
4. BaseCalendarRenderer will create a converter exception since the submitted value has no time

**Expected behavior**
Value successfully converts to LocalDateTime with a default time (LocalTime.MIN, etc). A use case would be adding an event to the schedule with all day option where it would be desirable to hide/disregard the time without having to change the backing value type.

melloware:


christophs78 (approved):
Looks good! Thanks for adding an additional test!

src/main/java/org/primefaces/component/calendar/BaseCalendarRenderer.java
```diff
@@ -188,7 +188,7 @@ protected Date convertToLegacyDateAPI(FacesContext context, UICalendar calendar,
     }
 
     protected Temporal convertToJava8DateTimeAPI(FacesContext context, UICalendar calendar, Class<?> type, String submittedValue) {
-        if (type == LocalDate.class || type == YearMonth.class) {
+        if (type == LocalDate.class || type == YearMonth.class || (type == LocalDateTime.class && !calendar.hasTime())) {
```
christophs78, line 191: Hm. Maybe we need to be cautious.
hasTime() looks at the pattern. DatePicker does not require to add HH:mm to pattern for LocalDateTime. You simply  do showTime=true.

melloware: No DatePickerBase overrides it for DatePicker and it checks the `showTime` prop
```java
    @Override
    public boolean hasTime() {
        return this.isShowTime() || this.isTimeOnly();
    }
```

melloware: Now a combination that would not would is LocalDateTime with timeOnly="true" but to me that is right not to make sense.  As you wouldn't set Month Day Year to 0/0/0 to give it a time.

christophs78: We added some magic to this for 10. Not totally sure how this reacts in all possible usecases. Did you check them?

christophs78: Thanks for checking! (I‘m currently on my iPad —> limited ability to check code)
I think we are safe. 😀 
I also agree LocalDateTime combined with timeOnly=true does not make sense.

melloware: I have tested many scenarios and I think this change is a good one and now in the right spot!  Unit tests all passing as well.

PR #7240, opened 2021-04-21
Fix #7222: SelectBooleanButton IL8N on/off labels

### LINKED ISSUE #7222: SelectBooleanButton: make onLabel/offLabel attributes optional
Currently, p:selectBooleanButton requires the attributes onLabel/offLabel. Could you please makes these attributes optional and use the default labels true/false in that case.

This would help to quickly migrate SelectOneCheckboxes by simply replacing h:selectBooleanCheckbox by p:selectBooleanButton.

melloware:
Decided to go with "True" and "False" for defaults?

### DIFF
--- a/docs/11_0_0/components/selectbooleanbutton.md
+++ b/docs/11_0_0/components/selectbooleanbutton.md
@@ -37,8 +37,8 @@ label | null | String | User presentable name.
 onchange | null | String | Callback to execute on value change.
 style | null | String | Inline style of the component.
 styleClass | null | String | Style class of the container.
-onLabel | null | String | Label to display when button is selected.
-offLabel | null | String | Label to display when button is unselected.
+onLabel | null | String | Label to display when button is selected. Default is "On".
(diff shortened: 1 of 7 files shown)


tandraschko:
why not on/off like the name of the property?

melloware:
I like ON/OFF better but the ticket the user was mentioning True and False.  I prefer On and Off.

tandraschko:
lets to on/off

melloware:
Yep just pushed the fix to ON OFF

jepsar:
NL:
primefaces.boolean.LABEL_ON = Aan
primefaces.boolean.LABEL_OFF = Uit

melloware:
@Rapster @stolp if you want to provide German and French translation that would be great for this PR

melloware:
OK updated to `primefaces.switch.` as it is a better name.

stolp:
DE:
primefaces.switch.LABEL_ON = Ein
primefaces.switch.LABEL_OFF = Aus

PR #7251, opened 2021-04-22
Fix #7244: AutoComplete add dropdownAriaLabel attribute

### LINKED ISSUE #7244: AutoComplete (dropdown="true"): Accessibility of button
**Describe the defect**
According to WCAG criterion 4.1.2 all buttons must have an accessible name.

**Environment:**
 - PF Version: _8.0_
 - Affected browsers: _ALL_

**Expected behavior**
The button must have a text in the span ui-button-text or in an aria- * attribute
which could be entered in an attribute of the component.

**Example HTML**
Current 
```html
<button class="ui-autocomplete-dropdown ui-button ui-widget ui-state-default ui-corner-right ui-button-icon-only"
	type="button" role="button" aria-disabled="false">
	<span class="ui-button-icon-primary ui-icon ui-icon-triangle-1-s"/>
	<span class="ui-button-text">&nbsp;</span>
</button>
```
Expected
```html
<button class="ui-autocomplete-dropdown ui-button ui-widget ui-state-default ui-corner-right ui-button-icon-only"
	type="button" role="button" aria-disabled="false" aria-label="Show options">
	<span class="ui-button-icon-primary ui-icon ui-icon-triangle-1-s"/>
	<span class="ui-button-text">Show options</span>
</button>
```

melloware:


melloware:
@Rapster @stolp if you want to provide German and French translation that would be great for this PR

tandraschko:
+1

PR #7254, opened 2021-04-23
#7253 add countBytesAsChars feature to inputText and inputTextarea

### LINKED ISSUE #7253: InputText and InputTextarea: Let the counter count bytes instead of characters
**Description**
I needed the counter in the textarea to count bytes instead of characters. 
I providing a PR as may be some one else find it usefull.
The countBytesAsChars feature is borrowed from the word count plugin for the ckeditor.

OBreidenbach:


christophs78:
Do you use a DB-System where field-lengths are defined in bytes instead of characters? (eg DB2 with UTF-8)

melloware:
We have a lot of legacy Oracle DBs where columns were VARCHAR2(200 byte) instead of (200 char) and so when you paste unicode characters from Word etc it gets a DB contraint error instead of a UI constraint and users get confused!

I actually think this is a good change and we should probably add it to any place we do character counts.  Not sure all where besides InputText and InputTextArea?

OBreidenbach:
We have ORACLE varchar2(4000 char) but oracle is set to max_string_size=STANDARD. So the string can only be 4000 bytes long.

docs/11_0_0/components/inputtext.md
```diff
@@ -83,6 +83,7 @@ title | null | String | Advisory tooltip information.
 type | text | String | Input field type.
 counter | null | String | Id of the label component to display remaining and entered characters.
 counterTemplate | {0} | String | Template text to display in counter. Placeholders: {0} = remaining chars, {1} = length, {2} = maxlength.
+countBytesAsChars | false | Boolean | Should bytes instead of characters be counted.
```
melloware, line 87: Don't you need to add this to InputTextArea docs as well?

src/main/java/org/primefaces/component/inputtextarea/InputTextareaBase.java
```diff
@@ -106,6 +107,14 @@ public void setCounterTemplate(String counterTemplate) {
...
     }
 
+    public boolean getCountBytesAsChars() {
+        return (Boolean) this.getStateHelper().eval(PropertyKeys.countBytesAsChars, false);
+    }
+
+    public void setCountBytesAsChars(boolean countBytesAsChars) {
+        this.getStateHelper().put(PropertyKeys.countBytesAsChars, countBytesAsChars);
+    }
+
```
melloware, line 117: also add to INputText.

src/main/resources/META-INF/primefaces-p.taglib.xml
```diff
@@ -14599,6 +14599,14 @@
...
             <type>java.lang.String</type>
         </attribute>
+        <attribute>
+            <description>
+                <![CDATA[Should bytes instead of characters be counted, default value is false.]]>
+            </description>
+            <name>countBytesAsChars</name>
+            <required>false</required>
+            <type>java.lang.Boolean</type>
+        </attribute>
```
melloware, line 14609: also to InputText

src/main/resources/META-INF/resources/primefaces/forms/forms.inputtextarea.js
```diff
@@ -142,7 +142,22 @@ PrimeFaces.widget.InputTextarea = PrimeFaces.widget.DeferredWidget.extend({
...
+     */
+    countBytes: function(text) {
+        var count = 0, stringLength = text.length, i;
+        text = String(text || "");
+        for (i = 0; i < stringLength; i++) {
+            var partCount = encodeURI(text[i]).split("%").length;
+            count += partCount === 1 ? 1 : partCount - 1;
+        }
+        return count;
+    },
```
melloware: Maybe move this to core.util.js

melloware (changes requested):

jepsar:
Cool! We've using Oracle with a legacy Trinidad application. There we added a validator which is kind of lame.

docs/11_0_0/components/inputtext.md
jepsar: .. and to https://github.com/primefaces/primefaces/blob/master/docs/11_0_0/gettingstarted/whatsnew.md

christophs78:
> We have a lot of legacy Oracle DBs where columns were VARCHAR2(200 byte) instead of (200 char) and so when you paste unicode characters from Word etc it gets a DB contraint error instead of a UI constraint and users get confused!
> 
> I actually think this is a good change and we should probably add it to any place we do character counts. Not sure all where besides InputText and InputTextArea?

Yes it´s a good change.
We just introduced this week our own `@Size` - Validation - Annotation (called @SizeDB2) in our in-house-framework which does this byte-count-validation additional to what `javax.validation.constraints.Size` does.

melloware (approved):

Issue #7051, opened 2021-03-01
Datatable:  Group row can't toggle columns correctly.   


halisyilboga:
**Describe the defect**
Datatable with a column grouprowed can't toggle columns correctly.
**Environment:**
 - PF Version: _8.0_
 - JSF + version: _e.g. Mojarra 2.2.8 or MyFaces 2.3.6_
 - Affected browsers: _e.g. Chrome, iOS, ALL_

**Reproducer**
[https://github.com/halisyilboga/primefaces-test/tree/datatable-grouprow-toggle-report](https://github.com/halisyilboga/primefaces-test/tree/datatable-grouprow-toggle-report)

**To Reproduce**

1. Go to 'datatable' page
2. Click on 'sort name header' if it is not already sorted. Group the rows according to name/ 
3. click toggle button on table and choose name tag to hide.

![Capture](https://user-images.githubusercontent.com/1422498/109492386-f8f11200-7a9b-11eb-9d45-ed9e426e969e.JPG)
![Capture2](https://user-images.githubusercontent.com/1422498/109492402-fee6f300-7a9b-11eb-9187-aba21fa65b88.JPG)

**Expected behavior**
Column grouped should be toggled and only toggled column should hide. But when there is a grouped row it hides more than one column in some rows. Thats why columns are float left arbitrary. You can see detail in the screen shot.  

**Example XHTML**
```html 
<p:dataTable  id="products"
			  value="#{dtBasicView.products}"
			  var="item"
			  rowIndexVar="row"
			  rowStyleClass="table-row-hilite"
			  emptyMessage="#{labels['base.record.notfound']}"
			  style="margin-top: 1em;" sortBy="#{item.name}">
<f:facet name="header">
<div class="p-d-flex p-jc-between p-ai-center">
				List of Products
 <div>
		<p:commandButton id="toggler" type="button" value="Columns" icon="pi pi-align-justify"/>
		<p:columnToggler datasource="products" trigger="toggler"/>
    </div>
</div>
</f:facet>
<p:column headerText="row number" styleClass="to-right" width="25">
	<h:outputText value="#{row + 1}" />
</p:column>

<p:column field="name" headerText="Name" sortBy="#{item.name}" groupRow="true">
	<h:outputText value="#{item.name}" /> 
</p:column>
<p:column field="#{item.code}" headerText="Code" sortable="false" filterable="false">
        <h:outputText value="#{item.code}" />
</p:column>
			       
<p:column headerText="Category">
[...]

PR #7256, opened 2021-04-24
Fix #7255: Column Group breaks visibility

### LINKED ISSUE #7255: Datatable: not visible columns within grouped columns become visible after sorting/filtering
A column with the attribute visible="false" that is part of a grouped column becomes visible after any click on the table, e.g., sort, filter, or page navigation. 

See attached repoducer. Open the page and you see three column where the first two are grouped. When you sort on the first column, the non-visible column become visible such that four columns are now shown (while the table header is kept stable und thus misaligned!)

[p-datatable-groupedcolumn.zip](https://github.com/primefaces/primefaces/files/6366414/p-datatable-groupedcolumn.zip)

The expected behavior is of course that the non-visible column stays non-visible (and it only used for the export).

Many thanks in advance for fixing.

melloware:
Ticket is related to this one: #7051

Basically grouped rows don't play nicely with Column Toggler and column visibility because in the test there are 5 row headers and 3 columns and the code is trying to figure out which ones are visible but the columns don't line up 5 vs 3 so it thinks column 2 is visible when its not.


tandraschko:
+1 if well tested

PR #7265, opened 2021-04-28
#7264 CommandLinkRenderer must not use the title as fallback for the …

### LINKED ISSUE #7264: CommandLink: Renders title as aria-label if aria-label is not defined
**Describe the defect**
The CommandLinkRenderer renderes the title as aria-label attribute if the aria-label attribute is not defined.
https://github.com/primefaces/primefaces/blob/c49696d66ff3216a849a6d81ccd821f553bcda2c/src/main/java/org/primefaces/component/commandlink/CommandLinkRenderer.java#L93-L98

In our application we have visible tags to indicate that the list is filtered with some properties. The link text is the filter value.  These tags are a commandLink to reset the filter. To inform mouse users about this functionality we have a title "reset filter".

With link text und title present a screen reader would read out: Property + reset filter.
But as the aria-label is set to the title, only 'reset filter' will be read out. If we have more of these tags visible all get read out with 'reset filter'.

If aria-label is not present screen reader will read the anchor text + title. 
See table at The Comparison first column JAWS 15: https://www.deque.com/blog/text-links-practices-screen-readers/

We get a failure in the german bitv test: https://testen.bitv-test.de/index.php?a=di&iid=117  (german)
It references https://www.w3.org/WAI/WCAG21/Understanding/label-in-name  (english)
It explains pretty well, how aria-label should be used.
> For this reason, when a visible label already exists, aria-label should be avoided or used carefully, ...

Therefor the title must not be used as fallback for the aria-label attribute.

**Environment:**
 - PF Version: _8.0_, _10.0_, _11.0.SNAPSHOT_
 - JSF + version: _does not matter_
 - Affected browsers: _does not matter_

**Expected behavior**
CommandLinkRenderer must not use the title as fallback for the aria-label attribute.

**Example XHTML**
```html
<h:form id="frmTest">
    <p:commandLink title="reset filter">
       filter value
    </p:commandLink>
</h:form>
```

If desired I can provide a PR.

OBreidenbach:
…aria-label attribute.
#7264

### DIFF
--- a/src/main/java/org/primefaces/component/commandlink/CommandLinkRenderer.java
+++ b/src/main/java/org/primefaces/component/commandlink/CommandLinkRenderer.java
@@ -90,12 +90,7 @@ public class CommandLinkRenderer extends CoreRenderer {
             writer.writeAttribute("id", clientId, "id");
             writer.writeAttribute("href", "#", null);
             writer.writeAttribute("class", styleClass, null);
-            if (link.getAriaLabel() != null) {
-                writer.writeAttribute(HTML.ARIA_LABEL, link.getAriaLabel(), null);
-            }
-            else if (link.getTitle() != null) {
-                writer.writeAttribute(HTML.ARIA_LABEL, link.getTitle(), null);
-            }
+            writer.writeAttribute(HTML.ARIA_LABEL, link.getAriaLabel(), null);
 
             if (ajax) {
                 request = buildAjaxRequest(context, link);


src/main/java/org/primefaces/component/commandlink/CommandLinkRenderer.java
```diff
@@ -90,12 +90,7 @@ public void encodeEnd(FacesContext context, UIComponent component) throws IOExce
             writer.writeAttribute("id", clientId, "id");
             writer.writeAttribute("href", "#", null);
             writer.writeAttribute("class", styleClass, null);
-            if (link.getAriaLabel() != null) {
-                writer.writeAttribute(HTML.ARIA_LABEL, link.getAriaLabel(), null);
-            }
-            else if (link.getTitle() != null) {
-                writer.writeAttribute(HTML.ARIA_LABEL, link.getTitle(), null);
-            }
+            writer.writeAttribute(HTML.ARIA_LABEL, link.getAriaLabel(), null);
```
melloware, line 93: Is this right?  Shouldn't it be...
```java
if (LangUtils.isNotBlank(link.getAriaLabel()) {
   writer.writeAttribute(HTML.ARIA_LABEL, link.getAriaLabel(), null);
}
```

melloware (changes requested):

src/main/java/org/primefaces/component/commandlink/CommandLinkRenderer.java
OBreidenbach: At least the HtmlResponseWriter does a not null check. 
https://github.com/eclipse-ee4j/mojarra/blob/e85dcd755bec27008dc70446b56bd30d148b2f2b/impl/src/main/java/com/sun/faces/renderkit/html_basic/HtmlResponseWriter.java#L712-L714

And in CommandButtonRenderer there is no extra null check.
https://github.com/primefaces/primefaces/blob/c49696d66ff3216a849a6d81ccd821f553bcda2c/src/main/java/org/primefaces/component/commandbutton/CommandButtonRenderer.java#L86

There is not isNotBlack check befor writing arie-label at any other place in primefaces.

melloware (approved):

PR #7283, opened 2021-05-01
Fix #7279: FileUpload: Add dropZone attribute 

### LINKED ISSUE #7279: FileUpload: Add dropZone attribute
Facet content should act as the drop zone for uploading files instead of the default component DOM output

jepsar:
Fix #7279

Showcase: https://github.com/primefaces/primefaces-showcase/pull/358

If `dropZone` attribute is present, automatically sets mode to "advanced", auto to true and dragDropSupport to true.

```xhtml
<h5>Drop Zone</h5>
<div class="ui-fluid">
    <div class="p-field">
        <p:inputTextarea id="customDropZone" widgetVar="textarea"
                         rows="5" value="#{fileUploadView.dropZoneText}"/>
        <small>
            Drop on the text area to upload, or
            <p:link value="choose" onclick="PF('uploadDnd').show();return false" style="font-size:inherit"/>.
        </small>
    </div>
</div>
<p:fileUpload widgetVar="uploadDnd" listener="#{fileUploadView.handleFileUploadTextarea}"
              sizeLimit="100000" allowTypes="/(\.|\/)(gif|jpe?g|png)$/" dropZone="customDropZone"
              style="display: none"/>
```

Screenshot 2021-09-23 at 20 25 22


src/main/java/org/primefaces/component/fileupload/FileUploadRenderer.java
```diff
@@ -151,6 +158,9 @@ protected void encodeAdvancedMarkup(FacesContext context, FileUpload fileUpload)
 
         //buttonbar
         writer.startElement("div", fileUpload);
+        if (fileUpload.hasDropZoneFacet()) {
+            writer.writeAttribute("style", "display:none", null);
+        }
```
jepsar: Not too sure if I like this. It does prevent needing to change any JavaScript... 🤔

jepsar:
I think we also would need an extra event for this to handle errors, for example file being too large.

Also I would like to be able to add a button as an alternative to choose a file on touch devices, but I'm not sure how to do it. Maybe add an attribute dropZoneChooseTarget where we move the the choose button from the button bar to using JavaScript? Or maybe nesting the `p:fileUpload`.. or introducing a new tag. Or simply make the entire zone clickable to trigger the file browser (I think this is the best option). @melloware WDYT?

To do: add `.on('dragenter')` / `.on('dragleave')` to toggle the `ui-active` state.

melloware:
So instead of a facet for DropZone couldn't it just be a selector like `dropZoneTarget="cardId"` ??? That way you could assign the dropzone to any p:panel or DIV you wanted to?

jepsar:
Could be, but I don't want the default UI to show. This could still be done by checking the attribute. Note that I want to add some JS in the mix to set a style class on dragenter and add a click listener to open the file browser. So maybe a facet is the best option.. right?

melloware:
OK yeah I can live with that.  It seems like a good change to me.

src/main/java/org/primefaces/component/fileupload/FileUpload.java
```diff
@@ -60,6 +61,9 @@
 
     public static final String CONTAINER_CLASS_SIMPLE = "ui-fileupload-simple ui-widget";
     public static final String FILENAME_CLASS = "ui-fileupload-filename";
+    public static final String DROPZONE_CLASS = "ui-fileupload-dropzone";
+
+    private Boolean hasDropZoneFacet;
```
tandraschko: avoid local variables in components for MyFaces ViewPooling

jepsar: I've seen that in several components. Should they all be removed?

Just  some random examples:

https://github.com/primefaces/primefaces/blob/4f85a65ff583bf5823df877566210e811d54e6c6/src/main/java/org/primefaces/component/datepicker/DatePickerBase.java#L44

https://github.com/primefaces-extensions/primefaces-extensions/blob/48d5750dd1f5efd38a9e67d4d873aab60c4fcee8/core/src/main/java/org/primefaces/extensions/component/counter/Counter.java#L66

tandraschko: yep, that should be avoided if not to performance risky. Otherwise resetted in saveState.

melloware: @jepsar We can do those others in a separate PR.  For now just only fix your issue for this PR.

jepsar: Yeah, I was planning to go over the ones in PFE in a separate PR

jepsar:
Decided to go with an attribute as @melloware suggested

melloware (approved):
I like the attribute a lot so you can make anything a drop zone for this upload!

docs/11_0_0/components/fileupload.md
```diff
@@ -250,13 +251,18 @@ FileLimit restricts the number of maximum files that can be uploaded.
...
+is `{name} {size}`.
+
+## Custom Drop Zone
+With the `dropZone` attribute your can specify which component should be used as the drop zone for the
+upload component. If you set this attribute, the following will be set by default: `mode="advanced"`,
+ `auto="true"` and `dragDropSupport="true"`.
 
 ## Skinning
-FileUpload resides in a container element which _style_ and _styleClass_ options apply. As skinning
+FileUpload resides in a container element which `style` and `styleClass` options apply. As skinning
```
jepsar, line 319: Note to self: add the new classes to the table below

src/main/java/org/primefaces/component/fileupload/FileUpload.java
jepsar: https://github.com/primefaces-extensions/primefaces-extensions/pull/392

tandraschko:
@jepsar conflicts

jepsar:
> @jepsar conflicts

@tandraschko resolved

tandraschko:
looks good to me
asinging Mert because of stylesheet names

melloware:
Looks like you have a conflict now after our latest fixes.

jepsar:
@melloware resolved

mertsincan:
Hi @jepsar, 

Thanks a lot for your contribution! Everything looks good. ;) @sqores Could you please check in your locale?

Best Regards,

buunsal:
I think the bottom FileUpload panel doesn't need to be rendered when dropzone attribute added. Everything else looks good. @jepsar WDYT?

jepsar:
It does not need to be visible, but it illustrates best what is happening. You do need the component to be rendered though, but you could style it to be `display:none`.

buunsal:
If the panel is displayed this way, users will expect to be able to set the "auto" variable to false. Just like any other dnd demo. But now it works as default auto and users cant change it. So i think if the panel needs to be displayed, we should change this behavior.

jepsar:
Okay, but then you manually need to handle messages like file size exceeds max. Initially I hid the panels, see https://github.com/primefaces/primefaces/pull/7283/commits/21bd535144f86beb76692ffd7ecd70e727ad45c3

Let me convert the example to a text area and handle messages (might require new events).

tandraschko:
@jepsar what about this PR?
i asked primetek to review those 2 "pending-review" PRs before RC1
so if you are ready, please fix conflicts and so on

jepsar:
@tandraschko I'll fix this one. Only showcase is to do. Had some ideas to create a text area, but I guess I'll keep it simple and maybe add a text area example later.

melloware (approved):

mertsincan:
Hi @jepsar, 

Thanks a lot for your contribution! Nice feature ;)

PR #7287, opened 2021-05-02
Resolves #7285

### LINKED ISSUE #7285: PickList: Add localization support
**Description**
With PrimeFaces 10, localization message files are now part of the distribution. Not all components do support that. 
For example, PickList shows fixed tooltip strings like 'Add',, 'Add all' etc. for the control buttons.

As a hack for localization, the following attributes are currently available in PickList: addLabel, addAllLabel, removeLabel, removeAllLabel, moveUpLabel, moveTopLabel, moveDownLabel, moveBottomLabel.

Message resource entries should be added for these labels and used instead of the hardcoded strings.

Additional suggestion: The existing label attributes should be deprecated in version 11 and removed in version 12.

Please comment.

I will follow up with a pull request resolving this.

stolp:


src/main/resources/org/primefaces/Messages_de.properties
```diff
@@ -30,6 +31,14 @@ primefaces.panel.aria.OPTIONS_MENU = Optionsmenü umschalten
 primefaces.panel.aria.TOGGLE = Bedienfeld ein- und ausblenden
 primefaces.password.INVALID_MATCH = {0}: Validierungsfehler.
 primefaces.password.INVALID_MATCH_detail = {0} sollte mit {1} übereinstimmen.
+primefaces.picklist.ADD = Hinzufügen
+primefaces.picklist.ADD_ALL = Alle hinzufügen
+primefaces.picklist.REMOVE = Entfernen
+primefaces.picklist.REMOVE_ALL = Alle entfernen
+primefaces.picklist.MOVE_UP = Nach oben verschieben
+primefaces.picklist.MOVE_TOP = Nach oben verschieben
```
jepsar: Why is the text the same for both?

stolp: My bad. It was way too late yesterday. Thanks for spotting!

stolp: I resolved the translation and added your NL translation as well.

src/main/resources/org/primefaces/Messages_de.properties
```diff
@@ -36,9 +36,9 @@ primefaces.picklist.ADD_ALL = Alle hinzufügen
 primefaces.picklist.REMOVE = Entfernen
 primefaces.picklist.REMOVE_ALL = Alle entfernen
 primefaces.picklist.MOVE_UP = Nach oben verschieben
-primefaces.picklist.MOVE_TOP = Nach oben verschieben
+primefaces.picklist.MOVE_TOP = An den Anfang verschieben
```
jepsar, line 39: FYI in Dutch I left out the equivalent of verschieben (as with German, translations can get long compared to English).

stolp: Your call - But as this is a tooltip there is enough room to be verbose I think

melloware (approved):

jepsar:
Maybe update https://github.com/primefaces/primefaces/blob/master/docs/11_0_0/core/localization.md

melloware:
@stolp While your updatig the 11.0 MD can you add two I forgot to add.
```
primefaces.autocomplete.aria.DROPDOWN_LABEL=Show Options
primefaces.switch.LABEL_ON = On
primefaces.switch.LABEL_OFF = Off
```

stolp:
I'm on it.

stolp:
Done and done.

melloware (approved):

PR #7312, opened 2021-05-11
Fix #7311: ToggleSwitch align with SelectBooleanCheckbox

### LINKED ISSUE #7311: ToggleSwich: Support switching value when clicking on an assigned outputLabel
**Description**
`p:toggleSwitch` does not behave like `p:selectBooleanCheckBox` when activated by clicking on an `p:outputLabel ` with a `for ` attribute set to the id of the ToggleSwitch.

ToggleSwitch should change its value like selectBooleanCheckBox  does when a user clicks on the assigned label.

**Additional context**

**Example XHTML**
```xml
<h:form id="frmTest">
  <p:outputLabel for="@next" value="Click Me" />
  <p:toggleSwitch value="#{bean.value}" />
</h:form>
```

melloware:
Compared with SelectBooleanCheckbox and aligned the two.  It was missing interface `InputHolder` interface which triggers OutputLabel to look for `_input` instead of the main component.


tandraschko:
+1 if well tested

PR #7322, opened 2021-05-16
Fix #7288: Upgrade Printer to Print.js

### LINKED ISSUE #7288: Printer: Unable to print Charts
**Describe the defect**
When printing my page using the p:printer component, Charts objects are missing from the print output.
If I use Ctrl-P, the page is rendered correctly.

**Reproducer**
See: https://github.com/stolp/primefaces-test/tree/PF-7285

**Environment:**
 - PF Version: _10.0_
 - JSF + version: _e.g. Mojarra 2.3.14_
 - Affected browsers: _ALL_

**To Reproduce**

1. Print page via Print Button
2. Print Page via CTRL-P
3. Compare

**Expected behavior**
The page should print correctly.

melloware:
The problem with the jQuery.Print library is it does not support `<canvas>` elements and CHartsJS is a canvas.  This Print.js library is one of the most popular out there and it supports canvas.

I am also adding a Showcase Example to "Charts Export" to show the Print option actually printing the chart correctly now that we are using ChartsJS 3.2


melloware:
@blutorange wouldn't mind a second set of eyes on this to make sure I did the TypeScript correctly as Print.JS has TypeScript included.

melloware:
Print.JS also allows you to delcare a "title" of the Print Window.  I think I am going to add this as a property of the p:print. The default right now is just "Document".

stolp:
I think this would be great. Maybe using the HTML page title as the default is a good idea here?

melloware:
Agreed I think that is what the current component does as on the Showcase the title of the window is "PrimeFaces Showcase".

melloware:
Added.  Defaults to `document.title` if no `title="Bar Chart"` attribute is given on the printer.

src/main/type-definitions/core.d.ts
```diff
@@ -9,6 +9,7 @@
 /// <reference types="jquery" />
 /// <reference types="jqueryui" />
 /// <reference types="moment-timezone" />
+/// <reference types="print-js" />
```
blutorange: @melloware Using declarations from other libraries is pretty confusing at time, and I had to look this up too, but basically : Referencing print-js like this doesn't work because it declares itself as a module, not as a global library (it contains a `export = printJS` statement). What we need to do is to add the global export ourself. Can you please delete this line and add the following (somewhere before `declare namespace PrimeFaces {`)

```typescript
 /**
  * Exposes the print-js library as a global variable. Normally you do not
  * have to use this youself, as PrimeFaces calls it for you when you use
  * the printer component. Also note that while PrimeFaces currently uses
  * this library for printing, this is subject to change in future releases.
  */
 declare const printJS: typeof import("print-js");
```

melloware: Thank you I will!  Much appreciated for the review

melloware: Changes Made!

tandraschko:
+1 if well tested

PR #7326, opened 2021-05-18
Fix #7323: Input widgets add enable()/disable() methods

### LINKED ISSUE #7323: Input Widgets: Support enable() and disable() widget methods
**Description**
I think all Input type widgets should support a client side `enable()` and `disable()` methods.  This not only makes selenium testing easier it also makes the components more scriptable.  I only discovered this while using `p:password` and decided we should review all input ones and add these widget convenience methods.

melloware:
Added utility methods to prevent the same cut and paste code everywhere.


tandraschko:
looks good 
just wonder if we should rename the core methods to e.g. enableInputWidget instead of enableWidget?

melloware:
Good idea. I will rename.

melloware:
Renamed.

Issue #11, opened 2015-04-12
Value attribute to Draggable and Droppable


cagataycivici:
Draggable and Droppable should have value attributes to easily identify what has been dragged and dropped onto.

452:
be wonderful to see some API methods like this:

``` java
public void onDragDrop(DragDropEvent event) {
    List<Car> selectedCarsToMove = (List<Car>) event.getDragObject();
    TreeNode dropNodeCategory = (TreeNode) event.getDropObject();
}
```

for others people old link of discussions:
https://code.google.com/p/primefaces/issues/detail?id=1469
http://forum.primefaces.org/viewtopic.php?f=3&t=5952

452:
on this moment I found only one workaround, but in my opinion this is bad way

``` html
<p:treeNode>
    <h:outputText id="selectedCategoryNode" value="#{node}" />
    <p:droppable for="selectedCategoryNode" scope="tabletotree"
    datasource="nomenclatureTable" activeStyleClass="ui-state-highlight" tolerance="touch">
        <p:ajax listener="#{catalogBean.onDragDropTree}" update="nomenclatureTable msgs"/>
        <f:attribute name="category" value="#{node}" />
    </p:droppable>
</p:treeNode>
<!-- <p:dataTable id="nomenclatureTable" -->
<p:column id="drag" style="width:20px">
    <h:outputText id="dragIcon" styleClass="ui-icon ui-icon-arrow-4" />
    <p:draggable for="dragIcon" scope="tabletotree" revert="true" helper="clone" />
</p:column>
```

``` java
    public void onDragDropTree(DragDropEvent event) {
        FacesMessage message;
        try {
            Nomenclature nomenclature = (Nomenclature) event.getData();
            Nomenclature category;

            if (event.getComponent().getAttributes().get("category") instanceof String) {
                category = new Nomenclature(0L, "Номенклатура");
            } else {
                category = (Nomenclature) event.getComponent().getAttributes()
                        .get("category");
            }

            nomenclatureService.move(category, nomenclature);
            nomenclatureList = nomenclatureService.getNomenclature(parentId);
            message = new FacesMessage(FacesMessage.SEVERITY_INFO,
                    "Переміщення", "Номенклатура " + nomenclature
                            + " переміщена в категорію " + category);
        } catch (Exception e) {
            message = new FacesMessage(FacesMessage.SEVERITY_ERROR, "Помилка",
                    "Виникла помилка при переміщенні");
        }
[...]

PR #7339, opened 2021-05-22
#7336: DataTable: Removed data (rows) remain after update when filtered before

### LINKED ISSUE #7336: DataTable: Removed data (rows) remain after update when filtered before
**Describe the defect**
When using a List (for example ArrayList) of DTOs for a dataTable and removing some elements from the list, updating the table throw ajax, everythink works fine.
But when I first filter the rows inside the dataTable throw column or global Filter and then remove some of the still visible elements, they remain after ajax update. But the elements are arleady removed from the Java List.

The Bug was also discussed in the forum: https://forum.primefaces.org/viewtopic.php?f=3&t=66918

**Reproducer**
Checkout https://github.com/MatthiasPischka/primefaces-test.git and run as described (mvn jetty:run for example).
After starting, visit the page http://localhost:8080/primefaces-test/data-table-remove-bug.xhtml (you get there throw navigation from default page, too).
Working and buggy example are described on the page.

**Environment:**
 - PF Version: _10.0.0_

christophs78:
May be called after adding, removing or updating rows.

```java
DataTable dataTable = (DataTable)FacesContext.getCurrentInstance().getViewRoot().findComponent("form:dt-products");
dataTable.reapplyFilter();
```

If we like this, we should add it to docu.

And we need to check corner cases like paging. (eg 10 rows per page, 11 rows in list, user on page 2, we remove row #11, ...)

### DIFF
--- a/src/main/java/org/primefaces/component/datatable/DataTable.java
+++ b/src/main/java/org/primefaces/component/datatable/DataTable.java
@@ -527,6 +527,14 @@ public class DataTable extends DataTableBase {
         setScrollOffset(0);
     }
 
+    /**
+     * Recalculates filteredValue (for a non-lazy value) after adding or removing rows to a filtered DataTable
+     */
+    public void reapplyFilter() {
+        FilterFeature filterFeature = (FilterFeature) getFeature(DataTableFeatureKey.FILTER);
+        filterFeature.filter(FacesContext.getCurrentInstance(), this);
+    }
+
     public RowExpansion getRowExpansion() {
         for (int i = 0; i < getChildCount(); i++) {
             UIComponent child = getChildren().get(i);
--- a/src/main/java/org/primefaces/component/datatable/DataTable.java
+++ b/src/main/java/org/primefaces/component/datatable/DataTable.java
@@ -528,11 +528,13 @@ public class DataTable extends DataTableBase {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


tandraschko:
@christophs78 can we close this?

christophs78:
Hm. Do we wan´t this as convenience-method instead of manually updating filteredValue after adding/removing/updating rows from DataTable-value?

tandraschko:
is there any case where a user must manually invoke this method?

christophs78:
> is there any case where a user must manually invoke this method?

https://github.com/primefaces-extensions/primefaces-integration-tests/blob/master/src/main/java/org/primefaces/extensions/integrationtests/datatable/DataTable021.java#L66

tandraschko:
we should at least find a better name/comments as it should only be used for non-lazy tables

christophs78:
I´m open to for a better name/JavaDoc....
`reapplyFilterAndSortForNonLazyDataTable`
`syncFilterAndSortForNonLazyDataTable`
`syncFilterValueForNonLazyDataTable`
???

tandraschko:
maybe we can just add a IF-statement inside? if (lazy) { throw Exception or just return } ?

and add it to the javadoc

tandraschko:
commited it with the additional IF

PR #7340, opened 2021-05-22
Fix #7338: ToolbarGroup: Add CSS style to fix align right 

### LINKED ISSUE #7338: ToolbarGroup: Align="right" doesn't work
**Describe the defect**
In Primefaces 8 I had many toolbars that have only buttons aligned to right, like the one below:

image

After updating to Primefaces 10, the alignment to right doesn't work anymore:

image

**Reproducer**
https://github.com/hfluz/primefaces-test

**Environment:**
 - PF Version: 10.0.0 and 11.0.0-SNAPSHOT
 - Affected browsers: all

**To Reproduce**
1. Just open the reproducer's test page.

**Expected behavior**
Align="right" property should be respected, but it's ignored.

Instead, the only way to align a toolbarGroup to the right is adding another toolbarGroup first, since the second one will be aligned to right. Since I didn't find anything about this change in the migration guide, I assume it wasn't intended.

I belive this is the commit that broke it: https://github.com/primefaces/primefaces/commit/5365f7c0c17c6cf6b6e03b0c80eb0abd9b0ef2d0#diff-29a14603d40c8eed580bde9d89fa970008bfe6763f254edd13682ce96e21831b

ui-toolbar-group-left and ui-toolbar-group-right have exactly the same styles after this change.

hfluz:
Added CSS style to fix issue https://github.com/primefaces/primefaces/issues/7338.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces/toolbar/toolbar.css
+++ b/src/main/resources/META-INF/resources/primefaces/toolbar/toolbar.css
@@ -10,3 +10,6 @@
     align-items: center;
     flex-wrap: wrap;
 }
+.ui-toolbar-group-right {
+    margin-left: auto;
+}


melloware:
@sqores please review since this was affected by your change...

buunsal:
Although the toolbar group on the left is empty, it needs to be rendered. Issue stems from this. No need to add an extra css. I'll fix it next week.

buunsal:
@hfluz fixed with https://github.com/primefaces/primefaces/pull/7347

PR #7346, opened 2021-05-24
Fix #7345: add resizeEnd event to the Splitter

### LINKED ISSUE #7345: Splitter: add resizeEnd event and onResizeEnd callback

buunsal:


src/main/java/org/primefaces/component/splitter/Splitter.java
```diff
@@ -38,4 +51,52 @@
     public static final String LAYOUT_HORIZONTAL_CLASS = "ui-splitter-horizontal";
     public static final String GUTTER_CLASS = "ui-splitter-gutter";
     public static final String GUTTER_HANDLE_CLASS = "ui-splitter-gutter-handle";
+
+    private static final String DEFAULT_EVENT = "onResizeEnd";
+    private static final Map<String, Class<? extends BehaviorEvent>> BEHAVIOR_EVENT_MAPPING = MapBuilder.<String, Class<? extends BehaviorEvent>>builder()
+            .put("onResizeEnd", SplitterResizeEvent.class)
```
melloware: I think here its just "resizeEnd" we don't put "on" here for AJAX events so it will be...
```xml
<p:ajax event="resizeEnd" update="" />
```

src/main/java/org/primefaces/component/splitter/Splitter.java
```diff
@@ -38,4 +51,52 @@
...
+    @Override
+    public void queueEvent(FacesEvent event) {
+        FacesContext context = getFacesContext();
+        Map<String, String> params = context.getExternalContext().getRequestParameterMap();
+
+        if (ComponentUtils.isRequestSource(this, context)) {
+            String eventName = params.get(Constants.RequestParams.PARTIAL_BEHAVIOR_EVENT_PARAM);
+            String clientId = getClientId(context);
+
+            if ("onResizeEnd".equals(eventName)) {
```
melloware: Use the CONSTANT for DEFAULT_EVENT here.

src/main/java/org/primefaces/component/splitter/Splitter.java
```diff
@@ -38,4 +51,52 @@
...
+                AjaxBehaviorEvent behaviorEvent = (AjaxBehaviorEvent) event;
+                List<Float> panelSizes = new ArrayList<>();
```
melloware: We usually don't use Floats we use Double.

src/main/resources/META-INF/resources/primefaces/splitter/splitter.js
```diff
@@ -174,6 +175,18 @@ PrimeFaces.widget.Splitter = PrimeFaces.widget.BaseWidget.extend({
 
         this.gutterElement.removeClass('ui-splitter-gutter-resizing');
         this.jq.removeClass('ui-splitter-resizing');
+
+        if(this.hasBehavior('onResizeEnd')) {
```
melloware: Here is also just resizeEnd

src/main/resources/META-INF/resources/primefaces/splitter/splitter.js
```diff
@@ -174,6 +175,18 @@ PrimeFaces.widget.Splitter = PrimeFaces.widget.BaseWidget.extend({
...
+            var sizesArr = this.panelSizes;
+            var ext = {
+                params: [
+                    {name: this.id + '_panelSizes', value: sizesArr.map(e => e.toFixed(2)).join('_')},
+                ]
+            };
+
+            this.callBehavior('onResizeEnd', ext);
```
melloware: Also here

src/main/java/org/primefaces/component/splitter/Splitter.java
melloware: Also use Constant here in the BEHAVIOR_EVENT_MAPPING use the DEFAULT_EVENT.

melloware (changes requested):

docs/11_0_0/components/splitter.md
```diff
@@ -149,6 +150,15 @@ local storage which can be configured using the ```stateStorage``` property.
...
 ```
 
+## Ajax Behavior Events
+The following AJAX behavior event are available for this component. If no event is specified the default event is called.
+
+**Default Event:** `onResizeEnd`
+
+| Event | Listener Parameter | Fired |
+| --- | --- | --- |
+| onResizeEnd | org.primefaces.event.SplitterResizeEvent | Is fired after the resizing end. |
```
melloware: Just `resizeEnd` here

melloware (changes requested):

src/main/java/org/primefaces/component/splitter/SplitterRenderer.java
```diff
@@ -122,7 +122,7 @@ protected void encodeScript(FacesContext context, Splitter splitter) throws IOEx
                 .attr("stateKey", splitter.getStateKey())
                 .attr("stateStorage", splitter.getStateStorage());
 
-        wb.callback("onResizeEnd", "function(event,ui)", splitter.getOnResizeEnd());
+        wb.callback("resizeEnd", "function(event,ui)", splitter.getOnResizeEnd());
```
melloware: I think this is still supposed to be `onResizeEnd` since this is the client side JS callback.

So client side would be `<p:splitter onresizeend="alert('test');" />` but the AJAX is `event="resizeEnd`.  Also I think the Client Side events are usually all lowercase like `onresizeend`

melloware (changes requested):

src/main/java/org/primefaces/component/splitter/SplitterRenderer.java
buunsal: I did that because the event name is "onResizeEnd" in primeng or primereact. And I think also "resizeEnd" :D If you think okay, I will merge

melloware: I think in most PF components the AJAX is always without "on" and the client side events always have "on" so we should keep that consistent.

buunsal: I agree. I wasn't sure about the naming anyway. In the end, I could not be sure which one should be what. And I did it all the same. To fix it when you notice

melloware (approved):

src/main/java/org/primefaces/component/splitter/SplitterRenderer.java
buunsal: We reviewed the codes with Mert and decided to do "onResizeEnd". The reason is that this is not html event. We have used camel cases in many places. accordion panel etc.

melloware: OK

PR #7348, opened 2021-05-24
Fix #7325: MultiViewState restore global filter

### LINKED ISSUE #7325: DataTable/TreeTable: Global Filter not preserved by MultiViewState
**Describe the defect**
As discussed earlier in the primefaces forum (https://forum.primefaces.org/viewtopic.php?f=3&t=66857): When i am using the globalFilter IntputField in combination with the MultiViewState, the table entries are still filtered after revisiting the same site but the value (search keywords) of the globalFilter InputField is resettet. 

**Reproducer**
Visit the PrimeFaces Showcase for example: https://www.primefaces.org/showcase/ui/data/datatable/multiViewState.xhtml, enter something at the right in the global filter field (first dataTable on page), go to another page in Showcase and come back again. The entries are still filtered but the global input field is resetted.

**Environment:**
 - PF Version: _10.0.0_

**To Reproduce**
See above (reproducer)

**Expected behavior**
The entered value should be preserved in case of using MultiViewState

melloware:
Tested using Showcase: https://www.primefaces.org/showcase/ui/data/datatable/multiViewState.xhtml

Type "Argen" for Argentine watch the table get filtered.  Leave the page and come back and the Global Filter has "Argen" in it.

### DIFF
--- a/src/main/java/org/primefaces/component/api/UITable.java
+++ b/src/main/java/org/primefaces/component/api/UITable.java
@@ -160,7 +160,7 @@ public interface UITable<T extends UITableState> extends ColumnAware, MultiViewS
             // #7325 restore global filter value
             if (FilterMeta.GLOBAL_FILTER_KEY.equals(entry.getKey())) {
                 UIComponent globalFilterComponent = SearchExpressionFacade
-                            .resolveComponent(context, (UIComponent) this, FilterMeta.GLOBAL_FILTER_KEY, SearchExpressionUtils.SET_NONE);
+                            .resolveComponent(context, (UIComponent) this, FilterMeta.GLOBAL_FILTER_COMPONENT_ID, SearchExpressionUtils.SET_NONE);
                 if (globalFilterComponent != null && globalFilterComponent instanceof ValueHolder) {
                     ((ValueHolder) globalFilterComponent).setValue(entry.getValue().getFilterValue());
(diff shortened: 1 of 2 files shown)


src/main/java/org/primefaces/component/api/UITable.java
```diff
@@ -203,7 +211,7 @@ default void updateFilterByWithGlobalFilter(FacesContext context, Map<String, Fi
                 ? SearchExpressionUtils.SET_IGNORE_NO_RESULT
                 : SearchExpressionUtils.SET_NONE;
         UIComponent globalFilterComponent = SearchExpressionFacade
-                .resolveComponent(context, (UIComponent) this, "globalFilter", hint);
+                .resolveComponent(context, (UIComponent) this, FilterMeta.GLOBAL_FILTER_KEY, hint);
```
tandraschko: Please create a new constant for that string: GLOBAL_FILTER_COMPONENT_ID

Thats not the map key

melloware: Done

tandraschko:
please put in UITable, the component isnt really related to the FilterMeta pojo

melloware:
Good suggestion.  Moved.

tandraschko:
looks good to me

PR #7355, opened 2021-05-26
Fix #7354: Media/PhotoCam remove Adobe Flash support

### LINKED ISSUE #7354: MediaPlayer/PhotoCam: remove Flash support
> With Flash's EOL announced, many browsers took steps to gradually restrict Flash content (caution users before launching it, eventually blocking all content without an option to play it). By January of 2021, all major browsers started blocking Flash without ability to play it. Only IE11, niche browser forks, and some browsers built for China plan to continue support. Furthermore, excluding the China variant of Flash, Flash execution software has a built-in kill switch which prevents it from playing Flash after January 12, 2021.[65] In January 2021, Microsoft released an optional update KB4577586 which removals Flash Player from Windows; in July 2021 this update will be pushed out as a security update and applied automatically to all remaining systems.[66]

https://en.wikipedia.org/wiki/Adobe_Flash

melloware:
Removes all flash support. Tested with showcase and Media and PhotoCam are working properly.


docs/migrationguide/11_0_0.md
```diff
@@ -42,3 +36,15 @@ and you can remove:
   - `setLineTension` was renamed to `setTension` (`LineChartDataSet`, `RadarChartDataSet`, `ElementsLine`)
   - `LineChartDataSet#getSteppedLine` was renamed to `LineChartDataSet#getStepped`
   - `PolarAreaChartOptions#startAngle` was moved to `RadialScales#startAngle`
+  
+## Dialog
```
jepsar, line 40: +1 for alphabet :-)

jepsar (approved):

PR #7365, opened 2021-05-27
Fix #7344: HotKey: Add bindMac attribute

### LINKED ISSUE #7344: HotKey: allow two different combinations for Mac and PC
On Mac you might want to use `command+e` (`meta+e`) while on PC you would want to use `ctrl+e` for the same action.

We could do:

1. introduce a new attributes `bindMac` and use `bind` for PC
2. introduce new attributes `bindMac` and `bindPC` and deprecate `bind`
3. allow JSON in `bind`, for example `{pc:'ctrl+e',mac:'meta+e'}`
4. introduce a new key name, like `metaCtrl` which translated to `command` or `ctrl` based on the OS

jepsar:
Fix #7344


src/main/java/org/primefaces/util/AgentUtils.java
```diff
@@ -30,30 +30,35 @@
...
-            return userAgent.contains("MSIE");
-        }
+    public static boolean isIE(FacesContext context) {
+        return userAgentContains(context, "MSIE");
     }
 
+    /**
+     * @deprecated it is unused
+     */
+    @Deprecated(since = "11.0.0")
```
jepsar: Or should I simply delete these methods? Or will that be part of removing IE support?

melloware: We will make them part of the other ticket. I will add a note there.

melloware (approved):

PR #7385, opened 2021-06-02
Fix #7384: Spinner buttons only on left click

### LINKED ISSUE #7384: Spinner: Only respect left click on buttons
**Description**
Spinner should only allow increment/decrement buttons when pressed with the left click not the right click or middle click.

Similar to: https://github.com/primefaces-extensions/primefaces-extensions/issues/423

melloware:


src/main/resources/META-INF/resources/primefaces/spinner/spinner.js
```diff
@@ -148,7 +152,6 @@ PrimeFaces.widget.Spinner = PrimeFaces.widget.BaseWidget.extend({
                 case keyCode.RIGHT:
                 case keyCode.TAB:
                     return;
-                break;
```
melloware, line 151: Break was marked as unreachable code.

PR #7386, opened 2021-06-02
#7339 DataTable:filterAndSort - only apply filter (and sort?) when DataTable is already filtered and/or sorted

christophs78:
Not sure how to do this. Just tried something.

### DIFF
--- a/src/main/java/org/primefaces/component/datatable/DataTable.java
+++ b/src/main/java/org/primefaces/component/datatable/DataTable.java
@@ -1116,8 +1116,10 @@ public class DataTable extends DataTableBase {
             return;
         }
 
-        FilterFeature filterFeature = (FilterFeature) getFeature(DataTableFeatureKey.FILTER);
-        filterFeature.filter(FacesContext.getCurrentInstance(), this);
+        if (getFilteredValue() != null) {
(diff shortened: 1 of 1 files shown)


tandraschko:
no idea, i still dont unterstand the manual method exactly ;D

christophs78:
Hm. Think about a CRUD-example like https://primefaces.org/diamond/crud.xhtml.
Backed by a `LazyDataModel` you have no issues. DataTable load´s during rendering what it needs.
Backed by eg ArraryList it get´s a bit complicated. As long as no filter is applied we only have one list. But after a filter is applied we have two list (value-list and filteredValue-list). So when we e.g. add a new row/object to our (primary) (value-)list DataTable won´t reflect this when it´s filtered. You also have to sync filteredValue-list in your project-code. (Because FilterFeature is only executed when the client modifies the filter.)
So IMO it makes sense to let the application developer trigger (in some way) the sync of filteredValued.
The current implementation of `filterAndSort` does this by calling the matching methods of `FilterFeature `and `SortFeature`.
Do we have a similar view of what we want to achieve? (IMO `LazyDataModel` should be preferred. But as long as we offer simple binding to List people use this and expect it doing the "right thing" when used together with `FilterFeature`.)

tandraschko:
will close this for now
please come back if you finished your ideas :)

PR #7394, opened 2021-06-08
Release Profile and Instructions

melloware:
- Added release instructions for PrimeTek to README.MD
- Consolidated Release Profile to only have the GPG plugin in the parent pom
- Removed all TABS from pom.xml's so its only spaces
- WAR Plugin 3.3.1 which is latest for Java 16 hopefully.


melloware:
Confirmed this fixes Java16!

PR #7396, opened 2021-06-08
Fix #7395: Firefox FOUC Fix

### LINKED ISSUE #7395: MoveScriptsToBottom: Firefox flash of unstyled content (FOUC)
**Description**
Optimizing a PrimeFaces app using the following tricks...

- PrimeFaces MOVE_SCRIPTS_TO_BOTTOM
- OmniFaces CombinedResourceHandler
- PrimeFaces Extensions CombinedResourceHandler Helper https://github.com/primefaces-extensions/primefaces-extensions/issues/293

Causes the Firefox Flash of Unstyled Content (FOUC) issue.   If you go to the PFE Showcase using Firefox: https://www.primefaces.org/showcase-ext/views/home.jsf and then start bouncing around different links you see a flashing effect. It looks like the page is being rendered without CSS and then the CSS is applied.  It is not an issue on any Chrome based browser or IE.  Its specific to Firefox

Firefox Issue: https://stackoverflow.com/questions/21147149/flash-of-unstyled-content-fouc-in-firefox-only-is-ff-slow-renderer

Firefox Bug open for years: https://bugzilla.mozilla.org/show_bug.cgi?id=1404468

OmniFaces Issue: https://github.com/omnifaces/omnifaces/issues/632

melloware:
Renders `<body><script>/*FIREFOX_FOUC_FIX*/</script>` right after the BODY tag only if browser is Firefox.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/application/resource/MoveScriptsToBottomResponseWriter.java
+++ b/primefaces/src/main/java/org/primefaces/application/resource/MoveScriptsToBottomResponseWriter.java
@@ -23,9 +23,11 @@
  */
 package org.primefaces.application.resource;
 
+import org.primefaces.util.AgentUtils;
 import org.primefaces.util.LangUtils;
 
 import javax.faces.component.UIComponent;
+import javax.faces.context.FacesContext;
 import javax.faces.context.ResponseWriter;
 import javax.faces.context.ResponseWriterWrapper;
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


tandraschko:
and that really fixes the issue? :D

melloware:
Yes from everything I read on the Firefox bug ticket it is something to do with their Developer Tools and the two teams have been arguing for years over whether its the browser team bug or developer tools team bug.  But either way by forcing the small empty <script> in there it tells the Developer Tools not to render until it has all CSS content.

tandraschko:
ok and the PFE showcase works fine then?

melloware:
> ok and the PFE showcase works fine then?

Yep with this fix the PFE showcase renders properly in Firefox without the flicker.

PR #7400, opened 2021-06-09
Fix #7387: DatePicker/Calendar do not apply mask if disabled

### LINKED ISSUE #7387: Calendar/DatePicker: Disabled or ReadOnly Mask breaks the page
**Describe the defect**
When the p:calendar component has readonly or disabled properties set to true the action attribute of p:commandButton no longer works.

I reported similar issue in the following post. https://forum.primefaces.org/viewtopic.php?f=3&t=66983 which resulted in the following issue [https://github.com/primefaces/primefaces/issues/7362](url)

**Reproducer**
[primefaces-test.zip](https://github.com/primefaces/primefaces/files/6588003/primefaces-test.zip)

**Environment:**
PF Version: 10.0
JSF + version: ALL
Affected browsers: ALL

Forum Post: [https://forum.primefaces.org/viewtopic.php?f=3&t=67098](url)

**To Reproduce**
1. Go to URL http://localhost:8080/primefaces-test-1.0-SNAPSHOT/index.xhtml
2. Click on button "Salvar" Save in my language
3. The action method should be executed.

**Expected behavior**
Show FacesMessage

**Example XHTML**
```<h:form id="form">
		
			<p:messages id="mensagem" showDetail="false" autoUpdate="true"/>
			
			<p:calendar id="dataVencimento" value="#{alunoMB.datnas}" pattern="dd/MM/yyyy" mask="true" locale="pt" readonly="true"/>
			
			<p:commandButton id="salvar" value="Salvar" icon="imagem-email-excluir" action="#{alunoMB.salvar}" />
								
			
		</h:form>
```

**Example Bean**
```
public void salvar() {

		FacesMessage message = new FacesMessage(FacesMessage.SEVERITY_INFO, "Metódo chamado", null);
		context.addMessage("Método chamado", message);

		System.out.println("------------------------------------");

	}
```

melloware:


educalize:
I downloaded the source code from the master. Generated primefaces-11.0.0-SNAPSHOT.jar. I installed it in my project and the components still have problems, not opening the popup to select the date and other features such as selecting the hour and minute. The previous problem, which was a problem with the use of a mask, has apparently been solved. It remains these problems that I reported.

melloware:
Yep then you definitely have other unrelated issue.  Please create a PrimeFaces Test reproducer for me to debug that actually runs with `mvn clean jetty:run` or else there is nothing I can do.

PR #7405, opened 2021-06-10
Fix #7398: Inplace: add keyboard support

### LINKED ISSUE #7398: Inplace: add keyboard support
Output should be focusable and enter should trigger a click.

See also: https://stackoverflow.com/questions/67901875/how-to-tab-correctly-to-pinplace-component-in-jsf-so-screenreader-can-read-the

jepsar:
Fix #7398


tandraschko:
code looks good
i think you tested well in all browsers?

jepsar:
> i think you tested well in all browsers?

Chrome, Firefox and Safari on MacOS

PR #7415, opened 2021-06-10
Test #7413: Datatable FilterFunction not called

### LINKED ISSUE #7413: DataTable: Column FilterFunction not being called
Describe the defect
filterFunction of column in dataTable not called when

    <context-param>
        <param-value>client</param-value>
    </context-param>

is set to "client"

Reproducer

https://github.com/mafulafunk/primefaces-test

Branch that works: master
Branch showing the issue: issue

Environment:

PF Version: 10.0.0
JSF + version: Default as in Reproducer
Affected browsers: Safari

In the reproducer got to:
http://localhost:8080/primefaces-test/

When the filter method is called. System out printlns are thrown out on the terminal.

There is explicitly no filter logic implemented in the method, it was only meant to show that toggling the context-parameter exposes the behaviour of not calling the "filterFunction"

melloware:
Integration test proving the column FilterFunction is not being called.

### DIFF
--- a/docs/10_0_0/components/column.md
+++ b/docs/10_0_0/components/column.md
@@ -19,47 +19,47 @@ treetable and more.
 | id | null | String | Unique identifier of the component
 | rendered | true | Boolean | Boolean value to specify the rendering of the component, when set to false component will not be rendered.
 | binding | null | Object | An el expression that maps to a server side UIComponent instance in a backing bean
-| style | null | String | Inline style of the column.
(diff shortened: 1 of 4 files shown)


melloware:
Failing as expected proving the columnFilterFunction is not called.
```
Failures: 
Error:    DataTable019Test.testColumnFilterFunctionByType:134->
            assertMessage:139 Messages was empty when a message was expected! ==> expected: <1> but was: <0>
```

PR #7418, opened 2021-06-11
Fix #6956: Drop Internet Explorer support

### LINKED ISSUE #6956: Core: Drop Internet Explorer support
With Internet Explorer dying in foreseeable future we may discuss dropping support with a future PrimeFaces-release.

This would allow use to use more modern JavaScript for new features and/or components. Maybe ES6 would be a proper base-level for some time.
Thanks to our integration-tests we have a good base to check compatibility against the 3 most wide-spreaded browsers. (Chrome, Safari and Firefox)

This suggestion is somehow inspired by Cagatay Civici who spoke about plans to replace JS-based animations within PrimeFaces with up2date CSS-animations.

melloware:
image


tandraschko:
must be discussed with PrimeTek

mertsincan:
Hi,
I think PrimeFaces' browser support is related to Jquery and Jquery UI. Jquery currently supports IE 9+.
https://jquery.com/browser-support/
https://jqueryui.com/browser-support/
That's why I'm not sure about this PR.

tandraschko:
@mertsincan what do you mean exactly?
We can decide independent of jQuery if our code supports IE or not.
If we remove IE hacks, the underlying jquery might still support IE or not, it doesnt matter.

tandraschko:
about the polyfills...
not sure if we can remove it
we also have some older ipad here, where replaceAll was not available

melloware:
I removed it from this file since in the latest master you have the new special polyfill file.  I looked up most of them and only ReplaceAll is needed by anyting modern.  All the other polyfills were just to support IE.

melloware:
I rebased with master.

christophs78:
Should we move this one to 12.0.0?

melloware:
Moved the PR and the Ticket to 12.0.0

tandraschko:
Yep

jepsar:
@melloware should this be mentioned in the migration guide?

melloware:
Yep updating now.

PR #7420, opened 2021-06-12
Fix #7413 - DataTable: Column FilterFunction not being called

### LINKED ISSUE #7413: DataTable: Column FilterFunction not being called

Rapster:
`FilterMeta#constraint` gets lost after deserialization as it is transient, then `FilterConstraint` must be serializable (as of today, only built-in constraint were retrieved after deserializaton)

### DIFF
--- a/primefaces/src/main/java/org/primefaces/model/FilterMeta.java
+++ b/primefaces/src/main/java/org/primefaces/model/FilterMeta.java
@@ -25,6 +25,7 @@ package org.primefaces.model;
 
 import org.primefaces.component.api.DynamicColumn;
 import org.primefaces.component.api.UIColumn;
+import org.primefaces.component.api.UITable;
 import org.primefaces.component.column.ColumnBase;
 import org.primefaces.component.datatable.feature.FilterFeature;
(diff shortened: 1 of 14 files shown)


melloware (approved):
Thanks for fixing! I was looking into this and couldn't quite figure out what was wrong.  The transient makes sense losing the constraint

melloware:
I am going to merge this then rebase my PR with your changes and verify the Integration Test passes.

PR #7428, opened 2021-06-15
Fix #7193: DataTable globalFilterOnly attribute

### LINKED ISSUE #7193: DataTable/TreeTable: GlobalFilter not working with filterable="false"
**Describe the defect**
Global filter works only on p:column that has atribute `filtarable="true"`, but it should work on every column that has `filterBy=""` atribute.
I have a datatable where I need only the global filter, not the column filters. To disable column filter I use `<p:column filterable="false"/>`, but this breaks the GlobalFilter in PF10, in PF8 it works as expected.
I haven't tested yet, but this may affect treetable too.

**Reproducer**
Clone repository https://github.com/stefan-dan/primefaces-test

**Environment:**
 - PF Version: _10.0_
 - JSF + version: _Mojarra 2.2.8 and MyFaces 2.3.6_
 - Affected browsers: _ALL_

**To Reproduce**
1. Go to 'http://localhost:8080/primefaces-test/'
2. Click on 'Global filter _Enter keyword'_
3. Write 'Stefan'
4. Works as expected
5. Write 'Moldova'
6. See error

**Expected behavior**
Expected behaviour: Global filter should work on all p:columns that has atribute filterBy="", despite of filterable="" value.

**Example XHTML**
```html
<p:dataTable var="customer" value="#{dtFilterView.customers1}" widgetVar="customersTable"
                             filteredValue="#{dtFilterView.filteredCustomers1}">

                    <f:facet name="header">
                        <div class="p-d-flex p-ai-center p-jc-between">
                            <span>List of Customers</span>
                            <p:inputText id="globalFilter" onkeyup="PF('customersTable').filter()" style="width:10rem" placeholder="Enter keyword" />
                        </div>
                    </f:facet>

                    <p:column headerText="Name" sortBy="#{customer.name}" filterBy="#{customer.name}" filterable="false">
                        <h:outputText value="#{customer.name}" />
                    </p:column>
</p:dataTable>
```

melloware:
Tested with reproducer:
[pf-7193.zip](https://github.com/primefaces/primefaces/files/6655980/pf-7193.zip)


tandraschko:
can you also add tests? or do we have already?

melloware:
Yep we have tests but let me add 1 more test to test this specific scenario.

melloware:
OK added new Integration Test that test the global filter against a column with `filterDisplayed="false"` to make sure the column is filtered properly.

melloware:
Also added same fix to TreeTable.

tandraschko:
@Rapster WDYT about the attr?

Rapster:
Looks good but it would be simpler to have an attr `globalFilterOnly` on dataTable no?

melloware:
But GlobalFilter can be used two different ways and we have Inegration Tests for both.  In the case where a user wants their columns to be filterable by Global Filter you respect column `filterable="true"` to allow which columns get included in the global Filter. So there is no way 1 attribute at the table level would work right?

This scenario is the customer wants a column filterable by gloabl filter but NOT displayed as an individual column filter.  I think the only way to be flexible is to have an attribute at the `column` level.

Rapster:
Having `globalFilterOnly` would only prevent display of column filters, not initialization of these.

> This scenario is the customer wants a column filterable by gloabl filter but NOT displayed as an individual column filter. I think the only way to be flexible is to have an attribute at the column level.

Only one or all?

melloware:
Yep but what if a customer wants all filters in the global filter but one of the columns to have also a column filter.  That is how I was testing and what our Integration test does.  It disallows a single filter from being a column filter but it is part of the global filter.  We just won't know exactly all combinations someone might want.

Rapster:
> Yep but what if a customer wants all filters in the global filter but one of the columns to have also a column filter

You mean, all columns have their filters displayed except one, and all of them should be taken account in global filter? (i know you said only one displayed, but I'm saying it in a way to make it easier for me to understand 😅 )

Basically, I would make it as simple as it can be. Giving a lot of flexibility creates too much of confusion for the end-user IMO. Having one column which doesn't have filter displayed, but the others have and I can still "global filter" it is very confusing and/or quite unexpected

melloware:
Well either way 1 displayed and all columns in global filter or all displayed except 1 but all used in the global filter.  But yeah you get the point.

Rapster:
Also why hiding filter using CSS, and not simply skip rendering on server side?

melloware:
You are right I should just not render the filter. I will update.

Rapster:
Cool, and again to make my thought clear. I'm more in favor for having `globalFilterOnly` on dataTable. Users will have these choices:
- Global filter with column filters _(e.g globalFilterOnly false, today's use case)_
- Global filter without column filters _(e.g globalFilterOnly true, what @stefan-dan wants)_

And by column filters, I mean columns having filterBy/field + filterable = true. So it's configurable at dataTable level not column level.

Thought on the side, what would be really cool, it's some kind of text highlighting for text matching using globalFilter 😄

melloware:
So you are saying you don't want it controllable by column level that its ALL or NOTHING at the datatable level?

Rapster:
Yes, there are at least 2 reasons to that:
- like I said, avoid confusion by having one column filter displayed, one column not displayed but still global filterable. Why would you allow it in one case and not the other? How end-user is supposed to know that one colum remains filterable using globalFilter etc.
- let's say I want to migrate all my dataTable to this new behavior _(which I must say is very tempting)_ I just need to set this attribute, instead of setting x _(nb columns)_ times

melloware:
Understood.  Let me modify this PR to that.  I can live with that.

Rapster (approved):

melloware:
New and old Integration Tests also passing with this fix.

Rapster:
Looks good! If you can, update filter showcase with a switch component allowing to change globalFilterOnly attr?

melloware:
Showcase button added to toggle the GlobalFilterOnly.

tandraschko:
TBH i liked the old flag (filterDisplayed) much more

will think about it

Rapster:
I just shared my thoughts on this as requested, but I'm all ears if you have a relevant use case. Maybe have both solution to make everyone happy? But i'm quite eager to use `globalFilterOnly` on our datatables

melloware:
Up to you guys. I can add both?

Rapster:
Yeah that should be possible

PR #7429, opened 2021-06-15
Fix #7419 Part 1 - use org.json instead of shaded

### LINKED ISSUE #7419: Development: multi-module + shade + IDE - issues
Not a issue for PrimeFaces-users, but a issue for people working on this OpenSource-project.

IntelliJ does not understand shaded classes in multi-module-projects: https://stackoverflow.com/questions/30407065/getting-intellij-to-import-shaded-dependencies-in-a-multi-module-maven-project/30898810

Causes e.g. issues for showcase-module:
image

And i can´t start showcase on Tomcat via IntelliJ-run-configuration any more. (so i´m able do debug showcase/PrimeFaces) I have to fallback to the jetty-plugin.

I don´t expect a short-term-solution for this. Just want to create some awareness of this issue.

christophs78:
... to avoid IDE-issues. (please keep #7419 opened; there seems to be more work to be done)

### DIFF
--- a/primefaces-showcase/pom.xml
+++ b/primefaces-showcase/pom.xml
@@ -187,6 +187,12 @@
             <version>${owasp-sanitizer.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.json</groupId>
+            <artifactId>json</artifactId>
+            <version>20210307</version>
+        </dependency>
+
         <!-- REST -->
         <dependency>
             <groupId>org.jboss.resteasy</groupId>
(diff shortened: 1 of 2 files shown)


melloware (approved):

melloware:
This is a good change.

tandraschko:
Tbh i dont like that 😂

melloware:
Its just for the Showcase so it untangles them.

melloware:
org.shaded classes should be considered almost private to PF and not be used by an external webapp.

PR #7468, opened 2021-06-26
update CI-badge und add badge for nightly integrationtests

christophs78:
The first modification is necessary because we renamed our job within build.yaml as part of the multi-module-thing.
Second badge is to show how our integration-tests run. (nightly.yml)
Maybe we should rename our jobs one more time to get good labels for our badges.


melloware:
How how just CI and IT for badge names to keep it short?

PR #7470, opened 2021-06-26
Fix #7469 - Datatable: support of custom dynamic filters

### LINKED ISSUE #7469: Datatable: support of custom dynamic filters
**Description**
As of today, working with filters and `<p:columns />` imply all columns share same filter component, which obviously is not ideal... `Columns#filterOptions` looks like an attempt to solve that problem but works only if a select component is what you want. I think there is a simple way to solve that "problem". Basically, facet filter expect one `ValueHolder` component, instead it should retrieve the first eligible/rendered component: it's up to the user to define the eligibility

```xml
<p:columns var="$col" value="#{controller.columnsModel}" field="#{$col.field}">
    <f:facet name="filter">
        <p:datePicker selectionMode="range" onchange="PF('#{widgetVar}').filter()" rendered="#{$col.type eq 'date'}"/>
        <p:selectOneMenu onchange="PF('#{widgetVar}').filter()" converter="omnifaces.SelectItemsConverter" rendered="#{$col.type eq 'enum'}">
            <f:selectItem itemValue="#{null}" itemLabel=""/>
            <f:selectItems value="#{$col.enumClass.enumConstants}" var="$o" itemLabel="#{i18n[$o]}"/>
        </p:selectOneMenu>
    </f:facet>
</p:columns>
```

If you also have faced that limitation, don't hesitate to share your thoughts maybe you came up with a better idea 😉 Any suggestions are welcome!

It also raises other questions like dynamic content, reusable filters for static datatable (defined in composites, facelet tags) etc

Rapster:


tandraschko:
looks good to me

PR #7490, opened 2021-06-29
Fix #7448: added milliseconds to datepicker

### LINKED ISSUE #7448: DatePicker: Add support for fractions of a second
To allow migrating some use cases from Calendar component, which is deprecated now, support for entering/selecting milliseconds should be added to DatePicker. It's a functionality that is present in Calendar, but not in DatePicker. While doing this, you may consider adding support for micro- or nanoseconds (any precision from 0 to 9).

wal-jan:
Added milliseconds to DatePicker (but not micros nor nonos since JS Date only supports milliseconds). Also added some integration tests and augmented DatePicker showcases.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.js
@@ -1344,6 +1344,7 @@
             hourText: 'Hour',
             minuteText: 'Minute',
             secondText: 'Second',
+            millisecondText: 'Millisecond',
             currentText: 'Current Date',
             ampm: false,
             year: 'Year',
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datepicker/DatePicker004.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datepicker/DatePicker004.java
@@ -39,12 +39,15 @@ public class DatePicker004 implements Serializable {
 
     private static final long serialVersionUID = 1L;
 
+    private LocalDateTime localDateTimeMilliseconds;
+
     private LocalDateTime localDateTimeSeconds;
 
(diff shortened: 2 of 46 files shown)


melloware:
This looks like it was a lot of work!   Nice PR we will all review it.

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.js
```diff
@@ -1344,6 +1344,7 @@
             hourText: 'Hour',
             minuteText: 'Minute',
             secondText: 'Second',
+            millisecondText: 'Millisecond',
```
jepsar, line 1347: We should add this to the locale files as well. @melloware as a separate PR?

melloware: I am fine with separate PR or if @wal-jan wants to add them to the local files in this PR.

wal-jan: @jepsar , @melloware: I'll add localized texts to PR (Google translate hopefully will do the translations)

melloware: Nice yeah at least that will be a good start and people can always complain and we can fix if its not perfect. Much appreciated.

wal-jan:
Done adding localized texts (see [068b70c](https://github.com/primefaces/primefaces/pull/7490/commits/068b70c71b7d2bdbd15a2f17d198b36bd0b1fcfe)) Do I have to issue a new PR or ist this change sufficient?

melloware:
Nope this PR is good. Thanks for updating those locales!

melloware (approved):

tandraschko:
great work and PR - thanks!

PR #7517, opened 2021-07-02
Fix #7516 - Datatable: introduce encodeDynamicCell method

### LINKED ISSUE #7516: Datatable: introduce encodeDynamicCell method
**Description**
Introduce this new method so it can be possible to override dynamic column rendering

Rapster:


tandraschko:
Why do you need this?

Please add a javadoc
Otherwise it may easy be removed by refactoring in the future

Rapster:
I know what you mean but it's not a method like `DataTable#encodeFrozenScrollableTable` (introduced for a premium themes AFAIR) which is never called, this one is and can't be removed. I could add javadoc, but name of the method is self-explanatory...

tandraschko:
I personally remove/refactor 1 liner methods which are only called once

Issue #2672, opened 2017-08-29
DataTable: Filtering and sorting does not work on ColumnGroup


kodmarcel:
The showcase for grouping [https://www.primefaces.org/showcase/ui/data/datatable/group.xhtml](https://www.primefaces.org/showcase/ui/data/datatable/group.xhtml) is used and modified to add sorting and filtering. The header columns show sorting arrows and input area for filtering but neither works on dynamic columns.

## 1) Environment
- PrimeFaces version: 6.0 (6.1, 6.2-SNAPSHOT(29.8.2017) are not working either)
- Does it work on the newest released PrimeFaces version? Version? NO
- Does it work on the newest sources in GitHub? (Build by source -> https://github.com/primefaces/primefaces/wiki/Building-From-Source) No
- Application server + version: Tested with : Wildfly-8.2.1 , Jetty-8.1.9.v20130131
- Affected browsers: Tested on Firefox 55, but should not work on others either.

## 2) Expected behavior

Sorting and filtering is working for the dynamic headers.

## 3) Actual behavior

Sorting and filtering are working for static header column. On dynamic header columns filtering is not invoked and sorting causes a null pointer exception.  

>SEVERE: java.lang.NullPointerException
	at org.primefaces.component.datatable.feature.SortFeature.decode(SortFeature.java:87)
	at org.primefaces.component.datatable.DataTableRenderer.decode(DataTableRenderer.java:64)
	at javax.faces.component.UIComponentBase.decode(UIComponentBase.java:831)
	at org.primefaces.component.api.UIData.processDecodes(UIData.java:287)
	at com.sun.faces.context.PartialViewContextImpl$PhaseAwareVisitCallback.visit(PartialViewContextImpl.java:573)
	at com.sun.faces.component.visit.PartialVisitContext.invokeVisitCallback(PartialVisitContext.java:183)
	at org.primefaces.component.api.UIData.visitTree(UIData.java:827)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1700)
	at javax.faces.component.UIForm.visitTree(UIForm.java:371)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1700)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1700)
	at com.sun.faces.context.PartialViewContextImpl.processComponents(PartialViewContextImpl.java:403)
	at com.sun.faces.context.PartialViewContextImpl.processPartial(PartialViewContextImpl.java:266)
[...]

PR #7525, opened 2021-07-02
Implement #7523: TreeTable: provide integrationtests

### LINKED ISSUE #7523: TreeTable: provide integrationtests
Please assign this one to me, i´ll try to work on this.

christophs78:
also adds generics to TreeNode

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/treetable/Document.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/treetable/Document.java
@@ -21,52 +21,27 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.primefaces.showcase.view.data.treetable;
+package org.primefaces.integrationtests.treetable;
 
-import org.primefaces.model.TreeNode;
-import org.primefaces.showcase.domain.Document;
-import org.primefaces.showcase.service.DocumentService;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import lombok.ToString;
 
-import javax.annotation.PostConstruct;
-import javax.faces.view.ViewScoped;
-import javax.inject.Inject;
-import javax.inject.Named;
 import java.io.Serializable;
 
-@Named("ttScrollableView")
-@ViewScoped
-public class ScrollableView implements Serializable {
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@ToString(onlyExplicitlyIncluded = true)
+public class Document implements Serializable, Comparable<Document> {
 
-    private TreeNode<Document> root1;
-    private TreeNode<Document> root2;
-    private TreeNode<Document> root3;
+    @ToString.Include
+    private String name;
+    private String size;
+    private String type;
 
-    @Inject
-    private DocumentService service;
-
-    @PostConstruct
-    public void init() {
-        root1 = service.createDocuments();
-        root2 = service.createDocuments();
-        root3 = service.createDocuments();
-
-        root1.getChildren().get(0).setExpanded(true);
-        root1.getChildren().get(1).setExpanded(true);
-    }
-
-    public TreeNode<Document> getRoot1() {
-        return root1;
-    }
-
-    public TreeNode<Document> getRoot2() {
-        return root2;
-    }
-
-    public TreeNode<Document> getRoot3() {
-        return root3;
-    }
-
-    public void setService(DocumentService service) {
-        this.service = service;
+    public int compareTo(Document document) {
+        return this.getName().compareTo(document.getName());
     }
 }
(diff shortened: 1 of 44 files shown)


primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/treetable/Document.java
```diff
@@ -0,0 +1,122 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.integrationtests.treetable;
+
+import java.io.Serializable;
+
+public class Document implements Serializable, Comparable<Document> {
```
melloware, line 37: Can just use @Data here and remove all this boilerplate!  @ToString(onlyExplicitylyIncluded=true) and then just put @ToString.Include above Name

primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/treetable/TreeTable001.java
```diff
@@ -0,0 +1,84 @@
...
+        root = service.createDocuments();
+
+        sortBy = new ArrayList<>();
+        sortBy.add(SortMeta.builder()
+                .field("name")
+                .order(SortOrder.ASCENDING)
+                .build());
+    }
+
+    public TreeNode getRoot() {
```
melloware: Don't need any of these setters or getters since you are using @Data.

primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/treetable/AbstractTreeTableTest.java
```diff
@@ -0,0 +1,108 @@
...
+    protected void assertRows(TreeTable treeTable, TreeNode root) {
+        List<Row> rows = treeTable.getRows();
+        assertRows(rows, root);
+    }
+
+    protected void assertRows(List<Row> rows, TreeNode<Document> root) {
+        int expectedSize = countRowsRecursive(root);
+        Assertions.assertNotNull(rows);
+        if (expectedSize == 0) {
+            Assertions.assertEquals(1, rows.size());; // No records found.
```
melloware: Double semicolon.

christophs78:
@tandraschko , @Rapster , @melloware - What´s the expected behaviour for multisort within treetable?
I´m currently a little bit surprised what showcase does. Are only leaves sorted or????
(single sort does what i expect)
image

melloware:
I would expect it to "sub-sort" the column within the main sort . So if I sort "Size" first it groups the 150KB together nd then if you make Name the second sort it flips flips the values sorted in the Name column.  it looks like it is doing it correctly to me.

christophs78:
Please take a second look.
IMO TreeTable does not take sortPriority into account. The later column always wins. So size first (column 2) and name second (column 1) works.
But not name first (column 1) and size second (column 2).
Look at https://github.com/primefaces/primefaces/blob/1f60dda906c252c33865fe38b7614662b8ef1308/primefaces/src/main/java/org/primefaces/component/treetable/TreeTableRenderer.java#L1187 and https://github.com/primefaces/primefaces/blob/2e79cbc628fa98885cad4688813fc7e82f6f3a97/primefaces/src/main/java/org/primefaces/model/TreeNodeComparator.java#L54.
IMO there´s currently a mismatch between what UI and showcase suggests and what happens in reality.

melloware:
Ahhh OK.   I think whatever the Showcase is doing for 10.0.0 is correct so something must have changed?

tandraschko:
maybe related to this commits: https://github.com/primefaces/primefaces/issues/2672
we partly also reworked TreeTable but it was not as good tested as DataTable

christophs78:
> maybe related to this commits: #2672
> we partly also reworked TreeTable but it was not as good tested as DataTable

Yes. It looks to be related to https://github.com/primefaces/primefaces/issues/2672. I'll open a separate issue for this.

tandraschko:
Perfect, let me know when we can merge this

christophs78:
From my side i´m done for V1.0. :-)
As always there´s way more to do.
There´s currently one disabled assertRows in TreeTable002Test#testMultiSortPlusPredefined we need to activate after https://github.com/primefaces/primefaces/issues/7537 is fixed.

tandraschko:
Perfect thanks

christophs78:
I need to check our TreeTable-integration-tests against FF in the evening....
```
Error:  Errors: 
Error:    TreeTable001Test.testSelectedEvent:301 » ElementNotInteractable Element <tr id...
Error:    TreeTable001Test.testShowSelectedNode:284 » ElementNotInteractable Element <tr...
Error:    TreeTable001Test.testUnselectedEvent:317 » ElementNotInteractable Element <tr ...
```

PR #7526, opened 2021-07-02
CI - switch back to Chrome

christophs78:
... because it´s way faster than Firefox

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,4 +35,4 @@ jobs:
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
       - name: Integration Tests
         if: matrix.java == 11
-        run: mvn -B -V clean install -fprimefaces-integration-tests/pom.xml -Pintegration-tests,headless,firefox,theme-saga,myfaces-2.3.x
+        run: mvn -B -V clean install -fprimefaces-integration-tests/pom.xml -Pintegration-tests,headless,chrome,theme-saga,myfaces-2.3.x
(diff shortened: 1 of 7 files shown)


tandraschko:
shoudlnt dialog.getYesButton() guard itself ajax?

tandraschko:
see CommandButton#click

christophs78:
> shoudlnt dialog.getYesButton() guard itself ajax?

i thougth so too until we had
```
Error:  Errors: 
Error:    ConfirmDialog001Test.testConfirmYes:119 » StaleElementReference stale element ...
```

When i debug this following code-path is selected:
image

tandraschko:
Doesnt it hit the guardAjax line?

christophs78:
Yep. He does not walk into guardAjax.
image

christophs78:
Maybe somehow related to the way `p:confirm` works.

tandraschko:
Maybe, yes
Thats a missing piece in pfselenium then, we must not workaround this in the tests

christophs78:
You are right. What do you about my last commit which tries to fix this for PF Selenium?

melloware (approved):

tandraschko:
Ist a confirm always Ajax?

christophs78:
> Ist a confirm always Ajax?

Nope.

Figuring out whether it is ajax or not is at least not easy. The button marked with red arrows needs to know how the button marked with green arrows is configured.
image

The easier thing is to handle this on the integration-test-side, what we had before.

tandraschko:
Yep i see

tandraschko:
would you like to switch back to guard in our tests?

PR #7544, opened 2021-07-06
Fix #7528: CommandLink remove duplicate onclick passthrough

### LINKED ISSUE #7528: CommandLink: doesn't fire action if "onclick" is defined with CSP enabled
**Describe the defect**
Currently we are migrating our project from Primefaces 8.0 to Primefaces 10.0.0. In Primefaces 8.0 everything worked fine. CSP is enabled since we migrated to PF 8.0.
We are using p:commandLink with an onclick event to check, whether there are changes in the form.. But since primefaces 10 the action of a commandLink isn't executed, if there is an onclick event defined. The onclick event of the commandLink can be filled or an empty String. As long as the attribute is defined the action doesn't get fired (But the onclick event does!).  Also everything works fine if we disable CSP. 

**Reproducer**

[commandButtonCSPExample.zip](https://github.com/primefaces/primefaces/files/6763133/commandButtonCSPExample.zip)

**Environment:**
 - PF Version: _10.0.0_
 - JSF + version: Mojarra 2.3.14
 - Affected browsers: ALL
 - CSP in web.xml is enabled

**To Reproduce**
1. Run the example as "mvn clean jetty:run -Pmojarra23"
2. In the Example click the first commandLink "works". It should print the current timestamp above.
3. Click the second commandLink "broken". It should first show an alert then update the timestamp without changing its value.
4. Disable CSP and restart the example. All buttons are executed correctly.

**Expected behavior**
The onclick event should be executed, if the commandlink is activated. After that if no return value is provided, the action should be fired.

In the example, I exchanged the mechanism to check, if there are changes with a simple alert to show that the onclick event is executed.

**Example XHTML**
```html
    <h:form>
	<h:outputText value="#{helloworld.getLastUpdate()}" id="textOutput" /> <br />
        <p:commandLink action="#{helloworld.doUpdate()}" update="textOutput">works </p:commandLink> <br />
        <p:commandLink onclick="alert('test');" action="#{helloworld.doUpdate()}" update="textOutput">broken</p:commandLink>
        <p:commandLink onclick="" action="#{helloworld.doUpdate()}" update="textOutput">also broken</p:commandLink>
    </h:form>
```

**Example Bean**
```java

@Named("helloworld")
@ViewScoped
public class HelloWorldBean implements Serializable {

	private String lastUpdate = "";

	public String getLastUpdate() {
		return lastUpdate;
	}

	public void doUpdate() {
		lastUpdate = LocalDateTime.now().toString();
	}
}
```

melloware:


primefaces/src/main/java/org/primefaces/component/commandlink/CommandLinkRenderer.java
```diff
@@ -112,13 +112,18 @@ public void encodeEnd(FacesContext context, UIComponent component) throws IOExce
             onclick.append(request);
 
             if (onclick.length() > 0) {
+                String onClickStr = onclick.toString();
                 if (link.requiresConfirmation()) {
-                    writer.writeAttribute("data-pfconfirmcommand", onclick.toString(), null);
+                    writer.writeAttribute("data-pfconfirmcommand", onClickStr, null);
```
tandraschko: its a if/else, i liked to old way better without extra var

melloware:
I put it back the way it was.  Sorry I was using it to debug what was in the onclick.

PR #7545, opened 2021-07-06
Fix #7531: PanelMenu store for the site

### LINKED ISSUE #7531: PanelMenu with global state
**Describe the defect**
With #6054 the state saving was changed from Cookies to LocalStorage. As discussed in https://github.com/primefaces/primefaces/issues/6054#issuecomment-874127828 ff. the new implementation behaves different in that the state is now saved on a per page basis vs. previously globally.

**Environment:**
 - PF Version: _10.0_
 - JSF + version: _MyFaces 2.3.9_
 - Affected browsers: ALL

**To Reproduce**
1. Create app with at least 2 pages and a shared main menu
2. The main menu should have multiple subsections
3. Open some of them on the first page
4. Navigate to 2nd page
5. The subsections are closed again.

**Expected behavior**
There should be an attribute that allows to configure global state saving for at least PanelMenus, possibly other components as well. This should allow to display the menu in the same state on both pages.

melloware:
Just a draft of what I was thinking to just restore the PanelMenu to previous behavior.

### DIFF
--- a/docs/11_0_0/components/carousel.md
+++ b/docs/11_0_0/components/carousel.md
@@ -45,6 +45,7 @@ Carousel is a multi purpose component to display a set of data or general conten
 | breakpoint | 560 | Integer | Breakpoint value in pixels to switch between small and large viewport.
 | touchable | true | Boolean | Enable touch support if browser detection supports it.
 | stateful | false | Boolean | When enabled, carousel state is saved in a LocalStorage.
+| statefulGlobal | false | Boolean | When enabled, carousel state is saved globally across pages. If disabled then state is stored per view/page.
 
 ## Getting Started with Carousel
 Carousel has two main use-cases; data and general content display. To begin with data iteration let’s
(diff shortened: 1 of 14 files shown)


tandraschko:
i think we should make it configurable
all components maybe be located in a template/global place

melloware:
so that means either new attribute in each component?  I am OK with that but just checking.  And what should we call it?

tandraschko:
Good question TBH

tandraschko:
maybe we just should add a doc hint, to overwrite the createStorageKey JS function?
or make the storageKey configurable via web.xml

im really not sure
it makes sense IMO to make it configurable per component but its not worth to add a attr to each of these components?

melloware:
I wonder if each component should have a`createStorageKey` method that can be overridden so for PanelMenu you could make it gloabl but for regular menu's leave them per page etc.  Then I can just update the docs showing how to override the value in each component?

melloware:
How does that look?

tandraschko:
im really really not sure
i think the best way is to add a attribute to each of those components, otherwise we may change it again in the future

either we do something like:
1) statefulGlobal=true|false
2) statefulStorageKey=....

melloware:
Added `statefulGlobal` property to the components and gave them smart defaults....
PanelMenu = true
Menu = false
Carousel = false

Which is their defaults in PF 8.0

tandraschko:
+1
i would just make all "false" per default

melloware:
OK all are now `false` by default.

PR #7555, opened 2021-07-07
DataTable: Integrationtest for filter + paging

christophs78:


primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datatable/DataTable001Test.java
```diff
@@ -163,6 +163,38 @@ public void testFilter(Page page) {
...
     @Test
     @Order(4)
+    @DisplayName("DataTable: filter plus paging")
+    public void testFilterPlusPaging(Page page) {
+        // Arrange
+        DataTable dataTable = page.dataTable;
+        dataTable.selectPage(1);
+        dataTable.sort("Name");
+        Select selectRowsPerPage = new Select(dataTable.getPaginatorWebElement().findElement(By.className("ui-paginator-rpp-options")));
+        // PrimeSelenium.guardAjax(selectRowsPerPage).selectByValue("2"); // does not work - maybe because of native Select
```
christophs78: any idea how we can get this working with guardAjax?

tandraschko: no idea sorry
i would think it should work like this but never used Select

christophs78: This one is because ByteBuddy can´t subclass `org.openqa.selenium.support.ui.Select`.

```
java.lang.RuntimeException: java.lang.NoSuchMethodException: org.openqa.selenium.support.ui.Select$ByteBuddy$fcqwp9fJ.<init>()

	at org.primefaces.selenium.internal.Guard.proxy(Guard.java:195)
	at org.primefaces.selenium.internal.Guard.ajax(Guard.java:102)
	at org.primefaces.selenium.internal.Guard.ajax(Guard.java:96)
	at org.primefaces.selenium.PrimeSelenium.guardAjax(PrimeSelenium.java:295)
	at org.primefaces.integrationtests.datatable.DataTable001Test.testFilterPlusPaging(DataTable001Test.java:173)
...
Caused by: java.lang.NoSuchMethodException: org.openqa.selenium.support.ui.Select$ByteBuddy$fcqwp9fJ.<init>()
	at java.lang.Class.getConstructor0(Class.java:3082)
	at java.lang.Class.getDeclaredConstructor(Class.java:2178)
	at org.primefaces.selenium.internal.Guard.proxy(Guard.java:192)
	... 62 more
```

tandraschko: because it doesnt have a default constructor?

christophs78: Yep. There must be some way to do this via ByteBuddy.

tandraschko: yeah
i will try something

christophs78: I added some magic to Guard#proxy. All test´s are green with this modification.
Maybe you find a better way to do this. :-)

tandraschko: i would try it like this: https://gist.github.com/tandraschko/b721d9ee1b8cde084b30ecb9a79afcf7

primefaces-selenium/primefaces-selenium-core/src/main/java/org/primefaces/selenium/internal/Guard.java
```diff
@@ -179,17 +179,25 @@ private static void waitUntilAjaxCompletes(WebDriver driver) {
...
+                .make()
+                .load(target.getClass().getClassLoader())
+                .getLoaded();
 
         try {
-            return proxyClass.getDeclaredConstructor().newInstance();
+            if (classProxyable && !classToProxy.getName().contains("$ByteBuddy$") &&
+                    target instanceof WrapsElement && !(target instanceof AbstractPrimePageFragment))  {
+                // e.g. org.openqa.selenium.support.ui.Select ("native" Selenium-components?)
+                return proxyClass.getDeclaredConstructor(WebElement.class).newInstance(((WrapsElement) target).getWrappedElement());
```
christophs78: For org.openqa.selenium.support.ui.Select (and maybe some others) we need to call the constructor with WebElement-argument.
see https://github.com/SeleniumHQ/selenium/blob/9b6231a45dd4ea3bd175a3a27709bcdd107ae7e9/java/client/src/org/openqa/selenium/support/ui/Select.java#L45

primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datatable/DataTable001Test.java
christophs78: good idea

tandraschko: looks more generic

christophs78: As expected - all test´s green with your more generic approach too.
Quite some fun to run my 8 core / 16 thread - notebook on 100% when the whole integration-test-suite runs. :-)
Would be even faster on a 16 core / 32 thread - workstation. ;-)

tandraschko: i even only have 4 / 8 :P

PR #7556, opened 2021-07-08
TreeTable: integrationtests - timing-issues on globalFilter

christophs78:


primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/treetable/AbstractTreeTableTest.java
```diff
@@ -113,4 +116,18 @@ protected void assertMessage(Messages messages, int index, String summary, Strin
...
+        /*
+         * TreeTable-filter ist triggered by keyup. (filterEvent is not configureable like for DataTable)
+         */
+
+        // maybe we can move some of this to PF Selenium (InputText?, TreeTable?)
+        inputGlobalFilter.clear();
+//        PrimeSelenium.guardAjax(inputGlobalFilter).sendKeys(filter); // does not work - causes org.openqa.selenium.TimeoutException: Timeout while waiting for AJAX complete!
+        inputGlobalFilter.sendKeys(filter);
+        PrimeSelenium.guardAjax(inputGlobalFilter).sendKeys(Keys.TAB);
+        PrimeSelenium.wait(500); // ugly work-around because PrimeSelenium.guardAjax does not work
```
christophs78: Better ideas/solutions are welcome.

tandraschko: could you add some logging to see what happens?
add start and end js logging in the guard and also catch PF ajax events with jquery and log it?

so we can see if the guard is around the ajax request or happens before

tandraschko: same for the other open PR

christophs78: Look´s like treetable-global-filter fires 2 ajax-request (when controlled via Selenium?)
image

The guard works right (for the first ajax-request)
image

But probably the second ajax-request causes timing-trouble.

christophs78: That´s our "bad guy":
image

christophs78:
Fixed this on the integration-test-side of things.
Wenn we think about PrimeSelenium.ajax it seem´s to be hard (impossible?) to catch usecases where one user-interaction causes more than one AJAX-request.

tandraschko:
i think it depends
when its queued, its handled by the PrimeFaces.ajax.Queue.isEmpty() check
if 1 ends and 2 starts short time after without queue, its quite hard

PR #7564, opened 2021-07-09
Fix #7563 corrected month names in DatePicker

### LINKED ISSUE #7563: Datepicker: numberOfMonths > 12 shows wrong month names + year
**Describe the defect**
An (inline) p:datePicker with numberOfMonths  > 12 shows wrong month names and wrong years after the second october

**Environment:**
 - PF Version: _10.0_
 - JSF + version: does not matter

**Expected behavior**
Correct months names and years

**Example XHTML**
```html
<h:form id="frmTest">
<p:datePicker id="calendars" numberOfMonths="15" inline="true"></p:datePicker>
</h:form>
```
See last 3 months (should be Nov 2021, Dec 2021 and Jan 2022)
datepickerFehler

The error is in 
https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
Line 342ff
the calculation does modulo 11 (and we have 12 months!?!). This only works (by chance) for the first year.

Fix: 
Calculate with modulo 12 and also calculate the correct year
```
createMonths: function (month, year) {
        var months = [];
        for (var i = 0; i < this.options.numberOfMonths; i++) {
            var m = month + i,
                y = year;
            if (m > 11) {
                y = year + Math.floor(m / 12);
                m = m % 12;
            }
            months.push(this.createMonth(m, y, i));
        }
        return months;
    },
```

wal-jan:
Bug fix and test

### DIFF
--- a/primefaces-integration-tests/src/main/webapp/datepicker/datePicker009.xhtml
+++ b/primefaces-integration-tests/src/main/webapp/datepicker/datePicker009.xhtml
@@ -1,0 +1,21 @@
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml"
+      xmlns:h="http://xmlns.jcp.org/jsf/html"
+      xmlns:f="http://xmlns.jcp.org/jsf/core"
+      xmlns:p="http://primefaces.org/ui">
+
+<f:view contentType="text/html;charset=UTF-8" encoding="UTF-8">
+    <h:head>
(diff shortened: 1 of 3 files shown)


melloware (approved):

melloware:
Fantastic and thanks for including the Integration Test!

PR #7567, opened 2021-07-10
Integrationtests: bring back Safari

christophs78:
Safari-IT as additional independent job; so it does not interfere with our IT-badge


melloware:
@christophs78 you think this one is ready?

christophs78:
> @christophs78 you think this one is ready?

Yep.

PR #7568, opened 2021-07-10
#7393 integrationtest

### LINKED ISSUE #7393: SelectOneMenu: client side function selectValue() doesn't work before menu was opened
**Describe the defect**
When calling the SelectOneMenu widget function [selectValue()](https://primefaces.github.io/primefaces/jsdocs/classes/src_primefaces.primefaces.widget.selectonemenu-1.html#selectvalue) before the menu was opened at least once, the function call fails.

````
Uncaught TypeError: Cannot read property 'eq' of undefined
    at c.selectValue (components.js.xhtml?…mefaces&v=10.0.0:20)
    at <anonymous>:1:45
selectValue	@	components.js.xhtml?…mefaces&v=10.0.0:20
(anonymous)	@	VM812:1
````
It seems like ``this.items`` in [forms.selectonemenu.js:1114](https://github.com/primefaces/primefaces/blob/10.0.0/src/main/resources/META-INF/resources/primefaces/forms/forms.selectonemenu.js#L1114) is undefined in this case.

After opening the menu, everything works fine.

**Reproducer**
Showcase for SelectOneMenu https://www.primefaces.org/showcase/ui/input/oneMenu.xhtml

**Environment:**
 - PF Version: 10.0.0
 - JSF + version: MyFaces 2.3.6
 - Affected browsers: ALL

**To Reproduce**
1. Open the SelectOneMenu showcase https://www.primefaces.org/showcase/ui/input/oneMenu.xhtml
2. Execute the ``selectValue()`` function in the console for one of the menus, e.g. 
````js
PrimeFaces.getWidgetById('j_idt302:option').selectValue('Option1')
````
3.  If you open the menu, e.g. ``PrimeFaces.getWidgetById('j_idt302:option').show()`` and then execute 2. again, it works.

**Expected behavior**
``selectValue()`` should be able to select a value even before the menu was opened.

### PR DESCRIPTION
**Pull request by christophs78**

### CODE REVIEW

**Review by melloware — APPROVED** [review]

File: primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/selectonemenu/SelectOneMenu001Test.java [hunk 1]
```diff
@@ -131,6 +132,24 @@ public void testEnablePanel(Page page) {
...
+    @Test
+    @Order(5)
+    @DisplayName("SelectOneMenu: selectValue via JavaScript")
+    public void testJsSelectValue(Page page) {
+        // Arrange
+        SelectOneMenu selectOneMenu = page.selectOneMenu;
+        Assertions.assertEquals("Lewis", selectOneMenu.getSelectedLabel());
+
+        // Act
+        PrimeSelenium.executeScript("PF('selectonemenu').selectValue('Max');");
```
melloware: We should probably add `selectValue` to to the SelectOne Selenium Component no?

melloware (changes requested):

primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/selectonemenu/SelectOneMenu001Test.java
```diff
@@ -131,6 +132,24 @@ public void testEnablePanel(Page page) {
...
+        selectOneMenu.select("Max");
```
melloware: Why do you also call `select` after calling `selectValue` ?

melloware (changes requested):

primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/selectonemenu/SelectOneMenu001Test.java
christophs78: Uuups. copying-issue.

primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/selectonemenu/SelectOneMenu001Test.java
christophs78: IMO the existing SelectOneMenu#select is fine. No need to expose the JS-function which does the same too.

christophs78: changed my mind ;-)

melloware (approved):

PR #7587, opened 2021-07-14
Fix #7586: Badge value empty

### LINKED ISSUE #7586: Badge: does not render with ui-badge-dot class on empty value

jepsar:
Fix #7586

Before:

Screenshot 2021-07-14 at 19 59 30

After:

Screenshot 2021-07-14 at 19 59 41

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/badge/BadgeRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/badge/BadgeRenderer.java
@@ -23,13 +23,12 @@
  */
 package org.primefaces.component.badge;
 
-import org.primefaces.renderkit.CoreRenderer;
-
+import java.io.IOException;
 import javax.faces.component.UIComponent;
 import javax.faces.context.FacesContext;
 import javax.faces.context.ResponseWriter;
-
-import java.io.IOException;
+import org.primefaces.renderkit.CoreRenderer;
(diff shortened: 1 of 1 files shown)


melloware (approved):

melloware:
Shouldn't the after the badge not be rendered at all? or at least not visible?

jepsar:
@melloware that was my initial idea, but then I ran into the line that added a class on empty values, so decided to fix that.

Maybe add a flag to hide if empty?

melloware:
That might be a good idea unless PrimeTek expected us to use the `rendered` flag?

jepsar:
The `rendered` attribute is useless for this, as you need to include children:
```xhtml
<p:badge value="1">
    <p:commandButton value="Fiets"/>
</p:badge>
```

melloware:
Got it.

melloware:
I am adding PrimeTek to this review to get their 2 cents...

jepsar:
If we would add a flag, I would suggest to make it a breaking change and hide the badge by default if it is empty, and show the dot if the flag is set.

tandraschko:
i think we can at least add this PR

melloware:
OK I am good with it if you want to merge it!

jepsar:
Opened a new issue for hiding the badge: https://github.com/primefaces/primefaces/issues/7676

PR #7589, opened 2021-07-15
Fix #7579: Tree: provide PF Selenium - wrapper and integrationtests

### LINKED ISSUE #7579: Tree: provide PF Selenium - wrapper and integrationtests
If someone else (@FlipWarthog?) wants to work on i´m happy. Else i'll try to care about this.

FlipWarthog:
Took a first pass at Tree. @melloware and I bounced some ideas back and forth regarding the best way to access the various elements within the p:tree. It's currently rough, but any feedback would be greatly appreciated, then I can incorporate them and build out some more tests. @christophs78


primefaces-selenium/primefaces-selenium-components/src/main/java/org/primefaces/selenium/component/model/tree/TreeNode.java
```diff
@@ -0,0 +1,73 @@
...
+import org.openqa.selenium.WebElement;
+
+public class TreeNode {
+
+    private WebElement webElement;
+    private String selector;
+    private String childSelector;
+
+    public TreeNode(WebElement webElement, String selector) {
+        this.webElement = webElement;
```
melloware, line 43: You might want to pass the owning Tree component into this as well.  You will probably need it.

tandraschko:
feels like RootNode should be removed (selector moved to Tree component) and reuse TreeNode instead

primefaces-selenium/primefaces-selenium-components/src/main/java/org/primefaces/selenium/component/model/tree/TreeNode.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public String getSelector() {
+        return selector;
+    }
+
+    public void setSelector(String selector) {
+        this.selector = selector;
+    }
+
```
melloware, line 80: What about a `toggle()` convenience method right here that calls `getTreeToggler.click()`.   The reason this will be good is for another test you will need to do this with AJAX "expand" and "collapse" behaviors and we will need to guard this `toggle`

primefaces-selenium/primefaces-selenium-components/src/main/java/org/primefaces/selenium/component/model/tree/RootNode.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebElement;
+
+public class RootNode {
+
+    public static final String CHILD_SELECTOR = ".ui-tree-container>li";
+    private WebElement webElement;
+
+    public RootNode(WebElement webElement) {
```
melloware: Pass the owning Tree component in here maybe?

melloware (changes requested):

christophs78:
Regarding (integration-)testing:
I personally try to add some AJAX-postbacks (via eg CommandButton).
With this postbacks we ensure correct re-rendinger of componentes (is still the same value / node selected?). And i also like to add some (info-)message(s) with the selected value / node from the ViewBean. And validate this messages as part of the integration-test.
Similar things for ajax-events.

melloware:
Yep we will definitely need tests that use the "expand" and "collapse" AJAX events with messages etc.

primefaces-selenium/primefaces-selenium-components/src/main/java/org/primefaces/selenium/component/model/tree/TreeNode.java
FlipWarthog: Added `toggle()` and `select()`, both with `guardAjax(...)`.

primefaces-selenium/primefaces-selenium-components/src/main/java/org/primefaces/selenium/component/model/tree/TreeNode.java
FlipWarthog: Added, as well as the parent `TreeNode`, if available.

FlipWarthog:
Comments addressed. Let me know if there's anything else. @melloware @tandraschko @christophs78

primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/tree/Tree002Test.java
```diff
@@ -0,0 +1,94 @@
...
+        first.select();
+
+        // Assert
+        assertMessage(page.messages, 0, "Selected", "Documents");
+        page.buttonShowSelectedNode.click();
+        assertMessage(page.messages, 0, "Selected node", "Documents");
+
+        // Act Pt. 2
+        Actions actions = new Actions(page.getWebDriver());
+        actions.keyDown(Keys.META).click(first.getLabel()).keyUp(Keys.META).perform();
```
christophs78: Maybe we should add some PrimeSelenium.guardAjax to avoid running into timing-issues.

christophs78:
Thanks for working on this!

The one thing i´m missing is something like
`<p:commandButton id="buttonUpdate" value="Update" update="@form"/>`
And after this some re-evaluation of the Tree.
Why? We should ensure state is kept through (AJAX-)update-cycles. (Sound´s trivial but is something new components sometimes fail because no one cares. And should not break after some refactoring or other changes.)

melloware (approved):

primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/tree/Tree002Test.java
FlipWarthog: I can wrap it around buttonShowSelectedNode, but if I try something like `actions.keyDown(Keys.META).click(PrimeSelenium.guardAjax(first.getLabel())).keyUp(Keys.META).perform();` Selenium seems to lock up when it performs the series of actions.

FlipWarthog:
@christophs78 Added a basic test for the AJAX update. There's still a ton of Tree functionality to test, specifically multi-select with both `List<TreeNode>` and `TreeNode[]` fields in the backing bean (which is what kicked off this exercise). I think I'll save those for a separate issue, if that's okay.

primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/tree/Tree001Test.java
```diff
@@ -78,13 +80,64 @@ public void testBasicAndToggling(Page page) {
...
+        Assertions.assertEquals("Documents", first.getLabelText());
+
+        TreeNode second = children.get(1);
+        Assertions.assertEquals("Events", second.getLabelText());
+
+        TreeNode third = children.get(2);
+        Assertions.assertEquals("Movies", third.getLabelText());
+
+        // Act
+        PrimeSelenium.guardAjax(page.buttonUpdate).click();
```
christophs78, line 115: Please expand (toggle) one node before clicking update-button. So we can check this node is still expanded after update.

FlipWarthog: Added.

melloware:
I just kicked off the GitHub Action tests to see if this runs and passes. If So I will merge. @christophs78 we can then modify and build other tests in other PR's if that is OK with you?

melloware:
Looks like we got 1 failure.  Most likely need an AJAX guard somewhere
```
 Errors: 
Error:    Tree002Test.testSingleSelection:73->AbstractTreeTest.assertMessage:41 » StaleElementReference
[INFO] 
Error:  Tests run: 360, Failures: 0, Errors: 1, Skipped: 1
```

christophs78:
> I just kicked off the GitHub Action tests to see if this runs and passes. If So I will merge. @christophs78 we can then modify and build other tests in other PR's if that is OK with you?

Yes it´s fine. We can always add more. :-)

FlipWarthog:
@melloware I added a commit to address the stale element reference, if you want to kick the build again.

PR #7596, opened 2021-07-15
SelectManyMenu: Integrationtest: filter

christophs78:


primefaces-selenium/primefaces-selenium-components/src/main/java/org/primefaces/selenium/component/SelectManyMenu.java
```diff
@@ -103,9 +106,17 @@ public boolean isSelected(String label) {
     }
 
     public List<String> getLabels() {
-        return getInput().findElements(By.tagName("option")).stream()
+        if (getWidgetConfiguration().has("filter") &&  getWidgetConfiguration().getBoolean("filter")) {
```
melloware: For efficiency maybe only get the widgetConfiguration once and store in a variable?

melloware (changes requested):

melloware (approved):

PR #7622, opened 2021-07-22
Fix #6575: selection could not be restored for lazy-mode + MVS

### LINKED ISSUE #6575: DataTable: clean up selection

fanste:
This fix adds two flags to restoreMultiViewState which are only relevant
for components that do not have a static data model but use lazy loaded
data. In that case the first flag indicates, that the data independent
state should be restored (e.g. filters in a datatable that is required
to load the data model). The second flag indicates, that the data is
available and the data dependent state could be restored.

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datatable/DataTable025.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datatable/DataTable025.java
@@ -1,0 +1,71 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2021 PrimeTek
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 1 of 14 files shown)


fanste:
As announced in https://github.com/primefaces/primefaces/issues/6575#issuecomment-883514213 I suggest these changes in order to fix a chicken-egg problem in the current MVS handling in DataTable.

tandraschko:
this requires deep review on your problem and the PR, probably i dont have much time for it the next weeks.

christophs78:
May we ask you for adding integrationtests for this?

We currently have

- DataTable015Test which does some MVS tests (sort, filter, pagination) with a non-lazy datasource
- DataTable004Test for non-lazy datasource with single-select
- DataTable005Test for non-lazy datasource with multi-select

(and a few more mutations with other selection-ui, ...)

IMO we should create additional integration-tests for DataTable + lazy datasource + single/multi-select + MVS.

And you may look at DataTable001Test how to have one integrationtest-class which runs the same test´s against multiple XHTML´s. (May help to have to same integrationtest-class to check lazy and non-lazy datasources.)

fanste:
Added a basic integration tests, that covers my changes to the data table. It will be executed for a lazy-loaded and a non-lazy-loaded data table.

tandraschko:
@fanste i migrated your tests, see #8275
but both, lazy and non-lazy fails
can you check them?

tandraschko:
tests are working fine now, propably because of other changes
will close this for now

please create a new issue and reproducer if you still have problems

PR #7626, opened 2021-07-23
Fix #7613: Tree: Focus / Navigation broken if tree is droppable

### LINKED ISSUE #7613: Tree: Focus / Navigation broken if tree is droppable
If droppable is active on a p:tree and you tab to it the whole tree container is focused. And in this case it is impossible to select a node via keyboard.

**Reproducer**
https://www.primefaces.org/showcase/ui/data/tree/basic.xhtml
https://www.primefaces.org/showcase/ui/data/tree/dragdrop.xhtml

**Environment:**
PF Showcase v10

**To Reproduce**
Tab to focus the tree on both pages.

**Expected behavior**
basic tree: first node is focused
dragdrop tree: whole tree container focused. impossible to focus a node via keyboard.

**Example XHTML**
---

**Example Bean**
---

FlipWarthog:
Drag/Drop mode on tree adds extra `li` elements that wreak havoc within the js keybinds. `getFirstNode`, `getNodeLabel`, and tree traversal with arrow keys were broken for both normal trees and Drag/Drop trees. This PR adds new integration tests that reveal the issue, as well as the fixes that make them green.

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/tree/Tree003.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/tree/Tree003.java
@@ -1,0 +1,54 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2021 PrimeTek
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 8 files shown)


melloware (approved):
Great job!  Really appreciate both the integration test and the quality fix!

Issue #7598, opened 2021-07-16
Responsive: PrimeFlex 3.0 Upgrade


melloware:
**Description**
PrimeFlex 3.0 was released and contains some breaking changes we would need to update in our components and our GridLayoutUtils.java

Migration Guide: https://www.primefaces.org/primeflex/migration

cc @christophs78 

TODOs:

- [x] Showcase
   - [x] all XHTMLs (use PrimeFaces CLI migration tool) 
   - [x] PrimeFlex section (documentation as part of XHTMLs)
- [x] Documentation
   - [x] Migration guide
   - [x] formlayout.md
   - [x] panelgrid.md
- [x] PrimeFaces (Core)
   - [x] PanelGrid
   - [x] GridLayoutUtils

christophs78:
Don´t know yet wether i like all this class-renaming or not.
At least this should allow developers to include PrimeFlex 2 and PrimeFlex 3 without walking through the whole application and replacing css-classes.
Maybe someone want´s to write a migration-tool which walks through all XHTMLs and does this replacement automatically.

melloware:
That is not a bad idea.  We were discussing something similar.

christophs78:
We may provide such a PrimeFlex-Migration-CLI-Tool as new PrimeFaces-module or as part of PF Extensions.
Better to be done once than each costumer on it´s own.

FlipWarthog:
@melloware Also this condition in PanelGridRenderer will need to be updated:
```java
if (!columnClass.contains("p-md-") && !columnClass.contains("p-col-")) {
   columnClass = columnClass + " " + GridLayoutUtils.getFlexColumnClass(columns);
}
```

christophs78:
Regarding PrimeFlex-Migration-CLI-Tool the first step seems to be creating a file (JSON/XML/CSV) which contains a mapping for all PrimeFlex 2.0 - classes to PrimeFlex 3.0 - classes. I´ll try to start a PR for this.

christophs78:
Do we still wan´t to do this for 11.0.0 or do we postpone this to 12.0.0?
When we wan´t to do this for 11.0.0 we would not to start early. Modify PrimeFaces Core, migrate Showcase incl. PrimeFlex - documentation.
From my personal view i would prefer a stable 11.0.0 release and do this afterwards.

melloware:
I am OK waiting until 12.0 on this one.

jepsar:
[...]

PR #7633, opened 2021-07-25
#7598: CLI-tool for PrimeFlex V2 --> V3 - migration

christophs78:
Very early days. :-)
Just to give you an idea of the direction i think.


tandraschko:
i like the idea because we can really resolve many migration problems

also we should think about using something like: https://github.com/remkop/picocli

christophs78:
Just converted my local PrimeFaces-showcase to PrimeFlex 3. Most things already look good. :-)
Someone who want´s to to have a look at this PR?

@mertsincan : This CLI-tool should propably work for PrimeNG, PrimeReact und PrimeVue too.

christophs78:
Also added Grid CSS to PrimeFlex 2 migration. (We still have this in use.)
So you may first migrate vom Grid CSS to PrimeFlex 2. And then later from PrimeFlex 2 to PrimeFlex 3.

melloware:
Nice!

tandraschko:
i would rework the readme a bit:

1) Intro: global stuff like (It´s suggested to have no uncommited... and here always may be corner-cases which need some additional manual work...)

2) PrimeFlex 2 --> PrimeFlex 3 migration

3) Grid CSS --> PrimeFlex 2 migration

tandraschko:
Otherwise great work!
we should probably add this to the docs and especially migration guide

FlipWarthog:
This looks amazing! Great work, @christophs78!

marcelocaser:
very nice ideia!! will be very util

melloware (approved):

tandraschko:
Thanks @christophs78, thats a nice contribution!

PR #7632, opened 2021-07-25
Fix #7617 fixed NPEs in FileUpload

wal-jan:
FileUpload now ignores empty parts in request (both in native and commons uploader).

NPE found in commons uploader actually was a configuration error. So I added a warning log output in that case.

Added some more FileUpload integration tests which required some tweeks in the projects configuration. Now it's possible to pick native or commons uploader via profile (default is auto). 

I think exception reported in #7616 is also fixed.

PR #7635, opened 2021-07-26
#7632 added nightly file upload integration tests

wal-jan:
Suggestion for nightly to run FileUpload integration tests in Chrome, Firefox with Mojarra, Myfaces, Commons, and Native mode.

### DIFF
--- a/.github/workflows/nightly-file.upload.yml
+++ b/.github/workflows/nightly-file.upload.yml
@@ -1,0 +1,42 @@
+name: IT FileUpload
+
+on: [push, pull_request]
+  #schedule:
+  #  - cron: '0 8 * * *'
+
+jobs:
+  nightly:
+    if: github.repository == 'primefaces/primefaces' && github.ref == 'refs/heads/master'
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


tandraschko:
@christophs78 should we do it like this and create a workflow only fileupload?

.github/workflows/nightly-file.upload.yml
```diff
@@ -0,0 +1,42 @@
+name: IT FileUpload
+
+on: [push, pull_request]
+  #schedule:
+  #  - cron: '0 8 * * *'
+
+jobs:
+  nightly:
+    if: github.repository == 'primefaces/primefaces' && github.ref == 'refs/heads/master'
```
christophs78: please disable this line temporary to check to workflow-run

christophs78:
> @christophs78 should we do it like this and create a workflow only fileupload?

I need to think about it. :-)
Maybe it´s worth the "price" compared to having an XXL-extensive matrix for checking all combinations as part of our primary IT-matrix.

christophs78:
I´m fine with this approach. IMO we don´t want to add this to our full IT-matrix. So it´s ok, to add another scheduled job to run file upload integration tests with this additional upload-specific-matrix.

PR #7637, opened 2021-07-26
Fix #7576 UICalendar.validateMinMax

### LINKED ISSUE #7576: Calendar: mindate and maxdate does not work
**Describe the defect**
As described here
https://forum.primefaces.org/viewtopic.php?f=3&t=67214
setting mindate/maxdate in p:calender get strange results. 

This one works:
`<p:calendar id="cal" mode="inline"
	value="#{data.selectedDate}"
	pattern="dd.MM.yyyy" locale="de"
	mindate="03.06.2021"
	maxdate="09.06.2021">
	<p:ajax event="dateSelect" listener="#{controller.dateSelectListener}" update="controlButtons"/>
</p:calendar>`

For this dates, I got a  error "Calendar : "cal" minimum date must be less than maximum date.": 
`<p:calendar id="cal" mode="inline"
	value="#{data.selectedDate}"
	pattern="dd.MM.yyyy" locale="de"
	mindate="27.05.2021"
	maxdate="10.06.2021">
	<p:ajax event="dateSelect" listener="#{controller.dateSelectListener}" update="controlButtons"/>
</p:calendar>`

**Reproducer**
Just open your showcase
https://www.primefaces.org/showcase/ui/input/calendar/calendar.xhtml
In your showcase mindate="4/1/14" and maxdate="7/1/14". It seems that p:calendar does not use these values.

**Environment:**
 - PF Version: 10.0.0
 - JSF + version: MyFaces 2.3.9
 - Affected browsers: all

**To Reproduce**
See "Reproducer" for details.

**Expected behavior**
mindate and maxdate should work as expected for ALL dates. I can't migrate to datepicker at the moment, which was suggested by another user.

**Example XHTML**
See "Reproducer" for details.

**Example Bean**
See "Reproducer" for details.

wal-jan:
`UICalendar#validateMinMax(FaceContext)` compared strings instead of formated dates.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/api/UICalendar.java
+++ b/primefaces/src/main/java/org/primefaces/component/api/UICalendar.java
@@ -23,6 +23,7 @@
  */
 package org.primefaces.component.api;
 
+import java.time.LocalDate;
 import java.time.chrono.IsoChronology;
 import java.time.format.DateTimeFormatterBuilder;
 import java.time.format.FormatStyle;
@@ -368,10 +369,9 @@ public abstract class UICalendar extends AbstractPrimeHtmlInputText implements I
         }
     }
 
-    @SuppressWarnings({"unchecked", "rawtypes"})
     public void validateMinMax(FacesContext context) {
-        Comparable minDate = (Comparable) getMindate();
-        Comparable maxDate = (Comparable) getMaxdate();
+        LocalDate minDate = CalendarUtils.getObjectAsLocalDate(context, this, getMindate());
+        LocalDate maxDate = CalendarUtils.getObjectAsLocalDate(context, this, getMaxdate());
         if (minDate != null && maxDate != null && maxDate.compareTo(minDate) < 0) {
             String id = getClientId(context);
             String component = this.getClass().getSimpleName();
--- a/primefaces/src/main/java/org/primefaces/component/api/UICalendar.java
+++ b/primefaces/src/main/java/org/primefaces/component/api/UICalendar.java
@@ -24,16 +24,11 @@
 package org.primefaces.component.api;
 
 import java.time.Instant;
-import java.time.LocalDate;
-import java.time.LocalDateTime;
-import java.time.LocalTime;
-import java.time.ZoneOffset;
 import java.time.chrono.IsoChronology;
 import java.time.format.DateTimeFormatterBuilder;
 import java.time.format.FormatStyle;
 import java.time.format.ResolverStyle;
 import java.util.Collection;
-import java.util.Date;
 import java.util.List;
 import java.util.Locale;
 
@@ -42,7 +37,6 @@ import javax.faces.FacesException;
 import javax.faces.application.FacesMessage;
 import javax.faces.context.FacesContext;
 
-import org.primefaces.component.datepicker.DatePicker;
 import org.primefaces.util.*;
 
 public abstract class UICalendar extends AbstractPrimeHtmlInputText implements InputHolder, TouchAware {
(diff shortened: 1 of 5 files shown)


primefaces/src/main/java/org/primefaces/component/api/UICalendar.java
```diff
@@ -368,10 +369,9 @@ protected void createFacesMessageFromValidationResult(FacesContext context, Vali
         }
     }
 
-    @SuppressWarnings({"unchecked", "rawtypes"})
     public void validateMinMax(FacesContext context) {
-        Comparable minDate = (Comparable) getMindate();
-        Comparable maxDate = (Comparable) getMaxdate();
+        LocalDate minDate = CalendarUtils.getObjectAsLocalDate(context, this, getMindate());
+        LocalDate maxDate = CalendarUtils.getObjectAsLocalDate(context, this, getMaxdate());
```
christophs78: IMO this is a bit more complicated. mindate/maxdate may be java.time.LocalDate, java.time.LocalDateTime, java.time.LocalTime, java.util.Date (deprecated) or String when we look into docu.
Currently it is broken for String. But we should not break this for LocalDateTime what this PR does. (mindate may be something like '27.05.2021 10:00' for a LocalDateTime-usecase)
Maybe we should explicit check for String and only do some explizit conversion for String.

melloware: Agreed.

melloware: These both should equate to an `Instant` somehow.

melloware (changes requested):

primefaces/src/main/java/org/primefaces/component/api/UICalendar.java
christophs78: … or we may convert everything to `LocalDateTime`.

wal-jan: I picked @melloware's suggestion to convert evrything to `Instant`
```
    private Instant toInstant(FacesContext context, Object date) {
        if (date == null) {
            return null;
        }
        if (date instanceof LocalDateTime) {
            return ((LocalDateTime) date).toInstant(ZoneOffset.UTC);
        }
        if (date instanceof LocalDate) {
            return ((LocalDate) date).atStartOfDay().toInstant(ZoneOffset.UTC);
        }
        if (date instanceof LocalTime) {
            LocalDate now = LocalDate.now();
            return now.atTime(((LocalTime) date)).toInstant(ZoneOffset.UTC);
        }
        if (date instanceof Date) {
            if (date instanceof java.sql.Date) {
                // java.sql.Date does not support toInstant
                return new Date(((java.sql.Date) date).getTime()).toInstant();
            }
            return ((Date) date).toInstant(); // implied UTC
        }
        if (date instanceof String) {
            return CalendarUtils.getObjectAsLocalDate(context, this, date).atStartOfDay().toInstant(ZoneOffset.UTC);
        }

        throw new FacesException(... " unsupported type ");
    }
```
@christophs78  I'm not sure if the `String` case is sufficient. Is it possible that the min/maxDate string may contain time parts?

wal-jan: @christophs78 Looks like it is possible to set time parts in min/maxDate string. At least `DatePicker.validateDateValue` takes it into account. So the `String` case has to be improved.

primefaces/src/main/java/org/primefaces/component/api/UICalendar.java
```diff
@@ -370,12 +376,53 @@ protected void createFacesMessageFromValidationResult(FacesContext context, Vali
...
+        Instant minDate = toInstant(context, getMindate(), "minimum");
+        Instant maxDate = toInstant(context, getMaxdate(), "maximum");
         if (minDate != null && maxDate != null && maxDate.compareTo(minDate) < 0) {
             String id = getClientId(context);
             String component = this.getClass().getSimpleName();
             throw new FacesException(component + " : \"" + id + "\" minimum date must be less than maximum date.");
         }
     }
+
+    private Instant toInstant(FacesContext context, Object date, String attr) {
```
melloware: I wonder if this method should be moved to CalendarUtils?

jepsar: Think so too

wal-jan: If `toInstant` has to be moved to `CalendarUtils` it should be renamed to somthing like `getObjectAsInstant`. The parameter `attr` (which is only used to construct a helpful FacesException) has to be reworked somewhat. And as in `CalendarUtils.getObjectAsLocalXxx` an additional parameter `UICalendar calendar` has to be added.
If the programming guidelines suggest that this code belongs in a util class, then I'll be happy to rebuild it.
What should the moved method be called?

melloware: i like getObjectAsInstant as it keeps the pattern the same as getObjectAsLocal...

melloware (approved):

PR #7644, opened 2021-07-28
Fixing global attribute default value

garcia-jj:
Reading the docs, the default value for `global` attribute is false. And reading the source code, actually the default value is true. Looks like a typo since in the default value column is also true.

For reference: https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/java/org/primefaces/component/fileupload/FileUploadBase.java#L450

### DIFF
--- a/docs/10_0_0/components/fileupload.md
+++ b/docs/10_0_0/components/fileupload.md
@@ -36,7 +36,7 @@ powered rich solution with graceful degradation for legacy browsers.
 | dragDropSupport | true | Boolean | Specifies dragdrop based file selection from filesystem, default is true and works only on supported browsers.
 | fileLimit | null | Integer | Maximum number of files allowed to upload.
 | fileLimitMessage | Maximum number of files exceeded | String | Message to display when file limit exceeds.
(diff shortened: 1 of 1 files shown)


docs/10_0_0/components/fileupload.md
```diff
@@ -36,7 +36,7 @@ powered rich solution with graceful degradation for legacy browsers.
 | dragDropSupport | true | Boolean | Specifies dragdrop based file selection from filesystem, default is true and works only on supported browsers.
 | fileLimit | null | Integer | Maximum number of files allowed to upload.
 | fileLimitMessage | Maximum number of files exceeded | String | Message to display when file limit exceeds.
-| global | true | Boolean | Global AJAX requests are listened by ajaxStatus component, setting global to false will not trigger ajaxStatus. Default is false.
+| global | true | Boolean | Global AJAX requests are listened by ajaxStatus component, setting global to false will not trigger ajaxStatus. Default is true.
```
jepsar, line 39: Don't we need this change in other versions as well?

PR #7656, opened 2021-08-01
Added Korean translations

N-Hidaya:
Added Korean translations to Message properties and locales


melloware:
I made some changes to your PR.  The official Java code for Korean is "ko" not "kr". I also added it to the Showcase so you can switch languages and tested it.  Nice Job. See screenshot below..

image

Issue #7620, opened 2021-07-22
Showcase: Update schema location of ehcache


fanste:
Trying to import this repository to eclipse to implement some fixes shows me an error regarding the ehcache example in the showcase. Root cause is, that they enforce HTTPS instead of HTTP and my eclipse 2021-03 installation is not happy about the 301 HTTP status.

After changing all `http://www.ehcache.org/...` to `https://www.ehcache.org/...`, everything is fine.

PR is on the way.

melloware:
thanks!

melloware:
I can't run locally now. I get this error...

```
Caused by: org.xml.sax.SAXParseException; systemId: 
file:/C:/dev/melloware-git/primefaces-korea/primefaces-showcase/target/classes/ehcache.xml; 
lineNumber: 5; columnNumber: 105; cvc-elt.1: Cannot find the declaration of element 'ehcache:config'.
        at com.sun.org.apache.xerces.internal.util.ErrorHandlerWrapper.createSAXParseException(ErrorHandlerWrapper.java:203)
```

If I revert the ehache.xml to what it was before I can run locally again.  Thoughts?

christophs78:
ehcache-core.xsd defines it´s targetNamespace with http://
![image](https://user-images.githubusercontent.com/10461942/127783745-c4bac2de-7ef9-41b2-ab70-15d9dca05a8a.png)

christophs78:
Switching to https makes IntelliJ 2021.2 unhappy
![image](https://user-images.githubusercontent.com/10461942/127783841-4916bda2-25fb-4702-b84a-9f13e1d146eb.png)
And it fails at runtime.

fanste:
Revert it and I will try to debug Eclipse to see, why it is unhappy with "HTTP" instead of "HTTPS" while other tools and IDEs accept it without any problem.

But this takes 2 weeks, as I will be on vacation.

melloware:
Reverted.

melloware:
any update on this or should we close it?  I think only the https://ehcache URL can be HTTPS and the W3C ones stay HTTP?

fanste:
The project business is currently taking up my time. This issue is on my list.

`xmlns:xsi` was changed by mistake, maybe thats already the fix for IntelliJ.

melloware:
Yep I am going to close for now we can re-open if you decide other changes are needed!

PR #7657, opened 2021-08-01
Showcase - fix ehcache.xml

christophs78:
~~probably necessary due to EHCache 3.8 - updates some time again~~


christophs78:
Hm. Maybe related to https://github.com/primefaces/primefaces/issues/7620.

christophs78:
Switching to https sounded good in theory but seems to be a bad idea in practice.

PR #7664, opened 2021-08-03
Fix #7648: Datatable/TreeTable ARIA sort labels

### LINKED ISSUE #7648: DataTable/TreeTable: In a sorted datatable the aria-label in Primefaces 10 always shows the text "undefined" after the column name.
In Primefaces 10 the `aria-label` for a sorted datatable is partly broken and always shows the text "undefined" after the column name.
In Primefaces 8 the `aria-label` works as expected.

 - PF Version: 8.0  works
 - PF Version: 10.0  don't work
 - PF Version: 11.0-SNAPSHOT  don't work
 - Affected browsers: ALL

**To Reproduce**
To see what is wrong go to the Primefaces showcase for DataTable Sort https://www.primefaces.org/showcase/ui/data/datatable/sort.xhtml
See in the HTML code the  `th` tags and the `aria-label` attribute.
All `aria-label` attribute always shows the text "undefined" after the column name.

**Expected behavior**
The `aria-label` should show the text like before in Primefaces 8

**That is wrong**.
In Primefaces 10 in the file src/main/resources/META-INF/resources/primefaces/datatable/datatable.js the aria label informations are get with this 3 lines of code
```
//aria messages
this.ascMessage = PrimeFaces.getAriaLabel('datatable.sort.SORT_ASC');
this.descMessage = PrimeFaces.getAriaLabel('datatable.sort.SORT_DESC');
this.otherMessage = PrimeFaces.getAriaLabel('datatable.sort.SORT_LABEL');
```
But the english default locale in the file src/main/resources/META-INF/resources/primefaces/core/core.js still has the old Primefaces 8 aria label keys for the datatable sort.
```
aria: {
    'paginator.PAGE': 'Page {0}',
    'calendar.BUTTON': 'Show Calendar',
    'datatable.sort.ASC': 'activate to sort column ascending',
    'datatable.sort.DESC': 'activate to sort column descending',
    'columntoggler.CLOSE': 'Close',
    'overlaypanel.CLOSE': 'Close'
 }
```

Btw.
The documentation for Primefaces 10 is also still on Primefaces 8 for these default datatable sort language keys.

So here is my fix for the file src/main/resources/META-INF/resources/primefaces/core/core.js:
```
aria: {
    'paginator.PAGE': 'Page {0}',
    'calendar.BUTTON': 'Show Calendar',
    'datatable.sort.SORT_ASC': 'activate to sort column ascending',
    'datatable.sort.SORT_DESC': 'activate to sort column descending',
    'datatable.sort.SORT_LABEL': 'activate to unsort column',
    'columntoggler.CLOSE': 'Close',
    'overlaypanel.CLOSE': 'Close'
 }
```

melloware:
@jepsar @christophs78 I am adding a new client side label if you guys want to send me Dutch and German translations I will add them to this PR.

### DIFF
--- a/docs/11_0_0/core/localization.md
+++ b/docs/11_0_0/core/localization.md
@@ -134,6 +134,7 @@ Side Validation, messages property is used as the bundle for the locale.
          'calendar.BUTTON': 'Show Calendar',
          'datatable.sort.ASC': 'activate to sort column ascending',
          'datatable.sort.DESC': 'activate to sort column descending',
+         'datatable.sort.NONE': 'activate to remove sorting on column',
          'columntoggler.CLOSE': 'Close',
          'overlaypanel.CLOSE': 'Close'
      },
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.js
@@ -1360,6 +1360,7 @@
                 'calendar.BUTTON': 'Show Calendar',
                 'datatable.sort.ASC': 'activate to sort column ascending',
                 'datatable.sort.DESC': 'activate to sort column descending',
+                'datatable.sort.NONE': 'activate to remove sorting on column',
                 'columntoggler.CLOSE': 'Close',
                 'overlaypanel.CLOSE': 'Close'
             }
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
@@ -465,9 +465,14 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
         this.sortableColumns.attr('tabindex', this.cfg.tabindex);
 
         //aria messages
-        this.ascMessage = PrimeFaces.getAriaLabel('datatable.sort.SORT_ASC');
-        this.descMessage = PrimeFaces.getAriaLabel('datatable.sort.SORT_DESC');
-        this.otherMessage = PrimeFaces.getAriaLabel('datatable.sort.SORT_LABEL');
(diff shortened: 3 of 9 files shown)


jepsar:
NL

'datatable.sort.UNSORTED': 'Activeer om kolomsortering op te heffen',

Although the key is somewhat misleading. I would have expected UNSORT instead of UNSORTED

christophs78:
German:
'datatable.sort.UNSORTED': 'Sortierung der Spalte entfernen',

melloware:
Thanks PR updated!

melloware:
@Rapster if you want to send me French translation as well i will add it.

stolp:
For German I think the following is more to the point:

'datatable.sort.UNSORTED': 'Spaltensortierung entfernen',

@Jepsar I agree that the label name UNSORTED is not the best as well. I suggest datatable.sort.NONE.

jepsar:
Removing the "activate to" part might be an option for all languages. In Dutch that would be:

        'datatable.sort.ASC': 'Kolom oplopend sorteren',
        'datatable.sort.DESC': 'Kolom aflopend sorteren',
        'datatable.sort.UNSORTED': 'Kolomsortering opheffen',

melloware:
OK made all the changes requested.  I like the idea of `NONE` instead of `UNSORTED`.

melloware:
A question for you guys.  Should I add the following to all other locales so people know its missing and needs to be translated?
```
datatable.sort.NONE=???
```

stolp:
Hmm, I think we should just let it fall back to the default properties automatically.

We already have this situation with other incomplete localizations as well, don't we?

melloware:
I know this works for Java Server Properties but wasn't sure if fallback works for this client side as well as I never tested it :)  I think you are right though i think it does work by checking the Locale File first then falling back to the defaults.

melloware:
Added Portugeuse and Spanish translations.

PR #7672, opened 2021-08-07
#1921 Integrationtest for DataTable: lazy loading pagination error

christophs78:
~~Look´s like this issue was already fixed during some other refactoring...
Please cross check whether i test the right things. :-)~~

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datatable/DataTable002.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datatable/DataTable002.java
@@ -25,10 +25,9 @@ package org.primefaces.integrationtests.datatable;
 
 import lombok.Data;
 import org.primefaces.event.SelectEvent;
+import org.primefaces.integrationtests.general.utilities.TestUtils;
 
 import javax.annotation.PostConstruct;
-import javax.faces.application.FacesMessage;
-import javax.faces.context.FacesContext;
 import javax.faces.view.ViewScoped;
 import javax.inject.Named;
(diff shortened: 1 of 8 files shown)


christophs78:
Figured out how to break it. Ready to get fixed. ;-)

christophs78:
@tandraschko , @Rapster : Now we have two mutations of the test. One works´s and one is broken. Compare the tiny little details between `ProgrammingLanguageLazyDataModel `and `ProgrammingLanguageLazyDataModelRowCountOtherImpl`. It currently makes a difference how `getRowCount` / `setRowCount` is implemented.

tandraschko:
yep, i see
i think ProgrammingLanguageLazyDataModel is correct
WDYT?

load should do 2 queries IMO:
- select count with all filters and apply it to setRowCount
- select entities with applied paging (first + pageSize) and filtering/sort

christophs78:
Yep. `ProgrammingLanguageLazyDataModel` is correct and should work. That´s what is done in most real-world-usecases which
fetch data from a database.
Ideally we should not break `ProgrammingLanguageLazyDataModelRowCountOtherImpl` in our attempt to fix this issue.

Currently IMO it´s not clearly defined what we expect from developers implementing LazyDataModel. Maybe we should extend JavaDoc of `LazyDataModel#load`. And add some note to https://primefaces.github.io/primefaces/11_0_0/#/components/datatable?id=lazy-loading.

tandraschko:
which one fails now?

christophs78:
ProgrammingLanguageLazyDataModel fails

PR #7675, opened 2021-08-08
Fix #7661: Avatar: Allow for fallback

### LINKED ISSUE #7661: Avatar: Fallback logic
**Description**

I'm just trying to get familiar with the new avatar widget in PrimeFaces 10.

From my point of view, the widget could become even more useful if it would implement some kind of fallback logic.

E.g.
* gravatar image -> letters or
* custom image -> gravatar image -> letters or
* custom image -> gravatar image -> icon

I tried to implement the first case using PF 10. However, this required two stacked avatar elements and lots of manual adjustments.

It would be great to have a single avatar widget that includes some sort of fallback logic.

* gravatar can be adjusted via `?d=404` to return a 404 HTTP status code, if there doesn't exist an image for a given e-mail address
* in this case the `img` can be hidden via its `onerror` attribute

jepsar:
Fix #7661

Basically we now render everything and set the position to absolute. As default we ask for a blank Gravatar image if not found to show the fallback text or icon. You can do the same for a child image or set rendered conditionally. 

```xhtml
<div class="card">
    <h5>Badge and Gravatar</h5>
    <p:badge value="4" severity="danger">
        <p:avatar size="xlarge" label="Walter">
            <p:graphicImage name="demo/images/organization/walter.jpg"/>
        </p:avatar>
    </p:badge>
    <p:avatar gravatar="contact@primetek.com.tr" gravatarConfig="d=mp" size="xlarge" styleClass="p-ml-3" />
    <p:avatar gravatar="notfound@avatar.com" label="Not Found" size="xlarge" styleClass="p-ml-3" />
    <p:avatar size="xlarge" label="Walter" styleClass="p-ml-3">
        <p:graphicImage name="demo/images/organization/walter.jpg" rendered="false"/>
    </p:avatar>
</div>
```
Screenshot 2021-08-08 at 09 00 40

### DIFF
--- a/docs/11_0_0/components/avatar.md
+++ b/docs/11_0_0/components/avatar.md
@@ -31,7 +31,7 @@ Avatar represents people using icons, labels and images.
 | lightness | 40 | Integer | Lightness of dynamicColor (0 .. 100).
 | alpha | 100 | Integer | Alpha of dynamicColor (0 .. 100).
 | gravatar | null | String | Gravatar e-mail address for the profile to be displayed.
-| gravatarConfig | d=blank | String | Optional Gravatar configuration for any other Gravatar API URL parameters.
+| gravatarConfig | null | String | Optional Gravatar configuration for any other Gravatar API URL parameters.
 
 ## Getting Started
 Avatar has three built-in display modes; "label", "icon" and custom content.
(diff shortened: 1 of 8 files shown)


melloware (approved):
Looks good to me.

PR #7677, opened 2021-08-08
Fix #7676: Badge: add visible attribute

### LINKED ISSUE #7676: Badge: add visible attribute
I'll create a PR for this.

jepsar:
Fix #7676


melloware:
This may be a nit but instead of `hidden="true"` wouldn't `visible="false"` be better?  visible obviously true by default?

Only reason I say that is because `visible` is what is used on p:column for similar behavior.

jepsar:
+1 for consistency, I'll change it

PR #7687, opened 2021-08-10
#7686 add touchswipe dependency to Data[Grid|List|View]

### LINKED ISSUE #7686: DataView: misses ResourceDependency to touch/touchswipe.js
**Describe the defect**
When using DataView with a pager in mobile/touch modus, there will be an exception in the console 
'Uncaught TypeError: a.swipe is not a function'.
The same is true for DataGrid and DataList. All components missing the ResourceDependency to touch/touchswipe.js.

**Reproducer**
See example code below.

**Environment:**
 - PF Version: 10.0
 - JSF + version: does not matter
 - Affected browsers:  FF, all I guess

**To Reproduce**
Use example code provided. In Firefox using developer tools, to simulate a touch device. Look at the JS-Console:
Uncaught TypeError: a.swipe is not a function

**Expected behavior**
No exception.

**Example XHTML**
```html
<h:form id="frmTest">
            <p:dataView id="dataView"
                        columns="1"
                        value="#{testView.entries}"
                        var="_it"
                        paginator="true"
                        paginatorAlwaysVisible="true"
                        widgetVar="dataGridWidgetVar">
                <p:column>#{_it}</p:column>
            </p:dataView>
</h:form>
```

**Example Bean**
```java
@Named
@ViewScoped
    private List<String> entries = Arrays.asList("a", "b", "c", "d", "e", "f", "g");

    public List<String> getEntries() {
        return entries;
    }
}
```

OBreidenbach:


melloware:
Good catch and thanks for the PR.

PR #7709, opened 2021-08-15
Fix #7457 DatePicker lazy model

### LINKED ISSUE #7457: DatePicker: lazy load disabled dates when panel is opened
**Describe the defect**
Lazy loading disabled dates was implemented in ticket #6355.
However, this doesn't work when opening the panel, no viewChangeEvent is fired. When changing months, the viewChangedEvent is fired correctly.

At the moment, when closing the panel, a viewChangeEvent is fired to set the the state of the dates of the current month to prepare the state the next time the panel is shown. However, this is not lazy. it should be initialized when showing the panel, not when hiding.

I think the following things should be changed;
When a LazyDateMetadataModel is used, ...
* a viewChangedEvent should be fired when the panel is opened.
* no viewChangedEvent should be fired when the panel is closed.

**Environment:**
 -  PF Version: 10.0
 -  JSF + version: Mojarra 2.3.9.SP09
 - Affected browsers: ALL (I could reproduce this with Chrome, Firefox and Edge, all updated to latest version. Haven't tested it with other browsers)

wal-jan:
View changed is not triggered any more if no event is present. This effects 2 JS functions `setDate` and `hideOverlay.onExited`. So view changed is not triggered any more by closing the popup.

Added attribute `refreshLazyModel` (default false). If `refreshLazyModel` is set to true and DatePicker has `viewChange` behaviour this behaviour is triggered before date panel is opened.

Added some DatePicker behaviour integration tests.

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datepicker/DatePicker011.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datepicker/DatePicker011.java
@@ -1,0 +1,105 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2021 PrimeTek
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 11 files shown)


jepsar:
I would like to see an answer on the questions in the issue first. If it is a bug, there is no need to add a new attribute.

tandraschko:
as discussed in the issue, i will close this for now

PR #7717, opened 2021-08-17
fix serializable when javax.faces.SERIALIZE_SERVER_STATE=true

redrede:
fix serializable when javax.faces.SERIALIZE_SERVER_STATE=true


christophs78:
Can you try to do your PR based on current MASTER-branch. (PrimeFaces was converted into a multimodule-project after 10.0.0-release.) This way we should avoid merge-conflicts.

Rapster:
Please create an issue first to describe your issue

tandraschko:
The change doesnt make sense for me
Please provide a issue and reproducer

PR #7723, opened 2021-08-19
#7336: one more integrationtest

christophs78:
... and better empty-message-handling for PrimeSelenium DataTable

### DIFF
--- a/primefaces-selenium/primefaces-selenium-components/src/main/java/org/primefaces/selenium/component/DataTable.java
+++ b/primefaces-selenium/primefaces-selenium-components/src/main/java/org/primefaces/selenium/component/DataTable.java
@@ -24,6 +24,7 @@
 package org.primefaces.selenium.component;
 
 import org.openqa.selenium.By;
+import org.primefaces.selenium.PrimeSelenium;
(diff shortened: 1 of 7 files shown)


primefaces-selenium/primefaces-selenium-components/src/main/java/org/primefaces/selenium/component/DataTable.java
```diff
@@ -37,10 +38,12 @@
 public abstract class DataTable extends AbstractTable<Row> {
 
     public List<Row> getRows() {
-        return getRowsWebElement().stream().map(rowElt -> {
-            List<Cell> cells = rowElt.findElements(By.tagName("td")).stream().map(cellElt -> new Cell(cellElt)).collect(Collectors.toList());
-            return new Row(rowElt, cells);
-        }).collect(Collectors.toList());
+        return getRowsWebElement().stream()
+                .filter(rowElt -> !PrimeSelenium.hasCssClass(rowElt, "ui-datatable-empty-message"))
```
christophs78, line 42: I added this to getRows. So we do not need to handle this later any more.

tandraschko:
looks good to me

PR #7728, opened 2021-08-20
Fix #7712: OutputPanel: Add skeleton facet to output panel

### LINKED ISSUE #7712: OutputPanel: Update deferred behavior based on Scaffolding
**Description**
Thanks to the great work of @jepsar PrimeTek has asked we port the Scaffolding component to core PF.

See: https://www.primefaces.org/showcase-ext/sections/scaffolding/basicUsage.jsf

This will allow the Skeleton component to be useful and provide a true Skeleton experience users have come to expect with other languages.

```xml
<p:scaffolding ready="#{not empty scaffoldingController.data}" id="scaffolding"
                loader="#{scaffoldingController.fetchData(2000)}">
    <f:facet name="loading">
        <p:skeleton width="3rem" height="3rem" shape="circle"/>
    </f:facet>
    <f:facet name="ready">
        <p:avatar label="#{scaffoldingController.data}" size="large" shape="circle" />
    </f:facet>
</p:scaffolding>
```

jepsar:
PR 1/2

Fix #7712 (I think we should create a new issue for the second PR)

### DIFF
--- a/docs/11_0_0/components/outputpanel.md
+++ b/docs/11_0_0/components/outputpanel.md
@@ -75,15 +75,15 @@ one part of the page that has components dealing with backend and taking time, w
 on, rest of the page is loaded instantly and time taking process is loaded afterwards with ajax.
 
 ### Skeletons
-Deferred loading can be used with the `skeletons` facet to show UI while the data is loaded. This
+Deferred loading can be used with the `skeleton` facet to show UI while the data is loaded. This
 is normally combined with the load ajax event and a listener which loads the data you want to show
 in your panel. For example:
 
 ```xhtml
 <p:outputPanel deferred="true">
     <p:ajax event="load" listener="#{bean.loadData}"/>
-    <f:facet name="skeletons">
-        <p:skeleton width="100px" height="21px" class="p-mb-4"/>
+    <f:facet name="skeleton">
+        <p:skeleton width="200px" height="21px" class="p-mb-4"/>
         <p:skeleton width="100%" height="63px"/>
     </f:facet>
     <h5>#{bean.data.title}</h5>
--- a/docs/11_0_0/components/outputpanel.md
+++ b/docs/11_0_0/components/outputpanel.md
@@ -75,8 +75,8 @@ one part of the page that has components dealing with backend and taking time, w
 on, rest of the page is loaded instantly and time taking process is loaded afterwards with ajax.
 
 ### Skeletons
-Deferred loading can be used with the `skeleton` facet to show UI while the data is loaded. This
-is normally combined with the load ajax event and a listener which loads the data you want to show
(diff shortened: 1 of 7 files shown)


docs/11_0_0/components/outputpanel.md
```diff
@@ -74,6 +74,23 @@ page scroll as well. This feature is very useful to increase page load performan
...
 
+### Skeletons
+Deferred loading can be used with the `skeletons` facet to show UI while the data is loaded. This
+is normally combined with the load ajax event and a listener which loads the data you want to show
+in your panel. For example:
+
+```xhtml
+<p:outputPanel deferred="true">
+    <p:ajax event="load" listener="#{bean.loadData}"/>
+    <f:facet name="skeletons">
```
melloware: facets usually are not plural so "skeleton" instead of "skeletons"?

melloware (changes requested):

melloware (approved):

primefaces/src/main/java/org/primefaces/component/outputpanel/OutputPanelRenderer.java
```diff
@@ -24,19 +24,33 @@
...
+import org.primefaces.util.ComponentUtils;
 import org.primefaces.util.WidgetBuilder;
 
 public class OutputPanelRenderer extends CoreRenderer {
 
     private static final String BLOCK = "div";
     private static final String INLINE = "span";
 
+    @Override
+    public void decode(final FacesContext context, final UIComponent component) {
```
tandraschko: remove "final"

primefaces/src/main/java/org/primefaces/component/outputpanel/OutputPanelRenderer.java
```diff
@@ -89,7 +109,7 @@ protected void encodeScript(FacesContext context, OutputPanel panel) throws IOEx
         wb.finish();
     }
 
-    protected void renderLoading(FacesContext context, OutputPanel panel) throws IOException {
+    protected void renderLoading(FacesContext context) throws IOException {
```
tandraschko: dont remove the panel param, might be used for overloading and doesnt hurt

melloware (approved):

Rapster:
Unless the facet have to be a skeleton, (but I don't think so from what I saw already) I don't think "skeleton" is the right name for this. Maybe "loading"?

jepsar:
It does not need to contain skeleton components per se, it can contain anything. I do like the term skeleton for the facet though.

tandraschko:
in general "loading" as name would be a bit nicer, also for the new datascroller
it is the content, when the component is "loading" the new content

i can rename if you like

Rapster:
Yes please 😉

jepsar:
I'll rename it

PR #7734, opened 2021-08-22
OutputPanel: skeleton -> loading

jepsar:


primefaces/src/main/java/org/primefaces/component/outputpanel/OutputPanelRenderer.java
```diff
@@ -121,10 +121,11 @@ protected boolean isDeferredNecessary(FacesContext context, OutputPanel panel) {
         if (!panel.isDeferred()) {
             return false;
         }
-        if (panel.getValueExpression(OutputPanel.PropertyKeys.loaded.name()) != null) {
+        boolean isAjaxRequest = context.getPartialViewContext().isAjaxRequest();
+        if (isAjaxRequest && panel.getValueExpression(OutputPanel.PropertyKeys.loaded.name()) != null) {
```
jepsar: @Rapster you mean like this?

melloware (approved):

primefaces/src/main/java/org/primefaces/component/outputpanel/OutputPanelRenderer.java
Rapster: Almost, what I had in mind was:
```java
    protected boolean isDeferredNecessary(FacesContext context, OutputPanel panel) {
        boolean deferred = !context.getPartialViewContext().isAjaxRequest() && panel.isDeferred();
        if (deferred && panel.getValueExpression(OutputPanel.PropertyKeys.loaded.name()) != null) {
            deferred = !panel.isLoaded();
        }
        return deferred;
    }
```

jepsar: Am I missing something? How is this going to work with Ajax requests?

Rapster: I'm guessing everyone missed that comment https://github.com/primefaces/primefaces/issues/7712#issuecomment-901944480 😋 Not really happy with that fix in the past, as I ended up with a bug (same as mentionned in the comment). I'll try something

jepsar: I'll revert this bit for now to reduce this PR to just the facet rename

Rapster: Sounds good

Issue #7737, opened 2021-08-23
Datatable throws ListDataModel cast exception


morvael:
I think there was a bug introduced somewhere due to #7391, because this didn't happen for us on 10.0.5 and suddenly appeared on 10.0.6. The problem happens when selecting rows in a filtered datatable with lazy model.

```
Caused by: java.lang.ClassCastException: class pl...DataModel cannot be cast to class java.util.List (pl...DataModel is in unnamed module of loader 'deployment...-web-1.war' @3428e777; java.util.List is in module java.base of loader 'bootstrap')
	at javax.faces.api@3.0.0.SP04//javax.faces.model.ListDataModel.setWrappedData(ListDataModel.java:200)
	at deployment...-web-1.war//org.primefaces.model.SelectableDataModelWrapper.<init>(SelectableDataModelWrapper.java:31)
	at deployment...-web-1.war//org.primefaces.component.datatable.DataTable.updateValue(DataTable.java:306)
	at deployment...-web-1.war//org.primefaces.component.datatable.feature.SelectionFeature.decode(SelectionFeature.java:55)
	at deployment...-web-1.war//org.primefaces.component.datatable.DataTableRenderer.decode(DataTableRenderer.java:65)
	at javax.faces.api@3.0.0.SP04//javax.faces.component.UIComponentBase.decode(UIComponentBase.java:509)
	at deployment...-web-1.war//org.primefaces.component.api.UIData.processDecodes(UIData.java:109)
	at com.sun.jsf-impl@2.3.14.SP04//com.sun.faces.context.PartialViewContextImpl$PhaseAwareVisitCallback.visit(PartialViewContextImpl.java:606)
	at com.sun.jsf-impl@2.3.14.SP04//com.sun.faces.component.visit.PartialVisitContext.invokeVisitCallback(PartialVisitContext.java:159)
	at deployment...-web-1.war//org.primefaces.component.api.UIData.visitTree(UIData.java:683)
	at javax.faces.api@3.0.0.SP04//javax.faces.component.UIComponent.visitTree(UIComponent.java:1468)
	at javax.faces.api@3.0.0.SP04//javax.faces.component.UIComponent.visitTree(UIComponent.java:1468)
	at javax.faces.api@3.0.0.SP04//javax.faces.component.UIComponent.visitTree(UIComponent.java:1468)
	at deployment...-web-1.war//org.primefaces.component.api.UITabPanel.visitTree(UITabPanel.java:931)
[...]

PR #7741, opened 2021-08-23
#7737: integrationtests (DataTable + lazy + filter + single/multiple selection)

christophs78:


christophs78:
IT for this combination with multiple selection already existed.

PR #7747, opened 2021-08-24
Schedule: Fix views options with extender (#7746)

### LINKED ISSUE #7746: Schedule: Does not respect "views" option set by an extender
**Describe the defect**

When an extender sets a value for `options.views`, the schedule widget does not respect that value, e.g.

```xml
<p:schedule widgetVar="schedule" extender="function(){ this.cfg.options.views = {dayGridDay: {dayMaxEvents: 6} } }" />
```

Then inspect `PF("schedule").cfg.options.views` in the browser console, it's missing the `dayMaxEvents` option set by the extender.

**Environment:**
 - PF Version: _10.0_

**Reproducer**

See the showcase https://www.primefaces.org/showcase/ui/data/schedule/extender.xhtml and use the following extender:

```javascript
$.extend(this.cfg.options, {
    dayMaxEvents: 6,
    views: {dayGridDay: {dayMaxEvents: 6} },
});
```

image

blutorange:
Fixes: #7746 Need to use `this.cfg.options.views` instead of `this.cfg.views`. I kept the fall back to `this.cfg.views` in order not to break existing code. Could be removed in the next major releaes.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/schedule/1-schedule.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/schedule/1-schedule.js
@@ -347,7 +347,9 @@ PrimeFaces.widget.Schedule = PrimeFaces.widget.DeferredWidget.extend({
             }
         }
 
-        this.cfg.options.views = this.cfg.views||{};
+        // Using this.cfg.views was a bug, but fall back to that for backwards compatibility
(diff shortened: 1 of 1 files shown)


melloware (approved):

melloware:
I don't think it hurts to leave the fallback in but thanks for the PR!

PR #7752, opened 2021-08-25
Fix #7336: MyFaces DataTable#filterAndSort

### LINKED ISSUE #7336: DataTable: Removed data (rows) remain after update when filtered before

christophs78:


melloware:
So are we expecting this failure until its fixed?
```
Error:  Failures: 
Error:    DataTable021Test.testGlobalFilterAndRemoveRowV2:237 expected: <1> but was: <2>
```

christophs78:
Yes. `DataTable#filterAndSort` does not work on MyFaces.
Maybe @tandraschko has some idea. Or we need to remove `DataTable#filterAndSort` we just added a few months ago.

tandraschko:
can you explain what exactly does not work?

christophs78:
I hoped it would be clear from looking at the failing integrationtest...
I try to describe.

DataTable is bound to a list auf 5 ProgrammingLanguages we remove one with the following code:
```java
public void removeV2(ProgrammingLanguage programmingLanguage) {
	this.progLanguages.remove(programmingLanguage);
	DataTable dataTable = (DataTable) FacesContext.getCurrentInstance().getViewRoot().findComponent("form:datatable");
	dataTable.filterAndSort();
}
```

When we debug FilterFeature#filter (called by DataTable#filterAndSort) it looks on Mojarra like this:
Mojarra_FilterFeature

On MyFaces (2.3; NOT next) it looks like this:
MyFaces_FilterFeature

This results in MyFaces calculating filteredValue-List wrong.

tandraschko:
it only fails in 2.3 but not in 2.3-next?

christophs78:
Can‘t say for 2.3-next. (Because wie can not replace MyFaces within TomEE yet.)

christophs78:
Look´s like `table.setValue(null);` has "limitations" on MyFaces. (setValue is implemented by UIData)

Mojarra is (in this point) straight forward:
```java
public void setValue(Object value) {
	setDataModel(null);
	getStateHelper().put(PropertyKeys.value, value);
}
```

MyFaces 2.3 tries to do some magic which seems to be worse in this usecase:
```java
public void setValue(Object value)
{
	getStateHelper().put(PropertyKeys.value, value );
	_dataModelMap.clear();
	_rowStates.clear();
	_isValidChilds = true;
}
```

tandraschko:
further research would be great

primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java
```diff
@@ -1100,10 +1100,17 @@ public void setWidth(String width) {
...
+        if (isLazy() || getFilteredValue() == null) {
             return;
         }
 
+        /*
+         * setDataModel is defined by UIData. So different implementations for Mojarra and MyFaces.
+         * But PrimeFaces comes with it´s own UIData which extends/modifies UIData provided by JSF-impl.
+         * But PrimeFaces UIData does not know all impl-specifics, so ....
+         */
+        setDataModel(null); // for MyFaces 2.3 - compatibility
```
christophs78, line 1112: Seems like we have kind of a solution.
But hard to find good words for PrimeFaces-UIData-thing.

tandraschko:
is the PR ready now?

christophs78:
> is the PR ready now?

Yes

melloware:
Looks like this is intermittently failing on MyFACES 2.3.X?  
```
 DataTable021Test.testGlobalFilterAndRemoveRowV1:278 » StaleElementReference st...
```

christophs78:
Think i have some timing-issue. (does not happy always & happens for different tests)

```
Error:  Errors: 
Error:    DataTable021Test.testGlobalFilterAndRemoveRowV1:278 » StaleElementReference st...

Error:  Errors: 
Error:    DataTable021Test.testGlobalFilterAndRemoveRowV2:335 » StaleElementReference st...
```

PR #7755, opened 2021-08-25
Carousel reimplementation minor updates

melloware:
1. Added touchswipe.js back in because this component uses Swipe it needs this JS included.

2. Added `refresh` method back in which if the component is updated it stops the current autoplay. This was is in the old component and definitely fixed a previously reported issue.


buunsal:
I think looks good 👍🏻

PR #7756, opened 2021-08-25
Fix #3926: Tree : Multiple selection checkbox propagation inconsistent with filter option

### LINKED ISSUE #3926: Tree : Multiple selection checkbox propagation inconsistent with filter option
## 1) Environment
- PrimeFaces version: 6.2 
- Application server + version:  Apache Tomcat/8.5.32
- Affected browsers: Firefox, Chrome

## 2) Expected behavior
With or without filtering, I expected to see the same results on selection. Filtering shouldn't change the selected nodes.

image

## 3) Actual behavior
All the parent nodes are selected automatically after filtering and only selecting one single node, not all the children of that node.
image

## 4) Steps to reproduce
Example is below.

## 5) Sample XHTML
    <h:form id="selection">
        <p:growl id="msgs" showDetail="true" escape="false"/>
        <h3>Checkbox</h3>
        <p:tree id="multiCourtTree" value="#{propagationSelectionView.root}" var="_doc"
                    selectionMode="checkbox" propagateSelectionUp="true" propagateSelectionDown="true"
                    filterBy="#{_doc.name}" filterMatchMode="contains"
                    selection="#{propagationSelectionView.selectedNodes}">
            <p:treeNode>
                <h:outputText value="#{_doc.name}"/>
            </p:treeNode>
        </p:tree>
    </h:form>

## 6) Sample bean

public class PropagationSelectionView {

    private TreeNode root;
    private TreeNode[] selectedNodes;

    @PostConstruct
    public void init() {
        root = createCheckboxDocuments();
    }

    public TreeNode createCheckboxDocuments() {
        TreeNode root = new CheckboxTreeNode(new Document("Files"), null);

        TreeNode documents = new CheckboxTreeNode(new Document("Documents"), root);
        TreeNode work = new CheckboxTreeNode(new Document("Work"), documents);
        TreeNode primefaces = new CheckboxTreeNode(new Document("PrimeFaces"), documents);

        //Documents
        TreeNode expenses = new CheckboxTreeNode(new Document("Expenses.doc"), work);
        TreeNode resume = new CheckboxTreeNode( new Document("Resume.doc"), work);
        TreeNode refdoc = new CheckboxTreeNode(new Document("RefDoc.pages"), primefaces);

        return root;
    }

    public TreeNode getRoot() {
        return root;
    }

    public TreeNode[] getSelectedNodes() {
        return selectedNodes;
    }

    public void setSelectedNodes(TreeNode[] selectedNodes) {
        this.selectedNodes = selectedNodes;
    }

    public class Document implements Serializable, Comparable<CheckboxDocument>{

        private String name;

        public Document(String name) {
            this.name = name;
        }

        public String getName() {
            return name;
        }

            this.name = name;
        }

        @Override
        public String toString() {
            return name;
        }

        public int compareTo(CheckboxDocument document) {
            return this.getName().compareTo(document.getName());
        }
    }
}

## 7)Conclusion

Selected Nodes should behave in the same way on whole tree structure regardless of filtering.

FlipWarthog:
Still needs work (See #3926 for detail).

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/tree/TreeRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/tree/TreeRenderer.java
@@ -669,6 +669,7 @@ public class TreeRenderer extends CoreRenderer {
 
         List<String> filteredRowKeys = tree.getFilteredRowKeys();
         boolean match = false;
+        boolean hidden = false;
         if (filter && !filteredRowKeys.isEmpty()) {
             for (String filteredRowKey : filteredRowKeys) {
                 String rowKeyExt = rowKey + "_";
@@ -688,7 +689,7 @@ public class TreeRenderer extends CoreRenderer {
             }
 
             if (!match) {
-                return;
+                hidden = true;
             }
         }
 
@@ -715,6 +716,10 @@ public class TreeRenderer extends CoreRenderer {
         }
         else {
             containerClass += " ui-treenode-unselected";
+        }
+
+        if (hidden) {
+            containerClass += " ui-treenode-hidden";
         }
 
         containerClass = uiTreeNode.getStyleClass() == null ? containerClass : containerClass + " " + uiTreeNode.getStyleClass();
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/tree/tree.css
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/tree/tree.css
@@ -71,6 +71,10 @@
     display: inline-block;
 }
 
+.ui-tree .ui-treenode-hidden {
+    display: none;
+}
+
 /** Horizontal Tree **/
 .ui-tree-horizontal {
     width:auto;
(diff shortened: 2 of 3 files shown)


melloware:
Looks good to me.  So is there 1 open scenario left or more than 1?

FlipWarthog:
Yeah, I still need to tackle the case where children are filtered and the parent checkbox is clicked. Currently it selects the parent and all the children, even hidden ones. I think it should only select the non-hidden ones and partial select the parent (assuming the hidden ones aren't already checked).

FlipWarthog:
Alright, @melloware, I was able to just borrow some of the logic from the upward propagation and apply it to the downward propagation (ignore my latest commit message, I had it backward).

So, in the same way that if you select a single child, it'll partially select the parent, now, when you apply a filter then select the parent, it selects all visible children then checks all the children to determine if the selected parent should be fully checked or partially checked.

Let me know what you think.

melloware:
This looks great to me! Nice work!

melloware (approved):

PR #7760, opened 2021-08-26
some 11.0.0 - preparations

christophs78:


tandraschko:
i think we can merge it?

christophs78:
Yes, we can merge this one.

(PrimeTek can do further updates afterwards, when they migrate showcase and/or PF-components to PrimeFlex 3)

PR #7766, opened 2021-08-27
Fix DataTable021Test

christophs78:


primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datatable/DataTable021Test.java
```diff
@@ -353,7 +353,7 @@ public void testGlobalFilterAndRemoveRowV2(Page page) {
         InputText globalFilter;
 
         @FindBy(id = "form:datatable:buttonDeleteRow")
-        InputText buttonDeleteRow;
+        CommandButton buttonDeleteRow;
```
christophs78, line 356: @melloware : Had a stupid mistake. So probably guardAjax did not kick in.

melloware: HA yep

PR #7771, opened 2021-08-30
Fix #6397:Slider with InputNumber two way sync

### LINKED ISSUE #6397: Slider: InputNumber two way sync is not working

melloware:
Use `input` event instead of `change` event.  The `val` method of Jquery does not fire any events so we must `trigger` one.

Also on the hidden input only call `input.slider` to trigger the slider input and not a general `input`.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js
@@ -178,7 +178,7 @@ PrimeFaces.widget.InputNumber = PrimeFaces.widget.BaseWidget.extend({
      * @param {string} value A value to set on the hidden input.
      */
     setValueToHiddenInput: function(value) {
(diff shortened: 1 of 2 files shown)


melloware:
@sqores i have tested the PRO users issue with this fix and it fixes the original issue and the PRO issue.

buunsal (approved):

PR #7783, opened 2021-09-02
Fix #7781: TriStateCheckbox add readonly attribute

### LINKED ISSUE #7781: TriStateCheckbox: Add readonly attribute
**Description**
The `<p:triStateCheckbox>` is missing a `readonly` attribute that allows to render the checkbox in a non-disabled fashion, still maintaining it read-only.
It's not even possible to simulate such a behaviour with something like `onclick="return false"` because this component is also missing the `onclick` attribute.

Adding the ability to set `<p:triStateCheckbox readonly="true" />` to make the tri state checkbox look enabled but readonly. This is also useful if you want to submit its value in a form although not letting the user change it.

It seems like this was implemented in primeng: https://github.com/primefaces/primeng/issues/2999

melloware:
I followed what PrimeNG was doing.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/tristatecheckbox/TriStateCheckboxRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/tristatecheckbox/TriStateCheckboxRenderer.java
@@ -124,9 +124,10 @@ public class TriStateCheckboxRenderer extends InputRenderer {
         ResponseWriter writer = context.getResponseWriter();
         String styleClass = createStyleClass(checkbox, null, HTML.CHECKBOX_BOX_CLASS) ;
         styleClass = (valCheck == 1 || valCheck == 2) ? styleClass + " ui-state-active" : styleClass;
+        styleClass = disabled ? styleClass + " ui-state-disabled" : styleClass;
         styleClass = readonly ? styleClass + " ui-chkbox-readonly" : styleClass;
 
-        //if stateIcon is defined use it insted of default icons.
+        //if stateIcon is defined use it instead of default icons.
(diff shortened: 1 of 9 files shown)


melloware:
Working on Integration tests next.

primefaces/src/main/java/org/primefaces/component/tristatecheckbox/TriStateCheckboxRenderer.java
```diff
@@ -119,11 +120,11 @@ protected void encodeInput(final FacesContext context, final TriStateCheckbox ch
     }
 
     protected void encodeOutput(final FacesContext context, final TriStateCheckbox checkbox, final int valCheck,
-                                final boolean disabled) throws IOException {
+                                final boolean disabled, boolean readonly) throws IOException {
         ResponseWriter writer = context.getResponseWriter();
         String styleClass = createStyleClass(checkbox, null, HTML.CHECKBOX_BOX_CLASS) ;
         styleClass = (valCheck == 1 || valCheck == 2) ? styleClass + " ui-state-active" : styleClass;
-        styleClass = disabled ? styleClass + " ui-state-disabled" : styleClass;
```
mauromol, line 126: I've not tested and I didn't investigate, but are you sure this line should be deleted?

melloware: No this is a mistake!

melloware:
Added integrations tests for enabled, disabled, and readonly and discovered a bug while writing those tests.

Issue #7729, opened 2021-08-21
Button: dialog refreshes the page after open


dima-fedotovs:
When you open a dialog via PF('dlg').show() - it appears on the screen for a moment and then the page is reloaded. Affected versions (ELITE) 10.0.5 and 10.0.6. **It isn't reproducible with 10.0.4**

The demo project can be found in GitHub - https://github.com/dima-fedotovs/PFDialogReproducer

**Environment:**
 - PF Version: __10.0.5__, __10.0.6__
 - JSF + version: __Mojarra 2.3.14.SP04__
 - App server: __WildFly 23.0.2__ _(no configuration required)_
 - Affected browsers: __ALL__ _(not a browser specific)_

**To Reproduce**
1. Go to 'index.xhtml'
2. Click on the 'Show Dialog' button
4. Dialog appears on the screen for a moment and then the page is reloaded

**Expected behavior**
The page shouldn't be reloaded after the dialog is open.

**Example XHTML**
```html
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml"
      xmlns:p="http://primefaces.org/ui"
      xmlns:h="http://xmlns.jcp.org/jsf/html">
<h:head>
    <title>Hello</title>
</h:head>
<h:body>
    <p:button onclick="PF('dlg').show()" value="Show Dialog"/>

    <p:dialog header="Greeting" widgetVar="dlg">
        <h1>HELLO</h1>
    </p:dialog>
</h:body>
</html>
```

**Example Bean**
No beans are required to reproduce

dima-fedotovs:
After further investigation, I've found that it isn't p:dialog. Seem in PF 10.0.5 the p:button changed its behaviour. Now it is navigating to the same view after click.

dima-fedotovs:
Probably related to #7461

melloware:
I agree @tandraschko this does look related to #7461

tandraschko:
its actually the same with 8.0
if you check the rendered xhtml:
`onclick="PF('dlg').show();window.open('\/primefaces\-test\/index.xhtml','_self')"`

tandraschko:
yeah, its by design
if you check h:button rendering, its the same
so you MUST add "return false" after your statement

mertsincan:
Hi,

In a semantic version such as 10.0.x, this change should not be made. So sorry for the confusion. I'll revert it on PF10.0.8.

[...]

PR #7788, opened 2021-09-03
Fix #7772: FileUpload: Explain show and user activation

### LINKED ISSUE #7772: FileUpload: Explain client side API show() and user activation
https://www.primefaces.org/showcase/ui/file/upload/basic.xhtml (so PrimeFaces 10.0.0, but I don't think that matters) 

```js
PF('widget_j_idt302_j_idt303').show()
```

Logs this error in the JS console:

```
<input> picker was blocked due to lack of user activation.
```

Firefox version: 91.0.2

No longer works with Chrome (it used to). Chrome reports:

```
File chooser dialog can only be shown with a user activation.
```

jepsar:
Fix #7772

Ended up in a documentation note

### DIFF
--- a/docs/10_0_0/components/fileupload.md
+++ b/docs/10_0_0/components/fileupload.md
@@ -421,3 +421,23 @@ Widget: _PrimeFaces.widget.FileUpload_
 | Method | Params | Return Type | Description |
 | --- | --- | --- | --- |
 show() | - | void | Shows file(s) chooser dialog.
+
+### `show` and user activation
+
+Note that `show()` can only be triggered from a user event in some browsers. For example, in Firefox the following
+code will **not** work due to a lack of user activation:
+
+```xhtml
+<p:fileUpload widgetVar="upload"/>
+<script>$(function(){PF('upload').show()})</script>
+```
+
+The following code **will** work in Firefox, because it is triggered from a click event triggered by the user:
+
+```xhtml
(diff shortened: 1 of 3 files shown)


Rapster:
` onclick="PF('upload')` _(without return false)_ works fine with firefox 92, it seems it was a bug in chromium that has been fixed 10 months ago

jepsar:
I was following https://github.com/primefaces/primefaces/issues/7729

I can take out `;return false` though, no problem

Rapster:
I think the only thing that has to be said is the version of chromium in which the issue has been raised and fixed: two things I can't find :( 

I might have misread something, not due to chromium but the browser itself and problem has been seen and fixed in 85

jepsar:
We can avoid using a `p:button` in the documentation and replace it with

```xhtml
<a href="javascript:PF('upload').show()">Choose</a>
```

melloware (approved):

Rapster:
I don't know... I don't feel there is not much to say in the doc... IMO, I wouldn't write anything since the problem is for a very specific version of Firefox. If user have this issue, they will google it, end up on that issue and get the answer, which is to update their web browser

jepsar:
Fine with me. @melloware shall I close it and can you remove the milestone and mark as won't fix?

melloware:
Yep close this and I will close the ticket as WontFix.

PR #7794, opened 2021-09-05
Fix #5492: jQuery UI 1.13

### LINKED ISSUE #5492: Upgrade jQuery UI 1.13

melloware:
Upgraded to jQuery UI 1.13.  Tested locally but submitting this PR to make sure all integration tests pass.


tandraschko:
It adds many new styles, not sure
Im sure they will conflict with ours or themes

I think we need more thinking how we should use jquery ui in general

melloware:
The styles can definitely be removed but so far in my testing I have not seen any issues.   And this fixes all the issues with JQMigrate making this version compatible with JQuery 3.6.0

tandraschko:
I think we should remove the styles as they are Not used
It just blows up styling on the client

melloware:
Styles removed.

tandraschko:
did you a extensive test in the showcase? If yes -> +1

melloware:
Yes I hit pretty much every function in the showcase looking for errors and saw none, tested Effects, Calendar (even though its hidden), and poked around on all the pages.

PR #7831, opened 2021-09-15
autocomplete.md doc updated

Abdullah8006:
Fixed the variable declaration.


melloware:
can you fix it in the 10 and 11 MD as well if its wrong?

Abdullah8006:
> can you fix it in the 10 and 11 MD as well if its wrong?

Sure I can. Can you please confirm if the fix is correct?

melloware:
yes those pipes are incorrect...probably a search and replace error.

melloware (approved):

PR #7833, opened 2021-09-15
Fix #7819: AjaxBehaviorListenerImpl#processAjaxBehavior: handle PropertyNotFoundException like MethodNotFoundException

### LINKED ISSUE #7819: AjaxBehaviorListenerImpl#processAjaxBehavior: handle PropertyNotFoundException like MethodNotFoundException
Complicated issue together wie Composite Components.

Mojarra "converts" MethodNotFoundException into PropertyNotFoundException with following code.
https://github.com/eclipse-ee4j/mojarra/blob/2.3.14-RELEASE/impl/src/main/java/com/sun/faces/facelets/el/ContextualCompositeMethodExpression.java#L176

PropertyNotFoundException ist not handeled by https://github.com/primefaces/primefaces/blob/2e79cbc628fa98885cad4688813fc7e82f6f3a97/primefaces/src/main/java/org/primefaces/behavior/ajax/AjaxBehaviorListenerImpl.java#L73
So PF-Fallback-code does not work.

IMO we also should catch PropertyNotFoundException and handle it like MethodNotFoundException.

There had been a similar issue in the past: https://github.com/primefaces/primefaces/issues/1375
This was adressed in a similar way.

christophs78:


melloware (approved):

primefaces/src/main/java/org/primefaces/behavior/ajax/AjaxBehaviorListenerImpl.java
```diff
@@ -77,6 +78,10 @@ public void processAjaxBehavior(AjaxBehaviorEvent event) throws AbortProcessingE
         catch (ArrayIndexOutOfBoundsException e) {
             processArgListener(context, elContext, event);
         }
+        // Mojarra hack for Composite Components, see #7819
+        catch (PropertyNotFoundException e) {
+            processArgListener(context, elContext, event);
+        }
```
melloware, line 80: Should this just be a multi catch `ArrayIndexOutOfBounds | PropertyNotFoundException` with the comment saying `// PropertyNotFoundException is Mojarra hack for Composite Components, see #7819`

tandraschko:
you need to add this 2 times
i currently have time, i copy paste it - sorry ;D

PR #7838, opened 2021-09-16
[Carousel, SpeedDial, Datepicker] Use JQuery for events instead of  native

### LINKED ISSUE #7824: New Component: SpeedDial
SpeedDial is a component that consists of many actions and a floating action button. When pressed, a floating action button can display multiple primary actions that can be performed on a page.

blutorange:
As was discussed we should always use `JQuery(...).on` for listening to events instead of the native `addEventListener`. I also removed the unused event param.

They are three components I found that did not use JQuery
* Carousel
* SpeedDial
* DatePicker JQueryUI widget

Tried it in the playground and does not seem to break anything.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/carousel/carousel.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/carousel/carousel.js
@@ -493,11 +493,11 @@ PrimeFaces.widget.Carousel = PrimeFaces.widget.DeferredWidget.extend({
         var $this = this;
 
         if (!this.documentResizeListener) {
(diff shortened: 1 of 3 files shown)


melloware (approved):

blutorange:
Not  sure about why that one run fails, but probably unrelated to this PR. Running `mvn -B -V clean install -fprimefaces-integration-tests/pom.xml -Pintegration-tests,headless,chrome,theme-saga,csp,myfaces-2.3.x` locally also works.

melloware:
Let me just kick it off again. Might be a hiccup.

PR #7842, opened 2021-09-17
DataTable/TreeTable: cellEditor saves changes too late #7776

### LINKED ISSUE #7546: DataTable: onCellEdit event fires too late
**Describe the defect**
Assume we have datatable whit cell we trying edit and some button below that submit that has ajax action
Then
first scenario: we edit datatable cell (enter new value), do mouse click on some form place (to imitate focus lost) - here onCellEdit 
 event will fire, then click the button.
So - **onCellEdit will fire first, then button action will fire second**
second scenario: we edit datatable cell (enter new value), then do mouse click directly on the button.
**The button action will fire first, onCellEdit will fire second**
But de-facto row cell also lost focus when we do click the button

**To Reproduce**
Go to edit datatable cell, change the value then click the mouse on the button that has ajax action event
Firs button action will be fired, then onCellEdit event

**Environment:**
 - PF Version: 10.0
 - Chrome 91

**Expected behavior**
OnCellEdit should fire BEFORE button action.

P.S.
I found the same problem description on StackOverflow, added long time ago:
https://stackoverflow.com/questions/26099330/pcelleditor-does-not-save-textarea-onblur-when-clicking-a-link
The author said he created a bug request in primefaces repo, but I did not found it in issues.

### LINKED ISSUE #7776: DataTable/TreeTable: cellEditor saves changes too late
**Describe the defect**
DataTable cellEditor saves changes when you click away from it. But if you click something like a commandButton, then the saving request gets queued after the button's, which is usually not what you want.

**Reproducer**
[primefaces-datatable-celledit-request-late.zip](https://github.com/primefaces/primefaces/files/7085911/primefaces-datatable-celledit-request-late.zip)

**Environment:**
 - JSF + version: _ALL_
 - Affected browsers: _ALL_

**To Reproduce**
2. Click on the "Initial" value in the table.
3. Change it to "123". Don't press Enter.
4. Click on the button.
5. Note the stdout:
```
Acting on value: Initial
Changing value to 123
```
The value was changed after the button's action.

**Expected behavior**
Saving request happens before whatever was triggered by the click away. Maybe by also handling focusin or mousedown, or mouseup?

**Example XHTML**
```
<h:form id="form">
	<p:dataTable id="datatable" value="#{testView.list}" var="row" style="width: 100px" editable="true" editMode="cell">
		<p:column headerText="Column">
			<p:cellEditor>
				<f:facet name="input">
					<p:inputText value="#{row.value}" />
				</f:facet>
				<f:facet name="output">
					#{row.value}
				</f:facet>
			</p:cellEditor>
		</p:column>
	</p:dataTable>
	<p:commandButton id="btn" value="Act" action="#{testView.act}" process="@this" />
</h:form>

```

**Example Bean**
```java
@Data
@Named
@ViewScoped

	private List<Bean> list = List.of(new Bean());

	public List<Bean> getList() {
		return list;
	}

	public void act() {
		System.out.println("Acting on value: " + list.get(0).getValue());
	}

	public static class Bean {

		private String value = "Initial";

		public String getValue() {
			return value;
		}

		public void setValue(String value) {
			System.out.println("Changing value to " + value);
			this.value = value;
		}

	}
    
}
```

VsevolodGolovanov:
Save/cancel on mouseup to que the event request before whatever was clicked reacts.

Got rid of incellClick and contextMenuClick, since the order of events changed. `if (currentCell.is(target) || currentCell.has(target).length)` takes care of their logic.
This does mean right clicking another cell would now end editing, but that seems acceptable.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
@@ -66,7 +66,6 @@
  * @prop {JQuery} checkAllTogglerInput DOM element of the `check all` checkbox in the header.
  * @prop {JQuery} clone Clone of the table header.
  * @prop {boolean} columnWidthsFixed Whether column widths are fixed or may be resized.
- * @prop {boolean} contextMenuClick Whether the context menu was clicked.
  * @prop {PrimeFaces.widget.ContextMenu} contextMenuWidget Widget with the context menu for the DataTable.
  * @prop {JQuery} currentCell Current cell to be edited.
  * @prop {number | null} cursorIndex 0-based index of row where the the cursor is located.
@@ -83,7 +82,6 @@
  * @prop {boolean} hasColumnGroup Whether the table has any column groups.
  * @prop {JQuery} headerTable The DOM elements for the header table.
  * @prop {JQuery} headers DOM elements for the `TH` headers of this DataTable.
- * @prop {boolean} incellClick Whether a click occurred inside a table cell.
  * @prop {boolean} isRTL Whether the writing direction is set to right-to-left.
  * @prop {boolean} isRowTogglerClicked Whether a row toggler was clicked.
  * @prop {boolean} liveScrollActive Whether live scrolling is currently active.
@@ -1310,7 +1308,6 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
             targetWidget.contextMenuCell.removeClass('ui-state-highlight');
         }
 
-        targetWidget.contextMenuClick = true;
         targetWidget.contextMenuCell = cell;
         targetWidget.contextMenuCell.addClass('ui-state-highlight');
     },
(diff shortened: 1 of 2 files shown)


melloware:
Yes that seems acceptable to me.  Can you check treetable.js to see if it needs the same fixes?

primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
```diff
@@ -3200,53 +3197,50 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
...
                             }
                         });
 
-            $(document).off('click.datatable-cell-blur' + this.id)
-                        .on('click.datatable-cell-blur' + this.id, function(e) {
-                            var target = $(e.target);
-                            if(!$this.incellClick && (target.is('.ui-input-overlay') || target.closest('.ui-input-overlay').length || target.closest('.ui-datepicker-buttonpane').length)) {
-                                $this.incellClick = true;
-                            }
+            // save/cancel on mouseup to que the event request before whatever was clicked reacts
```
melloware: `que` should be `queue`

melloware (changes requested):

VsevolodGolovanov:
@melloware, did the same for treeTable, tested a bit.

melloware (approved):

melloware:
I updated the ticket and the PR title to reflect its for both tables.

PR #7865, opened 2021-09-24
#7864 org.primefaces.selenium.component.SelectOneMenu#getLabels fails with filter

### LINKED ISSUE #7864: Selenium: org.primefaces.selenium.component.SelectOneMenu#getLabels fails with filter
@christophs78 i can see you changed the code for the filter case
i can commit a IT
would you like to check it?

christophs78:


tandraschko:
does this also fix test10? if yes, please enable it

PR #7886, opened 2021-09-28
fix IT for Java 17

christophs78:


tandraschko:
why removed 2.3-next?

christophs78:
Only removed as part of CI, is still part of IT. (To create less load on each commit.)
(If you want to have it back for CI - feel free!)

tandraschko:
i see, thats okay then :D

PR #7893, opened 2021-09-29
Tree004Test - improve stability

christophs78:


tandraschko:
cant we use ajaxguard?

christophs78:
> cant we use ajaxguard?

Hm. Thought there were some issues with "things" that run delayed. But local simple PrimeSelenium.guardAjax works. So let´s look whether we are green on Github Actions too.

PR #7899, opened 2021-09-30
Fix #7457: DatePicker integration tests

### LINKED ISSUE #7457: DatePicker: lazy load disabled dates when panel is opened

melloware:
cc @christophs78 cc @jepsar


melloware:
I will investigate what's failing

melloware:
@jepsar any idea why your new fix doesn't make @wal-jan Integration Tests pass?  I was assuming your fix accomplished the same as his PR?

melloware:
OK it looks like the problem is when "closing" the datepicker panel by clicking outside of the date picker it looks like @wal-jan was expecting that to fire the `viewChange` event which would clear the messages.  But that is not happening:

```xml
<p:datePicker id="datepicker2" pattern="yyyy-MM-dd" value="#{datePicker011.date2}"
                              model="#{datePicker011.model}" monthNavigator="true" yearNavigator="true" showButtonBar="true">
                    <p:ajax event="viewChange" listener="#{datePicker011.onViewChange}"/>
                </p:datePicker>
```

Part that is failing:

```java
        // Act - 3rd click outside panel
        if (DatePickerBehaviour.close == behaviour) {
            page.outside.click();
        }
        else {
            PrimeSelenium.guardAjax(page.outsideClean).click();
        }

        // Assert
        PrimeSelenium.waitGui().until(PrimeExpectedConditions.visibleAndAnimationComplete(datePicker));
        switch (behaviour) {
            case close:
                assertMessage(messages, behaviour);
                break;
            default:
                assertEmptyMessages(messages);
                break;
        }
```

When clicking outside of the datepicker he was expecting it to fire an AJAX even which would clear the Messages but it does so so the `assertEmptyMessages` is failing because its not empty.

jepsar:
Ha, I was just having a look. Yes, he introduced a flag to force an update of the model. What I did is not initially load data in the render and trigger a view change event on open and no long on close.

melloware:
So you think I should remove that check since no AJAX event is fired on close?

jepsar:
Yeah, with a lazy model we don't fire a change on close (changed in #7457). Without we do (not touched in #7457).

melloware:
@jepsar OK I see what you mean. I removed that check for Lazy and just always fire the ViewChange event.

primefaces-integration-tests/src/main/webapp/datepicker/datePicker012.xhtml
```diff
@@ -0,0 +1,32 @@
...
+<f:view contentType="text/html;charset=UTF-8" encoding="UTF-8">
+    <h:head>
+
+    </h:head>
+
+    <h:body>
+
+        <h:form id="form">
+            <p:datePicker id="datepicker1" value="#{datePicker012.date1}"
+                          model="#{datePicker012.lazyModel1}" refreshLazyModel="true"
```
jepsar, line 16: refreshLazyModel="true"? There is no such attribute.

melloware: Agreed there is no property for that...

PR #7912, opened 2021-10-03
cleanup test-classpath to avoid interference with embedded Tomcat

christophs78:


tandraschko:
jetty runs still fine?

christophs78:
Yep. Jetty still fine.
But i don´t get rid of the warnings regarding double initialization. (I hoped to improve this by doing this, but there seems to be another issue.)

tandraschko:
should we close or merge?

christophs78:
IMO merge

PR #7921, opened 2021-10-05
Fix #7871: Remove primefaces-selenium-runtime

### LINKED ISSUE #7871: Selenium: rework CSP support to remove runtime-module
the runtime module is required in the application itself, that makes remote testing unusable with CSP

the runtime module consists of 2 features:
- enable CSP testing support
- disable animations

we can also download lazy the disableanimations.css in selenium core
not sure about CSP

melloware:
Checking to see if everything passes.  I am worried about false positives since this is just checking `jQuery.events` for things like `click` but that doesn't mean it has to be AJAX I don't think.


melloware:
@tandraschko this is what I was afraid of... in AutoComplete001Test for example...

```java
/**
     * Sends the Tab-Key to jump to the next input. Attention: Pressing tab selects the first suggested value.
     */
    public void sendTabKey() {
        if (isOnchangeAjaxified()) {
            PrimeSelenium.guardAjax(getInput()).sendKeys(Keys.TAB);
        }
        else {
            getInput().sendKeys(Keys.TAB);
        }
    }
```

The `onchange` is being reported as AJAXified when it is not because AutoCompelte has a `onchange` javascript event that is not an AJAX event.   

I think the ONLY and bulletproof way to fix this is to add those csp.js methods to our real CSP js to actually keep that map when calling `PrimeFaces.register`. The map will take up very little memory and will be useful.

Thoughts?

melloware:
This is the script I am running but it can only tell me if Jquery has a `change` event registered but it might be just a client side `onchange="alert();"` i don't know whether its an AJAX event

```java
 private static final String CSP_SCRIPT = "var jq=jQuery(PrimeFaces.escapeClientId('%s'));" +
                "var events=jq?jQuery._data(jq[0],'events'):null;" +
                "return events?events.%s!==undefined:false;";
```

tandraschko:
cant we do a toString?

melloware:
> cant we do a toString?

Let me see if I can get the actual event.

melloware:
OK I can get the raw function JS code and I modified the script to this...

```javascript
var isAjaxified = false;
if (!PrimeFaces.csp.NONCE_VALUE) return false;
var jq = $(PrimeFaces.escapeClientId('%s'));
var events = jq ? $._data(jq[0], 'events') : null;
var hasEvents = events ? events.%s !== undefined : false;
if (!hasEvents) return isAjaxified;
$.each(events.%s, function(j, h) {
    var script = h.handler.toString();
    isAjaxified = isAjaxified || (script.indexOf("PrimeFaces.ab(") >= 0) ||
        (script.indexOf("pf.ab(") >= 0) ||
        (script.indexOf("mojarra.ab(") >= 0) ||
        (script.indexOf("jsf.ajax.request") >= 0);
});
return isAjaxified;
```

tandraschko:
i suggest to enhance our HeadRenderer to add:
PrimeFaces.settings.csp=true;

and only use the $__data when this is activated in #isAjaxified

melloware:
I can just short circuit the script with this line `if (!PrimeFaces.csp.NONCE_VALUE) return false`...  I do that in othre places in our code if the NONCE is not set you know we are not using CSP.

```javascript
if (!PrimeFaces.csp.NONCE_VALUE) return false;
var isAjaxified = false;
var jq = $(PrimeFaces.escapeClientId('%s'));
var events = jq ? $._data(jq[0], 'events') : null;
var hasEvents = events ? events.%s !== undefined : false;
if (!hasEvents) return isAjaxified;
$.each(events.%s, function(j, h) {
    var script = h.handler.toString();
    isAjaxified = isAjaxified || (script.indexOf("PrimeFaces.ab(") >= 0) ||
        (script.indexOf("pf.ab(") >= 0) ||
        (script.indexOf("mojarra.ab(") >= 0) ||
        (script.indexOf("jsf.ajax.request") >= 0);
});
return isAjaxified;
```

tandraschko:
thats ok, too!

melloware:
Looks like GitHub Actions is down right now so I will wait to see if this build and passes!

melloware:
OK next issue.  Because for some CommandButtons we wrapper the script with...

```
                // if the eventhandler return false, we must use preventDefault
                var jsWrapper = function(event) {
                    var retVal = js.call(element, event);
                    if (retVal === false && event.cancelable) {
                        event.preventDefault();
                    }
                };
```

So my code doesn't catch this as AJAX so I added these two lines because the JS could be compressed or uncompressed but this is very brittle because the `h` is not gauranteed to be the compressed variable name.

```java
+ " (script.indexOf('var retVal = js.call(element, event);') >= 0);" // CommandButton special handling
+ "(script.indexOf('h.cancelable){h.preventDefault()') >= 0);" // CommandButton special handling
```

christophs78:
Please keep in mind we don´t use CSP as part of CI. Maybe you temporary wan´t to modify build.yml.

melloware:
It looks like we do have CSP ON as part of CI?  I see the `csp`profile on.

christophs78:
Sorry, you are right. Maybe we enabled it during doing the multimodule-thing or so.

melloware:
I am really starting to think just adding the MAP to the core.csp.js is the cleanest and least brittle way to go.

christophs78:
> I am really starting to think just adding the MAP to the core.csp.js is the cleanest and least brittle way to go.

And use the MAP only when we set some JS(?)-switch from PF Selenium? (If there is a clean way to do this.)

melloware:
Well the Map shouldn't take up that much memory that we should worry about it right?

```javascript
                if (PrimeFaces.csp.NONCE_VALUE) {
                    if (!PrimeFaces.csp.REGISTERED_AJAXIFIED_EVENT_LISTENERS.has(id)) {
                        PrimeFaces.csp.REGISTERED_AJAXIFIED_EVENT_LISTENERS.set(id, new Map())
                    }
                    var script = js.toString();
                    var isAjaxified = (script.indexOf("PrimeFaces.ab(") >= 0) || 
                                      (script.indexOf("pf.ab(") >= 0) || 
                                      (script.indexOf("mojarra.ab(") >= 0) || 
                                      (script.indexOf("jsf.ajax.request") >= 0);
                    PrimeFaces.csp.REGISTERED_AJAXIFIED_EVENT_LISTENERS.get(id).set(jqEvent, isAjaxified);
                }
```

tandraschko:
+1 if only when != production

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.csp.js
```diff
@@ -63,9 +69,37 @@ if (!PrimeFaces.csp) {
                 };
 
                 $(element).off(jqEvent).on(jqEvent, jsWrapper);
+                
+                //Collect some basic information about registered AJAXified event listeners
+                if (!PrimeFaces.isProductionProjectStage()) {
+                    if (!PrimeFaces.csp.REGISTERED_AJAXIFIED_EVENT_LISTENERS.has(id)) {
+                        PrimeFaces.csp.REGISTERED_AJAXIFIED_EVENT_LISTENERS.set(id, new Map())
```
tandraschko: ; at end

melloware: Fixed letting it rebuild.

melloware:
Interesting. Getting some random test failures I was not expecting.

```
Failures: 
Error:    Tree004Test.testFilter:165 expected: <false> but was: <true>
Error:  Errors: 
Error:    Timeline001Test.testRangeChangedMoveLeft:114->assertRangeChanged:132 » StaleElementReference
Error:    Tree004Test.testFilterWithMultipleSelection:208 » ElementNotVisible element no...
```

I am having trouble producing them locally.

melloware:
I also just wrapped every `Arrays.asList` with `new ArrayList` because I was seeing Instantiation StackTraces on Datatable tests.

tandraschko:
@melloware i fixed already some tests like Tree004
so please dont mix your refactoring and test alignments in 1 PR

melloware:
Got it. I am going to close this PR and grab your latest changes.

melloware:
Did you fix wrapping th

PR #7930, opened 2021-10-06
Fix #7928: Respect javax.faces.SEPARATOR_CHAR

### LINKED ISSUE #7928: Column: getColumnKey() does not respect javax.faces.SEPARATOR_CHAR
**Describe the defect**
I have set javax.faces.SEPARATOR_CHAR to something different than the colon ":". I notice that the column filter does not work because Column.getColumnKey() doesn't work correctly, because getColumnKey() has the colon hardcoded:

```
public String getColumnKey(UIComponent parent, String rowIndex) {
     return getColumnKey().replace(parent.getId() + ":" + rowIndex + ":", parent.getId() + ":");
}
```

**Environment:**
 - PF Version: 11.0.0-RC1
 - Affected browsers: ALL

**To Reproduce**
1. Set javax.faces.SEPARATOR_CHAR to something different than the colon ":".
2. Try column filtering, it won't work

**Expected behavior**
Column filtering works.

melloware:
@tandraschko there is one place I am not sure about and that is ComponentUtils.escapeSelector().

```java
    public static String escapeSelector(String selector) {
        return selector.replace(":", "\\\\:");
    }
```

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/api/DynamicColumn.java
+++ b/primefaces/src/main/java/org/primefaces/component/api/DynamicColumn.java
@@ -221,7 +221,7 @@ public class DynamicColumn implements UIColumn {
 
     @Override
     public String getColumnKey(UIComponent parent, String rowIndex) {
-        return getColumnKey().replace(parent.getId() + ":" + rowIndex + ":", parent.getId() + ":");
+        return getColumnKey().replace(parent.getId() + separatorChar + rowIndex + separatorChar, parent.getId() + separatorChar);
     }
 
     @Override
--- a/primefaces/src/main/java/org/primefaces/component/column/Column.java
+++ b/primefaces/src/main/java/org/primefaces/component/column/Column.java
@@ -27,6 +27,7 @@ import java.io.IOException;
 import java.util.List;
 
 import javax.faces.component.UIComponent;
+import javax.faces.component.UINamingContainer;
 import javax.faces.context.FacesContext;
 
 import org.primefaces.component.celleditor.CellEditor;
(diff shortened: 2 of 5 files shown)


tandraschko:
findComponent really works with separatorChar and not just always :?

melloware:
Good question I assumed findComponent was looking for things like `:frmMain:txtUsername` right?

melloware:
Confirmed FindCompoennt uses sepChar

```java
public UIComponent findComponent(String expression) {
        if (expression == null) {
            throw new NullPointerException();
        }
        
        if (expression.isEmpty()) {
            // If an empty value is provided, fail fast.
            throw new IllegalArgumentException("\"\"");
        }

        final char sepChar = UINamingContainer.getSeparatorChar(FacesContext.getCurrentInstance());

        // Identify the base component from which we will perform our search
        UIComponent base = findBaseComponent(expression, sepChar);
        
        if (expression.charAt(0) == sepChar) {
            // Treat remainder of the expression as relative
            expression = expression.substring(1);
        }

        // Evaluate the search expression (now guaranteed to be relative)
        return evaluateSearchExpression(base, expression, String.valueOf(sepChar));
    }
```

tandraschko:
yes
but if you switch the seperator char to e.g. \, does this work:
`findComponent('\frmMain\txtUsername')` ?

melloware:
It looks like it always uses the `separatorChar`.

melloware:
Also according to a few things I read it said "you shouldn't need to change this but make sure its an HTML safe character like - or _".  Not sure changing it to slash `\` would be a good idea either way!

PR #7933, opened 2021-10-06
Fix #7932: CSP only call csp.init() if PrimeFaces available

### LINKED ISSUE #7932: CSP: Only initialize if PrimeFaces is available
Discovered during a failure of `Rating001Test.testInvalidNumberServerSide:253 » Timeout Timeout while waiting`

When navigating to `error.xhtml` while no PrimeFaces components are present the CSP script was generating this...

`PrimeFaces.csp.init('NGM3ZTNhOWMtYmM4YS00OWU1LTgzODQtZDk0ZDg2NmE2ZmVm');` and getting PrimeFaces not found!

When it should be this...

`if(pf){pf.csp.init('NGM3ZTNhOWMtYmM4YS00OWU1LTgzODQtZDk0ZDg2NmE2ZmVm');};`

melloware:
@tandraschko I fixed the root of the problem of CSP throwing an error on error.xhtml but also fixed the AJAX DebugInfo and put it back.  I tested it with this scenario where it was failing before and it now works.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/csp/CspPhaseListener.java
+++ b/primefaces/src/main/java/org/primefaces/csp/CspPhaseListener.java
@@ -23,17 +23,17 @@
  */
 package org.primefaces.csp;
 
-import org.owasp.encoder.Encode;
-import org.primefaces.context.PrimeApplicationContext;
-import org.primefaces.context.PrimeFacesContext;
-import org.primefaces.util.LangUtils;
-
 import javax.faces.context.ExternalContext;
 import javax.faces.context.FacesContext;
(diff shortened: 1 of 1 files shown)


christophs78:
On my dev-notebook the same issue like on Github Actions.
image

To me it looks like we have kind of a timing issue (with PrimeSelenium.guardAjax?) when the server decides to switch to errorpage.

melloware:
Yep the timeout is still there so I remove the guard change.  This only does the CSP fix if you look on the console of error.xhtml when its display it fails because there is no Primefaces.

PR #7934, opened 2021-10-06
Ratings IT Fix

melloware:
Just a draft to see if we can get all tests passing


tandraschko:
i will revert my guard change now, very complicated

PR #7946, opened 2021-10-09
Fix #7945: CascadeSelect align to item selected

### LINKED ISSUE #7945: CascadeSelect: subPanel should be aligned at the active item of parent panel instead of top of parent panel
While showing subpanel of CascadeSelect, this subpanel is aligned at the top of parentPanel. It should be aligned at the top of active item, which this subPanel belongs to.

Check the behaviour on http://www.primefaces.org:8080/showcase/ui/input/cascadeSelect.xhtml, if you choose third item of main panel, subpanel doesn't show "at" this item, but on the top. Imagine if there is more items in parentPanel..

SubPanel should be aligned at the top of activeItem. If you agree, look here:
forms.cascadeselect.js, method alignSubPanel, of position could be changed somehow like this > of: parentPanel.children('.ui-cascadeselect-item-active:first')

` 
subpanel.css({left:'0px', top:'0px'}).position({
                my: 'left top'
                ,at: 'right top'
                ,of: parentPanel.children('.ui-cascadeselect-item-active:first')
                ,collision: 'flipfit'
            });
`

thx

melloware:


mertsincan:
Nice! Thanks @melloware ;)

Issue #7950, opened 2021-10-12
DataTable: Payara IllegalArgumentException when using Array instead of List as selection value


ltlBeBoy:
**Describe the defect**
Multiple selection fails with an `IllegalArgumentException` using arrays as selection value for `DataTable`:

```
java.lang.IllegalArgumentException: Cannot convert [Ljava.lang.Object;@1a63f677 of type class [Ljava.lang.Object; to class [L<myClass>;
	at com.sun.el.lang.ELSupport.coerceToType(ELSupport.java:383)
	at com.sun.el.lang.ELSupport.coerceToType(ELSupport.java:339)
	at com.sun.el.parser.AstValue.setValue(AstValue.java:221)
	at com.sun.el.ValueExpressionImpl.setValue(ValueExpressionImpl.java:249)
	at org.jboss.weld.module.web.el.WeldValueExpression.setValue(WeldValueExpression.java:64)
	at org.jboss.weld.module.web.el.WeldValueExpression.setValue(WeldValueExpression.java:64)
	at com.sun.faces.facelets.el.TagValueExpression.setValue(TagValueExpression.java:95)
	at org.primefaces.component.datatable.feature.SelectionFeature.setSelection(SelectionFeature.java:248)
	at org.primefaces.component.datatable.feature.SelectionFeature.decodeMultipleSelection(SelectionFeature.java:153)
	at org.primefaces.component.datatable.feature.SelectionFeature.decodeSelection(SelectionFeature.java:89)
	at org.primefaces.component.datatable.feature.SelectionFeature.decode(SelectionFeature.java:69)
	at org.primefaces.component.datatable.DataTableRenderer.decode(DataTableRenderer.java:74)
	...
```

**Reproducer**
Simply replace `List<Product> selectedProducts` by `Product[] selectedProducts` in the showcase.

**Environment:**
 - PF Version: _11.0.0-RC1_
 - JSF + version: _Mojarra 2.3.14.payara-p2_
 - Affected browsers: _ALL_

**To Reproduce**
1. Go to DataTable view with multiple selection enabled that uses an array internally
2. Select a row
3. See error

**Expected behavior**
No exception should be thrown

**Solution**
Undo the changes in file [org.primefaces.component.datatable.feature.SelectionFeature](https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/java/org/primefaces/component/datatable/feature/SelectionFeature.java) at method `setSelection(...)` that have been made since version 10.0:

[...]

PR #7960, opened 2021-10-13
#7950 Integrationtests

christophs78:


melloware:
this looks good to me.  Let's hear back from OP on what they may be doing differently than this test.

PR #7961, opened 2021-10-13
Fix #7954: Datatable with virtual scrolling needs getRowCount()

### LINKED ISSUE #7954: DataTable: Table is empty when virtualScroll="true"
**Describe the defect**
`DataTable` is empty when attribute `virtualScroll` is set to `true`.

**Reproducer**
See _[Virtual Scrolling - 20000 Rows](https://www.primefaces.org/showcase/ui/data/datatable/scroll.xhtml)_ in showcase.

**Environment:**
 - PF Version: _11.0.0-RC1_
 - JSF + version: _Mojarra 2.3.14.payara-p2_
 - Affected browsers: _ALL_

**To Reproduce**
See _[Virtual Scrolling - 20000 Rows](https://www.primefaces.org/showcase/ui/data/datatable/scroll.xhtml)_ in showcase.

melloware:


christophs78:
Think loadLazyScrollData should call lazyModel.setRowCount(lazyModel.count(filterBy)); like loadLazyData.
(And maybe some more.)
Hm. Not totally sure. Data may have changed in the database behind. (Rows added, updated or removed.) Hard (impossible?) to deliver consistent results.

melloware:
I think for virtual scroll it always has to be the total rows not the filtered rows because it needs to know how to ask for more rows

tandraschko:
Sure?
I think it should depend on current Filters, but Not sure

tandraschko:
also note that getRowCount IS the the filtered rowCount for the normal case as we do
```
lazyModel.setRowCount(lazyModel.count(filterBy)); 
```

melloware:
OK testing a better fix. You are right.

melloware:
You guys were right the lazyScroll was just missing the call to getCount().  Fixed and verified Showcase is working properly.

PR #7968, opened 2021-10-14
Fix #7951 - DataTable: Sort is lost after data refresh

### LINKED ISSUE #7951: DataTable: Sort is lost after data refresh
**Describe the defect**
When using a DataTable with sortable columns, the sort order is lost after refreshing/replacing the underlying data model (List) in an AJAX request. The header still indicates the sorting in the column, but the data is no longer ordnered.
In PF 8.0 this still worked, but since 10.0 the sorting is no longer kept after the list is recreated.

Sometimes it is necessary to e.g. periodically update the displayed data from DB, but still keep the currently selected sorting.

**Environment:**
 - PF Version: 10.0.7 (also tested the current master 11.0.0-SNAPSHOT with commit 9ea38543)
 - JSF + version: Mojarra 2.3.9
 - Affected browsers: Firefox

**To Reproduce**
1. Display Datatable in a view
2. Sort by clicking a sortable header.
3. The data is correctly sorted by this column. THe column header indicates sorting.
4. Trigger reloading the data for this datatable by AJAX (and alos the update of the datatable), e.g. via a CommandButton
5. Now the table header still indicates that the data is sorted by this column. But the updated data is no longer sorted

**Expected behavior**
Even after a reloading of the data, I would expect the data in the table to be sorted by the column. This was also the behaviour in PrimeFaces 8.0.

**Example XHTML**
```html
<h:form>
    <p:dataTable value="#{dataBean.data}" var="data">
        <p:column headerText="Text" sortBy="#{data.text}">
            <h:outputText value="#{data.text}"/>
        </p:column>
        <p:column headerText="Number" sortBy="#{data.num}">
            <h:outputText value="#{data.num}"/>
        </p:column>
    </p:dataTable>
    <h:panelGroup>
        <p:commandButton value="Refresh" process="@this" update="@form" actionListener="#{dataBean.refresh}" />
    </h:panelGroup>
</h:form>
```

**Example Bean**
```java
@ViewScoped
@Named("dataBean")
public class DataBean implements Serializable {

    public static class Data {

        private final String text;

        private final int num;

        public Data(String text, int num) {
            this.text = text;
            this.num = num;
        }

        public String getText() {
            return text;
        }

        public int getNum() {
            return num;
        }
    }

    private List<Data> data;

    public List<Data> getData() {
        return data;
    }

    @PostConstruct
    public void refresh() {
        List<Data> newData = new ArrayList<>();
        for (int i = 0; i < 10; i++) {
            // Generate random letter and number to test sorting
            newData.add(new Data("Data " + ((char) ((int) 'A' + ThreadLocalRandom.current().nextInt(26))),
                    ThreadLocalRandom.current().nextInt(100)));
        }
        this.data = newData;
    }

}
```

Rapster:


christophs78:
We should also try to add an integrationtest for this issue. :-)

tandraschko:
It also breakes an existing test

Rapster:
Not sure I understand the existing test:
- 1,3,5 are selected (not sure it really matters)
- Delete Java row

When I have a look at DataTable010#deleteRow, it only remove the "Java" row and yet we expect Java and Javascript to be displayed? i'm confused

christophs78:
Feedback comes this evening. Maybe this weird test was implemented this way to compensate for a PF bug. And after you fixed the root cause we may need to adapt this test.

christophs78:
Up to your fix the (not ideal) behavior after deleting the Java-row was following:
image

This is reflected by the test and maybe the archived google-code-link.

--> I´ll checkout your PR and add a commit which adapts this test to the now correct behaviour.

primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datatable/DataTable010Test.java
```diff
@@ -132,18 +132,15 @@ public void testSelectionMultipleWithPagingAndFilteringAndDelete(Page page) {
...
+        Assertions.assertEquals(1, page.dataTable.getRows().stream().count());
+        Assertions.assertEquals("JavaScript", page.dataTable.getCell(0, 1).getText());
 
         // Act
         page.buttonSubmit.click();
 
         // Assert
         Assertions.assertTrue(page.messages.getMessage(0).getSummary().contains("Selected ProgrammingLanguage(s)"));
-        Assertions.assertEquals("3,5", page.messages.getMessage(0).getDetail());
+        Assertions.assertEquals("3,5", page.messages.getMessage(0).getDetail()); // TODO: this is still kind of weird because we deleted Java (ID 3) before
```
christophs78, line 143: Maybe you can have a look at this. (selection) Maybe i miss something or this aspect is still a bit weird.

tandraschko:
any news?
Your test PR passed, whats left now?

christophs78:
> any news? Your test PR passed, whats left now?

IMO it's with this PR better than before. (There is as often still place for future work, but that's another thing.)

PR #7981, opened 2021-10-17
Implement #7980: Dialog Framework: provide PhaseListener which allows using (JSF-)Flash with Dialog Framework

### LINKED ISSUE #7980: Dialog Framework: provide PhaseListener which allows using (JSF-)Flash with Dialog Framework
Flash can be used to pass data from one page to another.
On the way from a page to a dialog framework - page this is a bit more complicated. You need a PhaseListener which keep´s the Flash alive during opening the dialog framework - dialog. (eg DialogKeepFlashPhaseListener)
We already use such a PhaseListener inhouse and i´ll contribute it to PrimeFaces.

Known limitation:
Currently only works on Mojarra when combined with https://showcase.omnifaces.org/filters/GzipResponseFilter (there´s a bug within Mojarra which prevents it form working without OmniFaces GzipResponseFilter; i provided a PR https://github.com/eclipse-ee4j/mojarra/pull/4978 to fix this)

It work´s fine with MyFaces.

christophs78:


primefaces/src/main/java/org/primefaces/application/DialogKeepFlashPhaseListener.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public PhaseId getPhaseId() {
+        return PhaseId.RENDER_RESPONSE;
+    }
+
+    private boolean isInDialogPreparation() {
+        String dialogOutcome = (String) FacesContext.getCurrentInstance().getAttributes().get(Constants.DialogFramework.OUTCOME);
+        return dialogOutcome != null && dialogOutcome.length() > 0;
```
tandraschko: use our LangUtils

primefaces-showcase/src/main/java/org/primefaces/showcase/view/df/DFLevel1View.java
```diff
@@ -27,14 +27,30 @@
...
 @RequestScoped
 public class DFLevel1View {
 
+    private boolean flashChecked = false;
+
+    public void init() {
+        if (!flashChecked) {
+            LocalDateTime valueFromFlash = (LocalDateTime) FacesContext.getCurrentInstance().getExternalContext().getFlash().get("param1");
+            if (valueFromFlash != null) {
+                FacesContext.getCurrentInstance().addMessage("", new FacesMessage("value from flash",
```
melloware: Use `Constants.EMPTY_STRING` here

melloware (approved):

primefaces-showcase/src/main/java/org/primefaces/showcase/view/df/DFLevel1View.java
tandraschko: please dont use PF internal constants in showcase

melloware: Yep just realized that was a showcase class.

tandraschko:
can you also add a documentation in the DF part of the docs?

christophs78:
> can you also add a documentation in the DF part of the docs?

yes, comes this evening

PR #7992, opened 2021-10-19
Fix #7724: DataTable: components can't overflow in editable cells

### LINKED ISSUE #7724: DataTable: components can't overflow in editable cells
See https://github.com/primefaces-extensions/primefaces-extensions/issues/511

Screenshot 2021-08-20 at 11 58 36

This CSS rule sets overflow to hidden:

https://github.com/primefaces/primefaces/blob/65de61e60e6218b28ff147f947ed422e1ffd8916/primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.css#L29-L37

Shouldn't it be visible at least when a cell is editable?

https://github.com/primefaces/primefaces/blob/65de61e60e6218b28ff147f947ed422e1ffd8916/primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.css#L195-L201

jepsar:
Fix #7724

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.css
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.css
@@ -1,11 +1,7 @@
 .ui-datatable table {
-	border-collapse:collapse;
+    border-collapse: collapse;
     width: 100%;
     table-layout: fixed;
-}
-
-.ui-datatable-tablewrapper {
-    overflow: auto;
 }
 
 .ui-datatable .ui-datatable-header,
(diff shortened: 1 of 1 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.css
```diff
@@ -1,13 +1,9 @@
 .ui-datatable table {
-	border-collapse:collapse;
+    border-collapse: collapse;
     width: 100%;
     table-layout: fixed;
 }
 
-.ui-datatable-tablewrapper {
```
jepsar, line 7: This was needed to allow fields close to the top or bottom of the table to overflow

tandraschko:
We should Not apply such changes without Mert
It may breake cases again

melloware:
I assigned to Mert.

jepsar:
> We should Not apply such changes without Mert It may breake cases again

Yes, especially the change on the wrapper. I tested with the test repo and the showcase and there it is looking good. Happy to get feedback from Mert.

mertsincan:
My comment: https://github.com/primefaces/primefaces/issues/7724#issuecomment-957368921

PR #7998, opened 2021-10-21
Fix #7997: ColorPicker updated conversion functions

### LINKED ISSUE #7997: ColorPicker: Incorrect Translation of Inputs
**Describe the defect**

 I have a user complaining that some colors are not selectable and I confirmed even on the showcase : https://www.primefaces.org/showcase/ui/input/colorPicker.xhtml
If you try to input `RGB 0,48,108` or its hex value `00306C` you get `0,49,108` and `00316C`.
Not sure what happens here but I was expecting an input color to input my color without altering it ! 
after digging into colorPicker.js it looks like it always ends up setting values from HSB conversion, no mater where the user is updating value. 

```javascript
if (ev) {
                    fillRGBFields(col, cal.get(0));
                    fillHexFields(col, cal.get(0));
                    fillHSBFields(col, cal.get(0));
                }
```

where col is the HSB version of the selected color

melloware:
Updated to more accurate calculation of conversion functions.

See: https://stackoverflow.com/questions/17242144/javascript-convert-hsb-hsv-color-to-rgb-accurately

image

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/colorpicker/0-jquery.colorpicker.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/colorpicker/0-jquery.colorpicker.js
@@ -291,66 +291,68 @@
 				return hex;
 			},
 			HexToRGB = function (hex) {
-				var hex = parseInt(((hex.indexOf('#') > -1) ? hex.substring(1) : hex), 16);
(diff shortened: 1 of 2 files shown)


tandraschko:
we have a file somewhere where hacks are mentioned, please add it there

melloware:
Will do I also submitted my PR to this fork which hasn't been updated in 10 years but its for others to see: https://github.com/thorst/jQuery-ColorPicker/pull/2

melloware:
Updated the extensions.txt file with my change

PR #8008, opened 2021-10-22
Fix #8007: Tabview reload(index) widget method to reload cached panel

### LINKED ISSUE #8007: TabView: Widget method to reload dynamic panel
**Description**
Requested on Stack Overflow: https://stackoverflow.com/questions/69667691/can-a-primefaces-tab-be-manually-reset-to-force-a-reload

> I'm using a Primefaces tabView with the dynamic and cache attributes set to true - each tab is thus loaded the first time it's displayed and then cached so it does need to be reloaded. Is there a way to manually force a reload of an already cached tab so that the next time it is displayed it's content is fetched again from the server and not from the cache?

I have provided this MonkeyPatch fix which should be just first class in the TabView.

```javascript
if (PrimeFaces.widget.TabView) {
    PrimeFaces.widget.TabView.prototype.reloadTab = function(index) {
        var reloadPanel = this.panelContainer.children().eq(index);
        reloadPanel.data('loaded', false);
        this.select(index);
    }
}
```

melloware:


tandraschko:
+1 if well tested

PR #8009, opened 2021-10-22
Fix #8006: Autocomplete multiple=true forceSelection=false allow item

### LINKED ISSUE #8006: Autocomplete: forceselection="false" no longer working since primefaces release 10.0.1
**Describe the defect**
The autocomplete component with forceselection="false" doesn't add a new value (which is not on the suggestion list) after enter key is fired.
This worked with 10.0.0 but stopped working since 10.0.1.
Looks like this was the issue introduced in 10.0.1: https://github.com/primefaces/primefaces/issues/7211

**Reproducer**
I used the example code from the showcase:
https://www.primefaces.org/showcase/ui/input/autoComplete.xhtml

Only change I did was setting forceSelection="false".

**Environment:**
 - PF Version: 10.0.7

**To Reproduce**
Starting with release 10.0.1 we see a changed behaviour. When we enter a new value (which is not on the sugestion list) and hit 'enter' key -> nothing happens
image

**Expected behavior**
With release 10.0.0 it worked fine, we could add values from suggestion list and also new values could be entered (like 'aaa', 'bbb', ...):
image

This defect is related to https://forum.primefaces.org/viewtopic.php?f=3&t=69186

melloware:


tandraschko:
can we add a test here?
it feels like this part was touched multiple times and everytime something was destroyed

melloware:
yep let me see if I can add it.

melloware:
Integration test added.

PR #8010, opened 2021-10-22
Fix #8005: Exporter with draggableColumns=true

### LINKED ISSUE #8005: DataTable: All columns are missing in table exports when a "draggable=true" attribute is present
**Describe the defect**
When the data table attribute `draggable=true` is present, a different path through the `TableExporter.getExportableColumns()` method is taken, which is leading to the effect, that no exportable column will be found. 

This is due to the fact, that now columnMeta information is present in the dataTable which columnKey slightly differs from the columnKey of the column in the table.

 In that case meta.getColumnKey() aka columnKey yields something like "table:index:colName" whereas col.getColumnKey() yields "table:columnName".

This is the code in question:
```
            for (ColumnMeta meta : columnMetas) {
                String columnKey = meta.getColumnKey();
                for (UIColumn col : allColumns) {
                    if (col instanceof DynamicColumn) {
                        ((DynamicColumn) col).applyStatelessModel();
                    }
                    if (col.getColumnKey().equals(columnKey)) {
                        if (col.isRendered() && col.isExportable()) {
                            exportcolumns.add(col);
                        }
                        break;
                    }
                }
            }
```

There are several ways to fix this, probably stripping the column index before comparison is the correct one here.
I gladly provide a pull-request once this has been clarified.

**Reproducer**
Clone repository https://github.com/stolp/primefaces-test/tree/PF-8005 in order to reproduce the problem.

**Environment:**
 - PF Version: _11.0_
 - JSF + version: _Mojarra 2.2.20_
 - Affected browsers: _ALL_

**To Reproduce**
1. Open Reproducer
2. Click on any export button
3. Open export file
4. See that column information is missing

**Expected behavior**
Columns should be exported.

melloware:


primefaces/src/main/java/org/primefaces/component/export/TableExporter.java
```diff
@@ -98,14 +98,15 @@ else if (LangUtils.isNotBlank(column.getField())) {
         Map<String, ColumnMeta> columnMetadata = table.getColumnMeta();
 
         if (columnMetadata == null || columnMetadata.isEmpty()) {
-            for (UIColumn col : allColumns) {
+            table.forEachColumn(col -> {
                 if (col instanceof DynamicColumn) {
                     ((DynamicColumn) col).applyStatelessModel();
```
tandraschko: AFAIR this is now done directly into forEachColumn

melloware: You are right! removed

Issue #7999, opened 2021-10-21
DataTable:  filter/sort - wrong manipulation of list elements on Mojarra without filteredValue defined (non-lazy)


DazzlingBob:
### 1) Environment

PrimeFaces version: primefaces 10.0.0
Application server: jetty 9.4.44 + version: mojarra-2.3

### 2) Expected behavior

Run the sample project.

Follow these steps:
1. filter on name with value BB2
2. change all BB2 row values to BB3, press Save [result=OK]
3. remove filter BB2, press Save [result=OK]
4. sort on code, press Save [result=WRONG] data rows switch with primefaces 10!

The result must be:
509, EUR, BB, BB3, A
512, EUR, BB, BB3, B
515, EUR, BB, BB3, C
516, EUR, AA, AA, D
517, EUR, AA, AA, E

### 3) Actual behavior

The result is:
509, USA, AA, AA, A
512, USA, AA, AA, B
515, EUR, BB, BB3,  C
516, EUR, BB, BB3, D
517, EUR, BB, BB3, E

### 4) Steps to reproduce

See sample project.
This behavior is only in primefaces 10, primefaces 8 is fine.

As a remark i must mention this is only one example. Different combinations of sorting and filtering data
gives different results, even in combination with multiViewState. Also the behavior changes somehow in later primefaces versions. Still managed to get it wrong in PrimeFaces-10.0.7 elite (latest version).

### 5) Sample XHTML / Sample bean

See attached projects.

Primefaces 10:
[primefaces-test-datatable.zip](https://github.com/primefaces/primefaces/files/7389069/primefaces-test-datatable.zip)

Primefaces 8 (works as expected):
[primefaces-test-datatable-8.0.zip](https://github.com/primefaces/primefaces/files/7389071/primefaces-test-datatable-8.0.zip)

run with:
mvn clean jetty:run -Pmojarra23

tandraschko:
please try 11.0.0-RC1

DazzlingBob:
Just tried it with 11.0.0-RC1,  unfortunately same wrong result.

tandraschko:
@Rapster could you also check this? may be related to https://github.com/primefaces/primefaces/issues/7951

@melloware @christophs78 maybe you can also check it and provide a IT

i would really like to get rid of all those DT refactoring bugs, which made 10.0 so bad

melloware:
[...]

PR #8011, opened 2021-10-22
#7999 integrationtest

christophs78:


melloware:
Nice to see these results prove Mojarra fails and MyFaces works!

tandraschko:
I dont think we should merge this yet as it just contains a workaround?

melloware:
Oh I didn't consider `filtervalues` a workaround I have been using Mojarra for years and we ALWAYS have put `filterValues` on our datatables I thought it was required.  So this test is how all of my production code already is.

christophs78:
Can we log an error and/or add for jsf-dev-stage a faces-message when filter is used without defining filterValues?

melloware:
I personally like that idea.  @tandraschko Thoughts?

tandraschko:
I dont think that its an mandatory attr, so -1

Issue #7962, opened 2021-10-14
Schedule: Upgrade FullCalendar 5.10.0


melloware:
Release Notes: https://github.com/fullcalendar/fullcalendar/releases/tag/v5.10.0

Looks like it has lots of ARIA improvements.

cc @blutorange

PR #8013, opened 2021-10-23
#7962 Schedule: Upgrade FullCalendar 5.10.1

christophs78:


christophs78:
Look´s like there are some visual regressions. (Header)
image

melloware:
Can you updated thirdpartylibs.md in 12.0 docs I reverted your chnage here in the 11.0 docs

primefaces/src/main/resources/META-INF/resources/primefaces-saga/theme.css
```diff
@@ -1338,7 +1338,7 @@ body .fc th {
   background: #f8f9fa;
   border: 1px solid #dee2e6;
   color: #495057;
-  padding: 1rem;
+  /*padding: 1rem;*/
```
christophs78: We need @mertsincan to adjust all theme-css to work with FullCalendar >= 5.10.0

PR #8022, opened 2021-10-26
Test #7832: AutoComplete forceSelection=true autoSelection=false

### LINKED ISSUE #7832: AutoComplete: invalid behaviour for forceSelection=true and autoSelection=false
**Describe the defect**
With #3156 (PR #3162) the `autoSelection` property was implemented for `p:autoComplete`. The combination of `forceSelection=true` and `autoSelection=false` produces an invalid state. 

`forceSelection` ensures, that the autoComplete always submits valid values as returned by the `completeMethod`. If no valid item was selected (either manually or auto-selected), the autoComplete is cleared.

With disabled `autoSelection` the `forceSelection` algorithm still validates the entered label but the matching item is not selected resulting in an invalid state - the label instead of the item-value will be submitted to the server. This may cause an error in the backend due to e.g. a type missmatch (In my case: the converter expects an parseable number and not an string).

**Expected Behaviour**
In my opinion the autoComplete should be cleared or the documented logic of `forceSelection` is broken.

If `forceSelection` is `true` and you search a value but no suggested item is manually selected, always clear the state of the autoComplete when the input is blurred and `autoSelection` is `false`,

**Reproducer**
_I have to prepare a reproducer if required._

**Environment:**
 - PF Version: 10.0

melloware:
Trying to reproduce #7832

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/autocomplete/AutoComplete005.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/autocomplete/AutoComplete005.java
@@ -52,7 +52,11 @@ public class AutoComplete005 implements Serializable {
 
     private List<Driver> selectedDrivers;
 
+    private Driver selectedDriver;
+
     private boolean forceSelection = true;
+
+    private boolean autoSelection = true;
 
     @PostConstruct
     public void init() {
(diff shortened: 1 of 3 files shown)


tandraschko:
can we merge?

melloware:
Merged

PR #8025, opened 2021-10-27
Fix #8024: SelectOne javascript error on keypress

### LINKED ISSUE #8024: SelectOneMenu: ItemList is not initiliazed on focus by keyboard
**Describe the defect**
When a SelectOneMenu inputfield is focused by keyboard use it is not initializing it's itemList inside. So if the user want to choose the value by pressing the first char of the known item, it fails with an `Uncaught TypeError: $this.items is undefined`

**Reproducer**
Can be reproduced in official Primefaces Showcase.

**Environment:**
 - PF Version: _11.0.0-RC1_
 - Affected browsers: ALL_

**To Reproduce**
1. Go to 'http://www.primefaces.org:8080/showcase/ui/input/oneMenu.xhtml?jfwid=1ee74'
2. Click on 'inputfield of Editable'
3. Use 'SHIFT + TAB' to get the focus on the 'Grouping' inputfield
4. Press 'G' to get 'Germany'

**Expected behavior**
Germany is choosen and displayed in the inputfield

**Error message**
grafik

**Workaround**
If the mouse is used to focus the inputfield 'Grouping', it opens automatically the listed values and after that, the list is initialized and the error does not occure any longer.

melloware:
On focus we need to load the panel ONLY if its not an AJAX panel as that gets loaded dynamically when you click the drodpwn.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectonemenu.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectonemenu.js
@@ -288,9 +288,12 @@ PrimeFaces.widget.SelectOneMenu = PrimeFaces.widget.DeferredWidget.extend({
             e.preventDefault();
         });
 
-        this.focusInput.on('focus.ui-selectonemenu', function() {
+        this.focusInput.on('focus.ui-selectonemenu', function(e) {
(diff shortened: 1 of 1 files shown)


tandraschko:
+1 if well tested

Issue #8073, opened 2021-11-12
TreeTable: after filtering and changing value during ajax old filtered data is still shown


christophs78:
Same issue like https://github.com/primefaces/primefaces/issues/7845 for DataTable.
Worked with PF8, does not work with PF10.
I´ll provide a Integrationtest. (IT will show us whether the issue is still there with PF11. My personal feeling says we did not fix this yet for TreeTable.)

tandraschko:
i think we should fix it for 11.0
would be great if you can provide a PR

christophs78:
yep, should be fixed for 11.0.0

melloware:
@christophs78 @tandraschko it looks like this ticket is complete but its still open?

christophs78:
https://github.com/primefaces/primefaces/pull/8078#issuecomment-971826271

christophs78:
We added kind of a manual work-around but did not fix the root cause.

melloware:
Thanks I was having trouble figuring out where this stood!

tandraschko:
closing in favor of #12795

PR #8078, opened 2021-11-12
#8073 TreeTable: after filtering and changing value during ajax old filtered data is still shown 

christophs78:
Status:

- added integrationtest(s)
- my attempt to port https://github.com/primefaces/primefaces/commit/24c939e62d51a5ea0a341675b95f997cca1cc7e0?w=1 from DataTable to TreeTable failed
- so added TreeTable#filterAndSort to provide some kind of work-around

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/treetable/DocumentService.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/treetable/DocumentService.java
@@ -99,6 +99,16 @@ public class DocumentService {
         return root;
     }
 
+    public TreeNode createOtherDocuments() {
+        TreeNode root = new DefaultTreeNode(new Document("Projects", "-", "Folder"), null);
+
+        TreeNode applications = new DefaultTreeNode(new Document("Project A", "100kb", "File"), root);
+        TreeNode cloud = new DefaultTreeNode(new Document("Project B", "20kb", "File"), root);
+        TreeNode desktop = new DefaultTreeNode(new Document("Project C", "150kb", "File"), root);
+
+        return root;
+    }
+
     public TreeNode createCheckboxDocuments() {
         TreeNode root = new CheckboxTreeNode(new Document("Files", "-", "Folder"), null);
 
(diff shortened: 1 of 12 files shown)


christophs78:
Some late-night-notes...

I tried to add code to `TreeTableRenderer#preRender` similar to https://github.com/primefaces/primefaces/blob/c856470c09efbe6e84cb35554b453b7f27fa8e9d/primefaces/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java#L140-L147 but this breaks some existing filtering-integration-test. (And i tried and analysed a few more things without success.)

christophs78:
Adding code to `TreeTableRenderer#preRender`  works in combination with keeping references to original TreeNodes when cloning. Also we propagate expanded- and selected-state to the original TreeNodes.

tandraschko:
hmpf, looks like changes are getting to big, its not a simple fix anymore
also not sure about originalTreeNode in DefaultTreeNode

christophs78:
> hmpf, looks like changes are getting to big, its not a simple fix anymore
> also not sure about originalTreeNode in DefaultTreeNode

First thought about something like ClonedTreeNode. But not that easy because we have multiple implementations of TreeNode.

melloware:
Speaking of that there is this stack overflow recent issue around DefaultTreeNode

https://stackoverflow.com/questions/69938930/primefaces-defaulttreenode-seriaalization-exception-with-kryo

Not sure if it's related

christophs78:
DefaultTreeNode-Serialization-issue is not related to this.
We even don´t know what this guy puts into data-property of his DefaultTreeNode-instances.

https://github.com/primefaces/primefaces/issues/8073 already consumes way to much time and energy.

TreeTable has some special behaviour during filtering. It tries to expand nodes containing filtered nodes. This aspect makes it different (and more complicated) compared to DataTable. So indirect we loose existing expand-state during filtering. For some usecases this is propably a good thing for others i´m not that sure. (But it´s nothing we should change without of thinking about AND writing way more IntegrationTest for TreeTable than we have now.)

Maybe we should just add `TreeTable#filterAndSort` and rollback (almost) all other things.

tandraschko:
Yeah
I think the node cloneing isnt nice and we should just set visibility of the nodes on filter
But thats Not a small task

tandraschko:
see: https://github.com/primefaces/primefaces/issues/6173

christophs78:
Cleaned up this PR. We just add TreeTable#filterAndSort and keep everything else as-is for now.

PR #8081, opened 2021-11-12
Fix #8064: TextEditor respect PF themes

### LINKED ISSUE #8064: Theme: TextEditor buttons look disabled in dark themes Vela and Arya
Screenshot was taken from showcase running on 11.0.0-RC2

Screenshot 2021-11-10 at 08 56 50

melloware:
So no one really makes a QuillJS Dark Theme.  So the better solution was to just modify the SNOW theme with PrimeTek variables like `var(--text-color)`.  It looks great in Arya, Saga, Vela and every other theme I have tried.  So this was the simplest and most effective solution.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/texteditor/1-texteditor-theme.css
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/texteditor/1-texteditor-theme.css
@@ -1,400 +1,3 @@
-/*!
- * Quill Editor v1.3.7
- * https://quilljs.com/
- * Copyright (c) 2014, Jason Chen
- * Copyright (c) 2013, salesforce.com
- */
-.ql-container {
-  box-sizing: border-box;
-  font-family: Helvetica, Arial, sans-serif;
-  font-size: 13px;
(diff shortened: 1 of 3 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/texteditor/1-texteditor-theme.css
```diff
@@ -840,7 +443,7 @@
   font-size: 32px;
 }
 .ql-snow .ql-color-picker.ql-background .ql-picker-item {
-  background-color: #fff;
+  background-color: var(--surface-a);
```
jepsar: Should this be changed? Seems like this is on a fixed black background based on the rule below.

melloware: Good catch. I reverted that one.

tandraschko:
let @mertsincan review and decide if we should merge it
i would do it in 12.0

melloware:
Agreed

PR #8090, opened 2021-11-18
Fix #8076: Schedule: onDateSelect chooses wrong time

### LINKED ISSUE #7891: Schedule: FullCalendar is already timezone-aware since v5, PrimeFaces.toISOString should not be used
**Describe the defect**
When using `<p:schedule>` with a fixed client time zone as in `<p:schedule clientTimeZone="UTC">` then the lazy load and date select events still send wrong local dates back into the `LazyScheduleModel#loadEvents()` and `SelectEvent<LocalDateTime>` respectively.

**Reproducer**
1. Open https://www.primefaces.org/showcase/ui/data/schedule/configuration.xhtml
2. Adjust `clientTimeZone` option from `local` to a different timezone than your own
3. Click any time slot and in the modal you'll notice that the time is off from what's being selected

**Environment:**
 - PF Version: _10.0_
 - JSF + version: _ALL_
 - Affected browsers: _ALL_

**Cause**
https://github.com/primefaces/primefaces/blob/10.0.0/src/main/resources/META-INF/resources/primefaces/schedule/1-schedule.js is still using `PrimeFaces#toISOString()` in three places:
- https://github.com/primefaces/primefaces/blob/10.0.0/src/main/resources/META-INF/resources/primefaces/schedule/1-schedule.js#L135
- https://github.com/primefaces/primefaces/blob/10.0.0/src/main/resources/META-INF/resources/primefaces/schedule/1-schedule.js#L283
- https://github.com/primefaces/primefaces/blob/10.0.0/src/main/resources/META-INF/resources/primefaces/schedule/1-schedule.js#L284

**Solution**
Use `info.dateStr` instead of `PrimeFaces.toISOString(info.date)`. FullCalendar has already converted them to the correct ISO string based on the configured `clientTimeZone`. There is no need to have `PrimeFaces#toISOString` to convert them.
```
L135: var currentDate = dateClickInfo.dateStr;
L283: {name: $this.id + '_start', value: fetchInfo.startStr},
L284: {name: $this.id + '_end', value: fetchInfo.endStr}
```

### LINKED ISSUE #8076: Schedule: onDateSelect chooses wrong time
The selected time is always one hour less. You can easily verify this using the showcase:

https://www.primefaces.org/showcase/ui/data/schedule/basic.xhtml?jfwid=381ce

Just click on any slot. For instance, clicking on 6am will open the event details panel with starttime 5am.

Many thanks in advance for fixing this bug.

christophs78:


christophs78:
@tuerker: maybe you want to have a look at this

melloware:
Looks good to me!

tuerker:
> @tuerker: maybe you want to have a look at this

looks good. tomorrow, when i am back in the office, i will double check and let you know if there is still something.

would be good to replace the PrimeFaces.toISOString calls also in the timeline.js file. thanks and regards.

tuerker:
yes. that is the fix. thanks.

christophs78:
@balusc: Can you please have a look at this fix? Does it make sense to you? I would appreciate your feedback.

(When we look back it would have been better to switch to ZonedDateTime instead of LocalDateTime during our Java8-migration to make things more explicit. Maybe we should think about this again for a future PrimeFaces-version.)

tuerker:
would be great if this is merged into the master asap. thanks.

christophs78:
> would be great if this is merged into the master asap. thanks.

We will do this one for 12.0.0 after adding additional integrationtests.
This is too complicated for doing a quick fix and maybe causing new trouble otherwhere.

With this PR we create trouble for showcase in scenarios where the client has another timezone than the server. (Because we convert on the server into LocalDateTime and show LocalDateTime in editing-dialog. DatePicker (and Calendar) do not take client-timezone into account, ...
This are things we should be able to fix (or work-around) but it´s nothing we should do in the finish before a major-release.)

tuerker:
many thanks for the info.

christophs78:
@tuerker After i invested a bunch of sparetime for free it´s your turn. Please checkout this PR and enhance Schedule001Test with test´s with different settings for timeZone and clientTimeZone.

melloware:
@tuerker any progress on the request from @christophs78 ???

tuerker:
> @tuerker any progress on the request from @christophs78 ???

Sorry, I can only have a deeper look into it in the second week of January. The original fix was definitively working.

christophs78:
@tuerker : Which timezone-combination do you have? Are your client´s within the same timezone as your server?
(If your client´s are within a other timezone i´ll have to ask more questions.)

christophs78:
@BalusC: Which timezone-combination do you have in your usecases using PrimeFaces-Schedule? Are your client´s within the same timezone as your server?
(If your client´s are within a other timezone i´ll have to ask more questions because it get´s complicated when we allow editing events using e.g. PrimeFaces-DatePicker or PrimeFaces-Calendar which are not aware of client-timezone OOTB.)

erickdeoliveiraleal:
** @BalusC

melloware (approved):

melloware:
@tandraschko I would like to merge this.  Has integration tests and looks good to me.  Let me know your thoughts?

tandraschko:
Never used schedule or developed on it
So if your tests are fine, do it 🙏

melloware:
Done!

Issue #8062, opened 2021-11-09
LazyDataModel: split in count() and load() clashes with backend methods that do both


giemes:
**Description**
Prior to PF11, the LazyDataModel had a central load() method passing all relevant data for the load-function. This works as expected with backends that, upon a search service call, returned both the count of total results and the current slice/page of the data in one data structure.

With PF11, the loadLazyData() method has changed, calling count() to fetch the total count before loading the data:

```
Map<String, FilterMeta> filterBy = getActiveFilterMeta();
lazyModel.setRowCount(lazyModel.count(filterBy));

// further setup

List<?> data = lazyModel.load(first, rows, getActiveSortMeta(), filterBy);
```

This requires a backend API design that splits the count and retrieval of the slice, which we don't find beneficial, as it requires twice the amount of endpoints and allows for data modifications in between.

Simply moving the data retrieval from load() to count() also doesn't work, as count() is not passed all parameters required for the complete API call including the slice to fetch (sort/first/rows).

**Describe the solution you would like**
The DataModel should support the previous method of load functions that retrieve the count and the slice in one call, i.e. a singular load function for both. Alternatively, pass all information required to retrieve the slice to count(), so that data retrieval can be performed there, and load() then simply returns the cached data.

tandraschko:
the reason: https://github.com/primefaces/primefaces/issues/1921

as you can read in the linked issue, the DataTable itself requires the state of `rowCount for current filters` and `load current page with current filters`, thats how the DataTable itself works and LazyDataModel mirrors this

you can still just `return 0` in the count method but, like in the past, you have to calculate `first` in your `load` method and manually call 'setRowCount'
so i think your approach is still possible and we wont make the load method optional for now

we could either create a specialized LazyDataModel for your usecase or improveme the docs

tandraschko:
yeah, i think about something like this:

```
public class MySingleCallLazyDataModel<T> extends LazyDataModel<T>
{
    @Override
    public int count(Map<String, FilterMeta> filterBy) {
        return 0;
    }
    
    @Override
    public List<T> load(int first, int pageSize, Map<String, SortMeta> sortBy, Map<String, FilterMeta> filterBy) {
        List<T> data = null; //callApi...., apply sort and filter

        // setRowCount is required if #count method isnt implemented correctly
        int rowCount = data.size();
        setRowCount(rowCount);

        // fix first, see #1921
        if (rowCount > 0 && first >= rowCount) {
[...]

PR #8092, opened 2021-11-20
#8062 Docu DataTable, lazy, avoid separate count-statement against

christophs78:


tandraschko:
The sample misses atleast the setRowCount call and we should mention recalcuateFirst

Issue #8104, opened 2021-11-23
DataTable: clearFilters doesnt work correclty with p:selectBooleanCheckbox


mihaiDumea:
https://github.com/primefaces/primefaces/blob/f5643fc9b4a1574a8d2023a85794aedf390f74cb/primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js#L3966

tandraschko:
same
a issue without description and sample is invalid

mihaiDumea:
Same example ai in #8105 but using the clearFilters instead of filter:
<p:selectBooleanCheckbox id="toggleFilters" value="#{mBean.showFilters}" itemLabel="#{mBean.showFilters ? 'Hide filters' : 'Show filters'}">
        <p:ajax update="items" oncomplete="PF('items').**clearFilters**()" />
    </p:selectBooleanCheckbox>
After clearFilter is executed, further usage of the column filter component will submit their entries with empty value instead of original. Instead of value reset, uncheck is appropriate.

tandraschko:
is it a duplicate of https://github.com/primefaces/primefaces/issues/8071? clearFilters just currently doesnt work perfectly with non-basic inputs like h:*

mihaiDumea:
I just replaced: _this.thead.find('> tr > th.ui-filter-column > .ui-column-customfilter :input').val('');_ with
this.thead.find('> tr > th.ui-filter-column > .ui-column-customfilter :input').each((index, element) => {
        if (element.type === 'checkbox') {
            element.checked = false;
        }
        else {
            element.value = '';
        }
    });

I need more time to check #8071

tandraschko:
could you please check if it works fine for h:selectBooleanCheckbox? If yes, i would close this issue in favor of #8071, i also renamed the title there

mihaiDumea:
h:selectBooleanCheckbox renders as an checkbox so, proposed solution works: it unchecks and does not remove the "on" value.

melloware:
Yep let's close this and @mihaiDumea can you subscribe to #8071 we will track this issue there.

tandraschko:
@mihaiDumea can you retest pls?

Issue #8071, opened 2021-11-11
DataTable: clearFilters doesnt work for complex input widgets like p:selectOneMenu


ltlBeBoy:
**Describe the defect**
When using `p:selectOneMenu` in `DataTable` as custom filter, a call to `DataTable#clearFilters()` resets the filter of the `DataTable` but not the displayed value of the `p:selectOneMenu`. It works as expected when using `h:selectOneMenu` instead.

**Reproducer**
See [primefaces-test.zip](https://github.com/primefaces/primefaces/files/7520464/primefaces-test.zip).

**Environment:**
 - PF Version: _11.0.0-RC2_
 - JSF + version: _Mojarra 2.3.14.payara-p2_
 - Affected browsers: _ALL_

**To Reproduce**
1. Start the reproducer.
2. Select 'true' in column _c1_
3. Select 'true' in column _c2_
4. Click the filter reset button.
5. Filter for `DataTable` is reset, value of _c1_ filter is reset, but value of _c2_ filter is still 'true'.

**Expected behavior**
Value of custom filter with `p:selectOneMenu` should be reset too when calling `DataTable#clearFilters()`.

**Referenced issues**
#6669, #6670

tandraschko:
@melloware @Rapster could you check that?

melloware:
Yes I know this to be an issue.  ClearFilters just does this...

```javascript
 this.thead.find('> tr > th.ui-filter-column > .ui-column-filter').val('');
 this.thead.find('> tr > th.ui-filter-column > .ui-column-customfilter :input').val('');
$(this.jqId + '\\:globalFilter').val('');
```

Which sets all inputs regular and custom back to `''`.  However `p:selectOneMenu` is more complex than just settings its input to ''.  For SelectOne we would have to call custom widget method `PF('id').selectValue('');` or something like that.   And any other special widgets would need to have the same called like any would need a `clear()` method that can be called but I am not sure how we would determine the widget nuless the filter used a `data` attribute to put the widget ID in there.  So most likely @ltlBeBoy you will have to override `clearFilters` and write a custom clear for this table.  It gets even more complex for thinks like `SelectManyChekbox` etc.

tandraschko:
in theory we can look for `.ui-widget` and check `PF.getWidgetById`

melloware:
Oh slick!  Then we could check InstanceOf widget SelectOne and call its seletcValue or selectManyCheckbox like uncheckAll() wdget method etc!

[...]

PR #8107, opened 2021-11-23
Fix #8106: FilterMatchMode.IN check for empty string

### LINKED ISSUE #8106: FilterMatchMode.IN : allow to filter null values
[primefaces-test-master.zip](https://github.com/primefaces/primefaces/files/7590646/primefaces-test-master.zip)
**Description of the defect**
filterMatchMode="in" doesn't properly filter value of null. 
This issue is similar to https://github.com/primefaces/primeng/issues/5735 
* Column has empty cells (null cells)
* Filter selectCheckboxMenu has <f:selectItem itemValue="" itemLabel="(Empty)"/>
* If this (Empty) filter item is selected, cells were not reduced to only empty cells.

**Reproducer**
ZIP is attached.

**Environment:**
 - PF Version: _11.0.0-RC2
 - MyFaces 2.3
 - Affected browsers:  Edge

**To Reproduce**
1. Run demo with mvn jetty:run
2. Filter table by column 'Companies' - select '(Empty)' item in selectCheckboxMenu
3. All rows are still there, but there should be only two rows, where company cells are empty

**Expected behavior**
When `<f:selectItem itemValue="" itemLabel="(Empty)"/>` is selected in selectCheckboxMenu filter, empty cells in this column should remain.  

**Example XHTML**
```html
<h:form>
        <p:dataTable var="customer" value="#{dtFilterView.customers1}" widgetVar="customersTable"
                     filteredValue="#{dtFilterView.filteredCustomers1}">

            <p:column field="company" headerText="Companies" sortBy="#{customer.company}" filterBy="#{customer.company}"
                      filterMatchMode="in">
                <f:facet name="filter">
                    <p:selectCheckboxMenu onchange="PF('customersTable').filter()"
                                          label="Select"
                                          value="#{dtFilterView.selectedCompanies}" >
                        <f:selectItem itemValue="" itemLabel="(Empty)"/>
                        <f:selectItems value="#{dtFilterView.companies}" />
                    </p:selectCheckboxMenu>
                </f:facet>
            </p:column>

        </p:dataTable>
    </h:form>
```

melloware:
Tested with reproducer and its working as expected.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/model/filter/InFilterConstraint.java
+++ b/primefaces/src/main/java/org/primefaces/model/filter/InFilterConstraint.java
@@ -24,13 +24,16 @@
 package org.primefaces.model.filter;
 
 import java.util.*;
+
 import javax.faces.context.FacesContext;
+
+import org.primefaces.util.LangUtils;
 
 public class InFilterConstraint implements FilterConstraint {
 
     @Override
     public boolean isMatching(FacesContext ctxt, Object value, Object filter, Locale locale) {
-        if (filter == null || value == null) {
+        if (filter == null) {
             return false;
         }
 
@@ -47,6 +50,11 @@ public class InFilterConstraint implements FilterConstraint {
 
         for (Object o : collection) {
             if (Objects.equals(value, o)) {
+                return true;
+            }
+
+            // GitHub #8106 check for "" comparison
+            if (o instanceof String && LangUtils.isEmpty((String) o) && value == null) {
                 return true;
             }
         }


alinatoleeva (commented):
Looks great. Thanks for the very quick response and the test.

primefaces/src/main/java/org/primefaces/model/filter/InFilterConstraint.java
```diff
@@ -49,6 +50,11 @@ else if (filter instanceof Collection) {
             if (Objects.equals(value, o)) {
                 return true;
             }
+
+            // GitHub #8106 check for "" comparison
+            if (o instanceof String && o.toString().length() == 0 && value == null) {
```
tandraschko: can o.toString return null?
maybe better user LangUtils.isEmpty(o.toString())

alinatoleeva:
I found out, that with this change the clearFilters() function does not work anymore. 
```
if (PrimeFaces.widget.DataTable) {
	PrimeFaces.widget.DataTable.prototype.clearFilters = function() {
		this.thead.find('> tr > th.ui-filter-column > .ui-column-filter').val('');
		this.thead.find('> tr > th.ui-filter-column > .ui-column-customfilter :input').val(''); //this line
		console.log(undefined)
		$(this.jqId + '\\:globalFilter').val('');
	};
}
```

**this.thead.find('> tr > th.ui-filter-column > .ui-column-customfilter :input').val('')** is setting dropdown's value to empty string and that's why the  **<f:selectItem itemValue="" itemLabel="(Empty)"/>** is getting selected when the clearFilters is called, but expected, that all items in dropdown have to be deselected

tandraschko:
https://github.com/primefaces/primefaces/issues/8071

clearFilters needs to be enhanced for selectCheckboxMenu
lets then check the issue again

tandraschko:
@alinatoleeva 
your issue with clearFilters is unrelated to this issue actually?

alinatoleeva:
> @alinatoleeva your issue with clearFilters is unrelated to this issue actually?

@tandraschko clearFilters has stoped to work correctly when I icluded this fix (#8107) in my project. 
But then I added additional fix for selectCheckboxMenu from here: https://github.com/primefaces/primefaces/issues/8104#issuecomment-976456072  and it works again.

primefaces/src/main/java/org/primefaces/model/filter/InFilterConstraint.java
melloware: Yep switched to Langutils!  Good Catch

melloware:
@alinatoleeva just confirming your real issue had nothing to do with this fix correct?

alinatoleeva:
> @alinatoleeva just confirming your real issue had nothing to do with this fix correct?

@melloware my real issue is #8106 . So #8107 fixed it. 
clearFilters has nothing to do with fix #8107, correct.

PR #8108, opened 2021-11-24
Fix #8071: SelectCheckBoxMenu resetValue

### LINKED ISSUE #8071: DataTable: clearFilters doesnt work for complex input widgets like p:selectOneMenu

melloware:
@tandraschko 2 things.

1. I realize its SelectCheckboxMenu we use a lot like this...

```xml
<f:facet name="filter">
	<p:selectCheckboxMenu id="chkProfileStatusFilter" label="All" filter="true" filterMatchMode="startsWith"
						updateLabel="true">
		<p:ajax event="change" process="@this" oncomplete="PF('organizerTable').filter();" />
		<p:ajax event="toggleSelect" process="@this" oncomplete="PF('organizerTable').filter();" />
		<f:selectItems value="#{profileStatusFilterOptions}" />
	</p:selectCheckboxMenu>
</f:facet>
```

2. I added the `resetValue` using the custom code we have already been using but unfortunately it requires the dialog be visible. `Unselects all available options. Note that this required the overlay panel to be visible, use `show` for that.`

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
@@ -160,8 +160,6 @@ PrimeFaces.widget.SelectCheckboxMenu = PrimeFaces.widget.BaseWidget.extend({
                 this.label.attr('id', this.labelId);
                 this.keyboardTarget.attr('aria-expanded', false).attr('aria-labelledby', this.labelId);
             }
-
-            this.transition = PrimeFaces.utils.registerCSSTransition(this.panel, 'ui-connected-overlay');
         } else {
             // disabled
             if(!this.cfg.multiple) {
@@ -217,6 +215,7 @@ PrimeFaces.widget.SelectCheckboxMenu = PrimeFaces.widget.BaseWidget.extend({
      */
     renderPanel: function() {
         this.panel = $('<div id="' + this.panelId + '" class="ui-selectcheckboxmenu-panel ui-widget ui-widget-content ui-corner-all ui-helper-hidden ui-input-overlay" role="dialog"></div>');
+        this.transition = PrimeFaces.utils.registerCSSTransition(this.panel, 'ui-connected-overlay');
         if(this.cfg.panelStyle) {
             this.panel.attr('style', this.cfg.panelStyle);
         }
@@ -382,7 +381,7 @@ PrimeFaces.widget.SelectCheckboxMenu = PrimeFaces.widget.BaseWidget.extend({
                     return;
                 }
 
-                if($this.cfg.dynamic && !$this.isDynamicLoaded) {
+                if(!$this.isLoaded()) {
                     $this._renderPanel();
                 }
 
@@ -543,7 +542,7 @@ PrimeFaces.widget.SelectCheckboxMenu = PrimeFaces.widget.BaseWidget.extend({
             var keyCode = $.ui.keyCode,
             key = e.which;
 
-            if($this.cfg.dynamic && !$this.isDynamicLoaded) {
+            if(!$this.isLoaded()) {
                 $this._renderPanel();
             }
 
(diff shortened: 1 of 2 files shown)


tandraschko:
why must it be visible? makes it a bit weird

melloware:
It is what the note on the `uncheckAll` and `checkAll` methods say.... not sure why.

melloware:
```
/**
     * Selects all available options. Note that this required the overlay panel to be visible, use `show` for that.
     */
    checkAll: function() {
```

tandraschko:
maybe its such a lazy init/load thing like in selectOneMenu?

melloware:
Yep I think that is it and the code has `panel.is(:visible)` checks in the checkAll and unCheckAll.

tandraschko:
we should really recheck this 
it doesnt make sense to show first to uncheck
also UI will flicker

FlipWarthog:
@melloware Similar to the SelectCheckBoxMenu itself needing to be visible for uncheckAll to work, if `filter="true"` and a filter is applied, unckeckAll won't catch those filtered out values. I think we needed to grab the `filterInput` and clear its value.

melloware:
Agreed we need a deeper diver on this.

melloware:
@FlipWarthog maybe we can deep dive together on this and get this working.  This would be a nice addition.

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
```diff
@@ -217,6 +215,7 @@ PrimeFaces.widget.SelectCheckboxMenu = PrimeFaces.widget.BaseWidget.extend({
      */
     renderPanel: function() {
         this.panel = $('<div id="' + this.panelId + '" class="ui-selectcheckboxmenu-panel ui-widget ui-widget-content ui-corner-all ui-helper-hidden ui-input-overlay" role="dialog"></div>');
+        this.transition = PrimeFaces.utils.registerCSSTransition(this.panel, 'ui-connected-overlay');
```
melloware, line 218: This was a bug. The transition should be created right after the panel is created.  There was actually a bug with `dynamic=true` because of this.

melloware:
OK I cleaned it up and got it all working.  You no longer need to call Show or Hide to uncheck or check all the items. That was all around `dynamic="true"` which was actually broken anyway because the CSS transition was not working in Dynamic Mode.

So I have cleaned all this up and tested `dynamic` mode and non-dynamic.  I then tested widget methods `checkAll`, `uncheckAll` and the new `resetValue` all of them are working properly and do NOT need to display the dialog.

FlipWarthog:
@melloware but do we get any new integration tests for this? ;)

melloware:
> @melloware but do we get any new integration tests for this? ;)

That's a bigger task as there isn't even a base Selenium Component for this yet in the suite and 0 tests.    I will open a separate ticket.

tandraschko:
if you tested all cases -> +1

FlipWarthog (approved):
Tested this in our production code and its working as expected.

PR #8118, opened 2021-11-26
Fix #8113: Table filters now HTML5 search type

### LINKED ISSUE #8113: Filters: Add type="search" to text filters
**Description**
As requested on the forums: https://forum.primefaces.org/viewtopic.php?f=3&t=69745

Adding `type="search"` to filters to allow modern browers to add the "x" to clear filter icons.

PrimeVue issue: https://github.com/primefaces/primevue/issues/83

Simply adding `type="search"` to the generated input boxes for filter inputs

**WORKAROUND**

You can add this 1 line of Jquery code to your page to make all the filters have this...

```javascript
$('input.ui-column-filter.ui-inputfield.ui-inputtext').attr('type', 'search');
```

melloware:
Now the browser adds a useful "x" to filters and clicking the X clears the filter and fires the event.

image

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
@@ -792,6 +792,7 @@ public class DataTableRenderer extends DataRenderer {
         writer.startElement("input", null);
         writer.writeAttribute("id", filterId, null);
         writer.writeAttribute("name", filterId, null);
+        writer.writeAttribute("type", "search", null);
         writer.writeAttribute("class", filterStyleClass, null);
         writer.writeAttribute("value", filterValue, null);
         writer.writeAttribute("autocomplete", "off", null);
--- a/primefaces/src/main/java/org/primefaces/component/treetable/TreeTableRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/treetable/TreeTableRenderer.java
@@ -763,6 +763,7 @@ public class TreeTableRenderer extends DataRenderer {
             writer.startElement("input", null);
             writer.writeAttribute("id", filterId, null);
             writer.writeAttribute("name", filterId, null);
+            writer.writeAttribute("type", "search", null);
             writer.writeAttribute("class", filterStyleClass, null);
             writer.writeAttribute("value", filterValue, null);
             writer.writeAttribute("autocomplete", "off", null);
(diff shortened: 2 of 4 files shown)


melloware:
Also I removed the IE hack because as of IE10 it supports type='search' like all the browsers do so its universally supported.  That old X hack here was in for IE9 and lower.

tandraschko:
lets do it in 12.0

jepsar:
I think this should be optional. We have styled sorting to be displayed at the location of the cross icon. In our case onfocus select would be better.

melloware:
Can you give an example @jepsar ?  There are filters used in many components and I would hate to add a property for every one.  Is there any way you could adjust your code?  This is pretty standard HTML5 option.

jepsar:
Actually it is not that bad...

Screenshot 2021-12-13 at 19 31 11

melloware:
Yeah and you can actually control the "x" with CSS.  IT has a standard style.

mertsincan:
Hi,
I think we can add showClearButton or showClear attribute to Column.  So we can easily style it by themes.

showClear* : When enabled, a clear icon is displayed to clear the value.

https://www.primefaces.org/showcase/ui/input/inputText.xhtml?jfwid=d6123 (Right icon demo)

WDYT?

melloware:
Not a bad idea @mertsincan

tandraschko:
@melloware should we close it for now and transfer merts comment to the ticket?

melloware:
Yep I will do that today.

melloware:
Re-evaluating Mert's comments you can simply  style the "X" with a theme:

https://stackoverflow.com/questions/20804016/editing-input-type-search-pseudo-element-button-x

```css
input[type="search"] {
  border: 1px solid gray;
  padding: .2em .4em;
  border-radius: .2em;
}

input[type="search"].dark {
  background: #222;
  color: #fff;
}

input[type="search"].light {
  background: #fff;
  color: #222;
}

input[type="search"]::-webkit-search-cancel-button {
  -webkit-appearance: none;
  height: 1em;
  width: 1em;
  border-radius: 50em;
  background: url(https://pro.fontawesome.com/releases/v5.10.0/svgs/solid/times-circle.svg) no-repeat 50% 50%;
  background-size: contain;
  opacity: 0;
  pointer-events: none;
}

input[type="search"]:focus::-webkit-search-cancel-button {
  opacity: .3;
  pointer-events: all;
}

input[type="search"].dark::-webkit-search-cancel-button {
  filter: invert(1);
}
```

melloware:
@tandraschko @mertsincan i would like to merge this as it actually improves accessibility and UX

PR #8178, opened 2021-12-17
Fix #8177: InputMask proper autoClear handling

### LINKED ISSUE #8177: InputMask: InputMask does not clear mask symbols when autoClear=false
**Describe the defect**
`p:inputMask autoClear="false"` behaviour differs from how it worked in PrimeFaces 8 for empty field

**Environment:**
 - PF Version: _11.0_

**To Reproduce**
1. Create page with `inputMask autoClear="false"`
2. Observe the empty field behaviour compared to PF8

**Expected behavior**
When nothing is entered the mask should not be visible. 

***Example: PF8 vs PF11***
For mask: `FI99 9999 9999 9999 99`
- PF8
For invalid input the content is visible and mask is shown - this is expected

image

 When empty the mask is not shown
image

- PF11
This is how empty field looks like
image

melloware:
This is the proper fix I should have made originally to make it work like 8.0


primefaces/src/main/resources/META-INF/resources/primefaces/inputmask/1-inputmask.js
```diff
@@ -33,6 +33,11 @@ PrimeFaces.widget.InputMask = PrimeFaces.widget.BaseWidget.extend({
      */
     applyMask: function() {
         if(this.cfg.mask && !this.jq.is('[readonly]') && !this.jq.is(':disabled')) {
+            var autoClear = (this.cfg.autoClear === undefined) ? true : this.cfg.autoClear;
+            if (autoClear) {
+                this.cfg.clearMaskOnLostFocus = true;
```
arnis87: Shouldn't this `this.cfg.clearMaskOnLostFocus = true;` be set regardless of autoClear value for it to work when autoClear=false

melloware: Good catch. Fixed!   I will only set the other one if autoClear = true because its false by default

melloware:
OK all fixed now.

arnis87:
Will this be fixed in 11 elite patch as well?

PR #8197, opened 2021-12-24
Fix #7964: Exporter add exportHeader and Footer options

### LINKED ISSUE #7964: DataExporter: Option to globally suppress table header/footer export
**Description**
In my application, I like to see footers in some of my UI tables, but never in the corresponding exports (XLSX, PDF and XML).

I'm always frustrated when I have to rewrite my custom exporters for each new major PrimeFaces release as each new version brings a more sophisticated API (Nonetheless I really like the work done there!).

I am aware of the `exportFooterValue` attribute in p:Column, but this is not overly helpful due to the sheer amount of columns in use.

I like to have boolean exportFooter and exportHeader attributes on the DataExporter element which globally disables footer and header export.

It would also be great if defaults are sensible here:

PDF: exportFooter='false' exportHeader='true'
XLSX/XSL: exportFooter='false' exportHeader='true'
XML: exportFooter='false' exportHeader='false'

melloware:
I left everything TRUE by default so it matches previous PF versions but now offers the flexibility to turn them off.  cc @stolp

### DIFF
--- a/docs/12_0_0/gettingstarted/whatsnew.md
+++ b/docs/12_0_0/gettingstarted/whatsnew.md
@@ -6,3 +6,5 @@ This page contains a list of big features. Please check the GitHub issues for al
 
 
 Look into [migration guide](https://primefaces.github.io/primefaces/12_0_0/#/../migrationguide/12_0_0) for more enhancements and changes.
+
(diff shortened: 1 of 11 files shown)


melloware:
I made a better PR that kills two birds with one stone.

PR #8198, opened 2021-12-24
Fix #7966/#7964: Exporter `visibleOnly`, `exportHeader` and `exportFooter`

### LINKED ISSUE #7964: DataExporter: Option to globally suppress table header/footer export

### LINKED ISSUE #7966: DataExporter: Option to ignore columns which are hidden via ColumnToggler
**Describe the defect**
**Description**
Data exporters will always export all rendered, exportable columns but not columns excluded via ColumnToggler.
This is confusing for users who may want to tune their table content before exporting.

In the past, I have added checks to my custom exporters for column visibility (isVisible()). I'm always frustrated when I have to rewrite my custom exporters for each new major PrimeFaces release as each new version brings a more sophisticated API (Nonetheless I really like the work done there!).

**Reproducer**
Will follow.

**Environment:**
 - PF Version: _11.0_
 - JSF + version: _ALL_
 - Affected browsers: _ALL_

**To Reproduce**
1. Exclude columns via ColumnToggler
2. Export the table
3. Notice excluded columns appearing in the output

**Expected behavior**
Data exporters should omit columns which are hidden via ColumnToggler.
Perhaps this should be added via an additional attribute to be backwards compatible here.

melloware:
Do not export columns that are not visible.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/export/DataExporter.java
+++ b/primefaces/src/main/java/org/primefaces/component/export/DataExporter.java
@@ -59,6 +59,9 @@ public class DataExporter implements ActionListener, StateHolder {
     private ValueExpression encoding;
     private ValueExpression pageOnly;
     private ValueExpression selectionOnly;
+    private ValueExpression visibleOnly;
+    private ValueExpression exportHeader;
+    private ValueExpression exportFooter;
     private MethodExpression preProcessor;
     private MethodExpression postProcessor;
     private ValueExpression options;
@@ -67,22 +70,6 @@ public class DataExporter implements ActionListener, StateHolder {
 
     public DataExporter() {
         ResourceUtils.addComponentResource(FacesContext.getCurrentInstance(), "filedownload/filedownload.js");
-    }
-
-    public DataExporter(ValueExpression target, ValueExpression type, ValueExpression fileName, ValueExpression pageOnly,
-                        ValueExpression selectionOnly, ValueExpression encoding, MethodExpression preProcessor,
-                        MethodExpression postProcessor, ValueExpression options, MethodExpression onTableRender) {
-        this();
-        this.target = target;
-        this.type = type;
-        this.fileName = fileName;
-        this.pageOnly = pageOnly;
-        this.selectionOnly = selectionOnly;
-        this.preProcessor = preProcessor;
-        this.postProcessor = postProcessor;
-        this.encoding = encoding;
-        this.options = options;
-        this.onTableRender = onTableRender;
     }
 
     @Override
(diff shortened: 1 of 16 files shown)


tuerker:
@melloware I do not appreciate this change since this changes fundamental semantics. we have a lot of tables where we have "complex structured" columns that are not exported while we are providing for such columns a non-visible but exportable counterpart. if now non-visible by default implies non-exportable, then all these tables will not work anymore. i am pretty sure that many other users relying on the semantics implemented so far. 

the column attributes visible and exportable should definitively be kept independent. many thanks in advance for considering this argument.

melloware:
Great point @tuerker   let me see if I can make it configurable.

melloware:
OK added new `visibleOnly` option which is false by default so its like current PF8/10/11 behavior.

melloware:
@tandraschko I think I found a bug while testing this I would like to discuss when you have time.  Not sure if you are off on holiday.

melloware:
Fixed all my issues!

tandraschko:
hmmm
i dont like to increase max constructor params in checkstyle, maybe use a builder?

melloware:
I was thinking the same thing. I didn't like it either I will fix.

melloware:
Set checkstyle back to 10 params and switched to builder pattern

tandraschko:
the Builder doesnt look what we did for our SortMeta/Filter and so on, everything else looks good for me

melloware:
@tandraschko I followed the same pattern `ExportConfiguration.java` was using.  Should they both be modified to use the same builder pattern as SortMeta?

tandraschko:
i think so
its a a normal pattern that "setBlub" is void and a builder is just called "blub"

melloware:
OK updated both classes to use the common builder pattern.

primefaces/src/main/java/org/primefaces/component/export/DataExporter.java
```diff
@@ -249,30 +254,159 @@ public void restoreState(FacesContext context, Object state) {
...
+        values[13] = exporter;
 
         return (values);
     }
+
+    public static DataExporterBuilder builder() {
+        return new DataExporterBuilder();
+    }
+
+    public static class DataExporterBuilder {
```
tandraschko: just call it Builder?

primefaces/src/main/java/org/primefaces/component/export/ExportConfiguration.java
```diff
@@ -30,81 +30,170 @@
...
-        return outputFileName;
+    public ExportConfiguration() {
+        // NOOP
+    }
+
+    public static ExportConfigurationBuilder builder() {
+        return new ExportConfigurationBuilder();
+    }
+
+    public static class ExportConfigurationBuilder {
```
tandraschko: also just "Builder"?

melloware:
renamed both to builder

PR #8203, opened 2021-12-25
Fix #8202: AutoComplete: support LazyDataModel to search instead of completeMethod

### LINKED ISSUE #8202: AutoComplete: support LazyDataModel to search instead of completeMethod
You might have many `LazyDataModels` in your project which are almost suitable to serve as an `AutoComplete` suggestions
provider. They could be used with `AutoComplete` in combination with a field attribute. The query string could be applied on the provided `lazyModelField` using `MatchMode.CONTAINS`.

jepsar:
Fix #8202


jepsar:
Maybe we should just add an extra property to search for, instead of the interface.

tandraschko:
i like the approach to simple reuse a LazyDataModel but "lazyModelField" usage reads bit weird. Propably `#{customer.name}` would be better for endusers

tandraschko:
can you try `ValueExpression#getValueReference(elContext).getProperty()` instead str search?

jepsar:
I'm getting `Target Unreachable, identifier [customer] resolved to null` and we have no customer (var instance) to set there.

Rapster:
-1 for this fix. See my coment in original ticket. Furthermore, parsing property name from EL is a dirty trick _(no matter how well you try to get it, you'll always have a use case where it's too tricky to get it)_. Having an attribute for the targeted field would be better but like @tandraschko said, there is something off about it

jepsar:
I had a string field attribute before. See https://github.com/primefaces/primefaces/pull/8203/commits/c358a5f5aa66cc59314de2c5aea87c71be102f8f I agree the EL "parsing" is ugly. Could be done by a reg exp or we can revert to string.

I like the solution to be able to use an existing lazy model as is. If you would need to modify it or wrap it, you can just as well use it in the completeMethod.

tandraschko:
yeah, hmmm
we had similar issue in DataTable, where we ref the sort/filter field just by string (`field=name`). it works but a EL is better to read and code completion by IDE (`field=#{customer.name}`)

tandraschko:
@jepsar can you resolve conflicts?

jepsar:
@tandraschko done

PR #8214, opened 2021-12-29
Fix #2521: CommandButton: add option to disable button on Ajax requests

### LINKED ISSUE #2521: Prevent double click on UICommand components
This topic is quite recurrent, would it be possible to integrate a solution in PF?

https://stackoverflow.com/questions/10756426/how-to-do-double-click-prevention-in-jsf-2
http://ovaraksin.blogspot.fr/2011/05/prevent-multiple-clicks-in-jsf.html

Components:
- CommandButton
- CommandLink
- MenuItem
- SplitButton
- HotKey

### LINKED ISSUE #8213: CommandButton: add option to disable button on Ajax requests
Add option to disable the button on Ajax requests triggered by that button.

jepsar:
Fix #2521


Rapster:
Could we do this for every UICommand in PF?

jepsar:
If we would add `disable` to `SplitButton`, we can add it there. `MenuButton` has `deactivate` but it requires a menu item to disable. We might want to disable the entire component.

I don't think disabling other menus will make sense.

`CommandLink` has no widget.

Rapster:
As long it's possible to invoke an action, it makes sense. Would it be a lot of work to make commandLink a widget? But wait a sec, what does this fix then if not that issue https://www.primefaces.org/showcase-ext/sections/renderers/commandButtonSingleClick.jsf ?

jepsar:
We can have a look at adding a widget for `CommandLink` to disable it. #8218

I don't understand your last question. The renderer in PFE can be deprecated and removed if this PR is accepted.

Rapster:
Ok cool I just wanted to make sure of that

tandraschko:
i would like to see the "getSourceIs" in our ajax.js as util and maybe reimplement as "isXhrSource(...)"

jepsar:
> i would like to see the "getSourceIs" in our ajax.js as util and maybe reimplement as "isXhrSource(...)"

Okay. I'll also refactor the block UI JavaScript.

jepsar:
I'll create a new PR for block UI.. it was causing to much noise on this PR

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
```diff
@@ -148,6 +148,32 @@ if (!PrimeFaces.ajax) {
...
+            /**
+             * Checks whether the component ID from the provided configuration equals the source ID from the provided
+             * settings.
+             *
+             * @param {PrimeFaces.PartialWidgetCfg<TCfg>} cfg of the component to check for being the source.
+             * @param {JQuery.AjaxSettings} settings containing source ID.
+             * @returns {boolean} `true` if the component ID from the provided configuration equals the source ID from
+             * the provided settings.
+             */
+            isXhrSource: function(cfg, settings) {
```
tandraschko: i would either pass the id or the whole widget

tandraschko: not sure ;D i dont see any reason to pass a wrapper object from the middle of the stack, isntead of the whole widget or only the id

PR #8219, opened 2021-12-31
Fix #8218: CommandLink: add widget with disable / enable option

### LINKED ISSUE #8218: CommandLink: add widget with disable / enable option
Currently there is no widget for CommandLink.

jepsar:
Fix #8218

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/commandlink/CommandLinkRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/commandlink/CommandLinkRenderer.java
@@ -63,103 +63,102 @@ public class CommandLinkRenderer extends CoreRenderer {
 
     @Override
     public void encodeEnd(FacesContext context, UIComponent component) throws IOException {
-        ResponseWriter writer = context.getResponseWriter();
         CommandLink link = (CommandLink) component;
+
+        encodeMarkup(context, link);
+        encodeScript(context, link);
+    }
+
+    protected void encodeMarkup(FacesContext context, CommandLink link) throws IOException {
+        ResponseWriter writer = context.getResponseWriter();
         String clientId = link.getClientId(context);
         Object label = link.getValue();
 
-        if (!link.isDisabled()) {
-            String request;
-            boolean ajax = link.isAjax();
-            String styleClass = link.getStyleClass();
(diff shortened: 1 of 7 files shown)


primefaces/src/main/java/org/primefaces/component/commandlink/CommandLinkRenderer.java
```diff
@@ -63,103 +63,102 @@ public void decode(FacesContext context, UIComponent component) {
...
-            writer.writeAttribute(HTML.ARIA_LABEL, link.getAriaLabel(), null);
+        writer.startElement("a", link);
+        writer.writeAttribute("id", clientId, "id");
+        writer.writeAttribute("href", "#", null);
+        writer.writeAttribute("class", styleClass, null);
+        writer.writeAttribute(HTML.ARIA_LABEL, link.getAriaLabel(), null);
+        if (link.isDisabled()) {
+            writer.writeAttribute("tabindex", "-1", null);
+            writer.writeAttribute("aria-disabled", "true", null);
+        }
```
jepsar, line 104: Basically this is the only relevant change in this block. This replaces the disabled block.

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.commandlink.js
```diff
@@ -0,0 +1,68 @@
...
+        });
+    },
+
+    /**
+     * Disables this link so that the user cannot click the link anymore.
+     */
+    disable: function() {
+        this.jq.removeClass('ui-state-hover ui-state-focus ui-state-active')
+                .addClass('ui-state-disabled')
+                .attr('tabindex', '-1')
```
melloware, line 55: Hmmm shouldn't we save off the tabindex so if its reinstated the same tabindex is put back on in the enable method?

melloware (changes requested):

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.commandlink.js
jepsar: I introduced the tabindex and it is hard-coded.

melloware (approved):

PR #8235, opened 2022-01-05
Fix #8233: LinkButton allow widget disable/enable.

### LINKED ISSUE #8233: LinkButton: Enable/Disable via JS does not work correctly
**Describe the defect**

**Defect 1 (disable => enable):**

Create p:linkButton
`<p:linkButton id="btn" widgetVar="btn" disabled="true" value="test"/>`

Try to enable the button later with Javascript:
`PF('btn').enable()`

Expected behavior: Button works, Link is clickable
Actual behaviour: Button does not work, no clickable link

**Defect 2 (enable=> disable):**

Create p:linkButton
`<p:linkButton id="btn" widgetVar="btn" disabled="false" value="test"/>`

Try to disable the button later with Javascript:
`PF('btn').disable()`

Expected behavior: Button deactivated, no interaction possible
Actual behavior: Button does not respond to mouse (correct), but link can still be accessed by "tab" and activated by "enter/return"

**Environment:**
 - PF Version: _10.0_
 - Affected browsers: _ALL_

melloware:


tandraschko:
but the JS widget doesnt disable it in the in the init method?

melloware:
The renderer is already setting ui-disabled and and render pass through attributes is setting the disabled attribute so it's all working.

PR #8258, opened 2022-01-09
Fix #8246: Datatable: ComparableFilterConstraint should support numeric values

### LINKED ISSUE #8246: Datatable: filter on Integer uses String comparison
g.percelen is an Integer

The following xhtml:

```
<p:column headerText="#percelen" sortBy="#{g.percelen}" filterBy="#{g.percelen}" filterMatchMode="gt">
  #{g.percelen}
</p:column>
```
results in String comparison for gt instead of numeric comparison

**Environment:**
 - PF Version: _10.0.8
 - JSF + version: Mojarra on payara 5.2021.9
 - Affected browsers: firefox 95.0.1 linux

**To Reproduce**
Create a pojo with an Integer field percelen
Create a Datatable for it with a list of these objects in a controler
values for percelen to be in the list: 5 54 6 100
Create a xhtml page with the snippet above in it
Run the app
when entering filter value 54, 6 will be shown

**Expected behavior**
when entering filter value 54, 6 will not be shown

jepsar:
Fix #8246

### DIFF
--- a/primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java
+++ b/primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java
@@ -39,9 +39,8 @@ public abstract class ComparableFilterConstraint implements FilterConstraint {
             return false;
         }
         if (value instanceof Number) {
-            return getPredicate()
-                    .test(toBigDecimal(value, locale),
-                            toBigDecimal(filter, locale));
+            BigDecimal filterBigDecimal = toBigDecimal(filter, locale);
+            return filterBigDecimal != null && getPredicate().test(toBigDecimal(value, locale), filterBigDecimal);
         }
         else {
             return getPredicate()
--- a/primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java
+++ b/primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java
@@ -35,28 +35,31 @@ public abstract class ComparableFilterConstraint implements FilterConstraint {
 
     @Override
     public boolean isMatching(FacesContext ctxt, Object value, Object filter, Locale locale) {
-        if (value == null) {
+        if (value == null || filter == null) {
             return false;
         }
         if (!(value instanceof Comparable) || !(filter instanceof Comparable)) {
-            throw new IllegalArgumentException("Invalid type: " + value.getClass() + ". Valid type: " + Comparable.class.getName());
+            throw new IllegalArgumentException(
+                    "Invalid type: " + value.getClass() + ". Valid type: " + Comparable.class.getName());
         }
-
         if (value instanceof Number) {
-            return getPredicateBigDecimal()
-                    .test(new BigDecimal(value.toString()),
+            return getPredicate()
+                    .test(toBigDecimal(value, locale),
                             toBigDecimal(filter, locale));
         }
         else {
-            return getPredicateString()
+            return getPredicate()
                     .test(toString(value, locale),
                             toString(filter, locale));
         }
     }
 
     protected BigDecimal toBigDecimal(Object object, Locale locale) {
-        if (object == null) {
-            return null;
+        if (object == null || object instanceof BigDecimal) {
+            return (BigDecimal) object;
+        }
+        if (object instanceof Number) {
+            return new BigDecimal(object.toString());
         }
         try {
             DecimalFormat decimalFormat = (DecimalFormat) NumberFormat.getInstance(locale);
@@ -75,8 +78,6 @@ public abstract class ComparableFilterConstraint implements FilterConstraint {
         return object.toString().trim().toLowerCase(locale);
     }
 
-    protected abstract BiPredicate<BigDecimal, BigDecimal> getPredicateBigDecimal();
-
-    protected abstract BiPredicate<String, String> getPredicateString();
+    protected abstract BiPredicate<Comparable, Comparable> getPredicate();
 
 }
--- a/primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java
+++ b/primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java
@@ -38,10 +38,6 @@ public abstract class ComparableFilterConstraint implements FilterConstraint {
         if (value == null || filter == null) {
             return false;
         }
-        if (!(value instanceof Comparable) || !(filter instanceof Comparable)) {
-            throw new IllegalArgumentException(
-                    "Invalid type: " + value.getClass() + ". Valid type: " + Comparable.class.getName());
-        }
         if (value instanceof Number) {
             return getPredicate()
                     .test(toBigDecimal(value, locale),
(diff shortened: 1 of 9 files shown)


melloware (approved):

primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java
```diff
@@ -23,17 +23,60 @@
...
+            return false;
+        }
         if (!(value instanceof Comparable) || !(filter instanceof Comparable)) {
             throw new IllegalArgumentException("Invalid type: " + value.getClass() + ". Valid type: " + Comparable.class.getName());
         }
 
-        return super.isMatching(ctxt, value, filter, locale);
+        if (value instanceof Number) {
+            return getPredicateBigDecimal()
+                    .test(new BigDecimal(value.toString()),
```
tandraschko: shouldnt this use the locale as well?

primefaces/src/main/java/org/primefaces/model/filter/LessThanEqualsFilterConstraint.java
```diff
@@ -23,12 +23,18 @@
...
 package org.primefaces.model.filter;
 
+import java.math.BigDecimal;
 import java.util.function.BiPredicate;
 
 public class LessThanEqualsFilterConstraint extends ComparableFilterConstraint {
 
     @Override
-    protected BiPredicate<String, String> getPredicate() {
+    protected BiPredicate<String, String> getPredicateString() {
```
tandraschko: couldnt we just have one <Comparable, Comparable> Predicate instead of 2?

jepsar: I didn't try because comparables are typed and am not sure if we would be running into generic type issues. I can give it a try.

primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java
jepsar: No, this is the value from the model. The other string is the entered filter value and can have a decimal separator based on the locale.

tandraschko: but coulnd this destroy the number somehow? Like value is Double, you do a toString and new BigDecimal on it? What about the locale in this case?

jepsar: https://stackoverflow.com/questions/16216248/convert-java-number-to-bigdecimal-best-way

primefaces/src/main/java/org/primefaces/model/filter/LessThanEqualsFilterConstraint.java
jepsar: Okay, no issues with generics. Works with both numbers and strings.

primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java
```diff
@@ -23,17 +23,61 @@
...
 
-public abstract class ComparableFilterConstraint extends StringFilterConstraint {
+public abstract class ComparableFilterConstraint implements FilterConstraint {
 
     @Override
     public boolean isMatching(FacesContext ctxt, Object value, Object filter, Locale locale) {
+        if (value == null || filter == null) {
+            return false;
+        }
         if (!(value instanceof Comparable) || !(filter instanceof Comparable)) {
```
jepsar: We don't really need this check as the fallback is converting both sides to string

primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java
```diff
@@ -23,17 +23,50 @@
...
 
-        return super.isMatching(ctxt, value, filter, locale);
+    static BigDecimal toBigDecimal(Object object, Locale locale) {
+        if (object instanceof BigDecimal) {
+            return (BigDecimal) object;
+        }
+        if (object instanceof Number) {
+            return new BigDecimal(object.toString());
+        }
+        try {
```
jepsar: Maybe add an instanceof string check here?

jepsar:
Sorry for the additional commits. Ready for review now.

primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java
```diff
@@ -23,17 +23,52 @@
...
+        }
+    }
 
-        return super.isMatching(ctxt, value, filter, locale);
+    static BigDecimal toBigDecimal(Object object, Locale locale) {
+        if (object instanceof BigDecimal) {
+            return (BigDecimal) object;
+        }
+        if (object instanceof Number) {
+            return new BigDecimal(object.toString());
```
melloware, line 57: Can object be NULL here which would cause an NPE on the `toString() call?

primefaces/src/main/java/org/primefaces/model/filter/StringFilterConstraint.java
```diff
@@ -25,19 +25,23 @@
...
-        String str = filter == null ? null : filter.toString().trim().toLowerCase(locale);
-        String val = value == null ? null : value.toString().toLowerCase(locale);
-        return val != null && predicate.test(val, str);
+        return getPredicate()
+                .test(toString(value, locale),
+                        toString(filter, locale));
+    }
+
+    static String toString(Object object, Locale locale) {
+        return object.toString().trim().toLowerCase(locale);
```
melloware, line 44: Same here can object ever be NULL?

melloware (changes requested):

melloware:
Couple of comments but I love switching to `Comparable` it makes it so much cleaner and more flexible.

primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java
jepsar: No, it has been checked on line 38 and null instanceof Nummber is false.

primefaces/src/main/java/org/primefaces/model/filter/StringFilterConstraint.java
jepsar: No, it was checked on line 34.

melloware (approved):

jepsar:
@melloware the only thing that could be improved (but that's a bigger change which requires a new issue IMO) is that the filter value is processed for each row. Would be better if we could cache it in the request. The current component returns the entire table, so that cannot be used. We need the column, or a cache key based on the column to do that.

tandraschko:
looks good for me
@Rapster can you review plz?

tandraschko:
same here, a integration test would be nice

melloware:
@jepsar I would think this would be easy to modify one of our existing filtering examples to filter on a numeric field.

jepsar:
Can I just modify any data table test? It is hard to tell which test does what based on the name, and if I search on `filter` in `org.primefaces.integrationtests.datatable` I get results in 20 files.

melloware:
I would pick DataTable008Test since its a single test filtering on First Appeared maybe you can add another test there.

jepsar:
Shouldn't that test have failed, as this PR exactly fixes that type of filter (gte on a numeric value)?

melloware:
That is what I am wondering...

jepsar:
I'll create a new (non session scoped) test class and leave the disabled one as it is.

PR #8260, opened 2022-01-09
Add setup for separate checkstyle config for tests

krichter722:
Some basic rules which make sense to start with.

### DIFF
--- a/conf/checkstyle-test.xml
+++ b/conf/checkstyle-test.xml
@@ -1,0 +1,7 @@
+﻿<?xml version="1.0"?>
+<!DOCTYPE module PUBLIC
+    "-//Puppy Crawl//DTD Check Configuration 1.2//EN"
+    "https://checkstyle.org/dtds/configuration_1_2.dtd">
+<module name="Checker">
+</module>
+
--- a/conf/checkstyle-test.xml
+++ b/conf/checkstyle-test.xml
@@ -3,6 +3,32 @@
     "-//Puppy Crawl//DTD Check Configuration 1.2//EN"
     "https://checkstyle.org/dtds/configuration_1_2.dtd">
 <module name="Checker">
+
     <module name="FileTabCharacter"/>
+
+    <module name="SuppressWithPlainTextCommentFilter">
+        <property name="offCommentFormat" value="CHECKSTYLE:OFF"/>
+        <property name="onCommentFormat" value="CHECKSTYLE:ON"/>
+    </module>
+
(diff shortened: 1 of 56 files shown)


tandraschko:
we use the same checkstyle by design, so we wont merge this

krichter722:
@tandraschko I don't understand, currently you're not using any checkstyle for tests (this is due to includeTestSourceDirectory=false by default).

tandraschko:
then its currently by mistake
free to change includeTestSourceDirectory and fix the checkstyle but adding a second checkstyle isnt good

krichter722:
OK, I tried that at first, but you need to make some sacrifices for that, e.g. test methods can't have `_` in you current setup. Do you want the `_` to be removed from test method names (become even harder to read) or the rule to be dropped for prod code?

Maybe you want to think about the separate setup again based on this example with the test method names which would allow you to have almost the same setup except for some rules that just make sense for tests.

melloware:
I thought we excluded tests because they are just tests and yeah they would violate a lot of our check style rules.

tandraschko:
If its just _, i think we could just add it for prod code

krichter722:
I'm just doing (early) branch spring cleaning. It's sort of meditative work for me, for you it's your code base. I don't know if it's just `_`, however I think dropping a rule for prod has way more impact than adding a separate smaller ruleset for test which wasn't covered before.

Test methods with `_` could be surrounded by a suppression as well. However that's a lot of effort in comparison to the separate ruleset.

tandraschko:
for me it seems that its just "_" in method names
so i would go for add this in our global checkstyle and just fix tests

krichter722:
I'll try that. Just to make sure, because you talked about adding a rule twice. The rule to forbid `_` in method names will be **removed** from the ruleset for prod. Only that allows the same ruleset to pass for prod and test. It also allows (usually unwanted) method names with `_` in prod code.

tandraschko:
yeah
we can also copy the config + allow _ but everything else must be the same config

tandraschko:
applied our default checkstyle to the tests, with 2 modifications:
allow _ in methods
allow package scoped variables for our Page classes in the integration tests

krichter722:
OK. The work was also in the PR...

The removal of final is maybe not intended, usually you want as much immutability as possible, even on the local level. That might even be enforced with a rule.

I'd wrap the classes in `TerminalAutoCompleteTest` in `//CHECKSTYLE:OFF/ON` because they're test data rather than code.

PR #8269, opened 2022-01-12
Fix bad property name in virtual scroll example

YotillaPablo:
The DataTable documentation for PrimeFaces 10 shows an example of virtual scrolling with a wrong property: it says virtual instead of virtualScroll.


tandraschko:
can you also fix it in 11.0 + 12.0?

YotillaPablo:
> can you also fix it in 11.0 + 12.0?

Sure, there it is

tandraschko:
thanks!

PR #8284, opened 2022-01-15
#8257 Timeline: rework according to Schedule-component

### LINKED ISSUE #8257: Timeline: rework according to Schedule-component
see https://github.com/primefaces/primefaces/issues/8076, https://github.com/primefaces/primefaces/pull/8090

Currently it´s essentially broken due to the changes we did with https://github.com/primefaces/primefaces/pull/8090. (Broken means server- and client-time are out of sync.)

- [x] adapt Timeline-component to be in sync with https://github.com/primefaces/primefaces/pull/8090
- [x] adapt Showcase
- [x] provide integration-tests (like Schedule001Test)

christophs78:


christophs78:
Currently creates a new issue with daylight-saving:
image

Should be 23:00 instead of 22:00.

As of testing this (january 2022 we do not have daylight saving) this event is placed in august (with daylight saving).

(Schedule does not have this issue.)

christophs78:
Hm. This is more complicated (and different to Schedule / FullCalendar) ...

Because of https://github.com/primefaces/primefaces/commit/318b631919f623b4efe32618f9e38774cec912a4 / https://github.com/primefaces/primefaces/issues/5320.

--> We need integrationtests for Schedule and Timeline with and without daylight saving, ...
Way more work.

christophs78:
Now it´s up and working again for usecases where timeZone- and clientTimeZone-attributes are not set AND server and client are within the same timezone. (IMO the usecase of most people.)
clientTimeZone is a bit difficult. IMO because vis.js is maybe not perfect in this aspect. https://visjs.github.io/vis-timeline/docs/timeline/#Time_zone lets you define timezone. But this does not seem to keep up with understanding daylight savings.
@jasonex7 : Do you still use PrimeFaces? If yes --> may you look at this PR and see whether it´s still working for your usecase. (reference https://github.com/primefaces/primefaces/issues/5320) We did some changes to Schedule-component before which also affects Timeline due to shared code.
@tuerker: Maybe you can cross-check this against your usecases? (We introduced new issues during working on https://github.com/primefaces/primefaces/issues/8076)

jasonex7:
@christophs78 I'm still using PrimeFaces. We currently use timeline without setting any timezone parameter (client nor server). But I can use this PR and run our Serenity integration test with it and see if this is all right, I'll do that on Monday.

Recently I've updated from PrimeFaces 8.x to 10.x and have some issues, but now our Serenity tests pass again.

tuerker:
@christophs78  looks good to me. my use cases are running. thanks.

jasonex7:
> @christophs78 I'm still using PrimeFaces. We currently use timeline without setting any timezone parameter (client nor server). But I can use this PR and run our Serenity integration test with it and see if this is all right, I'll do that on Monday.

@christophs78 All my tests passed using your branch [christophs78:8257_Timeline](/christophs78/primefaces/tree/8257_Timeline). Although I don't know if this counts because as I say, we don't use timezone parameters in our project.

image

melloware (approved):

tandraschko:
please dont wait about a review from my side
you know schedule/timeline better then me

PR #8287, opened 2022-01-16
Fix #8286: Tree: add filterFunction for custom filtering

### LINKED ISSUE #8286: Tree: add filterFunction for custom filtering
So you can filter in your bean like:
```java
public boolean customFilter(TreeNode treeNode, Object filter, Locale locale) {
    // return true if the filter matches the tree node
}
```

jepsar:
Fix #8286

### DIFF
--- a/primefaces-showcase/src/main/java/org/primefaces/showcase/view/data/tree/BasicView.java
+++ b/primefaces-showcase/src/main/java/org/primefaces/showcase/view/data/tree/BasicView.java
@@ -72,12 +72,15 @@ public class BasicView implements Serializable {
         return root;
     }
 
-    public boolean customFilter(Object data, Object filter, Locale locale) {
-        String filterText = (filter == null) ? null : filter.toString().trim().toLowerCase(locale);
-        if (LangUtils.isBlank(filterText) || data == null) {
+    public boolean customFilter(TreeNode treeNode, Object filter, Locale locale) {
+        if (treeNode.getData() == null || filter == null) {
             return true;
         }
-        return ((String) data).toLowerCase(locale).contains(filterText);
+        String filterText = filter.toString().trim().toLowerCase(locale);
+        if (LangUtils.isBlank(filterText)) {
+            return true;
+        }
+        return ((String) treeNode.getData()).toLowerCase(locale).contains(filterText);
     }
 
 }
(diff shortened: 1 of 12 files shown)


primefaces/src/main/java/org/primefaces/component/tree/TreeRenderer.java
```diff
@@ -278,7 +279,10 @@ protected void encodeFilteredNodes(FacesContext context, Tree tree, TreeNode<?>
                 TreeNode childNode = node.getChildren().get(i);
                 requestMap.put(var, childNode.getData());
 
-                if (filterConstraint.isMatching(context, filterByVE.getValue(context.getELContext()), filteredValue, filterLocale)) {
+                Object value = tree.getFilterFunction() == null
+                        ? filterByVE.getValue(context.getELContext())
+                        : childNode.getData();
```
jepsar: I was doubting to just use the node instead of its data.

primefaces-showcase/src/main/java/org/primefaces/showcase/view/data/tree/BasicView.java
```diff
@@ -69,4 +71,13 @@ public void init() {
     public TreeNode getRoot() {
         return root;
     }
+
+    public boolean customFilter(Object data, Object filter, Locale locale) {
+        String filterText = (filter == null) ? null : filter.toString().trim().toLowerCase(locale);
+        if (LangUtils.isBlank(filterText) || data == null) {
```
melloware: Just a minor performance comment you can probably
Check data == null up front and bail out before checking filter text etc

melloware (changes requested):

primefaces-showcase/src/main/java/org/primefaces/showcase/view/data/tree/BasicView.java
jepsar: Okay. But what do you think about my previous remark? Should we pass the data or the tree node?

melloware: Hmm maybe the whole tree node for more flexibility?

jepsar: I was thinking the same.. will change it

primefaces/src/main/java/org/primefaces/component/tree/Tree.java
```diff
@@ -127,6 +127,10 @@ public UITreeNode getUITreeNodeByType(String type) {
         }
     }
 
+    public boolean isWithFilter() {
```
melloware: how about `isFiltering()` instead?

melloware (changes requested):

melloware (approved):

tandraschko:
it would be great to have some integration tests

Issue #8122, opened 2021-11-28
AutoComplete: implement inline AJAX status


nimo23:
**Description**
The [autoComplete](https://www.primefaces.org/showcase/ui/input/autoComplete.xhtml?jfwid=3a3d0) fetches a list of data from the backend and provides suggestions based on this data. This data can come from different sources (e.g. Rest-Endpoints), so it can take a while when the data is arrived. Actually, for such use cases, we can provide a global [ajax-status](https://www.primefaces.org/showcase/ui/ajax/status.xhtml?jfwid=3a3d0) but another alternative would be to provide a _local status_ for such use cases.

**Describe the solution you would like**
A new property for `autoComplete` called `status`, which represents a delay in milliseconds before displaying the ajax status on the right side (or left side) within the input-text-field. If this `status = null`, then no ajax-status is shown.

The `autoComplete`with its spinner can look like [Right Icon Example](https://www.primefaces.org/showcase/ui/input/inputText.xhtml?jfwid=3a3d0 )

melloware:
@nimo23 I think I get the idea.  Do you want it more like how PrimeReact does it where it shows the spinner icon right inside the input right?

See: https://primefaces.org/primereact/showcase/#/autocomplete

nimo23:
@melloware exactly like that.

jepsar:
I'll have a look at this, as it relates to the work I've been doing on the CommandButton

PR #8302, opened 2022-01-19
Fix #8301: CommandButton: add inline Ajax status indicator

### LINKED ISSUE #8301: CommandButton: add inline Ajax status indicator
The icon or text can be temporarily replaced with a spinner while an Ajax request is in progress.

commandbutton

Note that the speed of the GIF is slower than real time (don't want to spend time on getting the GIF right).

```xhtml
<h5>Ajax Disabling and Status Icon</h5>
<p:commandButton value="Disable" action="#{buttonBasicView.sleep()}" />
<p:commandButton value="Status" disableOnAjax="false" action="#{buttonBasicView.sleep()}" />
<p:commandButton value="Both + Icon" icon="pi pi-check" action="#{buttonBasicView.sleep()}" />
```

### LINKED ISSUE #8303: CommandButton: Ajax requests should toggle a CSS class allowing style overriding
I'm thinking to disable the button as we do now, but additionally have a CSS class while the Ajax request is in progress and based on that not set the opacity (will require a theme change). Not sure if this should be optional.

jepsar:
Fix #8301
Fix #8303

commandbutton

Note that the speed of the GIF is slower than real time (don't want to spend time on getting the GIF right).

```xhtml
<h5>Ajax Disabling and Status Icon</h5>
<p:commandButton value="Disable" action="#{buttonBasicView.sleep()}" />
<p:commandButton value="Status" disableOnAjax="false" action="#{buttonBasicView.sleep()}" />
<p:commandButton value="Both + Icon" icon="pi pi-check" action="#{buttonBasicView.sleep()}" />
```


melloware:
looks like the build failed?

melloware:
Similar to this request for AutoComplete: https://github.com/primefaces/primefaces/issues/8122

jepsar:
Going to reimplement this based on #8303

melloware (approved):

tandraschko:
i like how it works on the UI, i just need to think abit about making it easier to easy somehow

maybe somehow using p:ajaxStatus or create a attachable p:inlineAjaxStatus, or reuse p:ajax, dont know
at least for all other components it would make sense to reuse somehow p:ajax
commandButton is a bit different of course

tandraschko:
also maybe for disableOnAjax

we need to discuss first, before add this to more components

jepsar:
I'd prefer not to use child components, so you can easily set it globally. A context parameter could be used so you can enable it on all components that support inline Ajax status / disabling.

I already did `AutoComplete` (#8316) by the way. I was waiting on a review before doing `SplitButton` and `MenuButton`.

jepsar:
Maybe using `p:ajaxStatus` is the cleanest solution. We could add properties `disableTriggers` and `inlineStatus`, which we can make available as global JavaScript properties, which can be used by widgets supporting them. This allows us to  eliminate `inlineAjaxStatus` and `disableOnAjax` on various components.

jepsar:
Anyone other ideas or feedback?

tandraschko:
give me a week, had a busy week

jepsar:
@tandraschko Not trying to be pushy here, but I'd like to finish this PR. If you can find some time, can you please have a look?

tandraschko:
i thought a bit about it but i think the cleneast solution is to make usage of AjaxSource#global

if global = false, inline should be used when the component supports it
CommondButton, CommandLink, p:ajax all extends AjaxSource
therefore you can also configure it on each event

jepsar:
I'm cool with that. That still leaves you with the option to globally set all buttons to global="false" in your Application.

But, will global=false always trigger the inline status (so we drop inlineAjaxStatus)? Or do you still need to set inlineAjaxStatus together with global=false? Maybe we should keep it and default inlineAjaxStatus to true.

jepsar:
.. and maybe rename inlineAjaxStatus to localAjaxStatus?

tandraschko:
i think we have to clarify, what options do we need.
Can inline ajax status always be enabled? do a user need to disable it? How does PrimeVue etc. handle this?
Do we need a option enable both inline & global or disable both?


do we need a rule like only switch icons , if the component has a icon? (support only button 3 in your example above)
then we can always enable it IMO and only add ui-state-loading via JS to the component root
everything else is CSS

can you also post a gif how autocomplete looks currently?

jepsar:
Added a GIF to #8316

primefaces-showcase/src/main/webapp/ui/button/commandButton.xhtml
```diff
@@ -119,8 +119,43 @@
...
+                html body .custom.ui-button.ui-state-loading.ui-button-text-only .ui-button-text {
+                    opacity: 1;
+                    padding-left: 2rem;
+                }
+                html body .custom.ui-button.ui-state-loading.ui-button-text-only .ui-icon-loading {
+                    opacity: 1;
+                    left: 0;
+                    margin-left: 0.5rem;
+                }
+            </style>
```
jepsar: Example of custom CSS rule set

jepsar:
To answer your other questions:

In PrimeVue it is simply a loading boolean. I would suggest keep the attribute `inlineAjaxStatus`, but have three options:

1. `false` (default): no inline status indicator.
2. `local`: only if `global` is set to `false`.
3. `true`: for both local and global.

Stylingwise I would keep it as the PR currently is. I don't like the variant were the button widens to add an icon, although I've added that as an example of how to create a custom CSS rule set. So, I wouldn't offer properties in component to give stylings options. They can be done using the CSS class `ui-state-loading` that is toggled.

tandraschko:
WDYT about something like? names not final ofc

1) remove global=true|false
2) add ajaxEventPropagation=component,global,both,none to AjaxSource (p:ajax, commandButton, commandLink, ...)

So with "component", the ajax events are only sent to the component itself. If it does support inline ajax status, it will be used.
Default should be both IMO

jepsar:
It's cleaner when you can use a single attribute, but it is going to be a breaking change (though it's probably a simple find / replace). I haven't looked into AjaxSource yet, will need to look into that to get a better idea.

tandraschko:
we can deprecate global and remove for v13, not a breaking change then

jepsar:
Yeah let's keep global in 12; `false` can be translated to `none`. Maybe we should simply call it `ajaxStatus` instead of `ajaxEventPropagation`. And I think we should do a PR dealing with this change first. @melloware can I get your opinion on the name of the attribute?

tandraschko:
its actually really event-propagation if you check how ajaxStatus is implemented
ajaxStatus is just a consumer, like event/observer

jepsar:
Ok, let's  call it `eventPropagation` then. So you'll have
```xhtml
<p:ajax eventPropagation=...
```
instead of
```xhtml
<p:ajax ajaxEventPropagation=...
```

tandraschko:
i like this
just "eventPropagation" still sounds weird on p:commandButton - but IMO it was a bad decision that we built in p:ajax into commandButton itself

jepsar:
`disableOnAjax` now defaults to `true`. Removed `inlineAjaxStatus` attribute (active by default).

melloware (approved):

melloware:
This looks good and all integration tests are passing which is also good!

melloware:
Do we need something in `whatsnew.md` just to let people know this is here?

jasonex7:
> `disableOnAjax` now defaults to `true`. Removed `inlineAjaxStatus` attribute (active by default).

Hi @jepsar, I think that default value of `disableOnAjax` should be on the documentation of commandButton and commandLink because is still saying is false.

jepsar:
Good point. Must have overlooked it. Will fix it.

.. done https://github.com/primefaces/primefaces/pull/9163

PR #8315, opened 2022-01-22
Fix #8314: CLI PrimeFlex migration fixes

### LINKED ISSUE #8314: PrimeFaces CLI: PrimeFlex 2 -> PrimeFlex 3 migration issues
**Describe the defect**
The [PrimeFaces CLI 11.0.0](https://mvnrepository.com/artifact/org.primefaces/primefaces-cli/11.0.0) does not migrate the following _PrimeFlex 2_ to **PrimeFlex 3 CSS classes**:
```css
/* Flexbox - Flex Direction */
p-dir-col     -> flex-column
p-dir-col-rev -> flex-column-reverse
p-dir-row     -> flex-row
p-dir-rev     -> flex-row-reverse

/* Flexbox - Align Items */
p-align-start    -> align-items-start
p-align-end      -> align-items-end
p-align-center   -> align-items-center
p-align-baseline -> align-items-baseline
p-align-stretch  -> align-items-stretch

/* Flexbox - Align Self */
p-col-align-start    -> align-self-start
p-col-align-end      -> align-self-end
p-col-align-center   -> align-self-center
p-col-align-baseline -> align-self-baseline
p-col-align-stretch  -> align-self-stretch

/* Grid - Grid System */
p-offset-[0-12] -> col-offset-[0-12]

/* Typography - Whitespace */
p-text-nowrap -> white-space-nowrap
```

Also, it is accidentally migrating the following PrimeNG component selectors, **which do not exist in PrimeNG**
```html
<p-colorPicker></p-colorPicker>
<!-- is migrated into wrong selector  -->
<colorPicker></colorPicker>

<p-fieldset></p-fieldset>
<!-- is migrated into wrong selector -->
<fieldset></fieldset>

<p-columnFilter></p-columnFilter>
<!-- is migrated into wrong selector -->
<columnFilter><columnFilter>
```

**Environment:**
 - Angular version: _13.1.3_
 - PrimeNG version: _13.1.0_
 -  PrimeFaces CLI: _11.0.0_
 - Affected browsers: _ALL_

**To Reproduce**
Download Freya PrimeNG Layout **freya-ng-13.0.0.zip** at https://www.primefaces.org/store/templates.xhtml
and run
```powershell
Invoke-WebRequest -Uri https://repo1.maven.org/maven2/org/primefaces/primefaces-cli/11.0.0/primefaces-cli-11.0.0.jar -OutFile pfcli.jar
java -cp "pfcli.jar" org.primefaces.cli.migration.primeflex.PrimeFlexMigration .\src\app -e="scss,css,html,ts"
```
https://github.com/primefaces/primefaces/blob/11.0.0/primefaces-cli/README.md

**Expected behavior**
The PrimeFaces CLI 11.0.0: PrimeFlex 2 -> PrimeFlex 3 migration should migrate the above-listed PrimeFlex 2 CSS classes to PrimeFlex 3 and should not migrate the listed PrimeNG component selectors.

melloware:
Added unit tests proving the issue and fixed them

### DIFF
--- a/primefaces-cli/src/main/java/org/primefaces/cli/migration/primeflex/PrimeFlexMigration.java
+++ b/primefaces-cli/src/main/java/org/primefaces/cli/migration/primeflex/PrimeFlexMigration.java
@@ -57,11 +57,25 @@ public class PrimeFlexMigration extends AbstractPrimeMigration implements Runnab
         replaceRegex.put("p-col-([0-9]+?)", "col-$1");
         replaceRegex.put("p-col-fixed", "col-fixed");
         replaceRegex.put("p-col-offset-([0-9]+?)", "col-offset-$1");
+        replaceRegex.put("p-offset-([0-9]+?)", "col-offset-$1");
         replaceRegex.put("p-(xl|lg|md|sm)-offset-([0-9]+?)", "$1:col-offset-$2");
-        replaceRegex.put("p-col", "col");
         replaceRegex.put("p-(xl|lg|md|sm)-([0-9]+?)", "$1:col-$2");
 
-//        replaceRegex.put("_xxxPxxx_", "p");
+        replaceRegex.put("p-col-order-(first|0)", "flex-order-0");
+        replaceRegex.put("p-col-order-(1|2)", "flex-order-1");
+        replaceRegex.put("p-col-order-(3|4)", "flex-order-2");
+        replaceRegex.put("p-col-order-(5|6)", "flex-order-3");
+        replaceRegex.put("p-col-order-(7|8)", "flex-order-4");
+        replaceRegex.put("p-col-order-(9|10)", "flex-order-5");
+        replaceRegex.put("p-col-order-(11|12|last)", "flex-order-6");
+
+        replaceRegex.put("p-(xl|lg|md|sm)-order-(first|0)", "$1:flex-order-0");
+        replaceRegex.put("p-(xl|lg|md|sm)-order-(1|2)", "$1:flex-order-1");
+        replaceRegex.put("p-(xl|lg|md|sm)-order-(3|4)", "$1:flex-order-2");
+        replaceRegex.put("p-(xl|lg|md|sm)-order-(5|6)", "$1:flex-order-3");
+        replaceRegex.put("p-(xl|lg|md|sm)-order-(7|8)", "$1:flex-order-4");
+        replaceRegex.put("p-(xl|lg|md|sm)-order-(9|10)", "$1:flex-order-5");
+        replaceRegex.put("p-(xl|lg|md|sm)-order-(11|12|last)", "$1:flex-order-6");
 
         // display
         replaceRegex.put("p-d-none", "hidden");
@@ -79,19 +93,18 @@ public class PrimeFlexMigration extends AbstractPrimeMigration implements Runnab
         replaceRegex.put("p-d-(xl|lg|md|sm)-inline-flex", "$1:inline-flex");
 
         // flexbox
-        replaceRegex.put("p-flex-row", "flex-row");
-        replaceRegex.put("p-flex-column", "flex-column");
-        replaceRegex.put("p-flex-row-reverse", "flex-row-reverse");
-        replaceRegex.put("p-flex-column-reverse", "flex-column-reverse");
+        replaceRegex.put("p-flex-row|p-dir-row", "flex-row");
+        replaceRegex.put("p-flex-column|p-dir-col", "flex-column");
+        replaceRegex.put("p-flex-row-reverse|p-dir-rev", "flex-row-reverse");
+        replaceRegex.put("p-flex-column-reverse|p-dir-col-rev", "flex-column-reverse");
 
         replaceRegex.put("p-flex-(xl|lg|md|sm)-row", "$1:flex-row");
         replaceRegex.put("p-flex-(xl|lg|md|sm)-column", "$1:flex-column");
         replaceRegex.put("p-flex-(xl|lg|md|sm)-row-reverse", "$1:flex-row-reverse");
         replaceRegex.put("p-flex-(xl|lg|md|sm)-column-reverse", "$1:flex-column-reverse");
 
-        replaceRegex.put("p-order-([0-9])", "flex-order-$1");
-
-        replaceRegex.put("p-order-(xl|lg|md|sm)-([0-6])", "$1:flex-order");
+        replaceRegex.put("p-order-([0-6])", "flex-order-$1");
+        replaceRegex.put("p-order-(xl|lg|md|sm)-([0-6])", "$1:flex-order-$2");
 
         replaceRegex.put("p-flex-nowrap", "flex-nowrap");
         replaceRegex.put("p-flex-wrap", "flex-wrap");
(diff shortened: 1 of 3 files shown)


primefaces-cli/src/main/java/org/primefaces/cli/migration/primeflex/PrimeFlexMigration.java
```diff
@@ -57,8 +57,9 @@ protected void initReplaceRegEx() {
         replaceRegex.put("p-col-([0-9]+?)", "col-$1");
         replaceRegex.put("p-col-fixed", "col-fixed");
         replaceRegex.put("p-col-offset-([0-9]+?)", "col-offset-$1");
+        replaceRegex.put("p-offset-([0-9]+?)", "col-offset-$1");
         replaceRegex.put("p-(xl|lg|md|sm)-offset-([0-9]+?)", "$1:col-offset-$2");
-        replaceRegex.put("p-col", "col");
+        replaceRegex.put("p-col\s", "col\s");
```
jepsar: What about `class="whatever p-col"`?

melloware: ahhh crap your are right.  I was always assuming it was first or in the middle not on the end.

Any suggestions?

melloware: Maybe `(\s|\")` do a capture group of space or `"` ?

melloware:
OK this seems to fix it...

```java
replaceRegex.put("p-col(\s|\")", "col$1");
```

primefaces-cli/src/main/java/org/primefaces/cli/migration/primeflex/PrimeFlexMigration.java
jepsar: Something like
```js
"<p-fieldset class=\"p-fieldset\"".replace(/([^<])p-fieldset/, '$1fieldset')
```
(JavaScript)

primefaces-cli/src/main/java/org/primefaces/cli/migration/primeflex/PrimeFlexMigration.java
```diff
@@ -57,8 +57,9 @@ protected void initReplaceRegEx() {
         replaceRegex.put("p-col-([0-9]+?)", "col-$1");
         replaceRegex.put("p-col-fixed", "col-fixed");
         replaceRegex.put("p-col-offset-([0-9]+?)", "col-offset-$1");
+        replaceRegex.put("p-offset-([0-9]+?)", "col-offset-$1");
         replaceRegex.put("p-(xl|lg|md|sm)-offset-([0-9]+?)", "$1:col-offset-$2");
-        replaceRegex.put("p-col", "col");
+        replaceRegex.put("p-col(\s|\")", "col$1");
```
jepsar: What about `class='p-col'`?

melloware: Good catch fixed single quote as well and added test case

melloware:
I don't understand why I am getting this on GitHub Actions when it compiles and works fine for me locally.

```
Error:  COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
Error:  /home/runner/work/primefaces/primefaces/primefaces-cli/src/main/java/org/primefaces/cli/migration/primeflex/PrimeFlexMigration.java:[62,34] illegal escape character
Error:  /home/runner/work/primefaces/primefaces/primefaces-cli/src/main/java/org/primefaces/cli/migration/primeflex/PrimeFlexMigration.java:[195,36] illegal escape character
```

jepsar:
Isn't it better to try to skip `<p-xxx` and `</p-xxx`? What if one would have `styleClass="p-col#{condtion ? ' p-class' : ''}"`?

ThoSap:
> Isn't it better to try to skip `<p-xxx` and `</p-xxx`? What if one would have `styleClass="p-col#{condtion ? ' p-class' : ''}"`?

Yes that this is better, but the regex must not only exclude occurrences starting with `<`  as `p-col` does not _only_ exist in HTML element opening and closing tags in PrimeNG.
The term `p-col` can also exist in CSS/SCSS/TS files without `<` as my migration yesterday showed.

For some examples see
https://github.com/primefaces/primefaces/issues/8314#issuecomment-1019341626

Some files you could test against:
https://github.com/primefaces/primeng/blob/13.1.0/src/app/components/colorpicker/colorpicker.css
https://github.com/primefaces/primeng/blob/13.1.0/src/app/components/colorpicker/colorpicker.ts#L16-L32
https://github.com/primefaces/primeng/blob/13.1.0/src/app/components/colorpicker/colorpicker.spec.ts

https://github.com/primefaces/primeng/blob/13.1.0/src/app/components/table/table.css#L245-L311
https://github.com/primefaces/primeng/blob/13.1.0/src/app/components/table/table.ts#L4300-L4343
https://github.com/primefaces/primeng/blob/13.1.0/src/app/components/table/table.spec.ts

https://github.com/primefaces/primeng/blob/13.1.0/src/app/components/fieldset/fieldset.css
https://github.com/primefaces/primeng/blob/13.1.0/src/app/components/fieldset/fieldset.ts#L11-L87
https://github.com/primefaces/primeng/blob/13.1.0/src/app/components/fieldset/fieldset.spec.ts#L28-L48

Occurrences with `p-col` and `p-col-12` which should still work after a regex change:
https://github.com/primefaces/primeng/blob/13.1.0/src/app/components/dataview/dataview.ts#L31
https://github.com/primefaces/primeng/blob/13.1.0/src/app/components/table/table.spec.ts#L236-L249

ThoSap:
Can't we use something like this with a negative lookahead?
For `p-col` https://regexr.com/6drum
For `p-field` https://regexr.com/6drvb

Lucky for us PrimeFlex 2.0.0 `p-col` and `p-field` subclasses always used a **-** inbetween and not a \w character (example `p-col-offset-7` instead of `p-coloffset-7`) 😆

primefaces-cli/src/main/java/org/primefaces/cli/migration/primeflex/PrimeFlexMigration.java
```diff
@@ -133,6 +143,11 @@ protected void initReplaceRegEx() {
         replaceRegex.put("p-as-center", "align-self-center");
         replaceRegex.put("p-as-baseline", "align-self-baseline");
         replaceRegex.put("p-as-stretch", "align-self-stretch");
+        replaceRegex.put("p-col-align-start", "align-self-start");
```
ThoSap: Regarding my last comment https://github.com/primefaces/primefaces/pull/8315/files#issuecomment-1019457534, then you probably have to change the order of the specific `p-col-*` map entries to be first and the new negative lookahead regex must be last.

Otherwise for example `p-col-align-start` will be migrated to `col-align-start` instead of `align-self-start`.

primefaces-cli/src/main/java/org/primefaces/cli/migration/primeflex/PrimeFlexMigration.java
```diff
@@ -176,7 +192,7 @@ protected void initReplaceRegEx() {
         replaceRegex.put("p-shadow-(21|22|23)", "shadow-8");
 
         // form layout
-        replaceRegex.put("p-field", "field");
+        replaceRegex.put("p-field(\s|'|\")", "field$1");
         replaceRegex.put("p-formgrid", "formgrid");
         replaceRegex.put("p-formgroup-inline", "formgroup-inline");
         replaceRegex.put("p-field-checkbox", "field-checkbox");
```
ThoSap, line 194: To be consistent the same goes for specific `p-field-*` classes, even though the migration here would be correct (`p-field-checkbox` gets `field-checkbox`).

melloware:
@ThoSap I changed the `order` the way you suggested as well as switch to your two lookahead regexes they work like a charm!

BTW I use Regex 101 I like it a lot better only because it has a "Code Generator" option so it generated me the Java version of your regex. See https://regex101.com/r/lbCIQ9/1

Appreciate all your help on this!  I converted a PrimeReact project no problem but it was using a newer version of PrimeFlex 2.0.0 so I never ran into some of these issues but this will be a huge help to the whole PRIME community!

jepsar (approved):

ThoSap (changes requested):

primefaces-cli/src/main/java/org/primefaces/cli/migration/primeflex/PrimeFlexMigration.java
melloware: I moved both lookahead regexes to the very bottom so they are checked last.

PR #8316, opened 2022-01-23
Fix #8122: AutoComplete: Add inlineAjaxStatus

### LINKED ISSUE #8122: AutoComplete: implement inline AJAX status

jepsar:
Fix #8122

This will require a fix in the themes to set the `right` position of the load icon. I'll create a new issue when this is merged.

autocomplete

Again, the GIF is slower than real time, but it'll give you an idea.


jepsar:
First fix and merge #8302

jepsar:
Will need a followup theme issue to set the relative right position of the spinner. Will open this issue when this is merged.

melloware (approved):

PR #8321, opened 2022-01-24
Fix #8219: DatePicker issue with pasting mask

### LINKED ISSUE #8319: Datepicker: Change event is not called when pasting date and using mask="true"
**Describe the defect**
Additional issue in the datepicker like in issue #7176, where the change event is not fired when mask="true" and the date is **entered**, the change event is also not fired when mask="true" and a date is **pasted**. So when one pastes a date an leave the Datepicker input field (blur), the date is cleared. Also the date is not selected in the popout.

**Environment:**
 - PF Version: 10.0 and 11.0
 - JSF + version: MyFaces 2.2.12
 - Affected browsers: At least Firefox and Chrome

**To Reproduce**
1. Copy a correct formatted date (eg. in de _01.01.2022_) from somewhere to clipboard
2. Paste this content from clipboard into a datepicker with mask="true"
3. See that date is not selected in popout of the datepicker
4. Blur datepicker
5. See that pasted content was cleared

**Expected behavior**
When a correct formatted date is pasted into datepicker input field with mask="true", date is selected in the popout and pasted date stays when datepicker looses focus.

**Example XHTML**
```html
<h:form id="frmTest">
    <div class="p-grid" style="max-width:30em">
        <div class="p-col-6">
            Field for enter and copy something
        </div>
        <div class="p-col-6">
            <p:inputText />
        </div>
        <div class="p-col-6">
            Datepicker without masking
        </div>
        <div class="p-col-6">
            <p:datePicker mask="false" />
        </div>
        <div class="p-col-6">
            Datepicker with masking
        </div>
        <div class="p-col-6">
            <p:datePicker mask="true" />
        </div>
    </div>
</h:form>
```
[Datepicker.zip](https://github.com/primefaces/primefaces/files/7926628/Datepicker.zip)

melloware:


melloware:
This was the recommended workaround on InputMask site: https://github.com/RobinHerbots/Inputmask/issues/2431

jepsar:
Is this a workaround we need to remove when the 3rd party issue is fixed?

melloware:
Possibly. I subscribed to that issue so I can be notified if and when it gets fixed.

jepsar:
Maybe add a TODO in the code?

melloware:
TODO added good idea.

PR #8322, opened 2022-01-24
Fix #7901: Toggleswitch icons

### LINKED ISSUE #7901: ToggleSwitch: On/off icon
Is it possible to add icons like this?

ToggleSwitch

melloware:
Just a draft as it doesn't look quite correct yet.


jepsar:
After adding `line-height: 1.4em; text-align: center;` it's looking somewhat better. Obviously the text color needs to be changed in the theme and `content: ""` needs to be removed. But in general I must say that the circle is too small to contain icons. At normal text size the icon looks cramped. I feel it needs some padding. This could work if something like `size="large"` (like in `Avatar`) would be added.

melloware:
Mert this just needs the theme work mentioned by @jepsar above to be merged.

melloware:
I would actually like to merge this and open a Theme ticket like we have done for others?

jepsar:
> I would actually like to merge this and open a Theme ticket like we have done for others?

It does not break anything if you don't use this feature, so I think it's okay. Maybe add a note in what's new?

melloware:
Whats new added

Issue #8337, opened 2022-01-26
Tree: Improve node focus styling


jepsar:
Select the Saga theme both in PrimeFaces and PrimeVue:

https://www.primefaces.org/showcase/ui/data/tree/selection.xhtml
https://primefaces.org/primevue/showcase/#/tree/selection

The PrimeVue node focus styling is much clearer:

<img width="382" alt="Screenshot 2022-01-26 at 17 20 04" src="https://user-images.githubusercontent.com/7500178/151203226-ad3ed729-3ba3-46aa-a629-05fb39709ac5.png">

vs PrimeFaces:

<img width="298" alt="Screenshot 2022-01-26 at 17 23 48" src="https://user-images.githubusercontent.com/7500178/151203537-761a93f5-61cb-4a3c-b842-2262c4d813d1.png">

mertsincan:
Hi @jepsar,

I have added the necessary styles related to the theme. For a behavior like Primevue, it is necessary to make changes in the core. Even with this change, the theme will add proper styles.

jepsar:
Hi @mertsincan,

That's why I've created this PR https://github.com/primefaces/primefaces/pull/8350

PR #8350, opened 2022-01-29
Fix #8339: Tree: Transfer state classes and events from label to content

### LINKED ISSUE #8339: Tree: ui-treenode-outline and ui-state-highlight are applied on label instead of tree node content
This makes styling harder. See also #8337

jepsar:
Fix #8339

This is preparation for #8337 (which is a theme issue)


mertsincan:
Hi @jepsar, 

I made an improvement on the theme in #8337. If you've tested the PR on all Tree options, I think you can merge it. Please let me know if there is a problem with the Theme after you merged it.

melloware (approved):

PR #8354, opened 2022-01-30
Fix #8335: Captcha theme aware

### LINKED ISSUE #8335: Captcha: Make theme / prefers dark aware
Screenshot 2022-01-26 at 15 26 54

https://github.com/primefaces/primefaces/blob/18ae753d683fb3cc4c4dc45ecb5aaccce2ea8b79/primefaces/src/main/java/org/primefaces/component/captcha/CaptchaRenderer.java#L92

melloware:
Will default to user's OS theme preference for Captcha theme color.  Can still be properly set with the `theme` attribute this just gives a better default

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/captcha/captcha.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/captcha/captcha.js
@@ -3,7 +3,7 @@
  *
  * Captcha is a form validation component based on Recaptcha API V2.
  *
- * @typedef {"light" | "dark"} PrimeFaces.widget.Captcha.Theme Captcha features light and dark modes for theme.
+ * @typedef {"auto" | "light" | "dark"} PrimeFaces.widget.Captcha.Theme Captcha features light and dark modes for theme.
  *
  * @interface {PrimeFaces.widget.CaptchaCfg} cfg The configuration for the {@link  Captcha| Captcha widget}.
  * You can access this configuration via {@link PrimeFaces.widget.BaseWidget.cfg|BaseWidget.cfg}. Please note that this
@@ -31,6 +31,8 @@ PrimeFaces.widget.Captcha = PrimeFaces.widget.BaseWidget.extend({
     init: function(cfg) {
         this._super(cfg);
         this.cfg.language = this.cfg.language||'en';
+        this.cfg.theme = this.cfg.theme || PrimeFaces.env.getThemeContrast();
+
         var $this = this;
 
         window[this.getInitCallbackName()] = function() {
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.env.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.env.js
@@ -26,6 +26,16 @@ if (!PrimeFaces.env) {
          * @type {string}
          */
         browser : null,
+        /**
+         * `true` if the user's current OS setting prefers dark mode, `false` otherwise.
+         * @type {boolean}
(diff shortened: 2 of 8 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/captcha/captcha.js
```diff
@@ -31,6 +31,7 @@ PrimeFaces.widget.Captcha = PrimeFaces.widget.BaseWidget.extend({
     init: function(cfg) {
         this._super(cfg);
         this.cfg.language = this.cfg.language||'en';
+        this.cfg.theme = this.cfg.theme || PrimeFaces.env.getPreferredColorScheme();
```
jepsar: This would work in combination with `pe:lightSwitch`, but I was planning to consider the current theme to make this more bulletproof. I think you can do theme is Arya, Vela or ends with "-dark" or "-dim". See for example https://primefaces.org/diamond/javax.faces.resource/theme.css.xhtml?ln=primefaces-diamond-blue-dark

melloware: Ahh good idea.

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.env.js
```diff
@@ -70,6 +99,26 @@ if (!PrimeFaces.env) {
...
+        },
+
+       /**
+         * Based on the current PrimeFaces theme determine if light or dark contrast is being applied.
+         * 
+         * @return {string} either 'dark' or 'light'
+         */
+        getThemeContrast: function() {
+            var theme = PrimeFaces.env.getTheme();
+            var darkRegex = /(arya|vela|\S+-dim|\S+-dark)/gm;
```
jepsar: Shouldn't this be
```suggestion
            var darkRegex = /(^(arya|vela|.+-(dim|dark))$)/gm;
```

or simply
```
return (/(^(arya|vela|.+-(dim|dark))$)/gm).test(PrimeFaces.env.getTheme()) ? 'dark' : 'light';
```

melloware: Updated the regex.  I like leaving the separate lines it makes it easier to step through with the debugger than on one long line.

jepsar: Okay. In Chrome you can have multiple breakpoint on single line I think. I agree this is more readable.

primefaces/src/main/resources/META-INF/resources/primefaces/captcha/captcha.js
```diff
@@ -31,6 +31,7 @@ PrimeFaces.widget.Captcha = PrimeFaces.widget.BaseWidget.extend({
     init: function(cfg) {
         this._super(cfg);
         this.cfg.language = this.cfg.language||'en';
+        this.cfg.theme = this.cfg.theme || PrimeFaces.env.getThemeContrast();
```
jepsar, line 34: The only thing I was doubting if we shouldn't do something like
```js
if (this.cfg.theme === 'auto') {
    this.cfg.theme = PrimeFaces.env.getThemeContrast();
}
```

melloware: Great idea!

melloware:
Done and I made `auto` the new default and updated the migration guide and taglib.

jepsar (approved):

Issue #5037, opened 2019-07-31
DatePicker: Add range selection support with time


pouwelsjochem:
## 1) Environment
- PrimeFaces version: 7.0.5
- Does it work on the newest released PrimeFaces version? No
- Does it work on the newest sources in GitHub? No
- Application server + version: WildFly 17
- Affected browsers: All

## 2) Expected behavior
When selectionMode = 'range' and showTime = 'true' there should be two timepickers visible to be able to select a time selection within a day or over multiple days.

## 3) Actual behavior
There is only one timepicker which sets the time on both start and end date. I can't think of any practical usecase why someone would like to do that. So I can think the new behaviour can be the default behaviour.

## 4) Steps to reproduce
Create the sample XHTML

## 5) Sample XHTML
```
<p:datePicker selectionMode="range" showTime="true"/>
<p:datePicker selectionMode="range" timeOnly="true"/>
```

## 6) Sample bean
No bean required

kukel:
I would like to work on this, but what is the original source of the p:datePicker? I have some other enhancements too so the range can be used in a dataTable as a custom filter as well.

melloware:
@kukel  DatePicker is 100% built from scratch by PrimeTek and used in PrimeNG, PrimReact, PrimeFaces etc.  So there is no original source its just 0-datepicker.js.

kukel:
ok, thanks....

melloware:
Also recently brought up on Stack Overflow: https://stackoverflow.com/questions/67989856/primefaces-datepicker-range-selection-with-time/67989910

jepsar:
To be honest, I have no idea on how the UI should look like for this feature. I don't think we can easily recycle the single date time picker.

melloware:
None of the other PrimeLibs do this functionality and I agree with @jepsar that this UI would not be very friendly.

melloware:
Moved to:https://github.com/primefaces/primefaces/issues/5037

PR #8357, opened 2022-01-31
Fix #8355: DatePicker: Add option to convert to Date / LocalDateTime from a date only pattern

### LINKED ISSUE #8351: DatePicker: Time is not set to start / end of day when used as range filter
**Describe the defect**
A _DatePicker_ is used as custom filter in a _DataTable_'s timestamp column. The corresponding entity attribute is from type _java.util.Date_. When choosing a date range only the entries of the matching end date with a time value of _00:00:00_ are included. All other entries of the matching end date with a later time are ignored.

**Reproducer**
See [primefaces-test-master.zip](https://github.com/primefaces/primefaces/files/7963580/primefaces-test-master.zip)
_**HINT: Start with profile**_ `-Pmojarra23`!

**Environment:**
 - PF Version: _11.0.0_
 - PF Theme: _ALL_
 - JSF + version: _Mojarra 2.3_
 - Affected browsers: _ALL_

**To Reproduce**
1. Start reproducer
2. Filter column 'localDate' by yesterday / yesterday&today / today -> works!
3. Reset filter 'localDate'
4. Filter column 'Date' by yesterday / yesterday&today / today -> **From today only the entry with time value _00:00:00_ is inlucded!**
5. Reset filter 'Date'
6. Filter column 'localDateTime' by yesterday / yesterday&today / today -> **does not work at all -> why?**

**Expected behavior**
All entries with any time value of the selected end date should be included.

### LINKED ISSUE #8355: DataTable: Date / LocalDateTime filter requires a converter
class java.time.LocalDate cannot be cast to class java.time.chrono.ChronoLocalDateTime

Relates to #8351

jepsar:


jepsar:
Can anyone please have a look at the integration tests? I'm getting
```
[ERROR] Failures: 
[ERROR]   DataTable026Test.testFilterIn:291->assertEmployeeRows:391->assertEmployeeRows:397 expected: <3> but was: <8>
[ERROR] Errors: 
[ERROR]   DataTable026Test.testFilterBetweenDate:348->assertEmployeeRows:390 » StaleElementReference
[ERROR]   DataTable026Test.testFilterBetweenLocalDateTime:324->assertEmployeeRows:390 » StaleElementReference
```
Failure might be a macOS issue, as I did not touch any of the `in` code.

I don't get the `StaleElementReference`. How can I fix that? @christophs78 Do you know what's going on there?

jepsar:
Hmm.. `testFilterIn` is passing for me after switching to master. Strange.

christophs78:
> Can anyone please have a look at the integration tests? I'm getting
> 
> ```
> [ERROR] Failures: 
> [ERROR]   DataTable026Test.testFilterIn:291->assertEmployeeRows:391->assertEmployeeRows:397 expected: <3> but was: <8>
> [ERROR] Errors: 
> [ERROR]   DataTable026Test.testFilterBetweenDate:348->assertEmployeeRows:390 » StaleElementReference
> [ERROR]   DataTable026Test.testFilterBetweenLocalDateTime:324->assertEmployeeRows:390 » StaleElementReference
> ```
> 
> Failure might be a macOS issue, as I did not touch any of the `in` code.
> 
> I don't get the `StaleElementReference`. How can I fix that? @christophs78 Do you know what's going on there?

Some integrationtests are not 100% perfect regarding stability. Just re-run-them.
(And yes as almost always there is potential to improve our integrationtests. In terms of qualitity and quantitiy.)

christophs78:
Stop. :-) You are working on DataTable026Test. Need to have a second look.

christophs78:
One important thing ist to keep AJAX in mind. Filtering ist done server-side. So we need to wait until the AJAX-operation which does server-side filtering ist completed. 

This is usally done by `PrimeSelenium.guardAjax`. This guard looks waits until the AJAX-operation ist done.
image

Without this we´ll run into timing-issues and flickering tests. Means sometime the test passes, sometimes it fails.

jepsar:
> One important thing ist to keep AJAX in mind. Filtering ist done server-side. So we need to wait until the AJAX-operation which does server-side filtering ist completed.
> 
> This is usally done by `PrimeSelenium.guardAjax`. This guard looks waits until the AJAX-operation ist done. image
> 
> Without this we´ll run into timing-issues and flickering tests. Means sometime the test passes, sometimes it fails.

Thanks!

I was going to revert that. Although it was working for me locally. What fixed the issues with range was
```
        // Setting range via Selenium (sendKeys) causes JS-errors. So we don´t check for them.
//        assertConfiguration(dataTable.getWidgetConfiguration());
```

now need to figure out what causes `in` to fail

jepsar:
Oh.. it passes.. I don't get it. It was still failing locally.

christophs78:
> Oh.. it passes.. I don't get it. It was still failing locally.

There still must be some timing issue. It´s essential to add `guardAjax` to exact the one operation which triggers the Ajax-Request. Maybe cross-check browser-dev-tool (network-tab) against the integrationtest and step it through line for line.

primefaces/src/main/java/org/primefaces/model/filter/BetweenFilterConstraint.java
```diff
@@ -43,8 +47,28 @@ public boolean isMatching(FacesContext ctxt, Object value, Object filter, Locale
...
     protected boolean isBetween(Comparable value, List filter) {
-        Comparable start = (Comparable) filter.get(0);
-        Comparable end = (Comparable) filter.get(1);
+        Comparable<?> start = (Comparable) filter.get(0);
+        Comparable<?> end = (Comparable) filter.get(1);
+        if (start == null || end == null) {
+            return false;
+        }
+        if (end instanceof LocalDate) {
+            LocalDate date = toLocalDate(value);
```
tandraschko: wouldnt it just also work if we do "value= toLocalDate(value);" and remove the second statement in the IF?

tandraschko: also i wonder if we would need to do this for numbers?

jepsar: > wouldnt it just also work if we do "value= toLocalDate(value);" and remove the second statement in the IF?

I'll have a look. I don't really like overwriting parameters. We could eliminate a return statement though.

jepsar: > also i wonder if we would need to do this for numbers?

I think you are right. Same as the fix for lt, lte, gt, gte.

tandraschko: If you dont like overwriting, we can create a new method like convertValue and assign this to a new var "convertedValue"

jepsar: That's what I was planning to do ;-) I'll continue on this tonight.

primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datatable/DataTable026Test.java
```diff
@@ -302,6 +303,58 @@ public void testFilterIn(Page page) {
...
+        List<Employee> employeesFiltered = employees.stream()
+                .filter(e -> {
+                    LocalDate date = e.getLastLoginDateTime().toLocalDate();
+                    return date.equals(start) || date.equals(end) || (date.isAfter(start) && date.isBefore(end));
+                })
+                .collect(Collectors.toList());
+        assertEmployeeRows(dataTable, employeesFiltered);
+
+        // Setting range via Selenium (sendKeys) causes JS-errors. So we don´t check for them.
+//        assertConfiguration(dataTable.getWidgetConfiguration());
```
melloware: Should we investigate this?

melloware (changes requested):

primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datatable/DataTable026Test.java
jepsar: It was already like that for the between check of `LocalDate`s. But I can have a look what I get logged in my JS console.

primefaces/src/main/java/org/primefaces/model/filter/BetweenFilterConstraint.java
jepsar: > also i wonder if we would need to do this for numbers?

The only thing is that we don't have a component that supports a number range. Slider comes close but it does not hold a value. We could split a string so you can enter "0 / 10" or "0 - 10"

primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datatable/DataTable026Test.java
jepsar: @melloware I've create a new ticket for this: #8362

melloware: Excellent I will mark this convo resolved

melloware (approved):

primefaces/src/main/java/org/primefaces/model/filter/BetweenFilterConstraint.java
ltlBeBoy: I was thinking about a number range input these days, too. A text field with a separator is a start. But there are two downsides:

1. Each entered number triggers the filter event.
2. The filter field in the header requires the a double size of the column. (Linebreak in an input field is not possible)

So I was thinking about a component that works like the _DatePicker_:
- On click into the filter field a popup opens and shows two _InputNumber_ fields. On hiding the popup the filter will be triggered.
- The filter field could be a hidden input (`<input ` is requirement for filters). The displayed value could be in a `span`/`div` with a linebreak at the separator and a reduced font size.

melloware:
@jepsar pull my latest chnages into this PR it fixes the JS error.

primefaces/src/main/java/org/primefaces/model/filter/BetweenFilterConstraint.java
jepsar: Numbers will work if you'd write a converter which produces a list of the numeric type equaling the value type.

primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datatable/DataTable026Test.java
```diff
@@ -302,6 +301,58 @@ public void testFilterIn(Page page) {
...
+        List<Employee> employeesFiltered = employees.stream()
+                .filter(e -> {
+                    LocalDate date = CalendarUtils.convertDate2LocalDate(e.getLastLoginDate());
+                    return date.equals(start) || date.equals(end) || (date.isAfter(start) && date.isBefore(end));
+                })
+                .collect(Collectors.toList());
+        assertEmployeeRows(dataTable, employeesFiltered);
+
+        // Setting range via Selenium (sendKeys) causes JS-errors. So we don´t check for them.
+//        assertConfiguration(dataTable.getWidgetConfiguration());
```
melloware: Safe to uncomment this now

primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datatable/DataTable026Test.java
```diff
@@ -302,6 +301,58 @@ public void testFilterIn(Page page) {
...
+        List<Employee> employeesFiltered = employees.stream()
+                .filter(e -> {
+                    LocalDate date = e.getLastLoginDateTime().toLocalDate();
+                    return date.equals(start) || date.equals(end) || (date.isAfter(start) && date.isBefore(end));
+                })
+                .collect(Collectors.toList());
+        assertEmployeeRows(dataTable, employeesFiltered);
+
+        // Setting range via Selenium (sendKeys) causes JS-errors. So we don´t check for them.
+//        assertConfiguration(dataTable.getWidgetConfiguration());
```
melloware: Safe to uncomment this now

melloware (changes requested):

melloware (approved):

jepsar:
I tried with `DateTimeConverter` first, but it complained it could not convert to `LocalDateTime` (probably due to missing time).

ltlBeBoy:
@jepsar Please keep in mind that the time part of [LocalDateTime](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html) is represented to nanosecond precision. So I would recommend to do

`endDate = endDate.plusDays(1).minus(1, ChronoUnit.`**NANOS**`);`

instead of

`endDate = endDate.plusDays(1).minus(1, ChronoUnit.MILLIS);`

in `DatePickerRenderer#getConvertedValue(...)`.

I tested the modifications with `LazyDataModel`, and it works as expected now. Thank you very much!

Rapster:
Wouldn't be easier to convert into epoch time and compare from there?

jepsar:
Checked `Date` conversion with nano use and it's the same as with millis, so we can use the same conversion for both `Date` and `LocalDateTime`.

tandraschko:
i thought we have internal converts in p:datepicker?
would be great that @Rapster and @christophs78 could also review

jepsar:
Yes, there are internal converts, and that's a bit of the issue, you cannot force to set them. The converter is chosen based on the bound type. With a filter no value is bound. I do think the date picker code could use a bit of DRYing.

christophs78:
Yes, i agree DatePicker would benefit from refactoring. There´s a good (or not so good) amount of legacy, due to the old Calendar-component which shares some code,  support for java.util.Date and java.time, tons of features and design-decisions years ago.
Internal refactoring is IMO welcome when it´s backed up by a good unit- and/or integration-tests. For DatePicker we built https://github.com/primefaces/primefaces/blob/master/primefaces/src/test/java/org/primefaces/component/datepicker/DatePickerTest.java during our work on PF 8 and adding java.time - support. And there are some integration-tests.
Changing our API by removing internal converters or something like this is IMO no option. This would break too many existing usecases without clear benefit for long-term PrimeFaces-users. (IMO PrimeFaces has it´s position as long-term stable plattform which needs to avoid scaring  users by to much breaking changes. For some of our customer-base maybe already the step from PF8 to PF10/11 has been too much.)

jepsar:
I don't think anyone implied that we should remove the internal converters (could be done without breaking anything though). I just wanted to point out that automatic, internal conversion was not possible if you use the date picker as a filter without value (type) binding.

tandraschko:
should we maybe introduce something like "type=java...LocalDateTime" on datePicker, for a fallback without value-binding?
but i also see, that this PR follows more how it should be handled with JSF

maybe DatePicker should remove its internal conversion code and reuse those converters automatically?

jepsar:
That was also my idea, we could just call converters from the date picker code. I didn't want to touch the existing code that much though in this PR, as the focus was to set the end time of a (local) date range.

tandraschko:
yep
then lets create a new issue that DatePicker should remove its conversion code and automatically determine the right Converter to use

we should make the converters more global and dont call it Filter*Converter

jepsar:
Can't we rename these two converters and take care of the rest later? I was struggling finding a name for them. Basically they convert a date without time to `j.u.Date` or `LocalDateTime`. Maybe `MidnightDate` / `MidnightLocalDateTime`?

tandraschko:
yes
that way my idea
however, im also not sure if we need those converters
doesnt f:convertDateTime doesnt work in this case?

jepsar:
That was the first thing I tried, but it complained about time not being available. I'll give it a second look.

Rapster:
Few things has been added in JSF 2.3 regarding java.time, see https://omnifaces-fans.blogspot.com/2015/12/jsf-23-align-to-new-data-and-time.html, maybe that's why?

jepsar:
When I try
```xhtml
<p:datePicker selectionMode="range" onchange="PF('wgtTable').filter()" id="lastLoginDateTimeFilter"
              pattern="yyyy-MM-dd" rangeSeparator="-">
    <f:convertDateTime pattern="yyyy-MM-dd" type="localDateTime" />
</p:datePicker>
```
I'm getting
> Text '2021-02-09' could not be parsed: Unable to obtain LocalDateTime from TemporalAccessor: {},ISO resolved to 2021-02-09 of type java.time.format.Parsed

Rapster:
It should be localDate, or pattern should have a loalDateTime format (e.g ISO 8601) no?

jepsar:
If you would simply look at it from a date picker's perspective, then yes, `LocalDate` would be the logical type. But, it is used as a filter, so the type should match the type of the data that is being filtered. Hence I introduced two converters for common date types `j.u.Date` and `LocalDateTime`. That is the cleanest solution IMO. Earlier I converted in the filter https://github.com/primefaces/primefaces/pull/8357/files/c34194f0ca33711a02101237a525d6294214f0c8#diff-8d338d8057680452668deb6c0727bc9c8dbe58fbde541ec6a4bb8f4065456208

tandraschko:
im really totally +1 that filters should not convert and a input should have a attached converter (normally this works OOTB because of the assigned value-expression and Converter=forType but we dont have a value expression in this case)

im just not sure if we should provide a own converters here - or is the JSF 2.3 convertDateTime not that configurable enough?

jepsar:
I cannot get the standard JSF converters to work. See https://github.com/primefaces/primefaces/pull/8357#issuecomment-1043397674 this can be tested with IT 026.

tandraschko:
is this something wrong in the JSF impl or not compatible by spec?

ltlBeBoy:
Why don't you just change the output type of the `DatePicker` as followed **in case there is no value expression**:

1. `selectionMode="single / multiple"` & `showTime="false"` **->** `LocalDate`
2. `selectionMode="range"` & `showTime="false"` **->** `LocalDateTime` with time values `00:00:00.000000` - `23:59:59.999999`
3. `selectionMode="single / multiple"` & `showTime="true"` **->** `LocalDateTime`
4. `selectionMode="range"` & `showTime="true"` **->** `LocalDateTime` with time values as selected by user


And for case `2.` if the datamodel field is from type `LocalDate`, then the converter with `type="localDate"` can be specified:
```
<p:datePicker selectionMode="range" onchange="PF('wgtTable').filter()" id="lastLoginDateTimeFilter"
              rangeSeparator="-">
    <f:convertDateTime type="localDate" />
</p:datePicker>
```

ltlBeBoy:
OK, it was late and converters are just for `String` <> `Object` conversion...
Then `type` / `outputType` as suggested by @tandraschko [here](https://github.com/primefaces/primefaces/pull/8357#issuecomment-1041450895)

jepsar:
JSF spec does not help here. It does not say that if type is `localDateTime` that the pattern should also contain a time pattern. By that way, I previously tested with Mojarra. MyFaces does not work either. There I'm getting:

> '2021-02-09' could not be understood as a date and time. Example: 2022-02-20

Also, when I remove the `pattern` from the date picker, but do have a `pattern` on the converter, the date picker will fallback to a default format. Meaning you will need to have a duplicate pattern... Maybe worth opening a separate issue for.

I'll be renaming the converters for now.

tandraschko:
hmmmm
can we maybe improve the JSF converters here or extend it?

i mean, your converters are working and ok but we should NOT introduce own converter when a more generic converter is actually available
also currently they can only be used with DatePicker - whereas the f:convertTimeDate can be used on all components

jepsar:
@tandraschko You're right, now it's much cleaner.

primefaces/src/main/java/org/primefaces/convert/DateTimeConverter.java
```diff
@@ -25,16 +25,60 @@
...
+    }
+
+    public Object getAsObject(FacesContext context, DatePicker datePicker, String value) {
+        if (value == null) {
+            return null;
+        }
+        String type = getType();
+        boolean isDate = "date".equals(type);
+        boolean isLocalDateTime = "localDateTime".equals(type);
+        if ("range".equals(datePicker.getSelectionMode()) && (isDate || isLocalDateTime)) {
```
jepsar: We can probably remove the range check...

ltlBeBoy: Then you would reset the selected time to the start of day in case `showTime="true"`.

jepsar: Range and show time are two different things. But we should probably look at `showTime="false"` and else fallback to super.

ltlBeBoy: @jepsar No, they are not always different things. You can currently actually set `selectionMode="range"` and `showTime="true"` even if there is an issue (see #5037). And you can also use the `DatePicker` as filter with `type="date"` for `DateTimeConverter`, set `selectionMode="single"` and `showTime="true"`.

jepsar: +1 for `selectionMode="single"`. I'll check for that and add an IT.

tandraschko:
looks good to me now
@christophs78 can you review? especially the cheanges in DatePicker itself

christophs78:
> looks good to me now @christophs78 can you review? especially the cheanges in DatePicker itself

yes, will look at this in-depth at weekend

christophs78:
@jepsar - First of all many thanks for your work on this issue during the last month!

Green flag for everything except `org.primefaces.convert.DateTimeConverter`. For `org.primefaces.convert.DateTimeConverter` i´d like to ask some questions to ensure we don´t break something.

1. `org.primefaces.convert.DateTimeConverter#getAsObject` kicks in for `<f:convertDateTime type="localDateTime"/>` or `<f:convertDateTime type="date"/>` when used in combination with `p:DatePicker`?
2. For this two usage-scenarios we change the JSF-default behaviour as suggested by @tandraschko.

Is there the chance/risk of breaking other existing usage-scenarios where  `<f:convertDateTime type="localDateTime"/>` or `<f:convertDateTime type="date"/>` is used for non filter-between-scenarios?
What happens for something like `<f:convertDateTime type="localDateTime" pattern="some pattern"/>` Do we loose `pattern`-attribute? `DatePickerBase#calculatePattern` IMO only takes pattern-attribute of `p:DatePicker` into account. Not the one of `f:convertDateTime`.

jepsar:
> Is there the chance/risk of breaking other existing usage-scenarios where `<f:convertDateTime type="localDateTime"/>` or `<f:convertDateTime type="date"/>` is used for non filter-between-scenarios?

It shouldn't, but indeed we need to be sure of that. I'd like to add some ITs which include time zone, and support `ZonedDateTime`. Also some ITs to compare internal conversion with `f:convertDateTime` will be helpful. It wil also help cleaning up the date picker code.

> What happens for something like `<f:convertDateTime type="localDateTime" pattern="some pattern"/>` Do we loose pattern-attribute? `DatePickerBase#calculatePattern` IMO only takes pattern-attribute of `p:DatePicker` into account. Not the one of `f:convertDateTime`.

Indeed, date picker only looks at its own pattern. I think we should have a look at this while cleaning up the date picker code.

tandraschko:
about the pattern.... 
in theory we could also do something like this in the `getAsObject`:
```
if (this.getPattern == null) {
   this.setPattern(datePicker.getPattern());
}
return super.getAsObject()?
```

jepsar:
I'll override `getPattern` now for future use. Currently the date picker does not work properly without a pattern (unless you are using a default).

... wait, it should be the other way around. If the pattern of the date picker is empty, it should look at the converter (if set) and otherwise fallback to a default.

tandraschko:
will you also work on the topic from your last comment?

jepsar:
No, I think that should be done when we revise the date picker code. If there are no further comments, I hope this is finally it for this PR. 😅

jepsar:
@christophs78 I've added a test to compare internal conversion vs `f:convertDateTime`. I think this change is safe now, WDYT?

christophs78:
I´ll try to look at it next weekend. :-) Currently way to bussy on my day-job.

Currently not totally sure what we should do with all the other attributes of `f:convertDateTime` (see https://jakarta.ee/specifications/faces/3.0/jakarta-faces-3.0.html#a4697). I personally don´t use them, i even don´t use `f:convertDateTime` but can´t tell for other people...

jepsar:
As long as you don't make them conflict with what you've set on the date picker, you are good to go

melloware (approved):

christophs78 (approved):

christophs78:
The newly added and/or extended integrationtests fail on Firefox (see IT):

```
Error:  Failures: 
Error:    DataTable026Test.testFilterIn:293->assertEmployeeRows:449->assertEmployeeRows:455 expected: <3> but was: <8>
Error:  Errors: 
Error:    DataTable026Test.testFilterEquals:129 » ElementNotInteractable Element <input ...
Error:    DataTable026Test.testFilterGte:112 » ElementNotInteractable Element <input id=...
Error:    DataTable026Test.testFilterGt:95 » ElementNotInteractable Element <input id="f...
Error:    DataTable026Test.testFilterLte:78 » ElementNotInteractable Element <input id="...
Error:    DataTable026Test.testFilterLt:61 » ElementNotInteractable Element <input id="f...
```

melloware:
Its because of this...

```xml
            <style>
                html .ui-datatable table { table-layout: auto; }
                td, th { white-space: nowrap; max-width: 60px; font-size: .75rem; }
            </style>
```

In firefox the `max-width` is causing the column to be too small and its saying the Filter is not selectable.  If I increase `max-width` to `120` those tests pass and others fail with...

```
org.openqa.selenium.interactions.MoveTargetOutOfBoundsException: (1503, 79) is out of bounds of viewport width (1268) and height (909)
```

jepsar:
Oh, wow! I just added that styling to have it a bit more manageable when deployed. I'll remove it.

melloware:
We can try removing it but then my local tests failed but I didn't try the headless tests which use 1920 width screen

jepsar:
Removing it entirely failed. I had a look this morning, and it is really weird that with Firefox the content width seems to matter. We could break this up in two tests as a workaround.

melloware:
Breaking it up might be the way to go.

Issue #8159, opened 2021-12-09
DataTable filterBy/sortBy breaks when column ordering changes


markpalermo:
**Describe the defect**
When the column ordering has been changed on a DataTable, searching and filtering no longer function.
It would appear that the search/filter column name passed to the backing bean is no longer correct, likely using the column index to lookup the column name based on the initial ordering, rather than the updated ordering.

**Environment:**
Primefaces Showcase v11.0.0 RC2

**To Reproduce**
Reproducible in the Primefaces Showcase  DataTable Dynamic Columns page:
https://www.primefaces.org/showcase/ui/data/datatable/columns.xhtml

1. Go to https://www.primefaces.org/showcase/ui/data/datatable/columns.xhtml
2. Try filtering by "activity", filtering works.
3. Clear the "activity" filtering.
4. Move the "activity" column to be the first column, and click "Update" to refresh the table.
5. Try filtering or sorting any column
6. Note that neither filtering nor sorting function properly.

**Expected behavior**
Search and Sort should work after column ordering has changed

**Example XHTML**
```<div class="card">
    Available columns are <b>name</b>, <b>country</b>, <b>date</b>, <b>status</b> and <b>activity</b>.

    <h:form id="form">
        <div class="p-mt-3 p-mb-5">
            <p:inputText id="template" value="#{dtColumnsView.columnTemplate}" size="50" styleClass="p-mr-2"/>
            <p:commandButton update="customers" action="#{dtColumnsView.updateColumns}" value="Update"
                             process="@parent" icon="pi pi-refresh"
                             oncomplete="PF('customersTable').clearFilters()"/>
        </div>

        <p:dataTable id="customers" var="customer" value="#{dtColumnsView.customers}" widgetVar="customersTable"
                     filteredValue="#{dtColumnsView.filteredCustomers}">

            <p:columns value="#{dtColumnsView.columns}" var="column" headerText="#{column.header}" field="#{column.property}">
                <f:facet name="filter">
                    <p:datePicker selectionMode="range" onchange="PF('customersTable').filter()" rendered="#{column.type eq 'date'}"/>
                    <p:selectOneMenu onchange="PF('customersTable').filter()" rendered="#{column.type eq 'enum'}">
                        <f:selectItem itemValue="#{null}" itemLabel=""/>
                        <f:selectItems value="#{column.klazz.enumConstants}" />
                    </p:selectOneMenu>
                </f:facet>
            </p:columns>

        </p:dataTable>
    </h:form>
[...]

PR #8358, opened 2022-01-31
#8159 Fix Columns not sortable and filterable after reordering/adding

fcjm:
#8159. For further information please see my comment under the issue.


tandraschko:
as i said in the issue
your issue is a different new one, please provide a reproducer / a forked primefaces-test

fcjm:
I will try to create a reproducer and issue tomorrow. In the meanwhile I have a question. 

You closed #8159 by suggesting to reset the values of filterByAsMap/sortByAsMap to "null". The example-bean and .xhtml provided by the author were pretty much identical to the showcase "DataTable - Dynamic Columns", meaning that the bug also currently exists in this showcase. Does that mean that the Dynamic-Columns showcase has to be adjusted as well?

melloware:
@fcjm yep please submit a PR for showcase fix

PR #8367, opened 2022-02-01
Fix #8363: FilterConstraint: Don't convert values for each row

### LINKED ISSUE #8246: Datatable: filter on Integer uses String comparison

### LINKED ISSUE #8363: FilterConstraint: Don't convert values for each row
Now values are converted for each row, examples:

https://github.com/primefaces/primefaces/blob/a910479dd4febb60d99123f100265c5647b02ef7/primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java#L42

https://github.com/primefaces/primefaces/blob/a910479dd4febb60d99123f100265c5647b02ef7/primefaces/src/main/java/org/primefaces/model/filter/StringFilterConstraint.java#L39

with the id we can cache converted values as a request attribute

jepsar:
Fix #8363

### DIFF
--- a/docs/12_0_0/components/datatable.md
+++ b/docs/12_0_0/components/datatable.md
@@ -331,6 +331,11 @@ When a custom component is used as a filter facet, filtering needs to be called 
 preferred event such as `onchange="PF('carsTable').filter()"`. Also defining a converter might be
 necessary if the value of the filter facet is not defined.
 
+Please make sure that the filter is using the **same type as the column field** if you are using comparable
+filter match modes (like greater than). For example, if the column field is an integer, and you would like to
+add a greater than filter, make sure to convert the filter to integer as well. Do so by adding a `f:converter`
+(see example below).
+
 ```xhtml
 <p:dataTable id="dataTable" var="car" value="#{tableBean.carsSmall}" widgetVar="carsTable" filteredValue="#{tableBean.filteredCars}">
 
--- a/primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java
+++ b/primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java
@@ -37,7 +37,7 @@ public abstract class ComparableFilterConstraint implements FilterConstraint {
         if (!(value instanceof Comparable)) {
             throw new IllegalArgumentException("Value should be a java.lang.Comparable");
         }
-        if (!value.getClass().isAssignableFrom(filter.getClass())) {
+        if (!filter.getClass().isAssignableFrom(value.getClass())) {
             throw new IllegalArgumentException("Filter cannot be casted to value type. Forgot to add a converter?");
         }
         return getPredicate().test((Comparable) value, (Comparable) filter);
--- a/primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java
+++ b/primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java
@@ -34,20 +34,18 @@ public abstract class ComparableFilterConstraint implements FilterConstraint {
         if (value == null || filter == null) {
             return false;
         }
-        if (value.getClass() == filter.getClass() && value instanceof Comparable) {
-            return getPredicate().test((Comparable) value, (Comparable) filter);
+        if (!(value instanceof Comparable)) {
+            throw new IllegalArgumentException("Value should be a java.lang.Comparable");
         }
-        if (value instanceof Number) {
-            if (!(filter instanceof Number)) {
-                throw new IllegalArgumentException("Filter should be a number. Forgot to add f:converter?");
-            }
-            return getPredicate().test(((Number) value).doubleValue(), ((Number) filter).doubleValue());
+        if (!filter.getClass().isAssignableFrom(value.getClass())) {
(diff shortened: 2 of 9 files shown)


tandraschko:
does this also fix #8355 if we expect a converter?

jepsar:
> does this also fix #8355 if we expect a converter?

Not yet, I was planning to rework PR #8357 to use converters. Also for a date / local date time range (list) there is no converter available, so I was planning to create them.

docs/12_0_0/components/datatable.md
```diff
@@ -331,6 +331,11 @@ When a custom component is used as a filter facet, filtering needs to be called
 preferred event such as `onchange="PF('carsTable').filter()"`. Also defining a converter might be
 necessary if the value of the filter facet is not defined.
 
+Please make sure that the filter is using the **same type as the column field** if you are using comparable
```
jepsar, line 334: @melloware can you please check the wording? If okay, I'll add this paragraph to all versions.

primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java
```diff
@@ -38,9 +34,14 @@ public boolean isMatching(FacesContext ctxt, Object value, Object filter, Locale
         if (value == null || filter == null) {
             return false;
         }
+        if (value.getClass() == filter.getClass() && value instanceof Comparable) {
+            return getPredicate().test((Comparable) value, (Comparable) filter);
+        }
         if (value instanceof Number) {
-            BigDecimal filterBigDecimal = toBigDecimal(filter, locale);
```
tandraschko, line 42: should we do this check in general, not just for Number?
if (value.class not assignable to filter.class)?

jepsar: I do on line 37
```java
        if (value.getClass() == filter.getClass() && value instanceof Comparable) {
            return getPredicate().test((Comparable) value, (Comparable) filter);
        }
```
I could change it to an assignable check

jepsar: Ah.. to show an exception you mean. Yes!

docs/12_0_0/components/datatable.md
melloware: Looks good to me but maybe make comparable like `java.lang.Comparable` just so it is clear.

primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java
```diff
@@ -38,35 +34,18 @@ public boolean isMatching(FacesContext ctxt, Object value, Object filter, Locale
...
-        if (object instanceof String) {
-            try {
-                DecimalFormat decimalFormat = (DecimalFormat) NumberFormat.getInstance(locale);
-                decimalFormat.setParseBigDecimal(true);
-                return (BigDecimal) decimalFormat.parseObject((String) object);
-            }
-            catch (ParseException e) {
-                return null;
-            }
+        if (value instanceof String) {
```
jepsar: I was wondering if we cannot just drop the special string handling.

Rapster: I'm trying to figure out something about that whole thing... Somehow I feel we should always fallback on String comparison if value.class != filter.class. ~~But yeah, I think we should drop it, specially that it makes exact filter irrelevant since lowercase is used~~ (nvm I'm mixing up with my own changes 😄 )

Rapster (approved):

primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java
jepsar: We would have the fallback on line 48, only without to lowercase conversion (which I think is fine to leave out in this case).

melloware (approved):

primefaces/src/main/java/org/primefaces/model/filter/ComparableFilterConstraint.java
Rapster: Yes, I was trying if using an alphanumeric comparator would work but doesn't work too well in some cases... I wanted to avoid setting a converter in basic cases like numbers. So using converters make sense whenever you're using lt, gt etc and most probably dates?

jepsar: Yeah, see also #8355

PR #8378, opened 2022-02-02
Fix #5503: SelectOneRadio: custom layout keyboard navigation

### LINKED ISSUE #5503: SelectOneRadio: with custom layout keyboard navigation does not work properly
## 1) Environment
- PrimeFaces version: Showcase ->PrimeFaces-8.0-SNAPSHOT
- Affected browsers: FF, Chrome, ...

## 2) Expected behavior
In custom layout the radios are reachable by the keyboard tab as with other layout.

## 3) Actual behavior
After a radio in the custom layout is selected, the keyboard tab is ignoring the radios of this group.

## 4) Steps to reproduce
Got to the showcase: https://www.primefaces.org/showcase/ui/input/oneRadio.xhtml
1. Click on the radio at 'Responsive' - Amsterdam
1. Use the tab key to go to the following radios of the 'Custom Layout'
1. As long as you do not select any of thus radios it is working fine.
1. Select one of the radios of the custom layout
1. After selected one of the radios, 'Custom Layout' radios are ignored on tab key access.

## 5) Sample XHTML
Not provided. See showcase: https://www.primefaces.org/showcase/ui/input/oneRadio.xhtml

## 6) Sample bean
Not provided. See showcase: https://www.primefaces.org/showcase/ui/input/oneRadio.xhtml

jepsar:
Fix #5503

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectoneradio.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectoneradio.js
@@ -36,7 +36,7 @@ PrimeFaces.widget.SelectOneRadio = PrimeFaces.widget.BaseWidget.extend({
         //custom layout
         if(this.cfg.custom) {
             this.facet = this.jq.attr('role') === 'radiogroup';
-            this.originalInputs = this.jq.find((this.facet ? '.ui-helper-hidden ' : '') + ':radio');
+            this.originalInputs = this.jq.find((this.facet ? '.ui-helper-custom ' : '') + ':radio');
             this.inputs = $('input:radio[name="' + this.id + '"].ui-radio-clone');
             this.outputs = this.inputs.parent().next('.ui-radiobutton-box');
             this.labels = $();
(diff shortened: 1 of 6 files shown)


jepsar:
Needs work. Fix works with Chrome but fails with Safari and Firefox.

tandraschko:
the regex stuff feels bit weird :D
let me try to understand:
the original input in the showcase is: `j_idt312:customRadio:0` and `j_idt312:opt1` the mapped one
and we need to delegate the focus event from `j_idt312:customRadio:0` to the mapped input `j_idt312:opt1`

shouldnt it be possible without regex? something like:

```
        $(this.originalInputs).on('focus.selectOneRadio', function() {
            var originalInput = this;
            var index = $this.originalInputs.eq(originalInput);
            var mappedInput = $('#' + $this.id + PrimeFaces.settings.separatorChar + index + '_clone');
            mappedInput.trigger('focus');
        });
```

We would just need to add the separatorChar to the HeadRender

jepsar:
I was wondering about the separator char. Getting index like that is better :D. Only thing is that, if no item is selected, and you Tab to the component, the first element needs to be selected, and if you Shift+Tab to it, the last element must be selected. If there is a selection (hard to check currently, style on sibling of parent of the radio), you need to select that.

I think I need a totally different approach though as Firefox and Safari are handling focus on radio buttons in a different way. Need to figure out what is going on there first.

jepsar:
... and I'm considering to have a second iteration of improving custom layout which I hope makes this issue easier to handle. Need to think this through first. Will come back to this later.

jepsar:
The issue is worse, keyboard navigation on regular select one radio's doesn't work at all with Safari.

melloware:
Yikes!

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectoneradio.js
```diff
@@ -174,6 +174,21 @@ PrimeFaces.widget.SelectOneRadio = PrimeFaces.widget.BaseWidget.extend({
             e.preventDefault();
         });
 
+        // for custom radio buttons, transfer the focus from original to the matching custom button
+        $(this.originalInputs).on('focus.selectOneRadio', function() {
+            // 1 = JSF separator, 2 = index
+            var matches = this.id.match(/^.*([^\d])(\d)+$/);
```
tandraschko: please use seperatorChar instead of regex like discussed

jepsar: I've started over to address the Safari issue first. I won't be using the regex.

jepsar:
I took a step back and started over and came up with a way simpler fix which works with Safari, Firefox and Chrome.

What caused the weird behavior of first being able to select a button and no longer once a button was selected was caused by part of the buttons in the radio group were hidden. The only thing that had to change was having a distinct group name for the custom buttons, so that they won't interfere with the original, hidden radio buttons.

melloware (approved):

melloware:
Awesome fix!

PR #8384, opened 2022-02-03
Fix #8376: Datatable/Treetable return value onRowRightClick

### LINKED ISSUE #8376: Datatable: ContextMenu is duplicated by the native one when contextMenu behavior is present
onRowRightClick (https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js#L2451) should return a value, as is expected by https://github.com/primefaces/primefaces/blob/97b0180f488e109b3dc898b9920fe8cca661df88/primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js#L1276
The return value might be this.hasBehavior('contextMenu')

**Environment:**
 - PF Version: _11.0.0_
 - Affected browsers: ALL_

**Expected behavior**
Only the component context menu should be shown, not also the browser native one.

melloware:


mihaiDumea:
A minor mismatch between documentation and usage return value variable name: isContextMenuDelayed vs 'true if context menu was displayed'.

melloware:
OK I updated the doc can you check it.

mihaiDumea:
Perfect.

PR #8407, opened 2022-02-07
fix DataTable034Test

christophs78:
cc @tandraschko

### DIFF
--- a/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datatable/DataTable034Test.java
+++ b/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datatable/DataTable034Test.java
@@ -43,6 +43,9 @@ public class DataTable034Test extends AbstractDataTableTest {
 
         Assertions.assertEquals("Language 1", it.getValue());
 
+        it.clear();
+
+        it = PrimeSelenium.createFragment(InputText.class, By.id("form:datatable:0:name"));
         it.setValue("Language 666");
 
         it = PrimeSelenium.createFragment(InputText.class, By.id("form:datatable:0:name"));


primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datatable/DataTable034Test.java
```diff
@@ -43,6 +43,9 @@ public void test(Page page) throws InterruptedException {
 
         Assertions.assertEquals("Language 1", it.getValue());
 
+        it.clear();
```
christophs78, line 46: i call clear explizit because this already triggers the first ajax-refresh. after this we need to call `createFragment` again.

melloware (approved):

melloware:
I was looking at that too and trying to fix it "inside" the clear method but couldn't.   I agree with your way we have to recreate it after it gets cleared.

christophs78:
> I was looking at that too and trying to fix it "inside" the clear method but couldn't. I agree with your way we have to recreate it after it gets cleared.

IMO it´s a somewhat exotic usecase.

tandraschko:
I wonder why this happens anyway
I thought i proxy everything and always lookup the webelement

melloware:
Is somehow creating it like this different than when doing it as an injected field?

tandraschko:
thanks @christophs78 

i really dont know whats wrong here
i also made EVERYTHING lazy by

```
    @Override
    public WebElement getWrappedElement() {
        return PrimePageFragmentFactory.create(WebElement.class, getElementLocator());
    }

    public WebElement getRoot() {
        return getWrappedElement();
    }
```

in AbstractPrimePageFragment
so everything call to the WebElement triggers a new findElement and a stale MUST not happen actually

tandraschko:
nvm it would works, even without this change above, i will commit a small enhancement in the test now

it just doesnt work without the clear

tandraschko:
commited now

melloware:
Excellent !  You found a "hidden" bug.  Nice fix.

PR #8415, opened 2022-02-08
Fix #8414: monitorDownload for non-ajax exporter

### LINKED ISSUE #8414: Exporter: monitorDownload not functional for non-ajax request
Follow up for #3108.
Not tested with PF 11, as our application is not compatible with that version. But the source code is the same for PF 10 and 11 for this issue and the provided fix definitively fixes that issue.

**Describe the defect**
`PrimeFaces.downloadMonitor` is not functional for non-ajax request as triggert by `p:commandButton ajax="false"`.

**Reproducer**
No reproducer, as it is already "confirmed" by @melloware in https://github.com/primefaces/primefaces/issues/3108#issuecomment-861913021 ;)

**Environment:**
 - PF Version: _10.0.0_
 - Affected browsers: _ALL_

**Expected behavior**
`PrimeFaces.downloadMonitor` should be working for ajax and non-ajax request

**Proposed fix**
- File to be changed: https://github.com/primefaces/primefaces/blob/c198e1f9101f4ce7dbcfcd26015b04a41bbcced0/primefaces/src/main/java/org/primefaces/component/export/DataExporter.java#L235
- Use `ResourceUtils.getMonitorKeyCookieName(context, null)` instead of `Constants.DOWNLOAD_COOKIE` as it is already done for the ajax-part of the code https://github.com/primefaces/primefaces/blob/c198e1f9101f4ce7dbcfcd26015b04a41bbcced0/primefaces/src/main/java/org/primefaces/component/export/DataExporter.java#L224

fanste:


tandraschko:
can you also adjust the exporter docs for 12.0?

fanste:
In my opinion the docs are already correct. It already tells that ajaxStatus is not working (which is correct) and that downloadMonitor works as described in the docs for FileDownload. https://github.com/primefaces/primefaces/blob/master/docs/12_0_0/components/dataexporter.md#monitor-status

Or am I wrong?

tandraschko:
it also should use the method, right?

```externalContext.addResponseCookie(Constants.DOWNLOAD_COOKIE, "true", Collections.<String, Object>emptyMap());```

**Comment by fanste** [comment]
Why? It should not use `Constants.DOWNLOAD_COOKIE` directly as [you changed the whole cookie-thing](https://github.com/primefaces/primefaces/commit/3dc7320c34993a0756400c02988488d02a326997#diff-fc3861c44859c2cd5fad5f6d7c1675d8753f67b41661fe1129f7ab4341f40221R87) to include the viewId (refactored many times to the current utility method in `ResourceUtils`). Thats why the cookie name has to be generated. To set the cookie I used the utility function from PrimeFaces which does forward to `externalContext.addResponseCookie` with handling for additional properties.

**Comment by tandraschko** [comment]
the docu currently does NOT use ResourceUtils.getMonitorKeyCookieName

**Review by melloware — APPROVED** [review]

**Comment by fanste** [comment]
You need to be more precise with your requests ;) I did not see, that the `Custom Exporter` example also contained code for the download monitor.

Is this what you expected?

**Comment by fanste** [comment]
@melloware I had to force-push as the branch creation failed and did not copy the v12 docs. Therefore your approval points to a non-existing commit - just in case you want to correct that.

**Comment by melloware** [comment]
1 CheckStyle violation
`src/main/java/org/primefaces/component/export/DataExporter.java:[50,8] (imports) UnusedImports: Unused import - org.primefaces.util.Constants.`

**Comment by tandraschko** [comment]
+1 then

**Comment by fanste** [comment]
Thanks... time for closing if I fail to copy over three simple changes...

I currently cannot run maven for this repo due to network errors with the company proxy.

**Comment by melloware** [comment]
OK i have to look into this now we are getting an unrelated error in IT tests...

```
Password002Test.testFeedbackHide:69 expected: <none> but was: <block>
```

melloware (approved):

PR #8418, opened 2022-02-08
fix Password002Test

christophs78:


tandraschko:
@christophs78 thanks! i just wonder if this should directly be in `hideFeedback()` similar to in `showFeedback`?

melloware:
I agree if other users may get burned by it.

PR #8427, opened 2022-02-10
Fix #8423: DatePicker: with inline=true in Dialog causes position to be fixed

### LINKED ISSUE #8423: DatePicker: with inline=true in Dialog causes position to be fixed
```xhtml
<p:dialog widgetVar="myDialog">
    <p:datePicker inline="true"/>
</p:dialog>
<p:button value="Open" onclick="PF('myDialog').show();return false"/>
```

Screenshot 2022-02-10 at 17 34 18

jepsar:
Fix #8423

@melloware this was my idea


melloware (approved):

jepsar:
Shall I un-draft it?

melloware:
I think this is right.  this would only matter in 1 use case (or 2) and that is this inline calendar or inline datepicker.

All other calls to `PrimeFaces.utils.resolveAppendTo` are correctly determining for a dynamic popup.  This inline is basically a non dynamic popup.  It might be the only one.

PR #8438, opened 2022-02-12
Fix #8330: TreeTable: Swapping nodes with Collections.swap causes NPE

### LINKED ISSUE #8330: TreeTable: Swapping nodes with Collections.swap causes NPE
**Describe the defect**
If you swap items inside TreeNode#getChildren with Collections.swap, it causes an NPE with mojarra when the update="treeID" is triggered. With MyFaces NPE is only  thrown with quarkus for some reason, with jetty it doesn't throw an NPE but it doesn't seem to work either. @FlipWarthog has noticed that this is because TreeNodeChildren#.set clears the parent node  for some reason

```
java.lang.NullPointerException: Cannot invoke "org.primefaces.model.TreeNode.getRowKey()" because the return value of "org.primefaces.model.TreeNode.getParent()" is null
    at org.primefaces.component.treetable.TreeTableRenderer.encodeNode(TreeTableRenderer.java:528)
    at org.primefaces.component.treetable.TreeTableRenderer.encodeNodeChildren(TreeTableRenderer.java:849)
    at org.primefaces.component.treetable.TreeTableRenderer.encodeNodeChildren(TreeTableRenderer.java:836)
    at org.primefaces.component.treetable.TreeTableRenderer.encodeTbody(TreeTableRenderer.java:514)
    at org.primefaces.component.treetable.TreeTableRenderer.encodeRegularMarkup(TreeTableRenderer.java:386)
    at org.primefaces.component.treetable.TreeTableRenderer.encodeMarkup(TreeTableRenderer.java:262)
    at org.primefaces.component.treetable.TreeTableRenderer.render(TreeTableRenderer.java:145)
    at org.primefaces.component.treetable.TreeTableRenderer.encodeEnd(TreeTableRenderer.java:138)
    at javax.faces.component.UIComponentBase.encodeEnd(UIComponentBase.java:649)
    at javax.faces.component.UIComponentBase.encodeAll(UIComponentBase.java:530)
```

**Reproducer**
[primefaces-test.zip](https://github.com/primefaces/primefaces/files/7935585/primefaces-test.zip)

**Environment:**
 - PF Version: _11.0_
 - JSF + version: all profiles on the primefaces-test project
 - Affected browsers: _ALL_

**To Reproduce**
1. Run with mojarra
2. Click on 'Move up' button on the last row

FlipWarthog:


melloware:
The only thing I wonder is if this really does improve performance of sorting?  If this was in the original commit since 2013 we don't have any history of the Google Code ticket that created it. Either way I like this fix.

melloware:
Here was the original blog post announcement: https://www.primefaces.org/treetable-sorting/

FlipWarthog:
Right, so the way I made the change `set` now does what `setSibling` did, so the performance gain (whatever it was) is still there, but it also fixes any case where `set` is called on the collection, e.g. from the Java Collections API.

melloware:
Excellent

melloware (approved):

tandraschko:
can we add a integrationtest?

FlipWarthog:
@tandraschko This change specifically affects TreeTable sorting and there are already 5 sets of integration tests around TreeTable, some of which are testing sorting. Or do you mean tests that specifically test performance?

tandraschko:
i mean a case where Collections.swap does not break it :D

FlipWarthog:
Ahh, got it. Yup, let me add one specifically for this case.

tandraschko:
we can also just add unittests, no need to use integration tests pby

PR #8457, opened 2022-02-18
Fix #8456: Improve SelectCheckboxMenu keyboard controls

### LINKED ISSUE #8456: SelectCheckboxMenu: Keyboard navigation improvements
**Describe the defect**
Our application has recently been tested for accessibility. I am currently reviewing the components we use in terms of keyboard navigation. I think there can be some improvements made to the keyboard navigation of the SelectCheckboxMenu component to make it more accessible. I can also provide a PR for this. The problems I currently have with this component are the following:

- Focus is not reset to the trigger once the Overlay is closed
    - instead it gets set to the start of the page
- The user can tab out of the overlay on both ends (Shift+Tab/Tab)
    - since the overlay is located at the end of the page, the focus will be set there as well

**Reproducer**
The current behaviour can be tested on the showcase. 
https://www.primefaces.org/showcase/ui/input/checkboxMenu.xhtml?jfwid=b16ee

**Environment:**
 - PF Version: _11.0.0_
 - Affected browsers: _ALL_

**Expected behavior**
The focus should be set to the triggering element once the overlay is closed.
Inside the overlay the focus should be trapped.

fcjm:
- set Focus to trigger element after hiding the overlay
- trap the focus inside the overlay


primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
```diff
@@ -625,7 +625,12 @@ PrimeFaces.widget.SelectCheckboxMenu = PrimeFaces.widget.BaseWidget.extend({
...
             var togglerCheckboxInput = this.toggler.find('> div.ui-helper-hidden-accessible > input');
             this.bindCheckboxKeyEvents(togglerCheckboxInput);
-            togglerCheckboxInput.on('keyup.selectCheckboxMenu', function(e) {
+            togglerCheckboxInput.on('keydown.selectCheckboxMenu', function(e) {
+                if(e.which === $.ui.keyCode.TAB && e.shiftKey) {
+                    e.preventDefault();
+                }
+            })
+                .on('keyup.selectCheckboxMenu', function(e) {
                         if(e.which === $.ui.keyCode.SPACE) {
```
melloware: Just a minor styling thing here can you make the next .on line up with the paren above it so they look chained.  You can see other compoennts for examples.

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
```diff
@@ -1044,6 +1063,8 @@ PrimeFaces.widget.SelectCheckboxMenu = PrimeFaces.widget.BaseWidget.extend({
                     $this.postHide();
                 }
             });
+
+            this.keyboardTarget.focus();
```
melloware: Don't call `.focus()` that is deprecated jQuery call `.trigger('focus')` that is the recommended way now

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
```diff
@@ -651,7 +656,21 @@ PrimeFaces.widget.SelectCheckboxMenu = PrimeFaces.widget.BaseWidget.extend({
 
         var itemKeyInputs = this.itemContainer.find('> li > div.ui-chkbox > div.ui-helper-hidden-accessible > input');
         this.bindCheckboxKeyEvents(itemKeyInputs);
-        itemKeyInputs.on('keyup.selectCheckboxMenu', function(e) {
+        itemKeyInputs.on('keydown.selectCheckboxMenu', function(e) {
+            var input = $(this),
+                item = input.closest("li"),
```
melloware: I don't see where you are using this `item` variable anywhere here in the code?

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
```diff
@@ -651,7 +656,21 @@ PrimeFaces.widget.SelectCheckboxMenu = PrimeFaces.widget.BaseWidget.extend({
...
+                index = $this.items.index(item);
+
+            if(e.which === $.ui.keyCode.TAB && !e.shiftKey
+                && index === $this.items.length - 1) {
+                e.preventDefault();
+            }
+            else if(e.which === $.ui.keyCode.TAB && e.shiftKey
+                && !$this.cfg.showHeader && index === 0) {
+                e.preventDefault();
+            }
```
melloware, line 673: Slightly more readable might be...
```js
if (e.which === $.ui.keyCode.TAB) {
    if (!e.shiftKey && index === $this.items.length - 1) {
        e.preventDefault();
    } else if (e.shiftKey && !$this.cfg.showHeader && index === 0) {
        e.preventDefault();
    }
}
```

melloware (changes requested):

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
fcjm: I use this one line below to calculate the index. Should i just compress them into one variable? Effectively i only need the index variable.

melloware: on no you are ok then i didnt see that!

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
fcjm: good point, has been added in my new commit

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
fcjm: has been added

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
fcjm: has been added

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
fcjm: Ah i did not see that response, I now in my new commit compressed the three lines into one. Is it okay that way?

melloware (approved):

melloware:
@tandraschko i am going to merge this as well and then I am going to do some more refactoring on SelectCheckBoxmenu as there is some more code we can clean up.  Then after that I am hoping to eventually get to the ticket for creating the Selenium tests for this.

PR #8462, opened 2022-02-18
Fix #8460: SelectOneRadio aria-checked attribute

### LINKED ISSUE #8460: SelectOneRadio: ARIA issues
**Describe the defect**

In the [W3C Validator](https://validator.w3.org/nu) validate the SelectOneRadio Showcase page.

Results in: 
image

melloware:
Fixed the first issue complaining about missing `aria-checked` attribute

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/selectoneradio/SelectOneRadioRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/selectoneradio/SelectOneRadioRenderer.java
@@ -38,11 +38,7 @@ import javax.faces.model.SelectItem;
 import javax.faces.render.Renderer;
 
 import org.primefaces.renderkit.SelectOneRenderer;
-import org.primefaces.util.ComponentUtils;
-import org.primefaces.util.GridLayoutUtils;
-import org.primefaces.util.HTML;
-import org.primefaces.util.LangUtils;
-import org.primefaces.util.WidgetBuilder;
+import org.primefaces.util.*;
 
 public class SelectOneRadioRenderer extends SelectOneRenderer {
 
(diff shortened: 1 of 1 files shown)


jepsar:
Isn't this already fixed by the widget? Or at least, isn't the widget attempting to do so? Will check once I'm at my laptop.

melloware:
Yeah what I am doing is running the latest code in the showcase locally.  Then doing View Source on the page and copying that entire view source to the HTML5 Checker.  It its getting all these violations.

jepsar:
Yeah, see:

https://github.com/primefaces/primefaces/blob/c98a9857c1c3c695ef6112532418e50fff9001f3/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectoneradio.js#L95-L103

Called when setting up custom:

https://github.com/primefaces/primefaces/blob/c98a9857c1c3c695ef6112532418e50fff9001f3/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectoneradio.js#L65

And for regular:

https://github.com/primefaces/primefaces/blob/c98a9857c1c3c695ef6112532418e50fff9001f3/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectoneradio.js#L81-L83

If there is an issue, shouldn't we fix it there?

And shouldn't you use the source of the DOM tree instead of the result of view source? JS is not applied when you simply do view source.

melloware:
You are porbably right but the only way to test the Validator is by pasting in the HTML since we don't have a public showcase snapshot to point it at right?  Unless we have some other way to test it?

melloware:
I will close this PR let's dicusss in the ticket.

jepsar:
I copied the html node from the DOM inspector. The only thing missing will be the doctype.

melloware:
Oh good this is a much better way of doing it.

I think we need to go through all the components and fix these issues.  For example for AutoComplete I am now seeing this....

image

PR #8464, opened 2022-02-19
Fix #8463: Showcase remove script type="javascript"

### LINKED ISSUE #8463: Showcase: Cleanup HTML5 Script Warnings
**Description**
More and more accessibility reports come in using the [Nu HTML Checker](https://validator.w3.org/nu/) and I always use the PF Showcas to test the issues.  However it generates lots of noise like...

image

JUST FOR THE SHOWCASE I would like to filter these out so when diagnosing and HTML validation issues these warnings are not there.

melloware:
This allows us to debug HTML5 issues with much less noise!

### DIFF
--- a/primefaces-showcase/src/main/java/org/primefaces/showcase/application/Html5Context.java
+++ b/primefaces-showcase/src/main/java/org/primefaces/showcase/application/Html5Context.java
@@ -1,0 +1,43 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2021 PrimeTek
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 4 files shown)


mertsincan:
Great! Is it possible to make such an arrangement in core? WDYT? Maybe, a parameter can be added. But, I'm not sure :)

melloware:
We already made this part of PrimeFaces Extensions but I am more than happy to add it to PF core for others to use. We have had it in Extensions since 10.0

melloware:
I submitted another PR adding it to Core and user can simply turn it on if they want to with 1 line in their faces-config.xml

```xml
<faces-context-factory>org.primefaces.application.factory.Html5FacesContextFactory</faces-context-factory>
```

PR #8465, opened 2022-02-20
Fix #8463: Html5FacesContextFactory

### LINKED ISSUE #8463: Showcase: Cleanup HTML5 Script Warnings

melloware:
Moved into PF Core and renamed to include "Faces"

### DIFF
--- a/docs/12_0_0/core/accessibility.md
+++ b/docs/12_0_0/core/accessibility.md
@@ -33,3 +33,18 @@ bundle to provide your own values.
 - primefaces.rowgrouptoggler.aria.ROW_GROUP_TOGGLER = Toggle Row Group
 - primefaces.rowtoggler.aria.ROW_TOGGLER = Toggle Row
 
+## HTML5 Compliance
+
+Along with ARIA standards, PrimeFaces strives to be fully [W3C Validation](https://validator.w3.org/) compliant with the
+HTML5 specification.
+
+Before JSF 4.0 all script links required `type="text/javascript"`. However, according to the HTML5 specification
+this is not necessary as all scripts by default are Javascript.  If running an application on JSF lower than 4.0 you
+can remove the `type="text/javascript"` on all `<script>` links with the following line in your `faces-config.xml`.
+
+```xml
+<factory>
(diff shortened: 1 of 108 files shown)


tandraschko:
if its in core, please add a doc about it and also that its not required anymore in Faces 4.0

melloware:
Ok let me update docs and the Java doc comments in case anyone is reading them. 

Should we detect faces 4.0 and stop writing javascipt out in the few places in PF code where we write it out?

melloware:
Added docs blurb under accessibility.md

melloware:
Also while I was at it fixed all the showcase files. This works fine still on all browsers as we only care about IE11 anyway and it works there

docs/12_0_0/core/accessibility.md
```diff
@@ -33,3 +33,18 @@ bundle to provide your own values.
...
 - primefaces.rowtoggler.aria.ROW_TOGGLER = Toggle Row
 
+## HTML5 Compliance
+
+Along with ARIA standards, PrimeFaces strives to be fully [W3C Validation](https://validator.w3.org/) compliant with the
+HTML5 specification.
+
+Before JSF 4.0 all script links required `type="text/javascript"`. However, according to the HTML5 specification
+this is not necessary as all scripts by default are Javascript.  If running an application on JSF lower than 4.0 you
+can remove the `type="text/javascript"` on all `<script>` links with the following line in your `faces-config.xml`.
```
tandraschko, line 43: i would clarify this a bit, maybe something like:
If running an application on JSF lower than 4.0, the rendered HTML will always contain `type="text/javascript"` for JSF generated script tags. You can remove it this via our custom ResponseWrite via your `faces-config.xml`.

tandraschko:
| Should we detect faces 4.0 and stop writing javascipt out in the few places in PF code where we write it out?

We must detect faces 4 + check the doctype actually
but please create a new issue

melloware:
I will open a new issue

PR #8474, opened 2022-02-21
Fix #8473: SelectOneMenu fix aria-owns

### LINKED ISSUE #8473: SelectOneMenu: Advanced Mode ARIA Issue
**Describe the defect**
When using Advanced Mode in showcase the following ARIA issue is reported:

>  Attribute aria-activedescendant value should either refer to a descendant element, or should be accompanied by attribute aria-owns.

image

See: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role

melloware:


primefaces-showcase/src/main/webapp/WEB-INF/template.xhtml
```diff
@@ -10,13 +10,15 @@
...
     <h:head>
         <f:facet name="first">
-            <script async="async" src="https://www.googletagmanager.com/gtag/js?id=UA-93461466-1"></script>
-            <script>
-                window.dataLayer = window.dataLayer || [];
-                function gtag(){dataLayer.push(arguments);}
-                gtag('js', new Date());
-                gtag('config', 'UA-93461466-1');
-            </script>
+            <ui:fragment rendered="#{request.serverName != 'localhost'}">
```
melloware: I added this rendered so the Google Tag Manager is never loaded when running localhost showcase.

jepsar: Isn't `#{facesContext.application.projectStage ne 'Development'}"` better?

tandraschko: Actually we should only render it on production

melloware: I think the problem is we have the web.xml hardcoded to Production mode that is why I didn't do it.

melloware: And this was only for the showcase for us developers so i thought checking localhost was OK.

jepsar: Are they using a load balancer?

melloware: I have no idea I know they just run Jetty but not sure if they run it clustered or standalone.

melloware:
I reverted that change from this commit for now and maybe we can figure out a better way to do it than look for localhost.  Its not critical.

PR #8490, opened 2022-02-24
Fix #8381: DatePicker improve responsiveness for small screens

### LINKED ISSUE #8381: Datepicker: Panel not responsive
**Describe the defect**
The overlay panel of the datepicker component is not responsive and thus overflows small screens and can't be used to its full extent.

**Environment:**
 - PF Version: _10.0.0_

**To Reproduce**
1. Go to the [PrimeFaces datepicker showcase](http://primefaces.org/showcase/ui/input/datepicker/datePicker.xhtml?jfwid=d28da).
2. Reduce the screen width (e.g. to 320px).
3. Click on one of the datepicker components to open its overlay panel.
4. See error > overlay panel overflows the screen.

**Expected behavior**
The overlay panel on the datepicker component should adapt its width to the given screen width to be responsive. This is important to fulfil [WCAG accessibility success criterion 1.4.10](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html). The responsiveness should be guaranteed no matter what attributes are set (e.g. monthNavigator, yearNavigator and showButtonBar).

melloware:
Basically listen for `window.resize` events and if the screen size is lower than Media Query small which is typically 576px and will auto switch it to Touch UI.

image


primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -2228,6 +2239,16 @@
             }
         },
 
+        updateResponsiveness: function() {
+            if (this.options.autoDetectDisplay && !this.options.inline) {
+                var currentUI = this.options.touchUI;
+                this.options.touchUI = PrimeFaces.env.mobile || window.screen.width < 576;
```
jepsar: Shouldn't we move `window.screen.width < 576` to a utils function?

jepsar: .. and that value should be overridable (as with the first day of the week in the date picker)

melloware:
new property on datepicker `responsiveBreakpoint` that defaults to 576.  Added utility methods to `PrimeFaces.env`

primefaces/src/main/java/org/primefaces/component/datepicker/DatePickerBase.java
```diff
@@ -29,19 +29,22 @@
...
 import org.primefaces.model.datepicker.DateMetadataModel;
+import org.primefaces.util.CalendarUtils;
 
 public abstract class DatePickerBase extends UICalendar implements Widget, InputHolder, MixedClientBehaviorHolder {
 
     public static final String COMPONENT_FAMILY = "org.primefaces.component";
 
     public static final String DEFAULT_RENDERER = "org.primefaces.component.DatePickerRenderer";
 
+    public static final int DEFAULT_RESPONSIVE_BREAKPOINT = 576;
```
jepsar: Shouldn't this be a global default?

jepsar:
I'm wondering what would happen with multiple months. I think that's a separate problem though. Maybe we fix that by adding a horizontal scroll, or ignore multiple months under a certain screen width.

primefaces/src/main/java/org/primefaces/component/datepicker/DatePickerBase.java
melloware: Moved to Constants.java

tandraschko:
we should ask @mertsincan 
and the DEFAULT_RESPONSIVE_BREAKPOINT  in global Constants is weird without description and/or other usages

melloware:
It multi-month doesn't work then can just set `autoDetectDisplay` to `false` for that one.  Although who knows what that looks like on a mobile display now...

melloware:
I assigned to @mertsincan and I agree without a 2nd use case the DEFAULT_RESPONSIVE_BREAKPOINT does seem odd in Constants.  Any other suggestion to put it somewhere?

jepsar:
We don't have a second use now, but anything with a popup could use it. Color picker?

melloware:
For now I called it `RESPONSIVE_BREAKPOINT_SMALL` and Javadoc'ed it and moved it out Constants.java until we have a second use case.

Issue #8520, opened 2022-03-07
CSP: Report only


cnsgithub:
In large web applications it might be difficult to introduce Content Security Policy directly. Instead there might be the requirement that remaining CSP violations should be reported to an adequate endpoint as a start. Then the production system will show up problems regarding CSP even in the last corner of the webapp, driven by the users.

Unfortunately, PrimeFaces does not support the header `Content-Security-Policy-Report-Only` yet.

It would be nice to have additional context params to configure reporting, e.g. `primefaces.CSP_REPORT_ONLY` and primefaces.CSP_REPORT_ENDPOINT'.

tandraschko:
as discussed in the PR, please provide a new PR when you have time.

We should have 4 possible modes:

- CSP off (CSP=false)
- CSP on (CSP=true)
- CSP on with reporting (CSP=true, CSP_POLICY="report-uri /csp-violation-report-endpoint/")
- CSP off but reporting only

so the 4) is the only new mode now
and i think CSP="reportOnly" and CSP_REPORT_ONLY_POLICY="report-uri /csp-violation-report-endpoint/" is fine

melloware:
OK submitted PR basically handling the 4th case ` CSP="reportOnly" and CSP_REPORT_ONLY_POLICY="report-uri /csp-violation-report-endpoint/"`

PR #8521, opened 2022-03-07
add: CSP: Report only #8520

cnsgithub:
```xml
    <context-param>
        <param-name>primefaces.CSP_REPORT_ONLY</param-name>
    </context-param>
    <context-param>
        <param-name>primefaces.CSP_REPORT_ENDPOINT</param-name>
        <param-value>/csp-reporting</param-value>
    </context-param>
```

grafik

### DIFF
--- a/docs/12_0_0/core/contentsecuritypolicy.md
+++ b/docs/12_0_0/core/contentsecuritypolicy.md
@@ -19,6 +19,20 @@ CSP is disabled by default and a global parameter is required to turn it on.
     <param-value>true</param-value>
 </context-param>
 ```
+
+In large web applications you might decide to report CSP violations only as a start instead of directly forcing them. Browsers will report CSP violations to a user-defined HTTP endpoint then.
+
+```xml
+<context-param>
+    <param-name>primefaces.CSP_REPORT_ONLY</param-name>
+    <param-value>true</param-value>
+</context-param>
+<context-param>
+    <param-name>primefaces.CSP_REPORT_ENDPOINT</param-name>
+    <param-value>/csp-reporting</param-value>
+</context-param>
+```
+
 ## Policy
 There are many ways to configure CSP for different levels of security. Currently, PrimeFaces has chosen to
 support the NONCE (number used once) based checking for script evaluation only. Nonce attributes are automatically
--- a/primefaces/src/main/java/org/primefaces/csp/CspPhaseListener.java
+++ b/primefaces/src/main/java/org/primefaces/csp/CspPhaseListener.java
@@ -42,12 +42,18 @@ public class CspPhaseListener implements PhaseListener {
 
     private Lazy<Boolean> enabled;
     private Lazy<String> customPolicy;
+    private Lazy<Boolean> reportOnly;
+    private Lazy<String> reportEndpoint;
 
     public CspPhaseListener() {
         enabled = new Lazy<>(() ->
-                PrimeApplicationContext.getCurrentInstance(FacesContext.getCurrentInstance()).getConfig().isCsp());
+                PrimeApplicationContext.getCurrentInstance(FacesContext.getCurrentInstance()).getConfig().isCspOrCspReportOnly());
         customPolicy = new Lazy<>(() ->
                 PrimeApplicationContext.getCurrentInstance(FacesContext.getCurrentInstance()).getConfig().getCspPolicy());
+        reportOnly = new Lazy<>(() ->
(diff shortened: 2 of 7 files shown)


docs/12_0_0/core/contentsecuritypolicy.md
```diff
@@ -19,6 +19,20 @@ CSP is disabled by default and a global parameter is required to turn it on.
     <param-value>true</param-value>
 </context-param>
 ```
+
+In large web applications you might decide to report CSP violations only as a start instead of directly forcing them. Browsers will report CSP violations to a user-defined HTTP endpoint then.
```
melloware, line 23: Dont forget to also add to /gettingstarted/configuration.md file

primefaces/src/main/java/org/primefaces/csp/CspPhaseListener.java
```diff
@@ -67,7 +73,20 @@ public void beforePhase(PhaseEvent event) {
         CspState state = PrimeFacesContext.getCspState(context);
 
         String policy = LangUtils.isBlank(customPolicy.get()) ? "script-src 'self'" : customPolicy.get();
-        externalContext.addResponseHeader("Content-Security-Policy", policy + " 'nonce-" + state.getNonce() + "'");
+        policy += " 'nonce-" + state.getNonce() + "';";
+
+        if (LangUtils.isNotBlank(reportEndpoint.get())) {
+            // current browser compatibility
+            policy += " report-uri " + reportEndpoint.get() + ";";
+            // TODO forward compatibility via report-to directive and Report-To header left out for the moment to reduce complexity
```
melloware: Remove TODO?

melloware (changes requested):

melloware:
@cnsgithub long time!   Thanks for the PR this looks pretty good to me just a couple of comments.

cnsgithub:
> just a couple of comments.

Done.

Thanks for the quick review, @melloware

melloware (approved):

melloware:
The only other thing you might want to add is `/gettingstarted/whatsnew.md` just to let people know the new options are available.

tandraschko:
IMO we should introduce `CSP_POLICY_REPORT_ONLY` as new param for allowing setting multiple params, not just report-uri

and maybe instead of `CSP_REPORT_ONLY` `true|false`, change `CSP` to support `true|false|reportOnly` ?

melloware:
or we could just use 1 param...

```xml
    <context-param>
        <param-name>primefaces.CSP_REPORT_ONLY</param-name>
        <param-value>/csp-reporting</param-value>
    </context-param>
```

basically if the value for REPORT_ONLY exists its considered report only and we use the value as the endpoint?

tandraschko:
i dont like to introduce 1 param in web.xml for multiple header values (1:N), therefore i would follow `CSP_POLICY`

melloware:
No problem.  I like your idea then of `CSP` being a value.

cnsgithub:
> and maybe instead of `CSP_REPORT_ONLY` `true|false`, change `CSP` to support `true|false|reportOnly` ?

The disadvantage of this would be that CSP cannot be turned on together with reporting?

Where do you want to specify the `report-uri`? In both `CSP_POLICY_REPORT_ONLY ` and `CSP_POLICY`?

tandraschko:
hmmmm, ok we have 4 possible modes now:
- CSP off (`CSP=false`)
- CSP on (`CSP=true`)
- CSP on with reporting (`CSP=true`, `CSP_POLICY="report-uri /csp-violation-report-endpoint/"`)
- CSP off but reporting only

so the 4) is the only new mode now
and i think `CSP="reportOnly"` and `CSP_REPORT_ONLY_POLICY="report-uri /csp-violation-report-endpoint/"` is fine

tandraschko:
any update?

christophs78:
MDN says report-uri is deprecated: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri

cnsgithub:
> MDN says report-uri is deprecated: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri

That's true in theory but not in practice since not all browsers support the new `Report-To` header: https://caniuse.com/?search=report-to

grafik

cnsgithub:
> any update?

Sorry, currently no time, maybe someone else wants to take over, otherwise we would have to wait for a while.

tandraschko:
then i will close it for now

Issue #5866, opened 2020-05-06
dataTable: blockUI stop works after ajax update


frettarenan:
1) Environment
- PrimeFaces version: 7.0
- Application server + version: Spring boot tomcat embed 9
- Affected browsers: All

2) Expected behavior
blockUI should keep working

3) Actual behavior
blockUI stop works after ajax update

4) Steps to reproduce
After loading the page, use the pagination, see that it is working correctly.
But if you create a search, for example, that updates the records with update, when paging again, the blockui stops working.

5) Sample XHTML
```xml
<p:dataTable id="myDataTable"
					value="#{myLazyDataModel}" var="e" lazy="true"
					paginator="true">
<!-- Coluns here -->
</p:dataTable>
<p:blockUI block="#{p:component('myLazyDataModel')}" trigger="#{p:component('myLazyDataModel')}">
				<div class="p-grid ui-g p-align-center vertical-container p-justify-center sigor-blockui-loading">
        			<div class="p-col-fixed">
						<div class="blockui-loading-image ld ld-ring ld-spin" />
						<div class="blockui-loading-text">#{msg.get('blockuiMessage')}</div>
					</div>
				</div>
   			</p:blockUI>
```

6) Considerations
I know why the problem occurs, but I'm not managed to solve.
because when using the <p: ​​blockUI> component, it injects a jquery code, which needs to be re-executed after ajax events.
```javascript
$(document).on('pfAjaxComplete', function() {
     console.log('pfAjaxComplete');
     runOnEveryPartialPageLoad();
});
```
Do you have any suggestions on how do i solve this problem?

melloware:
Please follow the issue template or provide an executable example using the [PrimeFaces Test](https://github.com/primefaces/primefaces-test) project.  It is the only way developers can debug your problem to help.

melloware:
@frettarenan I will close the ticket if a reproducer is not created.

frettarenan:
Thank you very much for the support, I managed to solve the problem in another way.
```xml
<p:ajax event="page" onstart="showBlockUI('#{p:component('baseEstadoDataTable')}', '#{msg.get('blockuiMessage')}')" oncomplete="hideBlockUI('#{p:component('baseEstadoDataTable')}')" />
[...]

PR #8544, opened 2022-03-09
Fix #8494: BlockUI do not automatically destroy widget

### LINKED ISSUE #8494: BlockUI: widgetVar is reset to null after partial updates
I am using Primefaces 11, and I am adding this components in home.xhtml page:

```html
<p:dataTable id="dataTable" var="n" value="#{homeBean.dataMedel}" lazy="true" paginator="true" rows="10"
             paginatorTemplate="{CurrentPageReport} {FirstPageLink} {PreviousPageLink} {NextPageLink} {RowsPerPageDropdown}" 
             sortOrder="descending" rowsPerPageTemplate="10,15,20,30,50">
```
```html
<p:blockUI widgetVar="bui" block="dataTable" animate="false">
     <p:graphicImage value="ajax_loader.gif"/>
</p:blockUI>
```
```html
<h:selectOneMenu id="filterTopic" value="#{filterBean.topic}" label="topic" title="topic"
    <p:ajax event="change" update="dataTable" onstart="PF('bui').show();" oncomplete="PF('bui').hide()"
            listener="#{homeBean.onPreRender}"/>
    <f:selectItems value="#{topicBean.topicList}"/>
</h:selectOneMenu>
```

when I select one of the items in filterTopic menu, the ajax_loader.gif should fall on the dataTable until the data is fetch from homeBean. but it`s not working and i can see in the console the following error:

**Widget for var 'bui' not available!**  -->   core.js.xhtml?ln=primefaces&v=11.0.0:18

**home.xhtml Uncaught TypeError: Cannot read properties of null (reading 'hide')**  -->  home.xhtml

also when refresh page and run PrimeFaces.widgets in browser console, result is:

     bui: c {_super: undefined, cfg: {…}, id: 'form:j_id_bo', jqId: '#form\\:j_id_bo', jq: S.fn.init(1), …}

but when select one of the items in filterTopic menu, bui value set to null:

     bui: null

also this code worked with primefaces 10

melloware:
Do not automatically destroy the block UI.  Also upon `show` we must check and see if the blocker has been already removed from the DOM we need to create it again.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
@@ -42,6 +42,9 @@ PrimeFaces.widget.BlockUI = PrimeFaces.widget.BaseWidget.extend({
         if(this.cfg.blocked) {
             this.show();
         }
+
+        // GitHub #8494 do not remove BlockUI if its panel is removed from DOM
+        this.jq.off('remove');
     },
 
     /**
@@ -101,6 +104,10 @@ PrimeFaces.widget.BlockUI = PrimeFaces.widget.BaseWidget.extend({
      * milliseconds, respectively.
      */
     show: function(duration) {
+        if (!$.contains(document.documentElement, this.blocker.get(0))) {
+            // GitHub #8494 blocker is not longer in DOM so create another
+            this.refresh(this.cfg);
+        }
         this.blocker.css('z-index', PrimeFaces.nextZindex());
 
         //center position of content
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
@@ -1289,10 +1289,10 @@ if (!PrimeFaces.ajax) {
                     var widget = PF(widgetVar);
                     if (widget) {
                         if (widget.isDetached() === true) {
-                            PrimeFaces.widgets[widgetVar] = null;
                             widget.destroy();
 
                             try {
+                                delete PrimeFaces.widgets[widgetVar];
                                 delete widget;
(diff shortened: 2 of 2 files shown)


tandraschko:
I will check this, feels a bit hacky

melloware:
Agreed. thanks for looking at it.

melloware:
Some more info. This was also reported as #5866.  

I think the issue is the blockPanel gets created "inside" the object it's blocking so it gets removed from the DOM when the object is updated. Not sure if it is possible to render the BlockUi DIV outside of the thing it's blocking??

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
```diff
@@ -1289,10 +1289,10 @@ if (!PrimeFaces.ajax) {
                     var widget = PF(widgetVar);
                     if (widget) {
```
jepsar, line 1290: This can be `if (widget && widget.isDetached() === true) {`

primefaces/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
```diff
@@ -101,6 +104,10 @@ PrimeFaces.widget.BlockUI = PrimeFaces.widget.BaseWidget.extend({
      * milliseconds, respectively.
      */
     show: function(duration) {
+        if (!$.contains(document.documentElement, this.blocker.get(0))) {
```
jepsar, line 107: Don't we need to solve this with a listener in `bindTriggers`?

tandraschko:
who moves the blockUI this.jq / this.content into the panel itself? this is the issue actually

therefore the DOM element is removed, detached and widget will be the destroyed

tandraschko:
i will close this in favor of: https://github.com/primefaces/primefaces/issues/8548

PR #8552, opened 2022-03-10
Fix #8548: BlockUI Refactor

### LINKED ISSUE #5866: dataTable: blockUI stop works after ajax update

### LINKED ISSUE #8494: BlockUI: widgetVar is reset to null after partial updates

### LINKED ISSUE #8548: BlockUI: refactor + root element should not be moved in DOM
Currently the BlockUI script is a bit weird as it moves the root element to the target. Sometimes its also cloned, which should not be required.

I think it needs a bit of refactoring, also the code itself reads weird as it covers multiple block targets but the code is hard do unterstand.

This example shows the weird behavior:
```
        <h1>BLOCK UI Null Widget</h1>
        <h:form id="frmTest">
            <h:panelGroup id="panel">
                Panel
            </h:panelGroup>
            <p:blockUI widgetVar="blockPanel" block="panel" blocked="true"/>
            <p:commandButton value="Unblock"
                             update="panel"
                             oncomplete="PF('blockPanel').hide()"/>
        </h:form>
```

BlockUI is actually on the same level as commandButton and panelGroup but our BlockUI script moves the conent into the panel.
How does this work with multiple block target? It will probably duplicate the root element, which means we have multiple DIVs with the same ID, which is also wrong

- [x] Format code for readability
- [x] When using multiple targets prevent duplicate ID's in the DOM
- [x] Do not put the DOM element inside the Target (fixed NULL widget issue)
- [x] Test with showcase and reproducers

melloware:
@tandraschko this is part of my incremental refactoring just trying to untangle this component.  This PR documents the current process and fixes the issue of duplicate ID's when using multiple targets.

image

This now gives it unique `_block` and `_blockcontent` ids using the target id and not the BLockUI's id.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
@@ -179,30 +179,46 @@ PrimeFaces.widget.BlockUI = PrimeFaces.widget.BaseWidget.extend({
      * @private
      */
     render: function() {
-        this.blocker = $('<div id="' + this.id + '_blocker" class="ui-blockui ui-widget-overlay ui-helper-hidden"></div>');
+        var isMultiple = this.target.length > 1;
+        // there can be 1 to N targets
+        for (var i = 0; i < this.target.length; i++) {
+            var currentTarget = $(this.target[i]),
+                currentTargetId = currentTarget.attr('id') || this.id,
+                currentContent = this.jq;
 
-        if (this.cfg.styleClass) {
-            this.blocker.addClass(this.cfg.styleClass);
+            // create a specific blocker for this target
+            var currentBlocker = $('<div id="' + currentTargetId + '_blocker" class="ui-blockui ui-widget-overlay ui-helper-hidden"></div>');
+
+            // style the blocker
+            if (this.cfg.styleClass) {
+                currentBlocker.addClass(this.cfg.styleClass);
+            }
+            if (currentTarget.hasClass('ui-corner-all')) {
+                currentBlocker.addClass('ui-corner-all');
+            }
+
+            // when more than 1 target need to clone the blocker for each target
+            if (isMultiple) {
+                currentContent = currentContent.clone();
+                currentContent.attr('id', currentTargetId + '_blockcontent');
+            }
+
+            // configure the target positioning
+            var position = currentTarget.css("position");
+            if (position !== "fixed" && position !== "absolute") {
+                currentTarget.css('position', 'relative');
+            }
+
+            // ARIA 
+            currentTarget.attr('aria-busy', this.cfg.blocked);
+
+            // append the blocker to the target 
+            currentTarget.append(currentBlocker).append(currentContent);
         }
 
-        if (this.target.hasClass('ui-corner-all')) {
-            this.blocker.addClass('ui-corner-all');
-        }
-
-        if (this.target.length > 1) {
-            this.content = this.content.clone();
-        }
-
-        var position = this.target.css("position");
-        if (position !== "fixed" && position !== "absolute") {
-            this.target.css('position', 'relative');
-        }
-        this.target.attr('aria-busy', this.cfg.blocked).append(this.blocker).append(this.content);
-
-        if (this.target.length > 1) {
-            this.blocker = $(PrimeFaces.escapeClientId(this.id + '_blocker'));
-            this.content = this.target.children('.ui-blockui-content');
-        }
+        // assign all matching blockers to widget
+        this.blocker = $(this.target.find('.ui-blockui.ui-widget-overlay'));
+        this.content = $(this.target.find('.ui-blockui-content'));
     },
 
     /**
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
@@ -179,7 +179,8 @@ PrimeFaces.widget.BlockUI = PrimeFaces.widget.BaseWidget.extend({
      * @private
      */
     render: function() {
-        var isMultiple = this.target.length > 1;
+        var isMultiple = this.target.length > 1,
+            widgetId = this.id;
         // there can be 1 to N targets
         for (var i = 0; i < this.target.length; i++) {
             var currentTarget = $(this.target[i]),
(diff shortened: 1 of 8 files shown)


melloware:
OK I got everything working and the block ui now attach to `document.body` and tested with existing use case and showcase examples!

primefaces/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
```diff
@@ -179,30 +179,68 @@ PrimeFaces.widget.BlockUI = PrimeFaces.widget.BaseWidget.extend({
...
+            });
+            currentContent.css({
+                'height': height + 'px',
+                'width': width + 'px',
+                'left': position.left + 'px',
+                'top': position.top + 'px'
+            });
+
+            // append the blocker to the document 
+            $(document.body).append(currentBlocker).append(currentContent);
```
melloware, line 241: Append to document body instead of the target.

primefaces/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
```diff
@@ -179,30 +179,68 @@ PrimeFaces.widget.BlockUI = PrimeFaces.widget.BaseWidget.extend({
...
+                'width': width + 'px',
+                'left': position.left + 'px',
+                'top': position.top + 'px'
+            });
+            currentContent.css({
+                'height': height + 'px',
+                'width': width + 'px',
+                'left': position.left + 'px',
+                'top': position.top + 'px'
+            });
```
melloware: The reason it was putting the block panel inside the target element was so it could properly calculated position so this just preemptively copies the size and position to the block UI elements.

primefaces/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
```diff
@@ -179,30 +179,68 @@ PrimeFaces.widget.BlockUI = PrimeFaces.widget.BaseWidget.extend({
...
+
+            // append the blocker to the document 
+            $(document.body).append(currentBlocker).append(currentContent);
         }
+
+        // assign all matching blockers to widget
+        this.blocker = $('[data-bui-overlay~="' + widgetId + '"]');
+        this.content = $('[data-bui-content~="' + widgetId + '"]');
```
melloware, line 246: This properly grabs all the block panels the old code was actually only holding the "last" one instead of all of them in the case of multiple target.

tandraschko:
i think we can do it more easily
instead of move this.jq to the panel content, we can clone this.jq (there was a clone already!) and move this + change id

tandraschko:
the problem isnt that we put something inside the panel, the problem we remove this.jq from its original DOM position completely
but not sure whats more failsafe and cleaner (also think about positioning + scrolling)

melloware:
Yep the issue is by putting it inside the panel it's when the panel gets destroyed it takes the block ui DIV with it. So it needs to live outside the panel which is why it's in the doc body now. 

Also it's gets proper ids based on its target to prevent duplicate ids it was getting before.

tandraschko:
i think we should really avoid to move this.jq around, this would require the `appendTo` logic/cleanup and so on
what about not moving or cloning the content of this.jq but just re-position via css?

melloware:
So this.jq is the overlay  mask DiV. It's already being moved and postioned with CSS?  And it's cloned only if the target has more than 1 target it makes a copy. 

So not quite sure how to solve this unless we move that overlay mask outside of the target right?

tandraschko:
if the current fix also works with scrolling and so on, i'm ok with it

jepsar:
Since this was caused by regression, shouldn't we add ITs?

melloware:
Let me test scrolling today and see if I can add some IT tests.

melloware:
@jepsar great idea on the integration tests it helped me find a few more holes in this blockUI code and dom elements not properly being cleaned up on destroy and aggressive cloning when it was not necessary if there is no content to display.

PR #8561, opened 2022-03-14
Fix #8546: OverlayPanel `autoHide` attribute

### LINKED ISSUE #8546: OverlayPanel: add autoHide attribute
**Description**
To guarantee accessibility it has to be possible to hover over an opened overlay panel, mark its text with the mouse and copy said text. At the moment, this does not work as the overlay panel is hidden when the hover is not over the overlay panel's anchor anymore.

The overlay panel has to stay open when hovering over it, I suggest adding the "autoHide" attribute as is already implemented in PrimeReact (see https://primefaces.org/primereact/tooltip/).

- [x] Add `autoHide` property default to `true
- [x] Refactor DOM cleanup
- [x] Add `destroy` method to Selenium widget
- [x] Test `destroy` in Integration Test
- [x] Add showcase example

melloware:
- [x] Add `autoHide` property default to `true`
- [x] Refactor DOM cleanup
- [x] Add `destroy` method to Selenium widget
- [x] Test `destroy` in Integration Test
- [x] Add showcase example

### DIFF
--- a/docs/12_0_0/components/overlaypanel.md
+++ b/docs/12_0_0/components/overlaypanel.md
@@ -38,6 +38,7 @@ dismissable | true | Boolean | When set true, clicking outside of the panel hide
 showCloseIcon | false | Boolean | Displays a close icon to hide the overlay, default is false.
 modal | false | Boolean | Boolean value that specifies whether the document should be shielded with a partially transparent mask to require the user to close the Panel before being able to activate any elements in the document.
 appendTo | null | String | Appends the overlayPanel to the given search expression.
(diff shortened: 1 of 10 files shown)


tandraschko:
i wonder if we should make it the default behavior without configurable attr? no idea
but if other Prime* libs has it, +1

melloware:
Yeah other prime libs do this and my thought was leave it backward compatible and allow this feature to be enabled.  It really only matters when you are using custom mouse events as the default for overlay panel is it only goes away when you CLICK outside the overlaypanel. 

But in this use case the user is using mouse events `showEvent="mouseenter" hideEvent="mouseleave"` just like the Tooltip does.

PR #8573, opened 2022-03-16
Fix PF #8031. table.getColumnMeta() returns columns out of order and …

### LINKED ISSUE #8031: DataExporter: columnToggler breaks export ordering
@melloware In the latest 11.0.0 snapshot (as well as on a snapshot from September 27), the csv datatable export seems to be broken. The exported column order is incorrect. See example:

Screenshot from 2021-10-30 03-00-44

koan00:
Fix PF #8031. Avoid usage of table.getColumnMeta().

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/export/TableExporter.java
+++ b/primefaces/src/main/java/org/primefaces/component/export/TableExporter.java
@@ -24,8 +24,9 @@
 package org.primefaces.component.export;
 
 import java.util.*;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
-import java.util.stream.Collectors;
+
 
 import javax.el.MethodExpression;
 import javax.faces.component.UIComponent;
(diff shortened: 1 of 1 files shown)


primefaces/src/main/java/org/primefaces/component/export/TableExporter.java
```diff
@@ -95,33 +96,35 @@ else if (LangUtils.isNotBlank(column.getField())) {
...
-                                (!visibleColumnsOnly || (visibleColumnsOnly && column.isVisible()))) {
-                        exportcolumns.add(column);
-                    }
-                });
-            }
+        for (ColumnMeta meta : visibleColumnMetadata) {
+            String metaColumnKey = meta.getColumnKey();
+            table.invokeOnColumn(metaColumnKey, ((UIData) table).getRowIndex(), column -> {
+                exportcolumns.add(column);
+            });
```
melloware, line 126: Why remove the check for IsExportable here?   A column can be visible but not exportable?

primefaces/src/main/java/org/primefaces/component/export/TableExporter.java
```diff
@@ -95,33 +96,35 @@ else if (LangUtils.isNotBlank(column.getField())) {
...
+        AtomicBoolean hasNonDefaultSortPriorities = new AtomicBoolean(false);
+        final List<ColumnMeta> visibleColumnMetadata = new ArrayList<>(allColumns.size());
 
-        if (columnMetadata == null || columnMetadata.isEmpty()) {
-            table.forEachColumn(true, true, true, col -> {
-                if (col.isExportable() && (!visibleColumnsOnly || (visibleColumnsOnly && col.isVisible()))) {
-                    exportcolumns.add(col);
+        table.forEachColumn(true, true, true, column -> {
+            if (column.isExportable() && (!visibleColumnsOnly || (visibleColumnsOnly && column.isVisible()))) {
+                String metaColumnKey = column.getColumnKey();
```
melloware: I am testing your PR locally and its telling me the above `metaColumnKey` variable is unused and can be removed.

melloware (changes requested):

primefaces/src/main/java/org/primefaces/component/export/TableExporter.java
koan00: Sorry couldn't get the project loaded into my IDE. Just using a text editor doesn't highlight such oversights. I will correct it.

primefaces/src/main/java/org/primefaces/component/export/TableExporter.java
koan00: This code is looping over visibleColumnMetadata. This list is built in when we iterate over each table column at line 103. Line 104 contains the all of the exportable and visible checks. Only if these checks pass will the column be added to the visibleColumnMetadata. Maybe I should change the variable name to be more clear?

melloware:
OK everything looks good then just remove that 1 dead line and all my testing is passing.

koan00:
I pushed an additional commit with the requested change to my branch/fork. Normally it would just update the PR, but it doesn't seem to have this time?

melloware (approved):

melloware:
Actually I am not seeing your pushed commit?  I  still see the variable?

koan00:
> Actually I am not seeing your pushed commit? I still see the variable?

Yes, its in https://github.com/koan00/primefaces/commit/a97d7c3a501c1f5dafc4370e6297c172c317a7a4

https://www.githubstatus.com/ says there are currently issues with PR. Maybe the PR will update in a little time?

melloware:
OK yeah the build is running now

PR #8578, opened 2022-03-17
8577

### LINKED ISSUE #8577: Timeline: TimelineModel update() defect
**Describe the defect**
update() function in TimelineModel does not work any more.
The component updates the UI, but the events in the model (model.getEvents()) are not being changed.
This did work in PF 8.

Looking into sourcecode of PF 8, the update method looks like this:

    
    public void update(TimelineEvent<E> event, TimelineUpdater timelineUpdater) {
        int index = getIndex(event);
        if (index >= 0) {
            events.set(index, event);

            if (timelineUpdater != null) {
                // update UI
                timelineUpdater.update(event);
            }
        }
    }

In Primefaces 10 this method was adjusted like this:

    public void update(TimelineEvent<E> event, TimelineUpdater timelineUpdater) {
        if (timelineUpdater != null) {
            // update UI
            timelineUpdater.update(event);
        }
    }

So the events of the model are not getting adjusted any more....Why is this?

I can get the correct behaviour by calling model.delete and model.add instead of model.update.

**Environment:**
 - PF Version: _10.0.0_ , _11.0.0_
 - JSF + version: _e.g. Mojarra 2.3
 - Affected browsers: ALL_

kneringerjohann:
fixing https://github.com/primefaces/primefaces/issues/8577

### DIFF
--- a/primefaces/src/main/java/org/primefaces/model/timeline/TimelineModel.java
+++ b/primefaces/src/main/java/org/primefaces/model/timeline/TimelineModel.java
@@ -28,6 +28,7 @@ import org.primefaces.component.timeline.TimelineUpdater;
 import java.io.Serializable;
 import java.time.LocalDateTime;
 import java.util.*;
+import java.util.stream.IntStream;
 
 public class TimelineModel<E, G> implements Serializable {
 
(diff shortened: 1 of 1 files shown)


primefaces/src/main/java/org/primefaces/model/timeline/TimelineModel.java
```diff
@@ -151,9 +152,17 @@ public void update(TimelineEvent<E> event) {
      * @param timelineUpdater TimelineUpdater instance to update the event in UI
      */
     public void update(TimelineEvent<E> event, TimelineUpdater timelineUpdater) {
-        if (timelineUpdater != null) {
-            // update UI
-            timelineUpdater.update(event);
+        int index = IntStream.range(0, events.size())
```
melloware, line 156: Can you add a comment above this like `\\ GitHub #8577 must update events` so we don't remove this again in the future :)

melloware (changes requested):

melloware (approved):

PR #8624, opened 2022-04-02
add Java 18 to build-matrix

christophs78:


christophs78:
We keep all still supported Java-LTS-versions and add the latest non-LTS-version.

Issue #8436, opened 2022-02-11
DataTable: make lazy attribute optional when value is null


tuerker:
With the recent PF 12-Snapshots,  I am getting the following error on some screens that were working for many years:

```
Unable to automatically determine the `lazy` attribute. 
Either define the `lazy` attribute on the component or make sure the `value` attribute doesn't resolve to `null`. clientId: ...
```

Could you please revise the semantics such that no migration is required here:

if no lazy attribute is determined, it is by default lazy="false". finally, if the datatable value is null or an empty collection, the datatable should show nothing.

Thanks in advance.

tandraschko:
Its a warning, not a error and by design
But we should just log it and not throw a exception

tuerker:
The exception causes that the mentioned screens are not working anymore. Thanks.

tuerker:
Dear @tandraschko could please remove the warning which is neither needed nor helpful since the printed clientid is even not the page. I would suggest to replace

```
 if (type == null) {
                LOGGER.warning("Unable to automatically determine the `lazy` attribute, fallback to false. "
                        + "Either define the `lazy` attribute on the component or make sure the `value` attribute doesn't resolve to `null`. "
                        + "clientId: " + this.getClientId());
            }
            else {
                lazy = LazyDataModel.class.isAssignableFrom(type);
            }
```

by 

```
 if (type != null) {
                lazy = LazyDataModel.class.isAssignableFrom(type);
            }
```

This would clearly implement the documented optional attribute semantics where the default is lazy=false.

Many thanks in advance.

PR #8626, opened 2022-04-04
DataTable: Do not warn when optional attribute lazy as well as datatable content are null #8436

tuerker:
remove the warning since it pollutes the log for all legacy applications which rely on the default that the lazy attribute is false whenever not set explicitly

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/api/UIData.java
+++ b/primefaces/src/main/java/org/primefaces/component/api/UIData.java
@@ -90,12 +90,7 @@ public class UIData extends javax.faces.component.UIData {
             Class<?> type = ELUtils.getType(getFacesContext(),
                     getValueExpression("value"),
                     () -> getValue());
-            if (type == null) {
-                LOGGER.warning("Unable to automatically determine the `lazy` attribute, fallback to false. "
-                        + "Either define the `lazy` attribute on the component or make sure the `value` attribute doesn't resolve to `null`. "
-                        + "clientId: " + this.getClientId());
-            }
-            else {
+            if (type != null) {
                 lazy = LazyDataModel.class.isAssignableFrom(type);
             }
 


tandraschko:
-1 from my side
lazy should be a optional attribute, which is automatically determined now
better work with empty collections instead of returning null in your method

legacy is not an argument for me anyway in this case, there are breaking changes each version
compared with this changed, a log statement is harmless

tuerker:
sorry i cannot follow your argumentation. the lazy attribute is optional. but with your warning, you are forcing people to either explicitly use the lazy attribute or to make sure that the LazyDataModel is not null. This is completely not needed and represents an unnecesary enforcement.

tuerker:
there should be no difference whether the LazyDataModel is null or "empty".

tandraschko:
i dont like to have the lazy attribute, therefore we built this mechanism to automatically determine it
we might even remove the lazy attribute completely in future, so we MUST make sure, that we can determine the behavior based on the `value`

and a lazy DT with null `value` is just invalid, therefore its better to not a throw a exception and just give a hint that you should return a empty collection, not null, when you have a non-lazy DT; and thats we do currently

tuerker:
i see your point, but i would opt to announce a breaking change when you actually require that the LazyDataModel is not allowed to be null anymore. until then, i would not do that log which is really annoying. i am pretty sure that we are not alone having a huge application running for so many years now on PF based on the assumption that the LazyDataModel can be null.

tandraschko:
it just never made sense that a LazyDataModel can be null
there are actually 4 valid options currently:
lazy true:
- `value` binded to a LazyDataModel, it doesnt matter if it returns data or not

lazy false:
- `value` binded to a collection
- `value` binded to a empty collection 
- `value` binded to null


with this logging, we currently only "warn" the 4. case, which would be broken with a removed `lazy` attribute, too

tuerker:
we have already fully eliminated the usage of the attribute lazy within all p:datatables. so removing the attribute lazy of the datatable will have no effect at all. the isLazy() method is already doing what it should do. it delivers true if and only if the datatable content is a non-null LazyDataModel. that's fine and all what is needed. the warning is only annoying since it unnecessarily pollutes the log.

whenever you want to disallow non-null datatable content in general, you would have a warning "Non-null value is not allowed for datatable" triggered within the check of the datatable value.

tuerker:
maybe the title of the issue is confusing since the lazy attribute of the datatable should not be used anymore, since there is no need for it. the isLazy() method does the job.

tandraschko:
will close it for now
if you actually have a lazy DT and the value is initial null, lazy is evaluated wrong and saved in viewstate
if you later return a non-null LazyDataModel, the DT is "marked" as lazy=false
this leads to a real bug

and this warn message describes the possible error correctly
we can switch the message to "null values are not allowed" but thats just a different message; the current message describes much better what happens

we will revisit this message when we may remove the lazy attr in future

PR #8628, opened 2022-04-05
Fix #8619: Slider allow odd step/min/max

### LINKED ISSUE #8619: Slider: Issues with Step configuration (calculation)
**Describe the defect**
As I already wrote in the forum (https://forum.primefaces.org/viewtopic.php?f=3&t=71799)
I think there is a little bug in the step configuration or calculation of the slider element.

For my point of view the calculation of the steps is done based on the min value instead of zero base.
If you have a min value of 800 the next step is 1300 (800+500), I would expect 1000 (0 + 500 + 500).
Also when the max value isn't a value of the min value, you can not reach it.
My example here is 15000 end value of the Slider with it is 14800.

**Reproducer**
Reproducer is attached to the ticket, code sniped:
```java
@PostConstruct  
public void init() {
    integer = 3000;
}
```
```xhtml
<h:form id="frmTest">
         <p:inputText id="number" value="#{testView.integer}" />
         <p:slider for="number" minValue="800" maxValue="15000" step="500" />
    </h:panelGrid>
</h:form>
```

**Environment:**
 - PF Version: 11.0.0
 - JSF + version: _e.g. Mojarra 2.2.20
 - Affected browsers: my test is based on Chrome, but I think all

**Expected behavior**
My expected behavior is, that the calculation is based on zero and the first and the last point are calculated individual.

melloware:


melloware:
Added integration tests combining Slider + Spinner

PR #8632, opened 2022-04-05
Fix #8631: Spinner round to nearest step

### LINKED ISSUE #8631: Spinner: Issues with Step calculation
**Describe the defect**
In Issue https://github.com/primefaces/primefaces/issues/8619, I descprided and issue with the Slider Step calculation.
For the Spinner is a simliar behavior in place:

```html
            <p:spinner id="spinner1" value="#{sliderView.number3}" min="0" max="1000" stepFactor="50"
                       suffix=" €" valueChangeListener="#{sliderView.onInputChanged}">
                <p:ajax process="@this" update="growl"/>
            </p:spinner>
            <p:slider for="spinner1" minValue="0" maxValue="1000" step="50"/>
        </h:panelGrid>
```

If you start with 800 and go up, the steps are the following:
800 / 1300 / 1800 / 2300 / ... / 14800 / 15000

If you go down from 15000 you have the following sequence:
15000 / 14500 / 14000 / ... / 1000 / 800

For my point of view, the start value should make now difference in the step calculation.
It should be always based on zero and the first and the last point are calculated individual.

Means in that case:
800 / 1000 / 1500 / ... / 14500 / 15000
15000 / 14500 / 14000 / ... / 1000 / 800

**Reproducer**
Reproducer is attached to the ticket

**Environment**
PF Version: 11.0.0
JSF + version: Mojarra 2.2.20

Expected behavior
My expected behavior is, that the calculation is based on zero and the first and the last point are calculated individual.

melloware:


melloware:
I am merging this because I am writing integration tests in the other PR that use a Slider + Spinner combination.

PR #8653, opened 2022-04-09
IT: one user-action causes two ajax-calls (combination of p:ajax with eg p:commandButton) - ensure correct queuing

christophs78:
We need to discuss this a bit further within my organization, because i can´t reproduce the issue my colleagues pointed me on with 12.0.0-SNAPSHOT.


melloware:
@christophs78 is this safe to merge? Did you get to the bottom of your co-worker's issue?

christophs78:
Yes, it´s safe to be merged. Lately had less time for JSF/PrimeFaces but probably my co-worker´s may have issues with AutoUpdate. (eg. clicking a button which is updated by AutoUpdate because of some p:ajax triggered before; they may work around via ignoreAutoUpdate="true")

PR #8660, opened 2022-04-11
Fix #8659: TabView/Accordion add ui-state-error

### LINKED ISSUE #8659: TabView/AccordionPanel: error indication
When you have a `TabView` with multiple tabs on which validation can fail, it is really handy that the tabs on which errors occur, get an error highlighting.
For this, we use the following Javascript within our project:
```javascript
function markInvalidTabs(id) {
    id = PrimeFaces.escapeClientId(id);
    let c = $(id + '>ul>li');
    c.each(function () {
        let d = $('a', this);
        let tid = d.attr('href').substring(1);
        tid = PrimeFaces.escapeClientId(tid);
        let uiStateErrorElements = $(tid + ' .ui-state-error');
        let uiMessageErrorDetailElements = $(tid + ' .ui-message-error-detail');
        if (uiStateErrorElements.length + uiMessageErrorDetailElements.length > 0) {
            $(this).addClass('ui-state-error');
        } else {
            $(this).removeClass('ui-state-error');
        }
    });
}
```
and hence we call this method after an AJAX update, with `id` the `TabView` identifier.
It would be cool if this would be the default behavior of `TabView`.

melloware:
PF8659-Tab

PF8659-Accordion

### DIFF
--- a/docs/12_0_0/gettingstarted/whatsnew.md
+++ b/docs/12_0_0/gettingstarted/whatsnew.md
@@ -7,6 +7,7 @@ This page contains a list of big features. Please check the GitHub issues for al
 ### PrimeFaces
 
 * AccordionPanel: added support for `options` and `actions` facets of contained `Tab` components.
+* AccordionPanel: now `ui-state-error` class is added to tabs that contain invalid inputs.
 * AutoComplete: added `lazyModel` attribute for fetching autocomplete-suggestions (instead of completeMethod).
 * BreadCrumb: added `seo` attribute for creating JSON Link Data for better SEO.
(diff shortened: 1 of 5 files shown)


melloware:
```css
body .ui-tabs.ui-tabs-top > .ui-tabs-nav li.ui-tabs-header.ui-state-error {
  border-color: red;
}

body .ui-accordion .ui-accordion-header.ui-state-error {
    border-color: red;
    background: lightpink;
    color: red;
}
```

jepsar:
I would drop a note in what's new. Maybe also add this to the showcase to help Mert to do the themes?

melloware:
Good ideas MD updated and examples added to showcase.

PR #8668, opened 2022-04-12
#7598: PrimeFlex 3 (some more cleanup)

### LINKED ISSUE #8662: PrimeFlex 3-Support for PanelGrid
**Description**
Primefaces has the [PanelGrid](https://primefaces.github.io/primefaces/12_0_0/#/components/panelgrid)-component which offers a `layout`-attribute. The currently documented options are:

* `tabular` layout, 
* `grid` layout or 
* `flex` layout

We are using [PrimeFlex 3](https://www.primefaces.org/primeflex/) as our CSS layout engine.
No particular layout option works well with PrimeFlex 3.

* `tabular` generates `table`, `tr` and `td`,
* `grid` generates the deprecated `ui-g` classes and
* `flex` generates the previous major version of PrimeFlex 2 classes (`p-grid`)

An intermediate solution would be to support an option like `flex3` which would generate PrimeFlex 3 classes (like `grid`, `col`, etc.)
or probably a more scalable solution to let the developer inject his own classes like `layout="none"`.

**Example XHTML**
```xml
<h:form id="frmTest">
    <p:panelGrid
        layout="none"
        styleClass="grid"
        columns="2"
        columnClasses="col-9, col-3"
    >
        <div>left content</div>
        <div>right content</div>
</h:form>
```

christophs78:


melloware (approved):

melloware:
Nice work thanks for fixing!

PR #8676, opened 2022-04-15
Fix #8673/#8674: Filters use interface for non-ee environments

### LINKED ISSUE #8673: Should we add a CharacterEncodingFilter?
i always copy the class around my projects - even its used in themedesigner or showcase
actually we should also use it primefaces-test or integration tests

@melloware @christophs78 ?

melloware:
@tandraschko @FlipWarthog @jepsar had to switch to use `Filter` interface instead of `HttpFilter` for Jetty and possibly other non-ee environments.


tandraschko:
HttpFilter is servlet api, not EE

tandraschko:
but if you remove "final", we can also merge it
so we are compatible with older environments

melloware:
Yeah I thought it was odd too!  Removed `final` and merging.

PR #8677, opened 2022-04-15
Fix #8650: Tree: Selected ui-tree-nodes are not always classed correctly

### LINKED ISSUE #8650: Tree: Selected ui-tree-nodes are not always classed correctly
PF 11 and 12

Selected `ui-tree-node` with `selectionMode="checkbox"` has the class `ui-treenode-selected` which is correct:

Screenshot 2022-04-08 at 20 12 05

with `selectionMode="single"` or `multiple`, selected nodes are classed `ui-treenode-unselected` which is wrong:

Screenshot 2022-04-08 at 20 15 36

Relates to #8649

FlipWarthog:


FlipWarthog:
@jepsar @melloware Take a look, this seems to cover all the cases for `single` and `multiple` so they align with `checkbox`.

melloware (approved):

PR #8678, opened 2022-04-15
Fix #8333: Check parent for zIndex value as well

### LINKED ISSUE #8333: InputGroup: Command Button onkeypress not working
I have an InputGroup contains InputText and CommandButton and it is located on a dialog.  

If the dialog configured modal="true", onkeypress will not be triggered when key press on the CommandButton.  But when modal="false" the onkeypress is triggered.  Please take a look.

Attached is reproducer run mvn clean jetty:run and navigate to http://localhost:8080/primefaces-test/

[pf-test.zip](https://github.com/primefaces/primefaces/files/7939485/pf-test.zip)

modal="false" -> alert message shown
image

modal="true" -> nothing happened
image

PF Version: 11.0
JSF + version: _JAVAServer Face 2.2
Affected browsers: ALL

melloware:
Because the input group puts a new SPAN in there and the button is `position:relative` it causes the zIndex of the button to be 1.  I think I safe thing to check is the target zindex or its immediate parent zIndex.


melloware:
This will also fix: https://github.com/primefaces-extensions/primefaces-extensions/issues/804

PR #8680, opened 2022-04-16
Fix #8245: Datatable warn if filteredValues not being used

### LINKED ISSUE #8245: DataTable: Filter + Input component with ajax processing that row and the next one
**Describe the defect**
Whenever I set a filter value in the column that has an input component (like selectBooleanCheckbox, datePicker, inputText, etc.) and then I proceed to change that value by ajax, the table gets updated and it shows the affected values for the row that I clicked and the next one. This doens't happen when I don't set any filters.

**Reproducer**
Project attached: [primefaces-test-master.zip](https://github.com/primefaces/primefaces/files/7821915/primefaces-test-master.zip)

**Environment:**
 - PF Version: _11.0_
 - JSF + version: _Mojarra 2.3_
 - Affected browsers: _Chrome and FireFox_

**To Reproduce**
1. Go to 'test.xhtml'
2. Set filter value 'Yes' for 'Status column'
3. Click the first checkbox
image
image
5. Set filter value 'null' for 'Status column'
6. Changes where applied to clicked row and the next one
image

**Expected behavior**
Only the first row should be changed.

**Example XHTML**
```html
<p:dataTable id="dta-tbl" widgetVar="dta-tbl" var="object" value="#{testView.testList}" paginator="true"
			currentPageReportTemplate="Total: {totalRecords}" paginatorPosition="bottom" rowsPerPageTemplate="5,10,15" rows="5"
			pageLinks="3">
	<f:facet name="header">
		My table
	</f:facet>
	<p:column id="id-col" styleClass="" filterBy="#{object[0]}" headerText="Id.">
		#{object[0]}
	</p:column>
	<p:column id="status-col" styleClass="" filterBy="#{object[1]}" headerText="Status">
		<f:facet name="filter">
			<h:selectOneMenu id="status-sel" onchange="PF('dta-tbl').filter()">
				<f:selectItem itemLabel="" itemValue="#{null}" noSelectionOption="true" />
				<f:selectItem itemLabel="Yes" itemValue="#{true}" />
				<f:selectItem itemLabel="No" itemValue="#{false}" />
			</h:selectOneMenu>
		</f:facet>
		<p:selectBooleanCheckbox id="status-in" value="#{object[1]}">
			<p:ajax process="@this" partialSubmit="true" listener="#{testView.updateStatus(object[0], object[1])}"
				update="@none" oncomplete="PF('dta-tbl').filter()" />
		</p:selectBooleanCheckbox>
	</p:column>
</p:dataTable>
```

**Example Bean**
```java
@Data
@Named
@ViewScoped

	private String string;
	private List<Object[]> testList;

	@PostConstruct
	public void init() {
		string = "Welcome to PrimeFaces!!!";

		testList = new ArrayList<>();
		Object[] tmp = null;
		for (int idx = 0; idx < 5; idx++) {
			tmp = new Object[] { idx, true };
			testList.add(tmp);
		}

	}

	public void updateStatus(Integer idx, Boolean value) {

		System.out.println(!value);

	}

}
```

melloware:
Added a warning to the log.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java
+++ b/primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java
@@ -24,6 +24,8 @@
 package org.primefaces.component.datatable;
 
 import java.util.*;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
@@ -184,6 +186,7 @@ public class DataTable extends DataTableBase {
             .build();
 
     private static final Collection<String> EVENT_NAMES = BEHAVIOR_EVENT_MAPPING.keySet();
+    private static final Logger LOGGER = Logger.getLogger(DataTable.class.getName());
 
     private boolean reset = false;
     private List<UIColumn> columns;
@@ -305,6 +308,7 @@ public class DataTable extends DataTableBase {
                 // in older PF versions, we stored the filtered data in the viewstate but this blows up memory
                 // and caused bugs with editing and serialization like #7999
                 if (isFilteringCurrentlyActive()) {
+                    LOGGER.log(Level.WARNING, "DataTable with filtering should also set the 'filteredValue' attribute. ClientId: " + getClientId());
                     filterAndSort();
                 }
             }


tandraschko:
Actually it should work perfectly without

melloware:
So the only way to fix that is I think we have to prevent it from caling Filter 3 times instead of 1. But it looks like its looking for a `filterValues` VE to determine that?

tandraschko:
Why is it done 3 times? In think it should only be 1 time more as with filteredValue, as the value needs to be restored
But anyway, its not a problem, just a bit slower?
Maybe we should just add doc hints

melloware:
No in the ticket @FlipWarthog diagnosed that its recreating the filtered values multple times and thus the user's list is off by reference.  Once the user added `filterValues` the bug goes away.

tandraschko:
Then we need a real analysis

melloware:
OK closing the PR.

PR #8704, opened 2022-04-21
Fix sass plugin config

melloware:


melloware:
Moved it outside the execution so you can run `mvn dart-sass:watch-sass`

PR #8715, opened 2022-04-26
Fix #7906: FileUpload simple mode clear() widget method

### LINKED ISSUE #7906: FileUpload: Add option to clear selected file in simple mode
I can do something like:

```js
PF('widget').input.val(null);
$('.ui-fileupload-filename', PF('widget').jq).html('');
```

Would be nice to have something like `PF('widget').clear()` to clear the file.

melloware:


jepsar:
Sorry, missed this notification. Looking good! Great job!

PR #8720, opened 2022-04-26
Fix #8717: AutoComplete clear suggestions

### LINKED ISSUE #8717: AutoComplete: Invalid State with forceSelection=true
**Describe the defect**
A `p:autoComplete` with `forceSelection=true` will have an invalid state if the shown label does not match the search expression.

**Reproducer**
https://github.com/fanste/primefaces-test/tree/test-8717

**Environment:**
 - PF Version: 11.0.0, 12-SNAPSHOT

**To Reproduce**
_The reproducer uses 12-SNAPSHOT._

1. Open the testing page
2. The first autocomplete is empty, try to copy the label of the second autocomplete to the first one
3. Leave the first autocomplete - forceSelection will clear it, as no results could be found by the completion method
4. Cut the label of the second autocomplete
5. Leave the second autocomplete - everything is fine
6. Paste the label back to the second autocomplete
7. Leave the second autocomplete - forceSelection accepts the inserted label, but the hidden input is not updated with the correct item value and therefore the button will fail (stacktrace on the server side)

**Expected behavior**
The autocomplete should not accept the pasted label, as it does not match any item on the server side.

**Root cause**
The reason for that behavior is, that the javascript manages an array `currentItems` if `forceSelection=true`. This array is filled in `#setupForceSelection` and `#showSuggestions`. But the latter one is only updated if the server returned some results.

That is not the case in this example. The search expression differs from the label expression.

**Possible fix**
Clear the array `currentItems` if the server returned an empty result.

A quick'n'dirty fix is applied to the third autocomplete

melloware:


fanste (approved):
Your implementation matches my quick'n'dirty fix - looks good.

melloware:
Thanks for the suggestion!

PR #8731, opened 2022-04-28
Docs: replace apostrophe with singlequote

Epsilon02:
Inside the docs for `resolveWidgetVar` is an apostrophe used which throws an error in the application server.
The solution was to replace the outer single quotes with back ticks and use single quotes instead the apostrophe.


melloware:
Thanks for the PR!

PR #8738, opened 2022-04-29
Fix #8737: Added title tag for tooltip information.

### LINKED ISSUE #8737: Avatar: Missing title tag for tooltip information.
**Description**
Missing title tag for tooltip information.

**Example XHTML**
```xml
<p:avatar label="Johann Chrysostom Wolfgang Amadeus Mozart" title="Wolfgang Amadeus Mozart"/>
```

aripddev:


jepsar (changes requested):
Can you please update the documentation (in the parent project)?

jepsar:
I think the title will always be the same as the label. Do we want to explicitly set the title while it's the same as the label?

Maybe we can use "true" to copy the label value?

melloware:
I definitely like the flexibility of declaring a different title but maybe we should just default it to label and let the user override it to "" if they truly don't want a title?

aripddev:
Is it enough to set it to `label` instead of setting it to null?

```java
public String getTitle() {
   return (String) getStateHelper().eval(PropertyKeys.title, getLabel());
}
```

aripddev:
> Is it enough to set it to `label` instead of setting it to null?
> 
> ```java
> public String getTitle() {
>    return (String) getStateHelper().eval(PropertyKeys.title, getLabel());
> }
> ```

yeap, that is worked.

primefaces/src/main/java/org/primefaces/component/avatar/AvatarRenderer.java
```diff
@@ -74,6 +75,10 @@ public void encodeEnd(FacesContext context, UIComponent component) throws IOExce
             writer.writeAttribute("style", style, "style");
         }
 
+        if (title != null) {
```
jepsar:
```suggestion
        if (!LangUtils.isEmpty(title)) {
```

melloware (approved):

jepsar (approved):

jepsar:
Fixed the doc https://github.com/primefaces/primefaces/commit/dab97cb78f3d571865878d0bf8b2140ede38ac6c

PR #8745, opened 2022-04-30
Fix #8743: Targeted reset after update=@all

### LINKED ISSUE #8743: Core: FullAjaxExceptionHandler does not show error page when exception is thrown during ajax request
**Describe the defect**
Original report on OmniFaces: https://github.com/omnifaces/omnifaces/issues/686

**Reproducer:**
[pf-8743.zip](https://github.com/primefaces/primefaces/files/8597485/pf-8743.zip)

Run `mvn clean jetty:run` and click "Send AJAX Error".

**Environment:**
 - PF Version: _11.0.0_

Analysis done by @BalusC in that ticket should be implemented.

melloware:
Trying to make the reset on ALL more targeted than blowing away the whole `windows.PrimeFaces = null;` namepsace.


tandraschko:
i actually dont like the change at all as the new view doesnt use primefaces, so its quite heavy to fix that as many of the AJAX processing depends on it, after processing the update=ViewRoot
therefore i think its a dirty but valid workaround

PR #8747, opened 2022-05-01
Fix #8746: AutoComplete only toggle icon if the state changes

### LINKED ISSUE #8746: AutoComplete: multiple inline load indicators are shown (on slow queries)
Queries are not processed sequentially, but parallel. This results in multiple load indicators if you trigger multiple simultaneous queries.

jepsar:
Fix #8746


melloware:
Interesting we have another IT test that just started failing today May 1.  So I have a feeling its a problem with the test.

Schedule001Test.testEventSelect:140 expected: <Champions League Match> but was: <Birthday Party>

I will check it out

melloware:
OK I fixed the Schedule test.

melloware (approved):

PR #8749, opened 2022-05-01
Fix #8748: ui-state-loading based on number of Ajax requests

### LINKED ISSUE #8748: UICommand: multiple inline Ajax status icons with disableOnAjax="false" and async="true"
For example, change the showcase to

```xhtml
<p:commandButton value="Keep Enabled" async="true" disableOnAjax="false" action="#{buttonBasicView.sleep()}" />
```

jepsar:
Fix #8748

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.commandbutton.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.commandbutton.js
@@ -3,7 +3,7 @@
  * 
  * CommandButton is an extended version of standard commandButton with AJAX and theming.
  *
- * @prop {number} [ajaxCount] number of concurrent active Ajax requests.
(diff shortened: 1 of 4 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.commandbutton.js
```diff
@@ -3,6 +3,8 @@
  * 
  * CommandButton is an extended version of standard commandButton with AJAX and theming.
  *
+ * @prop {number} [ajaxCount] number of concurrent active Ajax requests.
```
melloware: I think to start your sentence the "n" in Number" has to be capitilized its failing "lint" check right now

melloware (changes requested):

melloware:
I fixed the JSDoc right on your branch.  Building now.

melloware (approved):

PR #8758, opened 2022-05-03
Fix #8520: CSP report only policy

### LINKED ISSUE #8520: CSP: Report only

melloware:


primefaces/src/main/java/org/primefaces/config/PrimeConfiguration.java
```diff
@@ -123,10 +124,19 @@ public PrimeConfiguration(FacesContext context, PrimeEnvironment environment) {
...
-            cspPolicy = externalContext.getInitParameter(Constants.ContextParams.CSP_POLICY);
+        value = Objects.toString(externalContext.getInitParameter(Constants.ContextParams.CSP));
+        switch (value) {
+            case "true":
+                csp = Boolean.TRUE;
+                cspPolicy = externalContext.getInitParameter(Constants.ContextParams.CSP_POLICY);
+                break;
+            case "reportOnly":
+                csp = Boolean.TRUE;
+                cspReportEndpoint = externalContext.getInitParameter(Constants.ContextParams.CSP_REPORT_ONLY_POLICY);
```
tandraschko: Endpoint or policy, where endpoint is just a property?

melloware: I used the same name that @cnsgithub was using but it really should be policy.  I will fix it

melloware:
I changed it to be reportPolicy.  Unless you want the property to be CSP_REPORT_ONLY_ENDPOINT ?

melloware:
I actually think all of this should be renamed `CSP_REPORT_ONLY_URI` since that is really what we want in that field.

tandraschko:
please see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only

it supports much more as the URI

Therefore `primefaces.CSP_REPORT_ONLY_POLICY=report-uri /my-csp-reports` is the way to go

melloware:
OK updated all docs and removed the hard coded `report-ui` allowing this to be a full policy string with the documentation recommending it be `report-uri /my-csp-reports`

primefaces/src/main/java/org/primefaces/csp/CspPhaseListener.java
```diff
@@ -67,7 +70,16 @@ public void beforePhase(PhaseEvent event) {
         CspState state = PrimeFacesContext.getCspState(context);
 
         String policy = LangUtils.isBlank(customPolicy.get()) ? "script-src 'self'" : customPolicy.get();
-        externalContext.addResponseHeader("Content-Security-Policy", policy + " 'nonce-" + state.getNonce() + "'");
+        policy += " 'nonce-" + state.getNonce() + "';";
```
tandraschko: i think this var should be moved in the `else` statement?

melloware: Agreed.  Moved

`

primefaces/src/main/java/org/primefaces/csp/CspPhaseListener.java
```diff
@@ -67,7 +70,15 @@ public void beforePhase(PhaseEvent event) {
         CspState state = PrimeFacesContext.getCspState(context);
 
         String policy = LangUtils.isBlank(customPolicy.get()) ? "script-src 'self'" : customPolicy.get();
```
tandraschko: i wonder of this should also be in the `else` statemen as customPolicy should not be used in report-only mode?

melloware: Ahhh wait a minute.  I think his original idea was correct.

So in report only mode it would generate...

```
script-src nonce-12344 report-uri \myreports
```

There is no way a user can generate `script-src nonce-12344 ` so the reports won't be reporting anything.   I think I have to put it back the way it was.

So `report-uri \myreports` won't report anything cause you have not turned on our CSP Script rules with NONCE.

melloware:
OK latest commit makes it the most correct where each block is only doing exactly what it needs to do.

PR #5391, opened 2019-12-01
 Fix #5060: DatePicker: No timeInput attribute like Calendar 

christophs78:
Showcase-PR: https://github.com/primefaces/showcase-facelift/pull/102

- [x] We should first care about: https://github.com/primefaces/primefaces/pull/5390, https://github.com/primefaces/primefaces/pull/5384, https://github.com/primefaces/primefaces/pull/5377.
- [x] After they are done i have to add timeValidation to this PR. 

@markusg80: Can you check this PR and provide feedback?
@mertsincan: Can you have a look at his PR? It´s "your" component and it would be great to get some valuable feedback.

Issue #5060, opened 2019-08-12
DatePicker: No timeInput attribute like Calendar


markusg80:
## 1) Environment
- PrimeFaces version: 6.2.21
- Does it work on the newest released PrimeFaces version? Version? No
- Does it work on the newest sources in GitHub? (Build by source -> No https://github.com/primefaces/primefaces/wiki/Building-From-Source)
- Application server + version: Payara 5.192
- Affected browsers: All

## 2) Expected behavior

Datepicker should have timeInput attribute to enter time/date manually like with calendar.

## 3) Actual behavior

There is no obvios way in actual API to get this 

## 4) Steps to reproduce

.. 

## 5) Sample XHTML

..

## 6) Sample bean

..

markusg80:
Any news regarding this issue? It is really important feature for us to change from Calendar to Datepicker component

tandraschko:
Not really :D
Feel free to provide a PR or sponsor it via PrimeFaces PRO!

markusg80:
Is there really no one else with the need to enter the time with Keyboard? It can be very boring to enter time value with mouse controls. To submit a pr i dont have the time st this moment and my company dont have any intension to spend money for one festure.

melloware:
There probably are others out there that need this feature but it will only get fixed if someone in the community submits a PR or sponsors it via PRO.  if your company doesn't want to then you will just have to wait and rely on others.

markusg80:
And why Primefaces has no intensions to add features to his components many people asking for. It is always an advantage for Primefaces to have useful features for the components! I think its not so a big effort for experienced programmer to implement this. I doesn‘t have experience in creating web components. so for me it will be a big effort to implement feature in a Webcomponent and to ensure to respect all Primefaces guidelines.

melloware:
I don’t disagree. I am just telling you your options for making it happen.

christophs78:
@markusg80 
Most of us do this work in our spare time for free. (As it´s the same with a good amount of other open source projects. There are some "big ones" which are sponsored. This means there´s a company behind which pays some developers. But for PrimeFaces at this point in time maybe 80% of the work is done by the "community" for free.)

Calendar has the following UI for date + time:
![grafik](https://user-images.githubusercontent.com/10461942/69898028-ff9e5000-1353-11ea-931c-43d6b234cc82.png)

DatePicker the following:
[...]

PR #8759, opened 2022-05-03
Fix #8430: DatePicker timepicker tab order

### LINKED ISSUE #8430: DatePicker: Time inputs can not be focussed by keyboard when inline
```xhtml
<p:datePicker id="inline" value="#{calendarJava8View.date1}" inline="true"
              showTime="true" timeInput="true"/>
```

melloware:


jepsar:
I'll test this tonight

jepsar:
It's working in Chrome and Safari (macOS). In Firefox you can focus the inputs, but everything else is not focusable (but that's already the case before this PR). So I think it is good to merge and we need a separate issue for the Firefox issues.

Do we want a review from @christophs78 as well?

jepsar (approved):

melloware:
I added him to this review.

christophs78:
The current state was introduced with https://github.com/primefaces/primefaces/pull/5391 which implemented https://github.com/primefaces/primefaces/issues/5060.
When this PR does not cause regressions for non-inline-mode (which is default) on Chrome i´m fine with this one.

melloware:
OK i fixed it for both inline and popup mode. popup mode was the tabindex it was before but inline mode is 0.

melloware:
Also the notes like `//type="number" min="0" max="59" does not work well on Firefox 70, so we don´t use it`  We are 30 versions past 70 should we implement this stuff the correct way now?

PR #8761, opened 2022-05-04
Fix #7767/#7763/#7765/#3781: AutoComplete/SelectOneMenu/RowExpansion/OrderList duplicate ids

### LINKED ISSUE #3781: OrderList: does not append index to clientId
## 1) Environment (same as in Primefaces test project)
- PrimeFaces version: 6.2
- Does it work on the newest released PrimeFaces version? Version? No. 6.2
- Application server + version: Jetty
- Affected browsers: All

## 2) Expected behavior

Make an orderList and put a Column with an outputText, give the outputText an ID. Render the component, the IDs of the outputTexts should be unique (prepended with an index).

## 3) Actual behavior

Each outputText hast the same ID, which makes it impossible to distinguish them and therefore you can't use elements like overlayPanel with it.

## 4) Steps to reproduce

See code below. I've also put a dataGrid into the code and a overlayPanel into both components, the code works for dataGrid but not for orderList.
Output of the outputTexts in the orderList:
`<span id="myoutputtext1">lala</span>`
Output of the outputTexts in the dataGrid:
`<span id="j_idt14:0:myoutputtext2">lala</span>`
[test.zip](https://github.com/primefaces/primefaces/files/2098087/test.zip)

## 5) Sample XHTML

```xml
<p:orderList value="#{testView.testStrings}" var="testString" itemValue="#{testString}">

    <p:column>
        <p:outputLabel class="section-list-label" onmouseover="PF('overlayPanel1').show('#{component.clientId}')"
                       onmouseout="PF('overlayPanel1').hide('#{component.clientId}')" >
            <h:outputText id="myoutputtext1" value="#{testString}" />
            <p:overlayPanel widgetVar="overlayPanel1" class="formpreview-settings-panel section-list-overlay">
                <p:outputPanel id="recurringStateDetail">
                    <h:outputText value="OverlayText" />
                </p:outputPanel>
            </p:overlayPanel>
        </p:outputLabel>
    </p:column>
</p:orderList>

<p:dataGrid value="#{testView.testStrings}" var="testString">

    <p:column>
        <p:outputLabel class="section-list-label" onmouseover="PF('overlayPanel1').show('#{component.clientId}')"
                       onmouseout="PF('overlayPanel1').hide('#{component.clientId}')" >
            <h:outputText id="myoutputtext2" value="#{testString}" />
            <p:overlayPanel widgetVar="overlayPanel1" class="formpreview-settings-panel section-list-overlay">
                <p:outputPanel id="recurringStateDetail">
                    <h:outputText value="OverlayText" />
                </p:outputPanel>
            </p:overlayPanel>
        </p:outputLabel>
    </p:column>
</p:dataGrid>
```

## 6) Sample bean
```java
package org.primefaces.test;

import java.io.Serializable;
import java.util.ArrayList;
import java.util.List;

@ViewScoped
   
    
    @PostConstruct  
    public void init() {
    }

    public List<String> getTestStrings(){
        return new ArrayList<String>(){{add("lala");add("bbbb");add("cccc");}};
    }
}
```

### LINKED ISSUE #7763: AutoComplete: duplicate ids with custom content
**Describe the defect**
An autocomplete with custom content renders duplicate ids in query results.

**Reproducer**
[primefaces-autocomplete-custom-content-duplicate-ids.zip](https://github.com/primefaces/primefaces/files/7067446/primefaces-autocomplete-custom-content-duplicate-ids.zip)
(Targets 10.0.0 instead of master-SNAPSHOT due to #7739.)

**Environment:**
 - PF Version: 6.2.30, 10.0.0
 - JSF + version: ALL
 - Affected browsers: ALL

**To Reproduce**
1. Open the reproducer.
2. Type "1" into the field.
3. See the error message in the browser console (a script detects duplicate ids and logs them).
image

**Expected behavior**
Client ids should probably include iteration, like with `ui:repeat`: "x:1:y", "x:2:y"...

**Example XHTML**
```html
<h:form id="form">
	<p:autoComplete id="autocomplete" completeMethod="#{testView.complete}"
		var="v" itemLabel="#{v}" itemValue="${v}">
		<p:ajax event="query" oncomplete="checkDuplicateIds()" />
		<p:column>
			<p:graphicImage value="#{resource['folder.png']}" />
			#{v}
		</p:column>
	</p:autoComplete>
</h:form>

<script>
	function checkDuplicateIds() {
		setTimeout(function() {
			var ids = {};
			$('[id]').each(function() {
				var id = this.id;
				if (id &amp;&amp; ids[id]) {
					ids[id].push(this);
					console.error('Duplicate id: ' + id + ' in elements', ids[id]);
				} else {
					ids[id] = [this];
				}
			});
		}, 100);
	}
	checkDuplicateIds();
</script>
```

**Example Bean**
```java
@Data
@Named
@ViewScoped
    
	public List<String> complete(String query) {
		return Stream.of("qwer", "asdf", "zxcv", "asdfg", "000", "111", "123").filter(e -> e.contains(query)).collect(Collectors.toList());
	}

}
```

### LINKED ISSUE #7765: DataTable: rowExpansion+rowEditor creates duplicate ids
**Describe the defect**
Using dataTable with rowExpansion and rowEditor creates duplicate ids.

**Reproducer**
[primefaces-datatable-rowexpansion-roweditor-duplicate-ids.zip](https://github.com/primefaces/primefaces/files/7068337/primefaces-datatable-rowexpansion-roweditor-duplicate-ids.zip)
(Targets 10.0.0 instead of master-SNAPSHOT due to #7739.)

**Environment:**
 - JSF + version: _ALL_
 - Affected browsers: _ALL_

**To Reproduce**
1. Open the reproducer.
2. Click rowEditor to start editing a row.
3. Finish editing (apply or cancel - doesn't matter).
4. See the error message in the browser console (a script detects duplicate ids and logs them).
image
For some reason rowExpansion content is cloned on row edit finish.

**Expected behavior**
No duplicate ids.

**Example XHTML**
```html
<h:form id="form">
	<p:dataTable id="datatable" value="#{[1,2]}" var="row" editable="true" expandedRow="true">
		<p:ajax event="rowEdit" oncomplete="checkDuplicateIds()" />
		<p:ajax event="rowEditCancel" oncomplete="checkDuplicateIds()" />

		<p:column width="80">
			<p:rowToggler />
		</p:column>

		<p:column width="80">
			<p:rowEditor />
		</p:column>

		<p:column>
			#{row} row
		</p:column>

		<p:rowExpansion>
			<h:outputText id="text" value="#{row} details"/>
		</p:rowExpansion>
	</p:dataTable>
</h:form>

<script>
	function checkDuplicateIds() {
		setTimeout(function() {
			var ids = {};
			$('[id]').each(function() {
				var id = this.id;
				if (id &amp;&amp; ids[id]) {
					ids[id].push(this);
					console.error('Duplicate id: ' + id + ' in elements', ids[id]);
				} else {
					ids[id] = [this];
				}
			});
		}, 100);
	}
	checkDuplicateIds();
</script>
```

**Example Bean**
None.

### LINKED ISSUE #7767: SelectOneMenu: duplicate ids with custom content
**Describe the defect**
Using selectOneMenu with custom content results in duplicate ids.

**Reproducer**
[primefaces-7767-selectonemenu-custom-content-duplicate-ids.zip](https://github.com/primefaces/primefaces/files/7068606/primefaces-7767-selectonemenu-custom-content-duplicate-ids.zip)
(Targets 10.0.0 instead of master-SNAPSHOT due to #7739.)

**Environment:**
 - JSF + version: _ALL_
 - Affected browsers: _ALL_

**To Reproduce**
1. Open the reproducer.
2. See the error message in the browser console (a script detects duplicate ids and logs them).
image

**Expected behavior**
Client ids should probably include iteration, like with ui:repeat: "x:1:y", "x:2:y"...

**Example XHTML**
```html
<h:form id="form">
	<p:selectOneMenu id="menu" var="v">
		<f:selectItems value="#{testView.selectItems}" />
		<p:column>
			<h:outputText id="item_id" value="#{v.id}" />
		</p:column>
		<p:column>
			<h:outputText id="item_name" value="#{v.name}" />
		</p:column>
	</p:selectOneMenu>
</h:form>

<script>
	function checkDuplicateIds() {
		setTimeout(function() {
			var ids = {};
			$('[id]').each(function() {
				var id = this.id;
				if (id &amp;&amp; ids[id]) {
					ids[id].push(this);
					console.error('Duplicate id: ' + id + ' in elements', ids[id]);
				} else {
					ids[id] = [this];
				}
			});
		}, 100);
	}
	checkDuplicateIds();
</script>
```

**Example Bean**
```java
@Data
@Named
@ViewScoped
    
	private List<TestBean> data = Arrays.asList(new TestBean(0, "Zero"),
			new TestBean(1, "One"), new TestBean(2, "Two"));

	public List<SelectItem> getSelectItems() {
		return data.stream().map(bean -> new SelectItem(bean, bean.getName())).collect(Collectors.toList());
	}

	public static class TestBean {

		private Integer id;
		private String name;

		public TestBean() {
		}

		public TestBean(Integer id, String name) {
			this.id = id;
			this.name = name;
		}

		public Integer getId() {
			return id;
		}

			this.id = id;
		}

		public String getName() {
			return name;
		}

			this.name = name;
		}

		@Override
		public String toString() {
			return getId() + " " + getName();
		}

	}

}
```

### LINKED ISSUE #8765: Core: Prevent duplicate IDs in DOM
### Describe the bug

PrimeFaces has a few places where duplicate IDs are present in the DOM.  This issue is to track all of them in 1 spot.  Most of the time its in places where PF uses `UIColumn` to iterate over some advanced rendering but not always.

**Why is it bad?**
https://softwareengineering.stackexchange.com/questions/127178/two-html-elements-with-same-id-attribute-how-bad-is-it-really

**Components Affected:**
- [x] AutoComplete (https://github.com/primefaces/primefaces/issues/7763
- [x] SelectOneMenu (https://github.com/primefaces/primefaces/issues/7767)
- [x] OrderList (https://github.com/primefaces/primefaces/issues/3781)
- [ ] DataTable RowExpansion (https://github.com/primefaces/primefaces/issues/7765)
- [ ] Diagram (https://github.com/primefaces/primefaces/issues/9551)

- [x] SelectCheckBoxMenu (https://github.com/primefaces/primefaces/issues/4696)

**Possible Solutions:**
1. Just ignore it until browser doesn't support it anymore
2. Commit a fast workaround (https://github.com/primefaces/primefaces/pull/8761)
3. Do it the right JSF way, which takes a long time using `UIRepeat / UINamingContainer`

### Reproducer

Great example from https://github.com/primefaces/primefaces/issues/7767 on how to test with SelectOne and use a script to check for duplicate ID's.

```xml
<h:form id="form">
	<p:selectOneMenu id="menu" var="v">
		<f:selectItems value="#{testView.selectItems}" />
		<p:column>
			<h:outputText id="item_id" value="#{v.id}" />
		</p:column>
		<p:column>
			<h:outputText id="item_name" value="#{v.name}" />
		</p:column>
	</p:selectOneMenu>
</h:form>

<script>
	function checkDuplicateIds() {
		setTimeout(function() {
			var ids = {};
			$('[id]').each(function() {
				var id = this.id;
				if (id &amp;&amp; ids[id]) {
					ids[id].push(this);
					console.error('Duplicate id: ' + id + ' in elements', ids[id]);
				} else {
					ids[id] = [this];
				}
			});
		}, 100);
	}
	checkDuplicateIds();
</script>
```

### Expected behavior

No components should generate duplicate ID's and all ID's in the DOM should be unique.

Client ids should probably include iteration, like with UI "x:1:y", "x:2:y"...

### PrimeFaces edition

Community

### PrimeFaces version

11.0.0

### Theme

ANY

### JSF implementation

All

### JSF version

ALL

### Browser(s)

ALL is

melloware:
This was the most elegant way to solve this without making a mess.  It basically captures the output in FastStringWriter and then appends the index on to each ID so `form:abc` becomes `form:abc:0` then `form:abc:1`

Tested against both reproducers and all is working properly.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
+++ b/primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
@@ -23,6 +23,7 @@
  */
 package org.primefaces.util;
 
+import java.io.IOException;
 import java.io.Serializable;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
@@ -41,6 +42,7 @@ import javax.faces.component.visit.VisitContext;
 import javax.faces.component.visit.VisitHint;
 import javax.faces.context.ExternalContext;
 import javax.faces.context.FacesContext;
+import javax.faces.context.ResponseWriter;
 import javax.faces.convert.Converter;
 import javax.faces.render.Renderer;
 
@@ -566,6 +568,34 @@ public class ComponentUtils {
 
         return false;
     }
+
+    /**
+     * GitHub #7763 Renders an in memory UIComponent and updates its ID with an index.
+     * For example id="form:test" becomes id="form:test:0".
+     *
+     * @param context the {@link FacesContext}.
+     * @param component the {@link UIComponent} to render.
+     * @param index the index number to append to the ID
+     * @throws IOException if any IO error occurs
+     */
+    public static void encodeIndexedId(FacesContext context, UIComponent component, int index) throws IOException {
+        // swap writers
+        ResponseWriter writer = context.getResponseWriter();
+        FastStringWriter fsw = new FastStringWriter();
+        context.setResponseWriter(writer.cloneWithWriter(fsw));
+
+        // encode the component
+        component.encodeAll(context);
+
+        // restore the original writer
+        context.setResponseWriter(writer);
+
+        // append index to all id's
+        String encodedComponent = fsw.toString();
+        encodedComponent = encodedComponent.replaceAll("\\sid=\"(.*?)\"", " id=\"$1:" + index + "\"");
+        writer.write(encodedComponent);
+    }
+
 
     public static Object getDynamicColumnValue(UIComponent component) {
         org.primefaces.component.api.UIColumn column =
--- a/primefaces/src/main/java/org/primefaces/component/autocomplete/AutoCompleteRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/autocomplete/AutoCompleteRenderer.java
@@ -550,19 +550,20 @@ public class AutoCompleteRenderer extends InputRenderer {
         writer.startElement("tbody", ac);
 
         if (items != null) {
+            int index = 0;
             if (ac.isClientQueryMode() || items instanceof Map) {
                 for (Map.Entry<String, List<String>> entry : ((Map<String, List<String>>) items).entrySet()) {
                     String key = entry.getKey();
                     List<String> list = entry.getValue();
 
                     for (Object item : list) {
-                        encodeSuggestionItemsAsTable(context, ac, item, converter, pojo, var, key);
+                        encodeSuggestionItemsAsTable(context, ac, item, converter, pojo, var, key, index++);
                     }
                 }
             }
             else {
                 for (Object item : (List) items) {
-                    encodeSuggestionItemsAsTable(context, ac, item, converter, pojo, var, null);
+                    encodeSuggestionItemsAsTable(context, ac, item, converter, pojo, var, null, index++);
                 }
 
                 if (ac.hasMoreSuggestions()) {
@@ -664,7 +665,7 @@ public class AutoCompleteRenderer extends InputRenderer {
     }
 
     protected void encodeSuggestionItemsAsTable(FacesContext context, AutoComplete ac, Object item, Converter converter,
-            boolean pojo, String var, String key) throws IOException {
+            boolean pojo, String var, String key, int index) throws IOException {
         ResponseWriter writer = context.getResponseWriter();
         Map<String, Object> requestMap = context.getExternalContext().getRequestMap();
         UIComponent itemtip = ac.getFacet("itemtip");
@@ -704,7 +705,7 @@ public class AutoCompleteRenderer extends InputRenderer {
                     writer.writeAttribute("class", column.getStyleClass(), null);
                 }
 
-                column.encodeAll(context);
+                encodeIndexedId(context, column, index);
 
                 writer.endElement("td");
             }
@@ -713,7 +714,7 @@ public class AutoCompleteRenderer extends InputRenderer {
         if (ComponentUtils.shouldRenderFacet(itemtip)) {
             writer.startElement("td", null);
             writer.writeAttribute("class", AutoComplete.ITEMTIP_CONTENT_CLASS, null);
-            itemtip.encodeAll(context);
+            encodeIndexedId(context, itemtip, index);
             writer.endElement("td");
         }
 
--- a/primefaces/src/main/java/org/primefaces/component/datatable/feature/RowExpandFeature.java
+++ b/primefaces/src/main/java/org/primefaces/component/datatable/feature/RowExpandFeature.java
@@ -35,6 +35,7 @@ import org.primefaces.component.datatable.DataTable;
 import org.primefaces.component.datatable.DataTableRenderer;
 import org.primefaces.component.datatable.DataTableState;
 import org.primefaces.component.rowexpansion.RowExpansion;
+import org.primefaces.util.ComponentUtils;
 import org.primefaces.util.LangUtils;
 
 public class RowExpandFeature implements DataTableFeature {
@@ -100,7 +101,7 @@ public class RowExpandFeature implements DataTableFeature {
             writer.startElement("td", null);
             writer.writeAttribute("colspan", table.getColumnsCount(), null);
 
-            table.getRowExpansion().encodeAll(context);
+            ComponentUtils.encodeIndexedId(context, table.getRowExpansion(), rowIndex);
 
             writer.endElement("td");
 
--- a/primefaces/src/main/java/org/primefaces/component/orderlist/OrderListRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/orderlist/OrderListRenderer.java
@@ -160,6 +160,7 @@ public class OrderListRenderer extends CoreRenderer {
         ResponseWriter writer = context.getResponseWriter();
         String var = old.getVar();
         Converter converter = old.getConverter();
+        int index = 0;
 
         for (Object item : model) {
             context.getExternalContext().getRequestMap().put(var, item);
@@ -187,7 +188,7 @@ public class OrderListRenderer extends CoreRenderer {
                             writer.writeAttribute("class", column.getStyleClass(), null);
                         }
 
-                        renderChildren(context, column);
+                        encodeIndexedId(context, column, index++);
                         writer.endElement("td");
                     }
                 }
--- a/primefaces/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java
@@ -476,7 +476,7 @@ public class SelectOneMenuRenderer extends SelectOneRenderer {
                         writer.writeAttribute("class", styleClass, null);
                     }
 
-                    renderChildren(context, column);
+                    encodeIndexedId(context, column, i);
                     writer.endElement("td");
                 }
             }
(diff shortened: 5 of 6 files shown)


primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
```diff
@@ -567,6 +569,34 @@ public static boolean isNestedWithinIterator(UIComponent component) {
...
+
+        // encode the component
+        component.encodeAll(context);
+
+        // restore the original writer
+        context.setResponseWriter(writer);
+
+        // append index to all id's
+        String encodedComponent = fsw.toString();
+        encodedComponent = encodedComponent.replaceAll("\\sid=\"(.*?)\"", " id=\"$1:" + index + "\"");
```
fcorneli, line 595: This captures also all the possible Javascript output?
How about the accidental single quote?

tandraschko:
we should really think about it, if we should really do this - also this way
duplicate ids is not a problem in general AFAIR, just a "bad pattern"

capturing the response + regex + write it is a performance downside

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
melloware: I don't think so I am specifically looking for HTML ID's only and theoretically JS code in these limited spots we do this should in theory not be looking for an exact id by name like `$('#form:abc`)` as it wouldn't find the correct one anyway as the Jquery spec by ID will return ONLY the first element matching that ID so the code would be wrong in this case either way.

melloware:
@tandraschko while I agree this is not ideal performance wise to me it seems like something we have to fix for the following reasons.

Cited: https://softwareengineering.stackexchange.com/questions/127178/two-html-elements-with-same-id-attribute-how-bad-is-it-really

1. PF is a library and it should always generate valid HTML IMHO. Here is a quote from a Stack Overflow user _"I would not consider using a template library that creates invalid HTML. "_

2. It a violation of the HTML4 and 5 spec.

3. Just because browsers are forgiving right now it doesn't mean they always will be or will have an HTML5 Strict mode in the future.

4. Behavior trying to use that ID as a selector, either in css or javascript, is unguessable and probably varies from browser to browser.

5. Experience says that `getElementById` in major browsers will return the first matched element in the document. But this may not always be the case in the future.

cc @VsevolodGolovanov @jepsar @christophs78 

I would love for others to weigh in here on whether this slight performance impact is worth generating valid HTML???

tandraschko:
its not just about performance - in case of a JSF library, its not the correct way of implementing this 
We actually should utilize stuff like UIRepeat / UINamingContainer, everything else is a hack and the ID is not available on the server side
And thats a lot of work to implement this

Therefore no idea if:
1) should we just ignore it until browser doesnt support it anymore
2) commit a fast workaround
3) do it the right way, which takes a long time

jepsar:
Yeah, I think a naming container is the way to go.

melloware:
> Yeah, I think a naming container is the way to go.

If its something that can be generically done that would be great because it seems there are at least 4 places where a PF component uses `UiColumn` for "advanced rendering" of a list that have this problem.  The rest of the component is fine in normal mode its just these Advanced modes using Columns that seem to have the issue.

jepsar:
> > Yeah, I think a naming container is the way to go.
> 
> If its something that can be generically done that would be great because it seems there are at least 4 places where a PF component uses `UiColumn` for "advanced rendering" of a list that have this problem. The rest of the component is fine in normal mode its just these Advanced modes using Columns that seem to have the issue.

Can't we introduce a new component you should use instead of `p:column`?

melloware:
Whiteout breaking backward compatibility would be nice if we could just inject something into the components something like if using Columns then wrap those <p:column> in a naming container automatically for them or something like that.  Basically make it completely backwards compatible.

melloware:
Closing this PR and have created new issue where we can track this discussion etc.

PR #8768, opened 2022-05-07
Fix #8767: Dialog Framework: fitViewport not recognised

### LINKED ISSUE #8767: Dialog Framework: fitViewport not recognised
### Describe the bug

fitViewport is documented and sent to the client, but the client-side JS does not recognize it.

### Reproducer

enhance primefaces-showcase/src/main/webapp/ui/df/viewProducts.xhtml

```xhtml
<h:body>
    <div style="background-color: red; height: 1500px">

    </div>

    <p:dataTable var="product" value="#{dfProductsView.products}">
        <p:column headerText="Code">
            <h:outputText value="#{product.code}"/>
        </p:column>
```

and enhance primefaces-showcase/src/main/java/org/primefaces/showcase/view/df/DFView.java
```java
public void viewProducts() {
    Map<String, Object> options = new HashMap<>();
    options.put("resizable", false);
    options.put("fitViewport", true);
    PrimeFaces.current().dialog().openDynamic("viewProducts", options, null);
}
```

### Expected behavior

there should be a vertical scrollbar

### PrimeFaces edition

Community

### PrimeFaces version

12.0.0-SNAPSHOT

### Theme

_No response_

### JSF implementation

_No response_

### JSF version

_No response_

### Browser(s)

_No response_

christophs78:


christophs78:
The second part of this issue is on the CSS-side of thing:

dialog.css defines
`.ui-dialog .ui-dialog-content.ui-df-content {overflow: hidden; padding: 0;}`

Not sure why `overflow` is set to `hidden`. To get it working it´s required to set overflow to auto with custom CSS. Should we modify CSS OOTB?

melloware:
I wonder if it is because DialogFramework renders inside an IFRAME right? But normal dialogs do not?  I remember there being some difference.  so for Dialog Framework dialogs maybe we need some extra OOTB CSS?

christophs78:
There seem´s to be miss-alignment between `max-height` of the outer `div` and `height` of the `iframe`. Together with `overflow: hidden` it creates sub-optimal result (no vertical scrollbar):
image

Removing `overflow:hidden` brings back the vertical scrollbar as expected.
image

Maybe it would be possible the improve our JS-code an set height or (max-height) of the iframe alligned with max-height of the outer div. But always at some risk of fixing this issue and creating some new issues.

christophs78:
@andresteingress: Maybe you can have a look at this PR and add your thoughts.

andresteingress:
@christophs78 the first commit is exactly the fix that I applied in our project. Regarding the overflow issue: I did not recognise that in our use-case. We have a dialog that basically has a search form and a data-table displaying the results. In that scenario, the scrollbar is shown. However, when resizing the window, the scrollbar is slightly "cut off" at the bottom, but that could as well be caused by one of our CSS styles. I need to investigate there, but your fix looks good to me.

andresteingress:
@christophs78 just investigated, the "cut off" vertical scrollbar is due to the fact that we use a relative width and height (both are set to 98%). It's exactly those 2% which are missing in height, when the window is resized and `fitViewport` is called. However, in our case, we can live with that.

melloware (approved):

PR #8789, opened 2022-05-12
Fix #8369: Column: Add converter attribute for filter and short notation

### LINKED ISSUE #8369: Column: Add converter attribute for filter and short notation
So you can do
```xhtml
<p:column headerText="ID lt" field="id" filterMatchMode="lt" converter="javax.faces.Integer"/>
```
instead of
```xhtml
<p:column headerText="ID lt" field="id" filterMatchMode="lt">
    <f:facet name="filter">
         <p:inputText onkeyup="PF('wgtTable').filter()">
              <f:converter converterId="javax.faces.Integer" />
          </p:inputText>
     </f:facet>
</p:column>
```

jepsar:
Fix #8369

### DIFF
--- a/docs/12_0_0/components/column.md
+++ b/docs/12_0_0/components/column.md
@@ -22,7 +22,7 @@ treetable and more.
 | ariaHeaderText | null | String | Label to read by screen readers, when not specified headerText is used.
 | caseSensitiveSort         | false              | Boolean          | Case sensitivity for sorting, insensitive by default.
 | colspan | 1 | Integer | Defines the number of columns the column spans.
-| converter | null | Converter/String | An el expression or a literal text that defines a converter for the component. When it’s an EL expression, it’s resolved to a converter instance. In case it’s a static text, it must refer to a converter id
+| converter | null | Converter/String | An EL expression or a literal text that defines a converter for the component. When it’s an EL expression, it’s resolved to a converter instance. In case it’s a static text, it must refer to a converter id
 | displayPriority           | 0                  | Integer          | Defines the display priority, in which order the columns should be displayed. Lower values have more priority.
 | draggable | true | Boolean | Defines if column is draggable if draggableColumns is set. Default true.
 | exportable | true | Boolean | Defines if the column should be exported by dataexporter.
--- a/docs/12_0_0/components/column.md
+++ b/docs/12_0_0/components/column.md
@@ -22,6 +22,7 @@ treetable and more.
 | ariaHeaderText | null | String | Label to read by screen readers, when not specified headerText is used.
 | caseSensitiveSort         | false              | Boolean          | Case sensitivity for sorting, insensitive by default.
 | colspan | 1 | Integer | Defines the number of columns the column spans.
+| converter | null | Converter/String | An el expression or a literal text that defines a converter for the component. When it’s an EL expression, it’s resolved to a converter instance. In case it’s a static text, it must refer to a converter id
 | displayPriority           | 0                  | Integer          | Defines the display priority, in which order the columns should be displayed. Lower values have more priority.
 | draggable | true | Boolean | Defines if column is draggable if draggableColumns is set. Default true.
 | exportable | true | Boolean | Defines if the column should be exported by dataexporter.
--- a/primefaces-integration-tests/src/main/webapp/datatable/dataTable026.xhtml
+++ b/primefaces-integration-tests/src/main/webapp/datatable/dataTable026.xhtml
@@ -20,34 +20,14 @@
             <p:dataTable id="datatable" widgetVar="wgtTable" value="#{dataTable026.employees}" var="employee"
                          filteredValue="#{dataTable026.filteredEmployees}" filterEvent="enter">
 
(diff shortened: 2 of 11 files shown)


melloware:
nice work and all Integration tests passing!

melloware (approved):

docs/12_0_0/components/column.md
```diff
@@ -22,6 +22,7 @@ treetable and more.
 | ariaHeaderText | null | String | Label to read by screen readers, when not specified headerText is used.
 | caseSensitiveSort         | false              | Boolean          | Case sensitivity for sorting, insensitive by default.
 | colspan | 1 | Integer | Defines the number of columns the column spans.
+| converter | null | Converter/String | An el expression or a literal text that defines a converter for the component. When it’s an EL expression, it’s resolved to a converter instance. In case it’s a static text, it must refer to a converter id
```
tandraschko: EL always upper case

primefaces/src/main/java/org/primefaces/component/column/ColumnBase.java
```diff
@@ -79,6 +83,8 @@ public enum PropertyKeys {
         width
     }
 
+    private Converter<?> converter;
```
tandraschko: im not sure if this caching works correctly with p:columns and converter=#{myModel.myConverterId} ?

docs/12_0_0/components/column.md
jepsar: I copied it. Was planning to find and replace it everywhere in a new PR. If I would do it now, it would pollute this PR.

primefaces/src/main/java/org/primefaces/component/column/ColumnBase.java
jepsar: I was struggling with the converter attribute to be honest. It now takes both a converter instance and a converterId... it all feels a bit like a hack TBO.

Regarding caching, I'll have a second look at it. Needs to be something like

https://github.com/apache/myfaces/blob/master/api/src/main/java/jakarta/faces/component/UIOutput.java
https://github.com/eclipse-ee4j/mojarra/blob/master/impl/src/main/java/jakarta/faces/component/UIOutput.java

tandraschko: I would not cache it for now and use stuff from ComponentUtils or introduce a new method there
And add to the Docs to use AppScoped converters

primefaces/src/main/java/org/primefaces/component/datatable/feature/FilterFeature.java
```diff
@@ -195,9 +195,11 @@ public void filter(FacesContext context, DataTable table) {
                 FilterConstraint constraint = filter.getConstraint();
                 Object filterValue = filter.getFilterValue();
                 if (filterValue instanceof String && column instanceof ColumnBase) {
-                    ColumnBase columnBase = (ColumnBase) column;
-                    if (columnBase.getConverter() != null) {
-                        filterValue = columnBase.getConverterObject().getAsObject(context, columnBase, (String) filterValue);
+                    try {
+                        filterValue = ComponentUtils.getConvertedValue(context, (ColumnBase) column, filterValue);
```
tandraschko: instead of accessing ColumnBase in the very generic ComponentUtils, maybe pass the converter via param and overload the method? Or even lambda if we need it lazy

melloware:
I really like where this PR landed using ComponentUtils!  Nice work.

melloware:
I really like where this PR landed using ComponentUtils!  Nice work.

tandraschko:
really looks good now, good work!
could you duplicate the test you changed? so we have one with the old style and one with the shortcut converter?

tandraschko:
really looks good now, good work!
could you duplicate the test you changed? so we have one with the old style and one with the shortcut converter?

jepsar:
> really looks good now, good work! could you duplicate the test you changed? so we have one with the old style and one with the shortcut converter?

I intentionally kept 1 column using the old style in the tests. Isn't that enough?

tandraschko:
didnt see it
+1 then, feel free to merge

primefaces-integration-tests/src/main/webapp/datatable/dataTable026.xhtml
```diff
@@ -20,34 +20,14 @@
...
-                </p:column>
+                <p:column headerText="ID lt" field="id" filterMatchMode="lt"
+                          converter="javax.faces.Integer"/>
+                <p:column headerText="ID lte" field="id" filterMatchMode="lte"
+                          converter="javax.faces.Integer"/>
+                <p:column headerText="ID gt" field="id" filterMatchMode="gt"
+                          converter="javax.faces.Integer"/>
+                <p:column headerText="ID gte" field="id" filterMatchMode="gte"
+                          converter="javax.faces.Integer"/>
                 <p:column headerText="ID equals" field="id" filterMatchMode="equals">
```
jepsar, line 31: FYI this test is still using the old verbose style

Issue #8857, opened 2022-06-03
Selenium: TimeoutException in complex mask with many update="@form"


nettozahler:
### Describe the bug

Let's try this (maybe) difficult one. Running a Selenium test on Firefox featuring a complex mask with many update="@form" fails with TimeoutException.

It does not happen always so it might be a timing issue or race condition or concurrency problem. I have seen it with

org.primefaces.selenium.component.InputText
org.primefaces.selenium.component.Calendar
org.primefaces.selenium.component.InputTextarea

but interestingly NOT with org.primefaces.selenium.component.InputNumber

In case you cannot reproduce it, here is some more information what happens as far as I understand it.

The PrimeSelenium Guard class sets this indicator before executing the command:
"pfselenium.xhr = 'somethingJustNotNull';"

In waitUntilAjaxCompletes it checks among other stuff:
pfselenium.xhr == null

In the error case it seems that pfselenium.xhr is not reset to null like this debug output shows:

Guard#ajax; ajaxDebugInfo **before methode.invoke**: document.readyState=complete, !window.jQuery=false, jQuery.active=0, !window.PrimeFaces=false, PrimeFaces.ajax.Queue.isEmpty()=true, PrimeFaces.animationActive=false, !window.pfselenium=false, **pfselenium.xhr=somethingJustNotNull**, pfselenium.anyXhrStarted=null, pfselenium.navigating=false

Guard#ajax; ajaxDebugInfo **in error catch clause**: document.readyState=complete, !window.jQuery=false, jQuery.active=0, !window.PrimeFaces=false, PrimeFaces.ajax.Queue.isEmpty()=true, PrimeFaces.animationActive=false, !window.pfselenium=false, **pfselenium.xhr=somethingJustNotNull**, pfselenium.anyXhrStarted=null, pfselenium.navigating=false

Therefore the condition for "Ajax complete" is never fulfilled and a TimeoutException occurs.

### Reproducer

Will provide a pull request with an appropriate integration test (hopefully) producing the error.

### Expected behavior

Selenium test runs fine without TimeoutException.

### PrimeFaces edition

Community

### PrimeFaces version

11

### Theme

Any

### JSF implementation

MyFaces

### JSF version

2.2, 2.3

### Browser(s)

Firefox 100

melloware:
I will say this definitely looks like a race condition.

[...]

PR #8858, opened 2022-06-03
Added a sometimes failing integration test for #8857

nettozahler:
Added a sometimes failing integration test for #8857

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/updateform/UpdateForm001.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/updateform/UpdateForm001.java
@@ -30,8 +30,6 @@ import java.util.Date;
 import javax.faces.view.ViewScoped;
 import javax.inject.Named;
 import javax.faces.event.AjaxBehaviorEvent;
-import org.primefaces.event.SelectEvent;
-import java.time.LocalDate;
 
 import lombok.Data;
 
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/updateform/UpdateForm001.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/updateform/UpdateForm001.java
@@ -1,0 +1,64 @@
+/*
+ * The MIT License
+ *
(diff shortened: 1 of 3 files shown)


melloware:
Can you re-sync with master and run this again?  I just made some fixes that fixed InputText and InputTextArea but locally I can't get the Calendar issue to happen.

nettozahler:
I hope I did the re-sync correctly since I am not yet familiar with the contribution workflow on GitHub.

melloware:
I will take a look tomorrow. I think something is still wrong with Calendar component.

melloware:
OK I got all your tests working locally I had to make some chnages to your test as some things were incorrect.  Can you re-sync with master again and we can watch it run?

melloware:
ok this is definitely something timing related or something because if I run this same test locally it passes.  Here its getting an AJAX timeout on the Calrendar `setValue` call for some reason.

stolp:
Welcome to hell. This matches my experiences with our integration tests.

Running tests locally has a much higher success rate as running them in a xvbf based test environment (both using Chrome). This may be related to the graphics performance of the local browser vs the slower performance of an emulated X-Framebuffer.

I spent some time trying to fix our tests today. Avoiding update="@form" helps getting fewer timeouts, but I cannot say that these are pinned to a certain component.

My gut feeling is, that the probability of hitting timeouts is more related to the amount of components being updated at one update action.

Currently, fixing tests is more like a game of Whack-a-mole. For every problem you solve a new one may evolve at a random other location at each new run.

We are on PrimeFaces/PrimeSelenium 11.

nettozahler:
Was there really a TimeoutException during the last test run? What I see in the log is a failed assertion: **UpdateForm001Test.testUpdateForm:59 expected: not null** This could be another manifestation of the root problem which I have also seen before: you set a field's value via PrimeSelenium and a short time later (e. g. when pressing "Submit") the previously entered field's value is suddenly missing.

melloware:
OK i have this test passing now except for 1 line `page.calendar.setValue(System.currentTimeMillis());`.

melloware:
OK now its all passing. I am going to merge this and we can watch it on the nightly build and see if we get "flickering" results.

PR #8861, opened 2022-06-05
Fix #8857: Calendar doesn't need AJAX protection on setDate

### LINKED ISSUE #8857: Selenium: TimeoutException in complex mask with many update="@form"

melloware:


tandraschko:
Can you explain? I thought this method fires the behavior event

melloware:
I looked in Calendar and it does not fire any event.  The `dateSelect` event is only fired when a date is picked from the popup panel.  It just does this...

```js
 setDate: function(date) {
        this.jqEl.datetimepicker('setDate', date);
    }
```

Setting the date on the jq picker.

melloware:
I will admit since we poured all our effort into DatePicker there are no Calendar tests at all so this has probably never worked.

PR #8866, opened 2022-06-07
Fix #8443/#8867: Menu model integration tests

### LINKED ISSUE #8443: Menubar/menu: might execute wrong (not clicked) items
**Describe the defect**
Clicking menu items while the ajax request is still in progress might lead to unexpected results. This might be exactly the case either in slow environment or if the user acts rapidly clicking fast. Please check the attached primefaces-test zip reproducing the problem, it explains the problem in details.

**Reproducer**
[primefaces-test-master.zip](https://github.com/primefaces/primefaces/files/8064098/primefaces-test-master.zip)

**Environment:**
 - PF Version: any
 - PF Theme: any
 - JSF + version: any
 - Affected browsers: any

**To Reproduce**
There are two examples in the attached app, just follow the instructions.

**Expected behavior**
I'd expect the menu behaving similar to the button component (the reproducer app includes the demonstration with buttons as well)

**The app screenshot giving a short overview**
image

### LINKED ISSUE #8867: Menu: Using RequestScoped model is broken
### Describe the bug

Using a `@RequestScoped` model breaks in 12.0.0 due to #8443

### Reproducer

[pf-8443.zip](https://github.com/primefaces/primefaces/files/8863936/pf-8443.zip)

Use the reproducer in `@ViewScoped` and everything works and with `@RequestScoped` everything fails.

### Expected behavior

MenuModel should be backwards compatible with previous PF while fixing bugs

### PrimeFaces edition

Community

### PrimeFaces version

12.0.0-SNAPSHOT

### Theme

ALL

### JSF implementation

All

### JSF version

ALL

### Browser(s)

ALL

melloware:
When using the MenuModel now that ticket #8443 has been implemented this Model must have a longer lived state than `RequestScoped`.

This is due to that fact that the old Menu had positional based lookup so "1_2" meant the first submenu and the 2nd menu item down. But now that we are uniquely identifying menu items so they can be looked up again the ID must remain so when the call is made from the UI it has to be able to find that `ID` again and not "assume its position".

### DIFF
--- a/docs/migrationguide/12_0_0.md
+++ b/docs/migrationguide/12_0_0.md
@@ -1,7 +1,7 @@
 # Migration guide 11.0.0 -> 12.0.0
 
 ## Core
-  * PrimeFlex 3.x is now the standard version of PrimeFlex that should be used. Use CLI migration tool to migrate from 2 to 3 (https://github.com/primefaces/primefaces/tree/master/primefaces-cli) 
+  * PrimeFlex 3.x is now the standard version of PrimeFlex that should be used. Use CLI migration tool to migrate from 2 to 3 (https://github.com/primefaces/primefaces/tree/master/primefaces-cli)
 
 ## DataTable
   * Row select mode `checkbox` is now `none`.
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/menubar/MenuBar002.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/menubar/MenuBar002.java
@@ -21,29 +21,28 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.primefaces.showcase.view.menu;
+package org.primefaces.integrationtests.menubar;
 
-import java.io.IOException;
 import java.io.Serializable;
-import java.util.concurrent.TimeUnit;
 
 import javax.annotation.PostConstruct;
-import javax.faces.application.FacesMessage;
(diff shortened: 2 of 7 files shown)


melloware:
cc @stolp

melloware:
@tandraschko should I add this to the release notes or should we look for a more clever way of storing the MenuModel?  Its doesn't seem like we can use ViewState for this?

tandraschko:
Hmm
Need to think about it, we may even revert or Look for a better alternative
Storing state should be then last alternative

melloware:
Yeah I think that is why it has always done that weird "0_1" positional thing which is also not ideal as it causes other issues.

christophs78:
From my personal view (you probably expect this from me) we should first add some more IT´s. We added a view (basic) MenuBar-IT´s back in april. (with static XHTML) Afterwards we can try to refactor.

melloware:
I will add an integration test for this scenario.

melloware:
Just to update here @stolp has gotten to the bottom of his issue and going to update a reproducer. He is basically creating his menu items dynamically on the fly not creating a model and storing it.  So it looks like the ID's will mismatch because everytime he re-generates there will be new and different UUID's is my suspicion.

melloware:
Added integration test using a `model` with AJAX and non-AJAX calls.

melloware:
Added failing `@RequestScoped` integration test proving the new bug

stolp:
I posted an updated reproducer for the dynamic case on Discord. Not sure where else to add it.

melloware:
That is fine I will check  it out and test my changes against your reproducer

melloware:
These tests will pass one the main fix PR is merged.

christophs78 (approved):

PR #8868, opened 2022-06-08
Fix #8867: Menu model restore legacy behavior as fallback

### LINKED ISSUE #8443: Menubar/menu: might execute wrong (not clicked) items

### LINKED ISSUE #8867: Menu: Using RequestScoped model is broken

melloware:
This makes it 100% compatible with scenarios from PF 11.0.0 and before. 

Basically the scenario this fixes is `@RequestScoped` models it falls back and uses the coordinates lookup like PF has always done.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/model/menu/BaseMenuModel.java
+++ b/primefaces/src/main/java/org/primefaces/model/menu/BaseMenuModel.java
@@ -28,6 +28,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
 
+import org.primefaces.util.Constants;
 import org.primefaces.util.LangUtils;
 
 /**
@@ -36,6 +37,7 @@ import org.primefaces.util.LangUtils;
 public class BaseMenuModel implements MenuModel, Serializable {
 
     public static final String ID_SEPARATOR = "_";
+    public static final String COORDINATES_SEPARATOR = "|";
 
     private static final long serialVersionUID = 1L;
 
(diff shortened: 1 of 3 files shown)


tandraschko:
TBH i dont like a new attribute
But i dont have enough time to think about it currently, we need to discuss in 2 or 3 weeks

melloware:
I can do it without the new attribute.  Was just trying to keep it separate I will fix.

melloware:
@tandraschko i have updated the PR and removed the new attribute. I have all tests passing with reproducers and Integration tests checking.

What I did was this...

**PF < 11:**
menuid was coordinates of menu like `2_1`

**PF 12 Update/Elite 11.0.5:**
menuid was made a UUID `561231c6-354a-4842-8086-81d00531fde3` however this does not work for RequestScoped beans as the menu id is lost.

**PF 12 With this fix:**
menuid is now`UUID|Coords` like `561231c6-354a-4842-8086-81d00531fde3|0_3` where it will try and use the ID and if the ID cannot be matched it falls back to the Coords check using `0_3` in the example above

PR #8873, opened 2022-06-09
Fix #8872: Faces 4.0 support of SameSite cookie

### LINKED ISSUE #8872: Cookies: Support SameSite=Strict for Faces 4.0
Currently PF creates 3 cookies that can be secured with SameSite=Strict in Jakarta Faces 4.0  

1. FileDownload Cookie
2. DataExporter cookie
3. Client Window Cookie

melloware:


tandraschko:
is there a reason to not always use strict like in MF?

melloware:
I originally had it as strict but provided this config failsafe just in case there were scenarios I coudn't think of why you would want to disable it or make it "Lax"?

tandraschko:
yeah ok
lets merge it and decide later, if we can remove or not

PR #8884, opened 2022-06-14
Fix #8883 Own DateTimeConverter should be considered

### LINKED ISSUE #8883: DatePicker: DateTimeConverter for custom date type no longer working with Primefaces 11
### Describe the bug

In our product we have our own Date, Time and DateTime classes. These classes has nothing to do with java.util.Date or the java.time.* classes. To support our own classes we have implemented own DateTimeConverters for them. These converters have worked fine in Primefaces 7 but do no longer with Primefaces 11.

The problem is located in the org.primefaces.util.CalendarUtils class in method getValue line 289. Here the type of the value are compared with the type that the converter provides. This is completely wrong as the type of the value is a Java class and the type the converter provides is not a Java class but a String that is either "date", "time" or "both".

The issue was introduced by https://github.com/primefaces/primefaces/pull/5925 and later modified by https://github.com/primefaces/primefaces/pull/6015.

The changes were made to fix a problem with the standard DateTimeConverter that is not able to handle java.time.* classes correctly. However the fix prevents that anybody can use their own DateTimeConverter for their own classes.

### Reproducer

_No response_

### Expected behavior

Our own DateTimeConverter should be called and not be ignored.

### PrimeFaces edition

Elite

### PrimeFaces version

11.0.2

### Theme

Freya

### JSF implementation

MyFaces

### JSF version

2.2.13

### Browser(s)

_No response_

weissreto:
Converter.getType() has nothing to do with the type of the value that
the converter can handle. Possible values are for type are 'date',
'time' and 'both'.

Instead, I try to use the converter and if it is not able to convert a
value I put it on a blacklist, so that the next time the Primefaces
codes is used. This is important since the standard DateTimeConverter
does not handle java.time.* classes correct. On the other hand our own
converter should be used for our own date time data types.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/util/CalendarUtils.java
+++ b/primefaces/src/main/java/org/primefaces/util/CalendarUtils.java
@@ -306,7 +306,8 @@ public class CalendarUtils {
             if (isGoodConverter(converter, value)) {
                 try {
                     return converter.getAsString(context, calendar, value);
-                } catch (ConverterException ex) {
+                }
+                catch (ConverterException ex) {
                     addBadConverter(converter, value);
                 }
             }
@@ -356,16 +357,16 @@ public class CalendarUtils {
     }
 
     private static void addBadConverter(Converter converter, Object value) {
-      Set<Class<? extends Converter>> converters = BAD_CONVERTERS.computeIfAbsent(value.getClass(), cl -> new HashSet<Class<? extends Converter>>());
-      converters.add(converter.getClass());
+        Set<Class<? extends Converter>> converters = BAD_CONVERTERS.computeIfAbsent(value.getClass(), cl -> new HashSet<Class<? extends Converter>>());
+        converters.add(converter.getClass());
     }
 
     private static boolean isGoodConverter(Converter converter, Object value) {
-      Set<Class<? extends Converter>> converters = BAD_CONVERTERS.get(value.getClass());
-      if (converters == null) {
-        return true;
-      }
-      return ! converters.contains(converter.getClass());
+        Set<Class<? extends Converter>> converters = BAD_CONVERTERS.get(value.getClass());
+        if (converters == null) {
+            return true;
+        }
+        return !converters.contains(converter.getClass());
     }
 
     /**
--- a/primefaces/src/main/java/org/primefaces/util/CalendarUtils.java
+++ b/primefaces/src/main/java/org/primefaces/util/CalendarUtils.java
@@ -63,7 +63,7 @@ public class CalendarUtils {
 
     private static final PatternConverter PATTERN_CONVERTER = new DateTimePatternConverter();
 
-    private static final Map<Class<?>, Set<Class<? extends Converter>>> BAD_CONVERTERS = new HashMap<>();
+    private static final String INVALID_CONVERTERS = CalendarUtils.class.getName() + ".invalidConverters";
 
     private CalendarUtils() {
     }
@@ -303,12 +303,12 @@ public class CalendarUtils {
         //first ask the converter
         Converter converter = calendar.getConverter();
         if (converter != null) {
-            if (isGoodConverter(converter, value)) {
+            if (isValidConverter(context, converter, value)) {
                 try {
                     return converter.getAsString(context, calendar, value);
                 }
                 catch (ConverterException ex) {
-                    addBadConverter(converter, value);
+                    addInvalidConverter(context, converter, value);
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


tandraschko:
1) i dont like a try-error approach; CC @christophs78 
2) a static cache can lead to mem leaks, a AppScoped map would be the way to go

christophs78:
Independent of the concrete solution please add a integration-test for this. This is the best way to avoid future issues regarding your kind of special usecase with custom DateTime-classes. 
(use e.g. https://github.com/primefaces/primefaces/blob/master/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datepicker/DatePicker013Test.java and https://github.com/primefaces/primefaces/blob/master/primefaces-integration-tests/src/main/webapp/datepicker/datePicker013.xhtml as template)

christophs78:
> 1. i dont like a try-error approach; CC @christophs78

Are we sure `javax.faces.convert.ConverterException` is only thrown in the case of invalid combination of class and coverter? (Or may it also thrown du to other reasons? https://docs.oracle.com/javaee/7/api/javax/faces/convert/ConverterException.html sounds a bit generic. With your modification we won´t call this converter anymore for a specific class after it has once thrown `javax.faces.convert.ConverterException`.)

> 2. a static cache can lead to mem leaks, a AppScoped map would be the way to go

weissreto:
I will try to provide an integration-test and also my converter. This will need some time. Hope I can work on it on Friday.

primefaces/src/main/java/org/primefaces/util/CalendarUtils.java
```diff
@@ -52,6 +63,8 @@ public class CalendarUtils {
 
     private static final PatternConverter PATTERN_CONVERTER = new DateTimePatternConverter();
 
+    private static final Map<Class<?>, Set<Class<? extends Converter>>> BAD_CONVERTERS = new HashMap<>();
```
melloware: INVALID_CONVERTERS

primefaces/src/main/java/org/primefaces/util/CalendarUtils.java
```diff
@@ -346,6 +356,19 @@ else if (value instanceof LocalTime) {
         }
     }
 
+    private static void addBadConverter(Converter converter, Object value) {
```
melloware: addInvalidConverter

primefaces/src/main/java/org/primefaces/util/CalendarUtils.java
```diff
@@ -346,6 +356,19 @@ else if (value instanceof LocalTime) {
...
+        Set<Class<? extends Converter>> converters = BAD_CONVERTERS.computeIfAbsent(value.getClass(), cl -> new HashSet<Class<? extends Converter>>());
+        converters.add(converter.getClass());
+    }
+
+    private static boolean isGoodConverter(Converter converter, Object value) {
```
melloware: addValidConverter

melloware (changes requested):

weissreto:
> > 1. i dont like a try-error approach; CC @christophs78
> 
> Are we sure `javax.faces.convert.ConverterException` is only thrown in the case of invalid combination of class and coverter? (Or may it also thrown du to other reasons? https://docs.oracle.com/javaee/7/api/javax/faces/convert/ConverterException.html sounds a bit generic. With your modification we won´t call this converter anymore for a specific class after it has once thrown `javax.faces.convert.ConverterException`.)
> 
The method getAsString from javax.faces.convert.DateTimeConverter only throws that Exception if a given Object cannot be converted to a String. Means DateFormat.format method throws an Exception. I don't see any other reason this Exception can be raised by the getAsString method of that class.

weissreto:
Here is our faces-config.xml with the javax.faces.DateTime converter configuration. And the implementation of our DateTimeConverter class. Also The JavaDateToIvyMappers that is used to map java.util.Date to our own Date, Time and DateTime classes.
[DateTimeConverter.zip](https://github.com/primefaces/primefaces/files/8925274/DateTimeConverter.zip)
w
Integration Test will come soon ...

melloware:
Nice Integration test!   Can you address my code review comments as well renaming "bad" and "good" to "valid" and "invalid"

melloware:
OK I am rerunning the integration tests now.

melloware (approved):

weissreto:
Thanks for your help and merging

melloware:
Thanks for the contribution and integration test!

tandraschko:
TBH i dont think this is an good change
A ConverterException says that the current value is not convertable -> e.g. a bad input on the UI and a error message will be shown the user
we now ignore the converter forever

either we introduce a own exception type or look for a nice and correct fix

user applied converters, doesnt matter if via annotation or component attr should be respected first

PR #8888, opened 2022-06-16
Feature Request #8887: Re-enabled row selection by button 

### LINKED ISSUE #8887: DataTable: Enable row selection per button even if disabledSelection="true"
### Description

When we worked with PrimeFaces 8.0 it was possible to set disabledSelection="true" on a p:dataTable so that the selection of a row was only possible by clicking on a designated button. With PrimeFaces 10.0 this was not possible anymore because if disabledSelection="true" was set, no rows were highlighted anymore, despite the current selection. So we could use that feature only with a nasty workaround. 

I think the use case for this is still valid, and maybe it can be re-enabled. I will attach a PR. Let me know what you think about it. 

### Additional context

For example, I use the dataTable as follow: 
`
<p:dataTable var="car"
           value="#{dtBasicView.cars}"
           multiViewState="true" 
           rowKey="#{car.number}"
           selection="#{dtBasicView.selectedCar}"
           selectionMode="single" 
           disabledSelection="true"
           rows="60" paginator="true"
           lazy="false"
`

Important is disabledSelection="true" so that the user can not select rows per click on a row or cell. It should only be possible with a click on a button that is included in every row. 

That is made with a column that include a button like this:
` 
<p:column
            headerText="Action"
            width="80">
            <p:commandButton value="Select"
                  update="@all"
                  process="@this">
               <f:setPropertyActionListener value="#{car}" target="#{dtBasicView.selectedCar}" />
            </p:commandButton>
       </p:column>
`

That is like an example which is shown in the PrimeFaces documentation. So with PrimeFaces 10 I can select the row with that button, but the selected row is **not** highlighted as it should, which worked with PrimeFaces 8 (I also checked PF11 and the latest code base). 

So my changes make that in that case the selected row is now highlighted.

I hope this will help to understand why I need that change. If you need more information, I will be happy to provide some.

marw1tt:
Re-enabled row selection by button when disabledSelection="true" on p:dataTable.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
@@ -1227,7 +1227,7 @@ public class DataTableRenderer extends DataRenderer {
             throws IOException {
 
         ResponseWriter writer = context.getResponseWriter();
-        boolean selectionEnabled = table.isSelectionEnabled() && !table.isDisabledSelection();
+        boolean selectionEnabled = table.isSelectionEnabled();
         boolean rowExpansionAvailable = table.getRowExpansion() != null;
         String rowKey = null;
         List<UIColumn> columns = table.getColumns();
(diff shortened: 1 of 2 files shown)


melloware (approved):

melloware:
I am OK with this.  @Rapster would you like to review?

Rapster:
I can review, just need some explanation to understand the issue

marw1tt:
For example, I use the dataTable as follow: 
`
<p:dataTable var="car"
           value="#{dtBasicView.cars}"
           multiViewState="true" 
           rowKey="#{car.number}"
           selection="#{dtBasicView.selectedCar}"
           selectionMode="single" 
           disabledSelection="true"
           rows="60" paginator="true"
           lazy="false"
`

Important is disabledSelection="true" so that the user can not select rows per click on a row or cell. It should only be possible with a click on a button that is included in every row. 

That is made with a column that include a button like this:
` 
<p:column
            headerText="Action"
            width="80">
            <p:commandButton value="Select"
                  update="@all"
                  process="@this">
               <f:setPropertyActionListener value="#{car}" target="#{dtBasicView.selectedCar}" />
            </p:commandButton>
       </p:column>
`

That is like an example which is shown in the PrimeFaces documentation. So with PrimeFaces 10 I can select the row with that button, but the selected row is **not** highlighted as it should, which worked with PrimeFaces 8 (I also checked PF11 and the latest code base). 

So my changes make that in that case the selected row is now highlighted.

I hope this will help to understand why I need that change. If you need more information, I will be happy to provide some.

Rapster:
Please see conversation in original issue where I gave you a feedback :wink:

PR #8914, opened 2022-06-21
Fix #8910: Optimize JS bundling

### LINKED ISSUE #8910: More lightweight PrimeFaces
### Description

Actually Primefaces JS files (components.js, jquery-plugin.js) are bigger and bigger along with following releases. In our company we use only few primefaces components, and we want our front to be as fast as possible - yes I know that in that case Primefaces is not the best option, but still we want to stay with it because of many reasons.

Maybe it will be possible to choose widgets we can enable or disable in settings and then on the front side only needed JS code will be loaded in both components.js and jquery-plugin.js.

### Additional context

_No response_

melloware:
By changing the following things was able to significantly reduce the size of the 3 major JS files

- Jquery Plugins only download the 35 necessary widgets
- Calendar now contains all the hacks and Jquery DatePicker/TimePicker in its own JS file
- Lightbox moved to its own JS file since it is deprecated
- Datatable moved to its own JS since its the largest component at 200KB

| Version  | jQuery Plugins |  Core  | Components  |
|--------|-------------|------|-----------|
| 10       | 262 KB         | 55 KB  | 477 KB      |
| 11       | 363 KB         | 90 KB  | 500 KB      |
| 12       | 146 KB         | 88 KB  | 396 KB      |

cc @jacekzwroclawia @stolp @tandraschko


primefaces/src/main/resources/META-INF/resources/primefaces/calendar/3-jquery.ui.datepicker.css
```diff
@@ -0,0 +1,266 @@
+/*! jQuery UI - v1.13.1 - 2022-06-21
+* http://jqueryui.com
+* Includes: core.css, datepicker.css
+* Copyright jQuery Foundation and other contributors; Licensed MIT */
+
+/* Layout helpers
+----------------------------------*/
+.ui-helper-hidden {
```
tandraschko: layout helpers are not related to datepicker only

primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java
```diff
@@ -61,8 +61,9 @@
 @ResourceDependency(library = "primefaces", name = "jquery/jquery.js")
 @ResourceDependency(library = "primefaces", name = "jquery/jquery-plugins.js")
 @ResourceDependency(library = "primefaces", name = "core.js")
-@ResourceDependency(library = "primefaces", name = "components.js")
 @ResourceDependency(library = "primefaces", name = "touch/touchswipe.js")
+@ResourceDependency(library = "primefaces", name = "datatable/datatable.js")
```
tandraschko: why making datatable a own file? one of the most used components

melloware: Only because its 200K so it reduces components.js by 100K and we do have pages that don't use DataTables so its loading that extra for no reason?  I think once a component is over 100K it should always be in its own file?

primefaces/src/main/resources/META-INF/resources/primefaces/calendar/3-jquery.ui.datepicker.css
melloware: OK this came down by default when I downloaded DatePicker by itself from Jquery UI let me remove.

primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java
tandraschko: the idea was to bundle most common components - ignoring size or everything else
i would not split it up
better remove uncommon components in future

melloware: OK Datatable changes reverted.

PR #8915, opened 2022-06-21
Fix #8912: DatePicker default view date inside min-max range

### LINKED ISSUE #8912: DatePicker: Initial view with maxdate in the past/mindate in the future
### Describe the bug

Because calendar widget is deprecated - I want to migrate to datePicker component, however I encoutered problems:

**TC 1.** When maxdate attribute is in the past and value is empty (not selected yet)

`<p:datePicker yearRange="1900:2021" maxdate="31-12-2021" pattern="dd-MM-yyyy" />`

When I open datePicker it shows current month with empty year filter, but I don't want this because our maxdate is in the past:

image

With deprecated `<p:calendar .. />` it works as expected - in that situation it shows last month from maxdate, not current month:

image

**TC 2.** When mindate is in the future

W don't use that case in our system, but it doesn't work properly on both p:calendar and p:datePicker.

### Reproducer

_No response_

### Expected behavior

Opened month in datePicker should take into accunt maxdate and mindate attributes values.

### PrimeFaces edition

Community

### PrimeFaces version

11.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.2

### Browser(s)

Chrome 102

melloware:
This ticket from @christophs78 already had this code in there but was only doing it for Time Only mode.  It now does it for both modes so the default view date is not `new Date()` if it does not fit inside the min max range.

Original Ticket: https://github.com/primefaces/primefaces/issues/5334

### DIFF
--- a/primefaces-integration-tests/src/main/webapp/datepicker/datePicker010.xhtml
+++ b/primefaces-integration-tests/src/main/webapp/datepicker/datePicker010.xhtml
@@ -14,6 +14,12 @@
         <h:form id="form">
             <p:datePicker id="datepicker" value="02.06.2021"
                           mindate="27.05.2021" maxdate="10.06.2021" pattern="dd.MM.yyyy"/>
+                          
+            <!-- GitHub #8912 default view date with no value should be October 2021 not NOW -->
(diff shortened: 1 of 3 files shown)


melloware:
Added Integration Test

PR #8918, opened 2022-06-22
Fix #8917: SelectOneMenu should include item groups during filter

### LINKED ISSUE #8917: SelectOneMenu: Include SelectItemGroup in filter
### Description

Currently the filter only searches SelectItems but no SelectItemGroups

It would be nice, if you could also filter for SelectItemGroups and it shows all items for that group.

### Additional context

A possible solution could be something like 

https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectonemenu.js#L1300-L1303

```javascript
if(this.filterMatcher(itemLabel, filterValue)) {
     show.push(item);
}
else if(!item.is('.ui-selectonemenu-item-group-children')) { // TODO configurable? this.cfg.filterIncludesGroup && ...
    hide.push(item);
}
else {
    itemLabel = this.cfg.caseSensitive ? option.parent().attr('label') : option.parent().attr('label').toLowerCase();
    if (this.filterMatcher(itemLabel, filterValue)) {
        show.push(item);
    }
    else {
        hide.push(item);
    }
}
```

This solution is quite simple and not the most performant. And maybe this should be configurable?

fanste:


melloware (approved):
Just waiting for build to run and complete

PR #8964, opened 2022-07-06
Fixes #8961 - DataExporter: Permit using custom DecimalFormat in ExcelOptions

### LINKED ISSUE #8961: DataExporter: Permit using custom DecimalFormat in ExcelOptions
### Description

It would be nice to allow a custom DecimalFormat to be used only in Excel files and not directly related to the locale used in JSF.

Add a new property in `ExcelOptions` called `decimalFormat` of type java.text.DecimalFormat which should be used when exporting to Excel files (xls, xlsx, xlsxstream)

### Additional context

_No response_

jasonex7:
Fixes #8961

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/export/ExcelOptions.java
+++ b/primefaces/src/main/java/org/primefaces/component/export/ExcelOptions.java
@@ -23,6 +23,8 @@
  */
 package org.primefaces.component.export;
 
+import java.text.DecimalFormat;
+
 public class ExcelOptions implements ExporterOptions {
 
     private String facetFontStyle;
@@ -44,6 +46,8 @@ public class ExcelOptions implements ExporterOptions {
     private boolean autoSizeColumn = true;
 
     private boolean stronglyTypedCells = true;
+
+    private DecimalFormat decimalFormat;
 
     public ExcelOptions() {
     }
@@ -156,5 +160,13 @@ public class ExcelOptions implements ExporterOptions {
 
     public void setStronglyTypedCells(boolean stronglyTypedCells) {
         this.stronglyTypedCells = stronglyTypedCells;
+    }
+
+    public DecimalFormat getDecimalFormat() {
+        return decimalFormat;
+    }
+
+    public void setDecimalFormat(DecimalFormat decimalFormat) {
+        this.decimalFormat = decimalFormat;
     }
 }
--- a/primefaces/src/main/java/org/primefaces/util/CurrencyValidator.java
+++ b/primefaces/src/main/java/org/primefaces/util/CurrencyValidator.java
@@ -27,7 +27,6 @@ import java.io.Serializable;
 import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.text.*;
-import java.util.Locale;
 
 /**
  * <p>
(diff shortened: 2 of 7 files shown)


melloware:
I will code review and test this in the morning.

primefaces/src/main/java/org/primefaces/util/CurrencyValidator.java
```diff
@@ -116,8 +115,9 @@ protected Object parse(String value, DecimalFormat formatter) {
                     buffer.append(pattern.charAt(i));
                 }
             }
-            formatter.applyPattern(buffer.toString());
-            parsedValue = parseFormat(value, formatter);
+            DecimalFormat copyFormatter = (DecimalFormat) formatter.clone();
```
melloware, line 118: Why clone here?

primefaces/src/main/java/org/primefaces/util/CurrencyValidator.java
```diff
@@ -161,9 +161,9 @@ protected Object parseFormat(String value, DecimalFormat formatter) {
      * @return The parsed <code>Number</code> converted to a <code>BigDecimal</code>.
      */
     protected BigDecimal processParsedValue(Number value, DecimalFormat formatter) {
-        BigDecimal decimal = null;
+        BigDecimal decimal;
         if (value instanceof Long) {
-            decimal = BigDecimal.valueOf(((Long) value).longValue());
```
melloware, line 166: Don't we want to keep the long value here?

primefaces/src/main/java/org/primefaces/component/export/ExcelOptions.java
```diff
@@ -45,6 +47,8 @@ public class ExcelOptions implements ExporterOptions {
 
     private boolean stronglyTypedCells = true;
 
+    private DecimalFormat decimalFormat;
+
```
melloware, line 51: Can you add a note in `\docs\12_0_0\gettingstarted\whatsnew.md` stating this new option exists in ExcelOptions for DataExporter.

Can you update `\docs\12_0_0\components\dataexporter.md` at the bottom where it shows the custom excel options add DecimalFormat in that example?

melloware (changes requested):

primefaces/src/main/java/org/primefaces/util/CurrencyValidator.java
jasonex7: It's because the `applyPattern` method is a modifier, so in order to not alter the original decimal format we clone it and modify that instead. That allow the user of CurrencyValidator to reuse the passed DecimalFormat.

melloware: But wont this be called over and over again on every cell cloning it every time?  Wouldn't it be better to clone it once?

primefaces/src/main/java/org/primefaces/util/CurrencyValidator.java
jasonex7: In this case the autoboxing feature (since Java 5) of the JVM do the unboxing for us and improves code readability.

primefaces/src/main/java/org/primefaces/util/CurrencyValidator.java
jasonex7: I don't know how to clone once inside CurrencyValidator class because "formatter" variable is a method parameter and CurrencyValidator is a singleton and for that cannot be altered without affected another user of the same instance. Maybe if CurrencyValidator was not a singleton we can receive DecimalFormat in the constructor once and I could clone it once, but in that scenario maybe there would be more instances of CurrencyValidator, once per DecimalFormat.

primefaces/src/main/java/org/primefaces/component/export/ExcelOptions.java
jasonex7: Of course, please check now.

melloware (approved):

PR #8971, opened 2022-07-07
Fixes #8970 - DataExporter: Align numeric cells at right in Excel export

### LINKED ISSUE #8970: DataExporter: Align numeric cells at right in excel export
### Description

Numeric cells that are not currency are aligned to left instead of right. Currency cells are already aligned to right.

image

image

### Additional context

I'm going to provide a PR.

jasonex7:
Fixes #8970


melloware (approved):
Excellent refactoring!!!

cocorossello:
The problem with this refactor is that now poi dependency is not enough, even if it's a simple pdf. The poi-ooxml is required now, it wasn't in 11.0.0

PR #8981, opened 2022-07-09
Fix #8950: DatePicker: year navigator input type number

### LINKED ISSUE #8950: DatePicker: use input number as year navigator
### Description

When you use the date picker to select a date of birth, you may need to navigate to years ago. With the current selector you either need to include a lot of years, or you need to navigate multiple times.

Instead of a select input it would be better (IMO) to have a numeric input.
```html
<input type="number" min="..." max="..." step="1" value="..."/>
```
This allows you to spin up and down, but also to simply enter a year.

### Additional context

_No response_

jepsar:
Fix #8950

This is working. However, somehow the spinner buttons are not showing on the input number. I don't get why. After some research it looks like they could be disabled by CSS, but I cannot find where. They are not inspectable using the DOM inspector (at least I didn't find a way), so it's hard to debug.

### DIFF
--- a/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datepicker/DatePicker011Test.java
+++ b/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datepicker/DatePicker011Test.java
@@ -293,7 +293,7 @@ public class DatePicker011Test extends AbstractDatePickerTest {
         }
 
         // Act - 7th next year via year drop down
-        datePicker.selectYearDropdown(LocalDate.now().getYear() + 1);
+        datePicker.navigateYear(LocalDate.now().getYear() + 1);
 
         // Assert
         PrimeSelenium.wait(SAFETY_WAIT_AFTER_DOUBLE_AJAX_CALL_MILLISECONDS);
--- a/primefaces-selenium/primefaces-selenium-components/src/main/java/org/primefaces/selenium/component/DatePicker.java
+++ b/primefaces-selenium/primefaces-selenium-components/src/main/java/org/primefaces/selenium/component/DatePicker.java
@@ -287,7 +287,7 @@ public abstract class DatePicker extends AbstractInputComponent {
     }
 
     /**
-     * Select a month from the drodown.
+     * Select a month from the dropdown.
      *
      * @param month the month to select
      */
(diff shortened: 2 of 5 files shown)


melloware:
Have you tried it with `primefaces.THEME` set to `none` to turn off the PF CSS to see if its something with that?

jepsar:
https://github.com/primefaces/primefaces/blob/d09ef48aa15aeda31678eb0c80d59bbd8349674b/primefaces-showcase/src/main/webapp/resources/showcase/css/layout/_core.scss#L68-L76

I'll have a look at the ITs

melloware:
Oh OK thta was added for the FloatLabel but in 12.0.0-SNAPSHOT I added a FloatLabel showcase page with every component that supports floatlabel.  Not sure why in particular it hid those for number fields?

melloware:
Actually I didn't add that looks like that has been in that Showcase CSS for a while but no idea why it would be needed.

jepsar:
I tried to fix the IT, but I cannot fix these errors:
```
[ERROR] Errors: 
[ERROR]   DatePicker011Test.testLazyMetaDataNoBehaviour1:147->testDatePickerPart1:296 » Timeout
[ERROR]   DatePicker011Test.testLazyMetaDataViewChangeBehaviour1:155->testDatePickerPart1:296 » Timeout
[ERROR]   DatePicker011Test.testMetaDataCloseBehaviour1:83->testDatePickerPart1:296 » Timeout
[ERROR]   DatePicker011Test.testMetaDataDateSelectBehaviour1:67->testDatePickerPart1:296 » Timeout
[ERROR]   DatePicker011Test.testMetaDataNoBehaviour1:59->testDatePickerPart1:296 » Timeout
[ERROR]   DatePicker011Test.testMetaDataViewChangeBehaviour1:75->testDatePickerPart1:296 » Timeout
```

jepsar:
> Actually I didn't add that looks like that has been in that Showcase CSS for a while but no idea why it would be needed.

For now I explicitly show spinners on the year input of a date picker.

jepsar:
Actually, I think I'll be closing the issue as won't fix. I'm not convinced that the input is a great improvement. Rerendering of the calendar (including the input) makes using the spinner awkward. And clearing the year will get you an empty calendar.

PR #8988, opened 2022-07-11
Fixes #8987 - DataExporter: OutputFormat doesn't render correctly on exports

### LINKED ISSUE #8987: DataExporter: OutputFormat doesn't render correctly on exports
### Describe the bug

When exporting a DataTable that contains <h:outputFormat> in header or footer facet, the exported value doesn't render correctly.

image

### Reproducer

```xhtml
<p:dataTable id="table" var="r" value="#{testView.records}">
    <f:facet name="header">
        <h:outputFormat value="{0} - {1}">
            <f:param value="#{testView.title}" />
            <f:param value="#{testView.customMessage}" />
        </h:outputFormat>
    </f:facet>
    <p:column headerText="Text Value">
        <h:outputText value="#{r.textValue}" />
    </p:column>
    <p:column headerText="Int Value">
        <h:outputText value="#{r.intValue}">
            <f:convertNumber type="number" />
        </h:outputText>
    </p:column>
    <f:facet name="footer">
        <h:outputFormat value="{0,number,#,##0} records.">
            <f:param value="#{testView.records.size()}" />
        </h:outputFormat>
    </f:facet>
</p:dataTable>

<h:commandLink immediate="true">
    <h:outputText value="xlsx" />
    <p:dataExporter type="xlsx" target="table" fileName="test" />
</h:commandLink>
```

### Expected behavior

image

### PrimeFaces edition

_No response_

### PrimeFaces version

11.0.0, 12.0.0-SNAPSHOT

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Browser(s)

_No response_

jasonex7:
Fixes #8987

### DIFF
--- a/primefaces/src/main/java/org/primefaces/renderkit/RendererUtils.java
+++ b/primefaces/src/main/java/org/primefaces/renderkit/RendererUtils.java
@@ -24,15 +24,8 @@
 package org.primefaces.renderkit;
 
 import java.io.IOException;
-import javax.faces.FactoryFinder;
-import javax.faces.application.Application;
-import javax.faces.application.ViewHandler;
-import javax.faces.component.UIViewRoot;
 import javax.faces.context.FacesContext;
 import javax.faces.context.ResponseWriter;
-import javax.faces.render.RenderKit;
-import javax.faces.render.RenderKitFactory;
-
 import org.primefaces.util.HTML;
 
 public class RendererUtils {
@@ -75,48 +68,6 @@ public class RendererUtils {
         writer.endElement("div");
 
         writer.endElement("div");
-    }
-
-    /**
-     * Duplicate code from OmniFaces project under apache license:
-     * <a href="https://github.com/omnifaces/omnifaces/blob/master/license.txt">https://github.com/omnifaces/omnifaces/blob/master/license.txt</a>
-     *
-     * Returns the {@link RenderKit} associated with the "current" view ID or view handler.
-     * <p>
-     * The current view ID is the view ID that's set for the view root that's associated with the current faces context.
-     * Or if there is none, then the current view handler will be assumed, which is the view handler that's associated
-     * with the requested view.
-     *
-     * @return The {@link RenderKit} associated with the "current" view ID or view handler.
-     * @throws NullPointerException When faces context is unavailable.
-     * @see <a href="https://github.com/omnifaces/omnifaces">Omnifaces</a>
-     */
-    public static RenderKit getRenderKit(FacesContext context) {
-        String renderKitId = null;
-        UIViewRoot view = context.getViewRoot();
-
-        if (view != null) {
-            renderKitId = view.getRenderKitId();
-        }
-
-        if (renderKitId == null) {
-            Application application = context.getApplication();
-            ViewHandler viewHandler = application.getViewHandler();
-
-            if (viewHandler != null) {
-                renderKitId = viewHandler.calculateRenderKitId(context);
-            }
-
-            if (renderKitId == null) {
-                renderKitId = application.getDefaultRenderKitId();
-
-                if (renderKitId == null) {
-                    renderKitId = RenderKitFactory.HTML_BASIC_RENDER_KIT;
-                }
-            }
-        }
-
-        return ((RenderKitFactory) FactoryFinder.getFactory(FactoryFinder.RENDER_KIT_FACTORY)).getRenderKit(context, renderKitId);
     }
 
 }
--- a/primefaces/src/main/java/org/primefaces/renderkit/RendererUtils.java
+++ b/primefaces/src/main/java/org/primefaces/renderkit/RendererUtils.java
@@ -24,8 +24,15 @@
 package org.primefaces.renderkit;
 
 import java.io.IOException;
+import javax.faces.FactoryFinder;
+import javax.faces.application.Application;
+import javax.faces.application.ViewHandler;
+import javax.faces.component.UIViewRoot;
 import javax.faces.context.FacesContext;
 import javax.faces.context.ResponseWriter;
+import javax.faces.render.RenderKit;
+import javax.faces.render.RenderKitFactory;
+
 import org.primefaces.util.HTML;
 
 public class RendererUtils {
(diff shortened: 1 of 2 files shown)


melloware:
I think you can eliminate most of this PR.  You don't need to do this `RenderKit` stuff if you simply want to write to a temporary writer we do this using our `FastStringWriter` all over in the code base.

```java

try (FastStringWriter fsw = new FastStringWriter()) {
                fsw.write("{");
                writeOption(fsw, "@context", model.getContext(), false);
                writeOption(fsw, "@type", model.getType(), true);
                encodeItems(fsw, model.getItemListType(), model.getItems(), true);
                fsw.write("}");

                writer.write(fsw.toString());
}
```

primefaces/src/main/java/org/primefaces/renderkit/RendererUtils.java
```diff
@@ -70,4 +77,46 @@ else if (partialSelected) {
...
+                renderKitId = application.getDefaultRenderKitId();
+
+                if (renderKitId == null) {
+                    renderKitId = RenderKitFactory.HTML_BASIC_RENDER_KIT;
+                }
+            }
+        }
+
+        return ((RenderKitFactory) FactoryFinder.getFactory(FactoryFinder.RENDER_KIT_FACTORY)).getRenderKit(context, renderKitId);
+    }
```
melloware, line 120: Ithink you can remove this method

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
```diff
@@ -607,4 +619,43 @@ public static String getDynamicColumnValueAsString(UIComponent component) {
...
+     * already doesn't have a HTML head nor body.
+     * @param component The component to capture HTML output for.
+     * @param context The current FacesContext.
+     * @return The encoded HTML output of the given component.
+     * @throws UncheckedIOException Whenever something fails at I/O level. This would be quite unexpected as it happens locally.
+     * @see <a href="https://github.com/omnifaces/omnifaces">Omnifaces</a>
+     */
+    public static String encodeHtml(UIComponent component, FacesContext context) {
+        ResponseWriter originalWriter = context.getResponseWriter();
+        StringWriter output = new StringWriter();
```
melloware: Just use our FastStringWriter.

melloware (changes requested):

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
jasonex7: FastStringWriter is not a javax.faces.context.ResponseWriter so I cannot pass directly to `context.setResponseWriter` method. I think that is the reason why the method needs the render kit instance, to create a ResponseWriter.

I've just commited a modifed version using FastStringWriter instead of StringWriter

jasonex7:
> I think you can eliminate most of this PR. You don't need to do this `RenderKit` stuff if you simply want to write to a temporary writer we do this using our `FastStringWriter` all over in the code base.
> 
> ```java
> ResponseWriter writer = context.getResponseWriter();
> 
> try (FastStringWriter fsw = new FastStringWriter()) {
>                 fsw.write("{");
>                 writeOption(fsw, "@context", model.getContext(), false);
>                 writeOption(fsw, "@type", model.getType(), true);
>                 encodeItems(fsw, model.getItemListType(), model.getItems(), true);
>                 fsw.write("}");
> 
>                 writer.write(fsw.toString());
> }
> ```

The most uses I see of StringWriter is in render classes to generate strings which at the end are going to be writed to `ResponseWriter` of FacesContext:

```java
public static void encodeModel(FacesContext context, JsonLDModel model) throws IOException {
    if (model != null) {

        try (FastStringWriter fsw = new FastStringWriter()) {
            fsw.write("{");
            writeOption(fsw, "@context", model.getContext(), false);
            writeOption(fsw, "@type", model.getType(), true);
            encodeItems(fsw, model.getItemListType(), model.getItems(), true);
            fsw.write("}");

            writer.write(fsw.toString());
        }
    }
}
```

melloware:
Hmmm but isn't that all you are doing here capturing the string of encodeAll to put in the Excel cell?

melloware:
If you search the code base for "setResponseWriter" you will see we do this in other places in the code

jasonex7:
OK, I used a cloned writer as in TimelineRenderer and UICacheRenderer

primefaces/src/main/java/org/primefaces/renderkit/RendererUtils.java
jasonex7: OK, I removed to favor the use of `cloneWithWriter` method

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
jasonex7: I used `cloneWithWriter` method to avoid the use of getRenderKit method

melloware:
Excellent that is what I meant!

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
```diff
@@ -99,7 +103,12 @@ else if (submittedValue != null) {
 
             ValueHolder valueHolder = (ValueHolder) component;
             if (value == UNDEFINED_VALUE) {
-                value = valueHolder.getValue();
+                if (valueHolder instanceof HtmlOutputFormat) {
+                    value = encodeHtml((HtmlOutputFormat) valueHolder, context);
```
melloware: You don't need to pass `(HtmlOutputFormat)valueHolder` here just pass `component` since its already a UIComponent that the signature of encodeHtml needs.

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
```diff
@@ -607,4 +616,40 @@ public static String getDynamicColumnValueAsString(UIComponent component) {
...
+     * added components, component event listeners, then it should mostly be safe.
+     * In other words, use this at most for "simple templates" only, e.g. a HTML based mail template, which usually
+     * already doesn't have a HTML head nor body.
+     * @param component The component to capture HTML output for.
+     * @param context The current FacesContext.
+     * @return The encoded HTML output of the given component.
+     * @throws UncheckedIOException Whenever something fails at I/O level. This would be quite unexpected as it happens locally.
+     * @see <a href="https://github.com/omnifaces/omnifaces">Omnifaces</a>
+     */
+    public static String encodeHtml(UIComponent component, FacesContext context) {
```
melloware: I would just call this `encodeComponent` instead of `encodeHtml` and check the Javadoc above to make sure its all still true?  Especially all the OmniFaces stuff.

melloware (changes requested):

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
jasonex7: Please check now

melloware (approved):

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
```diff
@@ -607,4 +616,29 @@ public static String getDynamicColumnValueAsString(UIComponent component) {
...
+     * Encodes the given component locally as HTML.
+     * @param component The component to capture HTML output for.
+     * @param context The current FacesContext.
+     * @return The encoded HTML output of the given component.
+     * @throws UncheckedIOException Whenever something fails at I/O level. This would be quite unexpected as it happens locally.
+     */
+    public static String encodeComponent(UIComponent component, FacesContext context) {
+        ResponseWriter originalWriter = context.getResponseWriter();
+        FastStringWriter output = new FastStringWriter();
+        context.setResponseWriter(originalWriter.cloneWithWriter(output));
```
jasonex7: @melloware this line throws NPE because originalWriter return null. I think that is why Omnifaces uses getRenderKit method to address that situations. Maybe is because the render response is not available yet when this method is executed

melloware:
Ahhh I see I bet its not set yet because DataExporter is a TagHandler not a Renderer.

melloware (approved):

PR #9048, opened 2022-07-30
Fix #9038: Other UI commands supporting disableOnAjax

### LINKED ISSUE #9038: UICommand: reenabled after disableOnAjax when disabled on server side in request
### Describe the bug

When a command button action willingly sets the disabled state of the button, the visual state of the button is still enabled. After another page update, this is reflected properly.

### Reproducer

Please clone: https://github.com/stolp/primefaces-test/tree/PF-9038

Form:
```xhtml
<h:form id="frmTest">

 <p>It is expected, that the first button disables itself after its action, but it doesn't.</p>
 <p>The second button only refreshes the form to prove that the disabled setting is indeed present.</p>

 <p:commandButton disabled="#{testView.isDisabled()}"
   value="Button Action disables this button" action="#{testView.sleep()}" partialSubmit="true" update="@form" />

 <p:commandButton
   value="Button Action updates the page" action="#{testView.refresh()}" partialSubmit="true" update="@form" />
</h:form>
```

### Expected behavior

The button state should be disabled after the action finishes.

### PrimeFaces edition

Community

### PrimeFaces version

12.0.0-RC2

### Theme

any

### JSF implementation

All

### JSF version

2.3

### Browser(s)

_No response_

jepsar:
Fix #9038

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/commandbutton/CommandButtonRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/commandbutton/CommandButtonRenderer.java
@@ -174,7 +174,7 @@ public class CommandButtonRenderer extends CoreRenderer {
         WidgetBuilder wb = getWidgetBuilder(context);
         wb.init("CommandButton", button)
             .attr("disableOnAjax", button.isDisableOnAjax(), true)
-            .attr("disabled", button.isDisabled(), false);
+            .attr("disabledAttr", button.isDisabled(), false);
 
         encodeClientBehaviors(context, button);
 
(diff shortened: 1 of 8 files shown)


primefaces/src/main/java/org/primefaces/component/commandbutton/CommandButtonRenderer.java
```diff
@@ -174,7 +174,7 @@ protected void encodeScript(FacesContext context, CommandButton button) throws I
         WidgetBuilder wb = getWidgetBuilder(context);
         wb.init("CommandButton", button)
             .attr("disableOnAjax", button.isDisableOnAjax(), true)
-            .attr("disabled", button.isDisabled(), false);
+            .attr("disabledAttr", button.isDisabled(), false);
```
jepsar, line 177: `this.cfg.disabled` was already used in `SplitButton` and `MenuButton`, so I opted to rename it on each component for consistency.

melloware (approved):

PR #9049, opened 2022-07-30
#9033 Dialog Framework: make dialogs more responsive

### LINKED ISSUE #9033: Dialog Framework: make dialogs more responsive
### Description

Dialog Framework - dialog should be responsive

1. adapt their own size to available screenspace
2. should allow to use PrimeFlex inside the dialog 

https://github.com/primefaces/primefaces/issues/723 only seem´s to be part of the truth.
(e.g. https://forum.primefaces.org/viewtopic.php?t=52907)

- [ ] add best-practice-showcase-example
- [ ] document already exisiting option 'styleClass'
- [ ] add option 'iframeStyleClass' (which may be set to max-w-screen provided by PrimeFlex)
- [ ] add DialogFrameworkOptions-class with builder as convenient way to build options-map

```java
Map<String, Object> options = new HashMap();
options.put("modal", true);
options.put("responsive", true);
// options.put("width", "auto");
options.put("contentWidth", "900");
// wichtig um responsive zu sein
options.put("styleClass", "max-w-screen");
```

### Additional context

_No response_

christophs78:


melloware:
This is a really nice improvement and I love that you made it backwards compatible with Options or the Builder.

christophs78:
image
image
image

primefaces/src/main/java/org/primefaces/model/DialogFrameworkOptions.java
```diff
@@ -0,0 +1,709 @@
...
+ * THE SOFTWARE.
+ */
+package org.primefaces.model;
+
+import org.primefaces.util.LangUtils;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class DialogFrameworkOptions {
```
tandraschko: Make it serializable

otherwise +1 in general
please just make sure it uses the same builder style as other models (didnt crosscheck)

christophs78: Done.

I used DefaultMenuItem as "template" for this DialogFrameworkOptions. Should be the same style.

melloware (approved):

PR #9050, opened 2022-08-01
Fix #8660: Tabview/Accordion check for non input error

### LINKED ISSUE #8659: TabView/AccordionPanel: error indication

melloware:
Fix #8660: Tabview/Accordion check for non input error

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/accordion/accordion.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/accordion/accordion.js
@@ -73,7 +73,7 @@ PrimeFaces.widget.AccordionPanel = PrimeFaces.widget.BaseWidget.extend({
         
         this.headers.each(function() {
             var containerId = PrimeFaces.escapeClientId(this.id.replace('_header', ''));
-            if ($(containerId + ' .ui-state-error').length > 0) {
+            if ($(containerId + ' .ui-state-error').length > 0 || $(containerId + ' .ui-message-error-detail').length > 0) {
                 $(this).addClass('ui-state-error');
             } else {
                 $(this).removeClass('ui-state-error');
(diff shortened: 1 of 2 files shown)


tandraschko:
Couldnt tree just render ui-state-error? A error message is a bit different meaning

melloware:
> Couldnt tree just render ui-state-error? A error message is a bit different meaning

I looked and its quite messy because it happens in `UITree.processValidators` here...

```java
validateSelection(context);
application.publishEvent(context, PostValidateEvent.class, this);
popComponentFromEL(context);
```

So I didn't see access to the component itself to add the `ui-state-error`  class to.  Since this is not an input component it looks like it works a lot differently but adding `ui-state-error` would be preferred.

tandraschko:
i see
lets leave your PR as it is, would be to much changes to those components

PR #9057, opened 2022-08-02
Fix #9054: TabView add focusOnError property

### LINKED ISSUE #9054: TabView/AccordionPanel: focusOnError
### Description

Now that we have #8659, we could take this up to the next logical step, which is to flip towards the (first) tab with an error indication.

This behavior should be activated via a `focusOnError` attribute being `true`.

### Additional context

_No response_

melloware:
Fix #9054: TabView add focusOnError property

@fcorneli you owe me a beer!


fcorneli:
@melloware When passing by in Brussels, let me know. Beers are on me. :)

jepsar:
Free beers are offered in Antwerp as well :P

Issue #9064, opened 2022-08-04
BlockUI: doesn't work when content changes


wsaca:
### Describe the bug

Whether due to changes in the size of the window or the content (ajax) the blocker does not adjust to the target component. This component was working correctly before the refactor in #8548.

### Reproducer

1. Go to https://www.primefaces.org/showcase/ui/misc/blockUI.xhtml
2. Resize the window (width = 600px)
3. Click the button to block the component

### Expected behavior

The target component should be blocked correctly even after content changes.

### PrimeFaces edition

Community

### PrimeFaces version

12.0.0-RC2

### JSF implementation

Mojarra

### JSF version

2.3

### Browser(s)

Firefox

PR #9067, opened 2022-08-04
Fix #9068: BlockUI: fix incorrect position of an overlay

### LINKED ISSUE #9068: BlockUI: Incorrect overlay placement
### Describe the bug

The overlay is displayed incorrectly, the size is correct, but position is shifted.

### Reproducer

_No response_

### Expected behavior

It seems it's a mistake to use `.position()` which returns the position of en element relative to its parent. But instead it should be relative to a page `.offset()`, because the overlay element is no more created next to a block element, but instead attached to `@(body)`.

### PrimeFaces edition

Community

### PrimeFaces version

12-RC2

### Theme

_No response_

### JSF implementation

MyFaces

### JSF version

3.0.0

### Browser(s)

Chrome 104

astappiev:
Affected version: 12-RC2

It seems it's a mistake to use `.position()` which returns the position of en element relative to its parent.

But instead it should be relative to a page, because the overlay element is no more created next to `block` element, but instead attached to `@(body)`.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
@@ -227,12 +227,12 @@ PrimeFaces.widget.BlockUI = PrimeFaces.widget.BaseWidget.extend({
             // set the size and position to match the target
             var height = currentTarget.height(),
                 width = currentTarget.width(),
-                position = currentTarget.position();
+                offset = currentTarget.offset();
             var sizeAndPosition = {
                 'height': height + 'px',
                 'width': width + 'px',
-                'left': position.left + 'px',
-                'top': position.top + 'px'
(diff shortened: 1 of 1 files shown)


melloware:
Is this related to https://github.com/primefaces/primefaces/issues/9064 or no?

astappiev:
Also, it would be nice to recalculate size of the overlay, because now it's “rendered” on page load and then only `show/hide` despite page or block element size change.

astappiev:
> Is this related to #9064 or no?

No, it's irrelevant. The error reproduces on any page where the `block` element is not directly in the `body`.

> Also, it would be nice to recalculate size of the overlay, because now it's “rendered” on page load and then only `show/hide` despite page or block element size change.

But this comment does :)

melloware:
OK let me test this!

jepsar:
@astappiev can you please create an issue to link this PR to?

PR #9112, opened 2022-08-17
Update of Messages_pl.properties

Emkas:
* Add missing properties
* Remove non-ASCII characters


melloware:
@Emkas did you also review the client side properties as well for PL?

melloware:
See: https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/resources/META-INF/resources/primefaces/locales/locale-pl.js

Emkas:
> @Emkas did you also review the client side properties as well for PL?

No...

I will look for it, and I can update it as well. No problem.

Emkas:
Thanks. I need a second for this. I'll provide next commit if needed.

melloware:
Perfect much appreciated.

Emkas:
OK, there's not so much to add there. Only two new properties.

PR #9115, opened 2022-08-18
Fix #9092: Widget lifecycle for PostConstruct/Refresh

### LINKED ISSUE #9092: Widget postConstruct lifecycle callback is invoked with non-fully constructed widget instance
### Describe the bug

Client-side widget lifecycle callbacks were added I believe in PF 9, which are super useful for various purposes. I was just using the `postConstruct` callback, when I noticed a minor issue with it: Currently, `postConstruct` [is called during the init method of the BaseWidget](https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.widget.js#L335-L337). But the init method is called by the individual widget during the `super` call (e.g. [see here for spitButton](https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.splitbutton.js#L55)). So at this point, the widget was not fully initialized yet, and thus lacks e.g. [various properties that are set only during the `init` method](https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.splitbutton.js#L57).

This isn't a major issue and can be worked around e.g. by using `setTimeout` or promises, but perhaps the callback could be invoked a bit later after the widget was initialized?

For example, the following code to add a `aria-label` attribute to the more button of the split button (for accessibility) does not work like this (this should probably be fixed on its own right, but that's just one example): 

```xml
<p:splitButton>
  <f:attribute name="widgetPostConstruct" value="this.menuButton.attr("aria-label", "Show all items")" />
</p:splitButton>
```

To make it work, we need to delay:

```xml
<p:splitButton>
  <f:attribute name="widgetPostConstruct" value="Promise.resolve().then(() => this.menuButton.attr("aria-label", "Show all items"))" />
</p:splitButton>
```

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

_No response_

### PrimeFaces version

11

### Theme

_No response_

### JSF implementation

_No response_

### JSF version

_No response_

### Browser(s)

_No response_

melloware:
Fix #9092: Widget lifecycle for PostConstruct/Refresh

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.js
@@ -635,12 +635,19 @@
                 //ajax update
                 if(widget && (widget.constructor === this.widget[widgetName])) {
                     widget.refresh(cfg);
+                    if (cfg.postRefresh) {
+                        cfg.postRefresh.call(widget, widget);
+                    }
                 }
                 //page init
                 else {
-                    this.widgets[widgetVar] = new this.widget[widgetName](cfg);
+		    var newWidget = new this.widget[widgetName](cfg);
+                    this.widgets[widgetVar] = newWidget;
                     if(this.settings.legacyWidgetNamespace) {
-                        window[widgetVar] = this.widgets[widgetVar];
+                        window[widgetVar] = newWidget;
+                    }
+                    if (cfg.postConstruct) {
+                       cfg.postConstruct.call(newWidget, newWidget);
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


blutorange:
@melloware Looks perfect to me! That should ensure the widget is fully initialized (also sorry I didn't get around to do this, was pretty busy in the last months, but it's getting better now)

tandraschko:
yep, perfect!

PR #9117, opened 2022-08-19
Fix #9093: InputNumber: decimalPlaces not defaulting correctly on numeric types with null value

### LINKED ISSUE #9093: InputNumber: decimalPlaces not defaulting correctly on numeric types with null value
### Describe the bug

If the value is `null` the type of the bound value should be checked

https://github.com/primefaces/primefaces/blob/fcced6df36579bc34f09b03b683f31fdee8f0ce3/primefaces/src/main/java/org/primefaces/component/inputnumber/InputNumberRenderer.java#L393-L395

### Reproducer

Bind to an `Integer` type of which the value is `null`

### Expected behavior

`decimalPlaces` should be 0 instead of 2, because the value is of type `Integer` although the value is `null`.

### PrimeFaces edition

Community

### PrimeFaces version

12

### Theme

_No response_

### JSF implementation

_No response_

### JSF version

_No response_

### Browser(s)

_No response_

jepsar:
Fix #9093

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/inputnumber/InputNumberRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/inputnumber/InputNumberRenderer.java
@@ -186,7 +186,8 @@ public class InputNumberRenderer extends InputRenderer {
         String styleClass = createStyleClass(inputNumber, InputNumber.PropertyKeys.inputStyleClass.name(), InputText.STYLE_CLASS) ;
         String inputMode = inputNumber.getInputmode();
         if (inputMode == null) {
-            String decimalPlaces = getDecimalPlaces(context, inputNumber, inputNumber.getValue());
+            boolean isIntegral = isIntegral(context, inputNumber, inputNumber.getValue());
+            String decimalPlaces = getDecimalPlaces(isIntegral, inputNumber);
             inputMode = "0".equals(decimalPlaces) ? "numeric" : "decimal";
             inputNumber.setInputmode(inputMode);
         }
@@ -224,6 +225,7 @@ public class InputNumberRenderer extends InputRenderer {
                     ? "."
                     : inputNumber.getDecimalSeparator();
 
+        boolean isIntegral = isIntegral(context, inputNumber, value);
 
         WidgetBuilder wb = getWidgetBuilder(context);
         wb.init(InputNumber.class.getSimpleName(), inputNumber);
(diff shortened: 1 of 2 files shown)


jepsar:
I think the failing test is wrong:
```java
        // NOTE: because its now NULL the default decimal places is set back to 2 because its no longer an Integer
        assertConfiguration(inputNumber.getWidgetConfiguration(), "2", "0.0000001", "999999");
```
That's exactly what has changed in this PR. Shall I change the test value from 2 to 0?

jepsar:
Done.

primefaces/src/main/java/org/primefaces/component/inputnumber/InputNumberRenderer.java
```diff
@@ -391,7 +394,24 @@ private String getMaximum(InputNumber inputNumber, Object value) {
...
+                || value instanceof Byte;
+    }
+
+    private boolean isIntegral(FacesContext context, InputNumber inputNumber) {
+        ValueExpression valueExpression = inputNumber.getValueExpression("value");
+        if (valueExpression == null) {
+            return false;
+        }
+        Class<?> type = valueExpression.getType(context.getELContext());
+        return type.isAssignableFrom(Long.class)
```
Rapster, line 409: Maybe you can simplify by moving these conditions in isIntegral(Object value) and then you can directly call in isIntegral(FacesContext context, InputNumber inputNumber)? To avoid repetition

melloware: Agreed!   Other than that looks great

melloware (changes requested):

melloware:
> I think the failing test is wrong:
> 
> ```java
>         // NOTE: because its now NULL the default decimal places is set back to 2 because its no longer an Integer
>         assertConfiguration(inputNumber.getWidgetConfiguration(), "2", "0.0000001", "999999");
> ```
> 
> That's exactly what has changed in this PR. Shall I change the test value from 2 to 0?

Yes you have fixed the code propertly so the unit test should be changed accordingly.

primefaces/src/main/java/org/primefaces/component/inputnumber/InputNumberRenderer.java
jepsar: Right, it's better now. I also optimized it a bit (it was called three times in a row)

melloware (approved):

PR #9119, opened 2022-08-20
Fix #9118: Signature padding removal

### LINKED ISSUE #9118: Signature: Drawing is off because of padding
### Describe the bug

See: https://stackoverflow.com/questions/73418237/primefaces-signature-line-drawn-from-pointer-is-not-at-the-tip

@FlipWarthog diagnosed...

> I think it's a bug. Specifically, the wrapper div for the signature component has ui-inputfield as one of its style classes which is applying a padding. Looks like that padding it throwing off the accuracy of determining pointer position.
> 
> If you check the showcase and use browser dev tools to set padding: 0; on the div, the lines appear exactly where expected.

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

Community

### PrimeFaces version

12.0.0-RC2

### Theme

_No response_

### JSF implementation

All

### JSF version

2.3

### Java version

11

### Browser(s)

_No response_

melloware:
Fix #9118: Signature padding removal

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/signature/signature.css
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/signature/signature.css
@@ -7,3 +7,7 @@
 .kbw-signature-disabled {
 	opacity: 0.35;
 }
+
+body .ui-inputfield.kbw-signature {
+    padding: 0;
+}


jepsar:
Should we assign Mert to review this?

PR #9157, opened 2022-08-30
Fix #9138: Document custom resource handler for cache busting

### LINKED ISSUE #9138: Configuration: primefaces.HIDE_RESOURCE_VERSION and updating PrimeFaces
### Description

#8752 -this is a nice feature, because it is bad practice to have that much visible library version for everyone, but it is problematic when you update Primeface and people have cached version of old scripts. 

Maybe instead of boolean, there can be configuration that could accept custom String, for example:

```
<context-param>
    <param-name>primefaces.RESOURCE_VERSION</param-name>
    <param-value></param-value>
</context-param>

<context-param>
    <param-name>primefaces.RESOURCE_VERSION</param-name>
    <param-value>49827659</param-value>
</context-param>
```

### Additional context

_No response_

melloware:
Fix #9138: Document custom resource handler for cache busting

### DIFF
--- a/docs/12_0_0/core/misc.md
+++ b/docs/12_0_0/core/misc.md
@@ -45,3 +45,128 @@ However it does not set no-cache headers for JSF views, which can be enabled wit
     <servlet-name>Faces Servlet</servlet-name>
 </filter-mapping>
 ```
+
+### Cache Busting Resource Handler
+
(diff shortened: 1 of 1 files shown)


tandraschko:
+1

PR #9158, opened 2022-08-30
Fix #9146: SelectOneMenu disable correct items

### LINKED ISSUE #9146: SelectOneMenu: SelectItemGroup combination defect
### Describe the bug

The SelectOneMenu does not properly render the checked/unchecked items. The items are only rendered correctly if I use either just the SelectItems (without grouping) or a list of objects.

Everything worked perfectly under PF 8, so I now have to adjust everything so that the clerks can work properly with it again.

[pfe-select_item_issue.zip](https://github.com/primefaces/primefaces/files/9443113/pfe-select_item_issue.zip)

### PrimeFaces edition

Community

### PrimeFaces version

11

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.2,x 2.3.x

### Java version

8

### Browser(s)

FF and Google Chrome

**Defect**:
grafik

**Correct**:
grafik
grafik

melloware:
Fix #9146: SelectOneMenu disable correct items

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectonemenu.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectonemenu.js
@@ -162,9 +162,7 @@ PrimeFaces.widget.SelectOneMenu = PrimeFaces.widget.DeferredWidget.extend({
         highlightedItem = this.items.eq(this.options.index(selectedOption));
 
         //disable options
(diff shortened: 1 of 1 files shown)


melloware:
Basically there is a subtle difference between `:disabled` and `[disabled]` attribute selecting and it was selecting options to disable the LI instead of just disabling the LI items.

tandraschko:
nice, thanks!

PR #9191, opened 2022-09-09
Fix #9171: UIData#visitTree should respect VisitResult#REJECT

### LINKED ISSUE #9171: DataTable: Not compatible with MyFaces' component tree manipulation
### Describe the bug

Our custom library utilizes the PF datatable in a very dynamic manner. Columns are populated dynamically based on runtime configurations. Everything worked fine with Mojarra, but because of other bugs, I tried MyFaces as the JSF implementation.

MyFaces performs way better and everything works as expected, but therefore I got a new bug: The contents of the dynamic columns is duplicated on view restore (e.g. pagination).

After digging around, the root cause is probably not MyFaces but PF itself.

During `DefaultFaceletsStateManagementStrategy#saveView` (MyFaces class) all dynamically created components are stored to the view state for restore at a later time. In `#saveStateOnMapVisitTree` a tree visit is started which tries to collect those dynamic components and if one is found, the state is serialized (including the children) and `VisitResult.REJECT` is returned, so the tree-visit does not visit the children again (see [here](https://github.com/apache/myfaces/blob/2.2.x/impl/src/main/java/org/apache/myfaces/view/facelets/DefaultFaceletsStateManagementStrategy.java#L1066) for the MyFaces source code).
 
 

But PF does not respect that `VisitResult.REJECT` in UIData. The problematic lines of code are https://github.com/primefaces/primefaces/blob/2b1384a3c6464c995e029757eeebd8cb6144f4d9/primefaces/src/main/java/org/primefaces/component/api/UIData.java#L771 and https://github.com/primefaces/primefaces/blob/2b1384a3c6464c995e029757eeebd8cb6144f4d9/primefaces/src/main/java/org/primefaces/component/api/UIData.java#L897.

The first code stops the processing of the children, as only the UIColumn itself should be visited at that place. But the result 'REJECT' is never stored. During the second code, the column content is visited which clearly is not expected, as the save-state-logic of MyFaces previously rejected the processing of the children of the column.

### Reproducer
https://github.com/fanste/primefaces-test/tree/test-9171
Just use the paginator and have a look at the second column.

### Expected behavior

_No response_

### PrimeFaces edition

Community

### PrimeFaces version

11.0.0

### Theme

_No response_

### JSF implementation

MyFaces

### JSF version

2.2.15

### Java version

8

### Browser(s)

_No response_

fanste:
Not the nicest solution, but the simpliest I came around.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/api/UIData.java
+++ b/primefaces/src/main/java/org/primefaces/component/api/UIData.java
@@ -723,11 +723,12 @@ public class UIData extends javax.faces.component.UIData {
                     return true;
                 }
 
-                if (requiresColumns() && visitColumnsAndColumnFacets(context, callback, visitRows)) {
+                Set<UIComponent> rejectedChildren = new HashSet<>();
+                if (requiresColumns() && visitColumnsAndColumnFacets(context, callback, visitRows, rejectedChildren)) {
                     return true;
                 }
 
-                if (visitRows(context, callback, visitRows)) {
+                if (visitRows(context, callback, visitRows, rejectedChildren)) {
                     return true;
                 }
 
@@ -760,7 +761,7 @@ public class UIData extends javax.faces.component.UIData {
         return false;
     }
 
-    protected boolean visitColumnsAndColumnFacets(VisitContext context, VisitCallback callback, boolean visitRows) {
+    protected boolean visitColumnsAndColumnFacets(VisitContext context, VisitCallback callback, boolean visitRows, Set<UIComponent> rejectedChildren) {
         if (visitRows) {
             setRowIndex(-1);
         }
@@ -771,6 +772,10 @@ public class UIData extends javax.faces.component.UIData {
                 VisitResult result = context.invokeVisitCallback(child, callback); // visit the column directly
                 if (result == VisitResult.COMPLETE) {
                     return true;
+                }
+                else if (result == VisitResult.REJECT) {
+                    rejectedChildren.add(child);
+                    continue;
                 }
 
                 if (child instanceof UIColumn) {
@@ -853,7 +858,7 @@ public class UIData extends javax.faces.component.UIData {
         return false;
     }
 
-    protected boolean visitRows(VisitContext context, VisitCallback callback, boolean visitRows) {
+    protected boolean visitRows(VisitContext context, VisitCallback callback, boolean visitRows, Set<UIComponent> rejectedChildren) {
         boolean requiresColumns = requiresColumns();
         int processed = 0;
         int rowIndex = 0;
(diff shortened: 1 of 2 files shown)


tandraschko:
Can you please add a Integration Test?

tandraschko:
also we should collect clientIds i think, not UIComponents

fanste:
I don't see  a requirement for collecting clientIds instead of UIComponents.

This would only be required, if there is any code that may replace a components instance and so the basic `equals` implemention fails. Then the clientId would be the better idea.
On the other side, `REJECT` was returned for exacly that instance on that child-index. 
- If another component is inserted before an already processed child, the clientId will change - wrong child is rejected.
- If an instance on that child-index is exchanged it is effectively another children which was not rejected.

I don't know the specs on how this is defined. But in this specific case: There is not much place for some _replace/manipulate that UIComponent_ between `visitColumnsAndColumnFacets` and `visitRows`.

But if you insist on it, I will change it to clientIds.

fanste:
I will try to implement a test. Last time this was a pain in the ... as our infrastructure here is really picky.

tandraschko:
im really not sure if the UIComponent cant be maniupliated/replaced as e.g. p:columns or we even support ui:repeat in some cases

therefore i just felt more safe with clientId

tandraschko:
do you have time to work on it? would like to see it in 12.0 final

fanste:
I was on vacation and will try to implement a test tomorrow.

UIComponents instead of clientIds can stay for now or does it need to be changed?

tandraschko:
i would feel more safe with clientId TBH

fanste:
I just changed it to the clientId, but then the behaviour might be wrong. That is now up to you, to decide it.

Using the UIComponent, the same instance is referenced in `visitColumnsAndColumnFacets` and `visitRows`.
But with the clientId, the returned ID in `visitColumnsAndColumnFacets` is without a row-index while in `visitRows` the returned ID is with a row-index.

During the processing of the column, the logic may return `REJECT` but that is mapped to the ÍD without a row-index. Later the row-index is included and to the column is not skipped (rejected) and the children of that column will be processed.

This observation is only true if `VisitHint#SKIP_ITERATION` is not used. If the hint is used, no row-index will be included. In the specific use-case of this issue, the hint is always used.

melloware:
@fanste I just added an Integration Test based on your PF reproducer for this issue. See: https://github.com/primefaces/primefaces/pull/9361

melloware:
As expected: 
image

melloware:
I merged your changes into my PR and now the Integration tests are passing!  @tandraschko I am going to merge and if you decide you want to refactor later for ID's you can but I see what @fanste is saying about using clientIds

PR #9200, opened 2022-09-12
Fix #9179: Exporter support dynamic column groups

### LINKED ISSUE #9179: DataExporter: support dynamic grouped columns with ui:repeat
### Describe the bug

The showcase page :
http://primefaces.org/showcase/ui/data/datatable/group.xhtml?jfwid=f547e 
shows the issue. 
After exporting data to  XLSX - button on the top right - the data export for second table (Dynamic Columns tab in export) does not show any header for dynamic columns ( 2010 2011 2012 2013 2014 - are missing).

Environment: v.12 ( tested also in v.11 - same issue)

image

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

_No response_

### PrimeFaces version

11.0.0 and up

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

11

### Browser(s)

all

melloware:
Fix #9179: Exporter support dynamic column groups

@tandraschko you were right had to use `forEachColumn` and now it works!

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/api/ColumnAware.java
+++ b/primefaces/src/main/java/org/primefaces/component/api/ColumnAware.java
@@ -31,14 +31,17 @@ import java.util.concurrent.atomic.LongAdder;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
+
 import javax.faces.FacesException;
 import javax.faces.component.UIComponent;
 import javax.faces.component.visit.VisitContext;
 import javax.faces.component.visit.VisitResult;
 import javax.faces.context.FacesContext;
+
 import org.primefaces.component.column.Column;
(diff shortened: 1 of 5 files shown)


tandraschko:
we should also change this for TreeTable exporter etc.?
Also we could add a new fuction #forEachColumnGroup to make the code easier and reusable

melloware:
Added new `forEachColumnGroup` callback to return each `row` of a column group.  This cleans up the code nicely and added TreeTable support.

tandraschko:
+1 if well tested
cool change!

tandraschko:
Just wonder if it shouldnt be forEachRow? as you pass the ColumnGroup

melloware:
I renamed it `forEachColumnGroupRow` I didn't want to throw someone off thinking `forEachRow` meant each row in the DT.

Issue #7900, opened 2021-09-30
Avatar: Problem with non-US chars


aripddev:
**Describe the defect**
There are two defects
1. Thrown `StringIndexOutOfBoundsException` when using non-US chars in surname initial.
2. No exception thrown but the initial not shown when non-US chars in first name initial.

Tried with Ş,Ö,Ç,İ,Ü

**Reproducer**
[issuereport.zip](https://github.com/primefaces/primefaces/files/7262144/issuereport.zip)

**Expected behavior**
Expected to see the original chars.

**Example XHTML**
```xhtml
<p:panelGrid columns="2">
    <h:outputText value="PrimeFaces Rocks"/>
    <p:avatar label="PrimeFaces Rocks"/><!-- works -->

    <h:outputText value="ŞPrimeFaces Rocks"/>
    <p:avatar label="ŞPrimeFaces Rocks"/><!-- does not work as expected -->

    <h:outputText value="PrimeFaces ŞRocks"/>
    <p:avatar label="PrimeFaces ŞRocks"/><!-- does not work as expected -->

    <h:outputText value="ŞPrimeFaces ŞRocks"/>
    <p:avatar label="ŞPrimeFaces ŞRocks"/><!-- thrown error -->
</p:panelGrid>

<p:panelGrid columns="2">
    <h:outputText value="#{testView.name1}"/>
    <p:avatar label="#{testView.name1}"/><!-- works -->
    
    <h:outputText value="#{testView.name2}"/>
    <p:avatar label="#{testView.name2}"/><!-- does not work as expected -->
    
    <h:outputText value="#{testView.name3}"/>
    <p:avatar label="#{testView.name3}"/><!-- does not work as expected -->
    
    <h:outputText value="#{testView.name4}"/>
    <p:avatar label="#{testView.name4}"/><!-- thrown error -->
</p:panelGrid>
```

**Example Bean**
```java
@Named
@ViewScoped
public class TestView implements Serializable {
    private String name1 = "PrimeFaces Rocks";
    private String name2 = "ŞPrimeFaces Rocks";
    private String name3 = "PrimeFaces ŞRocks";
    private String name4 = "ŞPrimeFaces ŞRocks";
    
    // getters
}
```

melloware:
Ahhh good catch. Should be easy to fix.

PR #9225, opened 2022-09-21
Java 19 instead of 18

christophs78:


christophs78:
AvatarRendererTest fails on Java 19. (Don´t understand the reason yet.)

christophs78:
Java 19 seems to interpret RegEx a bit different.

`private static final Pattern LETTER_PATTTERN = Pattern.compile("\\b[\\p{L}\\p{M}]");`
(part of AvatarRenderer)

Maybe kind of related to https://github.com/primefaces/primefaces/issues/7900

PR #9263, opened 2022-10-04
Fix #9259: DatePicker with mask prevent duplicate onChange

### LINKED ISSUE #9259: Datepicker: change event is triggered with `mask=true` when leaving input without changes
### Describe the bug

The first time I pick any date the ajax event is called twice, this can be checked through F12
the first ajax sends an empty date, and the second the value picked.
```
<p:datePicker id="value"
		      value="#{testView.date}"
		      mask="true"
		      pattern="dd/MM/yyyy"
		      showIcon="true"
		      showOnFocus="false"> 
	<p:ajax event="change" update="date"/>		 
</p:datePicker>	
```

### Reproducer

[primefaces-test-master.zip](https://github.com/primefaces/primefaces/files/9700633/primefaces-test-master.zip)

### Expected behavior

The event should be called just once with the correct value.

### PrimeFaces edition

Community

### PrimeFaces version

12.0.0.RC3

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2

### Java version

2.2.20

### Browser(s)

_No response_

melloware:
Fix #9259: DatePicker with mask prevent duplicate onChange

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/1-datepicker.js
@@ -258,6 +258,7 @@ PrimeFaces.widget.DatePicker = PrimeFaces.widget.BaseWidget.extend({
                 maskCfg.mask = this.cfg.mask;
             }
             this.input.inputmask('remove').inputmask(maskCfg);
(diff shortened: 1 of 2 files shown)


melloware:
The InputMask on its "blur" event was firing `input.trigger('change');` causing the duplicate events

Issue #9266, opened 2022-10-04
DatePicker: Min/Max not handling timeZone correctly.


melloware:
### Describe the bug

Start a server in UTC like this `mvn  jetty:run -Ddev -Duser.timezone="UTC"`.  This makes the whole server in UTC.

Then add a DatePicker in EST like this...

```xml
<p:datePicker id="timeMinMax" 
value="#{calendarView.date16}" 
showTime="true" 
timeZone="America/New_York" 
mindate="#{calendarView.minDateTime}" 
maxdate="#{calendarView.maxDateTime}"/>
```

This will generate an incorrect error 
![image](https://user-images.githubusercontent.com/4399574/193883727-45248da1-b10a-4eb0-8753-a922cce01e88.png)

This is because deep in the bowels of CalendarUtils it was using `ZoneId.getDefault()` instead of `calculateZoneId(calendar.getTimeZone())`

### PrimeFaces edition

Community

### PrimeFaces version

12.0.0-RC3

### JSF implementation

MyFaces

### JSF version

2.3

### Java version

11

PR #9270, opened 2022-10-04
#9266 - FileUpload; files.length > 0; Ajax-guard

christophs78:


christophs78:
https://www.githubstatus.com/
image

PR #9301, opened 2022-10-11
Fix filter.xhtml customerTable2 filter

stefan-dan:
Fixed filter by "Representative" from "Default Filter" dataTable.


melloware (approved):

melloware:
Thanks for the fix!

PR #9303, opened 2022-10-11
Fix #9302: Breadcrumb allow Home to be lastItemDisabled

### LINKED ISSUE #9302: BreadCrumb: Home icon not shown when using lastItemDisabled="true"
### Describe the bug

When using the breadcrumb configured with only lastItemDisabled="true" and one menuitem corresponding to the home link, the home icon is not shown.

### Reproducer

`<p:breadCrumb lastItemDisabled="true">
    <p:menuitem value="Home" />
 </p:breadCrumb>`

### Expected behavior

The home icon should be shown disabled, because it's the last item.

### PrimeFaces edition

Elite

### PrimeFaces version

12.0.0

### Theme

Saga

### JSF implementation

All

### JSF version

2.3

### Java version

11

### Browser(s)

All

melloware:
Fix #9302: Breadcrumb allow Home to be lastItemDisabled

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/breadcrumb/BreadCrumbRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/breadcrumb/BreadCrumbRenderer.java
@@ -67,9 +67,13 @@ public class BreadCrumbRenderer extends BaseMenuRenderer {
         List<JsonLDItem> ldItems = new ArrayList<>();
 
         //home icon for first item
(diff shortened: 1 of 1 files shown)


jepsar (approved):

jepsar:
Although I don't see why you want to disable home, even if the last item should be disabled and there are no other items...

melloware:
Agreed but to each his own I guess :)

PR #9316, opened 2022-10-16
Fix #7009: Refactor OverlayPanel dialog support

### LINKED ISSUE #7009: Overlay/ColorPicker: Refactor setupDialogSupport

melloware:
Fix #7009: Refactor OverlayPanel dialog support

Reviving this PR now that 12.0.0 is out

### DIFF
--- a/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/overlaypanel/OverlayPanel002Test.java
+++ b/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/overlaypanel/OverlayPanel002Test.java
@@ -94,7 +94,7 @@ public class OverlayPanel002Test extends AbstractPrimePageTest {
 
         // Assert
         Assertions.assertFalse(overlayPanel.isDisplayed());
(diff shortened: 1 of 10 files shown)


melloware:
OK good integration tests caught some issues I need to debug.

melloware:
@tandraschko this is ready for review all integration tests fixed and passing and OverlayPanel is actually correct now as before it was just doing 
```js
this.jq.appendTo(document.body);
```

Instead of setting `cfg.appendTo="@(body)"` before calling its registerDynamicOverlay.

melloware:
Going to merge this for now we can always refactor if we find issues we have a long time before 13.0.0 comes out!

PR #9318, opened 2022-10-17
Add correct pattern for responsive classes

### LINKED ISSUE #9317: PrimeFaces CLI: grid migration issue
I was migrating an old project from grid css to primeflex and I noticed responsive classe are not converted.
The grid.css pattern for responsive classes, as you can see on [the showcase](https://www.primefaces.org/showcase-v8/ui/panel/grid.xhtml) or [the documentation](https://primefaces.github.io/primefaces/8_0/#/components/gridcss), is ui-(xl|lg|md|sm)-_n_,
while the migration tool looks only for ui-g-(xl|lg|md|sm)-_n_:
https://github.com/primefaces/primefaces/blob/9c257a48b28464436e49c4bc782b2233ba23925e/primefaces-cli/src/main/java/org/primefaces/cli/migration/primeflex/GridCssMigration.java#L47

PR coming

NicolaIsotta:
fixes #9317

### DIFF
--- a/primefaces-cli/src/main/java/org/primefaces/cli/migration/primeflex/GridCssMigration.java
+++ b/primefaces-cli/src/main/java/org/primefaces/cli/migration/primeflex/GridCssMigration.java
@@ -41,10 +41,11 @@ public class GridCssMigration extends AbstractPrimeMigration implements Runnable
 
     @Override
     protected void initReplaceRegEx() {
-        // see https://www.primefaces.org/primeflex/migration
 
         // grid
-        replaceRegex.put("ui-g-(xl|lg|md|sm)-([0-9]+?)", "p-(xl|lg|md|sm)-([0-9]+?)");
+        replaceRegex.put("ui-(xl|lg|md|sm)-offset-([0-9]+?)", "p-$1-offset-$2");
+        replaceRegex.put("ui-g-offset-([0-9]+?)", "p-offset-$1");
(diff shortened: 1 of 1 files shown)


primefaces-cli/src/main/java/org/primefaces/cli/migration/primeflex/GridCssMigration.java
```diff
@@ -45,6 +45,7 @@ protected void initReplaceRegEx() {
 
         // grid
         replaceRegex.put("ui-g-(xl|lg|md|sm)-([0-9]+?)", "p-(xl|lg|md|sm)-([0-9]+?)");
```
jepsar: @melloware shouldn't groups be referenced in the replacement here? I would expect `"p-$1-$2"`.

melloware: Yes this looks wrong it should be $ like jepsar says

melloware (changes requested):

primefaces-cli/src/main/java/org/primefaces/cli/migration/primeflex/GridCssMigration.java
jepsar: In that case we don't have a unit test for that replacement, or it's wrong as well

melloware (approved):

NicolaIsotta:
fixed and tested locally with the showcase page, everything works fine.

PR #9325, opened 2022-10-18
Fix #9324: DataTable: Implement IN MatchMode for JpaLazyDataModel

### LINKED ISSUE #9324: DataTable: Implement IN MatchMode for JpaLazyDataModel
### Description

`JpaLazyDataModel` does not currently support `filterMatchMode="in"`

`filterMatchMode="in"` to be supported when using `JpaLazyDataModel`.

### Additional context

Comment on original ticket: https://github.com/primefaces/primefaces/issues/7667#issuecomment-894219333

FlipWarthog:
I had to push the cast to `Comparable` further into `createPredicate` since IN needs either an array or `Collection`, neither of which are `Comparable`.

Testing this change with the [new primefaces-test jpa branch](https://github.com/primefaces/primefaces-test/tree/jpa) by replacing `stringCol` with:

```xml
<p:column field="stringCol" filterMatchMode="in">
  <f:facet name="filter">
    <p:selectCheckboxMenu onchange="PF('jpaTable').filter()">
      <f:selectItem itemLabel="New York" itemValue="New York" />
      <f:selectItem itemLabel="Philadelphia" itemValue="Philadelphia" />
      <f:selectItem itemLabel="Detroit" itemValue="Detroit" />
    </p:selectCheckboxMenu>
  </f:facet>
</p:column>
```

### DIFF
--- a/primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
+++ b/primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
@@ -163,7 +163,8 @@ public class JpaLazyDataModel<T> extends LazyDataModel<T> implements Serializabl
                 Field filterField = LangUtils.getFieldRecursive(entityClass, filter.getField());
                 Object filterValue = filter.getFilterValue();
                 Object convertedFilterValue;
-                if (filterValue.getClass().isArray() || filterValue instanceof Collection) {
+                Class<?> filterClass = filterValue.getClass();
+                if (filterClass.isArray() || Collection.class.isAssignableFrom(filterClass)) {
                     convertedFilterValue = filterValue;
                 }
                 else {
--- a/primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
+++ b/primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
@@ -227,7 +227,9 @@ public class JpaLazyDataModel<T> extends LazyDataModel<T> implements Serializabl
                         ? cb.equal(fieldExpression, filterValueAsCollection.get().iterator().next())
                         : fieldExpression.in(filterValueAsCollection.get());
             case NOT_IN:
-                throw new UnsupportedOperationException("MatchMode.NOT_IN currently not supported!");
+                return filterValueAsCollection.get().size() == 1
+                        ? cb.notEqual(fieldExpression, filterValueAsCollection.get().iterator().next())
+                        : fieldExpression.in(filterValueAsCollection.get()).not();
             case BETWEEN:
                 throw new UnsupportedOperationException("MatchMode.BETWEEN currently not supported!");
             case NOT_BETWEEN:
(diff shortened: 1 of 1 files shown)


primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
```diff
@@ -204,15 +215,17 @@ protected <F extends Comparable> Predicate createPredicate(FilterMeta filter, Fi
...
             case GREATER_THAN_EQUALS:
-                return cb.greaterThanOrEqualTo(fieldExpression, filterValue);
+                return cb.greaterThanOrEqualTo(fieldExpression, (Comparable) filterValue);
             case IN:
-                throw new UnsupportedOperationException("MatchMode.IN currently not supported!");
+                return filterValueAsCollection.get().size() == 1
+                        ? cb.equal(fieldExpression, filterValueAsCollection.get().iterator().next())
+                        : fieldExpression.in(filterValueAsCollection.get());
             case NOT_IN:
                 throw new UnsupportedOperationException("MatchMode.NOT_IN currently not supported!");
```
melloware: You should be able to add Not In too right?

melloware (changes requested):

primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
FlipWarthog: I don't think there's a straight-forward `notIn` method, so I need to refresh myself on how to negate predicates in Criteria API. I wanted to get this in first before I go digging to see how we do it. But, yeah, it should be simple to tack in.

FlipWarthog: ... and [it was](https://github.com/primefaces/primefaces/pull/9325/commits/282310e1aece020c7898e5dc6a183f2cb8c1e344)

primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
```diff
@@ -158,12 +160,18 @@ protected void applyFilters(CriteriaBuilder cb,
                     continue;
                 }
 
-                String filterValue = filter.getFilterValue().toString();
                 Field filterField = LangUtils.getFieldRecursive(entityClass, filter.getField());
-                Object convertedFilterValue = convertToType(filterValue, filterField.getType());
+                Object filterValue = filter.getFilterValue();
+                Object convertedFilterValue;
+                if (filterValue.getClass().isArray() || filterValue instanceof Collection) {
```
melloware: In other spots in the code we use the safer `Collection.class.isAssignableFrom(filterValue.getClass())` so maybe...

```java
Class filterClass = filterValue.getClass();
if (filterClass.isArray() || ollection.class.isAssignableFrom(filterClass )) {
```

melloware (changes requested):

primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
FlipWarthog: Corrected.

melloware (approved):

PR #9327, opened 2022-10-18
Fix #9328: InputNumber minVal > 0

### LINKED ISSUE #8125: InputNumber: "minValue>0" shows js warning
**Describe the defect**
Using something like this:  

```
<p:inputNumber 
    value="#{bean.value}" 
    decimalPlaces="0" 
    minValue="1" 
    maxValue="20"/>
```

shows the following warning in the javascript console:

> Warning: You've set a `minimumValue` or a `maximumValue` excluding the value `0`. AutoNumeric will force the users to always have a valid value in the input, hence preventing them to clear the field. If you want to allow for temporary invalid values (ie. out-of-range), you should use the 'invalid' option for the 'overrideMinMaxLimits' setting.

Looking in https://github.com/autoNumeric/autoNumeric/issues/543#issuecomment-619542552, there is indeed the option. In my opinion, this should be the default (so that the warning disappears), if PF detects a "minValue > 0"

**Environment:**
 - PF Version: _11.00 RC2_
 - JSF + version: _e.g. Mojarra 2.2.20_
 - Affected browsers: ALL_

**To Reproduce**
1. Use `p:inputNumber` with minValue="1" (or greater than 0)
2. Click on javascript console (e.g. firefox) and the warning shows up

**Expected behavior**
No warning and if PF detects a "minValue > 0" in `p:inputNumber`, then it should use the the `invalid` option for the `overrideMinMaxLimits` setting for this.

### LINKED ISSUE #9328: InputNumber: "minValue>0 <1" shows js warning
Originally fixed in https://github.com/primefaces/primefaces/issues/8125

inputNumber is still broken if minValue is a decimal between 0 and 1, just try this to see:
```xhtml
<p:inputNumber minValue="0.01"/>
```
I think that here:
https://github.com/primefaces/primefaces/blob/e8d3e2968dac710b725a9f2e761178c22063f0d0/primefaces/src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js#L42-L45
the if condition should be:
```js
if (this.cfg.minimumValue > 0.0000001 || this.cfg.maximumValue < 0)
```

cc @NicolaIsotta

melloware:
Fix #9328: InputNumber minVal > 0
Fix #8125: InputNumber minVal > 0

Testing this PR against integration tests

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js
@@ -40,7 +40,7 @@ PrimeFaces.widget.InputNumber = PrimeFaces.widget.BaseWidget.extend({
         this.disabled = cfg.disabled;
 
         // GitHub #8125 minValue>0 shows js warning and quirky behavior
-        if (this.cfg.minimumValue >= 1 || this.cfg.maximumValue < 0) {
+        if (this.cfg.minimumValue > 0.0000001 || this.cfg.maximumValue < 0) {
             this.cfg.overrideMinMaxLimits = 'invalid';
         }
 


melloware:
Had to make it 0.00000001 that is what `@DecimalMin` sets and all the Integration tests are now passing.

PR #9333, opened 2022-10-19
Fix #9331 - UIColumn: remove filterOptions

### LINKED ISSUE #9331: UIColumn: remove filterOptions
### Description

You can do the same using dynamic filters introduced in https://github.com/primefaces/primefaces/issues/7469

Delete it

### Additional context

I realized that filterOptions was used to cover this point, see doc:

> If you’d like to use a dropdown instead of an input field to only allow
predefined filter values use _filterOptions_ attribute and a collection/array of selectitems as value.

But you can achieve the same by using facet filter no? Am I missing something?

Rapster:


melloware (approved):

melloware:
I realized I actually use this a lot when I have enums or true/false columns.

```xml
<p:column headerText="Active Flag" filterBy="#{row.active}" filterOptions="#{trueFalseFilterOptions}">
		<melloware:column-boolean value="#{row.active}" />
</p:column>
```

And I have this `ApplicationScoped` producer so all my boolean columns anywhere I can reuse.

```java
 /**
    * Produces a filter option list with True/False as display values for T/F
    * flags respectively.
    *
    * @return a List of SelectItem values for the filter, including All
    */
   @Produces
   @Named
   public List<SelectItem> getTrueFalseFilterOptions() {
      if (trueFalseFilterOptions == null) {
         trueFalseFilterOptions = new ArrayList<SelectItem>();
         trueFalseFilterOptions.add(FilterUtils.ALL_ITEM);
         trueFalseFilterOptions.add(new SelectItem(BooleanUtils.FALSE, StringUtils.capitalize(BooleanUtils.FALSE)));
         trueFalseFilterOptions.add(new SelectItem(BooleanUtils.TRUE, StringUtils.capitalize(BooleanUtils.TRUE)));
      }
      return trueFalseFilterOptions;
   }
```

Rapster:
Ok i see, but what you can easily do is to create a facet filter in your column-boolean component and pass your `trueFalseFilterOptions` as an input no?

melloware:
Yep just noting that we do use it a lot i thought we werent using it at all :)

tandraschko:
will check it the next weeks, im finally back from holiday

tandraschko:
@Rapster can you resolve conflicts? i finally reviewed it and checked the history; the change looks good!

PR #9344, opened 2022-10-26
Fix #9343: MenuItem respect iconPos

### LINKED ISSUE #9343: Menu: Respect iconPos attribute
Currently the `MenuItem` has the `iconPos` property but I can find no where in the code it is being used or respected.

For example:
```xml
<p:menuitem value="My Site" icon="pi pi-fw pi-external-link" iconPos="right"/>
<p:menuitem value="Help" icon="pi pi-fw pi-external-link" iconPos="right"/>
```

Renders like this...
image

Instead of this...
image

cc @FlipWarthog

melloware:
Fix #9343: MenuItem respect iconPos

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/menu/menu.css
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/menu/menu.css
@@ -65,6 +65,13 @@
     float:left;
 }
 
+.ui-menu .ui-menuitem-icon-right {
+    margin-left: 0.5rem;
+}
+.ui-menu .ui-menuitem-icon-left {
+    margin-right: 0.5rem;
+}
+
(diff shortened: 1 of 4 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/menu/menu.css
```diff
@@ -65,6 +65,13 @@
     float:left;
 }
 
+.ui-menu .ui-menuitem-icon-right {
+    margin-left: 0.5rem;
+}
+.ui-menu .ui-menuitem-icon-left {
+    margin-right: 0.5rem;
+}
```
melloware, line 73: smart defaults that can be overridden in the themes.

PR #9358, opened 2022-10-28
Fix #9357: Datatable disableSelection render checkbox/radio as disabled

### LINKED ISSUE #9357: DataTable: `disabledSelection` not properly styling checkbox. breaking radio
### Describe the bug

When `disabledSelection` is used on a dataTable with `selectionMode="multiple"`, the disabled rows' checkboxes are not properly styled (missing `ui-state-disabled`).

Additionally, if using single select, the same behavior happens while also allowing multiple selections.

Doing some digging, this may relate to changes made for #8279

### Reproducer

Using a local checkout of the PrimeFaces Showcase, add `disabledSelection="true"` to both "Checkbox" and "RadioButton" examples.

pf1
pf2
pf3
pf4

### Expected behavior

Non-selectable rows should be properly highlighted. 

### PrimeFaces edition

_No response_

### PrimeFaces version

13.0.0

### Theme

All

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

11

### Browser(s)

Chrome latest

melloware:
Fix #9357: Datatable disableSelection render checkbox/radio as disabled

### DIFF
--- a/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datatable/DataTable032Test.java
+++ b/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datatable/DataTable032Test.java
@@ -27,6 +27,7 @@ import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 import org.openqa.selenium.By;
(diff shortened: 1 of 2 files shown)


melloware:
Updated integration test to check for `ui-state-disabled` on the radio button.

Issue #9349, opened 2022-10-27
DataTable: when it is updated, the filter values are set empty. (LazyDataModel + "special" API-usage)


mertsincan:
As in the example below, after selecting a value from the selectcheckbox menu, the filter values are lost when the table is updated. It works fine with version 11.
```
<h:form>
      <p:dataTable var="customer" value="#{dtLazyView.lazyModel}" paginator="true" rows="10"
                   paginatorTemplate="{RowsPerPageDropdown} {FirstPageLink} {PreviousPageLink} {CurrentPageReport} {NextPageLink} {LastPageLink}"
                   rowsPerPageTemplate="5,10,15" selectionMode="single"
                   selection="#{dtLazyView.selectedCustomer}"
                   id="customerTable" widgetVar="_customerTable">

        <p:column headerText="Name" sortBy="#{customer.name}">
          <h:outputText value="#{customer.name}"/>
        </p:column>
        <!-- This filter hides correctly, but is not working when updated -->
        <p:column headerText="country " sortBy="#{customer.country}" filterBy="#{customer.country.name}">
          <f:facet name="filter">
            <p:selectCheckboxMenu id="multiple" value="#{dtFilterView.selectCountries}" style="min-width: 15rem"
                                  multiple="true" scrollHeight="250" updateLabel="true">
              <p:ajax event="change" process="@this" update="@form:customerTable"/>
              <f:selectItems value="#{dtLazyView.countries}"/>
            </p:selectCheckboxMenu>
          </f:facet>
          <h:outputText value="#{customer.country.name}"/>
        </p:column>
      </p:dataTable>

    </h:form>
```

tandraschko:
@mertsincan that reverts my fix for: https://github.com/primefaces/primefaces/issues/9040

mertsincan:
Hi @tandraschko, yes but unfortunately, there are at least 5 issues reported by PROs regarding https://github.com/primefaces/primefaces/issues/9040 
When they update the PF version, they completely lose the value they submitted in the filter. 
Exp; 
[primefaces-test.zip](https://github.com/primefaces/primefaces/files/9879004/primefaces-test.zip)
Maybe, you can have a look at this example to be compatible with your fix.
Note: As I said above, it works fine with PF11, PF10 and 8.

christophs78:
We need to add integrationtests for this specific situation. (or extend exisiting ones)

christophs78:
[...]

PR #9359, opened 2022-10-28
#9349 ITs + filter done the right way

christophs78:


melloware (approved):

christophs78:
Some more MyFaces-work to do....

christophs78:
> Some more MyFaces-work to do....

now green with MyFaces too

melloware:
Looks great!

PR #9365, opened 2022-10-31
DataExporter: Integration tests

melloware:
Checks for downloads of all 4 types + custom Text Exporter like showcase.

I compare equality for all types except for PDF as I could not find a good way to compare two PDF's


christophs78:
Great to see Integrationtests for DataExporter!

PDF fails because there are meta-data-differences between the reference PDF and to one generated by on-the-fly?
Maybe we can do some basic checks with https://librepdf.github.io/OpenPDF/docs-1-2-7/index.html?com/lowagie/text/pdf/parser/PdfTextExtractor.html? (or with some help by PDFBox https://stackoverflow.com/questions/23813727/how-to-extract-text-from-a-pdf-file-with-apache-pdfbox)

melloware:
Yep PDF must generate different timestamps etc inside the document.  XLSX does the same thing but I found a nice way using POI to compare to XLSX files.

melloware:
OK   I think it worked!

```java
 final PdfReader actualPdf = new PdfReader(actual.getAbsolutePath());
final String actualPdfContent = new PdfTextExtractor(actualPdf).getTextFromPage(1);

final PdfReader expectedPdf = new PdfReader(expected.getAbsolutePath());
final String expectedPdfContent = new PdfTextExtractor(expectedPdf).getTextFromPage(1);

Assertions.assertTrue(Objects.equal(actualPdfContent, expectedPdfContent), message);
```

primefaces-selenium/primefaces-selenium-core/src/main/java/org/primefaces/selenium/internal/DefaultWebDriverAdapter.java
```diff
@@ -85,15 +87,21 @@ public WebDriver createWebDriver() {
                 FirefoxOptions firefoxOptions = new FirefoxOptions();
                 firefoxOptions.setPageLoadStrategy(PageLoadStrategy.NORMAL);
                 firefoxOptions.setHeadless(config.isWebdriverHeadless());
-                if (!config.isWebdriverHeadless()) {
-                    firefoxOptions.setCapability(CapabilityType.LOGGING_PREFS, logPrefs);
-                }
+                firefoxOptions.setLogLevel(FirefoxDriverLogLevel.INFO);
```
christophs78, line 90: Are we sure we want to modify this?

christophs78: Beside this one thing everything fine and ready to be merged.

melloware: Yes that is marked as deprecated and a Chrome only option `LoggingPrefs` so I swapped it for the correct Firefox setting which actually still doesn't even work yet.  The Chrome driver is the only one that captures log output right now.

PR #9388, opened 2022-11-08
#9384 CommandLink: Invalid HTML when disabled

### LINKED ISSUE #9384: CommandLink: Invalid HTML when disabled
### Describe the bug

Since Primefaces 12 (Issue: #8218 / PR: #8219) disabled [CommandLinks](https://www.primefaces.org/showcase/ui/button/commandLink.xhtml?jfwid=d4529) are rendered with invalid HTML, because an attribute `disabled=true` is renderend. But this in not allowed on anker-tag (`<a>`).

> Attribute disabled not allowed on element a at this point.
> r Post</a><a id="j_idt343:disabled" href="#" class="ui-commandlink ui-widget ui-state-disabled" tabindex="-1" …meFaces.ab({s:&quot;j_idt343:disabled&quot;,f:&quot;j_idt343&quot;});return false;" disabled="true">Disab

pf_commandlink

### Reproducer

See showcase

### Expected behavior

Valid HTML is rendered.

### PrimeFaces edition

Community

### PrimeFaces version

12.0.0

### Theme

_No response_

### JSF implementation

All

### JSF version

2.2

### Java version

1.8

### Browser(s)

_No response_

jepsar:
Fix #9384

No CSS change was needed as we already have

https://github.com/primefaces/primefaces/blob/2e79cbc628fa98885cad4688813fc7e82f6f3a97/primefaces/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.css#L66-L69

### DIFF
--- a/primefaces/src/main/java/org/primefaces/util/HTML.java
+++ b/primefaces/src/main/java/org/primefaces/util/HTML.java
@@ -109,7 +109,6 @@ public class HTML {
         "charset",
         "coords",
         "dir",
-        "disabled",
         "hreflang",
         "rel",
         "rev",
--- a/primefaces/src/main/java/org/primefaces/util/HTML.java
+++ b/primefaces/src/main/java/org/primefaces/util/HTML.java
@@ -93,7 +93,6 @@ public class HTML {
         "charset",
         "coords",
         "dir",
-        "disabled",
         "hreflang",
         "rel",
         "rev",
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.commandlink.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.commandlink.js
@@ -79,8 +79,7 @@ PrimeFaces.widget.CommandLink = PrimeFaces.widget.BaseWidget.extend({
     disable: function() {
         this.jq.removeClass('ui-state-hover ui-state-focus ui-state-active')
                 .addClass('ui-state-disabled')
-                .attr('tabindex', '-1')
-                .attr('aria-disabled', 'true');
+                .attr('tabindex', '-1');
     },
 
     /**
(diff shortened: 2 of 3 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.commandlink.js
```diff
@@ -79,16 +79,15 @@ PrimeFaces.widget.CommandLink = PrimeFaces.widget.BaseWidget.extend({
     disable: function() {
         this.jq.removeClass('ui-state-hover ui-state-focus ui-state-active')
                 .addClass('ui-state-disabled')
-                .attr('tabindex', '-1')
-                .attr('aria-disabled', 'true');
+                .attr('tabindex', '-1');
```
melloware, line 82: Don't you want to leave these aria-disabled so if they use the widget to control disabled?

melloware (changes requested):

primefaces/src/main/java/org/primefaces/util/HTML.java
```diff
@@ -93,7 +93,6 @@ public class HTML {
         "charset",
         "coords",
         "dir",
-        "disabled",
         "hreflang",
```
melloware, line 96: What about `LINK_ATTRS_WITHOUT_EVENTS_AND_STYLE ` ??? It also has disabled?

melloware (changes requested):

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.commandlink.js
jepsar: There was a warning that you should not specify it on an anchor with an `href` attribute.

melloware: Oh OK

melloware (approved):

Bukama (approved):
`disabled` and `aria-disabled` attributes are getting removed, so 👍

PR #9402, opened 2022-11-13
#9279 SelectOneMenu: ARIA compliance

### LINKED ISSUE #9279: SelectOneMenu: ARIA compliance
### Description

[Primeface's SelectOneMenu](https://www.primefaces.org/showcase/ui/input/oneMenu.xhtml?) simulates a combobox using a `div` with `role="combobox"` and a hidden `input` element (in showcase it has the id `j_idt343:option_focus` ) for the current selected value.
To this hidden input element has the role `text` is assigned, as colored in the screenshot.

When using a screenreader on a page with such an combobox it reads out

> text input with auto-completion empty

This might be a bit confused so let's get the unproblematic parts ot it out of the way first:
* "auto-completion" comes from `aria-autocomplete="list"` which is absolutely correct set, see [Moz Dev, ARIA autocomplete values](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-autocomplete#values)
* "empty" comes from a not yet selected value and is updated when selecting one

So this leaves us with the first value "text input", which comes from the `role` attribute of the hidden input element and indicates that on the page there is an text input element instead of an combobox.

According to the accessibilty test team at my work this is problematic for users, who are visible impaired.

primefaces_role_selectOneMenu

Change the `role` attribute of the hidden input element from `text` to `combobox` so screen readers read out "combobox".

### Additional context

_No response_

christophs78:
Still reading doc´s, analysing examples, trying out things and talking to my disabled colleague.

### DIFF
--- a/primefaces-showcase/src/main/webapp/ui/input/oneMenu.xhtml
+++ b/primefaces-showcase/src/main/webapp/ui/input/oneMenu.xhtml
@@ -54,7 +54,7 @@
                     </div>
 
                     <div class="field col-12 md:col-4">
-                        <p:outputLabel value="Label"/>
+                        <p:outputLabel for="@next" value="Label"/>
                         <p:selectOneMenu id="labeled" value="#{selectOneMenuView.labeled}" label="Always Show Label"
                                         alwaysDisplayLabel="true">
                             <f:selectItems value="#{selectOneMenuView.cities}"/>
@@ -137,12 +137,12 @@
                     </div>
 
                     <div class="field col-12 md:col-4">
-                        <p:outputLabel value="Icons" />
+                        <p:outputLabel for="selectOneMenuIcon" value="Icons" />
                         <h:panelGroup styleClass="ui-inputgroup">
                             <h:panelGroup id="icon" styleClass="ui-inputgroup-addon">
                                 <i class="pi pi-#{selectOneMenuView.icon}"/>
                             </h:panelGroup>
-                            <p:selectOneMenu value="#{selectOneMenuView.icon}" var="item">
+                            <p:selectOneMenu id="selectOneMenuIcon" value="#{selectOneMenuView.icon}" var="item">
                                 <f:selectItems value="#{['flag','wallet','map','link']}" var="item" itemLabel="#{item}"
                                     itemValue="#{item}" />
                                 <p:column><i class="pi pi-#{item}" /> #{item} </p:column>
--- a/primefaces-showcase/src/main/webapp/ui/input/oneMenu.xhtml
+++ b/primefaces-showcase/src/main/webapp/ui/input/oneMenu.xhtml
@@ -54,8 +54,8 @@
                     </div>
 
                     <div class="field col-12 md:col-4">
-                        <p:outputLabel for="@next" value="Label"/>
-                        <p:selectOneMenu id="labeled" value="#{selectOneMenuView.labeled}" label="Always Show Label"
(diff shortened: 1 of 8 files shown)


christophs78:
Some notes to the current state:
- use focus-input or not?
   - PrimeNG, PrimeReact and PrimeFaces currently have a focus-input
   - PrimeVue does not have one (only uses a span marked with role="combobox")
   - https://github.com/primefaces/primefaces/issues/9279 and my disabled colleague using JAWS complain about focus input (which was originally introduced to improve accessibility)
   Because they see two inputs for one SelectOneMenu. (focus-input and div (or span) marked with role="combobox")
   - without focus-input clicking onto the label referencing the focus input via for-attribute does not focus the input (so i introduced some more java-script-work-arrounds to compensate this)
- my disabled colleague using JAWS complains there´s a delay of a few seconds when he opens the dropdown (http://primefaces.org/showcase/ui/input/oneMenu.xhtml); before and with this PR
- editable is currently broken in this PR (i´ll try to fix it)

Accessibility is tremendous hard. WCAG is one thing, what real disabled people tell you is sometimes a different thing. Maybe sometimes it even comes down to specific preferences of disabled people. (which screenreader do they use, which mode of the screenreader do they use, which customizing, ...)

melloware:
Appreciate the work on this.  Yeah this does seem like a very difficult problem.

christophs78:
> * my disabled colleague using JAWS complains there´s a delay of a few seconds when he opens the dropdown (http://primefaces.org/showcase/ui/input/oneMenu.xhtml); before and with this PR

After removing things like sourcecode-preview, ui-configuration and cutting down the menu this issue is gone and my disabled colleague is happy. So our new implementation looks positiv.
So i "only" have to get editable-mode working again.

christophs78:
Removing focus-input which distracts disabled people has some downsides...
When you have entries starting with the same letters (Germany, Greece; Slovakia, Slovenia, Spain) selection via typing is worse because only the last letter typed is recognized.

With focus-input we do not have this issue because focus-input knows what hat been typed up to this point.
image

So most not disabled people will dislike what i have done with this PR.

--> more work necessary to store in some variable what was typed and so on ...

christophs78:
Probably almost done - but some regression regarding alwaysDisplayLabel not covered yet by integrationtests.

primefaces-showcase/src/main/webapp/ui/input/oneMenu.xhtml
```diff
@@ -54,7 +54,7 @@
                     </div>
 
                     <div class="field col-12 md:col-4">
-                        <p:outputLabel value="Label"/>
+                        <p:outputLabel for="@next" value="Label"/>
```
christophs78: Ok. Adding  for="@next" caused the "regression". So i´ll remove this.

christophs78:
Done from my side. I´ll try to let this review my disabled colleage one more time.
And it would help if some of you do some manual testing too. SelectOneMenu is one of the most often used components where we should not introduce regressions.
@Bukama: Would it be possible to do some testing too?

melloware (approved):

christophs78:
Green light from my disabled colleague.

christophs78:
No styling-changes required.  (I checked against all existing free themes coming with PrimeFaces.)
For some exotic project specific styling there is a small risk that some minor adjustments are necessary. (Overall i did not change styleclasses or something like this.)

PR #9430, opened 2022-11-21
Fix #9429: rowSelectMode="none" don't style as selectable

### LINKED ISSUE #9429: DataTable: selectionMode=multiple with rowSelectMode=none still shows cursor pointer
### Describe the bug

DataTable Checkbox selection with `rowSelectMode="none"` still sets the `cursor: pointer` style on hover of the entire row, despite only being able to select via the Checkbox.

### Reproducer

This issue can be reproduced running the showcase locally. Modify the Checkbox example in the `ui/data/datatable/selection.xhtml` showcase page and set `rowSelectMode="none"`. Run the showcase and hover the rows in that example.

### Expected behavior

If a row cannot be selected by clicking on the row itself, the cursor should remain as the arrow rather than changing to a pointer.

### PrimeFaces edition

Community

### PrimeFaces version

13.0.0

### Theme

All

### JSF implementation

All

### JSF version

2.3

### Java version

11

### Browser(s)

All

melloware:
Fix #9429: rowSelectMode="none" don't style as selectable

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
@@ -1243,7 +1243,7 @@ public class DataTableRenderer extends DataRenderer {
         String rowStyleClass = getStyleClassBuilder(context)
                 .add(DataTable.ROW_CLASS)
                 .add(rowIndex % 2 == 0, DataTable.EVEN_ROW_CLASS, DataTable.ODD_ROW_CLASS)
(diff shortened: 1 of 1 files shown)


FlipWarthog:
It looks like `cursor: pointer` also gets added on hover. Should that be checked too?

https://github.com/primefaces/primefaces/blob/fd04840e0615b234028437c9f1701ae180fcf881/primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.css#L119-L123

melloware:
This fix doesn't work.  By removing `selectable` from the row it makes the checkboxes uncheckable.

PR #9434, opened 2022-11-22
Fix #9429: rowSelectMode="none" remove cursor pointer

### LINKED ISSUE #9429: DataTable: selectionMode=multiple with rowSelectMode=none still shows cursor pointer

melloware:
Fix #9429: rowSelectMode="none" remove cursor pointer

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
@@ -809,6 +809,9 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
             if(this.cfg.rowSelectMode !== 'none') {
                 this.bindRowEvents();
             }
+            else {
+                this.jq.find('tr.ui-datatable-selectable').css('cursor', 'default');
+            }
         }
         else if(this.cfg.selectionMode === 'checkbox') {
             this.bindCheckboxEvents();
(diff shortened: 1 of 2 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
```diff
@@ -809,6 +809,9 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
             if(this.cfg.rowSelectMode !== 'none') {
                 this.bindRowEvents();
             }
+            else {
+                this.jq.find('tr.ui-datatable-selectable').css('cursor', 'default');
```
jepsar, line 813: Why are there rows classed with `ui-datatable-selectable` when row select mode is `none`?

FlipWarthog: Take a look at the initial PR #9430. It's a difference between selecting via row click vs. actually clicking the checkbox, but removing the style affects the checkboxes too.

PR #9438, opened 2022-11-23
VirusScan: Provide defaults for host and port #9437

### LINKED ISSUE #9437: VirusScan: Provide defaults for host and port
### Description

Provide defaults for host and port such that no explicit configuration in web.xml is needed anymore when these defaults are fine.

Actually by looking deeper into this issue, I also detected a bug that the 

primefaces.fileupload.CLAM_AV_FILE

message is not correct since '{0}' cannot be replaced by the standard message formatter. here, the message should include ''{0}'' instead of '{0}'.

I will provide a PR immediately, implementing/fixing both issue and even generalizing the ClamDeamonScanner such that the implementation of a custom ClamDeamonScanner allows to access all virusscan paramters from the ClamDaemonClient.

As the other virus scan context parameter, make also host and port optional by providing according defaults:

* host=localhost
* port=3310

### Additional context

_No response_

tuerker:
This PR does the following:

1) Bugfix: In primefaces.fileupload.CLAM_AV_FILE '{0}' is replaced by ''{0}''.

2) Provide defaults for host and port (localhost:3310) such that no explicit configuration in web.xml is needed anymore when these defaults are fine.

3) Generalized ClamDeamonScanner such that the implementation of a custom ClamDeamonScanner allows to access all virusscan parameters from the ClamDaemonClient.


primefaces/src/main/resources/org/primefaces/Messages.properties
```diff
@@ -13,7 +13,7 @@ primefaces.datatable.SORT_ASC = Ascending
 primefaces.datatable.SORT_DESC = Descending
 primefaces.datatable.SORT_LABEL = Sort
 primefaces.dialog.aria.CLOSE = Close
-primefaces.fileupload.CLAM_AV_FILE = ClamAV has identified file '{0}' as potentially dangerous with message '{1}'
```
jepsar, line 16: Don't we need to do this in all message bundles and documentation?

https://github.com/primefaces/primefaces/search?p=1&q=primefaces.fileupload.CLAM_AV_FILE

primefaces/src/main/java/org/primefaces/virusscan/impl/ClamDaemonClient.java
```diff
@@ -48,34 +48,56 @@ public class ClamDaemonClient {
...
     // will reply with INSTREAM size limit exceeded and close the connection."
-    private static final int CHUNK_SIZE = 2048;
-    private static final int DEFAULT_TIMEOUT = 30000;
+    public static final int DEFAULT_BUFFER = 2048;
+    public static final String DEFAULT_HOST = "localhost";
+    public static final int DEFAULT_PORT = 3310;
+    public static final int DEFAULT_TIMEOUT = 30000;
 
-    private final String hostName;
+    private final int buffer;
```
jepsar, line 56: Why a rename of chunkSize and hostName? This would affect people that have extended this class.

jepsar (changes requested):
Thanks for your PR! Few small questions from my end though.

Oh, and can you please add this to https://github.com/primefaces/primefaces/blob/master/docs/13_0_0/gettingstarted/whatsnew.md

primefaces/src/main/resources/org/primefaces/Messages.properties
tuerker: Extended now whatsnew.md and also did some more fixes in the other messages.properties files

primefaces/src/main/java/org/primefaces/virusscan/impl/ClamDaemonClient.java
tuerker: this was not possible yet since these attributes were private uptonow.

tuerker: the test are not running through. they should not be affected by my commits.

jepsar: Ah yes, there were no getters.

primefaces/src/main/java/org/primefaces/virusscan/impl/ClamDaemonClient.java
```diff
@@ -48,34 +48,56 @@ public class ClamDaemonClient {
...
+    private final String host;
     private final int port;
     private final int timeout;
-    private final int chunkSize;
 
     /**
-     * @param hostName The hostname of the server running clamav-daemon
+     * @param host The hostname of the server running clamav-daemon
      * @param port The port that clamav-daemon listens to(By default it might not listen to a port. Check your clamav configuration).
      * @param timeout zero means infinite timeout. Not a good idea, but will be accepted.
```
melloware, line 64: Missing buffer param in JavaDoc

primefaces/src/main/java/org/primefaces/virusscan/impl/ClamDaemonScanner.java
```diff
@@ -57,8 +57,7 @@ public class ClamDaemonScanner implements VirusScanner {
 
     @Override
     public boolean isEnabled() {
-        ExternalContext ctx = FacesContext.getCurrentInstance().getExternalContext();
-        return ctx.getInitParameter(CONTEXT_PARAM_HOST) != null && ctx.getInitParameter(CONTEXT_PARAM_PORT) != null;
+        return true;
```
melloware: I have to take a deeper dive into where this enabled flag is used but does this turn on ClamAv automatically when it shouldn't?

melloware (changes requested):

primefaces/src/main/java/org/primefaces/virusscan/impl/ClamDaemonScanner.java
tuerker: the virusscan is controlled by the corresponding virusScan attribute of the fileUpload component, isn't it?

melloware: But what happens if VirusTotal and ClamAV both return true here it will automatically try to use this one even if a user doesn't want to right?

primefaces/src/main/java/org/primefaces/virusscan/impl/ClamDaemonClient.java
tuerker: added now

primefaces/src/main/java/org/primefaces/virusscan/impl/ClamDaemonScanner.java
tuerker: am i wrong that the virus scan is only triggered when somebody uses 

virusScan="true" in the fileUpload component?

melloware: You have that correct.  But its pluggable with 3 different virus scanning engines either VirusTotal, ClamAV, or a Custom one from the user.  This flag controls whether to "try" this virusanner or not.  Previously it would only be attempted if the user had configued ClamAV. Now you are forcing users using `viruscan="true"` to ALWAYS use ClamAV when they don't want to.

Documented here: https://primefaces.github.io/primefaces/12_0_0/#/components/fileupload?id=measures

tuerker: ok. then it would be needed a possibility to switch-off the default ClamAV scanner.

what about introducing

```
  <context-param>
    <param-name>primefaces.virusscan.CLAMAV_DEFAULT</param-name>
  </context-param>

```
wdyt?

melloware: But my original solution already worked.  You have to defined a server and port ANYWAY and that is what triggered the enablement.    So now you are introducing a new parameter for no reason all just to not have someone fill out server and port?

tandraschko: +1 melloware

tuerker: ok. i reverted that part such that the context parameter host and port are still required for enabling

melloware (approved):

PR #9445, opened 2022-11-24
VirusScan: Provide defaults for host and port #9437 (fix when ClamDaemonScanner is used without having a FacesContext)

tuerker:
Rational of this PR: I want to reuse the virus scanner configured with primefaces also from my Web services, i.e., without having a FacesContext and without introducing yet another dependency and configuration. Currently this does not work. Therefore, the ClamDaemonScanner is now extended to check whether a FacesContext is given. If yes, it uses the settings from web.xml; otherwise it takes its default settings.

@melloware: is there a "easy" way to access the init params of web.xml here in this place without having a FacesContext? That would be the next well-needed improvement. Then, the virus scanner would be used with a single configuration for both GUI as wellas Web services.


melloware:
That is a @tandraschko question I have always just followed the pattern of the rest of the code base in retrieving web.xml params.

tandraschko:
TBH i dont like that
ClamDaemonScanner received the params as constructor param, which is perfectly fine compared that it ready web.xml params by itself

melloware:
I agree this isn't saving much here its just moving pieces around on the chessboard.

tuerker:
@tandraschko how you would advice to make the virus scanner reusable also in case where the FacesContext is not given without being forced to maintain yet another dependency with possibly different configuration parameters?

tandraschko:
i think you should just use ClamDaemonClient and not the wrapper ClamDaemonScanner, which heavily depends on JSF APIs

it even has a constructor with your default params and so on

tuerker:
yes, this works fine and was exactly what i was doing. my method looks as follows:

```
      public static String virusScan(UploadedFile file) {
        ClamDaemonScanner clamDaemonScanner = new ClamDaemonScanner();
        ClamDaemonClient clamDaemonClient = clamDaemonScanner.getClamAvClient();
        try {
            clamDaemonClient.ping();
        } catch (IOException e) {
            String msg = Messages.get("virusScannerPingError").replace("{0}", clamDaemonClient.getHost()).replace("{1}", String.valueOf(clamDaemonClient.getPort()))
                .replace("{2}", String.valueOf(clamDaemonClient.getTimeout()));
            logger.severe(msg);
            return msg;
        }
        try {
            clamDaemonScanner.scan(file);
        } catch (VirusException exception) {
            String msg = Messages.get("virusScannerInfectedFileReport").replace("{0}", file.getFileName());
            logger.severe(msg);
            return msg;
        }
        return null;
    }
```

but when i invoke this method from my web service it does not work anymore since there is no FacesContext given which is needed to correctly construct the ClamDaemonClient. There it currently runs into a null pointer exception.

melloware:
@tuerker yes when I donated this class it worked standalone in my project the wrapper was only to allow it to work inside JSF and be pluggable in the FileUpload VirusScan SPI.

tandraschko:
we as PrimeFaces, rely on JSF config mechanism, which is FacesContext
we dont have any other way to get the ServletContext and receive context params and so on
and if you check CalmDeamonClient, thats exactly the thing you want
it doenst rely on JSF and you can pass your config, maybe via JAX-RS context or try to crap somehow ServletContext and pass it

tuerker:
this fix here avoids the NPE and uses the hard-coded default settings. as said above, ideally the configuration done in the web.xml could be also taken here. therefore, my question how to access here the web.xml params without having a FacesContext?

tandraschko:
you need a ServletContext, you must receive it from your WS framework
and even then, we should not use it in PrimeFaces as it wouldnt work in portlets - So the Scanner will always rely on JSF and FacesContext, the Client won't

tuerker:
I am not happy with that. My thought was to get rid of the additional dependency to import the ClamAV virus scanner explicitly. So, i can also stick to our old solution where all the file uploads are checked via a listener.

tandraschko:
i can just say that our code in PF is fine as the logic is a own class, the Client, which is indepdent of JSF
and the wrapper around with JSF dependency is in the Scanner
everything else like magically receiving configuration isnt possible for us

melloware:
+1 tandraschko

melloware:
@tuerker when I donated this code I actually kept the client separate because I use it in a CDI interceptor like this...

```java
@RequestScoped
public class VirusScannerService {

   /* Logger for this class */
   private static final Logger LOG = LoggerFactory.getLogger(VirusScannerService.class);

   /** The IP Address of the machine where ClamAV is running */
   @Inject
   @Configure(key = "com.melloware.web.virus.scanner.host", defaultValue = "127.0.0.1") // NOPMD
   String scannerHost;

   /** The Port of the machine where ClamAV is running */
   @Inject
   @Configure(key = "com.melloware.web.virus.scanner.port", defaultValue = "3310")
   Integer scannerPort;

   /**
    * The timeout of how much time to give CLamAV to scan the virus before
    * failing
    */
   @Inject
   @Configure(key = "com.melloware.web.virus.scanner.timeoutmilliseconds", defaultValue = "60000")
   Integer timeout;

   /**
    * Property to enable or disable the virus scanner. Useful for developers who
    * don't want to scan files locally.
    */
   @Inject
   @Configure(key = "com.melloware.web.virus.scanner.enabled", defaultValue = BooleanUtils.TRUE)
   Boolean scannerEnabled = Boolean.TRUE;

   /**
    * Scans an {@link InputStream} for any viruses. If any virus is detected an
    * exception will be thrown.
    *
    * @param filename the name of the file to be scanned
    * @param inputStream the {@link InputStream} of the file to scan
    * @throws MaliciousFileException if any virus is detected
    */
   public void scanFile(final String filename, final InputStream inputStream) throws MaliciousFileException {
      LOG.info("Starting the virus scan for file: {}", filename);
      if (!scannerEnabled) {
         LOG.warn("Virus Scanner is currently disabled!");
         return;
      }

      try {
         final ClamAvClient client = getClient();
         final byte[] reply = client.scan(inputStream);
         final String message = new String(reply, StandardCharsets.US_ASCII).trim();
         LOG.info("Scanner replied with message: {}", message);
         if (!ClamAvClient.isCleanReply(reply)) {
            final String error = String.format("Scan detected viruses in %s! Response = %s", filename, message);
            LOG.error("ClamAV {}", error);
            throw new MaliciousFileException(error);
         }
      } catch (final MaliciousFileException ex) {
         throw ex;
      } catch (final RuntimeException | IOException ex) {
         final String error = String.format("Unexpected error scanning file - %s",
                  ExceptionUtils.getRootCauseMessage(ex));
         throw new MaliciousFileException(error, ex);
      } finally {
         LOG.info("Finished virus scanning {}!", filename);
      }
   }

   /**
    * Returns a new ClamAvClient which can be overridden in unit tests.
    *
    * @return the {@link ClamAVClient}
    */
   ClamAvClient getClient() {
      return new ClamAvClient(scannerHost, scannerPort, timeout);
   }

}
```

tuerker:
thank you both for the explainations.

PR #9455, opened 2022-11-28
Fix #9453: CSP with PrimeExceptionHandler

### LINKED ISSUE #9453: AjaxExceptionHandler + CSP - OnClick is not triggered
### Describe the bug

If i use AjaxExceptionHandler with enabled CSP all inline JS (e.g. onClick) inside the Dialog will not be triggered. If i disable CSP its working. If i open the Dialog manually and let CSP enabled its also working. But both together not working.

### Reproducer

[producer.zip](https://github.com/primefaces/primefaces/files/10104942/producer.zip)

### Expected behavior

_No response_

### PrimeFaces edition

_No response_

### PrimeFaces version

12.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

X.X

### Java version

11

### Browser(s)

_No response_

melloware:
Fix #9453: CSP with PrimeExceptionHandler

I started this PR and it works except for 1 thing.  The `PrimeFaces.csp.init` is not being registered because that happens in RENDER_RESPONSE phase but this error is in INVOKE_APPLICATION phase and never gets to the Phase Listener.


melloware:
There is still one more issue I am having trouble figuring out what is happening.

melloware:
OK fixed all issues but I had to add a second check if its an AJAX request not just a PostBack.

primefaces/src/main/java/org/primefaces/csp/CspState.java
```diff
@@ -53,7 +53,7 @@ public CspState(FacesContext context) {
      */
     public String getNonce() {
         if (nonce == null) {
-            if (context.isPostback()) {
+            if (context.isPostback() || context.getPartialViewContext().isAjaxRequest()) {
```
tandraschko, line 57: isnt a AJAX request always a postback?

melloware: SO I don't fully understand it but in `PrimeExceptionHandler` when it goes to determine if its an AJAX request it does this `context.getPartialViewContext().isAjaxRequest()`.  So in this use case it kept generating a new CSP token even though this was an AJAX request not a new request. When I tracked it down `context.isPostBack` was `false` but `context.getPartialViewContext().isAjaxRequest()` was true!

melloware: I am not sure if this has to do with the "phase" the ExceptionListener is catching the error or something else?

tandraschko:
in general your change looks good, +1

i just wonder if there would be a case now where CSP would be initialized twice (somehow if the PhaseListener + the ExceptionHandler runs)
does it hurt to init 2x times? Otherwise i would add a security check + remeber the init in the facesContext attributes for the current reuqest

melloware:
Let me check if init 2X affects it or to defend against it.  I will test today.

melloware:
I added a check to make sure if doesn't accidentally get called twice.

PR #9462, opened 2022-12-01
Fix #9461 improve CSP nonce handling and validation

### LINKED ISSUE #9461: HTTP-Header Content-Security-Policy Nonce can be overwritten by client for Ajax requests
### Describe the bug

A content security policy with a secure directive is used. However, it is possible to influence the script-src directive and thus bypass the protection of the CSP header. The nonce value is set by the client-side parameter primefaces.nonce. This allows the **nonce value** in **POST requests to be determined by the user**, bypassing the protection of the header. 

--> The nonce value should be randomly generated on the server side for each request/response. The user must not be able to influence the nonce value.

### Sample Request:

**Header:**
POST https://localhost:8444/testTab.jsf HTTP/1.1
Host: localhost:8444
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:106.0) Gecko/20100101 Firefox/106.0
Accept: application/xml, text/xml, */*; q=0.01
Accept-Language: en-US,en;q=0.5
Content-Type: application/x-www-form-urlencoded; charset=UTF-8
Faces-Request: partial/ajax
X-Requested-With: XMLHttpRequest
Content-Length: 963
...

**Body:**
javax.faces.partial.ajax=true&javax.faces.source=quick-filters%3AquickFilter-form%3AselectOneMenuTestDropdown&javax.faces.partial.execute=quick-filters%3AquickFilter-form%3AselectOneMenuTestDropdown&javax.faces.partial.render=testTable-form&javax.faces.behavior.event=change&javax.faces.partial.event=change&quick-filters%3AquickFilter-form=quick-filters%3AquickFilter-form&javax.faces.ViewState=499211688213529577%3A-8085365838550146838&**primefaces.nonce=testscript1234**

### Response:

**Header:**
HTTP/1.1 200 OK
Content-Type: text/xml;charset=utf-8
Content-Security-Policy: default-src 'none'; base-uri 'self'; frame-ancestors 'none'; connect-src 'self'; img-src 'self'; font-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'strict-dynamic' **'nonce-testscript1234'**
Pragma: no-cache
...

**Body:**
```
<xml version='1.0' encoding='utf-8'?>
<partial-response id="j_id1"><changes><update id="quick-filters:quickFilter-form-messages">
<![CDATA[<div id="quick-filters:quickFilter-form-messages" class="ui-messages ui-widget" aria-live="polite"</div>]]>/update>
<update id="testTable-form:testTable"><![CDATA[<tr data-ri="32660" class="ui-widget-content ui-datatable-even" role="row"><td role="gridcell" class=""><span class=" ">*******</span></td>
<td role="gridcell" class=""></td><td role="gridcell" class="">*******</td>
<td role="gridcell" class=""><span class=" ">*******</span></td>
<td role="gridcell" class="pre-wrap break-word"><span class=" "></span></td>
<td role="gridcell" class="">*******</td>
<td role="gridcell" class="">*******</td>
<td role="gridcell" class="">*******</td>
<td role="gridcell" class="column-ac-tion"><button id="testTable-form:testTable:32660:view-button" name="testTable-%form:test-Table:32660:view-button" class="ui-button ui-widget ui-state-default ui-corner-all ui-button-icon-only secondary-%button" type="submit"><span class="ui-button-icon-left ui-icon ui-c ui-icon-pa-per"></span><span class="ui-button-text %ui-c">ui-button</span></button>
```
**<script id="testTable-form:testTable:32660:view-button_s" type="text/javascript" %nonce="testscript1234">** 
```
PrimeFaces.cw("CommandButton","widget_testTable_form_test-Table_32660_view_button",{id:"testTable-form:testTable:32660:view-button"});</script></td></tr>
```

### Reproducer

_No response_

### Expected behavior

The nonce value should be randomly generated on the server side for each request/response. The user must not be able to influence the nonce value.

### PrimeFaces edition

_No response_

### PrimeFaces version

10.0.10

### Theme

Bootstrap

### JSF implementation

All

### JSF version

2.2.20

### Java version

17.0.3

### Browser(s)

Firefox 106.0.5 (64-bit)

NicolaIsotta:
fix #9461

### DIFF
--- a/primefaces/src/main/java/org/primefaces/csp/CspState.java
+++ b/primefaces/src/main/java/org/primefaces/csp/CspState.java
@@ -54,11 +54,15 @@ public class CspState {
     public String getNonce() {
         if (nonce == null) {
             if (context.isPostback()) {
-                nonce = context.getExternalContext().getRequestParameterMap().get(Constants.RequestParams.NONCE_PARAM);
+                nonce = (String) context.getViewRoot().getViewMap().get(Constants.RequestParams.NONCE_PARAM);
+                if (nonce == null) {
+                    nonce = context.getExternalContext().getRequestParameterMap().get(Constants.RequestParams.NONCE_PARAM);
+                }
                 validate(nonce);
             }
             else {
                 nonce = Base64.getEncoder().encodeToString(UUID.randomUUID().toString().getBytes(StandardCharsets.UTF_8));
+                context.getViewRoot().getViewMap().put(Constants.RequestParams.NONCE_PARAM, nonce);
             }
         }
 
@@ -76,7 +80,8 @@ public class CspState {
             throw new FacesException("Missing CSP nonce");
         }
         try {
-            Base64.getDecoder().decode(nonce);
+            String decodedNonce = new String(Base64.getDecoder().decode(nonce), StandardCharsets.UTF_8);
+            UUID.fromString(decodedNonce);
         }
         catch (Exception e) {
             throw new FacesException("Invalid CSP nonce", e);


NicolaIsotta:
Some notes:

- the nonce is now stored in the ViewMap, so it can't be overwritten by the user. If for some reason it can't be retrieved from the ViewMap (eg stateless views) the old method is used as fallback
- in every case, the decode nonce value is validated using java's built-in [`UUID.fromString`](https://docs.oracle.com/javase/8/docs/api/java/util/UUID.html#fromString-java.lang.String-)

tandraschko:
i really dont remember anymore why we didnt use the viewmap (stateless is a reason but the right way would be like in your PR)
maybe we had a code, which tried to access CSP to early in JSF lifecycle?

NicolaIsotta:
According to [these comments](https://github.com/primefaces/primefaces/commit/c614e01cd4a5dca71487f5c0852478ddfd6a58dc#r33359932) the reason was stateless views. I couldn't find any other mention of ViewMap drawbacks, even in the issue ( #3505 )

melloware (approved):

melloware:
I am good with these changes.  We have a bunch of nightly integration tests that run in CSP mode and we have a lot of time before 13.0 is final and we can let this fix burn in.  So I am inclined to merge it @tandraschko ?

Hurled4505:
I still got some concerns, would it be possible to somehow get rid of the fallback algorithm?
As this could allow users to input their own nonce when the website is in either in a stateless state or there is no nonce saved in the ViewMap and the nonce follows the UUID representation.

melloware:
@Hurled4505 but in a purely stateless state there is no "view map" right?  so the NONCE has to be stored somewhere?

NicolaIsotta:
> I still got some concerns, would it be possible to somehow get rid of the fallback algorithm? As this could allow users to input their own nonce when the website is in either in a stateless state or there is no nonce saved in the ViewMap and the nonce follows the UUID representation.

Another approach I thinked about was encrypting the nonce with AES or something like that. Guess Mojarra and MyFaces do something similar with the viewstate.

melloware:
Enncrypting it depending on the encryption is a trade off of performance.  If we can encrypt the same way as MyFaces or Mojarra that would be OK I would think!

NicolaIsotta:
Both Mojarra ([ByteArrayGuard](https://github.com/eclipse-ee4j/mojarra/blob/e4c1f84456dac659fc31ef7af8bbfad711e0876a/impl/src/main/java/com/sun/faces/renderkit/ByteArrayGuard.java)) and MyFaces ([StateUtils](https://github.com/apache/myfaces/blob/93a47f3c78716f0ff3c01227ee1cd5972aeb965b/impl/src/main/java/org/apache/myfaces/application/viewstate/StateUtils.java)) use AES by default (although MyFaces is much more configurable). It should be pretty fast in modern JDKs.

Anyway, the current approach is fine for my requirements so I'm not investigating further, at least in the short term. If anyone wants to work on this, I think [ClientSideStateHelper](https://github.com/eclipse-ee4j/mojarra/blob/e4c1f84456dac659fc31ef7af8bbfad711e0876a/impl/src/main/java/com/sun/faces/renderkit/ClientSideStateHelper.java) from Mojarra can be a good inspiration.

melloware:
yeah i personally think it might be overkill for this feature but would like @tandraschko opinion.  We could probably take some code like `StateUtils` from MyFaces and implement it.

tandraschko:
What about a rembering and validate generated nonce in session scoped?

melloware:
but for stateless views there is no Session right?

tandraschko:
Depends but not necessarly

PR #9493, opened 2022-12-12
Fix #9485: ContentLength to long from int

### LINKED ISSUE #9485: DefaultStreamedContent.contentLength is Integer
I think Long would be better for bigger files.

melloware:
Fix #9485: ContentLength to long from int


tandraschko:
Can you uncomment all old externalContext calls and add a comment?

melloware:
I will update and add comments

melloware:
Old code commented out with comment above explaining why in both places.

tandraschko:
+1

PR #9508, opened 2022-12-13
Fix #9486: DatePicker year/month navigator ARIA labels

### LINKED ISSUE #9486: DatePicker: Year and month navigator don't have a label
### Describe the bug

Using year and monthnavigator = true for a datepicker element creates select items in the datepicker. These selects don't get a label for accessibility. 

### Reproducer

It can be reproduced using any Datepicker with 
`yearNavigator="true" monthNavigator="true"`

### Expected behavior

Year and Month select items should have an associated label.

### PrimeFaces edition

None

### PrimeFaces version

12.0.2

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.1

### Java version

11.0.14

### Browser(s)

_No response_

melloware:
Fix #9486: DatePicker year/month navigator ARIA labels

image

@jepsar @Rapster @tandraschko if you send me your language translation for the word "Select" like "Select Year" I can add it to the locales of this PR.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
@@ -72,7 +72,12 @@
                 monthNames: ["January", "February", "March", "April", "May", "June", "July", "August", "September", "October", "November", "December"],
                 monthNamesShort: ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"],
                 today: 'Today',
-                clear: 'Clear'
+                clear: 'Clear',
+                select: 'Select',
+                year: 'Year',
+                month: 'Month',
+                week: 'Week',
+                day: 'Day'
             },
             dateFormat: 'mm/dd/yy',
             yearRange: null,
@@ -1344,7 +1349,7 @@
 
         renderTitleMonthElement: function (month, index) {
             if (this.options.monthNavigator && this.options.view !== 'month' && index === 0) {
-                return '<select class="ui-datepicker-month">' + this.renderTitleOptions('month', this.options.locale.monthNamesShort, month) + '</select>';
+                return '<select class="ui-datepicker-month" aria-label="'+this.options.locale.select+' '+this.options.locale.month+'">' + this.renderTitleOptions('month', this.options.locale.monthNamesShort, month) + '</select>';
             }
             else {
                 return '<span class="ui-datepicker-month">' + this.escapeHTML(this.options.locale.monthNames[month]) + '</span>' + '&#xa0;';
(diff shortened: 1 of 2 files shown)


Rapster:
FR: `Sélectionner le mois` and `Sélectionner l'année`

jepsar:
nl
```
                select: 'Selecteer',
                year: 'Jaar',
                month: 'Maand',
```
Although labels should be like "Selecteer jaar", "Selecteer maand". So no capitals on the nouns.

Actually, I think we don't need to add "Select" at all. Simply "Jaar" or "Maand" will do.

primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -1344,7 +1349,7 @@
 
         renderTitleMonthElement: function (month, index) {
             if (this.options.monthNavigator && this.options.view !== 'month' && index === 0) {
-                return '<select class="ui-datepicker-month">' + this.renderTitleOptions('month', this.options.locale.monthNamesShort, month) + '</select>';
+                return '<select class="ui-datepicker-month" aria-label="'+this.options.locale.select+' '+this.options.locale.month+'">' + this.renderTitleOptions('month', this.options.locale.monthNamesShort, month) + '</select>';
```
jepsar: Why not simply do this?
```suggestion
                return '<select class="ui-datepicker-month" aria-label="'+this.options.locale.month+'">' + this.renderTitleOptions('month', this.options.locale.monthNamesShort, month) + '</select>';
```

primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -1363,7 +1368,7 @@
                     yearOptions.push(i);
                 }
 
-                return '<select class="ui-datepicker-year">' + this.renderTitleOptions('year', yearOptions, year) + '</select>';
+                return '<select class="ui-datepicker-year" aria-label="'+this.options.locale.select+' '+this.options.locale.year+'">' + this.renderTitleOptions('year', yearOptions, year) + '</select>';
```
jepsar: Why not simply do this?
```suggestion
                return '<select class="ui-datepicker-year" aria-label="'+this.options.locale.year+'">' + this.renderTitleOptions('year', yearOptions, year) + '</select>';
```

primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
melloware: You think its OK to just read out "Month" and not "Select Month" ?  I would be OK with that?

jepsar: Sure. For email input my label is "Email" and not "Input email". Same thing.

melloware:
I like this better @jepsar you are right less new parts and it gets the point across.

PR #9519, opened 2022-12-17
Fix #9518: JSF23 wrapped constructors

### LINKED ISSUE #9518: FacesWrapper implementations should push the wrapped instance to the constructor
### Describe the bug

Since JSF 2.3 the default constructor of `FacesWrapper` subclasses has been deprecated in order to force implementors to instead use the constructor taking the wrapped instance (and to raise their awareness), so that logically the inherited `getWrapped()` method will be used throughout the implementation instead of the local `wrapped` variable. This will ensure that the correct implementation is returned and correct behavior is performed might the `FacesWrapper` implementation itself being wrapped by yet another `FacesWrapper` implementation further down the chain. Because, when the `FacesWrapper` implementation incorrectly/accidentally uses the local `wrapped` variable instead of the `getWrapped()` method, then that other `FacesWrapper` implementation will basically be completely ignored, hereby breaking the decorator pattern.

For example, the `PrimeExceptionHandler` of 12.0.0 is implemented as follows:

```
    private final ExceptionHandler wrapped;
    private final Lazy<PrimeConfiguration> config;

    @SuppressWarnings("deprecation") // the default constructor is deprecated in JSF 2.3
    public PrimeExceptionHandler(ExceptionHandler wrapped) {
        this.wrapped = wrapped;
        this.config = new Lazy(() -> PrimeApplicationContext.getCurrentInstance(FacesContext.getCurrentInstance()).getConfig());
    }

    @Override
    public ExceptionHandler getWrapped() {
        return wrapped;
    }
```

This is not entirely correct. It should have been implemented as follows:

```
    private final Lazy<PrimeConfiguration> config;

    public PrimeExceptionHandler(ExceptionHandler wrapped) {
        super(wrapped);
        this.config = new Lazy(() -> PrimeApplicationContext.getCurrentInstance(FacesContext.getCurrentInstance()).getConfig());
    }
```

And wherever the local `wrapped` variable was referenced within the same class, it has to be replaced by `getWrapped()`.

And all other `FacesWrapper` implementations throughout the PrimeFaces library, if any, should follow the same pattern.

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

12.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0

### Java version

17

### Browser(s)

_No response_

melloware:
Fix #9518: JSF23 wrapped constructors

The note `// #6212 - don't remove it` is in reference to this old Google Code ticket: https://code.google.com/archive/p/primefaces/issues/6212

From 2013: "GlassFish 4/Weld logs a warning about it" so I think its safe to remove now in 2023.

### DIFF
--- a/primefaces-integration-tests/src/main/webapp/WEB-INF/web.xml
+++ b/primefaces-integration-tests/src/main/webapp/WEB-INF/web.xml
@@ -27,10 +27,6 @@
         <param-value>${primefaces.THEME}</param-value>
     </context-param>
     <context-param>
-        <param-name>primefaces.FONT_AWESOME</param-name>
-        <param-value>true</param-value>
-    </context-param>
-    <context-param>
         <param-name>primefaces.MOVE_SCRIPTS_TO_BOTTOM</param-name>
         <param-value>true</param-value>
     </context-param>
--- a/primefaces/src/main/java/org/primefaces/application/DialogViewHandler.java
+++ b/primefaces/src/main/java/org/primefaces/application/DialogViewHandler.java
@@ -24,23 +24,17 @@
 package org.primefaces.application;
 
 import java.util.Map;
+
 import javax.faces.application.ViewHandler;
 import javax.faces.application.ViewHandlerWrapper;
 import javax.faces.context.FacesContext;
+
 import org.primefaces.util.Constants;
 
 public class DialogViewHandler extends ViewHandlerWrapper {
 
-    private ViewHandler wrapped;
-
-    @SuppressWarnings("deprecation") // the default constructor is deprecated in JSF 2.3
     public DialogViewHandler(ViewHandler wrapped) {
-        this.wrapped = wrapped;
-    }
-
-    @Override
-    public ViewHandler getWrapped() {
-        return this.wrapped;
+        super(wrapped);
     }
 
     @Override
--- a/primefaces/src/main/java/org/primefaces/application/exceptionhandler/PrimeExceptionHandler.java
+++ b/primefaces/src/main/java/org/primefaces/application/exceptionhandler/PrimeExceptionHandler.java
@@ -33,6 +33,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+
 import javax.el.ELException;
 import javax.faces.FacesException;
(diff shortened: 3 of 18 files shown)


melloware:
Build is failing I need to investigate.

melloware:
OK @christophs78 this is weird.  When running locally with MyFaces the test that are failing are working for me so it seems like its a timing issue?

melloware:
The good news is the EXACT same tests are failing on all instances of Myfaces so its not random timeouts its specific to these use cases.

christophs78:
If it´s realy a MyFaces issue as @BalusC suggests with https://github.com/primefaces/primefaces/issues/9518#issuecomment-1356797344 maybe it can be fixed on the MyFaces - side by @tandraschko. (This should allow @BalusC to remove some of his workarrounds within https://github.com/omnifaces/omnifaces/blob/4.x/src/main/java/org/omnifaces/context/OmniPartialViewContext.java at some point in the future.)

But it´s still weird why we can´t reproduce it on our local dev-machines. (Same on my side.) I´m not convinced (yet?) calling it a Github actions timing issue. But what else is the cruical difference between our our dev-machines and Github Actions?

melloware:
Yeah weird. I am going to submit a MyFaces patch tomorrow and do some experimenting putting the PF fix back. Will let you know what I come up with.

melloware:
I am going to incrementally do one class at a time to narrow down the problem. It seems PartialResponseWriter was not the issue.

BalusC:
MyFaces has indeed fixed their PartialResponseWriter in the meanwhile: https://github.com/apache/myfaces/commit/9faf5888a6522fa69b0b6cac557b6fb4e0194720 It'll be another class then.

melloware:
@christophs78 I added 1 IT test nightly to run in JSF2.2 and its failing miserably with class not founds.  I wonder if you ran into this before.  See the run: https://github.com/primefaces/primefaces/actions/runs/3746873838/jobs/6362572220

tandraschko:
i thought CDI compatible ViewScoped was added in 2.2?

melloware:
me too I will try and run a single test with 2.2 locally and see what happens

PR #9534, opened 2022-12-20
Fix #9532: SelectOneRadio trigger change on unselect

### LINKED ISSUE #9532: SelectOneRadio: Not firing 'change' when unselectable="true" and unselect
### Describe the bug

SelectOneRadio can be enhanced with the option 'unselectable="true"' which enables the deselection of a selected option. Currently, however, no 'change' is fired when deselecting. 

However, this works with the SelectOneButton. 
See https://github.com/primefaces/primefaces/issues/2483 for the corresponding issue.

### Reproducer

Reproducer: https://github.com/dstucki/primefaces-test

1. Select Option '1'
2. Deselect Option '1'

### Expected behavior

It should print 'Value from bean: empty' on the page. Instead it prints 'Value from bean: 1'

### PrimeFaces edition

Community

### PrimeFaces version

12.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

see reproducer

### Java version

1.8

### Browser(s)

_No response_

melloware:
Fix #9532: SelectOneRadio trigger change on unselect


melloware:
@christophs78 Integration Test added for this scenario

PR #9539, opened 2022-12-21
Fix #9538: Inplace always mask passwords

### LINKED ISSUE #9538: Inplace: does not consider masked text of Password
### Description

Currently, the `p:inplace` does not consider to leave the text from an embeeded `p:password` masked (if enabled).

```
<p:inplace>
	<p:password value="#{user.password}" match="pw2" toggleMask="true" redisplay="true" />
</p:inplace>
```

The password is shown by `p:inplace` unmasked even if the `p:password` masks it.

If the user has something like this:

```
<p:inplace>
	<p:password value="#{user.password}" match="pw2" toggleMask="true" redisplay="true" />
</p:inplace>
```

Then `p:inplace` should leave the password text masked (in input and output mode). Only when the user toggles the mask by `p:password`, then the password should be unmasked.

### Additional context

_No response_

melloware:
Fix #9538: Inplace always mask passwords


primefaces/src/main/java/org/primefaces/component/inplace/InplaceRenderer.java
```diff
@@ -131,7 +132,12 @@ protected String getLabelToRender(FacesContext context, Inplace inplace) {
             return label;
         }
 
-        String value = ComponentUtils.getValueToRender(context, inplace.getChildren().get(0));
+        UIComponent editor = inplace.getChildren().get(0);
+        String value = ComponentUtils.getValueToRender(context, editor);
+        if (LangUtils.isNotBlank(value) && (editor instanceof Password || "password".equalsIgnoreCase(String.valueOf(editor.getAttributes().get("type"))))) {
+            value = "********";
```
jepsar: Maybe we could add a style class to the component instead and not change the value. With this class you can use CSS to add content (bullets by default I'd say).

melloware: I could see that just want to make sure the unmasked password is not in the HTML.

melloware: Maybe a `p-masked` style added to it.

jepsar: Sorry, of course we don't want to render the value in case of a password. We could do that by setting the value to null or blank, but that's a bit of a hack if you ask me.

PR #9540, opened 2022-12-21
fix #9531: InputNumber trigger change when empty and then input 0

### LINKED ISSUE #9531: InputNumber: Ajax-Change-Event not triggered
### Describe the bug

If the InputNumber is empty and you type 0 in the field, then no ajax-change-event is fired

```
<ui:composition template="/WEB-INF/template.xhtml"

    <f:metadata>
        <f:viewAction action="#{indexController.init()}"/>
    </f:metadata>

    <ui:define name="content">

        <h:form id="form">
            <p:inputNumber id="inputnumber" value="#{indexController.value}">
                <p:ajax process="@this" partialSubmit="true" update="currentValue"/>
            </p:inputNumber>

            <br/>
            Current Value:<br/>
            <h:outputText id="currentValue" value="#{not empty indexController.value ? indexController.value : 'empty'}"/>
        </h:form>

    </ui:define>

</ui:composition>
```

```
@Named
@ViewScoped
public class IndexController implements Serializable
{
    private Double value;

    public void init() {

    }

    public Double getValue()
    {
        return value;
    }

    public void setValue(Double value)
    {
        this.value = value;
    }
}
```

### Reproducer

type 0 and click anywhere
the output label shows empty instead of 0.0

type 5 and click anywhere
the output label shows 5.0

type 0 and click anywhere
label shows 0.0 (correct)

remove the value and click anywhere
label shows empty

if you type 0 again and click anywhere
the output label shows empty instead of 0.0

### Expected behavior

If the input number field is empty and you type 0 then the ajax-change-event is triggered.

### PrimeFaces edition

Community

### PrimeFaces version

13.0.0-SNAPSHOT

### Theme

_No response_

### JSF implementation

MyFaces

### JSF version

2.3

### Java version

11+

### Browser(s)

_No response_

stonewoodman:
@melloware
I could now fix this issue #9531 

Fix #9531

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js
@@ -158,7 +158,7 @@ PrimeFaces.widget.InputNumber = PrimeFaces.widget.BaseWidget.extend({
                 element.set(newValue, null, true);
 
                 // GitHub #8610: reset the raw values so we don't fire change event if 1.0 == 1.00
-                if (Number(element.rawValue) === Number(element.rawValueOnFocus)) {
+                if (element.rawValueOnFocus !== '' && Number(element.rawValue) === Number(element.rawValueOnFocus)) {
                     element.rawValueOnFocus = element.rawValue;
                 }
             }


melloware (approved):
Nice work!  I think when I was stepping through the debugger i was doing it later in the sequence and missed this.

PR #9548, opened 2022-12-24
Fix #9538 Inplace: does not consider masked text of Password

### LINKED ISSUE #9538: Inplace: does not consider masked text of Password

jepsar:
Fix #9538

@melloware this was the idea I had

### DIFF
--- a/primefaces-showcase/src/main/java/org/primefaces/showcase/view/input/InplaceView.java
+++ b/primefaces-showcase/src/main/java/org/primefaces/showcase/view/input/InplaceView.java
@@ -32,6 +32,7 @@ import java.io.Serializable;
 public class InplaceView implements Serializable {
 
     private String text = "PrimeFaces";
+    private String password = "secret";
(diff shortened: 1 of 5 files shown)


melloware (approved):
I like this so someone can then override the style

PR #9558, opened 2022-12-27
JSF 22 integration tests

melloware:
Have to exclude DatePicker013 becuase its using f:converter with LocalDateTime which does not exist in 2.2


melloware:
cc @christophs78

Issue #9581, opened 2023-01-03
QRCode: NPE


tuerker:
### Describe the bug

We have occasionally entries of the following form in our log:

```
Servlet.service() for servlet Faces Servlet threw exception
java.lang.NullPointerException
        at org.primefaces.application.resource.QRCodeHandler.handle(QRCodeHandler.java:50)
        at org.primefaces.application.resource.PrimeResourceHandler.handleResourceRequest(PrimeResourceHandler.java:98)
        at javax.faces.application.ResourceHandlerWrapper.handleResourceRequest(ResourceHandlerWrapper.java:171)
        at javax.faces.application.ResourceHandlerWrapper.handleResourceRequest(ResourceHandlerWrapper.java:171)
        at javax.faces.application.ResourceHandlerWrapper.handleResourceRequest(ResourceHandlerWrapper.java:171)
        at javax.faces.application.ResourceHandlerWrapper.handleResourceRequest(ResourceHandlerWrapper.java:171)
        at javax.faces.application.ResourceHandlerWrapper.handleResourceRequest(ResourceHandlerWrapper.java:171)
        at javax.faces.webapp.FacesServlet.executeLifecyle(FacesServlet.java:704)
        at javax.faces.webapp.FacesServlet.service(FacesServlet.java:451)
        at org.apache.catalina.core.StandardWrapper.service(StandardWrapper.java:1636)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:331)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:211)
        at org.glassfish.tyrus.servlet.TyrusServletFilter.doFilter(TyrusServletFilter.java:282)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:253)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:211)
        at org.bfabric.filter.LoginFilter.doFilter(LoginFilter.java:61)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:253)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:211)
        at org.apache.catalina.core.StandardWrapperValve.invoke(StandardWrapperValve.java:257)
[...]

PR #9582, opened 2023-01-03
QRCode: NPE #9581

tuerker:


tandraschko:
IMO we should just avoid the NPE and skip streaming as the renderer is the one who should only initialize it

tandraschko:
did a small refactoring and applied a null check
please see my commit

tuerker:
fine. thanks.

PR #9583, opened 2023-01-03
Fix #9517: TabView add focusOnLastActiveTab

### LINKED ISSUE #9517: TabView:  add restoreTabIndex to restore last active tab index
### Description

Would be nice to include a property `tabView.restoreTabIndex` which restores the active tab index when `tabView` is rerendered. Actually, users must bind a property and add a listener (with a form enclosed to tabView) (on server side) which is very roundabout. Besides, it is not always appropriate to have a form around the `tabView` when there are also forms within the tabs itself (nested forms are bad..).

It would be better to restore the `tabIndex` on **client side** by storing the actual index internally with `onTabChange` and activating this tab (by the stored index) after the `tabView` reInitialized. 

`<p:tabView restoreTabIndex="true" ..>`

### Additional context

_No response_

nimo23:
Fix https://github.com/primefaces/primefaces/issues/9517


melloware (approved):
LGTM!

PR #9589, opened 2023-01-05
Fix #9588: BlockUI respect padding

### LINKED ISSUE #9588: BlockUI: incorrect overlay position for block with padding
### Describe the bug

When a blocked element uses padding, the overlay is displayed incorrectly. The size is correct but the position should be shifted.
The behavior is correct when the padding is 0.

### Reproducer

The issue can be reproduced with the showcase of PrimeFaces 12:

1. Open the blockUI showcase (PrimeFaces 12): https://www.primefaces.org/showcase/ui/misc/blockUI.xhtml
2. Open developer tools to update the CSS (add padding on the blocked elements)
3. Example CSS for Basic: `body .ui-panel { padding: 20px; }`
4. Example CSS for Custom Content: `div.ui-datatable { padding: 20px; }`
5. Check if the Basic overlay covers the panel (Save button).
6. Actual result: Basic overlay size is OK but position should move (to bottom right)
7. Check if the Custom Content overlay covers the table (row navigation buttons)
8. Actual result: Custom Content overlay size is OK but position should move (to bottom right)

### Expected behavior

In both cases (Basic or Custom Content):

- The overlay should cover the blocked element (panel or table)
- The overlay position should take the padding into account (using outerWidth/Height rather than width/height)

### PrimeFaces edition

Community

### PrimeFaces version

12.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

11

### Browser(s)

Chrome 108.0.5359.125

melloware:
Fix #9588: BlockUI respect padding

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
@@ -276,8 +276,8 @@ PrimeFaces.widget.BlockUI = PrimeFaces.widget.BaseWidget.extend({
             }
 
             // set the size and position to match the target
-            var height = currentTarget.height(),
-                width = currentTarget.width(),
+            var height = currentTarget.outerHeight(),
+                width = currentTarget.outerWidth(),
                 offset = currentTarget.offset();
             var sizeAndPosition = {
                 'height': height + 'px',


melloware:
Checking to see if

```
One or more dependencies were identified with vulnerabilities that have a CVSS score greater than or equal to '5.0': 
Error:  
Error:  rome-utils-1.18.0.jar: CVE-2021-4277(5.3)
```

is a false positive

melloware:
Yep its a false positive and its been reported here: https://github.com/jeremylong/DependencyCheck/issues/5213

PR #9597, opened 2023-01-06
Fix #9594: DataTable: JpaLazyDataModel.convertToType throws FacesException for java.util.Date

### LINKED ISSUE #9594: DataTable: JpaLazyDataModel.convertToType throws FacesException for java.util.Date
### Describe the bug

When using `p:datePicker` in a column filter facet of a `p:dataTable` backed by a `JpaLazyDataModel`, filtering fails with a `FacesException` thrown by `JpaLazyDataModel` stating:

```
SEVERE: javax.faces.FacesException: Can not convert 2023-01-04 to type class java.util.Date; please create a JSF Converter for it or overwrite Object convertToType(String value, Class<?> valueType)!
        at org.primefaces.model.JpaLazyDataModel.convertToType(JpaLazyDataModel.java:337)
        at org.primefaces.model.JpaLazyDataModel.applyFilters(JpaLazyDataModel.java:163)
        at org.primefaces.model.JpaLazyDataModel.count(JpaLazyDataModel.java:120)
```

### Reproducer

The existing primefaces-test jpa branch shows this: https://github.com/primefaces/primefaces-test/tree/jpa

After starting the app, filter by the Date Col and you will see the exception in the terminal.

### Expected behavior

Date filters should work with the `JpaLazyDataModel`.

`filter.getFilterValue()` should contain the correct object already and not need conversion at all. I may be missing the need for `convertToType`, but I'm wondering if it should just default to the original value if it can't convert it.

### PrimeFaces edition

Community

### PrimeFaces version

13.0.0

### Theme

All

### JSF implementation

All

### JSF version

2.3

### Java version

11

### Browser(s)

Chrome latest

FlipWarthog:


melloware (approved):

tandraschko:
need to debug it, cant remember why i didnt use the filterValue object

melloware:
@tandraschko we hit this issue at one our client applications so let us know how to proceed when you get a chance to look at this.

tandraschko:
can you share me a debugger screenshot with all filters and filterValue from #applyFilters?
i thought in theory filters should always be string value but not sure

melloware:
Here is a runnable reproducer you can run and debug if need be.  Just pick the Date Column Filter.
[pf-9594.zip](https://github.com/primefaces/primefaces/files/10375642/pf-9594.zip)

FlipWarthog:
This is a screenshot of my debugger showing the `filterValue` as a `LocalDate`, as well as the converted `toString` value.
image

tandraschko:
i did a small refactoring and improved the whole handling
BUT
your case with the date column is still bit...
the field in your entity is Date but the filterValue is LocalDate

whould should happen? does JPA/Hibernate support this if i pass LocalDate to a Date field?

tandraschko:
please see my 2 commits and review

FlipWarthog:
@tandraschko I checked out your changes and it makes sense to me. I realized my debugger screenshot was after I'd reverted some changes in `primefaces-test`. Specifically, `p:datePicker` defaults to `LocalDate` _and_ the default `filterMatchMode` is "startsWith", so it was passing the date through then converting it to a String of "2023-01-10%" and doing a LIKE query.

I went back to what I had, binding the value of the `p:datePicker` to a `java.util.Date` field in the view:
```xml
<p:column field="dateCol" filterMatchMode="equals">
    <f:facet name="filter">
        <p:datePicker value="#{testView.filterDate}" onchange="PF('jpaTable').filter()" />
    </f:facet>
</p:column>
```

Now it's properly getting into this condition:
```java
// its already the same type
if (valueType.isAssignableFrom(value.getClass())) {
    return value;
}
```
And doing an EQUALS as expected.

NOTE: If I don't bind the `p:datePicker` (thus making the filter value be `LocalDate`) but _do_ switch the filterMatchMode to "equals", then Hibernate blows up on mismatched types (which is what I would expect).

So that was a long-winded way of saying, yup all is good for this issue!

PR #9603, opened 2023-01-06
Fix #9598: DataTable: Implement BETWEEN/NOT_BETWEEN MatchMode for JpaLazyDataModel

### LINKED ISSUE #9598: DataTable: Implement BETWEEN/NOT_BETWEEN MatchMode for JpaLazyDataModel
### Description

`JpaLazyDataModel` does not currently support `filterMatchMode="between"` or `filterMatchMode="notBetween"`

`filterMatchMode="between"` and `filterMatchMode="notBetween"` to be supported when using `JpaLazyDataModel`.

### Additional context

Follow up to implementing IN/NOT_IN: https://github.com/primefaces/primefaces/issues/9324

FlipWarthog:


primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
```diff
@@ -235,9 +236,13 @@ protected Predicate createPredicate(FilterMeta filter, Field filterField,
                         ? cb.notEqual(fieldExpression, filterValueAsCollection.get().iterator().next())
                         : fieldExpression.in(filterValueAsCollection.get()).not();
             case BETWEEN:
-                throw new UnsupportedOperationException("MatchMode.BETWEEN currently not supported!");
+                Iterator<Object> iter1 = filterValueAsCollection.get().iterator();
+                return cb.and(cb.greaterThanOrEqualTo(fieldExpression, (Comparable) iter1.next()),
+                    cb.lessThanOrEqualTo(fieldExpression, (Comparable) iter1.next()));
```
FlipWarthog: Inclusive on both sides based on the `isBetween` impl in `BetweenFilterConstraint`:
```java
    protected boolean isBetween(Comparable value, Comparable start, Comparable end) {
        return value.compareTo(start) >= 0 && value.compareTo(end) <= 0;
    }
```

primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
```diff
@@ -235,9 +236,13 @@ protected Predicate createPredicate(FilterMeta filter, Field filterField,
                         ? cb.notEqual(fieldExpression, filterValueAsCollection.get().iterator().next())
                         : fieldExpression.in(filterValueAsCollection.get()).not();
             case BETWEEN:
-                throw new UnsupportedOperationException("MatchMode.BETWEEN currently not supported!");
+                Iterator<Object> iter1 = filterValueAsCollection.get().iterator();
```
melloware: Minor but I like iterBetween and iterIn rather than 1amd 2

melloware (changes requested):

primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
FlipWarthog: Addressed this.

melloware (approved):

PR #9613, opened 2023-01-08
https://github.com/primefaces/primefaces/issues/9612 - CurrencyValida…

### LINKED ISSUE #9612: CurrencyValidator: restore validate(String value, Locale locale) method
### Description

Prior to PF12, in CurrencyValidator, we had 

https://github.com/primefaces/primefaces/blob/50a7ea981a449330944ceb2caceec0eb1eccae98/primefaces/src/main/java/org/primefaces/util/CurrencyValidator.java#L68-L78

https://github.com/primefaces/primefaces/blob/50a7ea981a449330944ceb2caceec0eb1eccae98/primefaces/src/main/java/org/primefaces/util/CurrencyValidator.java#L235-L252

https://github.com/primefaces/primefaces/blob/50a7ea981a449330944ceb2caceec0eb1eccae98/primefaces/src/main/java/org/primefaces/util/CurrencyValidator.java#L207-L217

It was changed by DecimalFormat and moved to BigDecimalValidator and all feature related to java.util.Locale was removed.

https://github.com/primefaces/primefaces/commit/a3817611579b3a7066be6e18883fbfc69cd8f5aa#diff-85ea31170393fe54c04fd1e167bd4c291b81f8009d8fd7ceed38b058dd87fc5bL67-L78

However, this is public method and external applications could rely on the public api and java.util.locale and decimal are not the same. In the migration plan, it wasn't mentioned as breaking changes but it should if we need to remove them (and provide alternative if there's any).

Previously, https://github.com/primefaces/primefaces/issues/9610 is restored, and this ticket is to restore the validate method with Locale as well as its depent methods.

BTW, this ticket shouldn't be marked as `feature` but that's the default setup in PF repo (either defect or feature), so please change the label if appropriate such as enhancement like #9610 .

### Additional context

_No response_

jxmai:
…tor: restore validate(String value, Locale locale) method

Resolve: https://github.com/primefaces/primefaces/issues/9612

### DIFF
--- a/primefaces/src/main/java/org/primefaces/util/CurrencyValidator.java
+++ b/primefaces/src/main/java/org/primefaces/util/CurrencyValidator.java
@@ -24,6 +24,7 @@
 package org.primefaces.util;
 
 import java.text.DecimalFormat;
+import java.math.BigDecimal;
 import java.util.Locale;
 
 /**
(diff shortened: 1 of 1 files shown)


melloware (approved):

melloware:
I will merge after the build passes. Thanks for the PR!

jxmai:
Thanks! I have built locally successfully. Let's see how the web hook result is :crossed_fingers:

jxmai:
all tests have passed:)

melloware:
Thanks!  Merges!

Issue #9629, opened 2023-01-10
SelectOneMenu: filter does not work for accented characters


kurkix:
### Describe the bug

Problem is turkish "i" character  when search in items. selectOneMenu does not filtered with lowercase characters. but problem is on "i" character only. not on "ü", "ş" etc. other turkish characters work. if word is uppercase "DESTİNATİON", and if i search "DESTİ", its okay. but if i  search "desti", there is no result in items.  same on reserve scenerio. if word is lowercase like "destination"  and try to search "DESTİ". it does not work. this problem is at selectOneMenu only. example no problem at dataTable.

[problem is here](https://ibb.co/dtrTgwD)

### Reproducer

[primefaces-test.zip](https://github.com/primefaces/primefaces/files/10385219/primefaces-test.zip)

open selectOneMenu and search "edirn"

### PrimeFaces edition

Community

### PrimeFaces version

12.0.0

### Theme

Serenity

### JSF implementation

Mojarra

### JSF version

2.2.11

### Java version

1.8

jepsar:
We could add an option to normalize strings.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize

Note: there are more components which offer filtering.

jepsar:
.. but maybe this is too exotic, and a (custom) `filterFunction` should be created by users?

melloware:
Originally asked on Stack Overflow 7 years ago too so this issue has been out there a while: https://stackoverflow.com/questions/29371104/primefaces-filter-not-working-with-%C4%B1-and-i-character

we should try and solve it.

jepsar:
```
> 'DESTİ'.normalize('NFD').replace(/\p{Diacritic}/gu, '').toLowerCase();
<- 'desti'
```
https://stackoverflow.com/questions/990904/remove-accents-diacritics-in-a-string-in-javascript

melloware:
MonkeyPatch thanks to @jepsar !  Just add this code to your project

```js
if (PrimeFaces.widget.SelectOneMenu) {

    PrimeFaces.widget.SelectOneMenu.prototype.normalize = function(string, lowercase) {
        if (!string) return string;
        var result = string.normalize('NFD').replace(/[\u0300-\u036f]/g, '');
        return lowercase ? result.toLowerCase() : result;
    }

    PrimeFaces.widget.SelectOneMenu.prototype.filter = function(value) {
[...]

PR #9651, opened 2023-01-12
#4696 Column support for SelectCheckboxMenu

### LINKED ISSUE #4696: SelectCheckboxMenu: Add ability to provide custom rendering of items in the dropdown menu
## 1) Environment
- PrimeFaces version: 6.2
- Application server + version: Tomcat 8

## 2) Expected behavior
`selectOneMenu` supports the ability to provide custom rendering of the items in the dropdown list, see "Advanced" example at: https://www.primefaces.org/showcase/ui/input/oneMenu.xhtml
I would expect / it would be nice to have the same ability with `selectCheckboxMenu`.

## 3) Actual behavior
`selectCheckboxMenu` does not allow to change the displaying of elements in the drop down menu.

## 4) Steps to reproduce
Add a `<p:selectCheckboxMenu>` to your facelet, like below.

## 5) Sample XHTML
```xml
<p:selectCheckboxMenu
  value="#{bean.myStrings}" multiple="true" filter="true"
  filterMatchMode="contains" dynamic="true">
    <f:selectItems value="#{bean.allStrings}" var="string"
      itemLabel="#{string.toUpperCase()}" />
    <p:column>
      <h:outputText value="Custom rendering: #{string}" />
    </p:column>
</p:selectCheckboxMenu>
```

## 6) Sample bean
```java
public class MyBean {
  private List<String> myStrings; // getter/setter omitted
  public List<String> getAllStrings() { 
    return Arrays.asList("one", "two", "three");
  }
}
```

### LINKED ISSUE #7406: SelectCheckboxMenu: Option to have multiple columns
**Description**

selectCheckboxMenu supports just single column, would be nice to have multiple columns like selectOneMenu 

**Sample**
```xml
<p:selectCheckboxMenu ... var="v">
...
                <p:column >
                  <h:outputText value="#{v.title1}" />
                </p:column>
                <p:column>
                  <h:outputText value="#{v.title2}" />
                </p:column>          
               
</p:selectCheckboxMenu>   
```

ltlBeBoy:
Fix #4696: I used `SelectOneMenu` as template to add column support for `SelectCheckboxMenu`. A demo is added to the showcase which almost works. But there is a problem: The first column is rendered outside of the checkbox menu. Can you please help me to fix this?

Fix #7406


primefaces/src/main/resources/META-INF/primefaces-p.taglib.xml
```diff
@@ -22889,6 +22889,14 @@
             <required>false</required>
             <type>java.lang.String</type>
         </attribute>
+        <attribute>
+            <description>
+                <![CDATA[]]>
```
melloware: We should add a description here is possible as IntelliJ apparently will ignore fields without a description.

melloware (changes requested):

jepsar (changes requested):
Please add documentation and a what's new entry.

melloware:
I will try and take a look and see why its creating the first column outside the menu

ltlBeBoy:
@jepsar Yes, of course. Tomorrow.

melloware:
@ltlBeBoy please add the following at the bottom of the renderer.   This tells it not to render its children which is why it was rendering that column outside of the context.

```java
@Override
public void encodeChildren(FacesContext facesContext, UIComponent component) throws IOException {
    //Rendering happens on encodeEnd
}

@Override
    return true;
}
```

melloware:
Also check your JS DOC errors... `mvn clean package -Pjsdoc` in the primfaces dir.

```
Type SelectCheckboxMenu declares property <labels>, but no such property is used in the code. Remove the doc comment, or use @forcedProp instead of @prop if you really want to include it.
```

ltlBeBoy:
OK, a first version is ready:

1. Added support for custom content in overlay panel.
2. Improved filtering according to #9629.
3. Added two demos to the showcase; one with "raw" entities and one with grouped `SelectItem`s.
4. Improved `org.primefaces.showcase.service.CountryService` to set reproducible IDs for `org.primefaces.showcase.domain.Country` (required for `org.primefaces.showcase.convert.CountryConverter`)
5. Extended documentation.
6. Entry for "whatsnew.md".

In my project I also added the following CSS to the theme (accordingly to `SelectOneMenu`). I guess this should be done by PrimeTek too:
```
/*
 * Columns for SelectCheckboxMenu
 */
body {
  .ui-selectcheckboxmenu-panel {
    .ui-selectcheckboxmenu-items {
      &.ui-selectcheckboxmenu-table {
        > thead tr th {
          background: $panelHeaderBg;
          border: $datatableHeaderCellBorder;
          color: $panelHeaderTextColor;
          font-weight: $panelHeaderFontWeight;
        }
      }

      .ui-selectcheckboxmenu-item {
        background: $datatableBodyRowBg;
        color: $datatableBodyRowTextColor;
        border: $datatableBodyRowBorder;

        > td {
          padding: $datatableBodyCellPadding;
        }

        &:first-child {
          border-top-color: transparent;
        }
      }
    }
  }
}
```

A few things that I noticed during the modifications (because I used `SelectOneMenu` as template):

1. `SelectOneMenu` and `SelectCheckboxMenu` now both support dynamic rendering and custom content in overlay panel. But there is one problem: These functions can not be used simultaneously because the custom content is always rendered by JSF! Shouldn't there be an exception or at least a logged warning that it's impossible to activate both at the same time?
2. `SelectOneMenu` fails to render `SelectItemGroup` children when custom content is used. Are you interested that I create another pull request?
3. `SelectOneMenu` fails to render custom content correctly when `omnifaces.GenericEnumConverter` is used. Are you interested that I create another pull request?
4. Is there a use case where a `SelectItem` with `noSelectionOption="true"` makes sense for `SelectCheckboxMenu`?
5. `SelectOneMenu` also supports RTL while `SelectCheckboxMenu` does not. Is that a requirement?
6. `SelectOneMenu` and `SelectCheckboxMenu` both have a variable to specified the height of the overlay panel. But at `SelectOneMenu` it's a `String` which is called `height`. And at `SelectCheckboxMenu` it's an `Integer` which is called `scrollHeight`...
7. `SelectOneMenu` and `SelectCheckboxMenu` both use a variable named `panelId` on the client side. But at `SelectOneMenu` it's the escaped client ID. And at `SelectCheckboxMenu` it's the "raw" client ID...

melloware:
Responses...

0. Please open another ticket with the CSS changes you need PrimeTek to make to the theme so I can mark is as "theme" and assign to PrimeTek
1. I don't understand this one and why thy both can't be used at the same time?  Maybe a ticket and a reproducer?
2. Open a ticket please. and then submit your PR against this ticket for review
3. Open a ticket please. and then submit your PR against this ticket for review
4. NO this use case makes no sense for CheckboxMenu I don't think
5. NO not a requirement just leave the TODO in there.  Not all components support RTL fully
6. It should be a `string` not an `integer` so it can take any CSS like "100vh" "100%" "100px" just make sure the code respects if its only a number it add "px" to it. I think you can find examples other places
7. It should probably be consistent?

ltlBeBoy:
@melloware 4. - 7. is done. For 1. - 3. I will open a ticket next week.

melloware (approved):

melloware:
This looks really good and a nice addition to SelectCheckboxMenu!

melloware:
@jepsar if you want to re-review before I merge that would be great.

primefaces/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
```diff
@@ -318,19 +333,343 @@ protected void encodeMenuIcon(FacesContext context, SelectCheckboxMenu menu, boo
...
+            if (selectItem.isNoSelectionOption() && !checked) {
+                return;
+            }
+
+            //attribute values
+            rowStyleClass = SelectCheckboxMenu.ROW_ITEM_CLASS;
+            rowStyleClass += checked ? " ui-selectcheckboxmenu-checked" : " ui-selectcheckboxmenu-unchecked";
+            if (selectItem.isNoSelectionOption()) {
+                rowStyleClass += " ui-noselection-option";
+            }
```
melloware, line 598: if (selectItem.isDisabled()) {
                rowStyleClass += " ui-state-disabled";
            }

melloware (changes requested):

melloware:
Based on this ticket: https://github.com/primefaces/primefaces/issues/9677 we need to disable rows if the selectitem is disabled.

primefaces/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenu.java
```diff
@@ -84,6 +96,19 @@ public String getDefaultEventName() {
         return DEFAULT_EVENT;
     }
 
+    public List<Column> getColumns() {
+        List<Column> columns = new ArrayList<>(5);
```
jepsar: Do we ever set initial capacity?

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
```diff
@@ -121,9 +122,12 @@ PrimeFaces.widget.SelectCheckboxMenu = PrimeFaces.widget.BaseWidget.extend({
         this.menuIcon = this.jq.children('.ui-selectcheckboxmenu-trigger');
         this.triggers = this.jq.find('.ui-selectcheckboxmenu-trigger, .ui-selectcheckboxmenu-label');
         this.disabled = this.jq.hasClass('ui-state-disabled');
-        this.inputs = this.jq.find(':checkbox');
-        this.panelId = this.id + '_panel';
+        //Only include checkboxes with an ID that starts like ID of the menu!
+        this.inputs = this.jq.find(':checkbox').filter('[id^="' + this.id.replace(/:/g,"\\:") + '"]');
```
jepsar: This is dangerous as the separator might be configured to a different character.

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
```diff
@@ -194,31 +197,39 @@ PrimeFaces.widget.SelectCheckboxMenu = PrimeFaces.widget.BaseWidget.extend({
...
+            this.itemContainer = this.itemContainerWrapper.children('.ui-selectcheckboxmenu-items');
+            this.items = this.itemContainer.find('.ui-selectcheckboxmenu-item');
+            this.groupHeaders = this.itemContainer.find('.ui-selectcheckboxmenu-item-group');
         }
-
-        this.renderItems();
 
         if (this.cfg.scrollHeight) {
-            this.itemContainerWrapper.height(this.cfg.scrollHeight);
+            this.itemContainerWrapper.css('max-height', this.cfg.scrollHeight);
```
jepsar, line 221: Are these height changes relevant to column support? Looks like a different issue to me.

primefaces/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
```diff
@@ -111,8 +127,7 @@ protected void encodeInputs(FacesContext context, SelectCheckboxMenu menu, List<
         writer.writeAttribute("class", "ui-helper-hidden", null);
 
         int idx = -1;
-        for (int i = 0; i < selectItems.size(); i++) {
-            SelectItem selectItem = selectItems.get(i);
+        for (SelectItem selectItem : selectItems) {
```
jepsar: https://github.com/primefaces/primefaces/wiki/Contributing-to-PrimeFaces#performance-guidelines

primefaces/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
```diff
@@ -318,19 +333,343 @@ protected void encodeMenuIcon(FacesContext context, SelectCheckboxMenu menu, boo
...
     }
 
+    protected void encodePanel(FacesContext context, SelectCheckboxMenu menu, List<SelectItem> selectItems) throws IOException {
+        ResponseWriter writer = context.getResponseWriter();
+        String panelStyle = menu.getPanelStyle();
+        String panelStyleClass = SelectCheckboxMenu.PANEL_CLASS;
+        if (menu.getPanelStyleClass() != null) {
+            panelStyleClass += " " + menu.getPanelStyleClass();
+        }
+        // TODO: "RTLAware"!
```
jepsar, line 346: There is still a TODO here

jepsar (changes requested):

primefaces/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
ltlBeBoy: Yes, I asked @melloware if RTL support is a requirement. He said no, but I should keep the "TODO" comment.

primefaces/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenu.java
melloware: No I think we only do it if we know the number of items specifically.

ltlBeBoy: @jepsar It's copy-paste from `SelectOneMenu`. But I will remove it.

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
ltlBeBoy: OK. Before it was `this.inputs = this.jq.find(':checkbox');`. But it's not exact enough because there is a toggler checkbox. Maybe I also try by class.

primefaces/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
melloware: Yep I have no idea what RTL should do in the case of a SelectCheckBoxMenu so I said leave it as a TODO.

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
ltlBeBoy: The problem is the code consistency. I used `SelectOneMenuRenderer` as template. There the height property is used to set the max height (when rendering custom content):

```
protected void encodePanel(FacesContext context, SelectOneMenu menu, List<SelectItem> selectItems) throws IOException {
        ...
        String height = null;
        try {
            height = Integer.parseInt(menu.getHeight()) + "px";
        }
        catch (NumberFormatException e) {
            height = menu.getHeight();
        }
        ...
        writer.startElement("div", null);
        writer.writeAttribute("class", SelectOneMenu.ITEMS_WRAPPER_CLASS, null);
        writer.writeAttribute("style", "max-height:" + height, null);
        ...
```

According to the documentation the similar property for `SelectCheckboxMenu` is `scrollHeight` which is also handed over to the client side in `SelectCheckboxMenuRenderer`:

```
protected void encodeScript(FacesContext context, SelectCheckboxMenu menu) throws IOException {
        wb.init("SelectCheckboxMenu", menu)
                .callback("onShow", "function()", menu.getOnShow())
                .callback("onHide", "function()", menu.getOnHide())
                .callback("onChange", "function()", menu.getOnchange())
                .attr("scrollHeight", menu.getScrollHeight(), Integer.MAX_VALUE) // now attr("scrollHeight", getMaxScrollHeight(menu), null)
```

But the filter function in `forms.selectcheckboxmenu.js` updates the CSS "height" attribute which would result in ignoring the property `scrollHeight`:

```
if(this.itemContainer.height() < this.cfg.initialHeight) {
    this.itemContainerWrapper.css('height', 'auto');
}
else {
    this.itemContainerWrapper.height(this.cfg.initialHeight);
}
```

This is why I choosed using CSS "max-height".

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
ltlBeBoy: It should be safe to search in the "hidden" `div` that contains the actual checkbox input elements:

`this.inputs = this.jq.filter('.ui-helper-hidden').find(':checkbox');`

What do you think?

jepsar: Better!

melloware (approved):

melloware:
Looks like we are getting Integration Test failures.

```
Error:  Failures: 
Error:    SelectCheckboxMenu001Test.testCheckAll:114->assertSelected:146 expected: <9> but was: <0>
Error:    SelectCheckboxMenu001Test.testEnabled:95->assertSelected:146 expected: <1> but was: <0>
Error:    SelectCheckboxMenu001Test.testSubmit:54->assertSelected:146 expected: <1> but was: <0>
Error:    SelectCheckboxMenu001Test.testUncheckAll:134->assertSelected:146 expected: <9> but was: <0>
Error:  Errors: 
Error:    DataTable002Test.testLazyFilterSelectCheckboxMenu:196 » IndexOutOfBounds Index...
Error:    DataTable002Test.testLazyFilterSelectCheckboxMenu:196 » IndexOutOfBounds Index...
Error:    DataTable039Test.testLazyFilterSelectCheckboxMenu:70 » IndexOutOfBounds Index:...
```

ltlBeBoy:
@melloware Yes, I fixed the problem already. Uploading soon.

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
ltlBeBoy: @jepsar FYI:
My proposed fix did not work and resulted in failed integration tests. Apparently because for the target element the CSS `display` is set to `none` initially which makes it invisible to the filter.
So I used this instead:
`this.jq.children('div.ui-helper-hidden:not(.ui-input-overlay)').children(':checkbox');`

melloware:
Interesting your last commit did not trigger a github actions build?

ltlBeBoy:
Maybe because there was a conflict with the master (the fix #9691).

ltlBeBoy:
@melloware Please update as stated here: https://github.com/primefaces/primefaces/issues/9691#issuecomment-1385653606
Then the conflict should be gone.

melloware:
Fixed.

docs/13_0_0/components/selectcheckboxmenu.md
```diff
@@ -90,6 +90,38 @@ function customFilter(itemLabel, filterValue) {
 }
 ```
 
+## Custom Content
+SelectCheckboxMenu can display custom content in overlay panel by using column component and the
+var option to refer to each item. Facets for column _header_ may also be used.
```
jepsar:
```suggestion
var option to refer to each item. Facets for column `header` may also be used.
```

primefaces/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
```diff
@@ -318,19 +335,348 @@ protected void encodeMenuIcon(FacesContext context, SelectCheckboxMenu menu, boo
...
+                    }
+                    else {
+                        //use initial values
+                        value = selectItem.getValue();
+                    }
+                    return !isSelected(context, menu, value, valuesArray, converter);
+                });
+
+        //toggler
+        encodeCheckbox(context, null, false, !notChecked, null, "Select All");
```
jepsar: Is this UI text? Should be loaded from a message bundle. https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/resources/org/primefaces/Messages.properties#L11

primefaces/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
```diff
@@ -318,19 +335,348 @@ protected void encodeMenuIcon(FacesContext context, SelectCheckboxMenu menu, boo
...
+        //filter
+        if (menu.isFilter()) {
+            encodeFilter(context, menu);
+        }
+
+        //closer
+        writer.startElement("a", null);
+        writer.writeAttribute("class", SelectCheckboxMenu.CLOSER_CLASS, null);
+        writer.writeAttribute("href", "#", null);
+        writer.writeAttribute(HTML.ARIA_LABEL, "Close", null);
```
jepsar: https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/resources/org/primefaces/Messages.properties#L19

primefaces/src/main/resources/META-INF/primefaces-p.taglib.xml
```diff
@@ -22851,11 +22851,11 @@
         </attribute>
         <attribute>
             <description>
-                <![CDATA[]]>
+                <![CDATA[Defines the maximum height of the scrollable area.]]>
             </description>
             <name>scrollHeight</name>
             <required>false</required>
-            <type>java.lang.Integer</type>
+            <type>java.lang.String</type>
```
jepsar: We should mention this in the migration guide, or accept Object. @melloware WDYT?

jepsar (changes requested):

melloware:
Yes please update Migration Guide and also agree on the "Select All" label.

primefaces/src/main/resources/META-INF/primefaces-p.taglib.xml
jepsar: @melloware so not change to Object?

primefaces/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
melloware: `MessageFactory.getMessage(InputRenderer.ARIA_FILTER)` for example if you look in the code for `MessageFactory`

primefaces/src/main/resources/META-INF/primefaces-p.taglib.xml
melloware: I think he made it consistent with how SelectOne Menu is so we should make it consistent. And since its a change in property type we should mention it in 13.0/migration guide.

primefaces/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
melloware: Looks like we have a `primefaces.messages.aria.CLOSE = Close` label: https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/resources/org/primefaces/Messages_en.properties

primefaces/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
melloware: Looks like we have `primefaces.datatable.aria.HEADER_CHECKBOX_ALL = Select All` label https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/resources/org/primefaces/Messages_en.properties

jepsar: Yeah, I already linked it (including line number) 😁

jepsar (approved):

jepsar:
@ltlBeBoy Thanks! Awesome job!

ltlBeBoy:
Can we close this PR? For the other issues I would open new PRs.
I also wanted to convert some integration tests from `SelectOneMenu`. But e.g. filtering is not yet supported for the IT. I would add support when I have time and add the ITs to the other PRs. OK?

melloware:
squashed and merged!

ltlBeBoy:
@melloware I just noticed that some basic CSS have to be added to `resources/primefaces/forms/forms.css` in PrimeFaces package. Should I add them here or is this a job for #9674 ?

jepsar:
form.css is not part of the themes, so you can open a PR for it

PR #9654, opened 2023-01-12
#9646 Core: Filter normalization should be optional

### LINKED ISSUE #9646: Core: Filter normalization should be optional
We now normalize options and filter text by default (#9629). This should be optional.

SelectOneMenu, SelectOneListbox, SelectManyMenu, SplitButton, PickList

jepsar:
Fix #9646

Added a draft before I fix all instances. @melloware WDYT? Fix all of them like this?


tandraschko:
I really wonder If we should make it only configurable global?
Not sure

jepsar:
> I really wonder If we should make it only configurable global? Not sure

Could be an option. We could apply it on server side filtering as well.

melloware:
what about like RTL where we turn it on with a web.xml param so the whole app is now normalized or not?  I think that is what @tandraschko is saying make it a global config and you can always pass the value to the widgets without allowing customization per widget just have them respect the global flag?

Only because I really don't think someone would want to turn this on for one component and not another??

tandraschko:
yeah
we have 3 different ways:
1) attr on component
2) global attrs (we should really make this into Faces 4.1)
3) context param

in general we have to think about it: is it really required to set it on each component? global always makes sense but better we go 2) then

also i wonder if we should name it filterNormalize? same as current caseSensitiv, shouldnt it be filterCaseSensitiv?

jepsar:
Something like this?

tandraschko:
i dont know TBH

Could it make sense to have it configurable on each component? Then we should go for 1+2

jepsar:
You mean https://primefaces.github.io/primefaces/12_0_0/#/core/globalattributes ? That is how it was before the last commit, right?

tandraschko:
yep - sorry for confusing ;D

jepsar:
Okay, back to attribute. I agree that `filterXxx` is better for filter related attributes and is in line with the existing filter attributes. I do think we should (breaking) change `caseSensitive` to `filterCaseSensitive`, but should we do that in this PR?

melloware (approved):

PR #9656, opened 2023-01-12
Fix #9653: OverlayPanel add cache attribute

### LINKED ISSUE #9653: OverlayPanel: Add "cache" attribute
### Description

We have the cache property for `p:dialog`:

https://primefaces.github.io/primefaces/12_0_0/#/components/dialog:

> cache | true | Boolean | Only relevant for dynamic="true": Defines if activating the dialog  should load the contents from server again. For cache="true" (default)  the dialog content is only loaded once.
> -- | -- | -- | --

Please add this cache property also for p:overlayPanel, so that we can use something like this:

```
<p:overlayPanel dynamic="true" cache="false"> 
    dynamic and uncached content.
</p:overlayPanel >
```

### Additional context

See https://github.com/primefaces/primefaces/issues/9652 for reproducer.

melloware:
Fix #9653: OverlayPanel add cache attribute

### DIFF
--- a/docs/13_0_0/components/overlaypanel.md
+++ b/docs/13_0_0/components/overlaypanel.md
@@ -23,22 +23,23 @@ id | null | String | Unique identifier of the component
 rendered | true | Boolean | Boolean value to specify the rendering of the component, when set to false component will not be rendered.
 binding | null | Object | An el expression that maps to a server side UIComponent instance in a backing bean
 widgetVar | null | String | Name of the client side widget.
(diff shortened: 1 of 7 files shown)


melloware:
Yep but it works perfectly it runs the ajax request every time it just only replaces the children. That is how all the lazy components seem to work. So I want you to test to see if you can make it work I assume there was some reason it was implemented this way.

nimo23:
Yes, `p:overlayPanel` with` cache=false` works.

PR #9672, opened 2023-01-13
Fix #8205: DataExporter useLocale attribute

### LINKED ISSUE #8205: XML Exporter: Generate element names for a fixed locale without breaking other exporter behavior
**Description**

In my multi-lingual application, I export tables to XML output and other formats like PDF and XLSX. The table column header string changes with the selected language.

The current implementation of DataTableXMLExporter derives tag names in the XML output from the corresponding column header value, which change according to the current locale setting used in the application.

This leads to different XML element names in the output for different locale settings, which really makes no sense at all.

When I set a fixed language exportHeaderValue string for the XMLTableExporter it will also appear for PDF, XSLX and other formats, so the user will stumble about column headers in a wrong language there. 

The application is currently using PrimeFaces 11.

Clearly, element names in the XML exporter and table headers in PDF/XLSX are two different things. I am looking for a way to specify both without having cross-dependencies between them.

One quick idea for a solution would be to default to a single locale in this case, which has the benefit to not having to add extra element name information on each  column.

Can this be circumvented by extending DataTableXMLExporter?

I think I need to override Exporter.exportValue(...) for this, but I am lost on how to force the locale to use there.

Would it make sense to at least pin down the locale used for element tags by an option to DataTableXMLExporter?

Whatever the solution is, it has to be on a per-Exporter basis, just targeting the XML export.

melloware:
Fix #8205: DataExporter useLocale attribute

usage:
```xml
 <p:dataExporter type="xml" target="tbl" fileName="products" useLocale="false"/>
```

TRUE (default):
```xml
<tbl>
	<product>
		<code>nvklal433</code>
		<name>Black Watch</name>
		<category>Accessories</category>
		<quantity>61</quantity>
		<price>$72.00</price>
	</product>
```

FALSE:
```xml
<?xml version="1.0"?>
<tbl>
	<product>
		<j_id_9h>waas1x2as</j_id_9h>
		<j_id_9k>Headphones</j_id_9k>
		<j_id_9n>Electronics</j_id_9n>
		<j_id_9q>8</j_id_9q>
		<j_id_9t>$175.00</j_id_9t>
	</product>
```

### DIFF
--- a/docs/11_0_0/components/dataexporter.md
+++ b/docs/11_0_0/components/dataexporter.md
@@ -19,6 +19,9 @@ type | null | String | Export type: "xls", "xlsx", "xlsxstream", "pdf", "csv", "
 target | null | String | Search expression to resolve one or multiple datatables.
 fileName | null | String | Filename of the generated export file, defaults to datatable id.
 pageOnly | false | Boolean | Exports only current page instead of whole dataset
+visibleOnly | false | Boolean | Exports only visible columns instead of whole dataset.
+exportHeader | true | Boolean | When enabled, the header will be exported.
+exportFooter | true | Boolean | When enabled, the footer will be exported.
 preProcessor | null | MethodExpr | PreProcessor for the exported document.
 postProcessor | null | MethodExpr | PostProcessor for the exported document.
 encoding | UTF-8 | String | Character encoding to use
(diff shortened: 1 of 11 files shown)


tandraschko:
why useLocale if it swichtes columnId/columnHeader? sounds bit weird

melloware:
You are correct.  @stolp and I have discussed a much better solution I am closing this PR and will open another.

PR #9689, opened 2023-01-16
Fix #8205: ExportTag and ExportRowTag attributes

### LINKED ISSUE #8205: XML Exporter: Generate element names for a fixed locale without breaking other exporter behavior

melloware:
Fix #8205: ExportTag and ExportRowTag attributes

For TreeTable and Datatable allowing you to do this..

```xml
                
<p:dataTable id="tbl" var="product" value="#{dataExporterView.products}" exportRowTag="rowval">
	<p:column headerText="Code" exportTag="code-tag" >
		<h:outputText value="#{product.code}"/>
	</p:column>
	<p:column headerText="Name" exportTag="name-tag" >
		<h:outputText value="#{product.name}"/>
	</p:column>
	<p:column headerText="Category" exportTag="cat-tag">
		<h:outputText value="#{product.category}"/>
	</p:column>
	<p:column headerText="Quantity" exportTag="quantity-tag">
		<h:outputText value="#{product.quantity}"/>
	</p:column>
	<p:column headerText="Price" exportTag="price-tag">
		<h:outputText value="#{product.price}">
			<f:convertNumber type="currency" />
		</h:outputText>
	</p:column>
</p:dataTable>
```

And output will look like this...

```xml
<?xml version="1.0"?>
<tbl>
	<rowval>
		<code-tag>mbvjkgip5</code-tag>
		<name-tag>Galaxy Earrings</name-tag>
		<cat-tag>Accessories</cat-tag>
		<quantity-tag>23</quantity-tag>
		<price-tag>$34.00</price-tag>
	</rowval>
	<rowval>
		<code-tag>250vm23cc</code-tag>
		<name-tag>Green T-Shirt</name-tag>
		<cat-tag>Clothing</cat-tag>
		<quantity-tag>74</quantity-tag>
		<price-tag>$49.00</price-tag>
	</rowval>
```

### DIFF
--- a/docs/13_0_0/components/column.md
+++ b/docs/13_0_0/components/column.md
@@ -32,6 +32,7 @@ treetable and more.
 | exportValue | null | String | Defines the value of the cell to be exported if you want something other than the cell contents or exportFunction.
 | exportRowspan | 0 | Integer | Defines the number of rows the column spans to be exported.
 | exportColspan | 0 | Integer | Defines the number of columns the column spans to be exported.
+| exportTag | null | String | If XML data exporter in use, this allows customization of the column tag in the XML.
 | field | null | String | Name of the field associated to bean "var". If not specified, filterBy-sortBy values are used to identify the field name.
 | filterable | true | Boolean | Boolean value to mark column as filterable.
 | filterBy | null | ValueExpr | ValueExpression to be used for filtering.
--- a/docs/13_0_0/components/columns.md
+++ b/docs/13_0_0/components/columns.md
@@ -30,6 +30,7 @@ Columns is used by datatable to create columns dynamically.
 | exportHeaderValue | null | String | Defines if the header value of column to be exported.
 | exportRowspan | 0 | Integer | Defines the number of rows the column spans to be exported.
 | exportValue | null | String | Defines the value of the cell to be exported if something other than the cell contents or exportFunction.
+| exportTag | null | String | If XML data exporter in use, this allows customization of the column tag in the XML.
 | exportable | true | Boolean | Defines if the column should be exported by dataexporter.
 | field | null | String | Name of the field associated to bean "var". If not specified, filterBy-sortBy values are used to identify the field name.
 | filterBy | null | ValueExpr | ValueExpression to be used for filtering.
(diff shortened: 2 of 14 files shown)


stolp:
Shouldn't dataTable also have a exportTag attribute? Or is this missing from the example here?

stolp:
It's not that I cannot live with that decision. But still, the rationale behind it escapes me.

In my use case, I set my table id to `projectTable`, but would like to export the table to XML with the tag `projects` for the document and `project` for the column elements. It feels not right, that I can now do the latter but not the former without changing the id.

As far as I can see, this has nothing to do with the use of EL in an id attribute.

Sorry, should I be a PITA here.

melloware:
yeah but its an immutable field?   You are introducing an extra tag field here when to me its not necessary. If you want your XML to be `projects` then simply set your `id="projects"` no?

melloware:
I guess my point is its already tag soup and adding this extra tag for this very specific use case when its...

A) not used in other exporters
B) immutable (no EL expression can change it)
C) Locale agnostic

stolp:
For me it just following the same design conventions as with the p:column/exportTag attribute.

There we already argued, that id and exportTag are similar but semantically different attributes.

Following the paradigm of least surprise I would expect the exportTag in dataTable and potentially other XML-exportable components as well.

If a user doesn't find it there, he will have to guess which other attribute is used instead. This feels inconsistent to me.

Besides there is a collision with id naming scheme vs. XML export schemes.

Other than that, your points are of course all valid and sound.

melloware:
I submitted a PR and Integration test adding `exportTag`

stolp:
Thank you. This relieves my soul. Sorry again for nagging.

PR #9709, opened 2023-01-19
Fix #9650: event.which to .key or .button

### LINKED ISSUE #9650: Core: KeyboardEvent.which/MouseEvent.which are deprecated
`KeyboardEvent.which` is deprecated and should be replaced with `KeyboardEvent.key` everywhere.

Here is a great Keyboard event viewer: https://w3c.github.io/uievents/tools/key-event-viewer.html

Also its universally supported now: https://caniuse.com/?search=event.key

Also changed `MouseEvent.which` to `MouseEvent.button`

melloware:
Fix #9650: event.which to .key or .button

also I modified the KeyFilter plugin to use the same code I wrote for PrimeReact so its much cleaner and removes a lot of the hacks that are no longer needed.


melloware:
OK very happy with this for review.  it makes the code more...

1. Modern
2. Supported in all browsers and mobile devices
3. Cleans up the code and makes it more maintainable.

PR #9711, opened 2023-01-19
Fix #8765/#7767/#7763/#7765/#3781: AutoComplete/SelectOneMenu/RowExpansion/OrderList duplicate ids

### LINKED ISSUE #7767: SelectOneMenu: duplicate ids with custom content

### LINKED ISSUE #8765: Core: Prevent duplicate IDs in DOM

melloware:
Fix #8765  AutoComplete/SelectOneMenu/RowExpansion/OrderList/SelectCheckboxMenu duplicate ids
Fix #7767/#7763/#7765/#3781: AutoComplete/SelectOneMenu/RowExpansion/OrderList/SelectCheckboxMenu duplicate ids

I am reviving this PR since @ltlBeBoy just added support for SelectCheckBoxMenu and @Bukama has pointed out duplicate ID's cause Accessibility issues.

@tandraschko I know you said this should be done with UIRepeat/NamingContainer but honestly I looked all over for examples and studied the code and I could not figure out how to make these components easily support that for this "Advanced" rendering feature.  Any hints would be great but I think we might need to go with this short term solution until a longer term solution can be found?


primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
```diff
@@ -608,6 +608,34 @@ public static boolean isNestedWithinIterator(UIComponent component) {
...
+
+        // encode the component
+        component.encodeAll(context);
+
+        // restore the original writer
+        context.setResponseWriter(writer);
+
+        // append index to all id's
+        String encodedComponent = fsw.toString();
+        encodedComponent = encodedComponent.replaceAll("\\sid=\"(.*?)\"", " id=\"$1:" + index + "\"");
```
jepsar: Separator char could be configured to something different then `:`

melloware: You are right!

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
```diff
@@ -630,8 +630,9 @@ public static void encodeIndexedId(FacesContext context, UIComponent component,
         context.setResponseWriter(writer);
 
         // append index to all id's
+        char separator = UINamingContainer.getSeparatorChar(context);
```
jepsar, line 633: Is this a costly operation? If so maybe cache the separator char?

melloware: I don't think its costly its done all over the code base.  Maybe a question for @tandraschko whether we want to cache it in `Lazy` in PrimeEnvironment?

jepsar: Just asking as we are iterating over a list of items.. 🙂

primefaces/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
```diff
@@ -46,11 +46,7 @@
 import org.primefaces.expression.SearchExpressionUtils;
 import org.primefaces.renderkit.InputRenderer;
 import org.primefaces.renderkit.SelectManyRenderer;
-import org.primefaces.util.ComponentUtils;
-import org.primefaces.util.Constants;
-import org.primefaces.util.HTML;
-import org.primefaces.util.MessageFactory;
-import org.primefaces.util.WidgetBuilder;
+import org.primefaces.util.*;
```
Bukama: Star-Import wanted or by accident?

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
```diff
@@ -608,6 +608,35 @@ public static boolean isNestedWithinIterator(UIComponent component) {
...
+        // encode the component
+        component.encodeAll(context);
+
+        // restore the original writer
+        context.setResponseWriter(writer);
+
+        // append index to all id's
+        char separator = UINamingContainer.getSeparatorChar(context);
+        String encodedComponent = fsw.toString();
+        encodedComponent = encodedComponent.replaceAll("\\sid=\"(.*?)\"", " id=\"$1" + separator + index + "\"");
```
Bukama, line 635: Just to avoid a mindgap. Does `id=\"$1"` already contains the first separator, so you only need to add the second and the index to achieve `id="form:test:0"`

Bukama (approved):

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
melloware: yes the capture group captures the entire current ID so can then append `:1`, `:2` etc onto it!

primefaces/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenuRenderer.java
melloware: Our standard policy on PrimeFaces is that once you have 5 imports from the same package switch it to *.  My IDE is set to do that automatically.

tandraschko:
Need to think about it

melloware:
No problem.. even if you could give me a hint on how it could be done a more proper way that would be great.

tandraschko:
Why do we need a ID anyway in such cases?

melloware:
@Bukama do you have a specific accessibility use case where having the duplicate id's causes issues?

Also I think the issue is @VsevolodGolovanov mentioned it in his issue its not how any other JSF component works and we can't stop people from doing this:

```xml
<p:selectOneMenu id="menu" var="v">
		<f:selectItems value="#{testView.selectItems}" />
		<p:column>
			<h:outputText id="item_id" value="#{v.id}" />
		</p:column>
		<p:column>
			<h:outputText id="item_name" value="#{v.name}" />
		</p:column>
	</p:selectOneMenu>
```

Maybe they need unique ID's for from JS code that have written or some Selenium Test code?

Bukama:
> @Bukama do you have a specific accessibility use case where having the duplicate id's causes issues?

Duplicate Ids are a) not valid HTML, therefor it validates SC 4.1.1 which is one criteria of the WVAG as valid HTML is THE base for assistence technologies. Having duplicate ids those technologies either can't identify the correct element and/or (screenreaders) tend to double speaking as multiple elements are read out loud. These kind of errors are making every accessibility test fail "instantly" as they are from the most weightend error class.

> Maybe they need unique ID's for from JS code that have written or some Selenium Test code?

Well it's coming to the same point as the assistence technologies - identifying the right element. So you could argue to gains at once ;)

melloware:
Thank you @Bukama for your insight!

tandraschko:
also need to think about it, currently i dont have much time for PF as i mostly help to bring Apache EE10 stuff out

melloware:
@tandraschko if you look at the current merge conflict you will see @christophs78 had to do exactly what I am proposing in this fix to make the AutoComplete properly accessible.

tandraschko:
yep
can you resolve conflicts? i'm ok to apply it if you aligned autoComplete with others

melloware:
Resolved and all IT's passing.

PR #9721, opened 2023-01-23
Fix #8472: HTML5 do not render `text/javascript`

### LINKED ISSUE #8472: Core: HTML5 and script types
**Description**
JSF 4.0 will no longer use 'type="text/JavaScript"'.  Therefore in the PF code base anywhere type is being used we should detect 4.0 and conditionally print it out or not.

melloware:
Fix #8472: JSF40 do not render `text\javascript`

Added new web.xml param since we cann't detect DocType easily.

```xml
<context-param>
    <param-name>primefaces.HTML5_COMPLIANCE</param-name>
</context-param>
```

### DIFF
--- a/primefaces/src/main/java/org/primefaces/renderkit/RendererUtils.java
+++ b/primefaces/src/main/java/org/primefaces/renderkit/RendererUtils.java
@@ -130,7 +130,7 @@ public class RendererUtils {
      * @param context the FacesContext
      * @throws IOException if any error occurs
      */
-    public static void encodeScript(FacesContext context) throws IOException {
+    public static void encodeScriptIfNecessary(FacesContext context) throws IOException {
         PrimeRequestContext requestContext = PrimeRequestContext.getCurrentInstance(context);
         PrimeApplicationContext applicationContext = requestContext.getApplicationContext();
         if (applicationContext.getConfig().isHtml5Compliant()) {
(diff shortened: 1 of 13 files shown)


melloware:
I don;t think we can do this since DocType is new in JSF 4.0?

```java
 UIViewRoot viewRoot = context.getViewRoot();

        if (viewRoot == null) {
            return false;
        }

        Doctype doctype = viewRoot.getDoctype();
```

tandraschko:
Yeah, thats new

melloware:
Instead of JSF4.0 check i think we should just add this as a `web.xml` param.  People will know whether they are using HTML5 or not??

primefaces/src/main/java/org/primefaces/renderkit/RendererUtils.java
```diff
@@ -119,4 +124,20 @@ public static RenderKit getRenderKit(FacesContext context) {
         return ((RenderKitFactory) FactoryFinder.getFactory(FactoryFinder.RENDER_KIT_FACTORY)).getRenderKit(context, renderKitId);
     }
 
+    /**
+     * HTML5 Doctype does not require the script type on JavaScript files.
+     *
+     * @param context the FacesContext
+     * @throws IOException if any error occurs
+     */
+    public static void encodeScript(FacesContext context) throws IOException {
```
tandraschko: please use a better name like encodeScriptTypeIfNecessary

PR #9722, opened 2023-01-23
Fix #4068: MenuItem allow action and link on same item

### LINKED ISSUE #4068: MenuItem: link and command at the same time

melloware:
Fix #4068: MenuItem allow action and link on same item

this allows a MenuItem like this...

```xml
<p:menuitem 
value="Create" 
url="http://jira.apache.org/create" 
actionListener="#{view.create}" 
update="dialog" 
oncomplete="PF('dlg').show();"/>
```

IF single clicking the link will cause an AJAX request to `view.create` and open a new dialog.  If you right-click and do "Open In New Tab" it opens the URL ` url="http://jira.apache.org/create"`


jepsar (approved):

jepsar:
Maybe add an example to the showcase? Can we do something similar for `CommandButton`?

melloware:
i will add a showcase example.  I need to loo at CommandLink probably more than ComandButton?

PR #9730, opened 2023-01-25
Fix #4068: Menu fix CTRL/CMD+click open in new tab

### LINKED ISSUE #4068: MenuItem: link and command at the same time

melloware:
Fix #4068: Menu fix CTRL/CMD+click open in new tab

### DIFF
--- a/primefaces/src/main/java/org/primefaces/renderkit/MenuItemAwareRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/renderkit/MenuItemAwareRenderer.java
@@ -112,6 +112,10 @@ public class MenuItemAwareRenderer extends OutcomeTargetRenderer {
                         : buildNonAjaxRequest(context, ((UIComponent) menuitem), form, ((UIComponent) menuitem).getClientId(context), true);
             }
 
+            if (isLink) {
+                // allow CTRL+CLICK link to open new tab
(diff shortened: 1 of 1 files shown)


primefaces/src/main/java/org/primefaces/renderkit/MenuItemAwareRenderer.java
```diff
@@ -112,6 +112,10 @@ protected void encodeOnClick(FacesContext context, UIComponent source, MenuItem
                         : buildNonAjaxRequest(context, ((UIComponent) menuitem), form, ((UIComponent) menuitem).getClientId(context), true);
             }
 
+            if (isLink) {
+                // allow CTRL+CLICK link to open new tab
+                command = "if(event.ctrlKey||event.metaKey){return true};" + command;
```
jepsar, line 117: I'm wondering if we should do `PF.isMetaKey(event)` (not sure how to name it) instead of `event.ctrlKey||event.metaKey`. If we would have a function, we could do OS checks and makes it easier to refactor. Something like
```js
return isMacOS ? event.metaKey : event.ctrlKey;
```
On macOS Control+Click will show the context menu by default.

melloware: Great idea. Let me fix.

PR #9734, opened 2023-01-26
Fix #4068: isMetaKey check based on OS

### LINKED ISSUE #4068: MenuItem: link and command at the same time

melloware:
Fix #4068: isMetaKey check based on OS

### DIFF
--- a/primefaces/src/main/java/org/primefaces/renderkit/MenuItemAwareRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/renderkit/MenuItemAwareRenderer.java
@@ -114,7 +114,7 @@ public class MenuItemAwareRenderer extends OutcomeTargetRenderer {
 
             if (isLink) {
                 // allow CTRL+CLICK link to open new tab
-                command = "if(event.ctrlKey||event.metaKey){return true};" + command;
+                command = "if(PF.metaKey(event)){return true};" + command;
             }
             onclick = (onclick == null) ? command : onclick + ";" + command;
         }
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
@@ -1,4 +1,13 @@
 if (!PrimeFaces.utils) {
+    
+   /**
+    * Shortcut for is this CMD on MacOs or CTRL key on other OSes. 
+    * @param {JQuery.TriggeredEvent} e The key event that occurred.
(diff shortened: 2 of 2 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
```diff
@@ -561,6 +561,15 @@ if (!PrimeFaces.utils) {
...
             }
         },
+        
+        /**
+         * Is this CMD on MacOs or CTRL key on other OSes. 
+         * @param {JQuery.TriggeredEvent} e The key event that occurred.
+         * @return {boolean} `true` if the key is a meta key, or `false` otherwise.
+         */
+        isMetaKey: function(e) {
+            return PrimeFaces.env.browser.mac ? event.metaKey : event.ctrlKey;
```
jepsar:
```suggestion
            return PrimeFaces.env.browser.mac ? e.metaKey : e.ctrlKey;
```

primefaces/src/main/java/org/primefaces/renderkit/MenuItemAwareRenderer.java
```diff
@@ -114,7 +114,7 @@ protected void encodeOnClick(FacesContext context, UIComponent source, MenuItem
 
             if (isLink) {
                 // allow CTRL+CLICK link to open new tab
-                command = "if(event.ctrlKey||event.metaKey){return true};" + command;
+                command = "if(PrimeFaces.utils.isMetaKey(event)){return true};" + command;
```
jepsar: It's not possible to have this as `PF.isMetaKey` (which is way shorter)?

melloware: PF is only for widgets so the only shorter way would be `PrimeFaces` only but then this is a utility and belongs in the utils namespace

```js
    PF = function(widgetVar) {
    	var widgetInstance = PrimeFaces.widgets[widgetVar];

    	if (!widgetInstance) {
	        PrimeFaces.widgetNotAvailable(widgetVar);
    	}

        return widgetInstance;
    };
```

melloware: OK check out my latest commit.

jepsar: > PF is only for widgets ...

I don't know what made me think it was an object. 😅

jepsar (approved):

PR #9736, opened 2023-01-26
Fixed #9733: Picklist accessibility updates

### LINKED ISSUE #9733: PickList: Accessibility
### Describe the bug

1. Using the picklist PrimeFaces will generate together with the picklist this HTML:

`<div id="j_idt318:pickList_ariaRegion" class="ui-helper-hidden-accessible" role="region" aria-live="polite" aria-atomic="true"></div> `

This is very useful for accessibility, but it's missing the aria-label that will allow the screen reader to identify it and read it.

-------------------------------------------------

Using the filter feature PrimeFaces will generate this HTML:
`<input id="j_idt318:pojoPickList_source_filter" name="j_idt318:pojoPickList_source_filter" type="text" autocomplete="off" class="ui-picklist-filter ui-inputfield ui-inputtext ui-widget ui-state-default ui-corner-all ui-source-filter-input" aria-label="Filter" tabindex="0">`

There are two problems with this:
2. The aria-label will be the same for sourceList and destinationList, as you can immagine for the user could be a bit confusing
3. The filter has a aria-label but there isn't the possibility to add a label or placeholder to it

### Reproducer

You can check it out using the showcase for PickList

### Expected behavior

1. The HTML generated for the element with `role=region` should include aria-label (as per [ARIA standard](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/region_role)):
`<div id="j_idt318:pickList_ariaRegion" class="ui-helper-hidden-accessible" role="region" aria-live="polite" aria-atomic="true" aria-label="MYPICKLISTREGIONDESCRIPTION"></div> `

The aria-label could be implemented in different ways:
- Insert  using a param on `<p:pickList>` like "description" or "roleAriaLabel". 
- Insert using f:facet like `<f:facet name="description">MYPICKLISTREGIONDESCRIPTION</f:facet>`
-------------------------------------------------

2. Add two params to the `<p:pickList>` like `sourceFilterLabel` and  `destinationFilterLabel` that will generate a label above it like this:

```
<div class="ui-picklist-filter-container">
  <label for="j_idt318:pojoPickList_source_filter">Available Countries Filter</label>
  <input
    id="j_idt318:pojoPickList_source_filter"
    name="j_idt318:pojoPickList_source_filter"
    type="text"
    autocomplete="off"
    class="ui-picklist-filter ui-inputfield ui-inputtext ui-widget ui-state-default ui-corner-all ui-source-filter-input"
    aria-label="Filter"
    tabindex="0"/>
<span class="ui-icon ui-icon-search"></span>
</div>
```

Furthermore should be able to:
- Must have a label for source and destination list. So the params will be `sourceFilterLabel` and `destinationFilterLabel`
- In case one of the "filterLabels" is not present the label can follow the rules of point 1, have as the default `Source List Filter` and  `Destination List Filter` in the .properties

-------------------------------------------------

3. And add another param called "filterPlaceholder" that will generate the input for the filter with the specified placeholder:

```
<div class="ui-picklist-filter-container">
  <input
    id="j_idt318:pojoPickList_source_filter"
    name="j_idt318:pojoPickList_source_filter"
    type="text"
    autocomplete="off"
    class="ui-picklist-filter ui-inputfield ui-inputtext ui-widget ui-state-default ui-corner-all ui-source-filter-input"
    aria-label="Filter"
    tabindex="0"/ placeholder="MYCUSTOMPLACEHOLDER">
<span class="ui-icon ui-icon-search"></span>
</div>
```
As before the default can be insert in the .properties and must have a label for source and another for destination

### PrimeFaces edition

Community

### PrimeFaces version

12.0.0

### Theme

_No response_

### JSF implementation

MyFaces

### JSF version

2.3

### Java version

1.8

### Browser(s)

_No response_

melloware:
Fixed #9733: Picklist accessibility updates

`aria-region` now gets `aria-label` when its updated.
'source' and 'target' filter placeholder which is exist replace the `aria-label`  if no placeholder then it defaults to what it is doing now with the message filter.

### DIFF
--- a/docs/13_0_0/components/picklist.md
+++ b/docs/13_0_0/components/picklist.md
@@ -55,6 +55,8 @@ label | null | String | A localized user presentable name.
 itemDisabled | false | Boolean | Specified if an item can be picked or not.
 showSourceFilter | false | Boolean | Displays and input filter for source list.
 showTargetFilter | false | Boolean | Displays and input filter for target list.
+sourceFilterPlaceholder | null | String | Placeholder for the source filter input element.
+targetFilterPlaceholder | null | String | Placeholder for the target filter input element.
 filterMatchMode | startsWith | String | Match mode for filtering, valid values are startsWith, contains, endsWith and custom.
 filterFunction | null | String | Name of the javascript function for custom filtering.
 filterNormalize | false | Boolean | Defines if filtering would be done using normalized values (accents will be removed from characters).
--- a/docs/13_0_0/gettingstarted/whatsnew.md
+++ b/docs/13_0_0/gettingstarted/whatsnew.md
@@ -31,6 +31,7 @@ This page contains a list of big features. Please check the GitHub issues for al
     * Added attribute `flex` to support PrimeFlex CSS instead of legacy Grid CSS.
 * PickList
     * Added attribute `filterNormalize` to allow normalized filtering (without accents).
+    * Added attributes `sourceFilterPlaceholder` and `targetFilterPlaceholder` for accessibility.
 * TabView
     * Added `focusOnLastActiveTab` if you want to focus on the tab that the user last activated.
     * Added `footer` facet to add a footer to the whole TabView not per tab.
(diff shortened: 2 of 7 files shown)


melloware:
I will respectfully diagree with needing a `<label>` for the filter.   HTML input are perfectly valid without labels as long as they have an `aria-label` they do not need an associated label.

See even Google has an input with no `<label>`

image

ZeroBrushV2:
I see where you come from.

To be honest I'm just telling you what a third party company employee, that does accessibility checks, told me.
Let's see if the placeholder will content them.

Thanks for your work :)

melloware:
Yep most of the time Placeholder will satisfy the needs of Accesibility and `aria-label`.  thanks for the report!

PR #9744, opened 2023-01-30
Fix #9282: Responsive instead of table layout defaults

### LINKED ISSUE #9282: Core: Deprecate `<table>` based layouts
### Describe the bug

When using the [SelectOneRadio](https://www.primefaces.org/showcase/ui/input/oneRadio.xhtml) with `layout` configuration set to `lineDirection`(wich is the default value!), then invalid HTML is rendered.

> The value of the for attribute of the label element must be the ID of a non-hidden form control

This is caused by the fact that with the rendering option `lineDirection` a table is rendered and the labels of the options point to the id of the table. But tables are not allowed to have a label.

radioInput_tablerender

### Reproducer

https://www.primefaces.org/showcase/ui/input/oneRadio.xhtml (Showcase Primefaces 12)

### Expected behavior

* Valid HTML is rendered, with the label pointing to the input field (hich could be problematic as it's hidden)
* No usage of layout tables, therefore change the default value to `responsive` or chaning the implementation of the `lineDirection` (and others)

### PrimeFaces edition

Community

### PrimeFaces version

11.0.0 at work (Showcase is using 12.0.0)

### Theme

Saga

### JSF implementation

All

### JSF version

2.2

### Java version

8

### Browser(s)

All

melloware:
Fix #9282: Responsive instead of table layout defaults

### DIFF
--- a/primefaces/src/main/java/org/primefaces/renderkit/CoreRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/renderkit/CoreRenderer.java
@@ -30,6 +30,7 @@ import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
 import javax.el.PropertyNotFoundException;
+import javax.faces.application.ProjectStage;
 import javax.faces.component.*;
 import javax.faces.component.behavior.ClientBehavior;
 import javax.faces.component.behavior.ClientBehaviorContext;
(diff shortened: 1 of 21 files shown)


jepsar:
I'm wondering if we should log warnings when you use deprecated features.

melloware:
Good idea I will log them in Stage DEVELOPMENT.

melloware:
Added a LOG warning to each of those only in DEVELOPMENT stage.

primefaces/src/main/java/org/primefaces/renderkit/CoreRenderer.java
```diff
@@ -878,4 +879,15 @@ protected String getIconOnlyButtonText(String title, String ariaLabel) {
         String text = (title != null) ? title : ariaLabel;
         return (text != null) ? text : "ui-button";
     }
+
+    /**
+     * Logs a WARN log message in ProjectStage == DEVELOPMENT.
+     * @param context the FacesContext
+     * @param message the message to log
+     */
+    protected void logDevelopmentWarning(FacesContext context, String message) {
```
jepsar, line 888: Maybe `Supplier<String>` instead of `String`? Saves some code execution in production.

melloware: will fix

PR #9763, opened 2023-02-04
avoid ViewMap creation (fix #9762)

### LINKED ISSUE #9762: CSP: nonce evaluation shouldn't create ViewMap
While fixing #9461 with #9462 I introduced a small flaw.
These calls to `getViewMap()`
https://github.com/primefaces/primefaces/blob/e709c351c81aa75777d245ec621b3fd47daa8cc9/primefaces/src/main/java/org/primefaces/csp/CspState.java#L58
and
https://github.com/primefaces/primefaces/blob/e709c351c81aa75777d245ec621b3fd47daa8cc9/primefaces/src/main/java/org/primefaces/csp/CspState.java#L66
create the ViewMap if it doesn't exist, but it doesn't make sense for stateless view.
They must be replaced with `getViewMap(false)` + a null check. PR coming.

Reference: https://docs.oracle.com/javaee/7/api/javax/faces/component/UIViewRoot.html#getViewMap-boolean-

NicolaIsotta:
fixes #9762

### DIFF
--- a/primefaces/src/main/java/org/primefaces/csp/CspState.java
+++ b/primefaces/src/main/java/org/primefaces/csp/CspState.java
@@ -55,7 +55,10 @@ public class CspState {
     public String getNonce() {
         if (nonce == null) {
             if (context.isPostback() || context.getPartialViewContext().isAjaxRequest()) {
-                nonce = (String) context.getViewRoot().getViewMap().get(Constants.RequestParams.NONCE_PARAM);
+                Map<String, Object> viewMap = context.getViewRoot().getViewMap(false);
+                if (viewMap != null) {
+                    nonce = (String) viewMap.get(Constants.RequestParams.NONCE_PARAM);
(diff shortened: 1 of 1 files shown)


tandraschko:
looks good to me

melloware (approved):

melloware:
Thanks for the fix!

NicolaIsotta:
This would be the correct fix:
```
} else {
    nonce = Base64.getEncoder().encodeToString(UUID.randomUUID().toString().getBytes(StandardCharsets.UTF_8));
    if (!context.getViewRoot().isTransient()) {
        context.getViewRoot().getViewMap(true).put(Constants.RequestParams.NONCE_PARAM, nonce);
    }
}
```
I guess this PR cannot be modified further? Should I make a new one?

melloware:
yep just make a new one.

PR #9766, opened 2023-02-05
Fix #9765 - TreeTable: move code into TreeTableFeature features

### LINKED ISSUE #9765: TreeTable: move code into TreeTableFeature features
### Description

TreeTable has implemented few TreeTableFeature so far and yet other features are there but don't implement this interface

Move code into a TreeTableFeature impl (just like DataTable does) 

### Additional context

_No response_

Rapster:


tandraschko:
How many TT IT do we have?

melloware (approved):
This looks great.  Thanks Rapster!

melloware:
i don't think we have nearly as many TT IT as we do DT IT.

tandraschko:
We only have 5 
Could you please add some ITs first for the moved features?
We must avoid breaking functionality

christophs78:
> We only have 5
> Could you please add some ITs first for the moved features?
> We must avoid breaking functionality

Hm. 5 Test-classes. But the first one contains something like 10 tests. We cover filter, sort, expand and to some degree paging. We do not cover features like edit. Far from ideal but not totally worse. More ITs always welcome.

tandraschko:
yeah
as i said, we should have some ITs for the moved features
and if row/cell edit is missing, we should add ITs first

Rapster:
TBH, I don't think I'll have time for writing an IT for this... If someone want to have a go at it, be my guest 😉

tandraschko:
@christophs78 @melloware would you like to help out?
I dont feel good to merge such big refactorings with that amount of ITs

christophs78:
added https://github.com/primefaces/primefaces/commit/32b0effaf2ad2ecc748b206be2ecf9af77d2f1d8 and https://github.com/primefaces/primefaces/commit/8a57cd13e20d95479e0e919ad72fb80a3f5a36dd

melloware:
@Rapster can you rebase this PR so we can watch the IT's run before merging it?

Rapster:
Looks good

PR #9778, opened 2023-02-09
Fix #9773: Add ui-datepicker-week-end class to weekends

### LINKED ISSUE #9773: Datepicker: CSS classes missing
### Describe the bug

[Server API](https://primefaces.github.io/primefaces/12_0_0/#/components/datepicker?id=skinning) lists css classes, which are not found on a rendered datepicker.

For example `.ui-datepicker-week-end` is listed at the API, but not found an the actual component.

The problem is also present in the [showcase datepicker](https://www.primefaces.org/showcase/ui/input/datepicker/datePickerJava8.xhtml)

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

Community

### PrimeFaces version

Master Snapshot

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

17

### Browser(s)

_No response_

melloware:
Fix #9773: Add `ui-datepicker-week-end` class to weekends

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
@@ -1481,16 +1481,16 @@
         },
 
         calculateWeekNumber: function(d) {
-            var dow = this.getFirstDayOfWeek(),
+            var firstDayOfWeek = this.getFirstDayOfWeek(),
                 doy = this.options.locale.firstDayWeekOffset,
-                weekOffset = this.firstWeekOffset(d.year, dow, doy),
+                weekOffset = this.firstWeekOffset(d.year, firstDayOfWeek, doy),
                 week = Math.floor((this.dayOfYear(d) - weekOffset - 1) / 7) + 1;
 
             if(week < 1) {
-                return week + this.weeksInYear(resYear, dow, doy);
+                return week + this.weeksInYear(resYear, firstDayOfWeek, doy);
             }
-            else if(week > this.weeksInYear(d.year, dow, doy)) {
-                return  week - this.weeksInYear(d.year, dow, doy);
+            else if(week > this.weeksInYear(d.year, firstDayOfWeek, doy)) {
+                return  week - this.weeksInYear(d.year, firstDayOfWeek, doy);
             }
             else {
                 return  week;
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
@@ -278,7 +278,8 @@
         },
 
         getSundayIndex: function () {
-            return 7 - this.getFirstDayOfWeek() % 7;
+            var firstDayOfWeek = this.getFirstDayOfWeek();
+            return firstDayOfWeek > 0 ? 7 - firstDayOfWeek : 0;
         },
 
         getSaturdayIndex: function () {
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
@@ -278,6 +278,11 @@
             return firstDayOfWeek > 0 ? 7 - firstDayOfWeek : 0;
         },
 
+        getSaturdayIndex: function () {
+            var firstDayOfWeek = this.options.locale.firstDay !== undefined ? this.options.locale.firstDay : this.options.locale.firstDayOfWeek;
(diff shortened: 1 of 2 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -278,6 +278,11 @@
             return firstDayOfWeek > 0 ? 7 - firstDayOfWeek : 0;
         },
 
+        getSaturdayIndex: function () {
+            var firstDayOfWeek = this.options.locale.firstDay !== undefined ? this.options.locale.firstDay : this.options.locale.firstDayOfWeek;
```
jepsar: I know it's a copied function, but isn't this the same?
```suggestion
            var firstDayOfWeek = this.options.locale.firstDay || this.options.locale.firstDayOfWeek;
```

primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -278,6 +278,11 @@
...
+            return firstDayOfWeek > 0 ? 7 - firstDayOfWeek - 1 : 6;
```
jepsar: Isn't this just
```suggestion
            return 7 - firstDayOfWeek - 1;
```
?

melloware:
Great idea i actually refactored and found a few more places that was being done.  I also updated the MD with some corrected and missing styles.

primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -273,11 +273,20 @@
             return dayIndex >= 7 ? dayIndex - 7 : dayIndex;
         },
 
+        getFirstDayOfWeek: function () {
+            return this.options.locale.firstDay !== undefined ? this.options.locale.firstDay : this.options.locale.firstDayOfWeek;
```
jepsar, line 277: This won't work?
```suggestion
            return this.options.locale.firstDay || this.options.locale.firstDayOfWeek;
```

primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -273,11 +273,20 @@
...
+        },
+
         getSundayIndex: function () {
-            var firstDayOfWeek = this.options.locale.firstDay !== undefined ? this.options.locale.firstDay : this.options.locale.firstDayOfWeek;
+            var firstDayOfWeek = this.getFirstDayOfWeek();
             return firstDayOfWeek > 0 ? 7 - firstDayOfWeek : 0;
```
jepsar, line 282:
```suggestion
            return 7 - this.getFirstDayOfWeek() % 7;
```

primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -273,11 +273,20 @@
...
         },
 
+        getSaturdayIndex: function () {
+            var firstDayOfWeek = this.getFirstDayOfWeek();
+            return 7 - firstDayOfWeek - 1;
```
jepsar:
```suggestion
            return 7 - this.getFirstDayOfWeek() - 1;
```

primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -1474,7 +1483,7 @@
         },
 
         calculateWeekNumber: function(d) {
-            var dow = this.options.locale.firstDay !== undefined ? this.options.locale.firstDay : this.options.locale.firstDayOfWeek,
+            var dow = this.getFirstDayOfWeek(),
```
jepsar: I think this variable name is misleading. Isn't something like `firstDow` better?

primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
melloware: no because if first day is `0` that is JS falsy so it will then go to first dayofWeek right it needs to check for undefined not just for falsiness?  And that is how it already was in the code I just moved it to a function.

jepsar: Ah! Alright. Need to keep that in mind.

primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
melloware: Fixed

PR #9786, opened 2023-02-11
Fix #9784: SelectItems should respect EL settings

### LINKED ISSUE #9784: SelectOneMenu: itemLabel attribute is not honored when using SelectItem as data type
### Describe the bug

The component should render the itemLabel as instructed by the provided EL, even if the backed data type is SelectItem. The default rendering should only be done when the attribute is missing.

Maybe the [getSelectItems](https://github.com/primefaces/primefaces/blob/816651f3837649ad25b2595689e3b5db45684011/primefaces/src/main/java/org/primefaces/renderkit/SelectRenderer.java#L123) should also test if itemLabel is missing.

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

12.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.2.19

### Java version

11

### Browser(s)

_No response_

melloware:
Fix #9784: SelectItems should respect EL settings

### DIFF
--- a/primefaces/src/main/java/org/primefaces/renderkit/SelectRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/renderkit/SelectRenderer.java
@@ -86,7 +86,7 @@ public abstract class SelectRenderer extends InputRenderer {
                             Object item = Array.get(value, j);
 
                             if (item instanceof SelectItem) {
-                                selectItem = (SelectItem) item;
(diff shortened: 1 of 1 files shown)


tandraschko:
Will review on monday

melloware:
Thanks!  Integration tests caught a bug too!

PR #9803, opened 2023-02-16
DataTable: remove features as singleton

### LINKED ISSUE #8061: Datatable: make Features replacable again
# Description
With Primefaces 11, the DataTable and its components saw some revisions that are clashing with how we used it in previous versions.

We have extended the SortFeature, and specifically modified how compare() works within it, as we are dealing with message keys whose localized messages should be compared for sorting, not the message key itself.

With PF11, compare() is now a static method, the DataTable Features are now singletons accessed by getInstance(), and those singletons reference each other (e.g. FilterFeature directly calls SortFeature.getInstance()) making minimally invasive modifications impossible:

```
// update filtered value accordingly to take account sorting
if (table.isSortingCurrentlyActive()) {
  SortFeature.getInstance().sort(context, table);
}
```

The DataTable FEATURES list is also static and unmodifiable, requiring us to overwrite decode() and encodeEnd() just to replace a feature in the list.

From PF11 onward, we have to basically copy and replace the entire implementation and all features, which will be a continuous pain with future updates.

# Describe the solution you would like
Refactor the structure to allow easy modification of features, for example by using protected methods to (1) retrieve the list of features, and (2) retrieve the instance of a feature, that subclasses can extend and modify. The reference implementation should use those methods instead of obtaining singletons directly.

Keep certain methods that can change with different use cases, like compare(), as instance methods instead of making them static.

Thank you for your consideration.

Rapster:
Fix #8061

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java
+++ b/primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java
@@ -26,7 +26,6 @@ package org.primefaces.component.datatable;
 import java.util.*;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
-
 import javax.el.ELContext;
 import javax.el.ValueExpression;
 import javax.faces.FacesException;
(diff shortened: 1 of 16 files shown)


melloware:
Does this solve this issue? https://github.com/primefaces/primefaces/issues/8061

Rapster:
I didn't mean to, but yeah seems like it will fix it

Rapster:
SPI is indeed a most common practise in our code base to replace stuff, but my intent at first was to remove singletons, I just introduced replace for "fun" (in case users wants to fix/workaround few things) I think it's good enough, but i'm open to discussion

melloware:
Yes this puts in more in line with how it used to work too.  SPI would be good but to me not necessary right now.

melloware (approved):

PR #9805, opened 2023-02-17
Update locale-uk.js

tsipiniuk:
updated dayNamesMin


melloware:
Much appreciated!  If you don't mind the UK translation is actually missing a lot of labels if you compare to locale-en.JS if you would like to complete it? 

We would appreciate it!

tsipiniuk:
@melloware Created a new pull #9806. Moved the full Ukrainian translation there.

PR #9809, opened 2023-02-17
Fix #9772 - Remove duplication of DataTable#loadLazyData

### LINKED ISSUE #9772: SONAR: code quality for 13 release
### Description

The codebase might need some attention since last time we focused on this, on this day:
- 131 critical
- 87 major

More details here: https://sonarcloud.io/project/issues?resolved=false&id=org.primefaces%3Aprimefaces

I see a lot of `Make "XXX" transient or serializable"` kind of errors, and I think it'd be great to know once and for all what to do with these warnings?

The rest should be quite easy to fix. If you have any questions, suggestions, feel free to share 😉  

See how we can split the work (it's redundant work and not so fun 😅 )

### Additional context

Also I see we have some singletons for DT/TT features, I'm planning to remove those for many reasons (essentially it's an anti-pattern, and we don't need to define them as singleton)

Rapster:
Out of topic but few things I'm seeing with scrolling, what exists so far:

<html>
<body>

property | default value | type | description
-- | -- | -- | --
liveScroll | false | Boolean | Enables live scrolling.
scrollRows | 0 | Integer | Number of rows to load on live scroll.
scrollable | false | Boolean | Makes data scrollable with fixed header.
virtualScroll | false | Boolean | Loads data on demand as the scrollbar gets close to the bottom. Default is false.
scrollOffset | 0 | Integer |  (not documented and only for live scrolling)

</body>
</html>

What it could be:

<html>
<body>

property | default value | type | description
-- | -- | -- | --
scrollMode | null| string| Possibles value: simple, live (aka infinite), virtual

</body>
</html>

`scrollRows`: suggests the number of rows fetched is different than the number of items displayed, why not reusing rows property, that's how pagination works and that's what scrolling does (loading a chunk is same as loading a page). If I look at the showcase, rows is 40 and scrollRows is 20, and what dt does is fetching scrollRows * 2... So I'm pretty sure this property is useless

`scrollHeight`: I don't see why user has to define this, it's actually annoying to do so

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/api/UIPageableData.java
+++ b/primefaces/src/main/java/org/primefaces/component/api/UIPageableData.java
@@ -252,10 +252,11 @@ public class UIPageableData extends UIData implements Pageable, TouchAware {
 
             if (rowCount > 0 && first >= rowCount) {
                 int numberOfPages = (int) Math.ceil(rowCount * 1d / rows);
+                int newFirst = Math.max((numberOfPages - 1) * rows, 0);
 
-                setFirst(Math.max((numberOfPages - 1) * rows, 0));
+                setFirst(newFirst);
 
-                return true;
+                return first != newFirst;
             }
         }
 
--- a/primefaces/src/main/java/org/primefaces/component/api/UIPageableData.java
+++ b/primefaces/src/main/java/org/primefaces/component/api/UIPageableData.java
@@ -331,11 +331,11 @@ public class UIPageableData extends UIData implements Pageable, TouchAware {
 
     public void updatePaginationData(FacesContext context) {
         setRowIndex(-1);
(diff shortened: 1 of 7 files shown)


tandraschko:
it looks good for me
any reason its still draft?

Rapster:
Yes, still a few tests I'd like to run

Rapster:
@tandraschko @melloware @ravihariharan I removed `DataTable#clearLazyCache`, not sure what was the purpose of that method, specially that wrappedData gets reset right after that method was called

melloware (approved):
LGTM!

tandraschko:
@Rapster you added 2 log statements when calculateFirst is required in loglevel warning
IMO its more like a normal case, i would switch them to fine
it even occurs in integration tests

Issue #9320, opened 2022-10-17
DataTable: JpaLazyDataModel ConverterException for numeric columns when filtering by non-numeric


FlipWarthog:
### Describe the bug

Numeric JPA columns (Long, BigDecimal, etc) cause `JpaLazyDataModel` to throw a `ConverterException` if non-numeric values are add to the filters for those columns.

### Reproducer

https://github.com/FlipWarthog/primefaces-test/tree/hibernate

Running this reproducer using `mvn clean jetty:run -Pmojarra23`, open the test page and filter either "Number Col" or "Decimal Col" with "abc" to get the following result:

```
SEVERE: javax.faces.convert.ConverterException: frmTest:tblJpa: 'abc' must be a number consisting of one or more digits.
        at javax.faces.convert.LongConverter.getAsObject(LongConverter.java:95)
        at org.primefaces.model.JpaLazyDataModel.convertToType(JpaLazyDataModel.java:326)
        at org.primefaces.model.JpaLazyDataModel.applyFilters(JpaLazyDataModel.java:163)
        at org.primefaces.model.JpaLazyDataModel.count(JpaLazyDataModel.java:120)
        at org.primefaces.component.datatable.DataTable.loadLazyData(DataTable.java:509)
        at org.primefaces.component.datatable.feature.FilterFeature.encode(FilterFeature.java:135)
        at org.primefaces.component.datatable.DataTableRenderer.encodeEnd(DataTableRenderer.java:85)
```

```
SEVERE: javax.faces.convert.ConverterException: frmTest:tblJpa: 'abc' must be a signed decimal number.
        at javax.faces.convert.BigDecimalConverter.getAsObject(BigDecimalConverter.java:96)
        at org.primefaces.model.JpaLazyDataModel.convertToType(JpaLazyDataModel.java:326)
        at org.primefaces.model.JpaLazyDataModel.applyFilters(JpaLazyDataModel.java:163)
        at org.primefaces.model.JpaLazyDataModel.count(JpaLazyDataModel.java:120)
        at org.primefaces.component.datatable.DataTable.loadLazyData(DataTable.java:509)
        at org.primefaces.component.datatable.feature.FilterFeature.encode(FilterFeature.java:135)
        at org.primefaces.component.datatable.DataTableRenderer.encodeEnd(DataTableRenderer.java:85)
```

### Expected behavior

[...]

PR #9816, opened 2023-02-20
Fix #9756: JpaLazyDataModel: Regression Issue for #9320

### LINKED ISSUE #9756: JpaLazyDataModel: Regression Issue for #9320
### Describe the bug

#9320 was fixed in 12.0.1, after updating to 12.0.3 the issue is back. I believe the refactoring to fix #9594 removed the original fix.

### Reproducer

Reproducers attached to original issues.

### Expected behavior

See original issues.

### PrimeFaces edition

Elite

### PrimeFaces version

12.0.3

### Theme

All

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

11

### Browser(s)

Chrome latest

FlipWarthog:
Fixes #9756

### DIFF
--- a/primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
+++ b/primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
@@ -366,13 +366,18 @@ public class JpaLazyDataModel<T> extends LazyDataModel<T> implements Serializabl
                     + "; Create a JSF Converter for it or overwrite Object convertToType(String value, Class<?> valueType)!");
         }
 
-        // first convert the object to string
-        String stringValue = sourceConverter == null
-                ? value.toString()
-                : sourceConverter.getAsString(context, UIComponent.getCurrentComponent(context), value);
+        try {
+            // first convert the object to string
+            String stringValue = sourceConverter == null
+                    ? value.toString()
+                    : sourceConverter.getAsString(context, UIComponent.getCurrentComponent(context), value);
 
-        // now convert the string to the required target
-        return targetConverter.getAsObject(context, UIComponent.getCurrentComponent(context), stringValue);
+            // now convert the string to the required target
+            return targetConverter.getAsObject(context, UIComponent.getCurrentComponent(context), stringValue);
+        }
+        catch (Exception e) {
+            return value;
+        }
     }
 
     protected Method getRowKeyGetter() {
--- a/primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
+++ b/primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
@@ -33,6 +33,7 @@ import java.util.*;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
+import javax.el.ELException;
 import javax.faces.FacesException;
 import javax.faces.component.UIComponent;
 import javax.faces.context.FacesContext;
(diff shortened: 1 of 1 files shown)


melloware (approved):

tandraschko:
What exception is thrown?
Please post msg and stacktrace
Catching everything and log nothing is a no-go

FlipWarthog:
When testing in primefaces-test with Jetty, the exception is

```
SEVERE: javax.el.ELException: Cannot convert [abc] of type [class java.lang.String] to [class java.lang.Long]
        at org.apache.el.lang.ELSupport.coerceToNumber(ELSupport.java:400)
        at org.apache.el.lang.ELSupport.coerceToNumber(ELSupport.java:379)
        at org.apache.el.lang.ELSupport.coerceToType(ELSupport.java:527)
        at org.apache.el.ExpressionFactoryImpl.coerceToType(ExpressionFactoryImpl.java:40)
        at javax.el.ELContext.convertToType(ELContext.java:325)
        at org.primefaces.model.JpaLazyDataModel.convertToType(JpaLazyDataModel.java:353)
        at org.primefaces.model.JpaLazyDataModel.applyFilters(JpaLazyDataModel.java:166)
        at org.primefaces.model.JpaLazyDataModel.count(JpaLazyDataModel.java:116)
```

But I do not believe that's the same as the original issue or what I've seen recently running in Wildfly (but I would need to confirm).

tandraschko:
But how should it ever work? Whats the usecase?
I think i need a reproducer

FlipWarthog:
The jpa branch of primefaces-test shows the issue (attached standalone .zip here). If you run that and try to filter Number Col with a non-numeric value, like "abc", it will throw the exception. This didn't work previously, but the fix for #9320 ([PR here](https://github.com/primefaces/primefaces/pull/9323)) addressed it, but the recent rework removed that fix. 

[PF9756-primefaces-test.zip](https://github.com/primefaces/primefaces/files/10787172/PF9756-primefaces-test.zip)


I just reproduced it in my own application running in Wildfly, and it looks like the Exception being thrown is a NumerFormatException wrapped in an ELException.

primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
```diff
@@ -366,13 +366,18 @@ protected Object convertToType(Object value, Class valueType) {
...
+        try {
+            // first convert the object to string
+            String stringValue = sourceConverter == null
+                    ? value.toString()
+                    : sourceConverter.getAsString(context, UIComponent.getCurrentComponent(context), value);
+
+            // now convert the string to the required target
+            return targetConverter.getAsObject(context, UIComponent.getCurrentComponent(context), stringValue);
+        }
+        catch (Exception e) {
```
melloware: So this should probably change to ELException?

melloware (changes requested):

primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
FlipWarthog: I can change it to ELException. Should I also include IllegalArgumentException, which NumberFormatException extends (among others that could come into play here)?

tandraschko:
i commited something, please try
ELContext must throw a ELException

this does however not solve a ConvertException from a converter, not sure if we should also catch it

PR #9833, opened 2023-02-25
#9825 AutoComplete disabled cursor

### LINKED ISSUE #9825: AutoComplete: disabled cursor is pointer
### Describe the bug

If the autocomplete is disabled, the wrong cursor is applied. I think there should be cursor default, not pointer. It's caused by this css class in [autcomplete.css](https://github.com/primefaces/primefaces/blob/24822d875d0eb8f658438dc4c74c56045b5d9faf/primefaces/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.css):
```
.ui-autocomplete {
    cursor: pointer;
}
```
I think this cursor: pointer property can be changed to default.

### Reproducer

1. Go to your autocomplete showcase
2. set class ui-state-disabled to Simple autocomplete's input element
3. mouse over

### Expected behavior

_No response_

### PrimeFaces edition

Community

### PrimeFaces version

12.0.0.

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3.18

### Java version

11

### Browser(s)

Chrome 110.0.5481.100

jepsar:
Fix #9825 

Was hoping to do this with CSS, but we needed the disabled class in the component wrapper.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/autocomplete/AutoCompleteRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/autocomplete/AutoCompleteRenderer.java
@@ -157,9 +157,14 @@ public class AutoCompleteRenderer extends InputRenderer {
         ResponseWriter writer = context.getResponseWriter();
         String clientId = ac.getClientId(context);
         boolean isDropdown = ac.isDropdown();
-        String styleClass = ac.getStyleClass();
-        styleClass = styleClass == null ? AutoComplete.STYLE_CLASS : AutoComplete.STYLE_CLASS + " " + styleClass;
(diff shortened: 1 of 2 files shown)


melloware (approved):

vonnai:
Hi, if the autocomplete is disabled, then the .ui-state-disabled class settings is overriden by:

https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/resources/META-INF/resources/primefaces/fileupload/fileupload.css

```
.ui-state-disabled .ui-state-disabled {
    opacity: 1;
    filter: alpha(opacity=100);
}
```

it's overridden only if the p:fileUpload is used in xhtml. It's caused by dual .ui-state-disabled class in the autocomplete span element and input element.

It's easy for me override .ui-state-disabled .ui-state-disabled, but I think there should not be this side effect. Please, check it, thanks.

PR #9838, opened 2023-02-26
Fix #9464: Cache don't print warning if disabled

### LINKED ISSUE #9464: Cache and move_scripts_to_bottom breaking the menu bar
### Describe the bug

Having move_scripts_to_bottom=true and caching a menu bar makes so that the menu items are not rendered anymore after the component is fetched from the cache.

### Reproducer

I checked out the PrimeFaces reproducer and put in the code of the menu bar showcase (https://www.primefaces.org/showcase/ui/menu/menubar.xhtml).

1. Click on any menu item except 'Logout'.
2. Click on 'Logout'.
3. Hover over the menu bar and no items are rendered.

With move_scripts_to_bottom=false, everything works as expected.

Here is the reproducer: [cache-move-scripts-to-bottom.zip](https://github.com/primefaces/primefaces/files/10141198/cache-move-scripts-to-bottom.zip)

### Expected behavior

The menu items should be rendered as it is the case with move_scripts_to_bottom=false.

### PrimeFaces edition

Community

### PrimeFaces version

12.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

8

### Browser(s)

_No response_

melloware:
Fix #9464: Cache don't print warning if disabled

### DIFF
--- a/primefaces-showcase/src/main/webapp/WEB-INF/tags/tabscode.xhtml
+++ b/primefaces-showcase/src/main/webapp/WEB-INF/tags/tabscode.xhtml
@@ -7,8 +7,7 @@
 
     <p:tabView var="entry" value="#{value}">
         <p:tab title="#{entry.title}" >
-            <p:cache region="sources" key="#{entry}"
-                     disabled="#{facesContext.application.projectStage eq 'Development'}">
+            <p:cache region="sources" key="#{entry}">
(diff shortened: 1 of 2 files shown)


Rapster:
Makes sense so +1

PR #9842, opened 2023-02-27
Fix #9282: SelectOneRadio pagedirection layout

### LINKED ISSUE #9282: Core: Deprecate `<table>` based layouts

melloware:
Added showcase example too:
image

Fix #9282

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/selectoneradio/SelectOneRadioRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/selectoneradio/SelectOneRadioRenderer.java
@@ -73,7 +73,7 @@ public class SelectOneRadioRenderer extends SelectOneRenderer {
             encodeLegacyTabularLayout(context, radio, layout);
         }
         else {
-            encodeResponsiveLayout(context, radio);
+            encodeResponsiveLayout(context, radio, layout);
         }
     }
 
@@ -92,7 +92,7 @@ public class SelectOneRadioRenderer extends SelectOneRenderer {
                 .finish();
     }
 
-    protected void encodeResponsiveLayout(FacesContext context, SelectOneRadio radio) throws IOException {
+    protected void encodeResponsiveLayout(FacesContext context, SelectOneRadio radio, String layout) throws IOException {
         ResponseWriter writer = context.getResponseWriter();
         String clientId = radio.getClientId(context);
         List<SelectItem> selectItems = getSelectItems(context, radio);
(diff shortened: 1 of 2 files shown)


primefaces/src/main/java/org/primefaces/component/selectoneradio/SelectOneRadioRenderer.java
```diff
@@ -119,9 +119,16 @@ protected void encodeResponsiveLayout(FacesContext context, SelectOneRadio radio
 
         Converter converter = radio.getConverter();
         String name = radio.getClientId(context);
-        int columns = radio.getColumns();
         String currentValue = ComponentUtils.getValueToRender(context, radio);
 
+        int columns = radio.getColumns();
+        if ("lineDirection".equals(layout)) {
```
jepsar: What happens is the label is long? It will probably start wrapping in the 1/12th column, right?

melloware: If gives it column `md-1` so column 13 should be wrapped on the next line.

jepsar: I'm not talking about 12+ items, but the length of a single label. So instead of "Option1" use "A label which you might expect in everyday use and might be longer".

melloware: ahh good point let me make a fix.

melloware: OK updated. The columns actually default to 12 unless you override it so it will be 12 but if you do `columns=2` with long labels it would make them both obviousle `md-6` making them much bigger.

jepsar: I think we should just drop using columns in that case.

melloware: I don't think we can else it won't be responsive anymore?

PR #9866, opened 2023-03-03
Fix #9863: ContextMenu cleanup DOM events on destroy

### LINKED ISSUE #9863: ContextMenu: cannot be conditionally toggled with ajax
### Describe the bug

If a context menu renderd value evaluates to false, it shouldn't be displayed even if it was displayed before.

### Reproducer

```xhtml
<h:form>
    <p:commandButton value="Toggle context menu" action="#{testView.toggleContextMenu()}" update="@form" />

    <p:outputLabel id="text" value="Context menu enabled: #{testView.enableContextMenu}"/>

    <p:contextMenu id="contextMenu2" for="text" rendered="#{testView.enableContextMenu}">
        <p:menuitem value="Homepage" url="https://www.primefaces.org" icon="pi pi-home"/>
    </p:contextMenu>
</h:form>
```

```java
@Data
@Named
@ViewScoped

    private boolean enableContextMenu;

    @PostConstruct  
    public void init() {
        enableContextMenu = true;
    }

    public void toggleContextMenu() {
        enableContextMenu = !enableContextMenu;
    }

}
```
To reproduce: open the page, press the button and right click the text: the context menu appears even if it shouldn't.

### Expected behavior

if rendered evaluates to false, context menu shouldn't be displayed

### PrimeFaces edition

Community

### PrimeFaces version

12.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3.19

### Java version

jdk-11.0.13

### Browser(s)

_No response_

melloware:
Fix #9863: ContextMenu cleanup DOM events on destroy

This adds the ability to remove all the subscribed DOM events from the document on destroy but in the scenario in the ticket I don't see why `destroy` is not being called?  cc @tandraschko  maybe i just don't understand enough about the `destroy` method.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/menu/menu.contextmenu.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/menu/menu.contextmenu.js
@@ -129,6 +129,46 @@ PrimeFaces.widget.ContextMenu = PrimeFaces.widget.TieredMenu.extend({
 
     /**
      * @override
+     * @inheritdoc
+     * @param {PrimeFaces.PartialWidgetCfg<TCfg>} cfg
+     */
+    refresh: function(cfg) {
+        this._cleanup();
+        this._super(cfg);
+    },
+
+    /**
+     * @override
+     * @inheritdoc
+     */
+    destroy: function() {
+        this._super();
+        this._cleanup();
+    },
+
+    /**
+    * Clean up this widget and remove events from the DOM.
+    * @private
+    */
+    _cleanup: function() {
+        if (this.cfg.target === undefined) {
+            var event = 'contextmenu.' + this.id + '_contextmenu';
+            $(document).off(event);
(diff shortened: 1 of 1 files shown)


tandraschko:
destroy is called when the widget  does not exist anymore after the ajax request
in case it still exists, refresh is called
didnt check your case above

melloware:
Yep it gets rendered="false" so it should be removed from the DOM "I think".

tandraschko:
thats weird then
can you check its called for older widgets?
hope nothing is broken in jquery ui core

melloware:
Yep I will do some research I just wanted to make sure I understood how its supposed to work. I will let you know what I find out.

melloware:
OK the issue is ContextMenu is attached to the body so it never calls destroy because 

```js
if (widget && widget.isDetached() === true) {
                        widget.destroy();
```

The isDetached will never be true as the Context Menu is always in the DOM.

PR #9881, opened 2023-03-07
Fix #9880: Password hide Edge icons

### LINKED ISSUE #9880: Password: Eye icon problem in MS Edge and IE
See: https://stackoverflow.com/questions/75661945/primefaces-password-togglemask-two-togglers-for-edge-browser

PrimeReact already fixed this: https://github.com/primefaces/primereact/issues/3756

melloware:
Fix #9880: Password hide Edge icons

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.css
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.css
@@ -829,6 +829,13 @@ div.ui-button, .ui-splitbutton {
     position: absolute;
 }
 
+.ui-password-masked.ui-password > input::-ms-reveal,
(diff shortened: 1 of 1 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.css
```diff
@@ -829,6 +829,11 @@ div.ui-button, .ui-splitbutton {
     position: absolute;
 }
 
+.ui-password::-ms-reveal,
```
jepsar: I think we should add a class if `toggleMask="true"` and only act if that class is present.

melloware:
Good suggestion

melloware:
We already had some classes to use so now it works properly

PR #9911, opened 2023-03-14
add display attribute to StaticMessage (#9908)

### LINKED ISSUE #9908: StaticMessage: support display property
### Description

`p:staticMessage` should support `display` property just like `p:message`

Implement rendering based on `display` property. Valid values would be: `both` (default), `text` and `icon`.
`tooltip` mode obviously cannot be implemented.

### Additional context

I'll work on this.

NicolaIsotta:
fixes #9908


melloware (approved):
LGTM but I will let others take a look

PR #9960, opened 2023-03-24
add disabled attribute to contextmenu (#9863)

### LINKED ISSUE #9863: ContextMenu: cannot be conditionally toggled with ajax

NicolaIsotta:
Advantages/differences over using `rendered`:

- ajax updates can target the context menu itself, no longer necessary to target the parent
- there's no need to call destroy on the widget (https://github.com/primefaces/primefaces/issues/9863#issuecomment-1454724751)

Fix #9863 #9926
CC @melloware

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/menu/menu.contextmenu.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/menu/menu.contextmenu.js
@@ -49,6 +49,7 @@
  * @prop {string} cfg.target Client ID of the target widget.
  * @prop {string} cfg.targetFilter Selector to filter the elements to attach the menu.
  * @prop {string} cfg.targetWidgetVar Widget variable of the target widget.
+ * @prop {boolean} cfg.disabled If true, prevents menu from being shown.
  */
 PrimeFaces.widget.ContextMenu = PrimeFaces.widget.TieredMenu.extend({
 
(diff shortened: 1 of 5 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/menu/menu.contextmenu.js
```diff
@@ -274,6 +275,10 @@ PrimeFaces.widget.ContextMenu = PrimeFaces.widget.TieredMenu.extend({
             return;
         }
 
+        if(this.cfg.disabled) {
```
jepsar: I would add this to the if in line 274

melloware (changes requested):
Looks good to me but I do like @jepsar suggestion.  I will review again once committed.

primefaces/src/main/resources/META-INF/resources/primefaces/menu/menu.contextmenu.js
NicolaIsotta: Yes, will do.

melloware (approved):

PR #9996, opened 2023-04-05
Fix #9988: SelectCheckBoxMenu register trigger for label

### LINKED ISSUE #9988: SelectCheckboxMenu: emptyLabel not considered as overlay-target and opens behind dialog
### Describe the bug

Dialog `bindEvents` checks for whether `mousedown` event target was `primefaces-overlay-target` and if not moves to top. This causes issues with clicks on `selectCheckboxMenu` `emptyLabel` which opens the menu/dropdown but dialog moves on top of it. When clicking on dropdown arrow or actual data label it works fine.

`SelectCheckboxMenu` `emptyLabel` fails this check on `line 379`:
https://github.com/primefaces/primefaces/blob/4e4430b40a2d1573137a975071edad1b3f953b66/primefaces/src/main/resources/META-INF/resources/primefaces/dialog/dialog.js#L377-L382
 

### Reproducer

Place selectCheckboxMenu inside popup and click on emptyLabel instead of dropdown arrow.
Menu opens but container dialog moves on top right after

### Expected behavior

Menu should open on top of the dialog

### PrimeFaces edition

Elite

### PrimeFaces version

12.0.3

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

8

### Browser(s)

_No response_

melloware:
Fix #9988: SelectCheckBoxMenu register trigger for label

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
@@ -148,9 +148,6 @@ PrimeFaces.widget.SelectCheckboxMenu = PrimeFaces.widget.BaseWidget.extend({
             this.bindEvents();
             this.bindKeyEvents();
 
-            //mark trigger and descandants of trigger as a trigger for a primefaces overlay
-            this.triggers.data('primefaces-overlay-target', true).find('*').data('primefaces-overlay-target', true);
-
             if (!this.cfg.multiple) {
                 this.label.attr('id', this.labelId);
                 this.keyboardTarget.attr('aria-labelledby', this.labelId);
@@ -239,6 +236,7 @@ PrimeFaces.widget.SelectCheckboxMenu = PrimeFaces.widget.BaseWidget.extend({
      */
     renderLabel: function() {
         if (!this.cfg.updateLabel) {
+            this.registerTrigger();
             return;
         }
         if (this.cfg.multiple) {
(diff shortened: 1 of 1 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectcheckboxmenu.js
```diff
@@ -147,9 +147,7 @@ PrimeFaces.widget.SelectCheckboxMenu = PrimeFaces.widget.BaseWidget.extend({
 
             this.bindEvents();
             this.bindKeyEvents();
-
-            //mark trigger and descandants of trigger as a trigger for a primefaces overlay
-            this.triggers.data('primefaces-overlay-target', true).find('*').data('primefaces-overlay-target', true);
+            this.registerTrigger();
```
arnis87: This seems redundant as now it is already done in renderLabel on line 159

melloware: But renderLabel has this
```js
        if (!this.cfg.updateLabel) {
            return;
        }
```

So if we are not updating the label the register never would get called.  But you need it called AFTER the label is created and rendered. So I guess I would have to do this.

```js
        if (!this.cfg.updateLabel) {
            this.registerTigger();
            return;
        }
```

melloware:
I updated the commit to move that registertrigger always inside the renderLabel.

arnis87 (approved):
Looks good to me

PR #10020, opened 2023-04-13
Update #9772: Sonacube fixes

melloware:
A few more Sonacube fixes


primefaces/src/main/java/org/primefaces/webapp/MultipartRequest.java
```diff
@@ -89,7 +89,7 @@ private String getItemString(FileItem item) {
             return (characterEncoding == null) ? item.getString() : item.getString(characterEncoding);
         }
         catch (UnsupportedEncodingException e) {
-            LOGGER.log(Level.SEVERE, "Unsupported character encoding " + getRequest().getCharacterEncoding(), e);
+            LOGGER.log(Level.SEVERE, e, () -> "Unsupported character encoding " + getRequest().getCharacterEncoding());
```
Rapster: Since you can't use interpolation and supplier, I would go with interpolation (just to be consistent with the other loggers). It's very likely that SEVERE will be enabled so no point to make "lazy"

PR #10023, opened 2023-04-14
Showcase: Add missing tabs to DF showcase

astappiev:
These files are used on showcase demo, but are missing in implementation tabs.

This PR aims to add all relevant sources to the showcases.

However, when I run it locally to verify that it works as expected, `viewResponsive.xhtml` appears empty, as well as other sources added, except `viewProductsLargerThanViewport.xhtml` it is rendered correctly. I have no clue why.

### DIFF
--- a/primefaces-showcase/src/main/webapp/ui/df/nested.xhtml
+++ b/primefaces-showcase/src/main/webapp/ui/df/nested.xhtml
@@ -53,4 +53,16 @@
 
     </ui:define>
 
+    <ui:define name="more-source-tabs">
+        <p:tab title="/ui/df/level1.xhtml">
+            <f:attribute name="flatten" value="true"/>
+        </p:tab>
+        <p:tab title="/ui/df/level2.xhtml">
+            <f:attribute name="flatten" value="true"/>
+        </p:tab>
+        <p:tab title="/ui/df/level3.xhtml">
+            <f:attribute name="flatten" value="true"/>
+        </p:tab>
+    </ui:define>
+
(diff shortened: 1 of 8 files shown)


melloware:
Hmmm what does `<f:attribute name="flatten" value="true"/>` do?  I noticed you have that on all the tabs?

astappiev:
What I understood from sources, it adds related beans used in the view to the showcase.

Without it, it still doesn't work, I tried.

melloware:
OK thanks for checking.

melloware:
OK I just made some changes to your branch.  Try again its working for me now.

melloware:
image

primefaces-showcase/src/main/webapp/ui/df/nested.xhtml
```diff
@@ -53,4 +53,16 @@
...
 
+    <ui:define name="more-source-tabs">
+        <p:tab title="/ui/df/level1.xhtml">
+            <f:attribute name="flatten" value="true"/>
+        </p:tab>
+        <p:tab title="/ui/df/level2.xhtml">
+            <f:attribute name="flatten" value="true"/>
+        </p:tab>
+        <p:tab title="/ui/df/level3.xhtml">
+            <f:attribute name="flatten" value="true"/>
```
melloware, line 64: You might want to check here I had to set flatten="false" on your others as it duplicated many of the Java tabs twice.

melloware:
Thanks all looks good and merged.

Issue #1248, opened 2016-03-17
p:imageCropper fails in portlets


stiemannkj1:
Steps to reproduce:
1. Add the following code to your portlet xhtml:
   
   ```
   <h:form>
   
       <h:graphicImage value="#{resource['example:example.png']}" />
       <p:imageCropper value="#{cropperView.croppedImage}" image="#{resource['example:example.png']}" initialCoords="10,10,85,25" />
       <h:panelGroup id="foo">
           <p:growl id="msgs" showDetail="true"/>
           <p:graphicImage id="cropped" rendered="#{not empty cropperView.newImageName}" name="#{cropperView.newImageName}.png" library="example" />
       </h:panelGroup>
   
       <p:commandButton value="Crop" action="#{cropperView.crop}" update="foo" icon="ui-icon-scissors" />
   
   </h:form>
   ```
2. Create **`CropperView.java`**:
   
   ```
   package com.liferay.faces.demos.bean;
   
   import java.io.File;
   import javax.faces.application.FacesMessage;
   import javax.faces.bean.ManagedBean;
   import javax.faces.context.ExternalContext;
   import javax.faces.context.FacesContext;
   import javax.imageio.stream.FileImageOutputStream;
   import org.primefaces.model.CroppedImage;
   
   @ManagedBean
   public class CropperView {
   
       private CroppedImage croppedImage;
   
       private String newImageName;
   
       public CroppedImage getCroppedImage() {
           return croppedImage;
       }
   
       public void setCroppedImage(CroppedImage croppedImage) {
           this.croppedImage = croppedImage;
       }
   
       public void crop() {
           if(croppedImage == null) {
               return;
           }
   
           setNewImageName(getRandomImageName());
           FacesContext facesContext = FacesContext.getCurrentInstance();
           ExternalContext externalContext = facesContext.getExternalContext();
           String newFileName = externalContext.getRealPath("") + File.separator + "WEB-INF" + File.separator + "resources" + File.separator + "example" + File.separator + getNewImageName() + ".png";
           System.out.println("!@#$" + newFileName);
   
           FileImageOutputStream imageOutput;
           try {
               imageOutput = new FileImageOutputStream(new File(newFileName));
               imageOutput.write(croppedImage.getBytes(), 0, croppedImage.getBytes().length);
               imageOutput.close();
           } catch (Exception e) {
[...]

PR #10029, opened 2023-04-16
Fix #9772 - SONAR: code quality for 13 release (ImageCropperRenderer)

### LINKED ISSUE #9772: SONAR: code quality for 13 release

Rapster:


primefaces/src/main/java/org/primefaces/component/imagecropper/ImageCropperRenderer.java
```diff
@@ -77,7 +85,7 @@ protected void encodeScript(FacesContext context, ImageCropper cropper) throws I
         String select = null;
 
         WidgetBuilder wb = getWidgetBuilder(context);
-        wb.initWithComponentLoad("ImageCropper", widgetVar, clientId, clientId + "_image")
+        wb.initWithComponentLoad("ImageCropper", widgetVar, clientId, image)
```
tandraschko, line 88: are you sure with this? this is by design to attach the onload to the image witself, not the component propably

Rapster: I'm sure: `String image = clientId + "_image";` 😉

primefaces/src/main/java/org/primefaces/component/imagecropper/ImageCropperRenderer.java
```diff
@@ -284,50 +228,38 @@ else if (stream != null) {
      * returns the value obtained by ResourceHandler.createResource().
      */
     private Resource getImageResource(FacesContext facesContext, ImageCropper imageCropper) {
```
tandraschko, line 230: i wonder in general if we need this method?
we could just evaluate the EL, without splitting and so on?

Rapster: No idea, related to #1248 and I never worked with portlets

tandraschko: TBH i would revert that
#{resource[test:test]} must just work, is specified by JSF and resolve internally to the correct portlet

Rapster: To discuss in another ticket then, here, the idea is just to clean up the code. But on a general note, I must admit it's kinda strange to see this only on ImageCropper, if it was really due to portlet environnement, I would picture that kinda change somewhere else

melloware (approved):
LGTM

Issue #10046, opened 2023-04-21
InputNumber: valueChange after updating without actually changing the value


jungm:
### Describe the bug

InputNumber fires valueChange event even though the value was not changed at all. This happens when it gains focus during an update. (e.g. tab from other input that caused the InputNumber to be updated)

### Reproducer

see: https://github.com/jungm/primefaces/commit/52683d1d9b76d60938a528955d1efe512602cbc4

### Expected behavior

valueChange event is not fired as the user made absolutely no inputs at all in the InputNumber

### PrimeFaces edition

None

### PrimeFaces version

13.0.0-SNAPSHOT

### JSF implementation

MyFaces

### JSF version

2.3

### Java version

11

melloware:
@jungm can you submit your Integration test please that you provided as a PR.

melloware:
@tandraschko i feel like this might be a new scenario as we already have this fix in the code:

https://github.com/primefaces/primefaces/blob/790a6e106d2413e32f429c622a484f2e51dd6eca/primefaces/src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js#L65-L66

And this fix:
https://github.com/primefaces/primefaces/blob/790a6e106d2413e32f429c622a484f2e51dd6eca/primefaces/src/main/resources/META-INF/resources/primefaces/inputnumber/1-inputnumber.js#L155-L158

To prevent too many onChange events from firing.

jungm:
@melloware created PR #10050

melloware:
OK did some debugging and submitted a fix.  Now letting the IT tests run.

melloware:
OK my fix and IT tests are working locally but still failing on the server I have no idea why.  @jungm please test my fix.

melloware:
MonkeyPatched reproducer showing the issue no longer happens.
[pf-10046.zip](https://github.com/primefaces/primefaces/files/11298778/pf-10046.zip)

Just run `mvn clean jetty:run`

jungm:
@melloware Can confirm this fixes #10046, tested in our app where we initially discovered that bug, the IT webapp and by running the ITs. No idea why the ITs fail in CI though

melloware:
Thanks for testing I am going to merge my fix with that test commented out for now.

PR #10050, opened 2023-04-21
Integration test for #10046

jungm:


melloware:
Thanks now I can look into this.

melloware:
Implemented in: https://github.com/primefaces/primefaces/pull/10051

PR #10051, opened 2023-04-21
Fix #10046: InputNumber do not fire change if value not changed

### LINKED ISSUE #10046: InputNumber: valueChange after updating without actually changing the value

melloware:
Fix #10046: InputNumber do not fire change if value not changed


melloware:
@jungm this is straneg this test case is passing for me locally with my fix.  Can you try my fixes to 1-inputnumber.js and see if it fixes your use case.  I don't understand why the test is failing here I can't get it to fail locally.

PR #10057, opened 2023-04-22
Fix #9969 - Core: Refactor Exporter - Part 2

### LINKED ISSUE #9969: Core: Refactor Exporter - Part 2
### Description

Lot of duplication and code is starting to be really messy

Cleaner code, less duplication in general but also  between Table/TreeTable exporter

### Additional context

#5053

### LINKED ISSUE #10027: DataExporter: XLS postProcessor fails
### Describe the bug

Based on the snapshot showcase: http://localhost:8080/showcase/ui/data/dataexporter/customizedDocuments.xhtml I have this exception when I click on XLS:

```
AVERTISSEMENT: /ui/data/dataexporter/customizedDocuments.xhtml @24,96 postProcessor="#{customizedDocumentsView.postProcessXLS}": java.lang.NullPointerException
javax.el.ELException: /ui/data/dataexporter/customizedDocuments.xhtml @24,96 postProcessor="#{customizedDocumentsView.postProcessXLS}": java.lang.NullPointerException
	at com.sun.faces.facelets.el.TagMethodExpression.invoke(TagMethodExpression.java:71)
	at org.primefaces.component.datatable.export.DataTableExcelExporter.postExport(DataTableExcelExporter.java:93)
	at org.primefaces.component.datatable.export.DataTableExcelXStreamExporter.postExport(DataTableExcelXStreamExporter.java:56)
	at org.primefaces.component.datatable.export.DataTableExporter.export(DataTableExporter.java:211)
	at org.primefaces.component.export.DataExporter.processAction(DataExporter.java:171)
	at javax.faces.event.ActionEvent.processListener(ActionEvent.java:72)
	at javax.faces.component.UIComponentBase.broadcast(UIComponentBase.java:495)
	at javax.faces.component.UICommand.broadcast(UICommand.java:211)
	at javax.faces.component.UIViewRoot.broadcastEvents(UIViewRoot.java:847)
	at javax.faces.component.UIViewRoot.processApplication(UIViewRoot.java:1396)
	at com.sun.faces.lifecycle.InvokeApplicationPhase.execute(InvokeApplicationPhase.java:58)
```

I believe this might be a regression of #9290. SXSSFSheet has some sort of cache on rows, once you exceed the initial capacity it flushes the first row (e.g 0), so whenever you do` sheet.getRow(0)` it returns null

I'll most probably fix it with a PR for #9969 (My Plan was to make it avaible for 14.0.0 but depending what you feel it could be merged for 13.0.0)  

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

13.0.0-SNAPSHOT

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

8

### Browser(s)

_No response_

Rapster:
This one is pretty big (but once you reviewed  `TableExporter`, the rest is boring) 😄 For review, I recommend to pull the branch as it's not easy to review through github diffs.

First, the most interesting part is `TableExporter`, basically it uses a template method pattern (so the whole export logic is there). It covers everything that can be possibly exported in DataTable. You can disable few things from there (see `TableExporter#supportedFacetTypes`)

Second,my plan was also to remove `TreeTable` code duplication but I realized exports does not keep tree structure, it's all flat, which makes me think not a lot of users are using that feature, whenever people will have feedbacks on this, I'll think about a Part 3... So far, it's all copy/paste from `DataTable` exporters

Besides that:
- `DataExporter#exporter` is deprecated, instead use `DataExporters#register()`
- New attribute `DataExporter#onRowExport` 

Fix #10027


melloware:
This looks like a really great refactoring.  And thanks for fixing the Integration Test!

melloware:
Testing your branch and in the Showcase the `Lazy" example when I export all Representatives are NULL?  Not sure if this is new issue or current showcase was like this.

```xml
	<customer>
		<name>Leon Z Rim</name>
		<country>Germany</country>
		<company>Rangoni Of Florence</company>
		<representative>null</representative>
	</customer>
	<customer>
		<name>Nicolas E Amigon</name>
		<country>Italy</country>
		<company>Chanay, Jeffrey A Esq</company>
		<representative>null</representative>
	</customer>
	<customer>
		<name>Nicolas U Vocelka</name>
		<country>United Kingdom</country>
		<company>Chanay, Jeffrey A Esq</company>
		<representative>null</representative>
	</customer>
```

Rapster:
Great, thanks for testing. The current showcase shows this:

```xml
<customer>
    <name>Ivar Q Oldroyd</name>
    <country>Australia</country>
    <company>Benton, John B Jr</company>
    <representative>nullAnna Fali</representative>
</customer>
<customer>
    <name>Arvin I Morasca</name>
    <country>Argentina</country>
    <company>Dorl, James J Esq</company>
    <representative>nullAsiya Javayant</representative>
</customer>
```

melloware:
Ahh yep i thought there was code in there to look for the printable value of a cell so the GraphicImage should be ignored and only the outputlabel picked up.

Rapster:
fixed

melloware:
Nice fix!

Rapster:
I'm thinking if it worth it to wait for 14.0.0 to merge this...

melloware:
Yep you can mark it for 14.0.   But do you have a short term fix for that other post processor bug?

Rapster:
I thought 13.0.0 would be released by now so I tagged it 14.0.0 but since it's not and considering how messy currently the code is, and it has that fix for #10027 we could merge it now instead of waiting for 14.0.0, and sooner we remove deprecated stuff better it is

melloware:
I like the idea but @tandraschko should weigh in.

tandraschko:
in general i would prefer 14.0 but i now pinged primtek about 13.0 release; TBH i hoped to see a release soon

melloware:
I am OK with it.  @tandraschko ?

melloware:
I am OK with it.  @tandraschko ?

tandraschko:
Will review next week

tandraschko:
Tbh i hope we can release next week a RC1 but we need to talk with primetek first

primefaces/src/main/java/org/primefaces/component/export/DataExporters.java
```diff
@@ -0,0 +1,97 @@
...
+
+        try {
+            return (Exporter<T>) exportClass.newInstance();
+        }
+        catch (ReflectiveOperationException e) {
+            throw new FacesException(e);
+        }
+    }
+
+    public static <T extends UIComponent, E extends Exporter<T>> void register(Class<T> targetClass, Class<E> exporter, String type) {
```
tandraschko, line 64: in theory its a bad pattern 
primefaces can likely be deployed in server/lib, where as the custom exporter class is likely inside war/lib

im sure we still have this in other places but we should NOT map custom classes in a static var
a solution is to map it on classloader level but that code isnt nice, so we should just store in the applicationMap which is bound to the webapp classloader

Rapster: But here, PF is not required to be deployed. Even if custom exporter is defined in an external lib, I don't think it'll be a problem is it?

tandraschko: if the external lib is deployed in webapp, and PF server/lib, it still is
using applicationMap is preferred to avoid such problems

Rapster: I'm not familiar with such scenarios, so if you have an example for applicationMap, I'll do the modifications required

tandraschko: facesContext.getExternalContext().getApplicationMap()

if you check e.g. myfaces, thats always used when storing webapp-associated data

tandraschko: or even better: PrimeApplicationContext

Rapster: Sounds good, i'll make the changes

tandraschko: will also do it for BV constraint mapping now: https://github.com/primefaces/primefaces/issues/10133

primefaces-showcase/src/main/java/org/primefaces/showcase/view/data/dataexporter/CustomizedDocumentsView.java
```diff
@@ -122,10 +120,10 @@ public void setService(ProductService service) {
         this.service = service;
     }
 
-    public void postProcessXLS(Object document) {
+    public void onRowExportXLS(Object document) {
```
tandraschko, line 123: the method originally was to hightight the header after all processing was done
now you hightlight the all rows, right?

shouldnt we create a different method/example here?

Rapster: Yeah i'm gonna change that, so it looks like strippedRows showcase

primefaces/src/main/java/org/primefaces/component/export/DataExporters.java
Rapster: Done with the changes, let me know what you think

Rapster: Is this okay or need more to time to review?

melloware:
Looks like there is a merge conflict?

melloware:
Now that RC1 is out should we consider merging this for RC2?  I am open to whatever you guys want to do.

tandraschko:
will review this week again

PR #10062, opened 2023-04-24
Fix #10046: InputNumber integration test

### LINKED ISSUE #10046: InputNumber: valueChange after updating without actually changing the value

melloware:
Just trying this now that the code is merged to see if the result is any different.

### DIFF
--- a/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/inputnumber/InputNumber007Test.java
+++ b/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/inputnumber/InputNumber007Test.java
@@ -39,9 +39,9 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 
 public class InputNumber007Test extends AbstractPrimePageTest {
 
-    // @Test
-    // @Order(1)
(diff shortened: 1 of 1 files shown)


melloware:
@christophs78 no rush on this but neither @jungm or myself can get this test to fail locally and he has verified the code works in his real application so I know the fix is right I am just stumped as to why the test fails?

christophs78:
Hm. Still the old JS code running inside my browser. --> First need to fix this.

christophs78:
There seem´s to be some aggregation issue with inputnumber.js:

1-inputnumber.js is copied into inputnumber.js multiple times. First your new code, afterwards the old (not fixed) code.

image
image

christophs78:
Some `clean install` later it work´s on my dev-machine. Let´s look what Github Actions thinks ...

melloware:
yeah i thought our Selenium tests always worrked with fresh browser cache

Issue #9040, opened 2022-07-28
DataTable: resetting FilterMetal#filterValue to null wont be applied to filter component


tandraschko:
### Describe the bug

the old value is still displayed but the filter applied correctly

### PrimeFaces version

12.0.0-RC2

Issue #8992, opened 2022-07-12
Column: FilterValue Mutator Not Called


hink084:
### Describe the bug

The defined Column filterValue within a DataTable does not store the value within the backing bean when processed - the mutator method is not called.  This behavior changed when migrating from PrimeFaces v8 to PrimeFaces v10. As of 7/12/2022 this issue is still present within the PrimeFaces 12.0-SNAPSHOT.

There is a workaround which is to manually define the filter utilizing the filter facet.  You can also bind the DataTable and access the filterValues via the DataTable API but that method is not shown here.

```
<h:form id="frmTest">
    <h2>1) Filter the Animal and Sound columns: {dog, woof}</h2>
    <p:dataTable id="dt_test" value="#{testView.items}" var="item" widgetVar="wgt_test">
        <!-- Broken FilterValue -->
        <p:column filterBy="#{item.animal}"
                  filterValue="#{testView.animalFilter}"
                  headerText="Animal"
                  id="col_animal">
            #{item.animal}
        </p:column>

        <!-- Workaround Filter -->
        <p:column filterBy="#{item.sound}"
                  headerText="Sound"
                  id="col_sound">

            <f:facet name="filter">
                <p:inputText value="#{testView.soundFilter}" onkeyup="PF('wgt_test').filter();"/>
            </f:facet>

            #{item.sound}
        </p:column>
   </p:dataTable>
   <br/><br/>

   <h2>2) Click the Button</h2>
   <p:commandButton id="btn_show"
                    partialSubmit="true"
                    process="@this, dt_test"
                    update="animalFilter, soundFilter"
                    value="Show Filter Values"/>
    <br/><br/>

    <h2>3) Observe Values</h2>
    Animal Filter Value: <h:outputText id="animalFilter"
                                       value="#{not empty testView.animalFilter ? testView.animalFilter : '- Empty -'}"/>
    <br/>
    Sound Filter Value: <h:outputText id="soundFilter"
                                      value="#{not empty testView.soundFilter ? testView.soundFilter : '- Empty -'}"/>
    <br/><br/>

    <h4>Column filterValue mutator was called in PrimeFaces v8 but not in PrimeFaces v10+</h4>
</h:form>
```

### Reproducer

1. Clean/Build the attached reproducer
2. Filter the DataTable Columns (Observe the console for which mutators are called during filtering)
3. Click the CommandButton (Observe the console for which mutators are called during processing)
4. Observe the Filter Values display on the screen
[...]

PR #10068, opened 2023-04-26
Fix #10067 - DataTable: set default filter either on UIColumn#filterValue or custom facet value

### LINKED ISSUE #10067: DataTable: set default filter either on UIColumn#filterValue or custom facet value
### Description

Based on this: https://github.com/primefaces/primefaces/issues/9349#issuecomment-1520843859

> In regards to when or when not to utilize the filterValue attribute...I was under the impression that the attribute was to be used if a filter was not explicitly defined in a filter facet via <f:facet name="filter">. Or in other words, the filterValue attribute was to be ignored if I did define a filter facet.

As there are two potential ways to define default value, users are confused which one to used, but more recently users tend to use custom facet filter value due to a "regression"

My solution is here https://github.com/primefaces/primefaces/issues/9349#issuecomment-1524099844

>So, maybe the best way to go about it is to support both by simply doing a coalesce 😄 I can propose a fix, but maybe in another issue, your issue is pretty clear as to what is not working /confusing

support both `UIColumn#filterValue` and custom facet value, `UIColumn#filterValue` being priority number 1

### Additional context

_No response_

Rapster:


melloware:
Looks like IT's are failing.

Rapster:
Yeah I noticed, funny thing is, it only works in 4.0

tandraschko:
yeah, i only copied some tests to test some basic functionality (startup/shading stuff)

Rapster:
I might need your insights on this, i'm not sure what's wrong... @melloware @christophs78  Another thing, how do I debug PF components, I can debug the IT itself but that's it, probably because they're running on another port?

melloware:
Yep that is what I usually do run it as a webapp and debug.

Rapster:
Ok so I think this ticket fixes a few things:

- #10067 
- similar to #8992 but the default value is a collection of complex objects (e.g Representative)
- #9349 like I mentionned in the original ticket (and restore fix for #9040)... trying to workaround the provided API will only lead to potential bugs

melloware:
nice!

melloware (approved):

christophs78:
Glad to see you got it working.
For the case of some open questions regarding debugging Integrationtests we may do a short videomeeting at some point in the future. (~15 min) There are no limitations. You can use the jetty-plugin to start integration-test-web-application like the showcase. Or you can run the integration-tests, set breakpoint´s into to integration-test-java-code, set breakpoint into the javascript code (do not use headless for this) and/or set breakpoint´s into primefaces-java-code, ...

PR #10075, opened 2023-04-28
Fix #10046: InputNumber must delay registering events until CSP is done

### LINKED ISSUE #10046: InputNumber: valueChange after updating without actually changing the value

### LINKED ISSUE #10074: IT: Nightly UpdateForm001Test failure
On the overnight full tests getting..

`UpdateForm001Test.testUpdateForm:44 » Timeout Timeout while waiting for AJAX complete!`

melloware:
Fix #10046: InputNumber must delay registering events until CSP is done
Fix #10074

Basically because CSP Register happens AFTER the widget init the `wrapEvents` method is not doing what is intended when running in CSP mode.

### DIFF
--- a/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/inputnumber/InputNumber007Test.java
+++ b/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/inputnumber/InputNumber007Test.java
@@ -39,9 +39,9 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 
 public class InputNumber007Test extends AbstractPrimePageTest {
 
-    // @Test
-    // @Order(1)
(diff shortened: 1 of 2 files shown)


melloware:
cc @christophs78 lets see how this goes....

melloware:
OK so this this looks like InputNumber007 test is now fixed but the UpdateFormOne still fails.  I will look into that next

melloware:
OK I think I nailed the both. It was a regression.  Running now...

melloware:
Yeah!  All integration tests old and new are passing now!

Issue #4106, opened 2018-10-03
Add model attribute to SplitButton


mertsincan:
A menu model instance to create the items of splitButton programmatically.

Rapster:
Great job! Is it possible to create the default action through the model instance?

Rapster:
Might close #766 and #4010

melloware:
@Rapster I agree it should definitely close those two with links to this ticket.

mertsincan:
@Rapster 

> Is it possible to create the default action through the model instance?

No, to create menuitems and submenus in panel

Issue #4105, opened 2018-10-03
Add filter support to SplitButton


mertsincan:
Added these attributes;
- **filter** -> Displays an input filter for the list
- **filterMatchMode** -> Match mode for filtering, valid values are startsWith (default), contains, endsWith and custom.
- **filterFunction** -> Client side function to use in custom filterMatchMode.
- **filterPlaceholder** -> Watermark displayed in the filter input field before the user enters a value in an HTML5 browser.

Issue #4104, opened 2018-10-03
Add p:submenu support to SplitButton


mertsincan:
Menuitems can be grouped with p:submenu.

PR #10089, opened 2023-05-06
Fix #10086 OverlayPanel improved SplitButton support

### LINKED ISSUE #10086: SplitButton: allow to open an OverlayPanel instead of MenuItems
### Description

Now you can only use MenuItems.

It would be great if you could link an OverlayPanel to a SplitButton. This would allow you to do something like GitLab's "create branch from issue":

Screenshot 2023-05-05 at 10 27 30

So, the button is the quick action, and the dropdown opens the OverlayPanel with advanced options. 

### Additional context

_No response_

jepsar:
Fix #10086

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/badge/BadgeRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/badge/BadgeRenderer.java
@@ -47,7 +47,7 @@ public class BadgeRenderer extends CoreRenderer {
             IOBiConsumer<FacesContext, T> contentRenderer, T component) throws IOException {
         BadgeModel badgeModel = Badge.getBadgeModel(badge);
         if (badgeModel != null) {
-            BadgeRenderer badgeRenderer = Badge.getDefaultRenderer(context);
+            BadgeRenderer badgeRenderer = new Badge().getRenderer();
             badgeRenderer.encodeOverlayBegin(context, null);
             contentRenderer.accept(context, component);
             badgeRenderer.encode(context, null, badgeModel, false);
@@ -61,7 +61,7 @@ public class BadgeRenderer extends CoreRenderer {
     public static void encode(FacesContext context, Object badge) throws IOException {
         BadgeModel badgeModel = Badge.getBadgeModel(badge);
         if (badgeModel != null) {
-            BadgeRenderer badgeRenderer = Badge.getDefaultRenderer(context);
+            BadgeRenderer badgeRenderer = new Badge().getRenderer();
             badgeRenderer.encode(context, null, badgeModel, false);
         }
     }
--- a/primefaces/src/main/java/org/primefaces/component/badge/BadgeRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/badge/BadgeRenderer.java
@@ -47,7 +47,7 @@ public class BadgeRenderer extends CoreRenderer {
             IOBiConsumer<FacesContext, T> contentRenderer, T component) throws IOException {
         BadgeModel badgeModel = Badge.getBadgeModel(badge);
         if (badgeModel != null) {
-            BadgeRenderer badgeRenderer = new BadgeRenderer();
+            BadgeRenderer badgeRenderer = Badge.getDefaultRenderer(context);
             badgeRenderer.encodeOverlayBegin(context, null);
             contentRenderer.accept(context, component);
             badgeRenderer.encode(context, null, badgeModel, false);
@@ -61,7 +61,7 @@ public class BadgeRenderer extends CoreRenderer {
     public static void encode(FacesContext context, Object badge) throws IOException {
         BadgeModel badgeModel = Badge.getBadgeModel(badge);
         if (badgeModel != null) {
-            BadgeRenderer badgeRenderer = new BadgeRenderer();
+            BadgeRenderer badgeRenderer = Badge.getDefaultRenderer(context);
             badgeRenderer.encode(context, null, badgeModel, false);
         }
     }
--- a/primefaces/src/main/java/org/primefaces/component/badge/BadgeRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/badge/BadgeRenderer.java
@@ -47,7 +47,7 @@ public class BadgeRenderer extends CoreRenderer {
             IOBiConsumer<FacesContext, T> contentRenderer, T component) throws IOException {
         BadgeModel badgeModel = Badge.getBadgeModel(badge);
         if (badgeModel != null) {
-            BadgeRenderer badgeRenderer = new Badge().getRenderer();
+            BadgeRenderer badgeRenderer = Badge.create(context).getRenderer();
             badgeRenderer.encodeOverlayBegin(context, null);
             contentRenderer.accept(context, component);
             badgeRenderer.encode(context, null, badgeModel, false);
@@ -61,7 +61,7 @@ public class BadgeRenderer extends CoreRenderer {
     public static void encode(FacesContext context, Object badge) throws IOException {
         BadgeModel badgeModel = Badge.getBadgeModel(badge);
         if (badgeModel != null) {
-            BadgeRenderer badgeRenderer = new Badge().getRenderer();
+            BadgeRenderer badgeRenderer = Badge.create(context).getRenderer();
             badgeRenderer.encode(context, null, badgeModel, false);
         }
     }
--- a/primefaces/src/main/java/org/primefaces/component/overlaypanel/OverlayPanel.java
+++ b/primefaces/src/main/java/org/primefaces/component/overlaypanel/OverlayPanel.java
@@ -28,6 +28,8 @@ import java.util.Map;
 import javax.faces.application.ResourceDependency;
 import javax.faces.context.FacesContext;
 import javax.faces.event.BehaviorEvent;
+import static org.primefaces.component.overlaypanel.OverlayPanelBase.COMPONENT_FAMILY;
+import static org.primefaces.component.overlaypanel.OverlayPanelBase.DEFAULT_RENDERER;
 
 import org.primefaces.util.ComponentUtils;
 import org.primefaces.util.MapBuilder;
@@ -52,6 +54,10 @@ public class OverlayPanel extends OverlayPanelBase {
             .put("loadContent", null)
             .build();
     private static final Collection<String> EVENT_NAMES = BEHAVIOR_EVENT_MAPPING.keySet();
+
+    public static OverlayPanelRenderer getDefaultRenderer(FacesContext context) {
+        return (OverlayPanelRenderer) context.getRenderKit().getRenderer(COMPONENT_FAMILY, DEFAULT_RENDERER);
+    }
 
     @Override
     public Map<String, Class<? extends BehaviorEvent>> getBehaviorEventMapping() {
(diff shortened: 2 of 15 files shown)


primefaces/src/main/java/org/primefaces/component/splitbutton/SplitButtonRenderer.java
```diff
@@ -98,10 +98,8 @@ protected void encodeMarkup(FacesContext context, SplitButton button) throws IOE
         }
 
         encodeDefaultButton(context, button, buttonId);
-        if (button.getElementsCount() > 0) {
```
melloware, line 101: By removing this check it will now always render the menu?  Or do we need to add more short circuits inside the encodeMenuIcon and encodeMenu methods?

melloware (changes requested):

primefaces/src/main/java/org/primefaces/component/splitbutton/SplitButtonRenderer.java
jepsar: Yes, it will always render the icon (drop down button) and the menu. I don't need the menu, and probably the code is now also toggling an empty div, but this was the least intrusive fix I could come up with.

melloware: Understood!  Thanks for the explanation.

melloware (approved):

primefaces/src/main/java/org/primefaces/component/splitbutton/SplitButtonRenderer.java
Rapster: -1 If you look at the history it revert 3 tickets, also if no element are present in splitButton it should be displayed as a simple button

melloware: If it reverts 3 tickets we need integration tests for those 3 tickets and this new one.

jepsar: It currently does not display as a simple button properly when there are no items (but that's a theme issue).

jepsar: So https://github.com/primefaces/primefaces/commit/07412082bb0218784032a022f5cf49fec0cc8050 solved 3 issues, but I'm pretty sure that removing the `if` won't regress https://github.com/primefaces/primefaces/issues/4104, https://github.com/primefaces/primefaces/issues/4105 and https://github.com/primefaces/primefaces/issues/4106

Rapster: Ok, I see that very same line had replaced what I did something in the past (see #2690) so there must be a reason for it. Turns out https://github.com/primefaces/primefaces/commit/07412082bb0218784032a022f5cf49fec0cc8050 introduced already a regression on #2690 (what a mess...)

jepsar: Alright. I can add a boolean attribute for it. But there also will need to be a change in the theme in that case. I'm not sure it only is the border radius, but at least the radius is 0 at the right side of the button (did not check RTL).

Let's add an IT for #2690 as well.

jepsar: Yeah, this shows the dropdown currently:
```java
<p:splitButton id="noRenderedChildren" value="No rendered children">
    <p:menuitem value="Rendered false" rendered="false"/>
</p:splitButton>
```

Screenshot 2023-05-09 at 16 41 15

Should I create a separate ticket for this? I will fix this in this PR though.

I will create an issue for the border radius.

Rapster: If you have plans to fix also the css part (or that's something up to PrimeTek?) maybe in another issue then, otherwise I don't mind if you fix it in this PR

jepsar: CSS must be fixed by PrimeTek (it's part of the theme), so that will need an issue for sure.

melloware: I haven't looked but now that this exists can we start fixing the CSS ourselves?  https://github.com/primefaces/primefaces-sass-theme

jepsar: Ha, yeah, could be. I could open a ticket there and see if I can provide a PR. And we need a readme.

melloware: Yep and I was gonna make a pom.xml that uses Dart SASS and minifies the CSS.

primefaces/src/main/resources/META-INF/primefaces-p.taglib.xml
```diff
@@ -27174,6 +27174,14 @@
             <required>false</required>
             <type>java.lang.String</type>
         </attribute>
+        <attribute>
+            <description>
+                <![CDATA[If true, the dropdown button will always be displayed (even if there are no menuitems). Default is 'false'.]]>
+            </description>
+            <name>allwaysShowDropdown</name>
```
melloware: its `always` not `allways`

melloware (changes requested):

primefaces/src/main/resources/META-INF/primefaces-p.taglib.xml
jepsar: Doh!

melloware (approved):

melloware:
Confirmed with Cagatay that that PrimeFaces SASS project we can start updating and it generates the themes so we can fix all the themes issues ourselves.

jepsar:
> whats the reason to render a splitbutton anyway without dropdown?

@Rapster mentioned it in https://github.com/primefaces/primefaces/pull/10089#discussion_r1187269402. This was introduced in #2690. It makes sense when you have menu items in the button which are conditionally rendered (based on permission).

tandraschko:
yep, understood

i like the change
i just  think about if overlayPanel could notifiy the splitButton on java-side, to render the dropdown (instead of using a manual attribute)
maybe we a prerender event

jepsar:
I think I can remove the attribute if we implement the renderer of the SplitButton to handle:
```xhtml
<p:splitButton value="OverlayPanel">
    <p:overlayPanel>...</p:overlayPanel>
</p:splitButton>
```
Like this we can also eliminate the `for` attribute of the OverlayPanel.

melloware:
GitHub Actions is struggling right now.  GitHub marked it as "degraded". https://www.githubstatus.com/

jepsar:
There is one test failing, but it's not using SplitButton nor OverlayPanel, and it passes with other Java versions, so I don't think it's related to this PR.

melloware:
I just re-ran your build and now its passing.

melloware (approved):

primefaces/src/main/java/org/primefaces/component/splitbutton/SplitButton.java
```diff
@@ -123,6 +125,7 @@ public List<MenuElement> getElements() {
         }
         else {
             return getChildren().stream()
+                        .filter(UIComponent::isRendered)
```
tandraschko: please check isInstance first, should be much faster as isRendered

jepsar: Ha, I was doubting in which order to do it. Will change it.

primefaces/src/main/java/org/primefaces/component/splitbutton/SplitButtonRenderer.java
```diff
@@ -64,6 +66,11 @@ public void decode(FacesContext context, UIComponent component) {
                 component.queueEvent(new ActionEvent(component));
             }
         }
+
+        OverlayPanel customOverlay = button.getCustomOverlay();
+        if (customOverlay != null) {
+            new OverlayPanelRenderer().decode(context, customOverlay);
```
tandraschko: get the Renderer instance via RenderKit, a user can overwrite it via faces-config

primefaces/src/main/java/org/primefaces/component/splitbutton/SplitButtonRenderer.java
```diff
@@ -98,12 +105,18 @@ protected void encodeMarkup(FacesContext context, SplitButton button) throws IOE
...
+        if (customOverlay != null || button.getElementsCount() > 0) {
             encodeMenuIcon(context, button, menuButtonId);
             encodeMenu(context, button, menuId);
         }
 
         writer.endElement("div");
+
+        if (customOverlay != null) {
+            customOverlay.setFor(clientId);
+            new OverlayPanelRenderer().encodeEnd(context, customOverlay);
```
tandraschko: same here

jepsar:
Alright. I think we are good to merge now.

primefaces/src/main/java/org/primefaces/component/overlaypanel/OverlayPanel.java
```diff
@@ -53,6 +53,10 @@ public class OverlayPanel extends OverlayPanelBase {
             .build();
     private static final Collection<String> EVENT_NAMES = BEHAVIOR_EVENT_MAPPING.keySet();
 
+    public static OverlayPanelRenderer getDefaultRenderer(FacesContext context) {
+        return (OverlayPanelRenderer) context.getRenderKit().getRenderer(COMPONENT_FAMILY, DEFAULT_RENDERER);
```
tandraschko: I would not make it static
make it non static and call it getRenderer() which does:
getFacesContext().getRenderKit().getRenderer(getFamily(), getRendererType());

primefaces/src/main/java/org/primefaces/component/badge/BadgeRenderer.java
```diff
@@ -47,7 +47,7 @@ public static <T extends UIComponent> void encodeOverlayed(FacesContext context,
             IOBiConsumer<FacesContext, T> contentRenderer, T component) throws IOException {
         BadgeModel badgeModel = Badge.getBadgeModel(badge);
         if (badgeModel != null) {
-            BadgeRenderer badgeRenderer = Badge.getDefaultRenderer(context);
+            BadgeRenderer badgeRenderer = new Badge().getRenderer();
```
tandraschko: i dont understand this in general
Badge#getRenderer should just return BadgeRenderer, which is equals to "this"?

tandraschko: nevermind
that code it a bit weird with statics, non-statics, functional and so on

we need at least use #createComponent API instead of new Badge

later we may refactor that part a bit

jepsar: Yeah, I just came back to this because we also constructed a renderer there. The thing with badges is that you want to be able to render a badge when you only have a badge model object. For example with a SpeedDial. We can improve this later.

PR #10092, opened 2023-05-08
Fix #9213: DatePicker prevent propagation on input click

### LINKED ISSUE #9213: DatePicker: Closing Itself Immediately
I'm using a DatePicker in a Primefaces dialog. The code for the field is 
```
        <p:datePicker id="date"
                      mask="true"
                      selectOtherMonths="true"
                      monthNavigator="true"
                      yearNavigator="true"
                      showButtonBar="true"
                      placeholder="mm/dd/yyyy"
                      pattern="MM/dd/yyyy"
                      value="#{beanName.property">
            <f:convertDateTime type="date" pattern="MM/dd/yyyy" />
            <f:validateBean for="date" validationGroups="group.package.and.name"/>
        </p:datePicker>
```

When the user clicks on the field it opens the datePicker, showing the current month and the dates. But the datePicker disappears when the mouse button is released. As long as the button is held down, the datePicker is visible. But it disappears as soon as the button is released. 
If the user clicks on the field a second time the datePicker will open and stay open. But they have to click twice every time.

I'm looking for a solution that will cause the datePicker to only need one click. It works normally on other pages in the same application and I haven't been able to identify any differences between this field and others. Can anyone offer a suggestion?

melloware:
Fix #9213: DatePicker prevent propagation on input click

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
@@ -1799,7 +1799,7 @@
             });
         },
 
-        onInputClick: function (event) {
+        onInputClick: function (event) {    
             if (this.documentClickListener) {
                 this.datepickerClick = true;
             }
@@ -2166,6 +2166,8 @@
                     },
                     onEntered: function() {
                         if (!$this.options.touchUI) {
+                            $this.datepickerClick = true;
+                            setTimeout(function () {$this.datepickerClick = false;}, 200);
                             $this.bindDocumentClickListener();
                             $this.bindWindowResizeListener();
 
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
@@ -1799,7 +1799,7 @@
             });
         },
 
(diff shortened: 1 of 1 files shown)


melloware:
Integration tests failing with this fix.

melloware:
@jepsar what is happening is when click the mouse using the mousepad as the popup is popping up the documentClickListener is also firing and since the value is `false` its immediately closing as if the user pressed a button somewhere outside the datepicker.  What this does is default the value to block outside clicks until the animation of bringing the popup is complete.

primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -2166,6 +2166,8 @@
                     },
                     onEntered: function() {
                         if (!$this.options.touchUI) {
+                            $this.datepickerClick = true;
+                            setTimeout(function () {$this.datepickerClick = false;}, 200);
```
jepsar, line 2170: I guess you probably already looked at a better way to fix this?

melloware: I have dug around and this is the best I could come up with since it only happens strangely with a mousepad and its a timing issue related to the animation etc.  There are so many complex events going on that this was the only thing I can do to get it working without breaking other things!

Issue #10112, opened 2023-05-11
TextEditor: improve blur and change events


jungm:
- blur event is triggered when interacting with the toolbar, imho this should not happen because the user is still interacting with the same TextEditor and doesn't even notice that it lost focus for a short moment by for example making text bold
- change event is called on every keystroke, making it behave more like an `input` event than a `change` event. I think we should align this with how e.g. InputText behaves

PR #10124, opened 2023-05-12
#10112 fix IT on Firefox

christophs78:
CC @jungm - better idea?


jungm:
@christophs78 seems like firefox behaves a bit differently to chromium/safari when sending keys to a WebElement that is not in focus. Clicking it before sending keys (to focus it) seems to align it to how chromium/safari behave

PR #10130, opened 2023-05-14
Fix #10129 - DataTable: detects automatically selectionMode based on selection VE/value

### LINKED ISSUE #10129: DataTable: detects automatically selectionMode based on selection VE/value
### Description

Just like `DataTable#lazy`, it should be possible to determine `selectionMode` value (single, mutiple) based on selection VE or value. If a `Collection`, we can assume it's multiple selection, otherwise single

**Automatic single selection** (`selectedProduct` being a single object):

```xml
<p:dataTable var="product" value="#{dtSelectionView.products}" selection="#{dtSelectionView.selectedProduct}">

</p:dataTable>
```

Will enable single selection

**Automatic multiple selection**  (`selectedProducts` being a collection of object):
```xml
<p:dataTable var="product" value="#{dtSelectionView.products}" selection="#{dtSelectionView.selectedProducts}">

</p:dataTable>
```

Will enable multiple selection

Rapster:
Fix #10129

### DIFF
--- a/docs/13_0_0/components/datatable.md
+++ b/docs/13_0_0/components/datatable.md
@@ -92,7 +92,7 @@ DataTable displays data in tabular format.
 | scrollWidth               | null               | Integer          | Scroll viewport width.
 | scrollable                | false              | Boolean          | Makes data scrollable with fixed header.
 | selection                 | null               | Object           | Reference to the selection data.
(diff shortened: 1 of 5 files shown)


tandraschko:
+1 If builds are running

Rapster:
If you're okay having this in 13.0.0, I can update docs

tandraschko:
yep, why not
it shouldnt break anything

PR #10134, opened 2023-05-15
Fix IT #10112 for firefox

### LINKED ISSUE #10112: TextEditor: improve blur and change events

jungm:
Fixes #10112 for firefox (see #10124 as well)

### DIFF
--- a/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/texteditor/TextEditor004Test.java
+++ b/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/texteditor/TextEditor004Test.java
@@ -60,17 +60,13 @@ public class TextEditor004Test extends AbstractPrimePageTest {
         WebElement quillEditor = page.editor.findElement(By.className("ql-editor"));
 
         // Act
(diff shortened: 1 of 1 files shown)


jungm:
General question I'm thinking of right now, shouldn't firefox (and possibly safari) ITs also be run on PRs? (at least one, don't need to run all jsf/java combinations in my opinion)

melloware:
Great question.  We used to try and run them all on every PR but there are too many combinations so we decided Chrome would be our PR run as its the "most stable" Selenium webdriver browser.

Safari is an absolute mess and @christophs78 and I were always chasing differences in safari and adding hacks to the tests that we kind of gave up as it was so much effort.  We current have the Safari tests completely disabled.

So at night we run the full suite of Firefox and Chrome as it takes over an hour.

PR #10152, opened 2023-05-18
Fix #9882: Dialog width for saveState account for scrollbar.

### LINKED ISSUE #8989: Dialog content with scrollbar does not resize correctly when dialog resized
### Describe the bug

If a dialog content has scroll bar, there will be a gap between dialog content and dialog border when dialog is resized.

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

_No response_

### PrimeFaces version

11.0.0

### Theme

Saga

### JSF implementation

_No response_

### JSF version

_No response_

### Browser(s)

Chrome 103

### LINKED ISSUE #9882: Dialog: resizing horizontally causes a gap between scrollbar and dialog right border
Resizing dialog horizontally causes a gap between scrollbar and dialog right border.

http://www.primefaces.org:8080/showcase/ui/overlay/dialog.xhtml

Steps to reproduce:
1. Show Model dialog (or others)
2. Resize dialog height to make vertical scroll show.
3. Then resize the dialog horizontally, a gap between the scroll and dialog right border will show.

melloware:
Fix #9882: Dialog width for saveState account for scrollbar.
Fix #8989 8989

Because `jquery.width()` is accounting for the scrollbar this content is constanly shrinking.  So we must try and use the already calculated `style="width: 308px"` to use 308 and only fall back to `content.width()` if that is null or 0.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/dialog/dialog.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/dialog/dialog.js
@@ -712,7 +712,7 @@ PrimeFaces.widget.Dialog = PrimeFaces.widget.DynamicOverlayWidget.extend({
         this.state = {
             width: this.jq.width(),
             height: this.jq.height(),
-            contentWidth: this.content.width(),
+            contentWidth: parseInt(this.content[0].style.width) || this.content.width(),
             contentHeight: this.content.height()
         };
 


patzi:
But isn't this a jQuery UI upstream bug that should be fixed in 1.13.3 see https://github.com/jquery/jquery-ui/issues/2083 so it should be fixed with updating jQuery UI in PrimeFaces?

melloware:
Jquery ui 1.13.3 does not exist yet.  And they don't seem to release it on any kind of schedule.   The fix I made works with or without jquery 1.13.3

patzi:
Sorry, my bad, i had read the jquery issue wrong and thought it had been released already. Anyways thanks for your fix and work.

PR #10153, opened 2023-05-18
Fix #10150: Log warning on missing locale-XX.js file

### LINKED ISSUE #10150: HeadRenderer: Client side localization fails for missing localization
### Describe the bug

Exception is thrown by [encodeJS](https://github.com/primefaces/primefaces/blob/10b13d49a1e79f9ea53124839c324ce8499f58cc/primefaces/src/main/java/org/primefaces/renderkit/HeadRenderer.java#L175) when a localization resource is missing.

### Reproducer

Set a locale that is not provided. See [META-INF/resources/primefaces/locales](https://github.com/primefaces/primefaces/tree/10b13d49a1e79f9ea53124839c324ce8499f58cc/primefaces/src/main/resources/META-INF/resources/primefaces/locales)

### Expected behavior

Maybe [encodeBegin](https://github.com/primefaces/primefaces/blob/10b13d49a1e79f9ea53124839c324ce8499f58cc/primefaces/src/main/java/org/primefaces/renderkit/HeadRenderer.java#L130) should retry with a default locale. Or skip throwing the exception if a default locale is already provided in JavaScript.

### PrimeFaces edition

Community

### PrimeFaces version

11.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.2.19

### Java version

11

### Browser(s)

_No response_

melloware:
Fix #10150: Log warning on missing locale-XX.js file

### DIFF
--- a/primefaces/src/main/java/org/primefaces/renderkit/HeadRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/renderkit/HeadRenderer.java
@@ -26,6 +26,8 @@ package org.primefaces.renderkit;
 import java.io.IOException;
 import java.util.List;
 import java.util.Locale;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 import javax.el.ELContext;
 import javax.el.ExpressionFactory;
@@ -65,6 +67,7 @@ import org.primefaces.util.LocaleUtils;
  */
 public class HeadRenderer extends Renderer {
 
+    private static final Logger LOGGER = Logger.getLogger(HeadRenderer.class.getName());
     private static final String LIBRARY = "primefaces";
 
     @Override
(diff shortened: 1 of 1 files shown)


primefaces/src/main/java/org/primefaces/renderkit/HeadRenderer.java
```diff
@@ -126,8 +129,16 @@ public void encodeBegin(FacesContext context, UIComponent component) throws IOEx
...
-            Locale locale = LocaleUtils.getCurrentLocale(context);
-            encodeJS(context, LIBRARY, "locales/locale-" + locale.getLanguage() + ".js");
+            try {
+                Locale locale = LocaleUtils.getCurrentLocale(context);
+                encodeJS(context, LIBRARY, "locales/locale-" + locale.getLanguage() + ".js");
+            }
+            catch (FacesException e) {
+                if (LOGGER.isLoggable(Level.WARNING)) {
+                    LOGGER.log(Level.WARNING,
+                            "Failed to load client side locale.js. {0}", e.getMessage());
```
tandraschko, line 139: I would log the filename inclusive language

melloware: Good idea

mihaiDumea: Maybe is a good idea to replace the condition `if (LOGGER.isLoggable(Level.WARNING)) {` with `if( FacesContext.getCurrentInstance( ).isProjectStage( ProjectStage.Development ) )` as nothing can be done to get rid of this message.

Regarding the message content, the filename and language is already included in exception message: `Failed to load client side locale.js. Error loading JavaScript, cannot find "locales/locale-zz.js" resource of "primefaces" library`

PR #10159, opened 2023-05-22
Fix #10157: Splitbutton proper icon CSS

### LINKED ISSUE #10157: SplitButton: RowSelect of TreeTable is triggered
### Describe the bug

`SplitButtonRenderer#encodeMenuIcon` [has hardcoded style classes](https://github.com/primefaces/primefaces/blob/e95160b1ae8f55c61d41da32c6aadad0e687ef60/primefaces/src/main/java/org/primefaces/component/splitbutton/SplitButtonRenderer.java#L204) for the menu icon which is missing `ui-c`.

I did not find any reference on why this marker style class was introduced, but it seems to be used by the tree table to exclude `span` elements from triggering the row selection feature.

Maybe use `HTML.BUTTON_LEFT_ICON_CLASS`?

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

Community

### PrimeFaces version

master

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.2

### Java version

8

### Browser(s)

_No response_

melloware:
Fix #10157: Splitbutton proper icon CSS

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/splitbutton/SplitButtonRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/splitbutton/SplitButtonRenderer.java
@@ -201,7 +201,7 @@ public class SplitButtonRenderer extends MenuItemAwareRenderer {
 
         //icon
         writer.startElement("span", null);
-        writer.writeAttribute("class", "ui-button-icon-left ui-icon ui-icon-triangle-1-s", null);
+        writer.writeAttribute("class", HTML.BUTTON_LEFT_ICON_CLASS + " ui-icon-triangle-1-s", null);
         writer.endElement("span");
 
         //text


primefaces/src/main/java/org/primefaces/component/splitbutton/SplitButtonRenderer.java
```diff
@@ -201,7 +201,7 @@ protected void encodeMenuIcon(FacesContext context, SplitButton button, String i
 
         //icon
         writer.startElement("span", null);
-        writer.writeAttribute("class", "ui-button-icon-left ui-icon ui-icon-triangle-1-s", null);
+        writer.writeAttribute("class", HTML.BUTTON_LEFT_ICON_CLASS + " ui-icon-triangle-1-s", null);
```
jepsar, line 204: Do we still need `ui-icon-triangle-1-s`?

melloware: Yep that is is the dropdown icon itself.  It was missing `ui-c` but instead of adding `ui-c` there is already a whole LEFT_ICON_CLASS just for putting the icon correctly, but naming which icon is still important.

jepsar:
```css
body .ui-splitbutton .ui-button.ui-splitbutton-menubutton .ui-icon:before
```
instead of 
```css
body .ui-splitbutton .ui-button.ui-splitbutton-menubutton .ui-icon.ui-icon-triangle-1-s:before
```
would also work

melloware: That is still missing `ui-c` right?  Which was the whole genesis of the ticket as Datatable ignores all things with `ui-c` in it?

jepsar: Yes, that's unrelated to the `ui-c` thing. Maybe this should be refactored in 14 when custom icons are introduced.

melloware: Yeah I think probably all the icons need to be revisted.

PR #10164, opened 2023-05-23
Fix #9922: Autocomplete ARIA labels

### LINKED ISSUE #9922: AutoComplete: uses hardcoded, english text in widget
https://github.com/primefaces/primefaces/blob/43f22012f9f2bb58ede5bbed83e8652e5b167fbf/primefaces/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js#L124-L125

melloware:
Fix #9922: Autocomplete ARIA labels

@Rapster need French
@jepsar need Dutch

`primefaces.autocomplete.RESULTS_MESSAGE=results are available, use up and down arrow keys to navigate`


jepsar:
Ha, I wanted to look into this one this week (back from holidays). One thing: is "results are available ...." prefixed with the number of results? Like "34 results are available ...."? If so, I don't think FR is correct (am I right @Rapster ?), otherwise "results" should start with a capital R.

Rapster:
Exactement! I updated my last coment

primefaces/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
```diff
@@ -121,8 +121,7 @@ PrimeFaces.widget.AutoComplete = PrimeFaces.widget.BaseWidget.extend({
         this.cfg.pojo = this.hinput.length == 1;
         this.cfg.minLength = this.cfg.minLength != undefined ? this.cfg.minLength : 1;
         this.cfg.cache = this.cfg.cache || false;
-        this.cfg.resultsMessage = this.cfg.resultsMessage || ' results are available, use up and down arrow keys to navigate';
-        this.cfg.ariaEmptyMessage = this.cfg.emptyMessage || 'No search results are available.';
+        this.cfg.ariaEmptyMessage = this.cfg.emptyMessage;
```
jepsar: `this.cfg.resultsMessage` is already used to display the message? (I see we do the fallback in Java now).

tandraschko:
couldnt we use client side localization? this we avoid render text each time for each component

melloware:
> couldnt we use client side localization? this we avoid render text each time for each component

We could but it was already server side localized so I was just updating what was there?

primefaces/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
melloware: yep

tandraschko:
yeah, i know
i was just thinking: f we touch it now, should we optimize it?

just place 3 autoComplete now on the view, we would render 3 times the same msg - instead of using the one from the JS file

PR #10183, opened 2023-05-29
#9129 OutputLabel: Keep required indicator visible with PrimeFlex

### LINKED ISSUE #9129: OutputLabel: required indicator placement with PrimeFlex can be hidden
### Describe the bug

Due to the PrimeFlex CSS rule `.field.grid>label { display: flex; ... }` labels do not overflow. This can push the required indicator behind the input if the label is too long.

Screenshot 2022-08-23 at 15 47 40

### Reproducer

See for example https://primefaces.org/diamond/formlayout.xhtml

Inspect a label of the horizontal example and make it too long + add a required indicator `<span class="ui-outputlabel-rfi">*</span>` within the `label` element.

You will end up with something like
```html
<div class="card ui-fluid">
    <h5>Horizontal</h5>
    <div class="field grid"><label id="j_idt87" class="ui-outputlabel ui-widget col-12 mb-2 md:col-2 mb-md-0" for="name3">Namewichiswaytoolong<span class="ui-outputlabel-rfi">*</span></label>
        <div class="col-12 md:col-10"><input id="name3" name="name3" type="text" class="ui-inputfield ui-inputtext ui-widget ui-state-default ui-corner-all" role="textbox" aria-readonly="false" aria-disabled="false">
        </div>
    </div>
    ...
</div>
```

### Expected behavior

The required indicator should always be visible.

Possible solution:

Make the label position relative and `.ui-outputlabel-rfi` absolute with right = 0 (might bite RTL).

### PrimeFaces edition

_No response_

### PrimeFaces version

11

### Theme

Diamond

### JSF implementation

All

### JSF version

2.3

### Java version

17

### Browser(s)

Chrome latest macOS

jepsar:
Fix #9129

### DIFF
--- a/primefaces-showcase/src/main/webapp/ui/misc/outputLabel.xhtml
+++ b/primefaces-showcase/src/main/webapp/ui/misc/outputLabel.xhtml
@@ -20,39 +20,25 @@
         <div class="card">
             <h:form>
                 <p:messages id="messages"/>
-                <h:panelGrid columns="2" cellpadding="7">
-                    <h:outputLabel for="name" value="Standard Label"/>
-                    <p:inputText id="name" required="true"/>
-
-                    <p:outputLabel for="@next" value="Extended Label"/>
-                    <p:inputText id="extended" required="true"/>
-
-                    <p:outputLabel for="@next" value="Number"/>
-                    <p:spinner id="number" value="0"/>
-
-                    <p:commandButton value="Submit" update="@form" styleClass="mt-3"/>
-                    <p:spacer />
-
-                    <!-- For testing, don't remove!
-                    <p:outputLabel for="@next" value="Label which considers @NotNull:" />
-                    <p:inputText id="nullable" value="#{outputLabelView.nullableValue}" />
-
-                    <p:outputLabel for="@next" value="Label which considers @NotNull:" />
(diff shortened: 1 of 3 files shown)


primefaces-showcase/src/main/webapp/ui/misc/outputLabel.xhtml
```diff
@@ -20,39 +20,25 @@
...
-                    <p:outputLabel for="@next" value="Extended Label"/>
-                    <p:inputText id="extended" required="true"/>
-
-                    <p:outputLabel for="@next" value="Number"/>
-                    <p:spinner id="number" value="0"/>
-
-                    <p:commandButton value="Submit" update="@form" styleClass="mt-3"/>
-                    <p:spacer />
-
-                    <!-- For testing, don't remove!
```
melloware, line 36: Thanks I think this were safe to remove as I added Integration Tests for all of these scenarios.

melloware (approved):

primefaces-showcase/src/main/webapp/ui/misc/outputLabel.xhtml
jepsar: Yeah, I checked

jepsar:
What should be the milestone for this? 14?

PR #10190, opened 2023-05-30
Fix #10189: Dialog should render not displayed

### LINKED ISSUE #10189: Dialog: Should render with display:none
Right now dialogs are rendered in the DOM but they are not set to `display:none`   in Quarkus this has a weird side effect of that the dialogs are being rendered on the screen before any `.show()` method is called.

image

The dialog is in the DOM. When you then click `.show()` and then `hide()` the appropriate `display:none` is added to the dialog but its not there on initial rendering.

melloware:
Fix #10189: Dialog should render not displayed

Right now the dialog CSS goes like this.

Initial Page Load: `no element style`

Show: `display:block`

Hide: `display: none`

Show: `display: block`

Hide: `display:none`

The initial state should be `display:none`

### DIFF
--- a/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/dialog/Dialog001Test.java
+++ b/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/dialog/Dialog001Test.java
@@ -43,6 +43,7 @@ public class Dialog001Test extends AbstractPrimePageTest {
     public void testShowWidget(Page page) {
         // Arrange
         Dialog dialog = page.dialog;
+        Assertions.assertFalse(dialog.isVisible());
 
         // Act
         dialog.show();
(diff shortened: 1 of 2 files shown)


tandraschko:
shouldnt we just add this as default style to ui-dialog or similar?

melloware:
Well i was looking through the commits it looks like it has `ui-hidden-container` on the Dialog class already but I can't find any CSS that actually works with that.  It looks like this has gone through many revisions over the years but it seems like it can't be CSS on the class itself because them the local element styling isn't overriding it.   The UI specifically does this..

```js
            if (this.positionInitialized === false) {
                this.jqEl.style.visibility = "hidden";
                this.jqEl.style.display = "block";
                this.initPosition();
                this.jqEl.style.display = "none";
                this.jqEl.style.visibility = "visible";
            }
```

So it looks like all the visibility and sizing and positioning is handled with styles right on the dialog and not in the CSS and I assume that is done for a reason?

PR #10195, opened 2023-05-31
Fix #10191: Spinner change links to buttons

### LINKED ISSUE #10191: Spinner/ScrollTop: SEO use HREF on anchor
### Description

I was checking our ecommerce shop with google PageSpeed insights and wondered why our SEO score had taken a hit reducing it from 100% to 92% and the problem was _p:spinner_ up and down anchors.

The problem is a missing _href_ attribute which google punishes from SEO perspective. The crawler does not know its not a real anchor.

grafik

**Other components**

Maybe other components are also affected?

**Ideas**

- add rel = “nofollow” and maybe google will ignore this anchor
- Or better dont use anchor tag at all

### Additional context

_No response_

melloware:
Fix #10191: Spinner change links to buttons


melloware:
@djmj this is a much better fix and it also make it more accessible with the ARIA changes.  

@newk5 if you don't mind PT translation
@jepsar Dutch
@Rapster French

Rapster:
```
'spinner.INCREASE': 'Augmenter la valeur',
'spinner.DECREASE': 'Diminuer la valeur'
```

jepsar:
NL
```
         'spinner.INCREASE': 'Verhoog waarde',
         'spinner.DECREASE': 'Verlaag waarde'
```

tuerker:
@melloware This PR ruins my layout and so the usability of the spinner. I assume that the problem is caused by the following change:

 ```
-  writer.startElement("a", null);
+ writer.startElement("button", null);
+ writer.writeAttribute("id", spinner.getClientId(context) + "-" + direction, "id");
+ writer.writeAttribute("type", "button", null);
```

For example, the screen

image

looks as follows after this PR:

image

This is surely because i have defined the following for all my buttons:

```
button[type="button"], button[type="reset"], button[type="submit"], input[type="reset"], input[type="submit"], input[type="button"] {
    font-family: Arial, Helvetica, sans-serif;
    background: #eeeff0;
    height: 20px;
    border: 1px solid #d4d7d9;
    cursor: pointer;
    min-width: 75px;
    color: #545e69;
}
```

Since I am sure that i am not the only one having such css definitions, my question is whether declaring the spinner as button was really needed. So it actually becomes a breaking change.

Do you have you an advice how to best solve this problem? Many thanks in advance.

**Comment by jepsar** [comment]
@tuerker what theme are you using? Are you using any custom CSS?

**Comment by tuerker** [comment]
Just an add-on: My theme config is:

```
  <context-param>
    <param-name>primefaces.THEME</param-name>
    <param-value>none</param-value>
  </context-param>
```

**Comment by jepsar** [comment]
Applying custom CSS always comes with a risk that your changes will not be compatible with a newer version of whatever component library you are using. This fix was an accessibility enhancement. And if you think about it, button HTML elements make more sense to represent UI buttons than anchor HTML elements. Custom button types  are not allowed https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button. I think this will be a matter of updating your custom styles.

**Comment by tuerker** [comment]
Yes but this approach has a massiv disadvantage. I want that all my "normal" buttons have a certain min-width and height, as shown in the css fragment above. Obviously, by "normal" buttons I am not referring at "spinner-buttons", where this min-width and height definitions make no sens and therefore should not be applied. So there must be an easy way to specify the layout of "normal" buttons that are not "spinner" buttons. Do you have a proposal how this can be done now with adding an additional styleClass attribute to several thousands buttons?

**Comment by tuerker** [comment]
it was meant: ... WITHOUT adding an additional styleClass attribute to several thousands buttons?

**Comment by jepsar** [comment]
Can you please try
```css
:not(.ui-spinner) > :is(button[type="button"], button[type="reset"], button[type="submit"], input[type="reset"], input[type="submit"], input[type="button"]) {
    font-family: Arial, Helvetica, sans-serif;
    background: #eeeff0;
    height: 20px;
    border: 1px solid #d4d7d9;
    cursor: pointer;
    min-width: 75px;
    color: #545e69;
}
```

tuerker:
many thanks for the hint to use negation in the css selector. 

i have the following 

image

instead of 

image

so, the spinner buttons look ok now but this definition breaks the autocomplete-drowdown button (see first component in the screenshots). so, i have to exclude this buttons too: 

```
:not(.ui-spinner, ui-autocomplete) > :is(button[type="button"], button[type="reset"], button[type="submit"], input[type="reset"], input[type="submit"], input[type="button"]) {
    font-family: Arial, Helvetica, sans-serif;
    background: #eeeff0;
    height: 20px;
    border: 1px solid #d4d7d9;
    cursor: pointer;
    min-width: 75px;
    color: #545e69;
}
```

i hope that there are not so much more cases to be filtered out.

thanks again for your quick response. best regards.

melloware:
Just to chime in I agree with @jepsar that when you "roll your own" CSS you can't expect a library to not make improvements to its code in this case ARIA and Accessibility change in my opinion are way more important than CSS changes.  I also believe the changes made here were the correct changes and that it should have been like this originally.

PR #10208, opened 2023-06-06
Fix #10207: BlockUI improve performance

### LINKED ISSUE #10207: BlockUI slows down the application
### Describe the bug

We noticed that our JSF application has become slower and slower while working with it. We found out that the problem lies in the processing of the 'alignOverlay' for BlockUI. Since PF12, BlockUI has been subscribed to all DOM update events so that it can be resized even if another DOM element has changed.

However, each time we update the component, the BlockUI is subscribed again, causing it to be processed by 'alignOverlay'. Our application performs much better with the old version of 'bindResizer', the one before [this fix](https://github.com/primefaces/primefaces/pull/9159/files).

### Reproducer

[primefaces-test-blockui-events.zip](https://github.com/primefaces/primefaces/files/11660741/primefaces-test-blockui-events.zip)
[blockui_test_case_github.patch](https://github.com/primefaces/primefaces/files/11660745/blockui_test_case_github.patch)

### Expected behavior

In the console we see each time we click the "reset" button a listener is added and there are more and more alignOverlay logs. Ths is, of course, not a problem for a simple small example, but it became relevant for complex applications.

### PrimeFaces edition

Community

### PrimeFaces version

since 12.0.0

### Theme

Babylon

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

11-17

### Browser(s)

_No response_

melloware:
Fix #10207: BlockUI improve performance

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
@@ -78,7 +78,7 @@ PrimeFaces.widget.BlockUI = PrimeFaces.widget.BaseWidget.extend({
         this.blocker.remove();
         this.jq.remove();
         this.target.attr('aria-busy', false);
-        $(document).off('pfAjaxSend.' + this.id + ' pfAjaxComplete.' + this.id);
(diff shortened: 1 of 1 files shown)


crocimi:
Thank you for the quick reply! I have tested it integrating the changes directly in my primefaces-custom.js and the issue seems to be fixed! The performance is constant and I see only one "alignOverlay" event for each blockUI, even after multiple updates.

Thx :-)

PR #10218, opened 2023-06-11
Fix #10209 - DataTable: Filter usage modify the value object type

### LINKED ISSUE #10209: DataTable: Filter usage modify the value object type
### Describe the bug

When using Datatable with the filter feature and using a ListDataModel for the datatable value, the value object is updated with a simple List<>.

A sample project as been created here to ilustrate the case:
https://github.com/rsi21/primefaces-test/tree/datatable_with_filter

The project displays only a simple datatable with sort on columns and a filter on a first column.
On each sort or filter the value type is displayed/updated using the binding feature of the datatable.

In resume, if you use sort feature, we can see that the value object type is not modified.
If we use the filter feature, then it is modified by a simple List<> object.

For information, also some exchanges has been made around the issue in the forum here:
https://forum.primefaces.org/viewtopic.php?f=3&t=81036

### Reproducer

https://github.com/rsi21/primefaces-test/tree/datatable_with_filter

### Expected behavior

Keep always the original object type in the datatable value.

### PrimeFaces edition

Community

### PrimeFaces version

12.0.4

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

11

### Browser(s)

_No response_

Rapster:
I tried to keep it very simple for now, but indeed we should keep DataTable#value consistent before/after filtering.

Out of topic: `CollectionDataModel` and `CollectionDataModel` should not be used in 2.3 and above, PF should only support it when it's necessary. IMO, they should be mark as deprecated

Fix #10209

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java
+++ b/primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java
@@ -1229,7 +1229,7 @@ public class DataTable extends DataTableBase {
                 }
             }
             catch (ReflectiveOperationException e) {
-                Logger.getLogger(FilterFeature.class.getName()).log(Level.SEVERE, e.getMessage());
+                Logger.getLogger(DataTable.class.getName()).log(Level.SEVERE, e.getMessage());
             }
         }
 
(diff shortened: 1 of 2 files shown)


melloware:
Does this same change need to get applied to TreeTable as well?

Rapster:
TreeTable is based on TreeNode not DataModel

melloware:
Perfect just checking whenever I see a DT change that we check TT

melloware (approved):

PR #10229, opened 2023-06-14
Fix #10226: CSV byte order mark for UTF-8

### LINKED ISSUE #10226: DataExporter: new attribute to set BOM when uses type CSV and encoding UTF-8
### Description

When you export a CSV-file using the Datatable component and DataExporter, the resulting file is encoded with UTF-8 without a BOM and there's no way to change this, causing encoding issues with latin language (For example with characteres á, é, í, ó ú)

This is an encoding problem for example when users open CSV File with Excel (The most common alternative to open this kind of files).

Note: Other types of encoding do not solve the problem ()

Related posts: 
https://github.com/primefaces/primeng/issues/9172
https://mkyong.com/java/java-how-to-add-and-remove-bom-from-utf-8-file/

Just exporting using UTF-8 is not enough. Users should be able to either change the encoding of the generated file or to add the BOM so the resulting CSV file is fully compatible with excel.

Something like this in DataTableCSVExporter class:

`
           
	        	
           @Override
	    public void doExport(FacesContext context, DataTable table, ExportConfiguration exportConfiguration, int index) throws IOException {
	        try (OutputStreamWriter osw = new OutputStreamWriter(getOutputStream(), exportConfiguration.getEncodingType());
                    if (exportConfiguration.isBomRequired()) { // Add new attribute bomRequired
	        	      writer.write('\ufeff'); //write the UTF-8 BOM to make Excel able to detect UTF-8 encoding
                      }
     
                      ...
                      ...
                      ...
`

### Additional context

_No response_

melloware:
Fix #10226: CSV byte order mark for UTF-8

### DIFF
--- a/primefaces-integration-tests/src/test/resources/org/primefaces/integrationtests/dataexporter/languages-pageonly.csv
+++ b/primefaces-integration-tests/src/test/resources/org/primefaces/integrationtests/dataexporter/languages-pageonly.csv
@@ -1,3 +1,3 @@
-"ID","Type","Name"
+﻿"ID","Type","Name"
 "4","INTERPRETED","TypeScript"
 "5","INTERPRETED","Python"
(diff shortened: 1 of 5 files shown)


raul810y (approved):
excellent!

PR #10237, opened 2023-06-15
#10227 SelectOneRadio: remove `plain` attribute

### LINKED ISSUE #10227: SelectOneRadio: remove `plain` attribute
Introduced in: https://code.google.com/archive/p/primefaces/issues/4760

But I don't see the point. No other components have such an attribute and it makes code harder to maintain (for example #9857). Without plain, the component still is accessible (if not we can / should improve that), and it is fully styleable.

Maybe we can deprecate it in 13 and remove in 14? WDYT?

jepsar:
Fix #10227

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/selectoneradio/SelectOneRadio.java
+++ b/primefaces/src/main/java/org/primefaces/component/selectoneradio/SelectOneRadio.java
@@ -43,7 +43,6 @@ public class SelectOneRadio extends SelectOneRadioBase {
     public static final List<String> DOM_EVENTS = LangUtils.unmodifiableList("onchange", "onclick");
 
     public static final String STYLE_CLASS = "ui-selectoneradio ui-widget";
-    public static final String NATIVE_STYLE_CLASS = "ui-selectoneradio ui-selectoneradio-native ui-widget";
 
     public String getRadioButtonId(FacesContext context, int index) {
         return this.getClientId(context) + UINamingContainer.getSeparatorChar(context) + index;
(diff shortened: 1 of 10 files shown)


primefaces/src/main/java/org/primefaces/component/selectoneradio/SelectOneRadio.java
```diff
@@ -43,7 +43,6 @@ public class SelectOneRadio extends SelectOneRadioBase {
     public static final List<String> DOM_EVENTS = LangUtils.unmodifiableList("onchange", "onclick");
 
     public static final String STYLE_CLASS = "ui-selectoneradio ui-widget";
-    public static final String NATIVE_STYLE_CLASS = "ui-selectoneradio ui-selectoneradio-native ui-widget";
```
melloware, line 46: Need to remove a bunch of styles in `forms.css` like the following:

```css
.ui-radiobutton.ui-radiobutton-native {
    margin: 0;
    padding: 0;
}

.ui-radiobutton.ui-radiobutton-native  .ui-helper-hidden-accessible {
    clip: auto;
    position: static !important;
    width: auto;
    height: auto;
    overflow: visible;
    margin: 0;
    padding: 0;
}

.ui-radiobutton.ui-radiobutton-native  .ui-radiobutton-box {
    display:none;
}

.ui-radiobutton.ui-radiobutton-native .ui-helper-hidden-accessible input {
    margin: 0;
    -webkit-transform: none; 
    -moz-transform: none; 
    transform: none;
}

.ui-selectoneradio-native label {
    margin: 0;
}
```

melloware (changes requested):

melloware (approved):

PR #10243, opened 2023-06-16
Fix #10242: DashboardColumn deprecation

### LINKED ISSUE #10242: Dashboard: DashboardColumn was completely removed, not moved

melloware:
Fix #10242: DashboardColumn deprecation

### DIFF
--- a/primefaces-showcase/src/main/java/org/primefaces/showcase/view/panel/DashboardView.java
+++ b/primefaces-showcase/src/main/java/org/primefaces/showcase/view/panel/DashboardView.java
@@ -58,9 +58,9 @@ public class DashboardView implements Serializable {
 
         // legacy
         legacyModel = new DefaultDashboardModel();
-        legacyModel.addColumn(new DefaultDashboardColumn(Arrays.asList("sports", "finance")));
-        legacyModel.addColumn(new DefaultDashboardColumn(Arrays.asList("lifestyle", "weather")));
-        legacyModel.addColumn(new DefaultDashboardColumn(Arrays.asList("politics")));
+        legacyModel.addWidget(new DefaultDashboardColumn(Arrays.asList("sports", "finance")));
+        legacyModel.addWidget(new DefaultDashboardColumn(Arrays.asList("lifestyle", "weather")));
+        legacyModel.addWidget(new DefaultDashboardColumn(Arrays.asList("politics")));
     }
 
     public void handleReorder(DashboardReorderEvent event) {
--- a/primefaces-showcase/src/main/java/org/primefaces/showcase/view/panel/DashboardView.java
+++ b/primefaces-showcase/src/main/java/org/primefaces/showcase/view/panel/DashboardView.java
@@ -58,9 +58,9 @@ public class DashboardView implements Serializable {
 
         // legacy
         legacyModel = new DefaultDashboardModel();
-        legacyModel.addWidget(new DefaultDashboardColumn(Arrays.asList("sports", "finance")));
-        legacyModel.addWidget(new DefaultDashboardColumn(Arrays.asList("lifestyle", "weather")));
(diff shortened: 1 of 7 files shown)


tandraschko:
why should we have a DefaultColumn but no Column? doesnt make sense for me

melloware:
Its really not a Column though its just a Widget and in responsive its laid out based on repsonsive design.  The Column is the old dahhboard concept but calling them columns to support both doesn't make sense.  Really its just a DashboardWidget.

tandraschko:
will come back later, checking other errors now

melloware:
ok no problem. I think I left it there for continuity with the legacy style dashboard.

```java
private DashboardModel responsiveModel;
    private DashboardModel legacyModel;

    @PostConstruct
    public void init() {
        // responsive
        responsiveModel = new DefaultDashboardModel();
        responsiveModel.addWidget(new DefaultDashboardWidget("bar", RESPONSIVE_CLASS));
        responsiveModel.addWidget(new DefaultDashboardWidget("stacked", RESPONSIVE_CLASS));
        responsiveModel.addWidget(new DefaultDashboardWidget("donut", RESPONSIVE_CLASS.replaceFirst("xl:col-\\d+", "xl:col-4")));
        responsiveModel.addWidget(new DefaultDashboardWidget("cartesian", RESPONSIVE_CLASS.replaceFirst("xl:col-\\d+", "xl:col-8")));

        // legacy
        legacyModel = new DefaultDashboardModel();
        legacyModel.addColumn(new DefaultDashboardColumn(Arrays.asList("sports", "finance")));
        legacyModel.addColumn(new DefaultDashboardColumn(Arrays.asList("lifestyle", "weather")));
        legacyModel.addColumn(new DefaultDashboardColumn(Arrays.asList("politics")));
    }
```

melloware:
I agree it was sort of a mess before I just tried to clean it up a little.  it goes like this.

`DashboardModel` = just the container
`DashboardColumn` 0...N elements that get laid out in legacy mode
`DashboardWidget` 0...N elements that get laid out in responsive mode

tandraschko:
what is the difference between widget<>column?

tandraschko:
then we should also deprecate DefaultDashoardColumn and addColumn etc. methods

melloware:
OK everything `Column` related is now deprecated with a comment saying use the `Widget` version

primefaces-showcase/src/main/java/org/primefaces/showcase/view/panel/DashboardView.java
```diff
@@ -58,9 +58,9 @@ public void init() {
 
         // legacy
         legacyModel = new DefaultDashboardModel();
-        legacyModel.addColumn(new DefaultDashboardColumn(Arrays.asList("sports", "finance")));
-        legacyModel.addColumn(new DefaultDashboardColumn(Arrays.asList("lifestyle", "weather")));
-        legacyModel.addColumn(new DefaultDashboardColumn(Arrays.asList("politics")));
+        legacyModel.addWidget(new DefaultDashboardColumn(Arrays.asList("sports", "finance")));
```
tandraschko: shouldnt it then also be DefaultDashboardWidget?

melloware: You are right and I just moved the code into DefaultDashboardWidget and instead of it was the other way around.

tandraschko:
what about the docs?
Also we should add more infos to the migration guide

melloware:
OK I added some more detail to the migration guide...let me know if you want me to add anything else.

tandraschko:
looks good for me
will continue further testing of my customers applications end of this week

tandraschko:
@melloware Dashboard#updateDashboardModel still uses this methods, cant we remove depreacted method usage?

melloware:
I will submit a PR

PR #10257, opened 2023-06-21
Fix #10251: FileDownload allow disabling of `no-store`

### LINKED ISSUE #10251: FileDownload: downloading .xhtml instead of PDF in Google Chrome (works in Mozilla Firefox)
### Describe the bug

When attempting to download a PDF file and opening it in a new window using the `contentDisposition="inline"` attribute with `p:fileDownload` and `ajax="false"` on `p:commandButton`, the file is downloaded as .xhtml instead of a PDF in Google Chrome and Microsoft Edge. This issue does not occur in Mozilla Firefox.

FileDownload

### Reproducer

[FileDownloadReproducer.zip](https://github.com/primefaces/primefaces/files/11792741/FileDownloadReproducer.zip)

1. Download the project.zip file;
2. Extract the contents of the project.zip file;
3. Deploy the project using a web application server (I'm using the Wildfly 27);
4. Open a web browser (Google Chrome or Microsoft Edge) and navigate to the following URL: "http://localhost:8080/filedownload/";
5. On the web page, click on the "Download" button;
6. Observe that the downloaded file has the incorrect extension (.xhtml) instead of being downloaded as a PDF.

### Expected behavior

The PDF file should be downloaded correctly with the correct file extension (.pdf) in all supported browsers, including Google Chrome and Microsoft Edge.

### PrimeFaces edition

Community

### PrimeFaces version

12.0.0 Jakarta

### Theme

AdminFaces

### JSF implementation

Mojarra

### JSF version

Jakarta Faces 4.0

### Java version

20

### Browser(s)

Google Chrome v114.0.5735.133; Brave Browser v1.52.126; Microsoft Edge v114.0.1823.51.

melloware:
Fix #10251: FileDownload allow enabling of caching.

Added this as an enhancement for the PDF use case. FALSE by default.

**Update:** this PR only allows for disabling the `no-store` option not the rest of the caching values.

### DIFF
--- a/docs/13_0_0/components/filedownload.md
+++ b/docs/13_0_0/components/filedownload.md
@@ -17,7 +17,6 @@ stream the binary data. FileDownload presents an easier way to do the same.
 | value | null | StreamedContent | A streamed content instance.
 | contentDisposition | attachment | String | Specifies display mode on non-AJAX downloads.
 | monitorKey | null | String | Optional key to support monitoring multiple filedownloads on same page.
-| cache | false | Boolean | Controls browser caching mode of the resource.
 
 ## Getting started with FileDownload
 A user command action is required to trigger the file-download process.
@@ -73,10 +72,6 @@ Please see our core documentation about it [Dynamic Content Streaming / Renderin
 On regular (non-AJAX) downloads, by default, content is displayed as an `attachment` with a download dialog box,
 another alternative is the `inline` mode, in this case browser will try to open the file internally without a prompt.
 Note that content disposition is not part of the http standard, although it is widely implemented.
-
-## Cache
-Most of the time you want to leave cache="false" which is the default to not cache the results.  However, in the case of PDF Downloads there have
-been issues with some browser like Chrome/Edge. See: https://github.com/primefaces/primefaces/issues/10251
 
 ## Monitor Status
 When fileDownload is used without AJAX, ajaxStatus cannot apply. Still PrimeFaces provides a feature
--- a/docs/13_0_0/components/filedownload.md
+++ b/docs/13_0_0/components/filedownload.md
@@ -17,6 +17,7 @@ stream the binary data. FileDownload presents an easier way to do the same.
 | value | null | StreamedContent | A streamed content instance.
 | contentDisposition | attachment | String | Specifies display mode on non-AJAX downloads.
 | monitorKey | null | String | Optional key to support monitoring multiple filedownloads on same page.
+| cache | false | Boolean | Controls browser caching mode of the resource.
 
 ## Getting started with FileDownload
 A user command action is required to trigger the file-download process.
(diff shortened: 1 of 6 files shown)


tandraschko:
what will happen if you click 2 times? the first file is downloaded?

i dont think this should be commited currently

melloware:
No i think all this does it allow it to work like PF < 10.0 used to work?  At least from my comparison of comparing 6.2 vs 12.0.0.

tandraschko:
but what happen if you click 2 times?
set a cache header to a dynamic generated file, which will be different on each click, just dont not make any sense and could cause very strange results

melloware:
It could definitely lead to strange results but its false by default so someone would have to turn this on and KNOW they wanted this behavior for PDF's right?

Bukama:
> It could definitely lead to strange results but its false by default so someone would have to turn this on and KNOW they wanted this behavior for PDF's right?

I guess we are all on the same side, saying that with dynamic content caching makes no sense and should always be false. But I assume that the `FileDownloadActionListener` can't find out by itself it the content is dynamic or not? If my assumption is right, than I think the default of `false` is the the best way, so users have to explicitly enable caching. Maybe we should point out in the docs section (the one you added) that enabling caching would only make sense if you provide static downloadable content, but not dynamic rendered one?

melloware:
@Bukama yep I put that in the docs mentioning the `cache` value and the only time you would want to use it.

mstichweh:
Completly disabling the cache does not seem not to be the right solution to fix the download issue on Chrome / Edge. Because dynamiclaly generated documents should not be cached at all, as @tandraschko mentioned.  
The problem seems to be only the "no-store" option of the "Cache-Control" header.

So the default for the p:fileDownload component should be without "no-store":

```
externalContext.setResponseHeader("Cache-Control", "no-cache, must-revalidate");
externalContext.setResponseHeader("Pragma", "no-cache");
externalContext.setResponseHeader("Expires", "0");
```

The "no-store" option could then be explicitly enabled for very secure files by an additional p:fileDownload attribute. Or the "Cache-Control" header becomes configurable, e.g. via the `StreamedContent` to overwrite the dedault.

melloware:
Good feedback let me update my PR.

tandraschko:
i think we should rename the "cache" attribute then to "store"
in general caching a dynamic file, which should be always true for a download (right?), doesnt make any sense - compared to graphicImage cache=true/false.

WDYT?

melloware:
Yep I can rename to store

tandraschko:
+1

atiqzaman-dsi:
Is it possible to get this fix in 13.x version?

melloware:
@tandraschko what do you think?  This is really an enhancement not a bug but i can see this being a roadblock should we backport this to 13.0.6?

Hint10:
> Is it possible to get this fix in 13.x version?

I will be also glad about a backport.

jepsar:
On the other hand, isn't PF 14 going to be released pretty soon?

melloware:
I think i will backport this to 13.0.6 only because 13 is the last version supporting Java 8 and JSF < 2.2 so I feel like 13 will be around for a WHILE.

tandraschko:
+1 for this small enhancement

PR #10265, opened 2023-06-23
Fix #10264 - Exporter: Custom registration issue

### LINKED ISSUE #10264: Exporter: Custom registration issue
CC @Rapster this is based on the refactoring done in #9969

We have a custom DataTable like this..

```java
public class MellowareDataTable extends DataTable {
```

When we go to Export in 13.0 we get this error...

```
java.lang.UnsupportedOperationException: 
Component class com.melloware.MellowareDataTable not supported. Use DataExporter#register()
```

The reason is before PF 13 the code was lazy and registered like this assuming it was a Datatable if not a TreeTable...

```java
if (targetClass != null && TreeTable.class.isAssignableFrom(targetClass)) {
            return TreeTableExporterFactory.getExporter(exportAs, exporterOptions);
}

return DataTableExporterFactory.getExporter(exportAs, exporterOptions);
```

But now everything is stored in a map by class and since it does `map.get(Datatable.class)` our custom instance is not found.  I can try and work around this by doing the following...

```java
DataExporters.register(MellowareDataTable.class, DataTablePDFExporter.class, "pdf");
DataExporters.register(MellowareDataTable.class, DataTableCSVExporter.class, "csv");
DataExporters.register(MellowareDataTable.class, DataTableXMLExporter.class, "xml");
DataExporters.register(MellowareDataTable.class, DataTableExcelXExporter.class, "xlsx");
DataExporters.register(MellowareDataTable.class, DataTableExcelXStreamExporter.class, "xlsxstream");
DataExporters.register(MellowareDataTable.class, DataTableCSVExporter.class, "csv");
```

However that fails because the generics are assuming a class with a "custom" exporter for that class.  What I want is to register a new class `MellowareDataTable.class` to an existing exporter `DataTableCSVExporter.class` and it will not allow it.

So there two possible ways to fix this:

1. Fix the generics?
2. When registering the Exporters when checking for an exporter type check the class then walk it's super classes checking for a match until NONE is found.  

Not sure what is the better way to fix it?

Rapster:
Fixes #10264

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/export/DataExporters.java
+++ b/primefaces/src/main/java/org/primefaces/component/export/DataExporters.java
@@ -47,13 +47,13 @@ public final class DataExporters {
 
         // since users might defined their own impl of components (e.g MyDataTable, MyTreeTable etc.)
         // retrieve the most eligible built-in component class first (e.g DataTable, TreeTable) to get corresponding exporters
-        Class<? extends UIComponent> targetClassTmp = primeAppContext.getExporters().keySet().stream()
+        Class<? extends UIComponent> resolvedTargetClass = primeAppContext.getExporters().keySet().stream()
                 .filter(k -> k.isAssignableFrom(targetClass))
                 .findFirst()
                 .orElseThrow(() -> new UnsupportedOperationException(
                         "Component " + targetClass + " not supported. Use DataExporters#register()"));
 
-        Map<String, Class<? extends Exporter<?>>> supportedExporters = primeAppContext.getExporters().get(targetClassTmp);
+        Map<String, Class<? extends Exporter<?>>> supportedExporters = primeAppContext.getExporters().get(resolvedTargetClass);
 
         String newType = type.toLowerCase();
         Class<? extends Exporter<?>> exportClass = Optional.ofNullable(supportedExporters.get(newType))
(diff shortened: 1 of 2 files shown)


primefaces/src/main/java/org/primefaces/component/export/DataExporters.java
```diff
@@ -43,15 +43,22 @@ private DataExporters() {
     }
 
     public static <T extends UIComponent> Exporter<T> get(Class<T> targetClass, String type) {
-        String newType = type.toLowerCase();
         PrimeApplicationContext primeAppContext = PrimeApplicationContext.getCurrentInstance(FacesContext.getCurrentInstance());
-        Map<String, Class<? extends Exporter<?>>> supportedExporters = Optional.ofNullable(primeAppContext.getExporters().get(targetClass))
+
+        // since users might defined their own impl of components (e.g MyDataTable, MyTreeTable etc.)
+        // retrieve the most eligible built-in component class first (e.g DataTable, TreeTable) to get corresponding exporters
+        Class<? extends UIComponent> targetClassTmp = primeAppContext.getExporters().keySet().stream()
```
melloware: this looks great just minor nitpick would be `resolvedTargetClass` instead of `targetClassTmp` ?

melloware (approved):

primefaces/src/main/java/org/primefaces/component/export/DataExporters.java
Rapster: Thank you! Sounds very good, will make the change

PR #10271, opened 2023-06-27
Fix #10270: Dialog/ConfirmDialog/ConfirmPopup refocus element on hide

### LINKED ISSUE #10270: Dialog: hiding dialog does not restore focus on the button that invoked it
### Describe the bug

When navigating the [Confirm Dialog showcase ]
[](https://www.primefaces.org/showcase/ui/overlay/confirmDialog.xhtml?jfwid=99d70) using only the keyboard, after selecting the 'Yes' button, the focus does not return to the 'Confirm' button. This poses an accessibility issue for keyboard-only users.

### Reproducer

1. Go to the [Confirm Dialog [](showcase](https://www.primefaces.org/showcase/ui/overlay/confirmDialog.xhtml?jfwid=99d70).)
2. Use the 'Tab' key to navigate to the 'Confirm' button.
3. Activate the 'Confirm' button (usually using the 'Enter' key).
4. Use the 'Tab' key again to navigate to the 'Yes' button in the confirmation dialog.
5. Select the 'Yes' button (usually using the 'Enter' key).
6. Notice that the focus does not return to the 'Confirm' button.

### Expected behavior

After the 'Yes' button is selected, the focus should return to the 'Confirm' button.

### PrimeFaces edition

None

### PrimeFaces version

12.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

11.0.18

### Browser(s)

_No response_

melloware:
Fix #10270: Dialog/ConfirmDialog/ConfirmPopup refocus element on hide

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/confirmpopup/confirmpopup.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/confirmpopup/confirmpopup.js
@@ -15,6 +15,7 @@
  * @prop {PrimeFaces.UnbindCallback} [hideOverlayHandler] Unbind callback for the hide overlay handler.
  * @prop {JQuery} icon The DOM element for the message icon.
  * @prop {JQuery} message DOM element of the confirmation message displayed in this confirm popup.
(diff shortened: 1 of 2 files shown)


tandraschko:
+1 if well tested

PR #10272, opened 2023-06-27
Fix #2705 - Fileupload: Directory structure

### LINKED ISSUE #2705: Fileupload: Directory structure
Hello!

with the jsf 2.2 passthrough attribute , you can set following attributes to the p:fileupload component:
pt:directory="" pt:webkitdirectory="" pt:mozdirectory=""

With this it is possible to upload directories, but in the fileuploadevent just the file name exist. 
Is it possible to add the directory structure?

Sure this is a feature request!

Thanks
Phil

Rapster:
I managed to achieve this without tweaking blueimp's fileupload which is great, by adding a new header. Ideally I would have prefer to see it in the uploaded part (along with Content-Disposition/Content-Type) but can't figure out how... Because of this, this feature is only available in advanced mode, I really would like to see it in simple mode also

Few notes:

- Why use ValidationException from esapi when we have a ValidatorException?
- FileUploadUtils#checkPathTraversal only used for ImageCropper?


melloware (approved):
Nice refactoring too.

christophs78:
Look good.
Did you run the integrationtests (https://github.com/primefaces/primefaces/tree/master/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/fileupload) with uploader-native and uploader-commons? Just to be sure we don´t break something. (Our default CI-pipeline only runs the default-behaviour.)

Rapster:
By default-behaviour, I'm guessing you mean "native" uploader, PR works fine with commons

christophs78 (approved):

melloware:
@Rapster some conflicts to resolve!

Rapster:
Thank you for resolving the conflicts @melloware

PR #10282, opened 2023-07-01
Fix sonar

Rapster:


primefaces/src/main/java/org/primefaces/component/fileupload/FileUploadRenderer.java
```diff
@@ -49,8 +48,8 @@ public class FileUploadRenderer extends CoreRenderer {
     public void decode(FacesContext context, UIComponent component) {
         if (!context.getExternalContext().getRequestContentType().toLowerCase().startsWith("multipart/")) {
             if (!context.isProjectStage(ProjectStage.Production)) {
-                LOGGER.fine("Decoding FileUpload requires contentType \"multipart/form-data\". Skipping... ClientId: "
-                        + component.getClientId(context));
+                LOGGER.fine(() -> "Decoding FileUpload requires contentType \"multipart/form-data\"." +
+                        " Skipping clientId: " + component.getClientId(context));
```
melloware, line 52: I don't know why this is ringing a bell but I thought when I tried before to use this LOGGER supplier method it did not work properly?

Rapster: I assume the only thing that could go wrong would be your log conf then. I browsed history here regarding this file, and changed into severe (I hesitated to throw an exception for this...)

melloware (approved):

PR #10286, opened 2023-07-03
Fix #10283: EncodeIndexed must replace all instances of the id

### LINKED ISSUE #10283: DataTable: tabView ajax events not rendered in p:rowExpansion
### Describe the bug

I'm currently evaluating PF13-RC2 and think I found a bug in datatable / rowexpansion feature:
If I open any toggle row, the further ajax events inside the p:rowexpansion are not rendered / ignored.
In my case, I have a tabview inside the togglerow. 
Instead of switching tabs in the tabview inside the togglerow, the IDs are added to the current URL in the browser,
and the Firefox debugger console shows there are no events (like click,focus,keydown) attached to the rendered <div> 
The same code works great with PF12 (or our outdated PF7) 
A very simple example is attached as reproducer (non-lazy table with 1 column and rowexpansion)

### Reproducer

```
<p:dataTable id="simpleeinheitenDebug" value="#{einheitenBean.listeHisEinheiten}"var="einheit" lazy="false" rowKey="#{einheit.ehtId}">
    <p:column>
        <p:rowToggler/>
    </p:column>
    <p:column headerText="EHT_ID"> <h:outputText value="#{einheit.ehtId}"/></p:column>
    <p:rowExpansion>
        <p:tabView>
            <p:ajax event="tabChange" listener="#{targetEcuBean.onTabChange}"/>
            <p:tab title="Details-1">
                <h:outputText value="Tab 1"/>
            </p:tab>
            <p:tab title="Details-2">
                <h:outputText value="Tab 2"/>
            </p:tab>
        </p:tabView>
    </p:rowExpansion>
</p:dataTable>

```

### Expected behavior

in the html code which is rendered (see EDITED output below)
" <div id="simpleeinheitenDebug:0:j_idt286" ", which is the first line in the table,
as well as all <li class="ui-tabs-header ui-state-default which are the 2 tabs inside the rowexpansion show they have a "event" assigned( click,rowtoggle etc) in firefox debugger with PF12 or 7.
They are not shown and the tab switch is not working in PF13 RC2.

```
<tbody id="simpleeinheitenDebug_data" class="ui-datatable-data ui-widget-content">
<tr data-ri="0" data-rk="101601" class="ui-widget-content ui-datatable-even ui-expanded-row"></tr>
<tr class="ui-expanded-row-content ui-widget-content">
    <td colspan="2">
<div id="simpleeinheitenDebug:0:j_idt286"
             class="ui-tabs ui-widget ui-widget-content ui-corner-all ui-hidden-container ui-tabs-top"
             data-widget="widget_simpleeinheitenDebug_0_j_idt286">
            <ul class="ui-tabs-nav ui-helper-reset ui-widget-header ui-corner-all" role="tablist">
                <li class="ui-tabs-header ui-state-default ui-corner-top ui-tabs-selected ui-state-active"
                    role="tab" aria-expanded="true" aria-selected="true" data-index="0" tabindex="0"><a
                        href="#simpleeinheitenDebug:0:j_idt286:j_idt287" tabindex="-1">Details-1</a></li>
                <li class="ui-tabs-header ui-state-default ui-corner-top" role="tab" aria-expanded="false"
                    aria-selected="false" data-index="1" tabindex="0"><a
                        href="#simpleeinheitenDebug:0:j_idt286:j_idt289" tabindex="-1">Details-2</a></li>
            </ul>
            <div class="ui-tabs-panels">
                <div id="simpleeinheitenDebug:0:j_idt286:j_idt287"
                     class="ui-tabs-panel ui-widget-content ui-corner-bottom" role="tabpanel"
                     aria-hidden="false" data-index="0" style="">Tab 1
                </div>



```

### PrimeFaces edition

Community

### PrimeFaces version

13 RC2

### Theme

saga

### JSF implementation

Mojarra

### JSF version

3.0

### Java version

11, 17

### Browser(s)

Firefox 102 ESR

melloware:
Fix #10283: EncodeIndexed must replace all instances of the id

Found the issue is that i was only replacing `id=_id_5:j_id_7:0:j_id_f"` with `id="_id_5:j_id_7:0:j_id_f:0` and not all instances of that ID found as it could be used in scripts.

For example..

```js
PrimeFaces.ab({s:&quot;j_id_5:j_id_7:0:j_id_f&quot;,f:&quot;j_id_5&quot;});return false;}]);
```
Should be:
```js
PrimeFaces.ab({s:&quot;j_id_5:j_id_7:0:j_id_f:0&quot;,f:&quot;j_id_5&quot;});return false;}]);
```

### DIFF
--- a/primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
+++ b/primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
@@ -32,6 +32,8 @@ import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.util.*;
 import java.util.function.Supplier;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.el.ValueExpression;
 import javax.faces.FacesException;
@@ -67,6 +69,9 @@ public class ComponentUtils {
 
     // marker for a undefined value when a null check is not reliable enough
     private static final Object UNDEFINED_VALUE = new Object();
+
+    // regex for finding ID's
+    private static final Pattern ID_PATTERN = Pattern.compile("\\sid=\"(.*?)\"");
 
     private ComponentUtils() {
     }
@@ -632,7 +637,22 @@ public class ComponentUtils {
         // append index to all id's
         char separator = UINamingContainer.getSeparatorChar(context);
         String encodedComponent = fsw.toString();
-        encodedComponent = encodedComponent.replaceAll("\\sid=\"(.*?)\"", " id=\"$1" + separator + index + "\"");
+
+        // find all id's to replace
+        Matcher matcher = ID_PATTERN.matcher(encodedComponent);
+
+        // grab all the unique ID's found
+        Set<String> ids = new HashSet<>(5);
+        while (matcher.find()) {
+            String id = matcher.group(1);
+            if (!ids.contains(id)) {
+                ids.add(id);
+                // replace each id with an indexed version
+                String replaceId = id + separator + index;
+                encodedComponent = encodedComponent.replaceAll(id, replaceId);
+            }
+        }
+
         writer.write(encodedComponent);
     }
 


primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
```diff
@@ -632,7 +634,24 @@ public static void encodeIndexedId(FacesContext context, UIComponent component,
...
+
+        // grab all the unique ID's found
+        Set<String> ids = new HashSet<>(5);
+        while (matcher.find()) {
+
+            ids.add(matcher.group(1));
+        }
+
+        // replace each id with an indexed version
+        for (String id : ids) {
```
jepsar: Shouldn't this be an index loop? https://github.com/primefaces/primefaces/blob/master/CONTRIBUTING.md#performance-guidelines

melloware: True but its not an ArrayList its a HashSet for uniqueness and there is no way to iterate over a HashSet without an iterator?

jepsar: Ha. My bad in that case. Sorry.

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
```diff
@@ -632,7 +637,22 @@ public static void encodeIndexedId(FacesContext context, UIComponent component,
...
+        Set<String> ids = new HashSet<>(5);
+        while (matcher.find()) {
+            ids.add(matcher.group(1));
+        }
+
+        // replace each id with an indexed version
+        for (String id : ids) {
+            String replaceId = id + separator + index;
+            encodedComponent = encodedComponent.replaceAll(id, replaceId);
+        }
```
Rapster: You could use one loop instead of two no?

melloware: I was thinking that wasn't suure if manipulating the underlying string would affect the matcher while its in progress.  Let me try it.

melloware: 1 loop works!

PR #10287, opened 2023-07-03
Fix #10284: Tree better check for isSelectionEnabled

### LINKED ISSUE #10284: Tree: Missing "selection" property not handled properly and ends in unexpected NPE
### Describe the bug

Hi,
we recently ran into an issue with a NullPointerException when having a `<p:tree>` component with `selectionMode="checkbox"`, but without definition of property `selection`, as we didn't use it at this place. Interestingly the NPE did not occur on loading/rendering the page, but while doing anything on the site _after_ it got loaded, e.g. simply going back to previous view.
The NPE started to show up after upgrading from PF 10 to PF 11, and after some debugging and research I found the issue related to the changes from #3887 within TreeRenderer's `decodeSelection` implementation, which itself is calling UITree's new method `getSelectionType`: [(link)](https://github.com/primefaces/primefaces/commit/1f60dda906c252c33865fe38b7614662b8ef1308#diff-0041c3354441976f17d07d557914536101498facff3a9b0361aeccad76b287c8R1119)
```
    public Class<?> getSelectionType() {
        ValueExpression selectionVE = getValueExpression(UITree.PropertyKeys.selection.toString());
        return selectionVE == null ? null : selectionVE.getType(getFacesContext().getELContext());
    }
```
The issue is caused by trying to access `UITree.PropertyKeys.selection.toString()` with `selection` not being defined (null) and hence the NullPointerException. However, in the docs at least the `selection` property is not explicitly mentioned as required.

For me I was able to find and fix the issue on our own code, so there is no full reproducer attached, but I think it'd be valuable for others to identify the root cause with self-explaining error output and hence handle the issue better in PF's code.

### Reproducer

```
<div id="layerSwitcher" class="treeContainer">
        <h1 class="layerSwitcherTreeHeader">#{cc.attrs.controller.layerSwitcherHeaderText}</h1>
        <p:tree value="#{cc.attrs.controller.layerSwitcherRootTreeNode}" var="node" selectionMode="checkbox"
               class="layerSwitcherTree">
            <p:ajax event="select" listener="#{cc.attrs.controller.onNodeSelectLayerStyle}"/>
            <p:ajax event="unselect" listener="#{cc.attrs.controller.onNodeUnselectLayerStyle}"/>

            <p:treeNode>
                <h:outputText value="#{node.get('label')}"/>
            </p:treeNode>
        </p:tree>
    </div>
```

### Expected behavior

I would expect a respective error message by PF ("missing mandatory property") rather than a completely unhelpful NullPointerException trace with literally no information to the causing issue ...

What do you think about it? Any chance to provide better check for mandatory properties, or at least for guarding access to the `selection` property with speaking error message rather NPE?

### PrimeFaces edition

Community

### PrimeFaces version

11.0.0+

### Theme

Omega

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

11

### Browser(s)

_No response_

melloware:
Fix #10284: Tree better check for isSelectionEnabled

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/api/UITree.java
+++ b/primefaces/src/main/java/org/primefaces/component/api/UITree.java
@@ -1126,7 +1126,10 @@ public abstract class UITree extends UIComponentBase implements NamingContainer 
 
     public Class<?> getSelectionType() {
         ValueExpression selectionVE = getValueExpression(UITree.PropertyKeys.selection.toString());
-        return selectionVE == null ? null : selectionVE.getType(getFacesContext().getELContext());
+        if (selectionVE == null) {
+            throw new FacesException("Selection must be set when SelectionMode is set for Tree " + getClientId());
+        }
+        return selectionVE.getType(getFacesContext().getELContext());
     }
 
     public boolean isCheckboxSelectionMode() {


melloware:
Let me put together a reproducer.

PR #10320, opened 2023-07-07
Fix #10315: SelectOneRadio flex=true assume layout=responsive

### LINKED ISSUE #10315: SelectOneRadio: in responsive mode (flex) - flex="true" not enough to enable
### Describe the bug

Flex="true" is not enough to enable flex layout. Responsive attribute has to be specified too. 

Please see method:
org.primefaces.component.selectoneradio.SelectOneRadioRenderer#encodeMarkup

Attribute responsive defaults to 'lineDirection' if left empty. 'Cusom' or 'grid' is also checked. You have to specify some other value like 'responsive' or 'flex' in responsive attribute to enable flex support.

### Reproducer

```
<!-- flex="true" not enough to enable FLEX. It creates tabular layout in markup. -->
<p:selectOneRadio id="first" value="#{testView.string}" columns="3" flex="true" >
  <f:selectItem itemLabel="Some short text 1" itemValue="1"/>
  <f:selectItem itemLabel="Some short text 2" itemValue="2"/>
</p:selectOneRadio>

<!--  Responsive attribute has to be specified. Please see method:
      org.primefaces.component.selectoneradio.SelectOneRadioRenderer#encodeMarkup

     Attribute responsive defaults to 'lineDirection' if left empty. 'Cusom' or 'grid' is also checked.
     You have to specify some other value like 'responsive' or 'flex' in responsive attribute to enable flex support. -->
<p:selectOneRadio id="second" value="#{testView.string}" columns="3" flex="true" layout="responsive">
  <f:selectItem itemLabel="Some short text 1" itemValue="1"/>
  <f:selectItem itemLabel="Some short text 2" itemValue="2"/>
</p:selectOneRadio>
```

### Expected behavior

I would expect that flex="true" would be enough to enable flex for component

### PrimeFaces edition

Elite

### PrimeFaces version

13.0.0-RC2

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3.19

### Java version

17

### Browser(s)

all

melloware:
Fix #10315: SelectOneRadio flex=true assume layout=responsive

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/selectoneradio/SelectOneRadioRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/selectoneradio/SelectOneRadioRenderer.java
@@ -97,8 +97,11 @@ public class SelectOneRadioRenderer extends SelectOneRenderer {
         String clientId = radio.getClientId(context);
         List<SelectItem> selectItems = getSelectItems(context, radio);
         String style = radio.getStyle();
+        boolean flex = ComponentUtils.isFlex(context, radio);
+        if (flex) {
+            layout = "responsive";
+        }
         boolean lineDirection = "lineDirection".equals(layout);
-        boolean flex = !lineDirection && ComponentUtils.isFlex(context, radio);
         String styleClass = getStyleClassBuilder(context)
                 .add(lineDirection, "layout-line-direction")
                 .add(GridLayoutUtils.getResponsiveClass(flex))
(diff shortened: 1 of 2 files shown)


primefaces/src/main/java/org/primefaces/component/selectoneradio/SelectOneRadioRenderer.java
```diff
@@ -97,8 +97,11 @@ protected void encodeResponsiveLayout(FacesContext context, SelectOneRadio radio
         String clientId = radio.getClientId(context);
         List<SelectItem> selectItems = getSelectItems(context, radio);
         String style = radio.getStyle();
+        boolean flex = ComponentUtils.isFlex(context, radio);
+        if (flex) {
+            layout = "responsive";
+        }
```
jepsar, line 103: Is this something to mention in the migration guide?

melloware: Good point I just added a note to migration guide

PR #10331, opened 2023-07-10
Fix #6773 - UIData: remove Mojarra hacks 

### LINKED ISSUE #6773: UIData: remove Mojarra hacks and rowStatePreserved backport for JSF 2.0
2.1 is there since 2010 and Mojarra should be stable enough now.

### LINKED ISSUE #10328: UIData: Backport Mojarra fixes
Originally discovered in this ticket: https://github.com/primefaces/primefaces/issues/7369

See Mojarra comment: https://github.com/eclipse-ee4j/mojarra/issues/5229#issuecomment-1616684697

Rapster:
Fix #6773
Fix #10328

* attempt to fix it
* based on Thomas firs draft
* Fix mojarra bug? I had to add getClientId and getContainerClientId
* Dynamic columns not working so far (Columns is not a UIColumn, state not saved/restored)


melloware:
Looks like you got it!

melloware:
Just 1 check style violation.

Rapster:
Just realized we don't a build for mojarra 2.2?

melloware:
I think the full nightly IT does Mojarra 2.2 but not regularly on commits.

tandraschko:
@Rapster 
you said "Dynamic columns not working so far (Columns is not a UIColumn, state not saved/restored)"
is this still true? all tests are running, maybe we should add some tests before then

melloware:
@Rapster what is left on this PR before removing it from Draft?

Rapster:
I'm playing with it so I can remember what went wrong... AFAICS:

- In theory, there should be a problem with Columns saving state since it's not a UIColumn (I remember fixing a problem for DataView since xxListItem/xxGridItem were no UIColumn), but I can see filtering not working properly in both impl
- With Myfaces, if I remove the last column, the header still show but not the header

Just to name a few issue

Rapster:
I don't think we can do much about this, and the whole problem is `<p:columns />` as it is not a `UIColumn`. Simply look at `UIData#saveDescendantState()` and you'll see that even if Columns was a UIColumn more  trick would be needed to make it work (e.g call on Columns#setRowIndex)

The closest solution I have in mind to get rid of our UIData, is to use `<c:foreach />` which will add UIColumn in the JSF tree, but there are several reasons as of why it's not ideal...

I'm looking at other components like <h:selectOneListbox /> and it has a `<f:selectItems />`, if only there was something similar, it would make things so much easier for us

tandraschko:
then lets close it for now

Rapster:
for now? till when then? I was kinda hoping we could discuss more so we could figure a solution for the near future, otherwise I don't see this ever be solved

tandraschko:
feel free to create a new issue, how we could make p:columns a UIColumn
i just dont like to have open PRs which are not mergeable soon

Issue #10336, opened 2023-07-12
Captcha: Incorrect error message if context parameters are not set


bvfalcon:
### Describe the bug

When Component Captcha is used without setting context parameters error `javax.el.ELException` with stacktrace

<details>
<summary>javax.el.ELException: Expression cannot be null</summary>

```java
jakarta.servlet.ServletException: Expression cannot be null
	at jakarta.faces.api//jakarta.faces.webapp.FacesServlet.executeLifecyle(FacesServlet.java:709)
	at jakarta.faces.api//jakarta.faces.webapp.FacesServlet.service(FacesServlet.java:449)
	at io.undertow.servlet@2.3.6.Final//io.undertow.servlet.handlers.ServletHandler.handleRequest(ServletHandler.java:74)
	at io.undertow.servlet@2.3.6.Final//io.undertow.servlet.handlers.ServletChain$1.handleRequest(ServletChain.java:68)
	at io.undertow.servlet@2.3.6.Final//io.undertow.servlet.handlers.ServletDispatchingHandler.handleRequest(ServletDispatchingHandler.java:36)
	at io.undertow.servlet@2.3.6.Final//io.undertow.servlet.handlers.RedirectDirHandler.handleRequest(RedirectDirHandler.java:68)
	at io.undertow.servlet@2.3.6.Final//io.undertow.servlet.handlers.SendErrorPageHandler.handleRequest(SendErrorPageHandler.java:52)
	at io.undertow.core@2.3.6.Final//io.undertow.server.handlers.PredicateHandler.handleRequest(PredicateHandler.java:43)
	at io.undertow.servlet@2.3.6.Final//io.undertow.servlet.handlers.ServletInitialHandler.handleFirstRequest(ServletInitialHandler.java:276)
	at io.undertow.servlet@2.3.6.Final//io.undertow.servlet.handlers.ServletInitialHandler$2.call(ServletInitialHandler.java:135)
	at io.undertow.servlet@2.3.6.Final//io.undertow.servlet.handlers.ServletInitialHandler$2.call(ServletInitialHandler.java:132)
	at io.undertow.servlet@2.3.6.Final//io.undertow.servlet.core.ServletRequestContextThreadSetupAction$1.call(ServletRequestContextThreadSetupAction.java:48)
	at io.undertow.servlet@2.3.6.Final//io.undertow.servlet.core.ContextClassLoaderSetupAction$1.call(ContextClassLoaderSetupAction.java:43)
[...]

PR #10337, opened 2023-07-12
fix 10336: Captcha: Incorrect error message if context parameters are not set

bvfalcon:
FIX for [#10336](https://github.com/primefaces/primefaces/issues/10336)

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/captcha/Captcha.java
+++ b/primefaces/src/main/java/org/primefaces/component/captcha/Captcha.java
@@ -48,6 +48,7 @@ import org.primefaces.util.MessageFactory;
 @ResourceDependency(library = "primefaces", name = "components.js")
 @ResourceDependency(library = "primefaces", name = "captcha/captcha.js")
 public class Captcha extends CaptchaBase {
+
     private static final Logger LOGGER = Logger.getLogger(Captcha.class.getName());
     public static final String COMPONENT_TYPE = "org.primefaces.component.Captcha";
 
@@ -128,7 +129,8 @@ public class Captcha extends CaptchaBase {
             if (privateKey != null) {
                 privateKey = context.getApplication().evaluateExpressionGet(context, privateKey, String.class);
             }
-        } catch (ELException e) {
+        }
+        catch (ELException e) {
             LOGGER.fine(() -> "Error to process context parameter " + Captcha.PRIVATE_KEY + " as EL-expression: " + e.getMessage());
         }
 
--- a/primefaces/src/main/java/org/primefaces/component/captcha/Captcha.java
+++ b/primefaces/src/main/java/org/primefaces/component/captcha/Captcha.java
@@ -121,8 +121,7 @@ public class Captcha extends CaptchaBase {
     }
 
     private String createPostParameters(FacesContext context, Object value) throws UnsupportedEncodingException {
-        String privateKey = context.getApplication().evaluateExpressionGet(context,
-                context.getExternalContext().getInitParameter(Captcha.PRIVATE_KEY), String.class);
+        String privateKey = context.getExternalContext().getInitParameter(Captcha.PRIVATE_KEY);
 
         if (privateKey == null) {
             throw new FacesException("Cannot find private key for catpcha, use primefaces.PRIVATE_CAPTCHA_KEY context-param to define one");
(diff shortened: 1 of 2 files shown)


primefaces/src/main/java/org/primefaces/component/captcha/Captcha.java
```diff
@@ -121,8 +121,7 @@ protected void validateValue(FacesContext context, Object value) {
     }
 
     private String createPostParameters(FacesContext context, Object value) throws UnsupportedEncodingException {
-        String privateKey = context.getApplication().evaluateExpressionGet(context,
-                context.getExternalContext().getInitParameter(Captcha.PRIVATE_KEY), String.class);
+        String privateKey = context.getExternalContext().getInitParameter(Captcha.PRIVATE_KEY);
```
melloware, line 128: You still need the other step though second `context.getApplication().evaluateExpressionGet` because if someone uses an EL Expression like...

```xml
<context-param>
    <param-name>primefaces.PRIVATE_CAPTCHA_KEY</param-name>
    <param-value>#{yourBean.privateKey}</param-value>
</context-param>
```

melloware (changes requested):

primefaces/src/main/java/org/primefaces/component/captcha/Captcha.java
bvfalcon: Thanks for explanation. Reading showcase, this moment is not obvious. I has added change to PR: context-param value is processed as EL-expression, and only as fallback-strategy - used "as is"

primefaces/src/main/java/org/primefaces/component/captcha/Captcha.java
```diff
@@ -121,11 +124,18 @@ protected void validateValue(FacesContext context, Object value) {
...
+        try {
+            if (privateKey != null) {
+                privateKey = context.getApplication().evaluateExpressionGet(context, privateKey, String.class);
+            }
+        }
+        catch (ELException e) {
+            LOGGER.fine(() -> "Error to process context parameter " + Captcha.PRIVATE_KEY + " as EL-expression: " + e.getMessage());
```
melloware:
```suggestion
            LOGGER.fine(() -> "Error processing context parameter " + Captcha.PRIVATE_KEY + " as EL-expression: " + e.getMessage());
```

primefaces/src/main/java/org/primefaces/component/captcha/CaptchaRenderer.java
```diff
@@ -101,6 +104,16 @@ protected void encodeScript(FacesContext context, Captcha captcha, String public
...
-        return context.getApplication().evaluateExpressionGet(context, context.getExternalContext().getInitParameter(Captcha.PUBLIC_KEY), String.class);
+        String publicKey = context.getExternalContext().getInitParameter(Captcha.PUBLIC_KEY);
+        if (publicKey == null) {
+            return publicKey;
+        }
+        try {
+            return context.getApplication().evaluateExpressionGet(context, publicKey, String.class);
+        }
+        catch (ELException e) {
+            LOGGER.fine(() -> "Error to process context parameter " + Captcha.PUBLIC_KEY + " as EL-expression: " + e.getMessage());
```
melloware:
```suggestion
            LOGGER.fine(() -> "Error processing context parameter " + Captcha.PUBLIC_KEY + " as EL-expression: " + e.getMessage());
```

melloware (changes requested):

primefaces/src/main/java/org/primefaces/component/captcha/CaptchaRenderer.java
```diff
@@ -101,6 +104,16 @@ protected void encodeScript(FacesContext context, Captcha captcha, String public
     }
 
     protected String getPublicKey(FacesContext context, Captcha captcha) {
-        return context.getApplication().evaluateExpressionGet(context, context.getExternalContext().getInitParameter(Captcha.PUBLIC_KEY), String.class);
+        String publicKey = context.getExternalContext().getInitParameter(Captcha.PUBLIC_KEY);
+        if (publicKey == null) {
```
tandraschko: just cosmetics, i like this more:

if (publicKey != null) {
   do some elStuff
}

return publicKey;

tandraschko:
you have some checkstyle violations

Issue #6409, opened 2020-10-13
Datatable: ColumnGroup refactoring


Rapster:
**Description**
Columns grouping follows a very "unconventional" approach: 13 columns have to be defined when only 5 are displayed! My guess, is that no solution have been found at the time to render nested tr/th. Consequence of this, is the current implementation makes the code very hard to maintain at very low level for us as PF developers, but also for developers using column grouping. I can see that ColumnGroup has been giving a really hard time since it's been there 😅 

FYI: I believe other PrimeUI librairies should do the same (if possible)

**Describe the solution you would like**
Could be done 100 times simpler (colspan and rowspan automatically calculated):
```xml
                
<p:dataTable id="tblSales" var="sale" value="#{dtGroupView.sales}" columnGroupLegacyEnabled="false"
                             scrollable="#{dtGroupView.frozenColumns}" frozenColumns="#{dtGroupView.nbFrozenColumns}"
                             scrollWidth="500">

    <f:facet name="header">
        <h:outputText value="Product Sales"/>
    </f:facet>

    <p:column headerText="Product" field="manufacturer" footerText="Table footer (tfoot) using column footer facet"/>

    <p:columnGroup headerText="Sale Rate">
        <p:columnGroup headerText="Sales">
            <p:column headerText="Last Year" field="lastYearSale" footerText="A"/>
            <p:column headerText="This Year" field="thisYearSale" footerText="B"/>
        </p:columnGroup>
        <p:columnGroup headerText="Profit">
            <p:column headerText="Last Year" field="lastYearProfit" footerText="C"/>
            <p:column headerText="This Year" field="thisYearProfit" footerText="D"/>
        </p:columnGroup>
    </p:columnGroup>

    <f:facet name="tfooter">
        <p:row >
            <p:column>Table footer (tfoot) using table tfooter</p:column>
            <p:column colspan="2" style="text-align:right">Totals:</p:column>
            <p:column>
                <h:outputText value="#{dtGroupView.lastYearTotal}">
                    <f:convertNumber type="currency" currencySymbol="$"/>
                </h:outputText>
            </p:column>
            <p:column>
                <h:outputText value="#{dtGroupView.thisYearTotal}">
                    <f:convertNumber type="currency" currencySymbol="$"/>
                </h:outputText>
            </p:column>
        </p:row>
    </f:facet>

    <f:facet name="footer">
        <h:outputText value="Data between 2013-2014"/>
    </f:facet>
</p:dataTable>
```

[...]

PR #10348, opened 2023-07-16
Fix #10334 - DataTable: dynamic columns have wrong order after sorting

### LINKED ISSUE #10334: DataTable: dynamic columns have wrong order after sorting
### Describe the bug

When you use a p:dataTable with dynamic, sortable, draggable columns (p:columns) the ordering gets lost:
- change order of columns by dragging
- click on a column header to sort
Result: the headers are still in the manually changed order, but the data is in the original order

This occurs (at least) since PF 10.0.

### Reproducer

This can be reproduced with the showcase project:
- edit src/main/webapp/ui/data/datatable/columns.xhtml: add draggableColumns="true"
- execute the steps described above

### Expected behavior

Expected result: column data should be in the manually changed order (aligned with the column headers)

### PrimeFaces edition

Community

### PrimeFaces version

13.0.0

### Theme

_No response_

### JSF implementation

MyFaces

### JSF version

2.3

### Java version

17

### Browser(s)

_No response_

Rapster:
I think the whole caching columns thing and resetDynamicColumns have to be revisited. For example, Columns#dynamicColumns should not be lazy loaded, depending the rowIndex set on DataTable, dynamic column clientIds can be wrong (that was the issue).

Or there is something I might be missing... I'll tag it as a draft since the real fix requires a lot more than that . i'll think about ist once #6409 is merged

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/api/ColumnAware.java
+++ b/primefaces/src/main/java/org/primefaces/component/api/ColumnAware.java
@@ -313,10 +313,7 @@ public interface ColumnAware {
             }
             else if (child instanceof Columns) {
                 Columns uiColumns = (Columns) child;
-                for (int j = 0; j < uiColumns.getRowCount(); j++) {
-                    DynamicColumn dynaColumn = new DynamicColumn(j, uiColumns, context);
-                    columns.add(dynaColumn);
-                }
+                columns.addAll(uiColumns.getDynamicColumns());
             }
         }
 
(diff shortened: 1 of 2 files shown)


melloware (approved):

tandraschko:
I think we should also revisit our tests if enough cases are covered
I think the reset was required when columns are changing in the same request

Rapster:
Yeah that's why I'm not so comfortable merging that PR... If by any chance you can find the ticket about the issue you mentionned, that would be great

PR #10358, opened 2023-07-19
Fix #10355: SelectOneMenu trigger icon should focus panel

### LINKED ISSUE #10355: SelectOneMenu: Click chevron icon blocks searching by text
### Describe the bug

I'm using version 13 the showcase site uses 12.

I've noticed the html elements differ between the version when inspection a selectOneMenu.
The input are in 13 is a `<span>` but in 12 is a `<label>`

Furter more in 13 when clicking the triangle icon to trigger the dropdown the search doesn't work.
If I click in the input are it works as expected but it doesnt when I click the icon. Clicking the icon doesnt set the focus or applies the css class.

### Reproducer

I copied the code from the showcase site to my local code.

### Expected behavior

Clicking the icon should set the focus, apply the css class and the search should work.

### PrimeFaces edition

Community

### PrimeFaces version

13.0.0

### Theme

Default

### JSF implementation

Mojarra

### JSF version

2.0, 3.0

### Java version

17

### Browser(s)

chrome, firefox

melloware:
Fix #10355: SelectOneMenu trigger icon should focus panel


christophs78 (approved):

tandraschko:
+1 if well tested

PR #10359, opened 2023-07-19
Fix #9616: Remove Watermark

### LINKED ISSUE #9616: Watermark: remove
since JSF 2.2 we have passthrough attributes and HTML placeholder is supported in all newer browsers

so the question is: when should we remove p:watermark for JSF2.0 and 2.1 users?

melloware:
Fix #9616: Remove Watermark


docs/14_0_0/gettingstarted/whatsnew.md
jepsar, line 1: Shouldn't this be changed?

melloware: Thanks fixed!

PR #10363, opened 2023-07-20
Fix #10362: Datatable allow multiple SummaryRow

### LINKED ISSUE #10362: Datatable: Allow multiple SummaryRows
@Rapster we just got this requirement in from a client. We are replacing an old desktop PowerBuilder application (yes PowerBuilder) and it has double SummaryRow's like in this screenshot.

image

When we tried to do 2 summaryRow we realized it only allowed for 1 so we couldn't render this result the way we wanted.  I am submitting a PR to allow for multiple SummaryRow.

melloware:
Fix #10362: Datatable allow multiple SummaryRow

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java
+++ b/primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java
@@ -664,15 +664,16 @@ public class DataTable extends DataTableBase {
         getStateHelper().put(InternalPropertyKeys.selectAll, selectAll);
     }
 
-    public SummaryRow getSummaryRow() {
+    public List<SummaryRow> getSummaryRows() {
+        List<SummaryRow> sumRows = new ArrayList<>(3);
         for (int i = 0; i < getChildCount(); i++) {
             UIComponent kid = getChildren().get(i);
             if (kid.isRendered() && kid instanceof SummaryRow) {
-                return (SummaryRow) kid;
+                sumRows.add((SummaryRow) kid);
             }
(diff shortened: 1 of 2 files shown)


Rapster (approved):

Rapster:
should it be 13.0.1 or 14.0.0 as it's no defect?

melloware:
I think its up to us. I need this for a client so normally i would ask to be patched through their PRO account to 13.0.1 but i think all of that is changing?

Rapster:
All i know is that, if we follow sem version, this fix should go to 13.1.0

melloware:
yep you are right let me move to 14.0.0 for now until we can discuss further on Discord?

tandraschko:
yep, i dont think we will have something like 13.1.0

primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java
```diff
@@ -664,15 +664,16 @@ public void setSelectAll(boolean selectAll) {
         getStateHelper().put(InternalPropertyKeys.selectAll, selectAll);
     }
 
-    public SummaryRow getSummaryRow() {
+    public List<SummaryRow> getSummaryRow() {
```
tandraschko: getSummaryRows

Rapster:
Should we go back to 2 digits then? PTek did the switch in 10 and always turned out to be useless,

tandraschko:
actually yes

PR #10366, opened 2023-07-20
Fix #10365: ImageCropper originalFilename for streamed content

### LINKED ISSUE #10365: ImageCropper: originalFilename does not match the name attribute of DefaultStreamedContent
### Describe the bug

`CroppedImage `object has an `originalFilename` attribute which is the name of the image that has been cropped.

If one is cropping an image which is obtained via a `StreamContent` rather than a static file, then there is no filename as such. Conveniently, a name can still be set to the `StreamContent`:

`StreamedContent imageAsStream = DefaultStreamedContent.builder()**.name("example")**.contentType("image/png").stream(() -> stream).build();`

yet, when one calls the method `getOriginalFileName `on the `CroppedImage`, it produces a `String `which is the result of the `toString()` method of the original `StreamContent`, instead of returning the `name` attribute that had been set in the builder.

Use case: I need to match the original pic with the cropped pic, and it would make my life easier if I could just use their names to match them.

### Reproducer

This case is deeply embedded in an app, can't create a reproducer.

### Expected behavior

The `originalFilename `attribute of `CroppedImage `should match the `name `attribute of the image that was cropped and which originated from a `DefaultStreamContent`.

### PrimeFaces edition

Community

### PrimeFaces version

12.0

### Theme

default theme

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

17

### Browser(s)

_No response_

melloware:
Fix #10365: ImageCropper originalFilename for streamed content

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/imagecropper/ImageCropperRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/imagecropper/ImageCropperRenderer.java
@@ -206,7 +206,20 @@ public class ImageCropperRenderer extends CoreRenderer {
             String format = guessImageFormat(stream.getContentType(), stream.getName());
             ImageIO.write(cropped, format, croppedOutImage);
 
-            return new CroppedImage(cropper.getImage().toString(), croppedOutImage.toByteArray(), x, y, w, h);
+            Object imageObject = cropper.getImage();
+            String originalFileName;
+            if (imageObject instanceof String) {
+                originalFileName = (String) imageObject;
+            }
+            else if (imageObject instanceof StreamedContent) {
+                StreamedContent streamedContentTmp = (StreamedContent) imageObject;
+                originalFileName = streamedContentTmp.getName();
+            }
+            else {
+                throw new IllegalArgumentException("ImageCropper 'image' must be either a String relative path or a StreamedObject.");
+            }
+
+            return new CroppedImage(originalFileName, croppedOutImage.toByteArray(), x, y, w, h);
         }
         catch (IOException e) {
             throw new ConverterException(e);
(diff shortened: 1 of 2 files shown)


primefaces/src/main/java/org/primefaces/component/imagecropper/ImageCropperRenderer.java
```diff
@@ -206,7 +206,20 @@ public Object getConvertedValue(FacesContext context, UIComponent component, Obj
...
+            }
+            else if (imageObject instanceof StreamedContent) {
+                StreamedContent streamedContentTmp = (StreamedContent) imageObject;
+                originalFileName = streamedContentTmp.getName();
+            }
+            else {
+                throw new IllegalArgumentException("ImageCropper 'image' must be either a String relative path or a StreamedObject.");
+            }
+
+            return new CroppedImage(originalFileName, croppedOutImage.toByteArray(), x, y, w, h);
```
Rapster, line 222: Can it be done in createStreamedContentForCropping method?

Rapster:
This can go in 13.0.1 no?

melloware:
Definitely

tandraschko:
+1

primefaces/src/main/java/org/primefaces/component/imagecropper/ImageCropperRenderer.java
Rapster: @melloware doesn't seem you saw this coment, but why not simply call:
```java
 return new CroppedImage(stream.getName(), croppedOutImage.toByteArray(), x, y, w, h);`
```

Rapster:
Whether it's a String or a StreamContent, it gets wrapped into a StreamContent, that's why I mentionned createStreamedContentForCropping

PR #10370, opened 2023-07-23
Fix #10365: ImageCropper originalFilename for streamed content

### LINKED ISSUE #10365: ImageCropper: originalFilename does not match the name attribute of DefaultStreamedContent

melloware:
Fix #10365: ImageCropper originalFilename for streamed content

Give the file a default name instead of NULL if the user forgets to set the name on the stream.

### DIFF
--- a/docs/13_0_0/components/imagecropper.md
+++ b/docs/13_0_0/components/imagecropper.md
@@ -69,7 +69,7 @@ about the crop process. Following table describes CroppedImage properties.
 
 | Property | Type | Description
 | --- | --- | --- |
-originalFileName | String | Name of the original file that’s cropped
+originalFileName | String | Name of the original file that’s cropped. If using StreamedContent it will be the name set on the stream or `unknown.png` if NULL.
 bytes | byte[] | Contents of the cropped area as a byte array
 left | int | Left coordinate
 right | int | Right coordinate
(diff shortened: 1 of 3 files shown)


seinecle:
Great thanks a lot!

PR #10384, opened 2023-07-25
Fix #10378: OutputLabel associate SelectCheckBoxMenu

### LINKED ISSUE #10378: SelectCheckboxMenu: Can't associate label to input
### Describe the bug

I am using a selectCheckboxMenu, but can't associate a label to it.
p:outputLabel and the label attribute don't work, and also accessibility still requires a label for this

### Reproducer

`
                        <p:outputLabel for="@next" value="Label:"/>
                        <p:selectCheckboxMenu ... label="label" ...
`

### Expected behavior

The label should be linked to the component, and also on click on the label the component should be selected (like on other inputs)

### PrimeFaces edition

None

### PrimeFaces version

13.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.x

### Java version

11

### Browser(s)

_No response_

melloware:
Fix #10378: OutputLabel associate SelectCheckBoxMenu

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenu.java
+++ b/primefaces/src/main/java/org/primefaces/component/selectcheckboxmenu/SelectCheckboxMenu.java
@@ -98,6 +98,26 @@ public class SelectCheckboxMenu extends SelectCheckboxMenuBase {
         return DEFAULT_EVENT;
     }
 
+    @Override
+    public String getInputClientId() {
+        return getClientId(getFacesContext()) + "_focus";
+    }
+
+    @Override
+    public String getValidatableInputClientId() {
+        return getClientId(getFacesContext());
+    }
+
+    @Override
+    public String getLabelledBy() {
+        return (String) getStateHelper().get("labelledby");
+    }
+
+    @Override
+    public void setLabelledBy(String labelledBy) {
+        getStateHelper().put("labelledby", labelledBy);
+    }
+
     public List<Column> getColumns() {
         List<Column> columns = new ArrayList<>();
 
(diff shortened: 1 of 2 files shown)


melloware:
Not sure if I love this but it does allow it to be associated correctly and clicking on the label focuses the SelectCheckBoxMenu correctly now.

primefaces/src/main/java/org/primefaces/component/outputlabel/OutputLabelRenderer.java
```diff
@@ -83,7 +83,11 @@ public void invokeContextCallback(FacesContext context, UIComponent target) {
                         inputHolder.setLabelledBy(clientId);
                     }
                     else {
-                        state.setClientId(target.getClientId(context));
+                        String clientId = target.getClientId(context);
+                        if (target instanceof SelectCheckboxMenu) {
+                            clientId = clientId + "_focus";
```
jepsar:
```suggestion
                            clientId += "_focus";
```

primefaces/src/main/java/org/primefaces/component/outputlabel/OutputLabelRenderer.java
```diff
@@ -83,7 +83,11 @@ public void invokeContextCallback(FacesContext context, UIComponent target) {
                         inputHolder.setLabelledBy(clientId);
                     }
                     else {
-                        state.setClientId(target.getClientId(context));
+                        String clientId = target.getClientId(context);
+                        if (target instanceof SelectCheckboxMenu) {
```
jepsar: Don't we have several components we have to correct for? Shouldn't we create an interface from which we can get the focus clientId?

melloware: So far this looks like the only one.  The others like InputNumber are handled more gracefully I think but this SelectCheckBoxMenu seems to be a special beast.  But we probably need to evaluate each component with a `Label` and click on it to make sure it is focused etc.

tandraschko: dont we have InputHolder for this?

melloware: Oh I just realized that is what it is for!  Let me look

melloware:
@tandraschko you rock.  Did not realize this was all already in place this was an easy change and works great.

PR #10409, opened 2023-07-28
Fixes #10404 - Exporter: Allow user can export a custom value type other than string

### LINKED ISSUE #10404: Exporter: Allow export a custom value type other than string
### Description

Actually with stronglyTypedCells = true, PrimeFaces tries to guess the type of the cell value from a string value. I suggest to allow the user could give the actual value with the appropriate java type in order PrimeFaces could handle more accurately instead of trying to guess. That would allow us to avoid exporting some cells as numeric and some others as text in a single column with Excel  Exporter (see image).

image

By using already provided ways to set custom export values in <p:column> tag like attributes `exportValue` and `exportFunction` one could set custom objects as values. By this way we could send Number or Date objects directly so PrimeFaces don't have to guess from string values.

Also in the future we could set some custom CellValueExporter (a hypothetical new interface) object in `ExcelOptions` to allow the user could customize `org.apache.poi.ss.usermodel.Cell` value.

### Additional context

[errror_reproducer.zip](https://github.com/primefaces/primefaces/files/12196012/errror_reproducer.zip)

### LINKED ISSUE #10464: Exporter: Wrong value in XLS export for Date / Calender
### Describe the bug

Exporting a datatable column with a Date field results in a wrong value in XLS when using convertDateTime with pattern "dd.MM.yyyy"
```
<p:column headerText="Date">
    <h:outputText value="#{product.created.time}">
        <f:convertDateTime pattern="dd.MM.yyyy"/>
    </h:outputText>
</p:column>
```

PDF and CSV export works as expected.

XLS Export works as expected when using pattern "yyyy-MM-dd"

Here the exported documents:
[products.csv](https://github.com/primefaces/primefaces/files/12292164/products.csv)
[products.pdf](https://github.com/primefaces/primefaces/files/12292165/products.pdf)
[products.xls](https://github.com/primefaces/primefaces/files/12292166/products.xls)

Issue also appears in PF 13.0.0

### Reproducer

Here is the reproducer using PF 12.0.4
[primefaces-test-export.zip](https://github.com/primefaces/primefaces/files/12292168/primefaces-test-export.zip)

### Expected behavior

XLS export should also show correct value for Date column.

### PrimeFaces edition

Elite

### PrimeFaces version

12.0.4

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

11

### Browser(s)

_No response_

jasonex7:
Fixes #10404

Fix #10464

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/export/ColumnValue.java
+++ b/primefaces/src/main/java/org/primefaces/component/export/ColumnValue.java
@@ -1,0 +1,64 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2023 PrimeTek Informatics
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.component.export;
+
+public class ColumnValue {
+    private final Object customValue;
+    private final String fallbackValue;
+
+    private ColumnValue(Object customValue, String fallbackValue) {
+        this.customValue = customValue;
+        this.fallbackValue = fallbackValue;
+    }
+
+    public static ColumnValue customValue(Object value) {
+        return new ColumnValue(value, null);
+    }
+
+    public static ColumnValue fallbackValue(String value) {
+        return new ColumnValue(null, value);
+    }
+
+    public Object getCustomValue() {
+        return customValue;
+    }
+
+    public String getFallbackValue() {
+        return fallbackValue;
+    }
+
+    public boolean isCustomValue() {
+        return customValue != null;
+    }
+
+    @Override
+    public String toString() {
+        if (customValue == null && fallbackValue == null) {
+            return null;
+        }
+        else {
+            return customValue != null ? customValue.toString() : fallbackValue;
+        }
+    }
+}
--- a/primefaces/src/main/java/org/primefaces/component/export/ColumnValue.java
+++ b/primefaces/src/main/java/org/primefaces/component/export/ColumnValue.java
@@ -23,6 +23,8 @@
  */
 package org.primefaces.component.export;
 
+import java.util.Objects;
+
 public class ColumnValue {
     private final Object customValue;
     private final String fallbackValue;
@@ -50,6 +52,24 @@ public class ColumnValue {
 
     public boolean isCustomValue() {
         return customValue != null;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        ColumnValue that = (ColumnValue) o;
+
+        if (!Objects.equals(customValue, that.customValue)) return false;
+        return Objects.equals(fallbackValue, that.fallbackValue);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = customValue != null ? customValue.hashCode() : 0;
+        result = 31 * result + (fallbackValue != null ? fallbackValue.hashCode() : 0);
+        return result;
     }
 
     @Override
--- a/primefaces/src/main/java/org/primefaces/component/export/ColumnValue.java
+++ b/primefaces/src/main/java/org/primefaces/component/export/ColumnValue.java
@@ -28,15 +28,15 @@ import org.primefaces.util.Constants;
 import java.util.Objects;
 
 public class ColumnValue {
-    private final Object customValue;
+    private final Object explicitlySetValue;
     private final String fallbackValue;
 
-    private ColumnValue(Object customValue, String fallbackValue) {
-        this.customValue = customValue;
+    private ColumnValue(Object explicitlySetValue, String fallbackValue) {
+        this.explicitlySetValue = explicitlySetValue;
         this.fallbackValue = Objects.toString(fallbackValue, Constants.EMPTY_STRING);
     }
 
-    public static ColumnValue customValue(Object value) {
+    public static ColumnValue explicitlySetValue(Object value) {
         return new ColumnValue(value, null);
     }
 
@@ -44,16 +44,16 @@ public class ColumnValue {
         return new ColumnValue(null, value);
     }
 
-    public Object getCustomValue() {
-        return customValue;
+    public Object getExplicitlySetValue() {
+        return explicitlySetValue;
     }
 
     public String getFallbackValue() {
         return fallbackValue;
     }
 
-    public boolean isCustomValue() {
-        return customValue != null;
+    public boolean isExplicitlySetValue() {
+        return explicitlySetValue != null;
     }
 
     @Override
@@ -63,24 +63,24 @@ public class ColumnValue {
 
         ColumnValue that = (ColumnValue) o;
 
-        if (!Objects.equals(customValue, that.customValue)) return false;
+        if (!Objects.equals(explicitlySetValue, that.explicitlySetValue)) return false;
         return Objects.equals(fallbackValue, that.fallbackValue);
     }
 
     @Override
     public int hashCode() {
-        int result = customValue != null ? customValue.hashCode() : 0;
+        int result = explicitlySetValue != null ? explicitlySetValue.hashCode() : 0;
         result = 31 * result + (fallbackValue != null ? fallbackValue.hashCode() : 0);
         return result;
     }
 
     @Override
     public String toString() {
-        if (customValue == null && fallbackValue == null) {
+        if (explicitlySetValue == null && fallbackValue == null) {
             return Constants.EMPTY_STRING;
         }
         else {
-            return Objects.toString(customValue != null ? customValue.toString() : fallbackValue, Constants.EMPTY_STRING);
+            return Objects.toString(explicitlySetValue != null ? explicitlySetValue.toString() : fallbackValue, Constants.EMPTY_STRING);
         }
     }
 }
--- a/primefaces/src/main/java/org/primefaces/component/export/ColumnValue.java
+++ b/primefaces/src/main/java/org/primefaces/component/export/ColumnValue.java
@@ -23,6 +23,8 @@
  */
 package org.primefaces.component.export;
 
+import org.primefaces.util.Constants;
+
 import java.util.Objects;
 
 public class ColumnValue {
@@ -31,7 +33,7 @@ public class ColumnValue {
 
     private ColumnValue(Object customValue, String fallbackValue) {
         this.customValue = customValue;
-        this.fallbackValue = fallbackValue;
+        this.fallbackValue = Objects.toString(fallbackValue, Constants.EMPTY_STRING);
     }
 
     public static ColumnValue customValue(Object value) {
@@ -75,10 +77,10 @@ public class ColumnValue {
     @Override
     public String toString() {
         if (customValue == null && fallbackValue == null) {
-            return null;
+            return Constants.EMPTY_STRING;
         }
         else {
-            return customValue != null ? customValue.toString() : fallbackValue;
+            return Objects.toString(customValue != null ? customValue.toString() : fallbackValue, Constants.EMPTY_STRING);
         }
     }
 }
--- a/primefaces/src/main/java/org/primefaces/component/export/ColumnValue.java
+++ b/primefaces/src/main/java/org/primefaces/component/export/ColumnValue.java
@@ -28,32 +28,25 @@ import org.primefaces.util.Constants;
 import java.util.Objects;
 
 public class ColumnValue {
-    private final Object explicitlySetValue;
-    private final String fallbackValue;
+    public static final ColumnValue EMPTY_VALUE = ColumnValue.of(Constants.EMPTY_STRING);
 
-    private ColumnValue(Object explicitlySetValue, String fallbackValue) {
-        this.explicitlySetValue = explicitlySetValue;
-        this.fallbackValue = Objects.toString(fallbackValue, Constants.EMPTY_STRING);
+    private final Object value;
+
+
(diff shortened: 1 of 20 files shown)


melloware:
Asking @Rapster to review this also as he is very familiar with this as well.  So far this would be a great change just want to make sure its not going to break anything.

primefaces/src/main/java/org/primefaces/component/export/ColumnValue.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public String getFallbackValue() {
+        return fallbackValue;
+    }
+
+    public boolean isCustomValue() {
+        return customValue != null;
+    }
+
```
melloware, line 66: Should probably implement equals() and hashcode() to be a good citizen

primefaces/src/main/java/org/primefaces/component/export/ColumnValue.java
```diff
@@ -0,0 +1,64 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.component.export;
+
+public class ColumnValue {
```
melloware, line 30: Should this be Serializable?

primefaces/src/main/java/org/primefaces/util/ExcelStylesManager.java
```diff
@@ -251,6 +338,51 @@ private CellStyle createCurrencyStyle() {
...
 
+    private CellStyle createDateStyle() {
+        CellStyle style = createDefaultCellStyle();
+        short dataFormat;
+        if (options.getExcelDateFormat() != null) {
+            String format = options.getExcelDateFormat();
+            dataFormat = wb.getCreationHelper().createDataFormat().getFormat(format);
+        }
+        else {
+            dataFormat = 14;
```
melloware: Where did this 14 come from?

primefaces/src/main/java/org/primefaces/util/ExcelStylesManager.java
```diff
@@ -251,6 +338,51 @@ private CellStyle createCurrencyStyle() {
...
+
+    private CellStyle createDateTimeStyle() {
+        CellStyle style = createDefaultCellStyle();
+        short dataFormat;
+        if (options.getExcelDateTimeFormat() != null) {
+            String format = options.getExcelDateTimeFormat();
+            dataFormat = wb.getCreationHelper().createDataFormat().getFormat(format);
+        }
+        else {
+            dataFormat = 22;
```
melloware: Same here we should Javadoc where these constants came from?

melloware (changes requested):

primefaces/src/main/java/org/primefaces/component/export/ColumnValue.java
jasonex7: I don't think is necessary, because that object is not saved anywhere.

primefaces/src/main/java/org/primefaces/util/ExcelStylesManager.java
jasonex7: It's a excel built-int format, documented in `org.apache.poi.ss.usermodel.BuiltinFormats`

primefaces/src/main/java/org/primefaces/util/ExcelStylesManager.java
jasonex7: I'm going to change the code to clarify where those values come from.

primefaces/src/main/java/org/primefaces/component/export/ColumnValue.java
jasonex7: done.

primefaces/src/main/java/org/primefaces/component/export/ColumnValue.java
```diff
@@ -0,0 +1,64 @@
...
+        this.fallbackValue = fallbackValue;
+    }
+
+    public static ColumnValue customValue(Object value) {
+        return new ColumnValue(value, null);
+    }
+
+    public static ColumnValue fallbackValue(String value) {
+        return new ColumnValue(null, value);
+    }
```
Rapster: Instead of having those 2 methods:

```java
public class ColumnValue {
        private final Object value;

        private ColumnValue(Object value) {
            this.value = value;
        }

        public static ColumnValue of(Object value) {
            return new ColumnValue(value);
        }

        public boolean isComplexValue() {
            return !isStringValue();
        }

        public boolean isStringValue() {
            return value instanceof String;
        }

        @Override
        public boolean equals(Object o) {
            if (this == o) return true;
            if (o == null || getClass() != o.getClass()) return false;
            ColumnValue that = (ColumnValue) o;
            return Objects.equals(value, that.value);
        }

        @Override
        public int hashCode() {
            return Objects.hash(value);
        }

        @Override
        public String toString() {
            return Objects.toString(value, Constants.EMPTY_STRING);
        }
    }
```

Makes you wonder how useful is this class... but it's better than working with Object type, and in the future maybe this class could be enriched? Also, toString() becomes quite handy here

primefaces/src/main/java/org/primefaces/util/ExcelStylesManager.java
```diff
@@ -87,14 +95,44 @@ public static ExcelStylesManager createExcelStylesManager(Workbook wb, Locale lo
...
-    public void updateCell(UIColumn column, Cell cell, String value) {
-        updateCell(cell, value);
+    public void updateCell(UIColumn column, Cell cell, ColumnValue value) {
+        if (value.isCustomValue()) {
+            updateCellCustomValue(cell, value.getCustomValue());
+        }
+        else {
+            updateCellFallbackValue(cell, value.getFallbackValue());
+        }
         applyColumnAlignments(column, cell);
```
Rapster: If the exportValue is a String, it is considered as a custom value when it's not, this could cause regressions for some users. You should use value.isComplexType instead, and that's what updateCellCustomValue will do, only deal with non String value

Rapster (changes requested):

primefaces/src/main/java/org/primefaces/util/ExcelStylesManager.java
jasonex7: That's not my intended use, I want to be capable of tell PrimeFaces I'm submitting a String value so ExcelExporter could treat like that and stop trying to guess what type is when stronglyTypedCells = true. I want to handle as customValue if the value was set with `exportValue` or `exportFunction` attribute in `<p:column>` despite of what type the value is. In all other cases, behave like before (ExcelExporter tries to figure what type is from the rendered value of the `<p:column>`)

Mabe I could rename the property as `explicitlySetValue` instead of `customValue`. I think that would show the real intention behind. What do you think?

primefaces/src/main/java/org/primefaces/component/export/ColumnValue.java
jasonex7: I want to try the value as custom if the user set it in the attributes `exportValue` or `exportFunction` no matter if is String, LocalDate, BigDecimal or whatever type. So ExcelExporter could handle appropriately, please see my example in https://github.com/primefaces/primefaces/issues/10404#issue-1826639737

Rapster: I understand that, but because you set exportValue or exportFunction doesn't necessarly mean it's a custom value, you might use that just because the column content is not supported by PF, example:

```xml
<p:column exportValue="#{car.color}>
    <div>
          <div>
              #{car.color}
          </div>
    </div>
</p:column>
```

If value is a String, you have to assume it's been properly formatted or formatted in a way that with stronglyTypedCells will try something smart. I understand you want to get rid of that tweak, but this is breaking change, for now stronglyTypedCells can simply be used as a fallback

primefaces/src/main/java/org/primefaces/util/ExcelStylesManager.java
Rapster: First 2 lines will cause regression as explained before. Don't think custom/fallback have a lot of meaning here, it's more whether it's the raw value (which will require conversion) or string (where the conversion has already been made, but stronglyTypedCells  should still be taken account to remain backward compatible)

primefaces/src/main/java/org/primefaces/component/export/ColumnValue.java
melloware: I agree with @Rapster. I actually do this in a few places where I may have custom format inside the column but the `exportValue` I use is a different field in the POJO that may be just a String that I want formatted by Strongly Typed Cells.

jasonex7: I've already change the name of the variable to "explicitlySetValue" instead of customValue to reflect what I want to mean. The problem with the actual version of PF is when I want to set some column as String and stronglyTypeCell tries to convert to number and strip characters and the excel column has some rows as text and some others as numbers, that is a bug for me. What do you suggest to solve that without stronglyTypeCell = false?

primefaces/src/main/java/org/primefaces/util/ExcelStylesManager.java
jasonex7: I've already change the name of the variable to "explicitlySetValue" instead of customValue to reflect what I want to mean. For now stronglyTypedCells is considered when the user don't use `exportValue` or `exportFunction` attributes of `<p:column>`, otherwise treats the value with type of the object. Maybe I could add another <p:column> attribute which permits disabling stronglyTypedCells per column basis.

melloware:
I have marked this as Breaking Change for now.

primefaces/src/main/java/org/primefaces/util/ExcelStylesManager.java
jasonex7: I think that new attribute in `<p:column>` could be called `exportOptions` and it will be a new interface similar to `org.primefaces.component.export.ExporterOptions` but for columns instead of whole DataTable.

primefaces/src/main/java/org/primefaces/component/export/ColumnValue.java
Rapster: If I understand well, you have a String as input (which look like a number) and is exported as a number: that's what stronglyTypeCell does. Now if you don't want that, then simply disable stronglyTypeCell. There is no bug here so far

jasonex7: I'm sorry there is no bug, I mean it's a limitation of stronglyTypedCell. I don't want to disable for entire DataTable just for a single column that it is a String no matter how if looks like number, for example a national ID which starts with "0"

melloware: So maybe on each column a new `exportStronglyTyped` boolean to allow it to be controlled per column?

Rapster: @jasonex7 @melloware  you can create a map and make custom options per column

jasonex7: I think that new attribute in `<p:column>` could be called `exportOptions` and it will be a new interface similar to `org.primefaces.component.export.ExporterOptions` but for columns instead of whole DataTable. That will allow more config in the future. I don't like the map option because it would be too much generic.

Rapster: Not sure what I had in mind at that time... Anyway, just to be clear, from what I understand you're trying to solve 2 issues no? First is, export raw values (as they are and no longer as String) and the other is to have, export options per column right?

jasonex7: I want primefaces receive the real value with exportValue or exportFunction properties in <p:colum> and handle that accordingly in exporter. But I understand that there will be a breaking change with String values set in that properties because now it's handle with stronglyTypeCell feature and with this PR as plain String. So in order to not introduce that breaking change, I suggest to add exportOptions property with stronglyTypedCell = true as default to preserve the current behavior.

Rapster: > But I understand that there will be a breaking change with String values set in that properties

No, if a string is set, it's handled as it was before. Why this should be suddenly any different?

jasonex7: You're right, I see now what I have to do. I'm going to change this PR that in case stronglyTypedCell = true and the value set was a String then it behave like it was before, any other case is handle by this PR proposal. With that there will be no breaking change.

jasonex7:
With the last commit when value set is String it consider stronglyTypedCell exporter option. So in my project I only have to disable that setting in order to treat String as String and other objects accordingly to their type. 

> I have marked this as Breaking Change for now.

@melloware please remove breaking change label.

melloware:
Label removed!

melloware (approved):

Rapster:
I'll review in 2 weeks

primefaces/src/main/java/org/primefaces/component/export/ExporterUtils.java
```diff
@@ -185,28 +183,38 @@ public static String getComponentFacetValue(FacesContext context, UIComponent pa
         return null;
     }
 
-    public static String getColumnFacetValue(FacesContext context, UIColumn column, TableExporter.ColumnType columnType) {
-        String textValue = null;
+    public static ColumnValue getColumnFacetValue(FacesContext context, UIColumn column, TableExporter.ColumnType columnType) {
+        ColumnValue columnValue = ColumnValue.EMPTY_VALUE;
         if (columnType == TableExporter.ColumnType.HEADER) {
-            textValue = Optional.ofNullable(column.getExportHeaderValue()).orElseGet(column::getHeaderText);
+            if (column.getExportHeaderValue() != null) {
```
Rapster: Keep the original code, you can do the same in 2 lines

primefaces/src/main/java/org/primefaces/component/export/ExporterUtils.java
```diff
@@ -185,28 +183,38 @@ public static String getComponentFacetValue(FacesContext context, UIComponent pa
...
+                columnValue = ColumnValue.of(column.getExportHeaderValue());
+            }
+            else {
+                columnValue = ColumnValue.of(column.getHeaderText());
+            }
         }
         else if (columnType == TableExporter.ColumnType.FOOTER) {
-            textValue =  Optional.ofNullable(column.getExportFooterValue()).orElseGet(column::getFooterText);
+            if (column.getExportFooterValue() != null) {
```
Rapster: Keep the original code, you can do the same in 2 lines

primefaces/src/main/java/org/primefaces/component/export/ExporterUtils.java
```diff
@@ -185,28 +183,38 @@ public static String getComponentFacetValue(FacesContext context, UIComponent pa
...
+            }
+            else {
+                columnValue = ColumnValue.of(column.getFooterText());
+            }
         }
 
         UIComponent facet = column.getFacet(columnType.facet());
-        if (LangUtils.isBlank(textValue) && ComponentUtils.shouldRenderFacet(facet)) {
-            textValue = getComponentValue(context, facet);
+        if (LangUtils.isBlank(columnValue.toString()) && ComponentUtils.shouldRenderFacet(facet)) {
```
Rapster, line 201: You could introduce `ColumnValue#isBlank` since ColumnValue should never be null

Rapster (changes requested):
Looks much better now, I left few notes

melloware:
@Rapster i fixed your comments and merged.

Issue #10333, opened 2023-07-11
Locale: Client Locale updates


melloware:
- [x] AutoComplete move server locale to client https://github.com/primefaces/primefaces/pull/10415
- [x] Paginator move server locale to client https://github.com/primefaces/primefaces/pull/10417
- [x] Picklist move server locale to client https://github.com/primefaces/primefaces/pull/10418
- [x] Check all server side messages which probably should be just Locale client messages.
- [x]  Rename `firstDay` to `firstDayOfWeek` to align with PrimeLocale https://github.com/primefaces/primefaces/pull/10416
- [x] Try and use or re-use PrimeLocale client locales as used by PrimeReact/Vue/Ng: https://github.com/primefaces/primelocale

PR #10415, opened 2023-07-30
#10333: Autocomplete move server labels to client labels

melloware:
This is the first phase of many to move server labels to client labels and align with PrimeLocale labels


tandraschko:
conflict

PR #10416, opened 2023-07-30
#10333: Locale rename firstDay to firstDayOfWeek

melloware:
This makes it consistent with other Prime libs


tandraschko:
conflict

PR #10419, opened 2023-07-31
Fix #10333: Use PrimeLocale

melloware:
Fix #10333: Use PrimeLocale

1. I always import PrimeLocale first at the top of the file untouched.
2. Then I merge in custom PF locale stuff. That way as PrimeLocale updates or changes we can just drop those changes in without much work!

A good example to look at is: https://github.com/primefaces/primefaces/blob/83dc572f03cf1ed039c8310e97e485a125eae4f8/primefaces/src/main/resources/META-INF/resources/primefaces/locales/locale-nl.js


jepsar (commented):
Looks okay to me.

PR #10424, opened 2023-08-01
Fix #10333: DataTable/CheckBoxMenu Select All Checkbox ARIA

melloware:
Fix #10333: DataTable/CheckBoxMenu Select All Checkbox ARIA


melloware:
This now puts it in line with other Prime Libs.

PR #10449, opened 2023-08-04
Fix #10444: Client Window Factory

### LINKED ISSUE #10444: JSF 2.2 removal: use client-window-factory
in the past we used lifecyclefactory to have JSF 2.2 compatiblity, now we can use client-window-factory directly
it seems it doesnt work at least in Mojarra 2.3, Myfaces 2.3 works fine

@melloware are you interessted to check that?

simple add a PrimeClientWindowFactory and return new PrimeClientWindow + register it in faces-config

melloware:
Fix #10444: Client Window Factory

### DIFF
--- a/primefaces/src/main/java/org/primefaces/clientwindow/PrimeClientWindowFactory.java
+++ b/primefaces/src/main/java/org/primefaces/clientwindow/PrimeClientWindowFactory.java
@@ -25,31 +25,20 @@ package org.primefaces.clientwindow;
 
 import javax.faces.context.FacesContext;
 import javax.faces.lifecycle.ClientWindow;
-import javax.faces.lifecycle.Lifecycle;
-import javax.faces.lifecycle.LifecycleWrapper;
+import javax.faces.lifecycle.ClientWindowFactory;
 
-// We can't use a ClientWindowFactory as this is registered in the faces-config.xml
-// if we would add the client-window-factory entry, pre JSF2.3 would crash
-public class PrimeClientWindowLifecycle extends LifecycleWrapper {
+public class PrimeClientWindowFactory extends ClientWindowFactory {
 
-    public PrimeClientWindowLifecycle(Lifecycle wrapped) {
+    public PrimeClientWindowFactory(ClientWindowFactory wrapped) {
         super(wrapped);
     }
 
     @Override
-    public void attachWindow(FacesContext facesContext) {
-        ClientWindow clientWindow = facesContext.getExternalContext().getClientWindow();
+    public ClientWindow getClientWindow(FacesContext context) {
+        ClientWindow clientWindow = context.getExternalContext().getClientWindow();
         if (clientWindow == null) {
             clientWindow = new PrimeClientWindow();
         }
-
-        try {
-            clientWindow.decode(facesContext);
-            facesContext.getExternalContext().setClientWindow(clientWindow);
-        }
-        catch (RuntimeException e) {
-            facesContext.getExternalContext().setClientWindow(null);
-            throw e;
-        }
+        return clientWindow;
     }
 }
--- a/docs/14_0_0/core/clientwindow.md
+++ b/docs/14_0_0/core/clientwindow.md
@@ -6,6 +6,20 @@ This allows for separating state not only per session but also on a per client-w
 
 
 Prime Client Window is activated in faces-config.xml
+
+**MyFaces:**
+```xml
+<faces-config>
+    ...
+    <factory>
+        <client-window-factory>org.primefaces.clientwindow.PrimeClientWindowFactory</client-window-factory>
+    </factory>
+</faces-config>
+```
+
+**Mojarra:** see: https://github.com/eclipse-ee4j/mojarra/issues/5297
+
+```xml
 ```xml
 <faces-config>
     ...
@@ -15,8 +29,9 @@ Prime Client Window is activated in faces-config.xml
 </faces-config>
 ```
 
-It´s a replacement for the standard JSF-mechanism activated in web.xml. So only add PrimeClientWindowLifecycleFactory
-but not add javax.faces.CLIENT_WINDOW_MODE!
+It´s a replacement for the standard JSF-mechanism activated in web.xml. So only add `PrimeClientWindowLifecycleFactory` or `PrimeClientWindowFactory`
+but do **NOT** add `javax.faces.CLIENT_WINDOW_MODE`!
+
 ```xml
 <context-param>
     <param-name>javax.faces.CLIENT_WINDOW_MODE</param-name>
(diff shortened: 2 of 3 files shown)


melloware:
@tandraschko 2 things.

1. Mojarra requires the web.xml context param `javax.faces.CLIENT_WINDOW_MODE` or it won't activate.
2. I am getting a stack trace that gets ignored...

```
java.lang.IllegalStateException: Committed
    at org.eclipse.jetty.server.HttpChannel.resetBuffer (HttpChannel.java:1093)
    at org.eclipse.jetty.server.HttpOutput.resetBuffer (HttpOutput.java:1467)
    at org.eclipse.jetty.server.Response.resetBuffer (Response.java:1219)
    at org.eclipse.jetty.server.Response.sendRedirect (Response.java:571)
    at org.eclipse.jetty.server.Response.sendRedirect (Response.java:505)
    at org.eclipse.jetty.server.Response.sendRedirect (Response.java:580)
    at com.sun.faces.context.ExternalContextImpl.redirect (ExternalContextImpl.java:725)
    at javax.faces.context.ExternalContextWrapper.redirect (ExternalContextWrapper.java:597)
    at org.primefaces.clientwindow.PrimeClientWindow.decode (PrimeClientWindow.java:82)
    at com.sun.faces.lifecycle.LifecycleImpl.attachWindow (LifecycleImpl.java:131)
    at javax.faces.webapp.FacesServlet.executeLifecyle (FacesServlet.java:706)
    at javax.faces.webapp.FacesServlet.service (FacesServlet.java:451)
```

primefaces/src/main/java/org/primefaces/clientwindow/PrimeClientWindowFactory.java
```diff
@@ -25,31 +25,28 @@
...
+    public ClientWindow getClientWindow(FacesContext context) {
+        ClientWindow clientWindow = context.getExternalContext().getClientWindow();
         if (clientWindow == null) {
             clientWindow = new PrimeClientWindow();
         }
-
         try {
-            clientWindow.decode(facesContext);
-            facesContext.getExternalContext().setClientWindow(clientWindow);
+            clientWindow.decode(context);
```
melloware: I think this logic needs to be examined whether it is still needed?

primefaces/src/main/java/org/primefaces/clientwindow/PrimeClientWindow.java
```diff
@@ -84,8 +85,6 @@ public void decode(FacesContext context) {
                 catch (IOException e) {
                     throw new FacesException("Could not send initial redirect!", e);
                 }
-
-                context.responseComplete();
```
melloware: Had to remove this as its no longer part of LifeCycle and it kept reporting "response is closed".

melloware:
I don't think we are allowed to redirect anymore either because its not a lifecycle?

tandraschko:
Actually we must NOT change the ClientWindow itself, only the factory part
Before we attached it via a own lifecycle because 2.2
Now we can simple use the real factory actually

tandraschko:
It also worked fine on MyFaces 2.3

melloware:
OK updated my PR now all integration tests are passing and I submitted a Mojarra ticket about needing the `web.xml` param to enable our custom factory.

tandraschko:
should we mention that in the docs, that only Mojarra requires the context-param?

melloware:
Note added about Mojarra and link provided.

melloware:
@tandraschko I think I am seeing one strange this with this.  Now with MyFaces I see it get windowID 1234 then immediately refresh and get 4567 for example. Like on first page load its double loading.  Do you think it is because of the web.xml `javax.faces.CLIENT_WINDOW_MODE` it seems like Myfaces default handling is happening and then our custom on kicks in?

tandraschko:
can you try to remove javax.faces.CLIENT_WINDOW_MODE and see what happens?

melloware:
Yes I will test and report back.

melloware:
Nevermind, MyFaces is doing the right thing its Mojarra that does the double redirect but it all related to that issue I already reported. Sorry!

tandraschko:
but that means that we can't remove the old LifecycleFactory :/

I think we should leave it and provide the alternative ClientWIndowFactory

then we can mention in the docs: Use ClientWindowFactory but in mojarra older then version x, please use the old lifecyclefactory

melloware:
OK I have put the LifeCycle back in as well and updated the docs to mention use direct Factory for MyFaces and LifeCycleFactory for Mojarra.

primefaces-integration-tests-jakarta/src/main/webapp/WEB-INF/web.xml
```diff
@@ -18,6 +18,10 @@
         <param-name>jakarta.faces.STATE_SAVING_METHOD</param-name>
         <param-value>${jakarta.faces.STATE_SAVING_METHOD}</param-value>
     </context-param>
+    <context-param>
+        <param-name>jakarta.faces.CLIENT_WINDOW_MODE</param-name>
```
tandraschko: shouldnt we remove this?

primefaces-integration-tests/src/main/webapp/WEB-INF/web.xml
```diff
@@ -18,6 +18,10 @@
         <param-name>javax.faces.STATE_SAVING_METHOD</param-name>
         <param-value>${javax.faces.STATE_SAVING_METHOD}</param-value>
     </context-param>
+    <context-param>
+        <param-name>javax.faces.CLIENT_WINDOW_MODE</param-name>
```
tandraschko: shouldnt we remove this?

primefaces-showcase/src/main/webapp/WEB-INF/web.xml
```diff
@@ -37,6 +37,11 @@
         <param-name>javax.faces.INTERPRET_EMPTY_STRING_SUBMITTED_VALUES_AS_NULL</param-name>
         <param-value>true</param-value>
     </context-param>
+    <!-- enable client window mode -->
+    <context-param>
+        <param-name>javax.faces.CLIENT_WINDOW_MODE</param-name>
```
tandraschko: shouldnt we remove this?

melloware: We can remove it from showcase but it means you can only ever run the showcase with MyFaces which is OK with me?

primefaces-integration-tests/src/main/webapp/WEB-INF/web.xml
melloware: Until Mojarra fixes it the integration tests don't work without this unless I put the LifeCycleMapping back in the faces-config.xml of the Integration tests?

primefaces-showcase/src/main/webapp/WEB-INF/web.xml
tandraschko: AFAIR you said this param even breaks Mojarra?
So i would just use the LifecycleFactory in our showcase/IT's but suggest the new one for MF users

melloware: OK i will revert these to use LifeCycle and then it will work for both again.

tandraschko:
+1

melloware:
I messed up my branch so closing and re-opening.

PR #10458, opened 2023-08-06
Fix #10457: queueMicroTask instead of setTimeout(0)

### LINKED ISSUE #10457: Core: queueMicroTask instead of setTimeout(0)
See: https://www.freecodecamp.org/news/queuemicrotask/

Supported on all Major browsers since we dropped IE support: https://developer.mozilla.org/en-US/docs/Web/API/queueMicrotask

melloware:
Fix #10457: queueMicroTask instead of setTimeout(0)

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/ajaxstatus/ajaxstatus.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/ajaxstatus/ajaxstatus.js
@@ -91,14 +91,9 @@ PrimeFaces.widget.AjaxStatus = PrimeFaces.widget.BaseWidget.extend({
         $this = this;
 
         doc.on('pfAjaxStart', function() {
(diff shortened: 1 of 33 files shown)


tandraschko:
what are the exact reason anyway to have timeout=0? maybe we can fix those things completely different in newer browsers

melloware:
If I had to guess most of the Timeout(0) are due to AJAX calls and needing to perform a task after the AJAX stuff is done but right before returning the browser event loop.  It might be possible to remove some of these but that would require much more extensive testing.

jepsar:
I've seen timeouts of 1, 2, 5 or 10 millis being replaces with queueTask. Those small delays are not relevant?

melloware:
I have been testing and so far they don't seem relevant. I think anything under 10ms is probably not noticeable and so far I have not see any difference. The ones specifically set higher I assume were done on purpose so I left those.

PR #10465, opened 2023-08-08
JSF 2.2 PartialResponse remove wrapper

melloware:
cc @tandraschko


primefaces/src/main/java/org/primefaces/context/PartialResponseWriterWrapper.java
```diff
@@ -23,96 +23,13 @@
...
-import java.io.IOException;
-import java.util.Map;
 import javax.faces.context.PartialResponseWriter;
 
 public class PartialResponseWriterWrapper extends PartialResponseWriter {
 
-    private PartialResponseWriter wrapped;
 
     public PartialResponseWriterWrapper(PartialResponseWriter wrapped) {
         super(wrapped);
```
melloware: Not even sure we need this class at all anymore?

tandraschko: we surely need this
the JSF specs only contains a ResponseWriterWrapper, not a PartialResponseWriterWrapper

melloware: Sorry I mean this is wrapping just the wrapper and then being extended when we can just extend PartialResponseWriter directly now?

melloware:
Maybe I don't understand but this wrapper was just wrapping and then calling each method overridden of PartialResponseWriter?  So our Impl wasn't doing anything was it?

tandraschko:
hmm i see
then you need to debug, to find out what is different in Mojarra and MF wrapping the writer

melloware:
@christophs78 may need some help with this one.  NONE of these test fail locally for me.  Very strange.

christophs78:
After checking out your feature-branch and some clean + install (PrimeFaces "Core") respective clean + package (PrimeFaces IT) later it also breaks on my machine with MyFaces 2.3 AND CSP enabled.
image

melloware:
Ahh CSP I think I was not enabled. Let me try that !

melloware:
OK was able to debug it and put my findings in the MyFaces ticket: https://issues.apache.org/jira/projects/MYFACES/issues/MYFACES-4536

tandraschko:
@melloware should we close this?
we have a reproducer in the MF issue and can't remove this for this, maybe in the next 2-3 major PF versions

melloware:
Yep we can close

PR #10479, opened 2023-08-11
Fix #10476: Accordion scroll tab into view

### LINKED ISSUE #10476: AccordionPanel: focus on clicked tab should be the default
### Description

The issue I have is also described in https://forum.primefaces.org/viewtopic.php?t=40788.

Actually, PF does not care of focusing a tab which was clicked by the user. That's not good because it's really not user friendly.

It should be the default to focus the tab when the user clicks a tab. PF should take care of displaying that element into the view (by auto-focusing). Otherwise **it is not user friendly and against accessibility**.

### Additional context
It's not really a new feature. In my opinion this issue is more of a **bug** because if a user clicks on a tab and the tab doesn't appear in the viewport because it wasn't focused when clicked, the user isn't able to do anything with the content of the clicked tab and needs to focus manually. So this is 2 clicks instead of one (more precisely: one click and one scroll). When a user clicks/activates a tab, he also expects that the activated tab is within his focus.

melloware:
Fix #10476: Accordion scroll tab into view

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/accordion/accordion.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/accordion/accordion.js
@@ -26,6 +26,7 @@
  * @prop {boolean} cfg.rtl `true` if the current text direction `rtl` (right-to-left); or `false` otherwise.
  * @prop {boolean} cfg.multiViewState Whether to keep AccordionPanel state across views.
  * @prop {number} cfg.toggleSpeed Speed of toggling in milliseconds.
+ * @prop {number} cfg.scrollIntoView Should the tab scroll into view. One of start, center, end, nearest, or NULL if disabled.
  */
 PrimeFaces.widget.AccordionPanel = PrimeFaces.widget.BaseWidget.extend({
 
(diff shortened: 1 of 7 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/accordion/accordion.js
```diff
@@ -100,16 +100,21 @@ PrimeFaces.widget.AccordionPanel = PrimeFaces.widget.BaseWidget.extend({
...
 
-                if(element.hasClass('ui-state-active')) {
+                if(header.hasClass('ui-state-active')) {
                     $this.unselect(tabIndex);
                 }
                 else {
                     $this.select(tabIndex);
-                    $(this).trigger('focus.accordion');
+                    header.trigger('focus.accordion');
+                    this.scrollIntoView({
```
jepsar: Makes this the header to scroll into view, or the tab contents?

melloware: even better because I have set "center" "center" it scrolls the value into the center of the viewport not the top or bottom of viewport.  To me that is expected behavior and it works great.

melloware:
Left you a comment in the ticket to try `nearest`

melloware:
Actually try

```js
                   header.next()[0].scrollIntoView({
                        behavior: "smooth",
                        block: "center" ,
                        inline: "center"
                    });
```

melloware:
Its now a config `scrollIntoView`

melloware:
Try this version: 
[pf-10476.zip](https://github.com/primefaces/primefaces/files/12324168/pf-10476.zip)

melloware:
Thanks for testing!  I am happy with the end result.

PR #10485, opened 2023-08-13
Fix #10484: BlockUI do not show block if AJAX queue empty

### LINKED ISSUE #10484: BlockUI: with `delay` will be activated even after ajax request was finished
### Describe the bug

Using a 

`<p:blockUI delay="2000" ..`

works in most cases. However, in relation with a `p:dataTable `and `<p:ajax event="rowSelect"` is does not work as expected, because the `blockUI` is activated even when a row was already selected (i.e. the ajax request completed successfully).

### Reproducer

[primefaces-test-jakarta.zip](https://github.com/primefaces/primefaces/files/12329257/primefaces-test-jakarta.zip)

### Expected behavior

The `blockUI` should **not** be activated if the ajax request was done before the `delay` of the `blockUI`.

### PrimeFaces edition

None

### PrimeFaces version

13.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4

### Java version

17

### Browser(s)

Firefox 115.0.2

melloware:
Fix #10484: BlockUI do not show block if AJAX queue empty

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
@@ -158,6 +158,11 @@ PrimeFaces.widget.BlockUI = PrimeFaces.widget.BaseWidget.extend({
 
         var delay = this.cfg.delay || 0;
         this.timeout = PrimeFaces.queueTask(function() {
+            // #10484: if delayed and AJAX event already finished
+            if ($this.cfg.triggers && PrimeFaces.ajax.Queue.isEmpty()) {
+                PrimeFaces.warn("BlockUI AJAX event completed before showing the block.");
+                return;
+            }
             $this.alignOverlay();
 
             var animated = $this.cfg.animate;


primefaces/src/main/resources/META-INF/resources/primefaces/blockui/blockui.js
```diff
@@ -158,6 +158,11 @@ PrimeFaces.widget.BlockUI = PrimeFaces.widget.BaseWidget.extend({
 
         var delay = this.cfg.delay || 0;
         this.timeout = PrimeFaces.queueTask(function() {
+            // #10484: if delayed and AJAX event already finished
+            if ($this.cfg.triggers && PrimeFaces.ajax.Queue.isEmpty()) {
+                PrimeFaces.warn("BlockUI AJAX event completed before showing the block.");
```
nimo23, line 163: Please remove the warning log. 

The user does not need such information as there is nothing they can do to avoid the warning and it only creates confusion. And by the way, it's perfectly fine if "BlockUI AJAX event completed before showing the block" - so it should be treated as a correct behavior and not be flagged as a warning.

melloware: Warning is ONLY displayed in Development mode not in Production Mode. If you look in your F12 console log you will see lot of debug information but only in Development mode.  It might help explain why they are not seeing their BlockUI if they don't understand why it s not being displayed.

nimo23: Oh ok, that's perfect. Did not know that it distincts between dev and prod mode.

PR #10488, opened 2023-08-13
Update autocomplete.js: avoid duplicate method calls

nimo23:
No need to call `PrimeFaces.escapeHTML` multiple times for the same variable.


melloware (approved):
LGTM

PR #10513, opened 2023-08-19
Fix #4685: DatePicker ARIA/WCAG and keyboard support

### LINKED ISSUE #4685: DatePicker: Accessibility 508 Compliance
## 1) Environment
- PrimeFaces version:6.1.19
- Application server + version:  Jboss 6.0
- Affected browsers: All browsers

## 2) Expected behavior

We are using https://www.primefaces.org/showcase/ui/input/datePicker.xhtml.
We are expecting it to be 508 / key board / screen reader compliant like
https://dequeuniversity.com/library/aria/date-pickers/sf-date-picker
...

## 3) Actual behavior

Date Picker is not accessibly via key board for date selections
..

## 4) Steps to reproduce

use any of https://www.primefaces.org/showcase/ui/input/datePicker.xhtml.
Date Picker is not accessibly via key board for date selections 
We are expecting it to be 508 / key board / screen reader compliant like
https://dequeuniversity.com/library/aria/date-pickers/sf-date-picker
## 5) Sample XHTML
<h:outputText  id="reportingPeriodStart"
	value="#{homeBean.startDate}" >
	<f:convertDateTime pattern="MM/dd/yyyy" />
</h:outputText>	
 

## 6) Sample bean

@ManagedBean(name = "homeBean")
@ViewScoped
public class HomeController {

private Date startDate;

public Date getStartDate() {
return startDate;
}

public void setStartDate(Date startDate) {
this.startDate = startDate;
}
}
..

melloware:
Fix #4685: DatePicker ARIA/WCAG and keyboard support

- ARIA labels for all
- Focus trap while the overlay panel is displayed
- All keyboard and ARIA specs listed here: https://primevue.org/calendar/#accessibility

@christophs78 I know you might be on holiday but when you return it would be great if you could have your co-worker validate this!

# Notes:
- Because of focus trap when overlay is shown the only way to do keyboard entry is when the overlay is closed  which makes sense from ARIA point of view.  If the overlay is up its the ACTIVE component.
- `showOnFocus="false"` if you want the datepicker editable immedately upon focus so the overlay does not pop up and grab focus
- `ARROW_DOWN, ENTER` when editing the input will trigger the overlay to open
- `ESC` key closes the overlay and focuses the input again for editing

So the above is different than the "current" behavior which lets you edit the field while the overlay is open.  Because the overlay is a focus trap now you are either in the overlay or you are editing the input but not both.  This is technically a `BREAKING CHANGE` from current behavior but is the correct behavior.

# TODO

- [x] TimeSelect demo on Java8 is throwing an exception in Showcase
- [x] Revisit ` //type="number" min="' + minHour + '" max="' + maxHour + '" - does not work well on Firefox 70, so we don´t use it` as Firefox 70 is long gone
- [x] Fix integration tests
- [x] @christophs78 sign off on testing manually


primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -1857,16 +1867,205 @@
             this.panel.off('click.datePicker-clearButton', clearButtonSelector).on('click.datePicker-clearButton', clearButtonSelector, null, this.onClearButtonClick.bind($this));
 
             var dateSelector = '.ui-datepicker-calendar td a';
-            this.panel.off('click.datePicker-date', dateSelector).on('click.datePicker-date', dateSelector, null, function (event) {
+            this.panel.off('click.datePicker-date keydown..datePicker-date', dateSelector).on('click.datePicker-date', dateSelector, null, function (event) {
```
jepsar:
```suggestion
            this.panel.off('click.datePicker-date keydown.datePicker-date', dateSelector).on('click.datePicker-date', dateSelector, null, function (event) {
```

primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -1857,16 +1867,205 @@
...
+                }
+                return;
+            }
+
+            var $this = this;
+            var currentElement = $(event.currentTarget);
+            var $tabbableElements = $this.panel.find('a:tabbable');
+            var currentIndex = $tabbableElements.index(currentElement);
+            switch (event.key) {
+                case 'Enter':
```
jepsar, line 1914: Don't we have a constant with key names?

melloware: I think you might be the thinking of the old Jquery constants for key codes but since these keys are the actual key names it makes the code readable without constants. For example it was keycode === 27 which would be $ui.keycode.ESCAPE.   

but now none of that is necessary as event.key === 'Escape' reads much better and is supported in all browsers.

melloware:
Also I know why the integration tests are failing but don't want to fix anything yet as I need to review what PrimeVue is doing as it's related to keyboard entry of dates.

primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
jepsar: It's just that, in Java, I never repeat the same string over and over. And I don't like things to be stringly typed. 😁

christophs78:
Thank´s for working on this!
I´ll be back at office at 28th of August and then check back with one of my co-workers.

melloware:
@christophs78 awesome!!!  Thank you so much.  I have tested most everything and it works great and it actually works a little better than PrimeVue as while testing PrimeVue I noticed some keyboard bugs 😄

christophs78:
1)
I have mixed feelings related to the implications of the focus-trap when the overlay is opened.
Especially because PF DatePicker opens the overlay per default when the input is focused. 
Some people may dislike this behaviour.
https://material.angular.io/components/datepicker/overview as reference also seems to have a focus-trap with similar implications. But it does not open the overlay OOTB.

2)
My co-worker (JAWS-user) tested this changes.
He tested following two usecases:
image
image
He tells me he has issues with navigating to the month-/year-selection. Repective to the time-input. Maybe i simply should not have enabled this features.

melloware:
1. Agree on the focus trap but that is what PrimeVue/React/Ng are doing and from an ARIA perspective most agree that if a popup is open it should have focus.  So you need to use `showOnFocus="false"` if you don't want it to popup on TAB.
Realtstically you then press ENTER/SPACE to bring up the popup.

2. I would like to get to the bottom of those issues I was able to TAB to all elements in the popup?  Arrow keys keep you on dates but TAB should cycle through all focusable elements?

christophs78:
Talked to my co-worker one more time.

> 2. I would like to get to the bottom of those issues I was able to TAB to all elements in the popup?  Arrow keys keep you on dates but TAB should cycle through all focusable elements?

He simply tried not hard enough. Tab works.
Afterwards he told me he expects the date (in the input) to change after selects a value from the month- or year-navigator. Told him this is not the intended way because you needs to select a day afterwards.
Long story, short summary: He personally simply prefers to type the date into the input. (He usually does not use datepicker-popus.) He is happy with a component which allows to enter the date into the input straight-forward.

jepsar:
> ... Realistically you then press `ENTER/SPACE` to bring up the popup and `ESC` to get out of the popup. ...

I can imagine that there are people using date formats containing a space.

melloware:
Sorry yes I had to remove SPACE as a trigger because of date range. It's only ENTER and ARROWDOWN that show the popup

tarwork:
thanks team for addressing this accessibility issue in the datepicker component. 

Ii is something our org is trying to address currently for an internal audit. Can this be included in a 13.x release ? Waiting till PF 14 will also require us significant work to upgrade to JSF 2.3 and we are not ready for that yet.

melloware:
@tarwork totally understand. I will discuss with PrimeTek about backporting to 13.  My only concern is its a change in behavior not just a bug fix.

PR #10536, opened 2023-08-22
Fix #10532: Java 11

### LINKED ISSUE #10532: Switch to Java11 as minimal requirement
The EOL of Java8 is already reached but most of the JDK's provide still further support.

The sad thing is that Webdriver-Manager and Selenium forces us to use Java11:
https://www.selenium.dev/blog/2023/java-8-support/

This means we have to upgrade to Java11.

melloware:
Fix #10532: Java 11 for builds

Still keeps Java8 as the compiled JAR language only updates the builds for Java8


tandraschko:
TBH i think we should go full java8
when still compiling on language level 8, we cant make sure anymore that it works on java8

melloware:
OK I bumped to 11 and marked it `breaking-change`

.github/workflows/build.yml
```diff
@@ -53,10 +53,10 @@ jobs:
       - name: Build
         run: mvn clean install -pl -primefaces-integration-tests-jakarta -T1C -DskipTests -Dcheckstyle.skip -Djsdoc.skip.typedoc=true -Dmaven.javadoc.skip=true --batch-mode --show-version
       - name: Integration Tests (serial)
```
christophs78: We do not need IT´s in serial order any more when we target Java 11+. (This was only necessary for Java 8 because of specific Java 8 internals.)

melloware: Good find I have removed the serial testing only parrallel now.

melloware:
The build is now significantly faster about 18 minutes down from 28 minutes with this change.

artragis:
Does it mean that you plan primefaces to be minimum java 11 soon?

tandraschko:
Yes

PR #10542, opened 2023-08-23
Fix #1435/#3992: SelectManyMenu/SelectOneListbox WCAG and Keyboard support

### LINKED ISSUE #1435: SelectManyMenu: add 508 keyboard support
I'm missing keyboard support for tag p:selectManyMenu like it is present for h:selectManyListbox (only Firefox does it exactly in the described way with CTRL/SHIFT keys):
- when focus is reached via TAB key one of the rows (the first row or first selected row) should be highlighted with a dotted frame
- pressing UP or DOWN key should move the highlighted row up/down. If neither CTRL nor SHIFT is pressed the new highlighted row should be marked.
- If CTRL is pressed together with UP/DOWN the old selection should stay and user can add/remove the currently highlighted row to the selection via SPACE key.
- If SHIFT is pressed together with UP/DOWN the old selection is changed to be the range from the first selected row to the reached row (like when clicking with SHIFT+mouse on a row)
- POS1/END should jump to the beginning/end of the items list with or without changing the selected item(s) depending on CTRL/SHIFT keys.
- PAGE UP/PAGE DOWN should scroll to previous/next page if not all items are visible
- If a p:outputLabel has a defined accesskey and is assigned to the p:selectManyMenu via its 'for' attribute the selectManyMenu should get focus (like when using TAB key to reach it) if the accesskey is pressed

I'm using PrimeFaces 6.0-RC3

### LINKED ISSUE #3992: SelectManyMenu: Filtered Item Selection
In my app I'd like to change the default selection strategy when filter is updated.

I primarily miss the **onFilterUpdate** event in client side API.

Then I would be grateful for two new methods:
- clearSelection() - clears the selection. If selectManyMenu is combined with the filter and some items are selected, these are still selected after updating the filter value, even they disappear from the list. It is confusing and even reported as a bug in my app.

- selectAllFiltered() - clears the previous selection and then select all filtered items. Handy if you need select everything in the filtered list (selectAll() selects also invisible items).

melloware:
Fix #1435: SelectManyMenu WCAG and Keyboard support
Fix #3992: Only selectAll/unselectAll visible (filtered) items

Followed PrimeVue as closely as I could. 

- [x] Implement PrimeVue like keyboard and ARIA attributes
- [x] @christophs78 verify with co-worker

### DIFF
--- a/docs/14_0_0/components/selectmanymenu.md
+++ b/docs/14_0_0/components/selectmanymenu.md
@@ -20,37 +20,37 @@ SelectManyMenu is an extended version of the standard SelectManyMenu.
 | Name | Default | Type | Description | 
 | --- | --- | --- | --- |
 id | null | String | Unique identifier of the component
-rendered | true | Boolean | Boolean value to specify the rendering of the component, when set to false component will not be rendered.
 binding | null | Object | An el expression that maps to a server side UIComponent instance in a backing bean
-value | null | Object | Value of the component referring to a List.
+caseSensitive | false | Boolean | Defines if filtering would be case sensitive.
+collectionType | null | String | Optional attribute that is a literal string that is the fully qualified class name of a concrete class that implements `java.util.Collection` or an EL expression that evaluates to either 1. such a String, or 2. the `Class` object itself.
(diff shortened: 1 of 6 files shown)


melloware:
Agreed at least it matches PrimeVue.  Thanks for the update going to merge!

PR #10546, opened 2023-08-24
Fix #10545: Custom sortFunction add sortMeta

### LINKED ISSUE #10545: DataTable: Optionally get the SortMeta of sortFunction
### Description

I need in sortFunction the info if asc or desc ordering.

```
<p:column headerText="Name" 
          sortable="true" 
          sortBy="#{row.name}" 
          sortFunction="#{indexController.onSortValue}">
    <h:outputLabel value="#{row.name}"/>
</p:column>
```

in most cases is this enough
```
    public int onSortValue(Object o1, Object o2) {
        // impl
    }
```

but i have a case, where i need the info if asc or desc
```
    public int onSortValue(Object o1, Object o2, SortMeta sortMeta) {
        // impl
    }
```
example: i want empty (null) values always on top (or bottom), no matter if sorted by asc or desc

### Additional context

_No response_

melloware:
Fix #10545: Custom sortFunction add sortMeta

This is what other places in the code do but not sure its the correct thing to do.

### DIFF
--- a/docs/14_0_0/components/datatable.md
+++ b/docs/14_0_0/components/datatable.md
@@ -244,7 +244,7 @@ Instead of using the default sorting algorithm which uses a java comparator, you
 own sort method as well.
 
 ```java
-public int sortByModel(Object car1, Object car2) {
+public int sortByModel(Object car1, Object car2, SortMeta sortMeta) {
     //return -1, 0 , 1 if car1 is less than, equal to or greater than car2
 }
 ```
--- a/docs/migrationguide/14_0_0.md
+++ b/docs/migrationguide/14_0_0.md
@@ -11,6 +11,9 @@
 
 ## DatePicker/Calendar
   * Locale `firstDay` renamed `firstDayOfWeek` to match other Prime libraries
+  
+## DataTable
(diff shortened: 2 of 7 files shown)


tandraschko:
did you test it?
the method-signature is also preconfigured in ColumnHandler

we should provide 2 small integration tests then, for both ways

melloware:
Integration test fails as it doesn't look like you have can have 2 columnHandlers MetaRules for the same property.

```java
    private static final MetaRule SORT_FUNCTION = new MethodRule(Column.PropertyKeys.sortFunction.name(),
            Integer.class,
            new Class<?>[]{Object.class, Object.class});

    private static final MetaRule SORT_ADVANCED_FUNCTION = new MethodRule(Column.PropertyKeys.sortFunction.name(),
            Integer.class,
            new Class<?>[]{Object.class, Object.class, SortMeta.class});
```

The last one added for the `sortFunction` column wins

melloware:
OK changes made, marked as `breaking-change` for 14.0.0.  I think this is a good change going forward as people may want to know NullSorter order or CaseInsenstive etc on the SortMeta.

PR #10550, opened 2023-08-24
Fix #10549 - UITable: cache columns when possible

### LINKED ISSUE #10334: DataTable: dynamic columns have wrong order after sorting

### LINKED ISSUE #10549: UITable: rework caching columns
### Description

After #10334 I recently realized that the cached columns was useless when used with dynamic columns, as it is invalidated all over the place whenever we iterate over rows. This is due to #6659, as there is no details in there, I tried to simplify few things based on this sample #2154

simplification

### Additional context

_No response_

Rapster:
Fix #10549 - UITable: Rework caching columns

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datatable/DataTable033.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datatable/DataTable033.java
@@ -23,12 +23,6 @@
  */
 package org.primefaces.integrationtests.datatable;
 
-import lombok.Data;
-
-import javax.annotation.PostConstruct;
-import javax.faces.context.FacesContext;
-import javax.faces.view.ViewScoped;
-import javax.inject.Named;
 import java.io.Serializable;
 import java.lang.reflect.Field;
 import java.time.LocalDate;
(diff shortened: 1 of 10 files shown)


tandraschko:
it seems we have some tests using p:columns
could you please check if we should add some more? i dont know TBH why we cached it anyway, maybe a performance thing

Otherwise the change looks good

Rapster:
Is draggable columns supported by PF selenium?

melloware:
I don't think we ever figured that out because moving the mouse and dragging is a PITA with Selenium if I remember.

melloware (approved):

Rapster:
Still wanna try one thing

Rapster:
Ok I gave up trying to reproduce the scenario... Like you said it's a PITA...

Rapster:
I finally came up with an IT where I test hybrid columns (mulitple p:columns + p:column)

melloware:
Fantastic!

PR #10556, opened 2023-08-25
Fix #10554 - DataTable: reduce number of calls to setRowIndex in encodeRows

### LINKED ISSUE #10554: DataTable: refactor isInSameGroup
### Description

Call to `setRowIndex` comes with a price as it save the whole state of `DataTable `and its children. In `isInSameGroup`, (which is called by `encodeRow`) `setRowIndex` is called way too many times (for no reason?)

limit number of calls of setRowIndex in isInSameGroup

### Additional context

_No response_

Rapster:


tandraschko:
We need some tests

Rapster:
DataTable037 does already the right job

PR #10560, opened 2023-08-26
Fix #10522: Consolidate file path traversal checks

### LINKED ISSUE #10522: FileUpload: sonar vulnerabilities
https://sonarcloud.io/project/issues?resolved=false&types=VULNERABILITY&id=org.primefaces%3Aprimefaces&open=AYnBpWp7DdSniu8YOW-R

@Rapster Can you check it it pls?

melloware:
Fix #10522: Consolidate file path traversal checks


melloware:
I will also retest the original image cropper report while I test this.

Rapster:
Ty very much @melloware!  Few things:
- Can we remove ValidationException, is there any value for it? JSF has a ValidatorException already
- Do we really need SafeFile? All it does, is extending File and perform few more checks but we could simply move those checks in `FileUploadUtils#requireValidFilePath`.

melloware:
Good idea. I will take a look at both of those items as well.

Rapster:
Also, regarding FileUplaodUtils#requireValidFilename, I think we can do better... I did something in the past:

```java
    // https://owasp.org/www-community/OWASP_Validation_Regex_Repository
    private static final Pattern INVALID_FILENAME_WINDOWS =
            Pattern.compile("^(?!^(PRN|AUX|CLOCK\\$|NUL|CON|COM\\d|LPT\\d|\\..*)(\\..+)?$)[^\\x00-\\x1f\\\\?*:\\\";|/<>]+$");

    private static final Pattern INVALID_FILENAME_LINUX = Pattern.compile(".*[/\0:*?\\\"<>|].*");

    private FileUploadUtils() {
        // NOOP
    }

    public static String requireValidFilename(String filename) {
        if (LangUtils.isBlank(filename)) {
            throw new FacesException("Filename cannot be empty or null");
        }

        if (isSystemWindows()) {
            if (!INVALID_FILENAME_WINDOWS.matcher(filename).find()) {
                throw new FacesException("Invalid filename: " + filename);
            }
        }
        else if (INVALID_FILENAME_LINUX.matcher(filename).find()) {
            throw new FacesException("Invalid filename: " + filename);
        }

        return filename;
    }

```

And some test:

```java
@Test
    public void getValidFilename_Linux() {

        try(MockedStatic<FileUploadUtils> fileUploadUtilsMockedStatic = Mockito.mockStatic(FileUploadUtils.class, Mockito.CALLS_REAL_METHODS)) {
            fileUploadUtilsMockedStatic.when(FileUploadUtils::isSystemWindows).thenReturn(false);

            // Check each invalid character individually
            String[] invalidCharacters = { "/", "\0", ":", "*", "?", "\"", "<", ">", "|" };
            for (String character : invalidCharacters) {
                String invalidFileName = "filename_with_" + character;
                assertThrows(FacesException.class, () -> FileUploadUtils.requireValidFilename(invalidFileName));
            }
        }
    }

    @Test
    public void getValidFilename_Windows() {
        try(MockedStatic<FileUploadUtils> fileUploadUtilsMockedStatic = Mockito.mockStatic(FileUploadUtils.class, Mockito.CALLS_REAL_METHODS)) {
            fileUploadUtilsMockedStatic.when(FileUploadUtils::isSystemWindows).thenReturn(true);

            // Check each invalid character individually
            String[] invalidCharacters = { "/", "\\\\", ":", "*", "?", "\"", "<", ">", "|" };
            for (String character : invalidCharacters) {
                String invalidFileName = "filename_with_" + character;
                assertThrows(FacesException.class, () -> FileUploadUtils.requireValidFilename(invalidFileName));
            }

            // Add reserved device names as invalid filenames
            String[] reservedDeviceNames = { "CON", "PRN", "AUX", "NUL", "COM1", "COM2", "COM3", "COM4", "COM5", "COM6",
                    "COM7", "COM8", "COM9", "LPT1", "LPT2", "LPT3", "LPT4", "LPT5", "LPT6", "LPT7", "LPT8", "LPT9" };
            for (String deviceName : reservedDeviceNames) {
                assertThrows(FacesException.class, () -> FileUploadUtils.requireValidFilename(deviceName));
            }
        }
    }
```

melloware:
Attached is the reproducer for the ImageCropper Path Traversal from #3286 and I confirmed its still being properly blocked.
[pf-3286.zip](https://github.com/primefaces/primefaces/files/12446544/pf-3286.zip)

melloware:
@Rapster I have implemented all your suggested changes exception the `SafeFile` one.  That is next on my list.

primefaces/src/test/java/org/primefaces/util/FileUploadUtilsTest.java
```diff
@@ -161,48 +164,36 @@ public void checkPathTraversal_AbsoluteFile() {
         // Act
         if (os.contains("win")) {
```
Rapster, line 166: I don't think this condition is ever filled, maybe used MockedStatic as in getValidFilename_Linux ?

primefaces/src/test/java/org/primefaces/util/FileUploadUtilsTest.java
```diff
@@ -263,4 +254,39 @@ public void getValidFilename_GitHub8359() {
         assertEquals(filename, result);
     }
 
+    @Test
+    public void getValidFilename_Linux() {
```
Rapster: rename into requireValidFilename_linux

primefaces/src/test/java/org/primefaces/util/FileUploadUtilsTest.java
```diff
@@ -263,4 +254,39 @@ public void getValidFilename_GitHub8359() {
...
+            String[] invalidCharacters = { "/", "\0", ":", "*", "?", "\"", "<", ">", "|" };
+            for (String character : invalidCharacters) {
+                String invalidFileName = "filename_with_" + character;
+                assertThrows(FacesException.class, () -> FileUploadUtils.requireValidFilename(invalidFileName));
+            }
+        }
+    }
+
+    @Test
+    public void getValidFilename_Windows() {
```
Rapster: same here

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
```diff
@@ -71,31 +73,18 @@ public static String requireValidFilename(String filename) {
...
+        else if (INVALID_FILENAME_LINUX.matcher(filename).find()) {
+            throw new FacesException("Invalid filename: " + filename);
         }
 
-        return name;
+        return FilenameUtils.getName(filename);
     }
 
-    public static String requireValidFilePath(String filePath) throws ValidationException {
+    public static String requireValidFilePath(String filePath, boolean mustExist) {
```
Rapster, line 97: Should we used ValidatorException instead, sounds appropriate don't you think?

Rapster (changes requested):
few notes

melloware:
Good feedback I will make those changes. I thought about ValidatorException but doesn't it require a field or client id?  I will take a closer look!

primefaces/src/test/java/org/primefaces/util/FileUploadUtilsTest.java
melloware: This is still needed or the unit test fails on Windows.   Because the path is `/C:/...`

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
melloware: ValidatorException requires `FacesMessages` not strings so I think FacesException is OK here?

Rapster: You can always do something like that:

```java
throw new ValidatorException(new FacesMessage(FacesMessage.SEVERITY_ERROR, "Validation Error", value + " is not a valid email;"));
```

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
```diff
@@ -132,6 +129,16 @@ public static boolean isSystemWindows() {
         return File.separatorChar == '\\';
     }
 
+    public static int containsUnprintableCharacters(String s) {
```
Rapster, line 146: This replaces what used to do SafeFile?

melloware: Yes all your other checks look like they already cover the other checks from what SafeFile was doing.

melloware: There is one check which they don't explain why they are doing it its the very specific about percents which I don't understand.

```java
 private static final Pattern PERCENTS_PAT = Pattern.compile("(%)([0-9a-fA-F])([0-9a-fA-F])");
```

should I add this check?

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
melloware: Switched to ValidatorException

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
melloware: OK I added the PERCENT check as well.

Rapster (approved):

Rapster:
How come you force every push?

melloware:
Oh I like it better than Squash commits when merging :)

melloware:
I still squash at the end but I like it all to look like 1 commit.

PR #10590, opened 2023-09-03
Fix #10575: MenuItem warning in dev mode

### LINKED ISSUE #10575: Menu : MenuItem oncomplete script is not called if no action is specified
### Describe the bug

Hi !
Since the Primefaces 13 release, if you define a menu with a menuitem with an `oncomplete` script and you don't define a bean `action` method, the `oncomplete` script is not called.

### Reproducer

```
<p:menu>  
    <p:menuitem value="Without action" oncomplete="console.log('Oncomplete not called')" />  
    <p:menuitem value="With action" action="#{testView.init()}" oncomplete="console.log('Oncomplete called')" />  
</p:menu>
```

### Expected behavior

The oncomplete script should be called independently of the action method.

### PrimeFaces edition

Elite

### PrimeFaces version

13.0.0 and 14.0.0-SNAPSHOT

### Theme

_No response_

### JSF implementation

All

### JSF version

4.0

### Java version

11

### Browser(s)

All

melloware:
Fix #10575: MenuItem warning in dev mode

Warning in DEV mode only if someone is trying to use the MenuItem as a Command without an `action` or `listener` defined.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/renderkit/MenuItemAwareRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/renderkit/MenuItemAwareRenderer.java
@@ -28,6 +28,8 @@ import java.util.*;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
+
+import javax.faces.application.ProjectStage;
 import javax.faces.component.UIComponent;
 import javax.faces.component.UIForm;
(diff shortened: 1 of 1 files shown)


tandraschko:
still not sure about changes here, the cleanest way is the way i mentioned in the ticket
need think about it

melloware:
No problem I just put this here for thought.

PR #10591, opened 2023-09-04
Fix #10584 - DataTable: custom filter does not work when set first

### LINKED ISSUE #10584: DataTable: custom filter does not work when set first
### Describe the bug

Reproducible in showcase (starting from PF13):
https://www.primefaces.org/showcase/ui/data/datatable/columns.xhtml

Rearrange the columns so that the column with the custom filter is first in the list:
"status name country date activity"

image

The filter for the STATUS column is not working

In short, any custom filter doesn't work if it's in the first column.

UITable#updateFilterByValuesWithFilterRequest:
```
            if (hasCustomFilter) {
                filterValue = ((ValueHolder) filterFacet).getLocalValue(); // returns NULL!!!
            }
```

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

Community

### PrimeFaces version

13.0.0

### Theme

_No response_

### JSF implementation

All

### JSF version

all

### Java version

all

### Browser(s)

all

Rapster:


melloware (approved):

tandraschko:
can we add a test based on the example?

melloware:
We should be able to modify an existing test and just put the selectOne in the first column?

tandraschko:
+1 @melloware

Rapster:
I think it works since #10328 😅

melloware:
i will give it a shot!

melloware:
Confirmed its all working right now with the UIData backport fixes!

PR #10601, opened 2023-09-05
Fix #10592: HideNoSelectionOption handle string boolean

### LINKED ISSUE #10592: selectOneMenu: hide no select option not working on first instance
### Describe the bug

https://www.primefaces.org/showcase/ui/input/oneMenu.xhtml?jfwid=95060

the Hide NoSelection Option example 
the drop down in meant to not show the select one option and if you link one of the options and then another eventally the select one option does start to hide but it doesnt hide at first

I've copied the code to my project and i'm seeing the same issue, so if there is a fix then please let me know

### Reproducer

you can see it on the showcase for v13.0.0

### Expected behavior

the select one option to not appear in the drop down at all 

### PrimeFaces edition

Community

### PrimeFaces version

13.0.0

### Theme

_No response_

### JSF implementation

All

### JSF version

2.1

### Java version

19

### Browser(s)

_No response_

melloware:
Fix #10592: HideNoSelectionOption handle string boolean

If you used `hideNoSelectionOption="true"` it would not work only `hideNoSelectionOption="#{true}"` which is confusing for most users including myself who was stumped why it didn't work.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/renderkit/SelectRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/renderkit/SelectRenderer.java
@@ -44,7 +44,7 @@ public abstract class SelectRenderer extends InputRenderer {
     protected boolean isHideNoSelection(UIComponent component) {
         Object attribute = component.getAttributes().get("hideNoSelectionOption");
         if (attribute instanceof String) {
-            attribute = Boolean.parseBoolean(Objects.toString(attribute));
+            attribute = Boolean.parseBoolean((String) attribute);
         }
         return Boolean.TRUE.equals(attribute);
     }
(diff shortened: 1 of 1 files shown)


primefaces/src/main/java/org/primefaces/renderkit/SelectRenderer.java
```diff
@@ -43,6 +43,9 @@ public abstract class SelectRenderer extends InputRenderer {
 
     protected boolean isHideNoSelection(UIComponent component) {
         Object attribute = component.getAttributes().get("hideNoSelectionOption");
+        if (attribute instanceof String) {
+            attribute = Boolean.parseBoolean(Objects.toString(attribute));
```
jepsar: Just curious, why didn't you do it like this?
```suggestion
            attribute = Boolean.parseBoolean((String) attribute);
```

PR #10604, opened 2023-09-06
Update since JSF 2.3

Rapster:


melloware:
Looks like Sonar is still failing?  

> Failed to execute goal org.sonarsource.scanner.maven:sonar-maven-plugin:3.9.1.2184:sonar (default-cli) on project primefaces: Project not found. Please check the 'sonar.projectKey' and 'sonar.organization' properties, the 'SONAR_TOKEN' environment variable, or contact the project administrator

Rapster:
Yeah, I'm trying to make the sonar build available for this PR. SONAR_TOKEN is configured on my repo but maybe i have to define  'sonar.projectKey' and 'sonar.organization'  as well?

melloware:
It was definitely working before but I had it running ONLY against `master` i see oyu switched it to run on PR branches which I don't think works?

Rapster:
Debug console

> ##[debug]Evaluating: secrets.SONAR_TOKEN
> ##[debug]Evaluating Index:
> ##[debug]..Evaluating secrets:
> ##[debug]..=> Object
> ##[debug]..Evaluating String:
> ##[debug]..=> 'SONAR_TOKEN'
> ##[debug]=> null
> ##[debug]Result: null

Is SONAR_TOKEN var set on this repo?

melloware:
Yes I think its because you removed this `&& github.ref == 'refs/heads/master' &&` I believe the token is only set on the master repo and not on branches and forks which is why we have always run it only on `master`

Rapster:
Turns out, it's not even supported... See https://portal.productboard.com/sonarsource/1-sonarcloud/c/50-sonarcloud-analyzes-external-pull-request

Rapster:
i'll simply revert...

primefaces/src/main/java/org/primefaces/model/file/NativeUploadedFile.java
```diff
@@ -40,7 +38,7 @@ public NativeUploadedFile() {
     }
 
     public NativeUploadedFile(Part source, Long sizeLimit, String webKitRelativePath) {
-        super(source, FileUploadUtils.requireValidFilename(source.getSubmittedFileName()), sizeLimit, webKitRelativePath);
+        super(source, source.getSubmittedFileName(), sizeLimit, webKitRelativePath);
```
Rapster, line 41: @melloware I think you introduce this check FileUploadUtils.requireValidFilename but it's already done in AbstractUploadedFile. I fear quality gate might failed because of that, and if so, I guess it'll be a false postive

melloware: Yep you are right its because Sonar can't tell that later on we actually do properly handle it.

melloware (approved):

PR #10612, opened 2023-09-07
Fix #10609: Colorpicker double conversion

### LINKED ISSUE #10609: ColorPicker: Submitted value set to converted value.
### Describe the bug

ColorPickerRenderer.decode() sets submitted value to converted value:
`colorPicker.setSubmittedValue(ComponentUtils.getConvertedValue(context, component, submittedValue))`

This causes problems when using a converter: getAsObject() is called twice, first time with correct (submitted) value and second time with already converted value.

Obviously, converter can be made to work around this, but I doubt this is expected behaviour.

### Reproducer

https://github.com/primefaces/primefaces/files/12552676/pf-10609.zip

### Expected behavior

_No response_

### PrimeFaces edition

Community

### PrimeFaces version

13.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0.3

### Java version

17

### Browser(s)

_No response_

melloware:
Fix #10609: Colorpicker double conversion

I have confirmed that removing this line causes it to be called only once and properly.

**Before:**
```
getAsString null
getAsObject #853232
getAsObject 853232
Color = 853232
```

**After:** 
```
getAsString null
getAsObject #853232
Color = 853232
```

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/colorpicker/ColorPickerRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/colorpicker/ColorPickerRenderer.java
@@ -32,7 +32,6 @@ import javax.faces.component.UINamingContainer;
 import javax.faces.context.FacesContext;
 import javax.faces.context.ResponseWriter;
 import javax.faces.convert.Converter;
-
 import org.primefaces.renderkit.InputRenderer;
 import org.primefaces.util.*;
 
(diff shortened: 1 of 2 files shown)


primefaces/src/main/java/org/primefaces/component/colorpicker/ColorPickerRenderer.java
```diff
@@ -161,7 +160,6 @@ protected void encodeScript(FacesContext context, ColorPicker colorPicker, Strin
                 .attr("locale", colorPicker.calculateLocale(context).toString())
                 .attr("mode", colorPicker.getMode())
                 .attr("defaultColor", value, null)
-                .attr("theme", colorPicker.getTheme())
```
melloware, line 164: This was declared twice.

Issue #10605, opened 2023-09-06
Core: convert component attributes to enums when fixed values are expected 


jepsar:
See for example #10576. We should only allow values from an enumeration. If any other value is used, an exception should be thrown.

PR #10623, opened 2023-09-08
#10605: Colorpicker switch to enums

melloware:
@jepsar et all I decided to use ColorPicker as the demo for what this might look like.  What is nice is EL already coerces to ENUM's so this will make the code way more robust in the future.

When you put an invalid value on the XHTML like `theme="xxx"` and that is not in the enum you get this error.

```
javax.el.ELException: Cannot convert [xxx] of type [class java.lang.String] to [class org.primefaces.component.colorpicker.ColorPickerTheme]
        at org.apache.el.lang.ELSupport.coerceToEnum(ELSupport.java:224)
        at org.apache.el.lang.ELSupport.coerceToType(ELSupport.java:536)
```


tandraschko:
i general i like it but not sure if its worth TBH

also e.g. "hex" actually needs by HEX by naming conventions and you have to introduce HEX("hex") for each enum
really not sure

melloware:
Yeah I know java conventions are `HEX("hex")` but man that would be a lot of work.  So maybe we should scrap this?

tandraschko:
doers the EL coerce work correctly for ColorPickerTheme#default?

melloware:
Crap I don't think it does. I didn't even realize that.

tandraschko:
Then we can just close it?

We can just validate enum like properties in the renderer then, in development stage

PR #10626, opened 2023-09-08
Fix #10557: Rating not setting value on cancel event

### LINKED ISSUE #10557: Rating: clear does not work correctly
### Describe the bug

image

Pressing on the clear icon clears the rating but only on the widget (exactly as on the https://www.primefaces.org/showcase/ui/input/rating.xhtml?jfwid=55d47)

image

While in the GUI it looks cleared, the value in the bean stays unchanged.

image

Adding <p:ajax> does not help either. In any case, the corresponding set attribute method of the bean is not invoked when the rating value is cleared.

Thanks in advance for looking into this issue. 

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

13.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

8

### Browser(s)

_No response_

melloware:
Fix #10557: Rating not setting value on cancel event

This fixes it but I still can't explain why this is happening.  This component is very simple and the encode is setting the `submittedValue(null)` but it still has the value.

- [x] Integration Tests Updated

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/rating/Rating001.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/rating/Rating001.java
@@ -51,7 +51,7 @@ public class Rating001 implements Serializable {
     }
 
     public void oncancel() {
-        FacesMessage message = new FacesMessage(FacesMessage.SEVERITY_INFO, "Cancel Event", "Rate Reset");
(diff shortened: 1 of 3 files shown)


tuerker:
@melloware would be great to have this fix already in 13.0.2. many thanks.

PR #10661, opened 2023-09-12
Fix #10656 - DataExporter: Introducing bufferSize to avoid OOM

### LINKED ISSUE #10656: DataExporter: OOM error as it's trying to load everything at once
### Description

### Additional context

_No response_

Rapster:


jasonex7:
I think rows value is not a suitable value for exporting. That "rows" value was intended to use as screen pagination. Maybe a new dataTable property should be defined.

melloware:
I think a new property is not worth it as the datatable already has too many props.  This should load N rows at a time and then allow them for garbage collection it will definitely be slower than what is happening now but will allow for less memory usage. Everything is a trade off.  So far in 13 years of using the exporter I have never run into this issue. 

If your table has 50000 rows you should be writing a REST service custom download and not just deferring to the DataExporter if you need fine grained control of your memory usage. 

Just my two cents.

jasonex7:
If you don't recommend to add a new property, I prefer provide a custom LazyDataModel implementation that limit the maximum rows loaded in memory and allow the user decide to use or not. I've done that in my apps and I wanted to contribute that to PrimeFaces. Please wait my PR and you can see how I've handled that for years.

jasonex7:
I've just submit my PR. I'm not agree with this PR because it will affect speed of data exporter way too much. I think it will do more harm that good.

Rapster:
I can understand loading `DataTable#rows` items at a time is not ideal (specially when the value is 10 like in the sample), @melloware this new property could be on DataExporter component or very simple, we force to load 100 items at a time?

Rapster:
> Ok we just have to be careful because JpaLazyModel will be firing queries for each call instead of one query. 

And that's a problem?

Rapster:
No, I don't think so... On many front, loading 5000 rows is a bad idea, whether it's fetching 5000 rows from your db, or having 5000 rows circulating on your network, or load 5000 rows in memory: all of this is bad. I'm not even sure if JPA can handle that much. There is also another decisive factor which is the size of your item...

Having n request fetching m items is usually the best way to go. We should simply decide what should be the default value, 200? 500?

melloware:
That has not been my experience with JPA and I can verify it quite easily.  My guess is with your change the export will be 10x-20x slower which will be easily reproducible

melloware:
Just imagine your database is in the cloud like azure or your mongo instance is Mongo Atlas their cloud offering. Every round trip is more overhead.

melloware:
Cc @FlipWarthog @jepsar i am happy to be overridden here for my argument of the current behavior should remain the default. 

Let me make the argument a different way. With this change:
- assumes there is a memory problem which in 13 years I have not hit yet
- Punishes everyone on JPA, JdBC, Mongo performance even if it's slight it could be drastic
- To me this is a breaking change.  We broke performance in the name of memory

Rapster:
So if you have a billion items to display, you load a billion entries? I don't think it's "breaking performance in the name of memory", it might actually be the opposite. `DataExporter#loadChunkSize` can be optional and `DataTable#rowCount` can be used instead

melloware:
OK yes if you make it a DataExporter option and if its omitted the chunk size is the whole table count I am good with that.

jasonex7:
I'm agree with @melloware to include a new property in <p:dataExporter> so one could set the maximum number of rows loaded in memory if the developer knows it is a high number of rows. The default should be all rows like it is now. It would be nice to set a global default value in web.xml maybe

Rapster:
Just trying to figure out a name, so far I came up with lazyChunkSize on dataExporter, any suggestions?

melloware:
@jasonex7 you should be able to set it globally in your application using: https://primefaces.github.io/primefaces/13_0_0/#/core/globalattributes?id=global-attributes

@Rapster maybe just `bufferSize` to be concise?

jasonex7:
> @jasonex7 you should be able to set it globally in your application using: https://primefaces.github.io/primefaces/13_0_0/#/core/globalattributes?id=global-attributes

That procedure only works for components, but <p:dataExporter> is a tag handler which it is different.

FlipWarthog:
@melloware @Rapster we've got primefaces-test jpa branch set up with Hibernate and JpaLazyDataModel. Might be worth adding data exporter and a lot more test data to see what it does. It's in memory H2 so it won't show network latency for queries, but we can at least see the volume of queries in testing.

primefaces/src/main/java/org/primefaces/component/export/ExportConfiguration.java
```diff
@@ -187,4 +194,8 @@ public MethodExpression getOnRowExport() {
     public OutputStream getOutputStream() {
         return outputStream;
     }
+
+    public Integer getbufferSize() {
```
jasonex7: Here it should be getBufferSize() with capital "B"

primefaces/src/main/java/org/primefaces/component/export/DataExporter.java
```diff
@@ -377,6 +386,11 @@ public Builder onRowExport(MethodExpression onRowExport) {
             return this;
         }
 
+        public Builder bufferSize(ValueExpression lazyChunSize) {
```
jasonex7: Method parameter should be named bufferSize as the property.

primefaces/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
```diff
@@ -94,34 +90,41 @@ protected void exportPageOnly(FacesContext context, DataTable table) {
...
-        int first = table.getFirst();
-        int rowCount = table.getRowCount();
-        int rows = table.getRows();
-        boolean lazy = table.isLazy();
-
-        if (lazy) {
-            LazyDataModel<?> lazyDataModel = (LazyDataModel<?>) table.getValue();
-            List<?> wrappedData = lazyDataModel.getWrappedData();
-
-            if (rowCount > 0) {
```
jasonex7, line 106: I think the method should maintain the validation to only call load methods if rowCount is greater that 0 to avoid hit an innecesary query to LazyDataModel.

primefaces/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
```diff
@@ -94,34 +90,41 @@ protected void exportPageOnly(FacesContext context, DataTable table) {
...
+        if (table.isLazy()) {
+            LazyDataModel<Object> lazyDataModel = (LazyDataModel<Object>) table.getValue();
+
+            // bufferSize is used to control how many items are fetched at a time.
+            // The purpose of using this variable is to retrieve the entire underlying dataset in smaller,
+            // manageable chunks rather than all at once. This approach is often used to improve performance
+            // and reduce the load on system resources when dealing with large datasets.
+            Integer bufferSize = exportConfiguration.getbufferSize();
+            boolean bufferized = bufferSize != null;
+            int batchSize = Objects.requireNonNullElseGet(bufferSize, table::getRowCount);
```
jasonex7: Sames as @melloware, I don't recommend to default to table.getRowCount as batch size. If the user didn't set bufferSize property in <p:dataExporter>, this method should load all rows as before.

melloware (approved):

primefaces/src/main/java/org/primefaces/component/export/ExportConfiguration.java
Rapster: Good catch

primefaces/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
Rapster: Not sure what you're saying... If bufferSize not set then load all rows. Am I missing something?

primefaces/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
Rapster: In case bufferSize set, no point to previously call getRowCount. Though, I can keep it if not bufferized

primefaces/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
melloware: yes the default behavior should be the current behavior.  ALL ROWS so as to not make it a breaking change. But I thought that is what `table.getRowCount()` did right?

Rapster: yes, but maybe LazyDataModel#count should be called instead

jasonex7: @Rapster you're OK `table.getRowCount()` has the total rows of LazyDataModel. I confused with "rows" property.

primefaces/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
jasonex7: Maybe batchSize could be the less value between bufferSize and rowCount so if rowCount is zero, then LazyDataModel will not be hit. Maybe this

```java
int batchSize = Objects.requireNonNullElseGet(bufferSize, table::getRowCount);
batchSize = Math.min(batchSize, table.getRowCount());
```

Rapster: current impl is good enough

melloware (approved):
LGTM

primefaces/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
```diff
@@ -94,34 +90,39 @@ protected void exportPageOnly(FacesContext context, DataTable table) {
...
-                table.setRows(rowCount);
-                table.loadLazyDataIfEnabled();
+        if (table.isLazy()) {
+            // bufferSize is used to control how many items are fetched at a time.
+            // The purpose of using this variable is to retrieve the entire underlying dataset in smaller,
+            // manageable chunks rather than all at once.
+            LazyDataModel<Object> lazyDataModel = (LazyDataModel<Object>) table.getValue();
+            Integer bufferSize = exportConfiguration.getBufferSize();
+            boolean bufferized = bufferSize != null;
+            int batchSize = Objects.requireNonNullElseGet(bufferSize, () -> lazyDataModel.count(table.getFilterByAsMap()));
```
Rapster: Note for myself: table.getActiveFilterMeta() instead

PR #10662, opened 2023-09-13
Fixes #10656 - DataExporter: OOM error as it's trying to load everything at once

### LINKED ISSUE #10656: DataExporter: OOM error as it's trying to load everything at once

jasonex7:
Fixes #10656 by providing a custom LazyDataModel that limit the maximum number of rows that could be loaded in memory.


melloware:
I think I prefer @Rapster idea of making it a DataExporter option which by default will be ALL rows unless overridden to a chunk size.

jasonex7:
I also prefer the proposed solution of new property in dataExporter tag handler, so I close this PR.

PR #10676, opened 2023-09-17
Fix regression after trying to fix #10656

### LINKED ISSUE #10656: DataExporter: OOM error as it's trying to load everything at once

Rapster:
Quite unlucky on this one, i've been testing the whole time with value 10 which is same value as rows value. Once bufferSize gets higher than pageSize, rowIndex is recalculated in LazyDataModel#setRowIndex (for reason I don't know...) and goes back to 0

@jasonex7 You can checkout this PR and try if it works for you (just to be sure this time 😄 )


melloware (approved):

jasonex7:
I've tested with last version and it works fine. Thanks @Rapster

PR #10678, opened 2023-09-18
Fix #10677: JPALazyDataModel case sensitive filters

### LINKED ISSUE #10677: Docs: Please add a reference to FlowLogix JPA LazyDataModel to docs
### Description

Hi,

I do believe that https://docs.flowlogix.com/#section-jpa-lazymodel is the best and simplest implementation of JPA LazyDataModel. I request it be added to the docs and showcase. Thank you

Please add https://docs.flowlogix.com/#section-jpa-lazymodel to docs and showcase

### Additional context

_No response_

melloware:
cc @FlipWarthog 

We took this code from our current implementation at a client to make the string filters case sensitive by default but can be disabled with `setCaseSensitive(false);`

### DIFF
--- a/docs/14_0_0/components/datatable.md
+++ b/docs/14_0_0/components/datatable.md
@@ -953,7 +953,17 @@ or provide a existing JSF `Converter`:
 new JpaLazyDataModel<>(MyEntity.class, () -> entityManager, myConverter);
 ```
 
+or use the builder pattern to enable case insensitive searching and other features:
+
+```java
+JpaLazyDataModel<MyEntity> lazyDataModel = new JpaLazyDataModel.Builder<>(MyEntity.class, () -> entityManager)
+    .rowKeyField("id")
+    .caseSensitive(false)
+    .build();
+```
+
 Also you can add global filters via:
+
 ```java
 new JpaLazyDataModel<>(MyEntity.class, () -> entityManager) {
     @Override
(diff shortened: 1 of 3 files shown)


FlipWarthog:
Couple notes:
1) This changes the default behavior (for the better, imo) but we still might want a note in the migration guide to call it out
2) In the current impl, EXACT is treated the same as EQUALS so this won't apply to EXACT as-is (which I think it should)

In our impl we do
```java
if (filter.getMatchMode() == MatchMode.EXACT) {
   return cb.like(newExpression.as(String.class), newFilterValue.toString());
}
```
and bypass the call to
```java
return super.createPredicate(filter, filterField, root, cb, newExpression, newFilterValue);
```
for this exact reason. 

But I think this can be handled by simply adding:
```java
case EXACT:
    return cb.equal(fieldExpressionAsString.get(), getStringFilterValue(filterValue));
```

melloware:
OK I fixed the `EXACT` and you are right caseSensitive=true should be the default so its not a breaking change.

melloware:
@tandraschko wasn't sure if this is what you were thinking for builder pattern.

```java
JpaLazyDataModel<MyEntity> lazyDataModel = new JpaLazyDataModel.Builder<>(MyEntity.class, () -> entityManager)
    .rowKeyField("id")
    .caseSensitive(false)
    .build();
```

tandraschko:
lets first discuss a bit further

melloware:
No problem. I figured as much and left it in draft mode.

tandraschko:
DefaultStreamedContent.builder() is built a bit different, can you align?

melloware:
@tandraschko when trying to align with DefaultStreamedContent I couldn't get the generics to work unless I am missing something?

DefaultStreamedContent does this:
```java
    public static Builder builder() {
        return new Builder();
    }
```

But for JPALazyDataModel I need the generics?
image


I don't know what to make `<T>` or `<?>` ???

melloware:
Thanks I will submit PR for case sensitive.

PR #10684, opened 2023-09-19
Fix #10681: JPALazyDataModel case sensitive filters

### LINKED ISSUE #10681: JpaLazyDataModel: allow case-sensitive configuration

melloware:
Fix #10681: JPALazyDataModel case sensitive filters

@tandraschko I updated the docs too if you don't mind reviewing what I updated?

### DIFF
--- a/docs/14_0_0/components/datatable.md
+++ b/docs/14_0_0/components/datatable.md
@@ -944,13 +944,27 @@ new JpaLazyDataModel<>(MyEntity.class, () -> entityManager);
 If you have selection enabled, you can either pass the rowKey field name in the constructor:
 
 ```java
-new JpaLazyDataModel<>(MyEntity.class, () -> entityManager, "id");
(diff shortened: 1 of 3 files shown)


melloware:
Also should we provide a setter for `caseSensitive` etc. because of this scenario where you want to override methods using the constructor way and not the builder way?

```java
JpaLazyDataModel<MyEntity> lazyDataModel= new JpaLazyDataModel<>(MyEntity.class, () -> entityManager) {
    @Override
    protected void applyGlobalFilters(CriteriaBuilder cb, CriteriaQuery<?> cq, Root<MyEntity> root, List<Predicate> predicates) {
        predicates.add(cb.isNull(root.get("id")));
    }
};

lazyDataModel.setCaseSensitive(false);

tandraschko:
yup why not
if we provide a setter, we can also provide getter

PR #10686, opened 2023-09-19
Fix #10685: JPALazyDataModel wildcard support

### LINKED ISSUE #10685: JpaLazyDataModel: Allow wildcard search
Another feature we have in our custom code is allowing using to use `*` and `?` in an EXACT search.

Example Name field with search "Melloware" will generate:

```
where name = 'Melloware`
```
Example Name field with search "Mello*" will generate:

```
where name like 'Mello%`
```

Example Name field with search "Mello?wa*" will generate:

```
where name like 'Mello_wa%`
```

cc @FlipWarthog

melloware:
Fix #10685: JPALazyDataModel wildcard support

cc @FlipWarthog please review since I am translating your original code.

### DIFF
--- a/docs/14_0_0/components/datatable.md
+++ b/docs/14_0_0/components/datatable.md
@@ -959,11 +959,19 @@ JpaLazyDataModel<MyEntity> lazyDataModel = JpaLazyDataModel.builder(MyEntity.cla
     .build();
 ```
 
-or use the builder pattern to enable case insensitive searching and other features:
+or enable case insensitive searching:
 
 ```java
 JpaLazyDataModel<MyEntity> lazyDataModel = JpaLazyDataModel.builder(MyEntity.class, () -> entityManager)
     .caseSensitive(false)
+    .build();
+```
+
+or enable wildcard support so `*` and `?` can be used like `Hello*` to find all words starting with `Hello`:
+
+```java
+JpaLazyDataModel<MyEntity> lazyDataModel = JpaLazyDataModel.builder(MyEntity.class, () -> entityManager)
+    .wildcardSupport(true)
     .build();
 ```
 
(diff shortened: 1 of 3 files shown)


tandraschko:
hmm, i really wonder about this one
you make a filter "exact" but support still wildcards
shouldnt you just use "like"?

melloware:
I will let @FlipWarthog chime in but I am pretty sure our use case was say on a Last Name field and we have 1,000,000 users.

If the box is Starts With or Contains then searching for last name "Smith" returns.
```
John Smith
Mike Smithson
Steve Naismith
```

We want it just to return `John Smith` for the user. But then the user might want to say search for `Smith*` and get 

```
John Smith
Mike Smithson
```

Or search for `*smith*` and get 

```
John Smith
Mike Smithson
Steve Naismith
```

All in 1 search box.  Plus when its an `=` query we use Oracle and the Oracle index is optimized for = vs a case insensitive UPPER index.

I hope that makes sense?

FlipWarthog:
@melloware you're exactly correct. There are related requirements for other fields but that sums it up. We need the benefit of proper indexing when searching for exact values, but also allow users to "fuzzy find" things via the filters too, without having full blown complex filter selection like the [JS libs offer](https://primereact.org/datatable/#basic_filter).

FlipWarthog:
For more context, this is the "Search Tips" dialog to inform users how it works. Obviously we'd need to update it based on this implementation, but I think that's for the better.

image

tandraschko:
i see
so please enhance the docu a bit that
1) * and % means any char/any amount
2) _ and ? means any chart/1 amount

tandraschko:
and shouldnt it be `Wild**c**ard`?

melloware:
Docs updated.

PR #10700, opened 2023-09-22
Remove unused UITable#isDefaultSort and UITable#isDefaultFilter

Rapster:
Not sure why it's been introduced in the first place... :/


melloware:
Can you look in Git Blame or Log?  I feel like this was to solve some MVS issue?

Rapster:
yeah I tried that, but the original ticket has too many commits to review 😫 ... But so far these property really seems useless, the PR shows that pretty clearly

melloware (approved):
LGTM and all tests are passing.

PR #10724, opened 2023-09-28
Fix #10684: JPA case sensitive use Faces Locale

melloware:
Fix #10684: JPA case sensitive use Faces Locale

### DIFF
--- a/primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
+++ b/primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
@@ -49,6 +49,7 @@ import org.primefaces.util.BeanUtils;
 import org.primefaces.util.Constants;
 import org.primefaces.util.LangUtils;
 import org.primefaces.util.Lazy;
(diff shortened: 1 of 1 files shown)


melloware:
Actually one more fix

primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
```diff
@@ -220,22 +221,24 @@ protected Predicate createPredicate(FilterMeta filter, Field filterField,
         Lazy<Collection<Object>> filterValueAsCollection = new Lazy(
                 () -> filterValue.getClass().isArray() ? Arrays.asList((Object[]) filterValue)
                         : (Collection<Object>) filterValue);
+        FacesContext context = FacesContext.getCurrentInstance();
```
tandraschko: createPredicate is also called in a loop

melloware: Moved it up higher outside the loops.

tandraschko:
+1

PR #10738, opened 2023-09-28
Fix #10737 - DataTable: binded rows value is not reset with his original value

### LINKED ISSUE #10737: DataTable: original value of rows if not reset with his original value
### Describe the bug

Let's say I declare a datatable like this:

```xml
<p:dataTable value="#{bean.value}" rows="#{bean.rows}">
       // something
</p:dataTable>

```

If I decide to change number of rows per page to 10, and then call `DataTable#reset`: number of rows per page is still 10 while it should be 3

### Reproducer

Based on dataTable001.xhtml from ITs: simply declare as the above

### Expected behavior

Calling `DataTable#reset` should reset `rows` to this original value

Additionnal context: #5465

### PrimeFaces edition

Community

### PrimeFaces version

14.0.0-SNAPSHOT

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

8

### Browser(s)

_No response_

Rapster:
Fix #10737

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datatable/DataTable001.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datatable/DataTable001.java
@@ -25,16 +25,14 @@ package org.primefaces.integrationtests.datatable;
 
 import java.io.Serializable;
 import java.util.List;
-
 import javax.annotation.PostConstruct;
 import javax.faces.context.FacesContext;
 import javax.faces.view.ViewScoped;
 import javax.inject.Inject;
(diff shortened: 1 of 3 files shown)


christophs78 (approved):
Looks good to me, no arguments against changing this.
On the IT-side i´ll prefere dedicated IT for this over just modifiying an existing one a bit. (Think there already should be some regarding reset-behaviour.)

melloware (approved):

Rapster:
@christophs78 You're right, as I'm planning to refactor the ITs (long work in progress... it'll be covered at that time)

Issue #9860, opened 2023-03-02
FileUpload: Apache Commons FileUpload 1.5 Security Update


poikilotherm:
### Describe the bug

There is a new version of fileupload that offers a new setting to limit number of files uploaded at the same time to avoid DoS attacks. Please find details in https://cve.circl.lu/cve/CVE-2023-24998

It's important to mention that one needs to configure the maximum file numbers - it's not enough to "just update the dependency".

### PrimeFaces edition

None

### PrimeFaces version

12.0.0

### JSF implementation

Mojarra

### JSF version

3.0.0

### Java version

11.0.18

tandraschko:
i think we just could add a new param to the FileUploadFilter called fileCountMax
and add it to the docs, thats all we can do

PR #10752, opened 2023-09-29
Fix #10751: Apache Commons FileUpload removed

### LINKED ISSUE #10751: FileUpload: Deprecate Apache Commons FileUpload
Now that 14.0.0 is on Servlet 3.0 and JSF2.3 we should drop all the code for Apache Commons FileUpload since the Native Servlet Parts work great and have been heavily tested over the years.

There should be no need for this anymore right?

melloware:
Fix #10751: Apache Commons FileUpload removed

- [x] Remove all code
- [x] remove special nightly integration tests for file uploads
- [x] Update all documentation
- [ ] Check old issues for it works on Apache Commons only issues


melloware:
@Rapster please confirm `UploadedFileCleanerListener ` was only needed for Servlet 2.5 like the docs said and is safe to remove now?

melloware:
@christophs78 do we need `nightly-file-upload.yml` work now since we no longer have native and Commons uploads it can just be part of the main IT's?

christophs78:
> @christophs78 do we need `nightly-file-upload.yml` work now since we no longer have native and Commons uploads it can just be part of the main IT's?

Yep. Think we can get rid of this.

melloware:
@christophs78 thanks I just removed it entirely!

tandraschko:
can you go through old issues and check if we had something like "works only with commons"?

melloware:
Yes I will go through all the old issues.

melloware:
The only real feature I can find that is not supported by native out of the box is https://github.com/primefaces/primefaces/issues/9860

But I assume since its Servlet spec there is some servlet property to control max file uploads?

PR #10766, opened 2023-10-04
Fix #10755 - AjaxStatus: Don't cancel timeout when redirect is received

### LINKED ISSUE #10755: AjaxStatus: Don't cancel timeout when redirect is received
### Description

We switched from custom js (responding to pfAjaxStart/pfAjaxSuccess/etc) to p:ajaxStatus a few days ago for displaying a loading indicator in our application for long running interactions.

Our setup now is basically just `<p:ajaxStatus onstart="..." onsuccess="..." delay="1000" />`

What we're experiencing is the following:
- user interacts with some UI component
- ajax response with a redirect is received (fairly quick, couple ms)
- user gets redirected (slow page, takes a couple of seconds to load), never actually sees any loading indicator

I think this feature request is similar to #10708, I'd like the timeout to only be cancelled if the xhr response is not a redirect allowing the "start" facet/callback to work in a case of fast ajax + slow redirect

### Additional context

_No response_

tandraschko:
Fix #10755


Rapster:
Please follow this convention: `Fix #10755 - AjaxStatus: Don't cancel timeout when redirect is received`

Otherwise only `#10755` will show in changelog

PR #10769, opened 2023-10-04
fix primefaces#10763: TreeTable: Filtering with CustomNode and cloneOnFilter does not work (primefaces#10763)

### LINKED ISSUE #10763: TreeTable: Filtering with CustomNode and cloneOnFilter does not work
### Describe the bug

The filtering on a TreeTable with cloneOnFilter="true" and custom TreeNodes does not work.
We have a simple p:treeTable
```
<p:treeTable id="treeTable"
                     value="#{testView.customTree}"
                     nodeVar="node"
                     var="data"
                     cloneOnFilter="true"
                     selectionMode="single"
                     emptyMessage="empty">

                <p:column headerText="Name"
                     filterMatchMode="contains"
                     filterBy="#{data}">
                     #{data}
                </p:column>   

</p:treeTable>
```
            
with a CustomNode type which extends DefaultTreeNode and implements the copy constructor required by cloneOnFilter="true"
```
public class CustomNode<T> extends org.primefaces.model.DefaultTreeNode<T>
{
    public CustomNode(String type, T data, TreeNode parent)
    {
        super(type, data, parent);
    }

    public CustomNode(CustomNode<T> other)
    {
        super(other.getType(), other.getData(), other.getParent());
    }
}
```

The filterMatchMode is contains which means it should show one column when searching for "X" but it shows an empty tree.
image
image

I think the problem is that in org.primefaces.component.treetable.feature.FilterFeature in the method protected TreeNode cloneTreeNode(TreeTable tt, TreeNode<?> node, TreeNode<?> parent) the newly cloned root is not set as parent of the cloned child nodes. Therefore all the nodes get cloned but set in the original treeRoot and not in the clonedTreeRoot, which is why the TreeTable is empty after filtering since the newly cloned root will be shown but it has no children.
```
...
            else {
                try {
                    Constructor<? extends TreeNode> ctor = node.getClass().getConstructor(node.getClass());
                    clone = ctor.newInstance(node); // <- copy constructor argument is oldChild which has the oldRoot, setParent(parent) is needed here
                }
                catch (NoSuchMethodException e) {
                    // ignore
                }
...
```

### Reproducer

I have used the primefaces-test jakarta version from here https://github.com/primefaces/primefaces-test/tree/jakarta
To start: mvn clean jetty:run -Pmyfaces40

[primefaces-test-jakarta.zip](https://github.com/primefaces/primefaces/files/12793477/primefaces-test-jakarta.zip)

The filterMatchMode is contains which means it should show one column when searching for "X" but it shows an empty tree.
image
image

### Expected behavior

It should show one row with "NameX" when the filter value is "X" or in other words filtering with custom treeNodes and cloneOnFilter should work

### PrimeFaces edition

None

### PrimeFaces version

13.0.1

### Theme

_No response_

### JSF implementation

MyFaces

### JSF version

4.0

### Java version

11

### Browser(s)

_No response_

Maverick-501:
Fix #10763

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/treetable/feature/FilterFeature.java
+++ b/primefaces/src/main/java/org/primefaces/component/treetable/feature/FilterFeature.java
@@ -243,6 +243,9 @@ public class FilterFeature implements TreeTableFeature {
                 try {
                     Constructor<? extends TreeNode> ctor = node.getClass().getConstructor(node.getClass());
                     clone = ctor.newInstance(node);
+                    if (parent != null) {
+                        parent.getChildren().add(clone);
+                    }
                 }
                 catch (NoSuchMethodException e) {
                     // ignore
--- a/primefaces/src/test/java/org/primefaces/component/treetable/feature/CustomNode.java
+++ b/primefaces/src/test/java/org/primefaces/component/treetable/feature/CustomNode.java
@@ -1,0 +1,14 @@
+package org.primefaces.component.treetable.feature;
+
+import org.primefaces.model.TreeNode;
+
+public class CustomNode<T> extends org.primefaces.model.DefaultTreeNode<T> {
+
+    public CustomNode(String type, T data, TreeNode parent) {
+        super(type, data, parent);
+    }
+
+    public CustomNode(CustomNode<T> other) {
+        super(other.getType(), other.getData(), null);
+    }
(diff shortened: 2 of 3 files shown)


melloware:
Thank you for the PR and the test!  Running the build now.

Issue #7699, opened 2021-08-12
Datatable: LazyDataModel+RowEditor breaks with Session serialized/restored


t-oster:
**Describe the defect**
When using Datatable with a lazy data model and a row editor, as soon as the session is serialized/deserialized between the requests (e.g. with spring session for load-balancing), the model does not get updated anymore before the RowEditEvent is processed.

**Reproducer**
See my clone of the primefaces-test repo.
This commit https://github.com/t-oster/primefaces-test/commit/aeee1eeb4195ce2730958c107e6446cd85f820b1 introduces the bug by just enabling session serializing/restoring on every request.

**To Reproduce**
Create a Datatable with Row editor and lazy data model. You can use this as working example:
https://github.com/t-oster/primefaces-test/commit/a77ad2ffbd95809d942c15df3378bb21c9c2c04e
If you edit a row, the updated value is shown in the growl.
Apply the next commit  https://github.com/t-oster/primefaces-test/commit/aeee1eeb4195ce2730958c107e6446cd85f820b1  and retry. Now the setSampleValue() is never called during postback, eventhough the values are sent as POST and the row edit event is fired.

tandraschko:
Can you please try a 11.0 Snapshot?
We fixed around 50 datatable bugs since 10.0

t-oster:
Thanks for the quick response. Can you give me an exact version number for the pom.xml?
I just tried 
```
<dependency>
     <groupId>com.github.primefaces</groupId>
     <artifactId>primefaces</artifactId>
     <version>master-SNAPSHOT</version>
 </dependency>
```
but I cannot get it to compile/run.

tandraschko:
Check our readme how to configure jitpack
Otherwise build from source

t-oster:
It downloads the latest version from jitpack and compiles. But when it starts it only complains about a number of duplicate jar files and bails out:
[...]

PR #10776, opened 2023-10-04
LazyDataModel: clean up

Rapster:
I'll try to sum up:

- There is no need to inherit from `ListDataModel `since #7699 
- `LazyDataModel` should be as much immutable as it can be (for internal purposes setRowCount and more can have setters)
- Builder with parameters is a repeat of the problem you're trying to solve (ctors with multiple parameters)
- Give the possibility to plug your own PD cache (in our app, we have a similar cache which is already huge, if you create another one, that might be problematic, also some users delegate their cache to PropertyUtils from beanutils)
- Rename `ReflectionLazyDataModel` into `ReflectionDataModel`: there is nothing lazy about it, using `LazyDataModel` is just a nice trick to improve perf
- Rename` ReflectionDataModel.Builder#rowKey` into `rowKeyField`: it's confusing at first, in PF code, whenever we mentionned rowKey: it's a value, here it's important to mention which field will be used to determine the rowKey
- Move sort lambda into proper Comparator: it's much more cleaner and reusable


tandraschko:
just a short review:

- as long as LazyDataModel is called LazyDataModel, we should stay on ReflectionLazyDataModel, also compared to javax.faces.model.ListDataModel its much more lazy
- +1 for PDCache SPI; we just use the ServiceLoader API
- +1 for rowKeyField (but it currently you seems to removed the usage)

Rapster:
- It's pageable yes, but not lazy... Having something called "xxxLazyModel" while it's in memory is very confusing, people might ask "How it is lazy then?", again there is nothing lazy about it.
- Issue that I have with SPI is, how to deal with 2 impl. How do you choose only one? I don't want to do something like our virus scanners where people need to explicitly enable a cache. I'm not 100% sure either about the order the services are stored in the ServiceLoader. There is also thread-safety to properly deal with... Having a simple context-param seems simple enough, don't you think?
- rowKeyField usage is there

tandraschko:
1) then we should think about renaming LazyDataModel and JPALazyDataModel and create a new ticket for it; for now should we should just stay consistent

2) i dont understand your point about 2 impls. if you put a file in your webapp, the hierachie wins. just like with resources. also there should be no problem about threadsafity as it should return a new instance. so the same like context params, just the default in java to get SPI instances. JSF impl, CDI impl, and so on, everyone is using it.

in general we should think about if it isnt overengineered.
We generate new PropertyDescriptor instances but the internal methods are already existing in the JVM. So the memory overhead is almost nothing.
And the same caches are existing in EL impl, JSF impl, CDI impl, they also dont have a SPI. For sure, the cache in PF is only a small fraction of those caches.

Rapster:
- I think names are fine, JPALazyDataModel is lazy, and LazyDataModel could be renamed but does not worth it to make a breaking change just for that. On the other hand, since ReflectionDataModel is brand new and is a very particular case, best to choose a proper name now than create a BC later
- ServiceLoader#load will return all impls of a service, not just one

Overengineered or not, it did crossed my mind, but on our side it comes with very low-cost and people like me will be happy to not have x cache spread everywhere doing the very same thing and consuming unecessary memory. Also, this could be used anywhere in PF using reflection not only for ReflectionDataModel (I think I use it for JPALazyModel)

tandraschko:
1) need to think about it; i dont like when its not consistent in naming and its just a specialized LazyDataModel
2) ServiceLoader#iterator will return a ordered iterator; the first one is the one with the highest prio, same as the classloader hierachie

Rapster:
- I'm not a big fan of the name actually... I mean reflection is how it works under the hood, it doesn't say much. Maybe it could be simple as NonLazyDataModel 😄  
- that was my question, like I said I wasn't 100% sure of it but if you are, I'm cool with that

melloware:
I like the name `Reflection` because it is descriptive of what it is doing.  I think most users would intuitively know that unless they just started using Java. 😄 

I will have to double check after this gets merged whether any changes are needed to our MongoLazyModel: https://github.com/primefaces-extensions/primefaces-extensions/blob/master/core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java

Rapster:
IMO what you're describing is more how I would pick a method name not a class name, class name should give meaningful context just like MorphiaLazyDataModel, UserService etc. By mentionning "reflection" you don't give much.. reflection is more how we make it happen

TBH, I thought about NonLazyPageableDataModel but that would imply to rename LazyDataModel into PageableDataModel to be consistent, but I don't think it's a BC that worth to be introduced or is it?

I don't see a branch on PFE for 14 normal?

melloware:
What about `InMemoryDataModel` instead of `Reflection` ?

Rapster:
99% of datamodel are in-memory 😄 I think `NonLazyPageableDataModel` is the right one, and rename LazyDataModel into PageableDataModel and to avoid BC right now, I can create a LazyDataModel that simply inherits from LazyDataModel

melloware:
The only reason  I don't like "NonLazy" is you are telling the user that its NOT something rather than what it is...

melloware:
Hibernate uses "Eager" as the opposite of "Lazy" so EagerDataModel?

Rapster:
If eager means not "lazy", it means that again 99% of datamodel is eager no? 😅 I realize more and more that LazyDataModel is not the right name... If it was PageableDataModel, we could simply call it: DefaultPageableDataModel and I guess it would be okay

primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java
```diff
@@ -335,214 +328,132 @@ protected Expression resolveFieldExpression(CriteriaBuilder cb, CriteriaQuery<?>
...
-            model.setRowKeyConverter(rowKeyConverter);
+            Objects.requireNonNull(model.entityClass, "entityClass not set");
+            Objects.requireNonNull(model.entityManager, "entityManager not set");
 
+            if (model.rowKeyProvider == null) {
+                if (model.rowKeyConverter != null) {
+                    model.rowKeyProvider = model::getRowKeyFromConverter;
+                }
+                else {
+                    Objects.requireNonNull(model.rowKeyField, "rowKeyField is mandatory if no rowKeyProvider nor converter is provided");
```
tandraschko: actually not true, its only required if you have selection enabled
therefore i made the initilization lazy

primefaces/src/main/java/org/primefaces/util/BeanPropertyComparator.java
```diff
@@ -0,0 +1,160 @@
...
+        }
+        catch (Exception e) {
+            throw new FacesException(e);
+        }
+    }
+
+    public static BeanPropertyMapper valueExprMapper() {
+        return (context, var, sortMeta, o1) -> {
+            ValueExpression ve = sortMeta.getSortBy();
+            context.getExternalContext().getRequestMap().put(var, o1);
```
tandraschko: cleanup missing?

primefaces/src/main/java/org/primefaces/util/BeanPropertyComparator.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+import javax.el.ValueExpression;
+import javax.faces.FacesException;
+import javax.faces.context.FacesContext;
+
+import org.primefaces.component.api.UITable;
+import org.primefaces.context.PrimeApplicationContext;
+import org.primefaces.model.SortMeta;
+
+public class BeanPropertyComparator implements Comparator<Object> {
```
tandraschko: currently the name sounds like a independent reusable class, but its bound to UITable/SortMeta.
Feels like we should move+rename

primefaces/src/main/java/org/primefaces/util/LangUtils.java
```diff
@@ -506,51 +513,55 @@ private static boolean withDecimalsParsing(final String str, final int beginIdx)
         return true;
     }
 
-    public static Field getFieldRecursive(Class<?> clazz, String name) {
-        if (clazz == null) {
-            throw new IllegalArgumentException("clazz must not be null!");
-        }
-        if (name == null) {
-            throw new IllegalArgumentException("name must not be null!");
+    public static Object convertToType(Object value, Class<?> valueType, Class<?> logContext) {
```
tandraschko: its not realy a "language util", ComponentUtils is a better place

primefaces/src/main/java/org/primefaces/util/PropertyDescriptorResolver.java
```diff
@@ -0,0 +1,96 @@
...
+        public DefaultResolver() {
+            pdCache = new ConcurrentHashMap<>();
+        }
+
+        @Override
+        public PropertyDescriptor get(Class<?> klazz, String expression) {
+            PropertyDescriptor pd = null;
+            Class<?> parent = klazz;
+
+            for (String field : NESTED_EXPRESSION_PATTERN.split(expression)) {
```
tandraschko, line 56: this can lead to problems
if you check the current ReflectionLazyDataModel#getFieldValue, it loops on the object instances instead of of the property-descriptor

if you have e.g.
company.user.field1

class Company {
    private User user;
}

but the only the actual user instance, (e.g. a class SpecializedUser) has field1, the current code will fail?
we need some tests here

tandraschko:
still completely unsure about the name

LazyDataModel is actually a appropriate name, as it lazy loads a page. PageableDataModel would be another good name.
Is our current "naming problem" enough to rename this very very old class? i dont think so.
The LazyDataModel is designed for datasources which allows filtering/sorting via queries. Otherwise everything else could be done via the DataTable, UI centric, "FilterFeature" and so on.

IMO this was already a imperfect solution as we now had filtering/sorting as "UI centric" feature and a second time in the DataModel.

What exactly is the new ReflectionLazyDataModel:
- its a enhanced DataModel, to influence sorting and filtering (compared to add new component attributes, it's actually more related to logic instead of UI behavior); we needed so skip objects from filtering and always show some objects as first items.
- its a good performance improvement as it doesnt require any UI state to be applied for each row; and we avoid storing big lists in longer living scopes

What could it be in the future:
- it could be a way to remove filtering and sorting from DataTable and move it to the single place


Based on this points above, i think we should first clarify the question, if we would like to remove the filter and sorting from the DataTable itself?
What does this mean? If filter/sorting is used, we need to throw a exception if the ???DataModel is not used OR we automatically wrap it if its just binded via a simple List.

The second big question is, if we should really provide the current ReflectionLazyDataModel in PF itself?
IMO we should, otherwise i need to enhance the dataTable by some listeners-attribute, e.g. to skip filtering for single objects.
i dont like to that as its actually more related to logic as to UI.
its a no-go that users have to extend Sort/FilterFeatures. Its something very internal, therefor i never liked to have this replaceable for users.  It should be inside the DataModel.

Based on decisions, we have the following possibilities:

1) If we decide to do this, the ???DataModel is responsible for the whole "query" (get, filter, sort, paging). For this case, i would suggest something like:
- QueryableDataModel (current LazyDataModel)
- JpaDataModel / JpaQueryableDataModel
- DefaultQueryableDataModel (current ReflectionDataModel)

2) Leave the ReflectionLazyDataModel in PF just as something really special (for influence sorting/filtering like skipFiltering and a nice performance compared to binding big lists), then i would suggest to do the following:
- LazyDataModel (not really worth to rename)
- JpaLazyDataModel
- ???

3) decide to remove ReflectionLazyDataModel, add some UI attributes like skipFiltering and state in our docs: if you like better performance, please implement your own LazyDataModel

Rapster:
Great feedbacks thanks, will look into that ASAP!

Regarding few things you say:
- If base name should be choosen, I still strongly believed `PageableDataModel` is the right one. Simply, look at the main method `LazyDataModel#load(int first, int pageSize, Map<String, SortMeta> sortBy, Map<String, FilterMeta> filterBy)` it returns a page. Whether it does a query or not, I don't think it matters, proof is this new datamodel you introduced.
- Should we do rename it? Personnally I never has any issues with any BC in PF, but I can understand depending the project(s) you're working on, it might be a pain to do so... 
- Should we replace features with the reflection datamodel: yes! Though, at first, it should be an experimental feature that users can enabled (via context-param), hopefully from there, we should have more visibility as of we should keep or not corresponding features.

Side note: `JpaLazyDataModel` should be renamed `JPALazyDataModel`

tandraschko:
IMO paging is just one of the features of the datamodel, it also filters and sorts. Therefore PageableDataModel would be the same as FilteringDataModel or SortingDataModel. We need a name which covers this all. Queryable would be a candidate, there may be others. We can check other languages / frameworks for naming.

Also pageSize should be nullable actually, so that paging is just something additional. Same as empty FilterMeta or empty SortMeta.

primefaces/src/main/java/org/primefaces/util/BeanPropertyComparator.java
Rapster: I used same name as before (PF 8 or 10). But indeed not a great name... I renamed into `SortMetaComparator` (which 100% correct) but if you have a better name i'm all ears

primefaces/src/main/java/org/primefaces/util/BeanPropertyComparator.java
Rapster: I reused same logic as before, it gets cleaned after sort

primefaces/src/main/java/org/primefaces/util/PropertyDescriptorResolver.java
Rapster: `ReflectionLazyDataModel#getFieldValue `does the same thing as `DefaultResolver#getValue`. You actually didn't need to call `getFieldType` since you got the value before. Now about the issues you mentionned, they're present for `JPALazyDataModel` (whether you use the new resolver or the present master code)

If inheritance present, it gets tricky to determine property types with only `entityClass`

tandraschko: JPALazyDataModel is something different, you dont filter/sort on existing objects, you build JPA expressions. you need to rely on class infos.

in case of the reflection, you can resolve/filter/sort on company.user.field1 if you work on the current object tree, same as current EL-based sorting/filtering.
we should just make sure via tests that it works, therfore it was more like a question :D

Rapster: > in case of the reflection, you can resolve/filter/sort on company.user.field1 if you work on the current object tree
we should just make sure via tests that it works, therfore it wa more like a question :D

I know

> JPALazyDataModel is something different, you dont filter/sort on existing objects, you build JPA expressions. you need to rely on class infos.

Yup, but as I am not a JPA user (at least not using JPQL, and criteria stuff) i'd rather let you know. If you take that sample

https://github.com/primefaces/primefaces/blob/4ad098c54c8d22f482e4e6ed3be3945ed32c230e/primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java#L384-L396

If `A` is abstract (no attributes) and `B` extends `A` and have an `id`, and you configure `A` as your `entityClass` and `id` as your rowKey, you'll have a problem. Same thing with field:

https://github.com/primefaces/primefaces/blob/4ad098c54c8d22f482e4e6ed3be3945ed32c230e/primefaces/src/main/java/org/primefaces/model/JpaLazyDataModel.java#L175

I don't think there is an easy way to workaround this, maybe worth to mention as a limitation in JPALazyDataModel

Rapster:
Not convinced by Queryable, if the model was taking a query as input, it would probably make sense but that's not the case, also users are more familiar with Pageable than Queryable because it's more explicit? (see also https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/domain/Pageable.html Looks quite familiar don't you think?) 

Anyway, if you're not convinced by all of this, I can't come up with better reasons 😅 I simply rename into `DefaultLazyDataModel` for now...

melloware:
Not sure if this helps but why `Lazy` at all?  PrimeReact refers to this just as `DatatableState` IN the TypeScript we define a SortEvent, FilterEvent, PageEvent but all three are called `DatatableStateEvent` so why not just `TableStateModel` ?

```js
interface DataTablePageEvent {
    /**
     * Index of the first row.
     */
    first: number;
    /**
     * Rows per page.
     */
    rows: number;
    /**
     * The page number of the datatable.
     */
    page?: number;
    /**
     * Total number of pages.
     */
    pageCount?: number;
    /**
     * Extra options.
     */
    [key: string]: any;
}

/**
 * Custom sort event.
 * @see {@link DataTableProps.onSort}
 * @event
 */
interface DataTableSortEvent {
    /**
     * Field to sort against.
     */
    sortField: string;
    /**
     * Sort order as integer.
     */
    sortOrder: 1 | 0 | -1 | null | undefined;
    /**
     * MultiSort metadata.
     */
    multiSortMeta: DataTableSortMeta[] | null | undefined;
    /**
     * Extra options.
     */
    [key: string]: any;
}

/**
 * Custom filter event.
 * @see {@link DataTableProps.onFilter}
 * @event
 */
interface DataTableFilterEvent {
    /**
     * Collection of active filters.
     */
    filters: DataTableFilterMeta;
    /**
     * Extra options.
     */
    [key: string]: any;
}

/**
 * Custom state event containing page, filter and sort states.
 * @see {@link DataTableProps.onFilter}
 * @see {@link DataTableProps.onSort}
 * @see {@link DataTableProps.onPage}
 * @event
 */
interface DataTableStateEvent extends DataTablePageEvent, DataTableFilterEvent, DataTableSortEvent {
    /**
     * Extra options.
     */
    [key: string]: any;
}
```

Rapster:
Don't know anything about react, and don't know what they associate `state` with, but at least in java, I tend to associate "state" as in statefull/stateless or something similar as in a state machine but here I would have simply call it `DataTableEvent` : clear and simple as that (unless state is important to mention ?)

i think it's important to keep `DataModel`, and also we can't simply associate it a component like DataTable since it's been by other component like `<p:autoComplete />` if I remember well

melloware:
No problem just throwing it out there!  Thanks for all your work on this @Rapster

Rapster:
Ty!

primefaces/pom.xml
```diff
@@ -1226,7 +1226,7 @@
             <properties>
                 <maven.test.skip>true</maven.test.skip>
                 <maven.javadoc.skip>true</maven.javadoc.skip>
-                <maven.source.skip>true</maven.source.skip>
+                <maven.source.skip>false</maven.source.skip>
```
melloware: Is this an accident?

melloware: Was this an accident flipping this flag?

melloware (changes requested):

primefaces/pom.xml
Rapster: yes xD

primefaces/src/main/java/org/primefaces/util/SortTableComparator.java
```diff
@@ -0,0 +1,162 @@
...
+
+    public static Comparator<Object> fieldBased(FacesContext context, UITable<?> table) {
+        return new SortTableComparator(context, table, fieldMapper(), false);
+    }
+
+    @Override
+    public int compare(Object o1, Object o2) {
+        AtomicInteger result = new AtomicInteger(0);
+        for (SortMeta sortMeta : sortBy) {
+            if (sortByVEBased) {
```
Rapster, line 73: @tandraschko we can do better, and I'll do that in another PR. But if it's not dynamic there is no point calling invokeOnColumn, also don't think the coment is really necessary since you can use p:columns with columngroup

lprimak:
BTW rowKeyField is unnecessary for JPA. It can be derived from JPA directly.
See https://github.com/flowlogix/flowlogix/blob/573a80ff5a2df3faad0d73b8540809013d52dbd7/jakarta-ee/flowlogix-datamodel/src/main/java/com/flowlogix/jeedao/primefaces/internal/JPAModelImpl.java#L491

Rapster:
I'm no expert in using JPA API, all I see is it seems required to build the where clause? Are you saying there is another way to achieve this without the rowKeyField?

tandraschko:
will review again next week, will be in holiday tomorrow and need to get some bugs done before

PersistenceUnitUtil#getIdentifier is only there to receive the id of a loaded entity
as @Rapster already said, we also need to build a query with where clause AND rowKey can be different then id - also a bit unlikely

tandraschko:
here you go: https://github.com/primefaces/primefaces/issues/10838

tandraschko:
can you pls update the docs?
+1 then to merge

PR #10784, opened 2023-10-07
TreeTable: removal of cloneOnFilter property

Rapster:
My take on this, is that `TreeTable#cloneOnFilter` seems unecessary. As a reminder, cloneOnFilter:

> Defines if nodes should be cloned on filter via Cloneable interface or Copy-Constructor (CustomNode(CustomNode original) or CustomNode(String type, Object data, TreeNode parent)). Normally the filtered nodes are new instanceof of DefaultTreeNode.

How can we expect user to know that? A simpler description would be: whether or not custom nodes is being used. And we 
 don't need a property for that. 

Also, if cloning fails through Cloneable or Constructor: it fallbacks to built-in instance, meaning you'll start with a tree of C nodes and end-up having a tree of A nodes. So, sooner or later there will be problem... like #6173

What I changed is: if clonable fails, then fail. If fallback argument constructor: fails. No point to fallback to default/checkbox nodes

### DIFF
--- a/docs/14_0_0/components/treetable.md
+++ b/docs/14_0_0/components/treetable.md
@@ -23,7 +23,6 @@ id | null | String | Unique identifier of the component
 allowUnsorting | false | Boolean | Defines whether columns are allowed to be unsorted. Default is false.
 binding | null | Object | An el expression that maps to a server side UIComponent instance in a backing bean
 cellSeparator | null | String | Separator text to use in output mode of editable cells with multiple components.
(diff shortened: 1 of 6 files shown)


melloware:
I agree this is not something we need to ask the user. I really like this refactoring and clean up!

tandraschko:
+1 about the clone change; it was always weird, i just improved it a bit somewhere between the last years

not sure about making the treenodes generic, a parent node must actually not have the same datatype as the child.

would be better to split up your PR

Rapster:
> not sure about making the treenodes generic, a parent node must actually not have the same datatype as the child.

I wouldn't say "must" hopefully they can have the same type, but I see what you mean child nodes data don't necessarely have the same type as parent node. I'll create a separate PR for it

PR #10794, opened 2023-10-10
Fix #10721: Handle TAB press with selectone panel

### LINKED ISSUE #10721: SelectOneMenu: The focus in the element is lost when selecting with the mouse.
### Discussed in https://github.com/orgs/primefaces/discussions/128

<sup>Originally posted by **hgfrey** September 27, 2023</sup>
In the ShowCase for SelectOneMenu, focus is lost when an entry is selected with the mouse. In version PF 11 the focus is still retained. Is something broken or do I need to adjust something?</div>

melloware:
Fix #10721: Handle TAB press with selectone panel

@FlipWarthog reported if you pressed the TAB key with the Panel open you got stuck in a "focus loop".  According to PrimeVue docs it should select the item and tab to the next field.  This fix restores that functionality while still fixing the original issue


FlipWarthog (approved):

FlipWarthog:
@melloware @christophs78 I'm going to take a look at the IT for this and see where the gap is. The "focus loop" was also happening when you click to open, then click away. But I think this fix will fix both situations.

melloware:
Thanks @FlipWarthog

PR #10823, opened 2023-10-13
DataTable: support of columns in colgroup

Rapster:


melloware:
I think the most only reason we did this is a user did it and we added it so that we could make sure the exporter exported correctly?  But I will have to look through the commits. 

But either way I agree we shouldn't be showing users a bad way to do something.

Rapster:
I tink you're refering to this https://github.com/primefaces/primefaces/issues/9179 I'll give it a try

Rapster:
Exporters works fine with p:columns and colgroup

melloware:
HA yes that was the ticket i was referring to!

Issue #10835, opened 2023-10-17
Tree: introduce nodeKey / get rid of the auto-generated rowKey


tandraschko:
In case of #10544 and #5856 i came across it again, that some parts of the Tree are not well designed compared to other models and components.

In case of the DataTable, a user must asign a rowKey for a row to make it identifiable. 
In case of the Tree, the rowKey is autogenerated and contains the cordinates (e.g. 0_1_9). 
I think it makes sense to also introduce something like a "nodeKey" attribute, to get rid of the this weird auto-generating.
Also the auto-generation is currently placed in different places (UITree and models).

I ported the LazyTreeNode to PF, which required some adjustments.
I introduced the LazyDefaultTreeNode and a interface LazyTreeNode, which is required to make our rowKey generation algorithm work for not-loaded child nodes.
We can propably remove the LazyTreeNode in the future, when the nodeKey has been added.

<del>Currently PF has DefaultTreeNode and CheckboxTreeNode, which can propably be unified.</del>

melloware:
I support this as the Tree model stuff has always been confusing when I did through the code for Tree bugs!

tandraschko:
moved a small part of this ticket to #12816

PR #10836, opened 2023-10-17
#10544,#10834: Tree: pre-selection breaks selection

### LINKED ISSUE #10544: Tree: pre-selection breaks selection
### Describe the bug

I need to pre-select some items in tree component,
but if i remove the selection on a child, then removes all selections in that path.

### Reproducer

```
    <f:metadata>
        <f:viewAction action="#{indexController.init()}"/>
    </f:metadata>

    <ui:define name="content">
        <h:form id="form">
            <p:tree id="attriubteGroupTree"
                    value="#{indexController.root}" var="node"
                    selectionMode="checkbox"
                    selection="#{indexController.selectedNodes}">
                <p:ajax event="select" process="@this" update="@this"/>
                <p:ajax event="unselect" process="@this" update="@this"/>
                <p:treeNode>
                    <h:outputText value="#{node}"/>
                </p:treeNode>
            </p:tree>
        </h:form>
    </ui:define>
```

```
@Named
@ViewScoped
public class IndexController implements Serializable
{
    private TreeNode<String> root;
    private TreeNode<String>[] selectedNodes;

    public void init() {
        root = new CheckboxTreeNode<>("Files", null);
        TreeNode<String> node0 = new CheckboxTreeNode<>("Documents", root);
        TreeNode<String> node1 = new CheckboxTreeNode<>("Events", root);
        TreeNode<String> node2 = new CheckboxTreeNode<>("Movies", root);

        TreeNode<String> node00 = new CheckboxTreeNode<>("Work", node0);
        TreeNode<String> node01 = new CheckboxTreeNode<>("Home", node0);

        node00.getChildren().add(new CheckboxTreeNode<>("Expenses.doc"));
        node00.getChildren().add(new CheckboxTreeNode<>("Resume.doc"));
        node01.getChildren().add(new CheckboxTreeNode<>("Invoices.txt"));
        node01.setSelected(true);

        TreeNode<String> node10 = new CheckboxTreeNode<>("Meeting", node1);
        node10.setSelected(true);
        TreeNode<String> node11 = new CheckboxTreeNode<>("Product Launch", node1);
        node11.setSelected(true);
        TreeNode<String> node12 = new CheckboxTreeNode<>("Report Review", node1);
        node12.setSelected(true);

        TreeNode<String> node20 = new CheckboxTreeNode<>("Al Pacino", node2);
        TreeNode<String> node21 = new CheckboxTreeNode<>("Robert De Niro", node2);

        node20.getChildren().add(new CheckboxTreeNode<>("Scarface"));
        node20.getChildren().add(new CheckboxTreeNode<>("Serpico"));

        node21.getChildren().add(new CheckboxTreeNode<>("Goodfellas"));
        node21.getChildren().add(new CheckboxTreeNode<>("Untouchables"));
    }
}
```

initial state
image

Remove the selection of "Product Launch", removes all selections of "Events" and its children.
image

### Expected behavior

only removes the selection of "Product Launch" and no other one.

**Hint:** If i don't pre-select any item, then works all correctly.

### PrimeFaces edition

Community

### PrimeFaces version

14.0.0-SNAPSHOT

### Theme

_No response_

### JSF implementation

MyFaces

### JSF version

2.3

### Java version

17

### Browser(s)

_No response_

### LINKED ISSUE #10834: Selenium / Tree: make webElement lookup lazy
Currently webElement is stored in the TreeNode
therfore if you work on the tree with AJAX updates, ElementStateExceptions will occur

we can simply switch it to lazy lookup while remember the rowkey

tandraschko:
Fixes #10544
Fixes #10834

### DIFF
--- a/primefaces/src/main/java/org/primefaces/model/LazyDefaultTreeNode.java
+++ b/primefaces/src/main/java/org/primefaces/model/LazyDefaultTreeNode.java
@@ -1,0 +1,137 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2023 PrimeTek Informatics
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.model;
+
+import org.primefaces.util.SerializableFunction;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Lazy version of the {@link DefaultTreeNode}.
+ * A lazy version of the {@link CheckboxTreeNode} is still missing.
+ *
+ * @param <T> T the data type
+ */
+public class LazyDefaultTreeNode<T> extends DefaultTreeNode<T> implements LazyTreeNode {
+
+    private SerializableFunction<T, List<T>> loadFunction;
+    private SerializableFunction<T, Boolean> isLeafFunction;
+    private boolean loaded;
+
+    // serialization
+    public LazyDefaultTreeNode() {
+    }
+
+    public LazyDefaultTreeNode(T data,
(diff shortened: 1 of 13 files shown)


tandraschko:
#10835 as follow up for the future

primefaces/src/main/java/org/primefaces/model/LazyDefaultTreeNode.java
Rapster, line 1: rename into DefaultLazyTreeNode

tandraschko: because of DefaultLazyDataModel?

Thats a different case here; we added a "Lazy" specialization of "DefaultTreeNode", so its actually the bit better name

but as i said in the linked story, we need to refactor the whole models in general

Rapster: No, because it implements LazyTreeNode and you're making a default impl, therefore: DefaultLazyTreeNode (simple as that)

Agreed, tree components require a lot of attention

tandraschko: its just a different view on the topic
a general naming pattern is to prefix the subclass with the new added functionality, thats what is mostly used in all projects i work on
the implementation of TreeNode is the DefaultTreeNode and we enhanced this, the DefaultTreeNode, with LazyLoading
doesnt matter if you decide by hierarchie or less code, the lazy functionality is the lowest hierarchie currently with that model inheritance

tandraschko: and actually we currently need a LazyCheckboxTreeNode too, just dont have time for more
just here to fix some bugs and get things done before holiday ;)

Rapster: Up to you, I'm just following consistency naming (most names containing default with start with Default) and also because in this case, it's the default lazy tree node?

tandraschko: yeah, i understand your view but i prefer technical namings and we just need to a big refactoring to get rid of that whole mess

the best case would be to just have:
interface TreeNode
class DefaultTreeNode (which provides non-lazy and lazy and also that stuff vom CheckboxTreeNode)

and get rid of the
CheckboxTreeNode, LazyTreeNode, LazyDefaulTreeNode

just a to big task now

PR #10848, opened 2023-10-18
Fix #10822: Hazelcast move to docs

### LINKED ISSUE #10822: Core: Drop Hazelcast
For discussion I have a CaffeineCache provider which uses the awesome Caffeine library.  Now that we are on Java 11 we can use it for 14.0.0+ forward.

HazelCast we can keep but as you can see again today another CVE reported against HazelCast failing our builds.

I would love to swap the two and just we can provide the source code for the HazelCastCache as its not that complex?

melloware:
Fix #10822: Hazelcast move to docs


tandraschko:
+1

melloware:
@tandraschko should I do the same for EhCache?

tandraschko:
would leave it for now as used in showcase?

PR #10852, opened 2023-10-18
Fix #10831 - Wizard: Add disableOnAjax attribute for Next/Back navigation

### LINKED ISSUE #10831: Wizard: Add `disableOnAjax` attribute for Next/Back navigation
### Description

Similar to Ajax enabled buttons, add attribute to Wizard that will disable Next/Back navigation buttons while Ajax is processing.

Similar to `disableOnAjax` on, e.g., CommandButton that applies to the Next/Back navigation buttons of Wizard.

### Additional context

_No response_

FlipWarthog:
Fix #10831 - Wizard: Add disableOnAjax attribute for Next/Back navigation

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/wizard/wizard.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/wizard/wizard.js
@@ -70,7 +70,7 @@ PrimeFaces.widget.Wizard = PrimeFaces.widget.BaseWidget.extend({
             //events
             this.backNav.on("click", function() {_self.back();});
             this.nextNav.on("click", function() {_self.next();});
-            if (this.cfg.disableOnAjax) {
+            if (this.cfg.disableOnAjax !== false) {
                 this.bindTriggers();
             }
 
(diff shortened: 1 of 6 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/wizard/wizard.js
```diff
@@ -70,6 +70,9 @@ PrimeFaces.widget.Wizard = PrimeFaces.widget.BaseWidget.extend({
             //events
             this.backNav.on("click", function() {_self.back();});
             this.nextNav.on("click", function() {_self.next();});
+            if (this.cfg.disableOnAjax) {
+                this.bindTriggers();
```
melloware, line 74: I am pretty sure because you default to true and the renderer says true is default it won't send the value in JS. So this value will never be true. In other widgets you will see we do if undefined make the value true.  This way it saves network traffic.

melloware (changes requested):
See comments

primefaces/src/main/resources/META-INF/resources/primefaces/wizard/wizard.js
FlipWarthog: Oh is that why the code I stole this from did an explicit `!== false`?

melloware (approved):

melloware:
Assigned to @jepsar for a quick review

jepsar (approved):

PR #10853, opened 2023-10-19
Fix #10847: AutoComplete closes on Enter when no item is highlighted

### LINKED ISSUE #10847: AutoComplete: DefaultCommand not triggered when Popup is opened
### Describe the bug

We have a use-case, where users can search our platform for a specific string. For that we use the autoComplete component with "autoHighlight=false". You can either enter any string you want and press the search button next to the autoComplete, or you select one of the suggestions, where a redirect to a specific resource gets triggered.

From the user perspective you would expect to perform the action of the button next to the autoComplete if you press Enter after entering your search term. For that we use a defaultCommand which triggers the action of the button. 

However, at the moment this doesn't seem to work as soon as the popup with the suggestions opens. If you close it with Escape and then hit Enter again, everything works fine.

### Reproducer

[reproducer.zip](https://github.com/primefaces/primefaces/files/12999738/reproducer.zip)

1. enter any string into the autoComplete
2. wait for the suggestions to show up
3. hit Enter -> nothing happens
4. hit Escape
5. hit Enter again -> button action gets performed

### Expected behavior

DefaultCommand should also work when the suggestion popup is opened.

### PrimeFaces edition

Community

### PrimeFaces version

13.0.2

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

11

### Browser(s)

_No response_

fcjm:
Fix #10847.

Changes the behaviour of the AutoComplete on pressing Enter when no item is highlighted as proposed by WAI.

https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/#kbd_label

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
@@ -512,6 +512,8 @@ PrimeFaces.widget.AutoComplete = PrimeFaces.widget.BaseWidget.extend({
                             $this.preventInputChangeEvent = true;
                             highlightedItem.trigger("click");
                             $this.itemSelectedWithEnter = true;
+                        } else {
+                            $this.hide();
                         }
 
                         e.preventDefault();


fcjm:
@melloware Could you please take a look at the failed check? I am not sure whether this was caused by my commit or not.

melloware (approved):

PR #10872, opened 2023-10-22
Fix #10866: Converter return empty string

### LINKED ISSUE #10866: SelectOneMenu: Float Label does not work with selecItem value=null
### Describe the bug

 Float Label does not work with p:selectOneMenu when there is a converter.
Even if there is no value selected, the ui-inputwrapper-filled class is always present.

### Reproducer

```html
<div class="field">
<span class="ui-float-label">
	<p:selectOneMenu
		converter="#{produtoConverter}"
	>
		<f:selectItem   />
	</p:selectOneMenu>
	<p:outputLabel
		for="@previous"
		value="AAAAA"
	>
	</p:outputLabel>
</span>
</div>
```

### Expected behavior

The floating label should appear inside the selectOneMenu when there is nothing selected.
The  ui-inputwrapper-filled class shoud be set only when a value is selected.

### PrimeFaces edition

Community

### PrimeFaces version

13.0.2

### Theme

default

### JSF implementation

Mojarra

### JSF version

JavaServer(TM) Faces 2.2 API

### Java version

8

### Browser(s)

_No response_

melloware:
Fix #10866: Converter return empty string

### DIFF
--- a/primefaces-showcase/src/main/java/org/primefaces/showcase/convert/CountryConverter.java
+++ b/primefaces-showcase/src/main/java/org/primefaces/showcase/convert/CountryConverter.java
@@ -63,7 +63,7 @@ public class CountryConverter implements Converter<Country> {
             return String.valueOf(value.getId());
         }
         else {
-            return null;
+            return "";
         }
     }
 }
--- a/primefaces-showcase/src/main/java/org/primefaces/showcase/convert/LocaleConverter.java
+++ b/primefaces-showcase/src/main/java/org/primefaces/showcase/convert/LocaleConverter.java
@@ -65,7 +65,7 @@ public class LocaleConverter implements Converter<Country> {
             return String.valueOf(value.getId());
         }
         else {
-            return null;
+            return "";
         }
     }
 }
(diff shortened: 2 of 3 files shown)


tandraschko:
We should check this, i also always use null and never had problems

melloware:
It is in the JSF Javadoc: https://javadoc.io/static/jakarta.faces/jakarta.faces-api/2.3.1/javax/faces/convert/Converter.html

> Returns:
a zero-length String if value is null, otherwise the result of the conversion

FlipWarthog:
Adding some context, the original issue has to do with how the elements are being generated based on the return value from the converter here:
https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/java/org/primefaces/component/selectonemenu/SelectOneMenuRenderer.java#L600

Which generates (null vs. ""):
```html
<option selected="selected" data-escape="true">&nbsp;</option>
vs.
<option value="" selected="selected" data-escape="true">&nbsp;</option>
```

So in core.utils.js, when it checks if the input is empty, that missing value attribute is causing it to get the actual content of the element:
https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js#L929

`input.val()`, if `value` is not set on the `option`, is returning the nbsp so it's seeing it as non-empty and adding the class.

tandraschko:
Yeah
I think we can just Support both here

primefaces-showcase/src/main/java/org/primefaces/showcase/convert/CountryConverter.java
```diff
@@ -63,7 +63,7 @@ public String getAsString(FacesContext context, UIComponent component, Country v
             return String.valueOf(value.getId());
         }
         else {
-            return null;
+            return "";
         }
```
Rapster, line 67: Constants.EMPTY_STRING maybe?

PR #10885, opened 2023-10-24
#10866

### LINKED ISSUE #10866: SelectOneMenu: Float Label does not work with selecItem value=null

tandraschko:
Fixes #10866

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/chips/chips.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/chips/chips.js
@@ -120,7 +120,7 @@ PrimeFaces.widget.Chips = PrimeFaces.widget.BaseWidget.extend({
      * @private
      */
     updateFloatLabel: function() {
-        PrimeFaces.utils.updateFloatLabel(this.jq, this.input, this.hasFloatLabel);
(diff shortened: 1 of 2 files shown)


melloware (approved):
Do you want this in 13.0.3 or should we leave it in 14.0

tandraschko:
feel free to port it to 13.x, but please retest then :)

melloware:
OK will test today while I am fixing the focus issue.

PR #10892, opened 2023-10-24
#10838

### LINKED ISSUE #10838: JpaLazyDataModel: automatically try to resolve @Id as rowKey
When `@Id` is a primitive, we can easily resuse it as default rowKeyField.
This doesnt work for clustered primary keys or embedded ids but many cases will just work.
We need a good exception hanlding if it's not a primitive.

for `getRowData(String rowKey)` we could maybe use JPA metamodel:
https://docs.oracle.com/javaee%2F7%2Fapi%2F%2F/javax/persistence/EntityManagerFactory.html#getMetamodel--
to resolve the field itself, otherwise use reflection to scan `@Id`

for `getRowKey(Object entity)` we can use PersistenceUnitUtil#getIdentifier

tandraschko:
Fixes #10838

i also contains a small change to a static logger for performance reasons if you check Logger.java:
```
public static Logger getLogger(String name) {
      return Logger.getLogger(name, Reflection.getCallerClass());
}
```

### DIFF
--- a/docs/14_0_0/components/datatable.md
+++ b/docs/14_0_0/components/datatable.md
@@ -1006,67 +1006,59 @@ JPALazyDataModel<MyEntity> lazyDataModel = JPALazyDataModel.<MyEntity>builder())
         .build();
 ```
 
-If you have selection enabled, you can either pass the rowKey via JPA metamodel:
+If you have selection enabled, we need a `rowKey`. 
+It's very likely that this is the mapped JPA `@Id`, so we try to auto-lookup per default.
+Otherwise you can also define it:
 
 ```java
 JPALazyDataModel<MyEntity> lazyDataModel = JPALazyDataModel.<MyEntity>builder())
-        .entityClass(MyEntity.class)
-        .entityManager(() -> entityManager)
-        .rowKeyField(MyEntity_.id)
-        .build();
-```
-
-or as plain String:
-
-```java
-JPALazyDataModel<MyEntity> lazyDataModel = JPALazyDataModel.<MyEntity>builder())
-        .entityClass(MyEntity.class)
-        .entityManager(() -> entityManager)
-        .rowKeyField("id")
-        .build();
+        ...
+        .rowKeyField(MyEntity_.id) // JPA MetaModel
+        .rowKeyField("id") // or as String
+        ...
 ```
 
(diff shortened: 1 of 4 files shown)


melloware (approved):

docs/14_0_0/components/datatable.md
```diff
@@ -1006,67 +1006,59 @@ JPALazyDataModel<MyEntity> lazyDataModel = JPALazyDataModel.<MyEntity>builder())
...
 
 ```java
 JPALazyDataModel<MyEntity> lazyDataModel = JPALazyDataModel.<MyEntity>builder())
-        .entityClass(MyEntity.class)
-        .entityManager(() -> entityManager)
+        ...
         .wildcardSupport(true)
-        .build();
+        ...
 ```
```
Rapster, line 1051: Not a fan of all these paragraphs, cumbersome and not so great to read... I think it'd be better (more readable) if you show every options within a table. Also, I don't think you need to show every use case with code associated. Considering a builder is being used, it speaks by itself

melloware: @Rapster i think in this case people need examples and show exactly what the features will do else we will get questions.

Rapster: Just to be clear, I meant a markdown table. I would simply describe that, there is a builder to instanciate a datamodel, give a very basic example, and then describe all the features users can enable contained in a md table (just my 2 cents tbh, builders are supposed to be self documented)

tandraschko:
if you feel that we should rework the docs here, please lets do it in a own issue/PR. it just spend some minutes to strip it.

@Rapster you introduced rowKeyProvider and rowKeyField in the builder. i dont think a user should use it and also dont need it. its just some additional, which can be automatically detected. Can i remove it?

tandraschko:
i dont like it that much actually because it just confuses users
auto lookup doesnt work -> means that the reflection access doesnt work, which means that much more is broken

but i will reaadd as it doesnt hurt

Rapster:
Don't see how it will confuse users, like you said, it requires just some javadoc and/or doc. I mentionned in case it doesn't work, but it can also be for other reasons, it's not unsual for users to come up with their own logic

PR #10871, opened 2023-10-21
Fix #10870 - ChartJS: new approach

Rapster:
Fix #10870

Issue #10870, opened 2023-10-21
ChartJS: new approach


Rapster:
### Description

Based on discussions #8819 and #10849

I'll try to sum up the big lines:
- Have one and only component: `<p:chart />`. There is no need to have multiple components, at worse `<p:chart />` should redirect to the proper renderer depending the type: lucky for us chartJS does the job for us.
- Model abstraction: As of today, PF is responsible to implement chartJS API into Java. First, chartJS API is freaking big! Second, It's hard to follow chartjs model changes from one version to another, which makes us left behind compare to the version we are already using: this is frustrating for users.
- KISS:  `<p:chart />` could be a very simple JSF wrapper of chartJS, take a String as input, and handle a few ajax events out of the box. As it takes a String, users are free to either implement their own API and serialize it into JSON, or use existing solutions like https://github.com/xdev-software/chartjs-java-model (which is already 4.4 compatible and java 11 compatible since 1.0.2) 
- Deprecate current impl for 14 

### Describe the solution you would like

**XHTML**

```xml
<div class="card">
    <h5>Plain JSON</h5>
    <p:chart value="#{chartJsView.barModel}" style="width: 100%; height: 500px;">
        <p:ajax event="itemSelect" listener="#{chartJsView.itemSelect}" update="growl"/>
    </p:chart>
</div>
<div class="card">
    <h5>XDEV ChartJS model</h5>
    <p:chart value="#{chartJsView.barModel2}" style="width: 100%; height: 500px;">
        <p:ajax event="itemSelect" listener="#{chartJsView.itemSelect}" update="growl"/>
    </p:chart>
</div>
```

**JAVA**
```java
    public void createBarModel() {
        barModel = "{" +
            "    \"type\": \"bar\"," +
            "    \"data\": {" +
            "        \"datasets\": [" +
            "            {" +
            "                \"type\": \"bar\"," +
            "                \"data\": [" +
            "                    65," +
            "                    59," +
            "                    80," +
            "                    81," +
            "                    56," +
            "                    55," +
            "                    40" +
            "                ]," +
            "                \"label\": \"My First Dataset\"," +
            "                \"hidden\": false," +
            "                \"backgroundColor\": \"rgba(255, 99, 132, 0.2)\"," +
            "                \"borderColor\": \"rgb(255, 99, 132)\"," +
            "                \"borderWidth\": 1" +
            "            }," +
            "            {" +
            "                \"type\": \"bar\"," +
            "                \"data\": [" +
[...]

PR #10903, opened 2023-10-25
ChartJS Refactor

### LINKED ISSUE #5665: ChartsJS: TimePoint value type
## 1) Environment
- PrimeFaces version: 8.0

## 2) Expected behavior
e.g.
x = 2020-01-01-12-00-00
y = 123.45

## 3) Actual behavior

### LINKED ISSUE #10849: Charts: unified and consistent data interface classes
### Description

As a developer of charts that often stack multiple elements (bar, line) to show the relationship between 2 or more data sets where my number of data points is not consistently equal; I want a consistent way to feed data to the Cartesian chart types that preserve my x / y tuples so that I do not need to pre parse all my data points to get a distinct list of X axis labels, then build each data set ensuring that I provide a value for each X value I found across all datasets, but instead let chart.js figure it out for me, instead of needing to write a potentially error prone hack to massage all the data points to fit the primefaces wrapper when the underlying chart.js can handle the tuples.

For each of the ChartDataSet implementations, have the data field accept a List<Object> instead of List<Number> (or a list of an abstraction that makes sense to collect x / y tuples)  like the LineChartDataSet currently does so that if a developer wants to pass in a list of NumericPoint, or a class of their own that has a toString implementation that produces JSON that will output correctly when processed by ChartUtils.writeDataValue() the same way BubblePoint and NumericPoint do.

### Additional context

The specific classes that would need to be modified to need a minimally viable fix are:  BarChartDataSet, PolarAreaChartDataSet, and RadarChartDataSet.

asdef:
Defined a new Abstract Chart point and translation to json in the ChartUtils that has 3 sample implementations for Numeric x values, String x values and Date x values.  Updated the cartesian chart types to accept a list of type Object instead of Number to allow for use of the new Point implementations.  WIP implementation in the showcase Mixed Chart.

Fix #10849 

**Remaining Tasks:**
- [ ] format the x axis dates
- [ ] integration test fixes
- [ ] additional doc?

### DIFF
--- a/primefaces-showcase/src/main/java/org/primefaces/showcase/view/chartjs/ChartJsView.java
+++ b/primefaces-showcase/src/main/java/org/primefaces/showcase/view/chartjs/ChartJsView.java
@@ -23,20 +23,10 @@
  */
 package org.primefaces.showcase.view.chartjs;
 
-import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-
-import javax.annotation.PostConstruct;
-import javax.enterprise.context.RequestScoped;
-import javax.faces.application.FacesMessage;
-import javax.faces.context.FacesContext;
-import javax.inject.Named;
-
 import org.primefaces.event.ItemSelectEvent;
 import org.primefaces.model.charts.ChartData;
 import org.primefaces.model.charts.axes.cartesian.CartesianScales;
+import org.primefaces.model.charts.axes.cartesian.CartesianTime;
 import org.primefaces.model.charts.axes.cartesian.linear.CartesianLinearAxes;
 import org.primefaces.model.charts.axes.cartesian.linear.CartesianLinearTicks;
 import org.primefaces.model.charts.axes.radial.RadialScales;
@@ -49,6 +39,7 @@ import org.primefaces.model.charts.bar.BarChartOptions;
 import org.primefaces.model.charts.bubble.BubbleChartDataSet;
 import org.primefaces.model.charts.bubble.BubbleChartModel;
 import org.primefaces.model.charts.data.BubblePoint;
+import org.primefaces.model.charts.data.DatePoint;
 import org.primefaces.model.charts.data.NumericPoint;
 import org.primefaces.model.charts.donut.DonutChartDataSet;
 import org.primefaces.model.charts.donut.DonutChartModel;
@@ -73,6 +64,14 @@ import org.primefaces.model.charts.radar.RadarChartDataSet;
 import org.primefaces.model.charts.radar.RadarChartModel;
 import org.primefaces.model.charts.radar.RadarChartOptions;
 import org.primefaces.model.charts.scatter.ScatterChartModel;
+
+import javax.annotation.PostConstruct;
+import javax.enterprise.context.RequestScoped;
+import javax.faces.application.FacesMessage;
+import javax.faces.context.FacesContext;
+import javax.inject.Named;
+import java.io.Serializable;
+import java.util.*;
 
 @Named
 @RequestScoped
@@ -161,7 +160,7 @@ public class ChartJsView implements Serializable {
         ChartData data = new ChartData();
 
         PolarAreaChartDataSet dataSet = new PolarAreaChartDataSet();
-        List<Number> values = new ArrayList<>();
+        List<Object> values = new ArrayList<>();
         values.add(11);
         values.add(16);
         values.add(7);
@@ -348,7 +347,7 @@ public class ChartJsView implements Serializable {
         BarChartDataSet barDataSet = new BarChartDataSet();
         barDataSet.setLabel("My First Dataset");
 
-        List<Number> values = new ArrayList<>();
+        List<Object> values = new ArrayList<>();
         values.add(65);
         values.add(59);
         values.add(80);
@@ -436,7 +435,7 @@ public class ChartJsView implements Serializable {
         barDataSet.setBackgroundColor("rgba(255, 99, 132, 0.2)");
         barDataSet.setBorderColor("rgb(255, 99, 132)");
         barDataSet.setBorderWidth(1);
-        List<Number> values = new ArrayList<>();
+        List<Object> values = new ArrayList<>();
         values.add(65);
         values.add(59);
         values.add(80);
@@ -451,7 +450,7 @@ public class ChartJsView implements Serializable {
         barDataSet2.setBackgroundColor("rgba(255, 159, 64, 0.2)");
         barDataSet2.setBorderColor("rgb(255, 159, 64)");
         barDataSet2.setBorderWidth(1);
-        List<Number> values2 = new ArrayList<>();
+        List<Object> values2 = new ArrayList<>();
         values2.add(85);
         values2.add(69);
         values2.add(20);
@@ -502,7 +501,7 @@ public class ChartJsView implements Serializable {
         HorizontalBarChartDataSet hbarDataSet = new HorizontalBarChartDataSet();
         hbarDataSet.setLabel("My First Dataset");
 
-        List<Number> values = new ArrayList<>();
+        List<Object> values = new ArrayList<>();
         values.add(65);
         values.add(59);
         values.add(80);
@@ -572,7 +571,7 @@ public class ChartJsView implements Serializable {
         BarChartDataSet barDataSet = new BarChartDataSet();
         barDataSet.setLabel("Dataset 1");
         barDataSet.setBackgroundColor("rgb(255, 99, 132)");
-        List<Number> dataVal = new ArrayList<>();
+        List<Object> dataVal = new ArrayList<>();
         dataVal.add(62);
         dataVal.add(-58);
         dataVal.add(-49);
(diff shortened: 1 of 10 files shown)


asdef:
WIP fix for https://github.com/primefaces/primefaces/issues/10849

melloware:
OK can you make a checklist of what issues you have remaining?

asdef:
remaining issue list:

- set the format of the x axis date values using a parser library (chartjs-adapter-date-fns) - in progress trying to get everything just right so the resource handler will load the js library for me.
- integration test fixes
- additional doc?

To the documentation question, I've updated the javadoc for the data lists  to describe how to use the Point tuple class, and I'm working on an example implementation in the showcase.  What additional places need doc updates?

asdef:
@jepsar thanks for the pointer.  I'll get that updated with my next push.

melloware:
I updated the checklist in at the top of the ticket.

melloware:
> remaining issue list:
> 
> * set the format of the x axis date values using a parser library (chartjs-adapter-date-fns) - in progress trying to get everything just right so the resource handler will load the js library for me.
> * integration test fixes
> * additional doc?
> 
> To the documentation question, I've updated the javadoc for the data lists to describe how to use the Point tuple class, and I'm working on an example implementation in the showcase. What additional places need doc updates?

That should be good enough for docs.  Most people use the Showcase examples.

asdef:
I need the date adapter so I can set a time > displayformat on the x axis scale like this:

x: {
      type: "time",
      time: {
        displayFormats: {
          millisecond: "ss.SSS",
          second: "mm:ss",
          minute: "HH:mm",
          hour: "HH:mm",
          day: "MM-dd",
          week: "YYYY",
          month: "MMM-yyyy",
          quarter: "MMM-yyyy",
          year: "yyyy"
        },
        unit: "month"
      },
      adapters: {
        date: {
          locale: en
        }
      }
    }

asdef:
working on extending the charts definition in [PrimeFaces.d.ts](https://github.com/primefaces/primefaces/blob/13.X/docs/PrimeFaces.d.ts) to import chart.js and chartjs-adapter-date-fns and I'm wondering why the PrimeFaces.widget namespace is declared multiple times, each time wrapping one class instead of declaring it once and exporting all the classes in that one declaration?  I think there's probably a really good architectural reason for it, but I'm not understanding it.  Could anyone enlighten me as to why the file is structured the way it is?

blutorange:
@asdef Not so much an architectural reason, but more a technical reason: `PrimeFaces.d.ts` is an **auto-generated file**. PrimeFaces currently uses plain vanilla JavaScript, which does not make it easy to extract type declarations. We tried some tools that can do that automatically, but they did not work as we wanted. That's why we currently have [a custom set of scripts](https://github.com/primefaces/primefaces/tree/13.X/primefaces/src/main/type-definitions) that parses all JavaScript files and JSDoc comments and extracts the type declarations from that (see link for some documentation). And it was much easier to just generate multiple namespace declarations for each widget, rather than to add additional logic to merge these namespaces; especially since TypeScript doesn't care whether it's declared once or multiple times.

That said, if you want to adjust the the type declarations:

* see [chartjs/1-widgets.js](https://github.com/primefaces/primefaces/blob/13.X/primefaces/src/main/resources/META-INF/resources/primefaces/chartjs/1-widgets.js) for the type declarations of the PrimeFaces widget itself
* see [type-definitions/core.d.ts](https://github.com/primefaces/primefaces/blob/13.X/primefaces/src/main/type-definitions/core.d.ts) for some additional custom type declarations that are included in the main file. But keep in mind that this must be a [global script file and not a module](https://www.typescriptlang.org/docs/handbook/2/modules.html#how-javascript-modules-are-defined), i.e. you can't use any `import`s or `export`s.
* you can add a *.d.ts module file with imports next to the widget file like [mindmap/mindmap.d.ts](https://github.com/primefaces/primefaces/blob/13.X/primefaces/src/main/resources/META-INF/resources/primefaces/mindmap/mindmap.d.ts), which gets picked up automatically. The script [automatically recognizes it as a module file](https://github.com/primefaces/primefaces/blob/13.X/primefaces/src/main/type-definitions/src/ts-classify-declaration-file.js#L22) and outputs it to the companion [docs/PrimeFaces-module.d.ts](https://github.com/primefaces/primefaces/blob/13.X/docs/PrimeFaces-module.d.ts) file.

It's not ideal, but that's what we have for now. Feel free to ask if you have any questions or if you have some use case that doesn't work currently with the type declarations.

On a related note, I've been thinking about whether we can convert the JavaScript to TypeScript, but that's another story. It would, among other things, have the benefit that we would not need the custom extractor script anymore and could just use standard tools for that.

asdef:
@blutorange Thanks again for all the documentation.  I've been reading up on the different module systems to try and fit what I'm seeing in primefaces, and while it makes a little more sense to me, I still don't see the path forward to do what chartJs says to do at the bottom of their [getting started integrations page](https://www.chartjs.org/docs/latest/getting-started/integration.html) with nested require statements.  I did try unsuccessfully to use nested require statements and to add common-js to get access to the asynchronous require functionality.  Can you recommend an approach for importing both chartjs and a datetime adapter for chartjs (moment, chartjs-adapter-date-fns or one of the others from [here](https://github.com/chartjs/awesome#adapters)) nested in a similar way to what is occuring with the require example?  This is what I get for not keeping up with the javascript world and living largely in Java 8 :-( .

blutorange:
I didn't follow the entire thread so I'm probably not understanding the goal here, but: what is wrong with how chart.js is currently loaded? PrimeFaces doesn't really use any module systems, it has a bunch of global script files (e.g. [galleria](https://github.com/primefaces/primefaces/tree/master/primefaces/src/main/resources/META-INF/resources/primefaces/galleria)) that mostly expose functions to the global scope; and are then [minified and possibly joined together](https://github.com/primefaces/primefaces/blob/master/primefaces/pom.xml#L569)

From their integrations page, that would be option 1: a script tag. Scripts tags aren't loaded asynchronously (without an `async` / `defer` flag); and code joined together is also executed sequentially, so you can ignore that note regarding asynchronous loading. `require(...)` is only asynchronous because it needs to fetch the files from the network.

For example

```html
<html>
<body>
<script src="https://unpkg.com/chart.js@4.4.0/dist/chart.umd.js"></script>
<script src="https://raw.githubusercontent.com/primefaces/primefaces/master/primefaces/src/main/resources/META-INF/resources/primefaces/moment/moment.js"><script>
<script src="https://unpkg.com/chartjs-adapter-moment@1.0.1/dist/chartjs-adapter-moment.js"><script>
</body>
</html>
```

should work just fine

* `moment.js` doesn't depend on anything and it shouldn't matter whether it's loaded before or after chart.js. The script, when run, defines a global `moment` variable.
* `chart.js` defines a global `Chart` variable.
* `chartjs-adapter-moment`, as you can see when you take a look at the file, accesses `global.moment`, so the moment script must have been executed beforehand. Similarly, it also accesses `global.Chart`, so the chart.js script must have run beforehand.

And of course, you don't manually add `<script>` tags, that's taken care of by Jakarta Faces via the `@ResourceDependency` on the component:

> https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/java/org/primefaces/component/barchart/BarChart.java#L37C45-L41

As you can see, it already declares `moment` before `chart.js`. So you only need to add `chartjs-adapter-moment.js` after `chart.js` and it should work.

melloware:
Great explanation @blutorange

asdef:
@blutorange thanks for the explanation, since its all global scoped, I'm going to try a pure HTML implementation to shake out any bugs in the config I'm generating for chartjs as the reason it's not working yet is probably there and not with the inclusion of the adapter.  When I know more about what I screwed up, I'll post again...

Rapster:
Not sure how far you want to go with this since the current component is very likely to be deprecated... My advise would be to focus on the new one and see what could be improved

asdef:
@Rapster what parts are being replaced for chart functionality in Primefaces?

melloware:
It is this PR that he is proposing: https://github.com/primefaces/primefaces/pull/10871

Rapster:
The whole component, meaning also the model behind it, see #10870

asdef:
@Rapster I see how the simplification allows me to pass in the tuples, which was half of what I need for my primefaces application, the other half is time series axes.  Given that we already have moment, available I think all we're missing is the chartjs-moment-adapter.  Since you already moved up to the latest version of chartjs in your PR, it looks like you're going to get there faster than I am, so how can I help add the stuff I need out of a chartjs refactor?

Rapster:
What exactly do you need for the time series? I'm no expert in chartjs so you might have to enlight me on a lot of things 😅

melloware:
@asdef make sure the files were like this..

```
0-chartjs.min.js
1-chartjs-adapter-moment.js
2-widgets.js
```

our bundler will make that `chart.js` and does it in the order they are alphanumerically.

Rapster:
If adding chartjs-adapter-moment.js is your issue, that should be easy, now I wonder if it should be added conditionnaly if time series is used...

Rapster:
I added chartjs-adapter-moment.

@melloware this file contains a header. Are we supposed to remove header for perf reasons? In the unified file, coments are still there

melloware:
Yep so if the header is a comment with a bang `!` the optimizer will not remove those as they are intentionally marked as don't strip me out.  I have seen a lot of libs use it for license headers etc.

Rapster:
Ok so I don't touch it right?

melloware:
Yeah i would leave it.

asdef:
closing in favor of [this solution](https://github.com/primefaces/primefaces/pull/10871)

PR #10908, opened 2023-10-25
PanelGrid remove deprecation warning for table layout

melloware:
In our 70+ apps we use this everywhere as this used to be the only way to do it.  We already made responsive the default but I want to remove this warning as I don't think we should remove this from PanelGrid anytime soon.  All the others I am still OK with removing.


stolp:
Yes please. It was the biggest issue on my list of looming future deprecation changes.

PR #10912, opened 2023-10-25
Fix #10345: Badge add icon support

### LINKED ISSUE #10345: Badge: add icon support
### Description

I am trying to use the badge to display a plus icon instead of a number. That is no problem since + is a normal character but it would look better if i could use the prime or font awesome plus icon which is larger and bold and would be more flexible in other use cases.

grafik

Similar to button _icon_ attribute could be added alongside text _value_. Maybe enhance it with _iconPos_ aswell.

### Additional context

_No response_

melloware:
Fix #10345: Badge add icon support

image

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/badge/BadgeRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/badge/BadgeRenderer.java
@@ -89,12 +89,15 @@ public class BadgeRenderer extends CoreRenderer {
         String value = model.getValue();
         boolean valueEmpty = LangUtils.isEmpty(value);
         String severity = model.getSeverity();
+        String icon = model.getIcon();
+        boolean iconEmpty = LangUtils.isEmpty(icon);
+        String iconPos = model.getIconPos();
         String size = model.getSize();
         String styleClass = getStyleClassBuilder(context)
                     .add(Badge.STYLE_CLASS)
                     .add(model.getStyleClass())
                     .add(!valueEmpty && value.length() == 1, Badge.NO_GUTTER_CLASS)
-                    .add(valueEmpty, Badge.DOT_CLASS)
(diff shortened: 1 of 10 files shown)


primefaces/src/main/java/org/primefaces/component/badge/BadgeRenderer.java
```diff
@@ -128,6 +145,24 @@ protected void encode(FacesContext context, Badge badge, BadgeModel badgeModel,
         }
     }
 
+    protected void encodeIcon(FacesContext context, Badge badge, String icon) throws IOException {
+        ResponseWriter writer = context.getResponseWriter();
+        String iconStyleClass = getStyleClassBuilder(context)
+                .add(Badge.ICON_CLASS)
+                .add(icon)
+                .build();
+        writer.startElement("i", null);
```
jepsar: Accessibility-wise you should use semantic tags. `i` is deprecated and `em` should be used instead. However, that's not semantic in this case. Something like `span` should be used instead. https://www.w3.org/TR/WCAG20-TECHS/G115.html

primefaces/src/main/java/org/primefaces/component/badge/BadgeRenderer.java
```diff
@@ -117,8 +120,22 @@ protected void encode(FacesContext context, Badge badge, BadgeModel badgeModel,
             writer.writeAttribute("style", model.getStyle(), "style");
         }
 
-        if (!valueEmpty && model.isVisible()) {
-            writer.writeText(value, "value");
+        if (model.isVisible()) {
```
jepsar: I would have delegated this to a new method, for better readability and reduced complexity (nesting).

jepsar (commented):
A few suggestions.

melloware:
@jepsar i made all your suggested changes!

jepsar (approved):

PR #10919, opened 2023-10-26
Added community page

### LINKED ISSUE #10691: Docs: add reference to FlowLogix JPA LazyDataModel
### Description

Hi,

I do believe that https://docs.flowlogix.com/#section-jpa-lazymodel is the best and simplest implementation of JPA LazyDataModel. I request it be added to the docs and showcase. Thank you

Please add https://docs.flowlogix.com/#section-jpa-lazymodel to docs and showcase

### Additional context

Many ecosystem projects such as PrimeFaces support their community by listing projects that use and enhance the main product.
PrimeFaces can do the same.

lprimak:
Added community page and some community projects
Continuation of 10828

fixes #10691

### DIFF
--- a/docs/14_0_0/components/datatable.md
+++ b/docs/14_0_0/components/datatable.md
@@ -1060,6 +1060,20 @@ JPALazyDataModel<MyEntity> lazyDataModel = JPALazyDataModel.<MyEntity>builder())
         })
         ...
 ```
+### FlowLogix JPALazyDataModel (PrimeFaces Community)
+`JPALazyDataModel` implementation that's fully integrated with Jakarta EE and `@Inject`able.
+
+```java
+@Named
+@ViewScoped
+public class UserViewer implements Serializable {
+    @Inject
+    // optional configuration annotation
+    @LazyModelConfig(caseInsensitive = true)
+    JPALazyDataModel<UserEntity> lazyModel;
+}
+```
+Developer's guide is available at https://docs.flowlogix.com/#section-jpa-lazymodel. Project page is available at https://github.com/flowlogix/flowlogix
 
 ### DefaultLazyDataModel
 
--- a/docs/14_0_0/gettingstarted/community.md
+++ b/docs/14_0_0/gettingstarted/community.md
@@ -1,0 +1,12 @@
+# PrimeFaces Community
+
+PrimeFaces has a vibrant community that enhances its functionality. The list below is by no means comprehensive.
+
+## Community Projects
+
+* [PrimeFaces Extensions](https://primefaces-extensions.github.io) includes many more components such as `GChart` and `ImageZoom`
+* [Arquillian Primefaces](https://github.com/omnifaces/arquillian-primefaces) easier integration testing of PrimeFaces components
+* [FlowLogix JPALazyDataModel](https://docs.flowlogix.com/#section-jpa-lazymodel)
+  fully integrates with Jakarta EE and `@Inject`able
+* [OptimusFaces](https://github.com/omnifaces/optimusfaces) service-based implementation of `JPALazyDataModel`
+* [QuarkusFaces](https://github.com/melloware/quarkus-faces) run PrimeFaces Showcase on Quarkus
--- a/docs/14_0_0/menu.md
+++ b/docs/14_0_0/menu.md
@@ -7,6 +7,7 @@
   - [Configuration](/gettingstarted/configuration.md)
   - [Hello World](/gettingstarted/helloworld.md)
   - [Whats New](/gettingstarted/whatsnew.md)
+  - [Community Projects](/gettingstarted/community.md)
   - [Migration Guide](../migrationguide/migrationguide.md)
 
 - Core


docs/14_0_0/gettingstarted/community.md
```diff
@@ -0,0 +1,11 @@
...
+
+PrimeFaces has a vibrant community that enhances its functionality. The list below is by no means comprehensive.
+
+## Community Projects
+
+* [PrimeFaces Extensions](https://primefaces-extensions.github.io) includes many more components such as `GChart` and `ImageZoom`
+* [Arquillian Primefaces](https://github.com/omnifaces/arquillian-primefaces) easier integration testing of PrimeFaces components
+* [FlowLogix JPALazyDataModel](https://docs.flowlogix.com/#section-jpa-lazymodel)
+  fully integrates with Jakarta EE and `@Inject`able
+* [OptimusFaces](https://github.com/omnifaces/optimusfaces) Service-based implementation of `JPALazyDataModel`
```
melloware: Can you add `Quarkus Faces` here: https://github.com/melloware/quarkus-faces

melloware (changes requested):

docs/14_0_0/gettingstarted/community.md
lprimak: done

melloware (approved):
I will let Thomas review and see if he wants you to move anything.

docs/14_0_0/components/datatable.md
```diff
@@ -1060,6 +1060,20 @@ JPALazyDataModel<MyEntity> lazyDataModel = JPALazyDataModel.<MyEntity>builder())
         })
         ...
 ```
+### FlowLogix JPALazyDataModel (PrimeFaces Community)
+`JPALazyDataModel` implementation that's fully integrated with Jakarta EE and `@Inject`able. It does not require optional configuration lambdas to be serializable, helping with usability.
```
tandraschko: > It does not require optional configuration lambdas to be serializable, helping with usability.

i would just remove that. Otherwise please explain exactly why your lambdas dont need to be serializable and where they are stored.
Even if they are not serialized, not sure how this improves usability.

everything else looks ok for me.

lprimak: Im not sure that explanation in this blurb is quite necessary there. (it's quite lengthy)
Basics about how to deal with final transient variables are described in this article: https://medium.com/@lprimak/how-to-deal-with-non-serializable-fields-in-java-correctly-4ffecad98f15

As you can see here, https://github.com/flowlogix/flowlogix/blob/a915f3343ce9b5e6afac8261cd15d2cbe289d75a/jakarta-ee/flowlogix-datamodel/src/main/java/com/flowlogix/jeedao/primefaces/internal/JPAModelImpl.java#L102

All configuration lambdas are marked as transient and final, thus don't need to be serializable. The only two lambdas that are required to be serializable are builder itself and (optionally-passed) entity manager.
During deserialization builder is invoked and all the config lambdas are rebuilt locally.

This way you get a lot less issues on accidentally bring in non-serializable constituents in the config lambdas.

What do you think?

tandraschko: In general i think we should just remove that statement. Everything else makes our implementation bad - but - its just not a problem at all and doesnt improve usability. Our lambdas are serialized, they dont have member fields like inner classes could have. So no real big state or anything will be serialized.

 just would like to understand your statements here because basically its just like this:
- JPALazyDataModel is serializable and if its placed inside a e.g. ViewScoped bean, it will be serialized
- Therefore currently our lambdas are serialized. If they are transient, they would be null after deserialization.
so where do you "store" this to overcame this, if its transient on your side?

lprimak: Did you read the medium article?? Is something there unclear somehow?

melloware: @lprimak i read the medium article and I don't understand how it solves the problem?

So if you have a transient field NAME and the value is "melloware".  When the EE container decides to serialize it, for example sample clustering between servers...because the value is transient it is now lost?  The other server can't rehydrate it as "Melloware" it would be NULL. 

Your solution seems like it only gets around Serialization errors but doesn't actually solve the serialization problem. Am I not reading that correctly?

lprimak: > Am I not reading that correctly?

:) no... Do you understand what `readResolve()` does? The article (IMHO) pretty clearly explains how it is able to restore the transient final variables based on the current state of the deserialized object, using Lombok's toBuilder().

JPALazyDataModel code does `readResolve()` and then executes the resulting builder to reconstitute al the other config lambdas without having to serialize them.

lprimak: I may write another test at some point to demonstrate this.
This functionality came out from actual applications not working. For example, the config lambdas can refer to `@RequestScoped` objects, or other non-serializable fields and method references inside `@ViewScope`d objects, which are not serializable.

melloware: I read it and I don't see how it possible solves the problem. If I have two servers and the EE container needs to serialize and send over the network so my second server gets the object and deserializes it there is NO way it can have the value "Melloware" and nothing Lombok can do can fix that?

If your solution does work in that scenario then I don't understand serialization. 

Your article should really test clustering with JSF and verify if it does work great!  However, if it doesn't your article is incorrect and misleading. 

The second EE will automatically ignore transient fields so your Lombok magic has no way of knowing about the value "Melloware" because it was never sent over the wire.  I HIGHLY recommend you test this. 

A simpler test would be to serialize your object to a flat file on disk. Shut your JVM down and restart your JVM and deserialize that file on disk back into an object.  I just can't see how that would possibly work?

melloware: Ok just read this https://www.baeldung.com/java-serialization-readobject-vs-readresolve

It has now convinced me that what you are doing with readResolve is violating the intention of transient and serialization to "work around" the issue of properly designing your objects.  

I would personally NOT recommend this approach and tout it as a benefit it's really a "hack" so developers can be lazy. 

So when you state it's helping with usability that is not accurate it is simply masking potential issues from developers and allowing them to be lazy. 

I agree with @@

lprimak: Ok, great, I think you now understand the solution.

> developers and allowing them to be lazy.

Yes!!! The framework needs to do the "heavy lifting" and allow the app developers to be "lazy"

Also, it's next to impossible to find these serialization issues in development. The always come up only in production when deploying on a real cluster. Possibly after days and days of running the code. It's good to do anything and everything to prevent these problems from appearing in the first place.

> simply masking potential issues

Can you elaborate on that? What "potential issues" are you talking about exactly?

In any case, this solution wasn't "invented" by me. Quarkus, Weld/CDI and Spring all use it to "fix" constructor injection after deserialization. This is actually a well-known method to fix objects after deserialization and existed well before 2011 when I started using it for this purpose.

Again, I want to reiterate, this does not solve a "hypothetical problem" but has worked for years with actual production applications and countless screens. These lessons have been learned over the last 10 years or so. Well before "native" `JPALazyDataModel` appeared.

lprimak: I have added a test demonstrating how you can use non-serializable class (e.g. RequestScoped) with config lambdas: https://github.com/flowlogix/flowlogix/blob/0b4e3ab2f5aae67cfcb4d63341bb730899742de2/jakarta-ee/flowlogix-datamodel/src/test/java/com/flowlogix/jeedao/primefaces/ModelTest.java#L384

tandraschko: I didnt read yet article in detail yet (it's weekend), i just wanted to hear 2-3 sentences where you store/restore your lambdas. no enough concentration on the couch via smartphone to read code examples.

i think we just should remove those sentences from the docs, as it makes our impl more bad. If yours is better, thats ok, you can just not expect that the main project adds this to the docs.
its like: hey users, our impl is not perfect, please use another community project.

if we have a problem here, we can fix it when we get the first issue in the issue tracker. You are also really welcome to contribute here.

docs/14_0_0/menu.md
```diff
@@ -8,6 +8,7 @@
   - [Hello World](/gettingstarted/helloworld.md)
   - [Whats New](/gettingstarted/whatsnew.md)
   - [Migration Guide](../migrationguide/migrationguide.md)
+  - [Community Projects](/gettingstarted/community.md)
```
melloware, line 10: Can you move this up 1 line too I think the Migration Guide should be last so Community Projects can go under Whats New

melloware (changes requested):

docs/14_0_0/menu.md
lprimak: done

lprimak:
@tandraschko I understand now. I have removed the sentence. Thanks for being patient with me.

melloware (approved):

PR #10923, opened 2023-10-27
Fix #10922: Spinner prefix breaks with min value

### LINKED ISSUE #10922: Spinner: Gets the min value when the prefix is defined
### Describe the bug

1. Set the value, min, max and the prefix
2. Click the up or down button
3. Input takes the min value

Without prefix there is no problem. I did not provide any test code since it can be seen from the showcase after changing the code as below.

### Reproducer

```xhtml
<p:spinner id="prefix" value="15" prefix="$" min="10" max="50"/>
<p:spinner id="prefix" value="15.20" prefix="$" min="10" max="50" stepFactor="0.1" decimalPlaces="2"/>
```

### Expected behavior

up/down: 15 to 16/14 (not 15 to 10)
up/down: 15.20 to 15.30/15.10 (not 15.20 to 10.00)

### PrimeFaces edition

Community

### PrimeFaces version

14.0.0-SNAPSHOT

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3.21

### Java version

17

### Browser(s)

Safari 17.0, Firefox 119.0

melloware:
Fix #10922: Spinner prefix breaks with min value

### DIFF
--- a/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/spinner/Spinner004Test.java
+++ b/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/spinner/Spinner004Test.java
@@ -51,7 +51,7 @@ public class Spinner004Test extends AbstractPrimePageTest {
         page.button.click();
 
         // Assert
-        Assertions.assertEquals("€0.33", spinner.getValue());
+        Assertions.assertEquals("€0,33", spinner.getValue());
         assertOutputLabel(page, "0.33");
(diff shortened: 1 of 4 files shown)


melloware:
this actually fixed a bug we already had an integration test for and a NOTE 😄 

```java
Assertions.assertEquals("€1456.78", spinner.getValue()); // TODO: this expected result seems odd to me
```

aripddev:
Thank you

PR #10929, opened 2023-10-29
Refactoring of #8369

### LINKED ISSUE #8369: Column: Add converter attribute for filter and short notation

Rapster:
Fix #8369

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/api/UITable.java
+++ b/primefaces/src/main/java/org/primefaces/component/api/UITable.java
@@ -35,6 +35,7 @@ import javax.faces.component.UINamingContainer;
 import javax.faces.component.ValueHolder;
 import javax.faces.component.search.SearchExpressionHint;
 import javax.faces.context.FacesContext;
+import javax.faces.convert.ConverterException;
 
 import org.primefaces.component.column.ColumnBase;
 import org.primefaces.component.headerrow.HeaderRow;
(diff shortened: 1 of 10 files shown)


primefaces/src/main/java/org/primefaces/component/columns/ColumnsBase.java
```diff
@@ -475,4 +476,12 @@ public void setTitle(String title) {
         getStateHelper().put(PropertyKeys.title, title);
     }
 
+    @Override
+    public String getConverter() {
```
tandraschko: Why string here and not object?

Rapster: bad copy/paste :D

primefaces/src/main/java/org/primefaces/component/api/UITable.java
```diff
@@ -203,6 +204,18 @@ default void updateFilterByValuesWithFilterRequest(FacesContext context, Map<Str
                 filterValue = params.get(valueHolderClientId);
             }
 
+            if (filterMeta.isShortFieldNotation()) {
```
tandraschko: can you add a comment here, why its only done when #isShortFieldNotiation?

everything else looks good -> +1

melloware (approved):

tandraschko:
do we need a taglib change?

PR #10935, opened 2023-11-02
#10793

### LINKED ISSUE #10793: DataTable: Initialized filteredValue collection leads to problems
### Describe the bug

When initializing the `filteredValue` attribute with an empty list, the `DataTable` displays all values of the value list, but any actions in the displayed rows do not work.

I can understand a bit how initializing the `filteredValue` attribute with an empty list might be a problem. If no code-based fix is possible, maybe you could add a prominent hint in the documentation and the attribute description?

A reproducer is attached.
[primefaces-test-datatable-filteredvalue-nonnull.zip](https://github.com/primefaces/primefaces/files/12856309/primefaces-test-datatable-filteredvalue-nonnull.zip)

### Reproducer

1. Open the "Select Action" MenuButton of any row.
2. Click on "Show Message"

### Expected behavior

The `FacesMessage` added in the action method should be displayed by the `p:growl` component.

### PrimeFaces edition

None

### PrimeFaces version

13.0.1

### Theme

_No response_

### JSF implementation

All

### JSF version

4.0

### Java version

17

### Browser(s)

_No response_

tandraschko:
Fixes #10793

### DIFF
--- a/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datatable/DataTable028Test.java
+++ b/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datatable/DataTable028Test.java
@@ -79,12 +79,7 @@ public class DataTable028Test extends AbstractDataTableTest {
         getButtonSave().click();
 
         // Assert
-        if (xhtml.contains("Without")) {
-            assertAfterBb3UpdateSorted();
-        }
-        else {
-            assertAfterBb3Update();
-        }
+        assertAfterBb3UpdateSorted();
 
         assertNoJavascriptErrors();
     }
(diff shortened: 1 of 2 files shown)


tandraschko:
@melloware can you also check the failing test pls?
for me it seems the test is wrong
i runned via jetty and it feels like the test does ignore the new sort order?

melloware:
Let me look.

melloware:
OK @christophs78 wrote this test and I am a little confused on the failure and what the expected result was correct or not orginally?

christophs78:
Good question why i did https://github.com/primefaces/primefaces/blob/master/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datatable/DataTable028Test.java#L82 two years ago.
My current best bet after looking ~15 minutes into this is to agree with @tandraschko .
Probably my motivation for introducing this if-clause was to make the IT passing. I should have left some comment regarding this behavior is not the expected one but simply what PF did at this point in time.

-->

Old:
```java
if (xhtml.contains("Without")) {
	assertAfterBb3UpdateSorted();
}
else {
	assertAfterBb3Update();
}
```

New:
```java
assertAfterBb3UpdateSorted();
```

melloware:
I made the change suggested and re-running the tests.

PR #10938, opened 2023-11-03
Preserve original filename when calling FileTypeDetector

### LINKED ISSUE #10937: FileUploadUtils#isValidFileContent obscurs file name/extension
FileUploadUtils#isValidFileContent obscurs the file name/extension [here](https://github.com/primefaces/primefaces/blob/13.0.2/primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java#L226)

This makes the default FileTypeDetector implementations provided by the JDK effectively useless, as those all seem to check the file extension against some form of known file extensions. See:
- [MimeTypesFileTypeDetector](https://github.com/openjdk/jdk11u/blob/master/src/java.base/unix/classes/sun/nio/fs/MimeTypesFileTypeDetector.java) (All Unix-like Systems; [Linux](https://github.com/openjdk/jdk11u/blob/master/src/java.base/linux/classes/sun/nio/fs/LinuxFileSystemProvider.java#L104), [MacOSX](https://github.com/openjdk/jdk11u/blob/master/src/java.base/macosx/classes/sun/nio/fs/MacOSXFileSystemProvider.java#L48),  [Solaris](https://github.com/openjdk/jdk11u/blob/master/src/java.base/solaris/classes/sun/nio/fs/SolarisFileSystemProvider.java#L87))
- [UTIFileTypeDetector](https://github.com/openjdk/jdk11u/blob/master/src/java.base/macosx/classes/sun/nio/fs/UTIFileTypeDetector.java) (MacOSX)
- [RegistryFileTypeDetector](https://github.com/openjdk/jdk11u/blob/master/src/java.base/windows/classes/sun/nio/fs/RegistryFileTypeDetector.java) (windows)

I noticed this when using `p:fileUpload` with `validateContentType="true"`, which forces me to provide a custom FileTypeDetector implementation to work

Reproducer (just start and try to upload any xml file): [fileupload-validatecontenttype.zip](https://github.com/primefaces/primefaces/files/13248939/fileupload-validatecontenttype.zip)

jungm:
This fixes #10937

### DIFF
--- a/docs/14_0_0/components/fileupload.md
+++ b/docs/14_0_0/components/fileupload.md
@@ -74,7 +74,7 @@ powered rich solution with graceful degradation for legacy browsers.
 | uploadButtonTitle | null | String | Native title tooltip for upload button
 | uploadIcon | pi pi-upload | String | The icon of upload button
 | uploadLabel | Upload | String | Label of the upload button.
-| validateContentType | false | Boolean | Whether content type validation should be performed, based on the types defined in the accept attribute. Default is false.
+| validateContentType | false | Boolean | Whether content type validation should be performed, based on the types defined in the accept attribute. Default is false. **NOTE:** By default the file type gets detected using the file extension, provide a custom FileTypeDetector implementation for more sophisticated detection methods.
 | validator | null | MethodExpr | A method expression that refers to a method validating the input.
 | validatorMessage | null | String | Message to be displayed when validation fails.
 | value | null | Object | Value of the component than can be either an EL expression of a literal text.
--- a/docs/14_0_0/components/fileupload.md
+++ b/docs/14_0_0/components/fileupload.md
@@ -74,7 +74,7 @@ powered rich solution with graceful degradation for legacy browsers.
 | uploadButtonTitle | null | String | Native title tooltip for upload button
 | uploadIcon | pi pi-upload | String | The icon of upload button
 | uploadLabel | Upload | String | Label of the upload button.
-| validateContentType | false | Boolean | Whether content type validation should be performed, based on the types defined in the accept attribute. Default is false. **NOTE:** By default the file type gets detected using the file extension, provide a custom FileTypeDetector implementation for more sophisticated detection methods.
+| validateContentType | false | Boolean | Whether content type validation should be performed, based on the types defined in the accept attribute. Default is false.
 | validator | null | MethodExpr | A method expression that refers to a method validating the input.
 | validatorMessage | null | String | Message to be displayed when validation fails.
 | value | null | Object | Value of the component than can be either an EL expression of a literal text.
--- a/primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
+++ b/primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
@@ -30,7 +30,6 @@ import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
-import java.util.UUID;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
@@ -242,8 +241,7 @@ public class FileUploadUtils {
             return true;
         }
 
-        String tempFilePrefix = UUID.randomUUID().toString();
-        Path tempFile = Files.createTempFile(tempFilePrefix, null);
+        Path tempFile = Files.createTempDirectory("pf-fileupload").resolve(fileName);
 
         try {
             try (InputStream in = new PushbackInputStream(new BufferedInputStream(stream))) {
--- a/primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
+++ b/primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
@@ -27,6 +27,7 @@ import java.io.*;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
@@ -242,8 +243,8 @@ public class FileUploadUtils {
             return true;
         }
 
-        Path tempDirectory = Files.createTempDirectory("pf-fileupload");
-        Path tempFile = Files.createTempFile(tempDirectory, UUID.randomUUID().toString(), null);
+        String tempFilePrefix = UUID.randomUUID().toString();
+        Path tempFile = Files.createTempFile(tempFilePrefix, null);
 
         try {
             try (InputStream in = new PushbackInputStream(new BufferedInputStream(stream))) {
@@ -261,8 +262,13 @@ public class FileUploadUtils {
             // If this first attempt failed we try again, but now while preserving the original file name/extension
             // to e.g. make the JDK default FileTypeDetector work
             if (contentType == null) {
-                tempFile = Files.move(tempFile, tempDirectory.resolve(fileName));
-                contentType = context.getFileTypeDetector().probeContentType(tempFile);
+                String fileExtension = getFileExtension(fileName);
+
+                if (fileExtension != null) {
+                    Path newTempFile = Files.createTempFile(tempFilePrefix, fileExtension);
+                    tempFile = Files.move(tempFile, newTempFile, StandardCopyOption.REPLACE_EXISTING);
+                    contentType = context.getFileTypeDetector().probeContentType(tempFile);
+                }
             }
 
             if (contentType == null) {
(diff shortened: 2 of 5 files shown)


primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
```diff
@@ -242,8 +241,7 @@ private static boolean isValidFileContent(PrimeApplicationContext context, FileU
             return true;
         }
 
-        String tempFilePrefix = UUID.randomUUID().toString();
-        Path tempFile = Files.createTempFile(tempFilePrefix, null);
+        Path tempFile = Files.createTempDirectory("pf-fileupload").resolve(fileName);
```
melloware: Just checking what happens if the directory already exists?

tandraschko:
+1

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
Rapster: I like the idea of having a single directory for PF (meant for internal stuff). Is there anywhere else we do this? If yes, create a global variable

Rapster: AFAIR, extension was removed on purpose to avoid content type checking based on extension. Does tika behave the same if extension present in filename?

primefaces/src/test/java/org/primefaces/util/FileUploadUtilsTest.java
```diff
@@ -148,10 +148,6 @@ public void isValidTypeContentTypeCheck() throws IOException {
         when(fileUpload.getAccept()).thenReturn("image/tiff");
         Assertions.assertFalse(FileUploadUtils.isValidType(appContext, fileUpload, createFile("test.png", "image/png", png)));
 
-        //Tampered - Apache Tika or Mime Type must be in the classpath for this to work
-        when(fileUpload.getAccept()).thenReturn("image/gif");
-        Assertions.assertFalse(FileUploadUtils.isValidType(appContext, fileUpload, createFile("test.gif", "image/gif", exe)));
-        Assertions.assertTrue(FileUploadUtils.isValidType(appContext, fileUpload, createFile("test.png", "image/png", gif)));
```
Rapster: why removed?

melloware (changes requested):

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
jungm: tika by default also tries to detect content type by file extension: https://github.com/apache/tika/blob/2.9.1/tika-java7/src/main/java/org/apache/tika/filetypedetector/TikaFileTypeDetector.java

jungm: This won't happen, [every call to Files#createTempDirectory creates a new dir](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Files.html#createTempDirectory(java.lang.String,java.nio.file.attribute.FileAttribute...)). `pf-fileupload` really is just a prefix for the directory name to be generated

primefaces/src/test/java/org/primefaces/util/FileUploadUtilsTest.java
jungm: will investige this further on monday, removed those because these tests were failing and I assumed no proper FileTypeDetector impl was present on classpath - I only checked for tika and not for mime-types

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
melloware: Thanks!

Rapster: Indeed not great, I think there should be 2 pass then: one without extension, and another one with extension in case resolved content type is null

jungm: I actually like your idea also a lot more than what I originally had in mind, forcing FileTypeDetector impls to do a deep inspection by default instead of going the lazy way of looking at the file extension

primefaces/src/test/java/org/primefaces/util/FileUploadUtilsTest.java
jungm: I reverted that change; the test was failing because exposing the file name to mime-types caused it to go the lazy route of looking up the file extension instead of inspecting the file content

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
```diff
@@ -242,8 +242,8 @@ private static boolean isValidFileContent(PrimeApplicationContext context, FileU
             return true;
         }
 
-        String tempFilePrefix = UUID.randomUUID().toString();
-        Path tempFile = Files.createTempFile(tempFilePrefix, null);
+        Path tempDirectory = Files.createTempDirectory("pf-fileupload");
+        Path tempFile = Files.createTempFile(tempDirectory, UUID.randomUUID().toString(), null);
```
Rapster: is this tmp directory really necessary?

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
```diff
@@ -292,26 +303,31 @@ private static boolean isValidFileContent(PrimeApplicationContext context, FileU
...
-                    try {
-                        tempFile.toFile().deleteOnExit();
-                    }
-                    catch (Exception ex1) {
-                        if (LOGGER.isLoggable(Level.WARNING)) {
-                            LOGGER.log(Level.WARNING, String.format("Could not register temporary file %s for deletion on JVM exit",
-                                    tempFile.toAbsolutePath()), ex1);
-                        }
+            deleteTempFile(tempFile);
+            deleteTempFile(tempDirectory);
```
Rapster: if tempfile in tempdirectory, deleting the directory should be enough?

jungm: I delete the file separately first because:

>If the file is a directory then the directory must be empty.

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Files.html#delete(java.nio.file.Path)

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
jungm: I kept it because when the second pass is run I completely keep the original filename received from the upload and wanted to avoid "collisions" when two files with the same name are uploaded at exactly the same time, creating/overriding something like `/tmp/file.jpg` twice

melloware (approved):

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
Rapster: What about extracting extension first, and on the second run, rename tempfile: tempfile name + extension?

jungm: Sounds good, PR is updated

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
```diff
@@ -314,6 +331,15 @@ private static boolean isValidFileContent(PrimeApplicationContext context, FileU
         return true;
     }
 
+    private static String getFileExtension(String fileName) {
```
Rapster: You can use `org.apache.commons.io.FilenameUtils#getExtension`

jungm: done

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
```diff
@@ -252,8 +253,24 @@ private static boolean isValidFileContent(PrimeApplicationContext context, FileU
...
+            // this is why we intentionally hide the original file name/extension at first because both
+            // tika and mime-types go the easy way of looking up the file extension if possible.
+            // If this first attempt failed we try again, but now while preserving the original file name/extension
+            // to e.g. make the JDK default FileTypeDetector work
+            if (contentType == null) {
+                String fileExtension = getFileExtension(fileName);
+
+                if (fileExtension != null) {
+                    Path newTempFile = Files.createTempFile(tempFilePrefix, fileExtension);
+                    tempFile = Files.move(tempFile, newTempFile, StandardCopyOption.REPLACE_EXISTING);
```
Rapster: Can you try renaming the file without creating a new temp file. I found this:

```java
Path source = Paths.get("Notes.txt");
Files.move(source, source.resolveSibling("Keynotes.txt"),
            StandardCopyOption.REPLACE_EXISTING);
```

Source: https://www.codejava.net/java-se/file-io/how-to-rename-move-file-or-directory-in-java

jungm: done

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
```diff
@@ -252,8 +252,25 @@ private static boolean isValidFileContent(PrimeApplicationContext context, FileU
...
+            // We want to force the FileTypeDetector impl to try to detect the file type by its content,
+            // this is why we intentionally hide the original file name/extension at first because both
+            // tika and mime-types go the easy way of looking up the file extension if possible.
+            // If this first attempt failed we try again, but now while preserving the original file name/extension
+            // to e.g. make the JDK default FileTypeDetector work
+            if (contentType == null) {
+                String fileExtension = FilenameUtils.getExtension(fileName);
+
+                if (!fileExtension.isEmpty()) {
+                    String newFileName = FilenameUtils.removeExtension(tempFile.getFileName().toString()) + "." + fileExtension;
```
Rapster: tempFile is supposed to be without extension no?

jungm: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Files.html#createTempFile(java.lang.String,java.lang.String,java.nio.file.attribute.FileAttribute...)

> suffix - the suffix string to be used in generating the file's name; may be null, in which case ".tmp" is used

Maybe passing `""` as the suffix could work

jungm: Yeah that seems to work, new code is pushed

docs/14_0_0/components/fileupload.md
```diff
@@ -74,7 +74,7 @@ powered rich solution with graceful degradation for legacy browsers.
 | uploadButtonTitle | null | String | Native title tooltip for upload button
 | uploadIcon | pi pi-upload | String | The icon of upload button
 | uploadLabel | Upload | String | Label of the upload button.
-| validateContentType | false | Boolean | Whether content type validation should be performed, based on the types defined in the accept attribute. Default is false.
+| validateContentType | false | Boolean | Whether content type validation should be performed, based on the types defined in the accept attribute. Default is false. **NOTE:** By default the file type gets detected using the file extension, provide a custom FileTypeDetector implementation for more sophisticated detection methods.
```
Rapster: no longer true isn'it?

jungm: yup

PR #10947, opened 2023-11-06
#10946

### LINKED ISSUE #10946: CSP: add noonce only to JSF forms
Nonce input is added to all forms, in some cases this may not be desirable. How about adding an attribute like ignore-primefaces-none to a form as attribute, if that exists the nonce is not attached dynamically.

tandraschko:
Fixes #10946

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.csp.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.csp.js
@@ -37,6 +37,10 @@ if (!PrimeFaces.csp) {
             var forms = document.getElementsByTagName("form");
             for (var i = 0; i < forms.length; i++) {
                 var form = forms[i];
+                if (!PrimeFaces.csp.isFacesForm(form)) {
+                    continue;
+                }
+
                 var input = form.elements[PrimeFaces.csp.NONCE_INPUT];
                 if (!input) {
                     input = document.createElement("input");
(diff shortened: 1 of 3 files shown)


melloware (approved):

tandraschko:
@melloware do we have CSP tests in normal CI? i didnt test it currently TBH :D

melloware:
Yes we do it runs CSP.

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.csp.js
```diff
@@ -48,6 +52,27 @@ if (!PrimeFaces.csp) {
...
+         * @param {HTMLInputElement} [form] The form to check.
+         * @return {boolean} true if the form is a Faces form.
+         */
+        isFacesForm: function(form) {
+            if (form.method === 'post') {
+                var children = form.children;
+                for (var i = 0; i < children.length; i++) {
+                    var child = children[i];
+                    if (child instanceof HTMLInputElement) {
+                        if (child.name.indexOf(PrimeFaces.VIEW_STATE) !== -1) {
```
jepsar, line 66: Why not use && here?

melloware: it could be this

```js
isFacesForm: function (form) {
    return form.method === 'post' && Array.from(form.children).some(child => {
        return child instanceof HTMLInputElement && child.name.includes(PrimeFaces.VIEW_STATE);
    });
}
```

tandraschko: i will provide a small refactored version; nevertheless i like loops more then lambas for such checks ;D

melloware: No worries!

PR #10959, opened 2023-11-07
Cleanup FileUploadUtils#isValidFileContent

Rapster:


melloware (approved):

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
```diff
@@ -224,111 +222,105 @@ protected static String convertJavaScriptRegex(String jsRegex) {
...
             if (contentType == null) {
-                String fileExtension = FilenameUtils.getExtension(fileName);
-
-                if (!fileExtension.isEmpty()) {
-                    String newFileName = tempFile.getFileName().toString() + "." + fileExtension;
-                    tempFile = Files.move(tempFile, tempFile.resolveSibling(newFileName));
-
-                    contentType = context.getFileTypeDetector().probeContentType(tempFile);
-                }
+                tempFile = Files.move(tempFile, tempFile.resolveSibling(fileName));
```
jungm: I think this could cause a collision if two files with the same name are uploaded simultaniously:
1. `test.jpg` is uploaded
2. It gets saved in a temporary file with a randomized name, e.g. `/tmp/test3887755495119540932` (fine)
3. Then it's moved according to its original filename: `/tmp/test.jpg` (problematic, could exist already because of another request, some other program writing temporary files, etc.)

Really what i think should happen is that the tempFile name is taken as is (we can expect it to be unique because of the random name) and the file extension gets appended

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
```diff
@@ -224,111 +222,105 @@ protected static String convertJavaScriptRegex(String jsRegex) {
...
+    private static void deleteFile(Path tempFile) {
+        try {
+            Files.delete(tempFile);
+        }
+        catch (Exception ex) {
+            if (LOGGER.isLoggable(Level.WARNING)) {
+                LOGGER.log(Level.WARNING, String.format("Could not delete temporary file %s, will try to delete on JVM exit",
+                        tempFile.toAbsolutePath()), ex);
+                try {
+                    tempFile.toFile().deleteOnExit();
```
jungm: Not exactly related to your changes, I didn't notice this as well before
But I really do think the deleteOnExit fallback should be called, even if `Level.WARNING` is not loggable

Rapster:
Great feedbacks @jungm! Thanks 👍

jungm (approved):

PR #10974, opened 2023-11-10
DatePicker filter range example clarification

melloware:
Based on Discord conversation and we ran into this in the real world as well

### DIFF
--- a/primefaces-showcase/src/main/webapp/ui/data/datatable/columns.xhtml
+++ b/primefaces-showcase/src/main/webapp/ui/data/datatable/columns.xhtml
@@ -33,7 +33,7 @@
 
                     <p:columns value="#{dtColumnsView.columns}" var="column" headerText="#{column.header}" field="#{column.property}">
                         <f:facet name="filter">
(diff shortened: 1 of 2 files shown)


primefaces-showcase/src/main/webapp/ui/data/datatable/columns.xhtml
```diff
@@ -33,7 +33,7 @@
 
                     <p:columns value="#{dtColumnsView.columns}" var="column" headerText="#{column.header}" field="#{column.property}">
                         <f:facet name="filter">
-                            <p:datePicker selectionMode="range" onchange="PF('customersTable').filter()" rendered="#{column.type eq 'date'}"/>
+                            <p:datePicker selectionMode="range" converter="javax.faces.DateTime" onchange="PF('customersTable').filter()" rendered="#{column.type eq 'date'}"/>
```
FlipWarthog, line 36: You didn't have to change any types in the Java?

melloware: The Java is using `java.time.LocalDate`  I have a feeling its when you use `java.util.Date` this happens like we do but it doesn't hurt giving it the converter "hint".

FlipWarthog (approved):

PR #10978, opened 2023-11-14
Fix #10967 - Datatable: support custom components as filter

### LINKED ISSUE #10967: Datatable: header filters do not render custom components
### Description

I am using Primefaces 13.01 and I am attempting to put a slider to do range filtering on a column in a datatable similar to the way that the datepicker component can be used to filter a date range. The output text renders but the slider does not.  I can see the controller filter function being called on page load  but the slider isn't rendered. Here is the the code snippet:

```xml
<p:column field="activityRating" headerText="Activity Rating" filterMatchMode="between"  filterFunction="#{controller.filterByRating}">
    <f:facet name="filter">
        <h:outputText id="displayRange" value="Between #{controller.filterMin} and #{controller.filterMax}" />
        <p:slider for="min,max" display="displayRange" minValue="1f" maxValue="5f" step=".1" style="width: 100px" range="true" displayTemplate="Between {min} and {max}" converter="javax.faces.Float" />
        <h:inputHidden id="min" value="#{controller.filterMin}" />
        <h:inputHidden id="max" value="#{controller.filterMax}" />
    </f:facet>
</p:column>
```

Unless I am doing something wrong, and that is entirely possible, I think being able to use custom components like datepicker to filter would be a good enhancement

### Additional context

_No response_

Rapster:
Fix #10967

### DIFF
--- a/primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
+++ b/primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
@@ -794,27 +794,4 @@ public class ComponentUtils {
             return value;
         }
     }
-
-    public static <T> T findComponent(UIComponent root, Class<T> childType) {
-        if (root != null) {
-            if (childType.isInstance(root)) {
-                return (T) root;
-            }
-
-            return findComponentInChildren(root, childType);
-        }
-
-        return null;
-    }
-
-    public static <T> T findComponentInChildren(UIComponent parent, Class<T> childType) {
-        if (parent.getChildCount() > 0) {
-            for (UIComponent child : parent.getChildren()) {
-                if (childType.isInstance(child) && child.isRendered()) {
-                    return (T) child;
-                }
-            }
-        }
-        return null;
-    }
 }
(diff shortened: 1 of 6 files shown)


melloware (approved):

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
```diff
@@ -794,4 +794,27 @@ public static Object convertToType(Object value, Class<?> valueType, Logger logg
             return value;
         }
     }
+
+    public static <T> T findComponent(UIComponent root, Class<T> childType) {
```
tandraschko: just use ComponentTraversalUtils#first

melloware (approved):

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
melloware: Nice hint about ComponentTraversalUtils!

Rapster: I'm not too sure yet about using ComponentTraversalUtils (as it traverses the children's children) By introducing `ComponentUtils#findComponent` (which only visits self or children and take account rendered attr) i was planning to reuse that method in various places where sometimes we need to check the presence of a component (for instance: celleditor, rowexpansion etc.)

tandraschko: then we need a better name for it
findComponent is something different already in JSF

melloware (approved):

PR #10990, opened 2023-11-18
Replace with new ComponentTraversalUtils#firstChildRendered

Rapster:


melloware (approved):

tandraschko:
+1

also please switch the firstChildRendered to a index loop, thats what we use everywhere when looping through JSF DOM

PR #10999, opened 2023-11-21
Fix #10996 - FileUpload: add Inline ajax status

### LINKED ISSUE #10996: Simple FileUpload: add Inline ajax status
Like for commandButton

tandraschko:
Fixes #10996

this also moves the code to a utils method, as its already used in several places


melloware (approved):
Nice work. @jepsar also please review

tandraschko:
@jepsar if you have further ideas to improve the code or also unify the status for autocomplete/commandlink, please create a further issue/PR
i will merge this for now, tested all cases in the showcase

jepsar:
I was on vacation this week. Love the refactor!

jepsar:
No style changes were needed?

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.js
```diff
@@ -378,6 +378,62 @@
...
+            $(document).on('pfAjaxSend.' + widget.id, function(e, xhr, settings) {
+                if (isXhrSource.call(this, widget, settings)) {
+                    widget.ajaxCount++;
+                    if (widget.ajaxCount > 1) {
+                        return;
+                    }
+
+                    button.addClass('ui-state-loading');
+
+                    if (widget.hasOwnProperty('disable')
```
jepsar, line 404: What does / should this line do? It is not passed for the "Ajax Disabling" buttons in the showcase.
```
console.log(widget.hasOwnProperty('disable'));
console.log(widget.cfg.disableOnAjax);
```
Results in
```
false
undefined
```

jepsar: Is it plain JS? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty ?

jepsar: How can I check this with file upload? I think we might be able to use `typeof widget.disable === 'function'`

melloware: SHouldn't that first one be `widget.cfg.hasOwnProperty` not `widget.hasOwnProperty` ?

PR #11000, opened 2023-11-21
Fix #10992 - Simple FileUpload: add property to hide filename (especially for basic & auto mode)

### LINKED ISSUE #10992: Simple FileUpload: add property to hide filename (especially for basic & auto mode)
https://www.primefaces.org/showcase/ui/file/upload/basicAuto.xhtml?jfwid=3b1c7

in this case the file is automatically uploaded after selecting the file. in normal cases, the uploaded file name is displayed right after the button but only for a very short time.

IMO we can completely skip rendering the uploaded file name in basic+auto mode. Or should we make it configurable?

CC @melloware

tandraschko:
Fix #10992


Rapster:
Why validation messages should be hidden?

tandraschko:
I dont rely on client side validation, i do all on server side.
Story is to just have a small as possible upload button, in a column and add global p:messages via server side. And avoid that Filename flickering.

 I could also refactor to 'displayFilename' for now, doesnt matter for me for currently.

In general the csv validation could also be moved to real PrimeFaces CSV with p:messages support. But thats much more work for now.

melloware (approved):

tandraschko:
I think i will refactor to displayFilename tomorrow and create a new issue to refactor simple Upload CSV

tandraschko:
follup ticket: https://github.com/primefaces/primefaces/issues/11001

PR #11017, opened 2023-11-22
Fix #10998: SelectOneMenu loading indicator

### LINKED ISSUE #10998: SelectOneMenu: dynamic=true add LoadingIndicator during first adding elements to DOM
### Description

Hi all,
it'll be nice having, like commandbutton,  a LoadingIndicator rotating after the first click on a SelectOneMenu with dynamic=true, just for notify a "load in progress" status tu the final user.

A LoadingIndicator  in the text field will be perfect

top tier: show an empty panel of the selectOneMenu with a bigger LoadingIndicator  inside and then (after loading) hiding it and showing the list

God Tier: both

### Additional context

_No response_

melloware:
Fix #10998: SelectOneMenu loading indicator

image

Tested in RTL mode as well.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.css
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.css
@@ -288,6 +288,20 @@
     cursor: default;
 }
 
+/* loading */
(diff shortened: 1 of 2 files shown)


tandraschko:
+1

PR #11020, opened 2023-11-23
Fix #11001 - Simple FileUpload: refactor client side validation

### LINKED ISSUE #11001: Simple FileUpload: refactor client side validation
Currently Simple-FileUpload displays some basic validation errors inside the span '.ui-fileupload-filename', which is quite weird.
We should check if we can implement it this with real PrimeFaces CSV to also enabled usage of p:message(s) and p:growl

tandraschko:
Fix #11001


melloware (approved):

tandraschko:
it now also contains code to correclty handling guarding when CSV is used on the button
this is why the tests failes because it expects http/ajax requests even when validation failed

PR #11021, opened 2023-11-23
Fix total size count for data scroller

### LINKED ISSUE #6087: DataScroller: Lazy loading losing track of items loaded
Running clean version 8.   2nd chunk replaces item association with first chunk. The issue is the exact same issue as #5354 which was closed with a pull request, but I noticed that the changes were reverted by a subsequent change three months later (I think #5612).

As soon as I work through the issues I'm having building my own version of 9.0-SNAPSHOT I'll retest it there, but the code looks unchanged.

**Environment:**
 - PF Version: _8.0_
 - JSF : jboss-jsf-api_2.3_spec-3.0.0.Final.jar (WildFly 11)

### LINKED ISSUE #11019: DataScroller: with lazyloading only load one time
### Describe the bug

When I try to use a dataScroller with lazy loading, the data load only once.  

### Reproducer

I've made a reproducer : https://github.com/Samuel-StO/primefaces-datascroller-issue

In the reproducer, I have a list of 100 cities that I'm trying to display in a lazy loading datascroller. 
I've set the number of elements per chunk to 25. 

The behavior I'm observing is that at first there are 25 elements, then if I scroll it loads the next 25. But if you keep scrolling, it won't load the next 25. 
In fact, it isn't capable of loading more data than once, even if there's more data to load. 

The video shows problematic behavior :
https://github.com/primefaces/primefaces/assets/86346301/c9a18e2c-85b3-4355-80a5-d9b90c44c515

### Expected behavior

The desired behavior is for the datascroller to continue loading data when you scroll.

### PrimeFaces edition

Community

### PrimeFaces version

13.0.3

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

8

### Browser(s)

_No response_

Samuel-StO:
Fix #11019
Fix #6087 

Use the count method for data scroller when we are in lazy

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/datascroller/DataScrollerRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/datascroller/DataScrollerRenderer.java
@@ -181,7 +181,7 @@ public class DataScrollerRenderer extends CoreRenderer {
         WidgetBuilder wb = getWidgetBuilder(context);
         wb.init("DataScroller", ds)
                 .attr("chunkSize", chunkSize)
-                .attr("totalSize", ds.getRowCount())
+                .attr("totalSize", getTotalSize(ds))
                 .attr("loadEvent", loadEvent)
                 .attr("mode", ds.getMode(), "document")
                 .attr("buffer", ds.getBuffer())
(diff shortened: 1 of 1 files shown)


primefaces/src/main/java/org/primefaces/component/datascroller/DataScrollerRenderer.java
```diff
@@ -192,6 +192,16 @@ protected void encodeScript(FacesContext context, DataScroller ds, int chunkSize
 
         wb.finish();
     }
+    
+    private static int getTotalSize(DataScroller ds) {
```
tandraschko: private static -> protected

primefaces/src/main/java/org/primefaces/component/datascroller/DataScrollerRenderer.java
```diff
@@ -192,6 +192,16 @@ protected void encodeScript(FacesContext context, DataScroller ds, int chunkSize
...
+        if ( ds.isLazy() ) {
```
tandraschko: our styleguide is `if (ds.isLazy()) {` (remove spaces)

tandraschko:
checkstyle still failing, i wonder if you even built the source on your local machine before?

melloware (approved):
build is passing now.  LGTM

tandraschko:
looks good for me
@melloware can you apply it and port to master?

melloware:
Yep will do!

PR #11025, opened 2023-11-24
Fix sonar code smells

Rapster:


melloware (approved):

primefaces/src/main/java/org/primefaces/component/fileupload/FileUploadBase.java
```diff
@@ -491,12 +491,7 @@ public void setDropZone(String dropZone) {
     }
 
     public boolean isDisplayFilename() {
-        return ComponentUtils.eval(getStateHelper(), PropertyKeys.displayFilename, () -> {
-            if (getMode().equals("simple") && isAuto()) {
-                return false;
-            }
-            return true;
-        });
+        return ComponentUtils.eval(getStateHelper(), PropertyKeys.displayFilename, () -> !getMode().equals("simple") || !isAuto());
```
tandraschko, line 494: IMO its not nice to read
`!(getMode().equals("simple") && isAuto())` would be ok for me if you would like to switch to 1 liner

tandraschko: @Rapster it even breaks the functionality. Per default it should not be displayed when simple&&auto, not just one of both

Rapster: Weird, the condition looks the same to me, and the MR passed before I merged. Let's see with the new MR

tandraschko: could also be that i have a "thinking error" in my brain :D but with your new PR it just reads more cleaner

and we dont have a test for this functionality :/

PR #11028, opened 2023-11-24
13.0.4 Remove OWASP Dependency Check

melloware:


riyadomf:
@melloware Why did we remove the OWASP dependency check? Will the TextEditor `secure="false"` work without the OWASP library?

melloware:
OWASP dependency check was scanning PrimeFaces libraries for vulnerabilities. We don't need that anymore as GitHub Dependabot already alerts us to dependencies and security violations.

TextEditor uses OWASP Santizer...so I think you are confusing the libraries.  OWASP Dependency Check vs OWASP Sanitizer.

All documented here: https://primefaces.github.io/primefaces/15_0_0/#/components/texteditor?id=getting-started-with-the-editor

PR #11060, opened 2023-11-30
Czech translation for primefaces.data.EMPTY_MESSAGE

milansie:
Added missing czech translation for primefaces.data.EMPTY_MESSAGE .


tandraschko:
Thanks!

PR #11061, opened 2023-11-30
sonar issues

tandraschko:
sonar issues

@melloware can you please review?


primefaces/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
```diff
@@ -980,11 +980,16 @@ public void encodeTbody(FacesContext context, DataTable table, boolean dataOnly,
...
-            int virtualScrollRowCount = (table.getScrollRows() * 2);
-            rowCountToRender = Math.min(virtualScrollRowCount, rowCount);
+            rowCountToRender = Math.min(table.getScrollRows() * 2, rowCount);
+        }
+        else if (table.isLiveScroll()) {
+            rowCountToRender = table.getScrollRows() + table.getScrollOffset();
+        }
+        else {
+            rowCountToRender = rows == 0 ? rowCount : rows;
         }
```
melloware, line 993: I think this logic needs to be 

```java
int rowCountToRender;

if (rows == 0) {
    if (table.isVirtualScroll()) {
        rowCountToRender = Math.min(table.getScrollRows() * 2, rowCount);
    }
    else if (table.isLiveScroll()) {
        rowCountToRender = table.getScrollRows() + table.getScrollOffset();
    }
    else {
        rowCountToRender = rowCount;
    }
} else {
    rowCountToRender = rows;
}
```

melloware (changes requested):

primefaces/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
tandraschko: the virtualScoll if was outside before
wait 1 min, maybe this is more clear

melloware: Yeah it looks like they were specifically looking for 0 rows before attempting the other checks?

melloware: Sorry I updated my code to say "virtualScroll" i had LiveScroll twice.

tandraschko: not sure TBH
nows its 100% the old logic again

melloware (approved):

PR #11088, opened 2023-12-07
Fix #11076: Colorpicker use nextZIndex

### LINKED ISSUE #11076: Color Picker: text input in nested dialogue
Hello, 

I noticed in my application and even in the primefaces showcase that the color picker, if used inside a dialog, has a non-functioning input field.

image

I can copy and paste value to the color field, but I cannot manually write or delete values. I do not think this is the intended behaviour since in the other color pickers in the showcase it is possible to do so.

### JSF version

2.3.9

### Java version

21

melloware:
Fix #11076: Colorpicker use nextZIndex

cc @tandraschko


tandraschko:
did you try to open the dialog several times?

melloware:
Yep this works because I delay updating the zIndex until the color picker popup is opened so its after the dialog.  It is working if I open it many times.

Issue #11036, opened 2023-11-26
CSV: enable (command-)button only after client side validation passes


christophs78:
### Description

Modern JS clients often enable save - buttons only after all necessary information is provided by the user. We may provide similar functionality for PF-CSV.

### Describe the solution you would like

validation.common.js already contains validate-function which is currently called by clicking (command-)buttons with `validateClient="true"`.
We may provide a new attribute like `enabledByValidateClient="true"` which enables the buttons after CSV is passed successful.
We´ll need to find the right events to trigger this ongoing background-validation which enables / disables the (command-)button.
( Or a bit on the ugly side we´ll need to do the validation based on something like https://developer.mozilla.org/en-US/docs/Web/API/setInterval.)
What are your thoughts?

tandraschko:
+1 in general

we may also refactor this slightly after this new feature: #11040

tandraschko:
Basically the whole thing should be easy to implement
The user must attach p:clientValidator to each input and we have 99% of the whole mechanism: https://www.primefaces.org/showcase/ui/csv/event.xhtml

we only attach the validation state to the parent form ($form.data('p-validationfailed', true))
p:clientValidator can then also jquery events and p:commandButton can listen on it


the big question is: 
can we make it easier for the user, like attaching a p:clientValidator to the form isntead of each input?

christophs78:
> we only need to attach the validation state to the parent form ($form.data('p-validationfailed', true))
> p:clientValidator can then also fire jquery events and the p:commandButton can listen on it

Think this is a bit more complicated. IMO we have to look at process-attribute of the button because not always the whole form is submittet.

And we may need some kind of initialization after initial page-load and partial ajax-updates. (Some good idea´s where/how we can add this?)

Just to get some feeling how large the "iceberg" before jumping into this adventure.

PR #11097, opened 2023-12-09
#11036: CSV: enable (command-)button only after client side validation passes (ui-state-error + ui-state-valid)

christophs78:
Just started working on this. Will probably take at least weeks maybe even a month or tow.


tandraschko:
can you explain a bit the mechanism?

christophs78:
> can you explain a bit the mechanism?

Basic idea is to plug into following events:

- full page load (not yet implemented)
- ajax-update (not yet implemented)
- `p:clientValidator` - event

Whenever one of this events is triggered we look for instances of `p:commandButton` marked with `enabledByValidateClient="true"`. Each `p:commandButton` marked such way has following data-attributes: `data-pf-validateclient-ajax="true" data-pf-validateclient-process="j_idt286:selectOneMenu" data-pf-validateclient-update="j_idt286:grid"`

We use this information / metadata to run CSV and decide afterwards whether to enable or disable the button.

primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/csv/Csv002Test.java
```diff
@@ -0,0 +1,124 @@
...
+
+        // Assert
+        Assertions.assertEquals("", page.msgSelectBooleanCheckbox.getText());
+
+        // Act
+        page.selectBooleanCheckbox.check();
+        page.selectBooleanCheckbox.uncheck();;
+
+        // Assert
+        Assertions.assertEquals("yes/no: Validation Error: Value is required.",
```
christophs78: Any idea what CSV should do with `p:selectBooleanCheckbox` and `p:triStateCheckbox` set to required?
image
IMO it does currently nothing.

(I´m adding CSV-IT´s as part of this PR too. This IT fails currently.)

christophs78:
As usual there are more things we may do.
But IMO we are at a good point to review this PR. Additional i provided our first IT´s for CSV as part of this PR. (Two of them disabled for now because required="true" does not seem to work (make sense?) for some components like `SelectBooleanCheckbox` and `TriStateCheckbox`.)

melloware:
I know required doesn't make sense for Select Boolean Checkbox but Tristate one might make sense?

christophs78:
Yep required=true makes IMO sense for TriStateCheckbox. Even for SelectBooleanCheckbox there may be some usecases where it makes sense. e.g. you need to check this checkbox to continue during some registration form. But there may be usecases where it does not make sense. (No one forces you to use required=true, so there may be room to implement required=true even for SelectBooleanCheckbox )
IMO we should open separate issues for this once we figured out what´s "common sense".

melloware:
Agreed

tandraschko:
do we really need additional css classes? styles should be the same, indepdent of CSV and SSV. If you need to know if it was triggered by CSV, maybe add a data attr?

christophs78:
> do we really need additional css classes? styles should be the same, indepdent of CSV and SSV. If you need to know if it was triggered by CSV, maybe add a data attr?

In an earlier iteration i had data-attributes. Then i refactored it to styles classes.
Our current style-classes only cover the invalid-state (`ui-state-error`) but IMO we want/need also style-classes for the valid state. So i took some inspiration from https://angular.io/guide/form-validation#control-status-css-classes.

My target was to allow something like following with JSF/PrimeFaces
image

An option would be to stick with `ui-state-error` and add something new like `ui-state-valid`. Implicit knowing this is only set (at least for now) via CSV.

tandraschko:
+1 for ui-state-valid for now
if you need to know that it failed/succeeded in CSV, we can still later check the best solution

christophs78:
Behaviour before i start reworking `ui-state-xxx` css-classes.
https://github.com/primefaces/primefaces/assets/10461942/60e4e9c2-311a-4620-853d-4c062d760172

christophs78:
After CSS-class rework (reuse `ui-state-error` + introduce `ui-state-valid` instead of `ui-state-csv-valid` + `ui-state-csv-invalid`)
I personally dislike the red borders we get this way at inital loading.
https://github.com/primefaces/primefaces/assets/10461942/d51121b5-8620-415d-a8d6-0c51b4b4677c

christophs78:
> @christophs78 but if its `ui-state-error` doesn't it have to have those borders?

Yep. From a UX-point-of-view i´d like to rollback the last commit. (back to `ui-state-csv-valid` + `ui-state-csv-invalid` and have more flexibility how to style this)
I hope it´s now transparent what our trade-off´s are. A bit of UX vs new/additional CSS-classes which may cause a bit of confusion elsewhere.)

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.commandbutton.js
```diff
@@ -23,6 +25,18 @@ PrimeFaces.widget.CommandButton = PrimeFaces.widget.BaseWidget.extend({
         PrimeFaces.skinButton(this.jq);
 
         this.bindTriggers();
+
+        if (cfg.validateClientDynamic) {
+            let that = this;
+
+            // init enabled/disabled-state after initial page-load
+            setTimeout( function() {
```
tandraschko: should we use the new task api @melloware introduced?

tandraschko:
so you would like to introduce a new style for csv-invalid and csv-valid? this means we need to change the themes and themecreator propably too

does PrimeVue has such a feature? can you compare style classes?

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.commandbutton.js
melloware: Yes please use `PrimeFaces.queueTask`

christophs78: Thanks for pointing me on this. Reminded there was some replacement for window.setTimeout but could not figure out myself what anymore.

christophs78:
> so you would like to introduce a new style for csv-invalid and csv-valid? this means we need to change the themes and themecreator propably too
> 
> does PrimeVue has such a feature? can you compare style classes?

The new css class(es) - ui-state-valid respective ui-state-csv-valid + ui-state-csv-invalid are IMO a additional offering which allow you to do some project-specific styling. Does not necessary mean we need to add something to themes and themecreator.

I´ll try to look what the other Prime... - libraries do.

christophs78:
What do other Prime-libraries?

- https://primeng.org/inputtext#invalid
- https://primevue.org/inputtext/#invalid

primefaces/src/main/resources/META-INF/resources/primefaces/validation/validation.common.js
```diff
@@ -511,6 +550,36 @@ if (window.PrimeFaces) {
...
+         * Bind to Ajax-Complete-events to update CSV-state after an Ajax-call may have changed state.
+         * @internal
+         */
+        bindAjaxComplete: function() {
+            if (this.ajaxCompleteBound) return;
+
+            var doc = $(document),
+                $this = this;
+
+            doc.on('pfAjaxComplete', function(e, xhr, settings, args) {
```
melloware:
```suggestion
            $(document).on('pfAjaxComplete', function(e, xhr, settings, args) {
```

No need to store `doc` variable as its not used anywhere else?

melloware (changes requested):

primefaces/src/main/resources/META-INF/resources/primefaces/validation/validation.common.js
christophs78: good catch - thanks!

primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/csv/Csv002Test.java
christophs78: `@AssertTrue` helps (partial) for SelectBooleanCheckbox.

christophs78:
Do we miss something on this PR?
When it helps i may provide another PR for the other CSS-variante. (Should be easy to provide this mutation.)
IMO both variants are backward-compatible. There are simple some new additional css classes which you may use in your project for custom styling when you want.

melloware (approved):

melloware:
If @tandraschko is ok with it we can get it in for the 14 RC1 release.

christophs78:
State after loading:
(ui-state-error kicks in immediately)
image

State after some inputs (ui-state-valid is added after successful CSV)
image

PR #11099, opened 2023-12-10
#10988 StaticMessage closable

### LINKED ISSUE #10988: StaticMessage: Add closable attribute
### Description

I have a static message taking up screen estate. When a user has read it, there is now no way to dismiss it.

Add `closable` attribute (and `onClose` event) to close (hide) the message.

jepsar:
Fixes #10988

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/staticmessage/StaticMessage.java
+++ b/primefaces/src/main/java/org/primefaces/component/staticmessage/StaticMessage.java
@@ -55,13 +55,13 @@ public class StaticMessage extends StaticMessageBase {
     }
 
     @Override
-    public void queueEvent(final FacesEvent event) {
+    public void queueEvent(FacesEvent event) {
         FacesContext context = getFacesContext();
         Map<String, String> params = context.getExternalContext().getRequestParameterMap();
         String eventName = params.get(Constants.RequestParams.PARTIAL_BEHAVIOR_EVENT_PARAM);
 
         if (eventName != null && event instanceof AjaxBehaviorEvent) {
-            final AjaxBehaviorEvent ajaxBehaviorEvent = (AjaxBehaviorEvent) event;
+            AjaxBehaviorEvent ajaxBehaviorEvent = (AjaxBehaviorEvent) event;
 
             if ("close".equals(eventName)) {
                 customEvents.put(eventName, ajaxBehaviorEvent);
--- a/primefaces/src/main/java/org/primefaces/component/staticmessage/StaticMessage.java
+++ b/primefaces/src/main/java/org/primefaces/component/staticmessage/StaticMessage.java
@@ -56,9 +56,9 @@ public class StaticMessage extends StaticMessageBase {
 
     @Override
     public void queueEvent(final FacesEvent event) {
-        final FacesContext context = getFacesContext();
-        final Map<String, String> params = context.getExternalContext().getRequestParameterMap();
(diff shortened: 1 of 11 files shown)


primefaces/src/main/java/org/primefaces/component/staticmessage/StaticMessage.java
```diff
@@ -23,11 +23,61 @@
...
+    }
+
+    @Override
+    public Collection<String> getUnobstrusiveEventNames() {
+        return getEventNames();
+    }
+
+    @Override
+    public void queueEvent(final FacesEvent event) {
+        final FacesContext context = getFacesContext();
```
tandraschko: please without "final", we dont use it in PF

primefaces/src/main/resources/META-INF/resources/primefaces/staticmessage/staticmessage.js
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * @override
+     * @inheritdoc
+     * @param {PrimeFaces.PartialWidgetCfg<TCfg>} cfg
+     */
+    init: function(cfg) {
+        this._super(cfg);
+        this.bindEvents();
+    },
+
```
melloware, line 22: We probably need ARIA attributes on the Close Icon letting the screen reader know its a close button?  role="button" and AriaLabel = close?   We should check other close buttons like in Sidebar?

melloware (changes requested):

primefaces/src/main/resources/META-INF/resources/primefaces/staticmessage/staticmessage.js
melloware:
```js
        if(this.cfg.showCloseIcon) {
            this.closeIcon.attr('role', 'button');
            this.closeIcon.attr('aria-label', PrimeFaces.getAriaLabel('close'));
        }
```

Is what sidebar.js does in the client side

primefaces/src/main/java/org/primefaces/component/staticmessage/StaticMessageBase.java
```diff
@@ -50,6 +59,14 @@ public String getFamily() {
         return COMPONENT_FAMILY;
     }
 
+    public String getWidgetVar() {
+        return (String) getStateHelper().eval(PropertyKeys.widgetVar, null);
+    }
+
+    public void setWidgetVar(final String widgetVar) {
```
tandraschko: same, still many final in some places

primefaces/src/main/resources/META-INF/resources/primefaces/staticmessage/staticmessage.js
jepsar: Hmm.. too bad that we have one file for front end and one for back end. Would be great if we had one file (backend messages) and transformed that to front end JSON by splitting keys on `.` or something.

melloware: We have moved 99% of ARIA label stuff to the UI client side where it belongs.  It makes translations easier and makes more sense to translate "on the glass" instead of half in the backend and half in the frontend.   it also aligns us with PrimeVue, PrimeReact etc

melloware (approved):

tandraschko:
looks good for me, +1

PR #11131, opened 2023-12-15
Fix #11130: Respect OS prefers-reduced-motion setting

### LINKED ISSUE #11130: Animation: Respect CSS `prefers-reduced-motion`
Using animations on the web in the most accessible way requires a little extra consideration:

In general, it's best to use animations on the web in a complementary way rather than the only way to note a state change in UI. Remember to update the text of actions and change elements' other visual aspects too!

And when it comes to an animation's timing and duration, avoid changing the state of something at a rate of three times per second or greater - which can trigger photosensitive epilepsy.

All included animations now support and leverage the prefers-reduced-motion [CSS media feature (opens new window)](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion)to detect if a user has requested that the system minimize the amount of non-essential motion it uses.

When a `prefers-reduced-motion` is set to `reduce`, it indicates that the user prefers less motion, which may trigger discomfort for those with vestibular motion disorders on the page.

image

melloware:
Fix #11130: Respect OS prefers-reduced-motion setting

### DIFF
--- a/docs/14_0_0/gettingstarted/whatsnew.md
+++ b/docs/14_0_0/gettingstarted/whatsnew.md
@@ -10,6 +10,7 @@ Look into [migration guide](https://primefaces.github.io/primefaces/14_0_0/#/../
 
 * Core
     * `MOVE_SCRIPTS_TO_BOTTOM` adds new option `defer` to defer loading scripts
+    * OS settings for `prefers-reduced-motion: reduce` is now respected and PF disables all animations
 
 * Accordion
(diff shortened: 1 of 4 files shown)


FlipWarthog (approved):

melloware:
@jepsar i will have to check if `disableOnAjax` spinner's respect this?   FontAwesome spinner's don't spin when this setting is on using a Media Query.

melloware:
Also I should add a note in What's New and Migration guide about this.

PR #11157, opened 2023-12-20
Message(s): add composite support (#3796)

### LINKED ISSUE #3796: Message(s): not working correctly with composite components
## 1) Environment
- PrimeFaces version: 6.2
- Application server + version: All
- Affected browsers: All

## 2) Expected behavior
Element p:message is properly printed for composite component with <cc:editableValueHolder name="input"/> .

## 3) Actual behavior

<p:message> is not printed

## 4) Steps to reproduce

Reproducer: https://github.com/kapitanrum/primefaces-test/tree/demo-label-input-messages

## 5) Sample XHTML

Using like:

````
  <p:outputLabel for="customInputTextTooltipHolder" value="Label for custom InputTextTooltipHolder" />
  <mycomp:inputTextTooltipHolder id="customInputTextTooltipHolder" value="#{testView.testInputCustomHolder}" tooltip="Test tooltip text holder" required="true" />
  <p:message for="customInputTextTooltipHolder" display="text" />

````

resources/mycomp/inputTextTooltipHolder.xhtml:
````
      xmlns:h="http://xmlns.jcp.org/jsf/html">

    <cc:interface>
        <cc:attribute name="value" type="java.lang.String" />
        <cc:attribute name="tooltip" type="java.lang.String" />        
        <cc:attribute name="required" type="java.lang.Boolean" />
        <cc:editableValueHolder name="input" />
    </cc:interface>

    <cc:implementation>
        <span id="#{cc.clientId}">
            <h:inputText id="input" value="#{cc.attrs.value}" required="#{cc.attrs.required}">
                <cc:insertChildren />
            </h:inputText>
            <p:tooltip for="#{cc.clientId}" value="#{cc.attrs.tooltip}" />
        </span>
    </cc:implementation>
</html>
````

primefaces-messages-issue

### PR DESCRIPTION
**Pull request by NicolaIsotta**
fixes #3796

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/message/MessageRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/message/MessageRenderer.java
@@ -28,15 +28,17 @@ import java.util.Iterator;
 import java.util.Objects;
 
 import javax.faces.application.FacesMessage;
+import javax.faces.component.ContextCallback;
 import javax.faces.component.UIComponent;
 import javax.faces.context.FacesContext;
 import javax.faces.context.ResponseWriter;
 
-import org.primefaces.component.api.InputHolder;
 import org.primefaces.context.PrimeApplicationContext;
 import org.primefaces.expression.SearchExpressionUtils;
 import org.primefaces.renderkit.UINotificationRenderer;
+import org.primefaces.util.CompositeUtils;
 import org.primefaces.util.HTML;
+import org.primefaces.util.StyleClassBuilder;
 import org.primefaces.util.WidgetBuilder;
 
 public class MessageRenderer extends UINotificationRenderer {
@@ -46,10 +48,22 @@ public class MessageRenderer extends UINotificationRenderer {
         Message uiMessage = (Message) component;
 
         UIComponent target = SearchExpressionUtils.contextlessResolveComponent(context, uiMessage, uiMessage.getFor());
-        String targetClientId = target.getClientId(context);
+        final MessageState state = new MessageState();
+        state.setTargetId(target.getClientId(context));
+        state.setTooltipTargetId(getTooltipTargetId(target, context));
 
-        encodeMarkup(context, uiMessage, targetClientId);
-        encodeScript(context, uiMessage, target);
+        if (!hasDisplayableMessage(uiMessage, target, context) && CompositeUtils.isComposite(target)) {
+            ContextCallback callback = (FacesContext fc, UIComponent comp) -> {
+                if (hasDisplayableMessage(uiMessage, comp, fc)) {
+                    state.setTargetId(comp.getClientId(fc));
+                    state.setTooltipTargetId(getTooltipTargetId(comp, fc));
+                }
+            };
+            CompositeUtils.invokeOnDeepestEditableValueHolder(context, target, callback);
+        }
+
+        encodeMarkup(context, uiMessage, state.getTargetId());
+        encodeScript(context, uiMessage, state.getTooltipTargetId());
     }
 
     protected void encodeMarkup(FacesContext context, Message uiMessage, String targetClientId) throws IOException {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: primefaces/src/main/java/org/primefaces/util/MessageState.java [hunk 1]
```diff
@@ -0,0 +1,47 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.util;
+
+public class MessageState {
```
melloware: MessageState should probably go right in `/component/message` like other components do? Or if we are going to use it for all three Growl, Messages, and Message then it should go in `/component/api` ?

melloware (changes requested):

primefaces/src/main/java/org/primefaces/util/MessageState.java
NicolaIsotta: I put it under `/util` like its inspiration, `EditableValueHolderState`, but it can safely be moved to `/component/message` (I do not think Messages/Growl are gonna use it).

melloware: Cool yeah usually the State's are located near their component.

primefaces/src/main/java/org/primefaces/component/message/MessageRenderer.java
```diff
@@ -46,20 +49,30 @@ public void encodeEnd(FacesContext context, UIComponent component) throws IOExce
...
 
     protected void encodeMarkup(FacesContext context, Message uiMessage, String targetClientId) throws IOException {
         ResponseWriter writer = context.getResponseWriter();
         String display = uiMessage.getDisplay();
         boolean iconOnly = "icon".equals(display);
         String style = uiMessage.getStyle();
-        String containerClass = "tooltip".equals(display) ? "ui-message ui-helper-hidden" : "ui-message";
-        String styleClass = uiMessage.getStyleClass();
-        styleClass = styleClass == null ? containerClass : styleClass + " " + containerClass;
+        StyleClassBuilder styleClassBuilder = new StyleClassBuilder(context)
```
tandraschko: just use #getStyleClassBuilder from CoreRenderer

primefaces/src/main/java/org/primefaces/component/message/MessageRenderer.java
```diff
@@ -46,20 +49,30 @@ public void encodeEnd(FacesContext context, UIComponent component) throws IOExce
...
-        String targetClientId = target.getClientId(context);
+        final MessageState state = new MessageState();
+        state.setTarget(target);
 
-        encodeMarkup(context, uiMessage, targetClientId);
-        encodeScript(context, uiMessage, target);
+        if (!hasDisplayableMessage(uiMessage, state.getTarget(), context) && CompositeUtils.isComposite(target)) {
+            ContextCallback callback = (FacesContext fc, UIComponent comp) -> {
+                if (hasDisplayableMessage(uiMessage, comp, fc)) {
+                    state.setTarget(comp);
```
tandraschko: in theory you should extract all required infos from the component here (targetClientId and inputClientId) as e.g. EL with #{cc} is not resolveable outside this callback anymore

melloware:
we will definitively need an integration test for this as well...I can help with that @NicolaIsotta

melloware (approved):

melloware:
Failing integration test which should pass when this is merged: https://github.com/primefaces/primefaces/pull/11168

Issue #7130, opened 2021-03-15
Datatable: lazy component isn't skipped when its parent has rendered property to false


oarenas-GI:
**Describe the defect**
It looks like lazy datatable isn't skipped when its parent has rendered property to false. When the page is loading, for some reason, it's calling savestate method from Datatable and fails. This error didn't appear on Primefaces 8.0.7.

**Environment:**
 - Server: Payara Server 5.2021.1 (community)
 - PF Version: 10.0
 - JSF + version: included in the specific Payara version
 - Affected browsers: ALL

**To Reproduce**
1. Run application server
2. Go to dummy-page.xhtml
3. See error
`Error Rendering View[/dummy-page.xhtml]
javax.el.ELException: /dummy-page.xhtml @13,63 rendered="#{dummyBean.checked}": java.lang.RuntimeException: Unexpected property read
	at com.sun.faces.facelets.el.TagValueExpression.getValue(TagValueExpression.java:77)
	at javax.faces.component.ComponentStateHelper.eval(ComponentStateHelper.java:170)
	at javax.faces.component.UIComponentBase.isRendered(UIComponentBase.java:313)
	at org.primefaces.component.api.ColumnAware.forEachColumn(ColumnAware.java:58)
	at org.primefaces.component.api.ColumnAware.forEachColumn(ColumnAware.java:52)
	at org.primefaces.component.api.ColumnAware.resetDynamicColumns(ColumnAware.java:315)
	at org.primefaces.component.datatable.DataTable.saveState(DataTable.java:935)
	at com.sun.faces.application.view.FaceletPartialStateManagementStrategy$3.visit(FaceletPartialStateManagementStrategy.java:475)
	at com.sun.faces.component.visit.FullVisitContext.invokeVisitCallback(FullVisitContext.java:127)
	at org.primefaces.component.api.UIData.visitTree(UIData.java:691)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1468)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1468)
	at javax.faces.component.UIForm.visitTree(UIForm.java:347)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1468)
	at javax.faces.component.UIComponent.visitTree(UIComponent.java:1468)
	at com.sun.faces.application.view.FaceletPartialStateManagementStrategy.saveView(FaceletPartialStateManagementStrategy.java:464)
[...]

PR #11167, opened 2023-12-21
Fix #10639: UIData do not visit unrendered column

### LINKED ISSUE #10639: DataTable: Value of component  in non rendered column gets evaluated after ajax update.
### Describe the bug

As the title says, component which should not be rendered gets evaluated after ajax update of parent DataTable. Problem lies in  `org.primefaces.component.api.UIData.visitRows` and in `org.primefaces.component.api.UIData.visitColumnContent` .  `org.primefaces.component.api.UIData.visitColumnContent`  gets called on a child component whether or not the column `isVisitable`.

### Reproducer

[primefaces-test-master-nonjakarta.zip](https://github.com/primefaces/primefaces/files/12764091/primefaces-test-master-nonjakarta.zip)

### Expected behavior

Expected behavior would be that child component does not get evaluated if the column is not rendered

### PrimeFaces edition

Community

### PrimeFaces version

13.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0.4

### Java version

Adoptium Temurin jdk-17.0.8.1+1

### Browser(s)

_No response_

melloware:
Fix #10639: UIData do not visit unrendered column

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/api/UIData.java
+++ b/primefaces/src/main/java/org/primefaces/component/api/UIData.java
@@ -881,9 +881,12 @@ public class UIData extends javax.faces.component.UIData {
     }
 
     protected boolean visitColumnContent(VisitContext context, VisitCallback callback, UIComponent component) {
-        if (component.getChildCount() > 0) {
+        if (component.getChildCount() > 0 && component.isRendered()) {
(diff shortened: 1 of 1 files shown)


melloware:
Fails the IT written for: https://github.com/primefaces/primefaces/issues/7130

PR #11177, opened 2023-12-23
Refactor more FacetUtils

melloware:


tandraschko:
Its actually both "component" and "facet"
So not sure where to place it

melloware:
Yeah I saw that too but since it was facet related it felt like we could declutter the Component Utils some more?

tandraschko:
Not sure
It feels more related to common components
Not just exclusive to facets

melloware:
No problem I will close thanks for checking!

Issue #99, opened 2015-04-19
Non-Ajax submit support for Global ConfirmDialog


cagataycivici:
Currently using `ajax=false` with Confirm Dialog does not work.
```xml
<p:commandButton id="delete" 
                 action="#{trafficExpenseItemsMBean.deleteExpenseItemsGroup}" 
                 icon="ui-icon ui-icon-trash" 
                 value="Delete" 
                 ajax="false"
                 title="DeleteButton">
    <p:confirm header="Delete Record" 
               message="Are you sure about deleting this record?" 
               icon="ui-icon-alert"/>
</p:commandButton>
```

zlaoz:
workaround for commandButtons, setting ajax=true and reload page via oncomplete=window.location.replace():

``` html
<p:commandButton id="delete"
    value="Delete"
    actionListener="#{advertHome.remove}"
    rendered="#{advertHome.managed}"
    immediate="true"
    ajax="true"
    oncomplete="window.location.replace(window.location.href);">

    <p:confirm icon="ui-icon-alert" message="Delete?" header="!"/>                         
</p:commandButton>
```

Rapster:
@cagataycivici could you give us some clues about this one, I'm facing this problem and would like to fix this. we're using `confirmDialog` with global option and a lot of non-ajax action

Rapster:
@tandraschko Do you have a clue how to fix this? I must say I don't know why ajax is not working, `pfconfirmcommand` looks good to me...

tandraschko:
No idea sry, i would have to check the implementation details of p:confirm.

PR #11178, opened 2023-12-24
#10258 Minimum Ajax load animation duration

### LINKED ISSUE #10258: CommandButton/etc: ajaxLoadingIndicator should have a minimum display time
Currently if you click on it, and the request e.g. just takes 50ms, you just see a weird flickering

what about setting a minimum display like 500ms?

- [ ] Minimum display so it shows either for the minimum duration or the full duration if it is longer than the minimum
- [ ] Delay duration where say a 100ms delay is set it won't show the spinner until 100ms and checks if `AjaxQueue.isEmpty() to see if the AJAX already finished to prevent short displays or flickering.

jepsar:
Fixes #10258

I was not sure how and where to define the minimum number of milliseconds to show Ajax load animations.


primefaces/src/main/resources/META-INF/resources/primefaces/core/core.js
```diff
@@ -421,19 +421,29 @@
...
-                    button.removeClass('ui-state-loading');
-
-                    if (widget.hasOwnProperty('enable')
-                        && widget.cfg.disableOnAjax !== false
-                        && !widget.cfg.disabledAttr) {
-                        widget.enable();
-                    }
-
-                    button.find('.ui-icon-loading').remove();
+                    setTimeout(PrimeFaces.buttonEndAjaxDisabled,
```
melloware:
```suggestion
                    PrimeFaces.queueTask(PrimeFaces.buttonEndAjaxDisabled,
```

melloware (changes requested):

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.js
```diff
@@ -400,9 +400,9 @@
                     }
 
                     button.addClass('ui-state-loading');
+                    button.data('astart', Date.now());
```
melloware: Should `astart` be a more recognizable name like `ajaxstart` or something?

jepsar:
This should also fix a regression introduced with #10999

melloware (changes requested):

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.js
jepsar: How can you pass parameters using this? I'm now getting `Uncaught TypeError: Cannot read properties of undefined (reading 'removeClass')`

jepsar: Alright, I think I've got it... @melloware please have a look

melloware (changes requested):
This looks much better!  Looks like TypScript issue with args

```
TypedError: [tagOptionalForRest] Parameter args was marked as optional, but code specifies this as a rest parameter (...args) and such arguments cannot be optional.
```

tandraschko:
Build failing

I will review next week

jepsar:
I think the minimum delay (of 500ms) is causing some tests to fail. I think I'll need to update those tests. Executing this JS before clicking buttons should fix it.
```js
PrimeFaces.ajax.minLoadAnim = 0;
```

melloware:
@jepsar in the Integration tests I have 1 spot where I disableAnimations you could just put the code there to lower the timeout.

tandraschko:
Why does it cause failing tests?
Do we have tests which verifies the loading indicator

jepsar:
> Why does it cause failing tests? Do we have tests which verifies the loading indicator

ConfirmDialog tests checked if buttons were enabled while they were not (yet) due to the minimum loading time.

I'll add a test for the loading indicator.

jepsar:
> @jepsar in the Integration tests I have 1 spot where I disableAnimations you could just put the code there to lower the timeout.

This is not really an animation, so I don't think it should be controlled by that function.

jepsar:
@melloware WDYT, show we mention `PrimeFaces.ajax.minLoadAnim` in the documentation somewhere?

primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.commandlink.js
```diff
@@ -56,6 +57,7 @@ PrimeFaces.widget.CommandLink = PrimeFaces.widget.BaseWidget.extend({
                 }
                 if (PrimeFaces.ajax.Utils.isXhrSource($this, settings)) {
                     $this.jq.addClass('ui-state-loading');
+                    $this.ajaxStart = Date.now();
```
melloware, line 60: Should this value be cleared somewhere once its no longer used so the value in memory is not misleading?

melloware (changes requested):

melloware:
I feel like in the AJAX section you could mention it somewhere like here: https://primefaces.github.io/primefaces/13_0_0/#/core/ajaxJavascriptApi

melloware (approved):

jepsar:
> I feel like in the AJAX section you could mention it somewhere like here: https://primefaces.github.io/primefaces/13_0_0/#/core/ajaxJavascriptApi

I also need to add this to what's new!

primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/confirmdialog/ConfirmDialog001Test.java
```diff
@@ -47,6 +47,7 @@ class ConfirmDialog001Test extends AbstractPrimePageTest {
     void showDialog(Page page) {
         // Arrange
         ConfirmDialog dialog = page.dialog;
+        noMinLoadAnim();
```
tandraschko: Why is this required? Makes the tests a bit hacky

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
```diff
@@ -86,6 +86,12 @@ if (!PrimeFaces.ajax) {
             'onco': 'oncomplete'
         },
 
+        /**
+         * Minimum number of milliseconds to show Ajax load animations.
+         * @type {number}
+         */
+        minLoadAnim : 500,
```
tandraschko: lets use "Animation" instead of "Anim", no need to shorten it actually
min/max is common

primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/confirmdialog/ConfirmDialog001Test.java
jepsar: ConfirmDialog tests checked if buttons were enabled while they were not (yet) due to the minimum loading time.

tandraschko: you mean the assertClickable(page.confirm);?

maybe we should instead improve this and check the animination state and wait for it

jepsar: Alright. I'll also add some additional tests to the IT I've added.

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
jepsar: Will do.

melloware:
merged my Confirm Dialog changes.  @jepsar when you test locally can you uncomment out the two lines in the test 

```java
        // Assert
        //PrimeSelenium.waitGui().until(PrimeExpectedConditions.visibleInViewport(page.message));
        //assertEquals("Full page submitted", page.message.getMessage(0).getDetail());
```

This passes for me fine locally but fails on GitHub actions I just want to see if it passes for you also.

jepsar:
> This passes for me fine locally but fails on GitHub actions I just want to see if it passes for you also.

@melloware Yeah, all tests work for me locally.

melloware:
OK comment those lines back out please this may be a debug job for @christophs78 !

melloware (approved):

jepsar:
I don't get why it's failing on one instance. Race condition?

christophs78:
Sometimes it helps to add something like `@RepeatedTest(100)` to make a issue kind of reproduceable.
Additional we create since a few days screenshots on Github Actions whenever a IT fails.
image

I´ll try to look into into. Maybe a have to continue next week when it takes more time.

christophs78:
> merged my Confirm Dialog changes. @jepsar when you test locally can you uncomment out the two lines in the test
> 
> ```java
>         // Assert
>         //PrimeSelenium.waitGui().until(PrimeExpectedConditions.visibleInViewport(page.message));
>         //assertEquals("Full page submitted", page.message.getMessage(0).getDetail());
> ```
> 
> This passes for me fine locally but fails on GitHub actions I just want to see if it passes for you also.

@melloware 
Your fix for https://github.com/primefaces/primefaces/issues/99, PR https://github.com/primefaces/primefaces/pull/11228 works only with CSP disabled. (CI enables CSP, Showcase has it disabled OOTB)
Spent a bit too much time on it because i looked into other things first. But thats the reason.

christophs78:
> I don't get why it's failing on one instance. Race condition? Maybe invert the if conditions...

Looking into this the next few days.

melloware:
Thanks @christophs78 i will debug with CSP on.

melloware:
@jepsar i just updated your branch with my latest change that fix the tests for my issue #99

PR #11186, opened 2023-12-27
Fix #11044 - FileUpload: move server side validation to a own p:validateFile

### LINKED ISSUE #11044: FileUpload: move server side validation to a own p:validateFile
and reuse this on the clientSide, too

#11001

this can even be reused for h:inputFile

tandraschko:
Fix #11044


tandraschko:
This duplicates the validator now in a own p:validateFile. Old attributes on p:fileUpload are deprecated, can be removed later. For now both are working.

Some notes:
1) The only downside now is that i removed e.g. fileSizeMessage. Do we really need it?
2) advanced fileupload doesnt reuse this CSV logic for now
3) i also fixed the showcase search that "file" now finds all file upload samples, too

melloware (approved):

melloware:
Looks good.. do we need a Migration Guide note for anything?

primefaces/src/main/java/org/primefaces/component/fileupload/FileUploadRenderer.java
```diff
@@ -328,34 +327,31 @@ protected void encodeInputField(FacesContext context, FileUpload fileUpload, Str
...
-                public String getValidatorId() {
-                    return "primefaces.File";
+                long sizeLimit = fileUpload.getSizeLimit();
+                if (sizeLimit != Long.MAX_VALUE) {
+                    fileValidator.setSizeLimit(sizeLimit);
                 }
-            });
+
+                String allowTypes = fileUpload.getAllowTypes();
+                if (!StringUtil.isBlank(allowTypes)) {
```
Rapster: Where this StringUtil comes from?

primefaces/src/main/java/org/primefaces/validate/FileValidator.java
```diff
@@ -0,0 +1,323 @@
...
+                UploadedFile uploadedFile = (UploadedFile) value;
+
+                validateUploadedFile(context, fileUpload, uploadedFile);
+            }
+            else if (value instanceof UploadedFiles) {
+                UploadedFiles uploadedFiles = (UploadedFiles) value;
+
+                validateUploadedFiles(context, fileUpload, uploadedFiles);
+            }
+        }
```
Rapster, line 106: throw exception if not FileUpload?

primefaces/src/main/java/org/primefaces/component/fileupload/FileUpload.java
```diff
@@ -80,7 +83,9 @@ public void broadcast(FacesEvent event) throws AbortProcessingException {
     protected void validateValue(FacesContext context, Object newValue) {
         super.validateValue(context, newValue);
 
-        if (isValid() && ComponentUtils.isRequestSource(this, context)) {
+        boolean hasFileValidator = Arrays.stream(getValidators()).anyMatch(v -> v instanceof FileValidator);
+
+        if (isValid() && ComponentUtils.isRequestSource(this, context) && !hasFileValidator) {
```
Rapster: `!hasFileValidator` should come first in the condition

primefaces/src/main/java/org/primefaces/component/fileupload/FileUploadBase.java
```diff
@@ -274,8 +274,7 @@ public void setOncomplete(String oncomplete) {
     }
 
     public String getOnvalidationfailure() {
-        Object eval = getStateHelper().eval(PropertyKeys.onvalidationfailure, null);
-        return (String) eval;
+        return (String) getStateHelper().eval(PropertyKeys.onvalidationfailure, null);
```
Rapster, line 277: Is it normal that `onvalidationfailure` is not camel case?

Rapster (commented):
Quick review

primefaces/src/main/java/org/primefaces/component/fileupload/FileUploadBase.java
tandraschko: not really :/ but not my task currently, just make the getter look like all our getters

primefaces/src/main/java/org/primefaces/validate/FileValidator.java
tandraschko: i will try to implement h:inputFile support, otherwise will do it !

tandraschko:
fixed the feedback and added h:inputFile support
what shall we do with the accept attribute and validateContentType? 
it would be enough to add a accept attr to the validator; if set -> validateContentType, otherwise skip.
the downside is just that a user would need to add accept to p:validateFile and the input.

tandraschko:
and should we drop fileLimitMessage etc? no JSF builtin validator supports this.

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
```diff
@@ -237,16 +237,10 @@ protected static String convertJavaScriptRegex(String jsRegex) {
         return LangUtils.substring(jsRegex, start, end);
     }
 
-    private static boolean isValidFileContent(PrimeApplicationContext primeAppContext, FileUpload fileUpload, String fileName,
+    private static boolean isValidFileContent(PrimeApplicationContext primeAppContext, String accept, String fileName,
```
Rapster: acceptTypes

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
```diff
@@ -284,22 +278,22 @@ private static boolean isValidFileContent(PrimeApplicationContext primeAppContex
...
         //Comma-separated values: file_extension|audio/*|video/*|image/*|media_type (see https://www.w3schools.com/tags/att_input_accept.asp)
-        String[] accepts = fileUpload.getAccept().split(",");
+        String[] accepts = accept.split(",");
         final String filenameLC = fileName.toLowerCase();
         final String contentTypeLC = contentType.toLowerCase();
 
         boolean accepted = Stream.of(accepts)
                 .map(String::trim)
-                .anyMatch(accept -> {
+                .anyMatch(acceptPart -> {
```
Rapster: acceptType

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
tandraschko: Not sure about this one
Its just called accept in HTML and on p:fileUpload/h:inputFile, we should propably stick to it.

Rapster: It's a lot more explicit than just accept, it's also important to show plurarity

tandraschko: how is it more then just accept?
Its a list of allowed content-types, same like the HTML accept attribute.

tandraschko:
removed the messages attributes, added docs and migration guide.
also introduced the accept attribute on p:validateFile, to enable validation of it.

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
tandraschko: also for me its very important to stick with native HTML names; just think about div vs h:outputPanel

Rapster: Exactly, it's a list... based on what you modify after, how do you make a difference between acceptPart and accept? That's not really clear at that point... That's why I would emphase plurality with acceptTypes, and then use acceptType to emphase it is one accept-chunk from accepTypes string

tandraschko:
i'm happy now with: `p:validateFile contentType=true`
and we try to extract the accept attr from the attached component

melloware (approved):

docs/migrationguide/14_0_0.md
```diff
@@ -22,9 +22,10 @@
   * Locale `firstDay` renamed `firstDayOfWeek` to match other Prime libraries
 
 ## FileUpload
-### mode='simple'
-  * filename won't be displayed per default when auto='true'. Please set displayFilename='true' if you need it.
-  * validation won't be displayed right after the button anymore. The validation has been reimplemented with PrimeFaces CSV and support for `p:message(s)` and `p:growl`.
+  * `invalidSizeMessage`, `invalidFileMessage`, `fileLimitMessage` are deprecated and will be removed in 15.0.0. PF i18n will be used for it.
+  * `maxFileSize`, `fileLimit`, `allowedTypes` are deprecated and will be removed in 15.0.0. Please attach `p:validateFile` to the `p:fileUpload` component.
+  * Simple FileUpload: Client side validation has been reimplemented with PrimeFaces CSV and support for `p:message(s)`/`p:growl`. Please make sure to enable CSV to use it.
+  * Simple FileUpload: filename won't be displayed per default when `auto`=`true`. Please set `displayFilename`=`true` if you need it.
```
Rapster: filename is not displayed by default when `auto`=`true`, set `displayFilename`=`true` instead if you need it.

docs/14_0_0/validators/validateFile.md
```diff
@@ -0,0 +1,35 @@
...
+|------------------|-----------------------------------------------|
+| Tag              | validateFile                                  |
+| Validator Class  | org.primefaces.validate.FileValidator         |
+| Validator Id     | primefaces.File                               |
+
+## Attributes
+
+| Name           | Default | Type    | Description     |
+|----------------| ------- |---------| ----------------- |
+| allowTypes     | null    | String  | Regular expression for accepted file types, e.g. /(\.|\/)(gif|jpe?g|png)$/
```
Rapster: worth mentionning it's a javascript regex (at it is server validator, people would expect java regex) Being said, it doesn't make a lot of sense if people use this for `h:inputFile` If I remember correctly, it's a js regex because of FileUpload widget

melloware:
@Rapster I fixed your textual issues mentioned above as well as a few minor English corrections.

PR #11210, opened 2024-01-01
Fix #3551: Datatable remember multiple selection between AJAX updates

### LINKED ISSUE #3551: DataTable: Multiselection with shift-click doesn't work after update

melloware:
Fix #3551: Datatable remember multiple selection between AJAX updates

This PR stores the `rowMeta` and not just the `index` of the rows selected between AJAX updates.  This allows a lookup by the `rowKey` to find the rows selected again or if they have changed (aka removed) indexes in between AJAX updates it can still find the correct rows by row key.

- Store `rowMeta` with `rowKey` instead of `rowIndex`
- After AJAX update if these have values lookup the new row by its `rowKey`
- Select rows that were selected before the AJAX update

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
@@ -68,7 +68,7 @@
  * @prop {JQuery} [contextMenuCell] DOM element of the table cell for which the context menu was opened.
  * @prop {PrimeFaces.widget.ContextMenu} contextMenuWidget Widget with the context menu for the DataTable.
  * @prop {JQuery} currentCell Current cell to be edited.
- * @prop {number | null} cursorIndex 0-based index of row where the the cursor is located.
+ * @prop {PrimeFaces.widget.DataTable.RowMeta | null} cursorRowMeta 0-based index of row where the the cursor is located.
  * @prop {string} [descMessage] Localized message for sorting a column in descending order.
  * @prop {JQuery} dragIndicatorBottom DOM element of the icon that indicates a column is draggable.
  * @prop {JQuery} dragIndicatorTop DOM element of the icon that indicates a column is draggable.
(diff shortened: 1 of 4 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
```diff
@@ -797,17 +797,33 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
...
             $this.updateSelectionAria($(this))
         });
 
         this.bindSelectionEvents();
+        
+        //shift key based range selection
+        this.originRowMeta = this.originRowMeta ? this.getRowMeta(this.tbody.find("[data-rk='" + this.originRowMeta.key + "']")) : null;
+        if (this.cursorRowMeta) {
+            // #3551 lookup row after update by row key
+            var crow = this.tbody.find("[data-rk='" + this.cursorRowMeta.key + "']");
```
tandraschko: crow? is this a common name for vars? :D

tandraschko:
could we add a test here?

melloware:
Yes let me add a test!

melloware:
Integration test added.  I confirmed the test fails without this change and passes with this change.

PR #11213, opened 2024-01-02
Fix #11211: CLIENT_SIDE_LOCALISATION true by default

### LINKED ISSUE #11211: enable CLIENT_SIDE_LOCALISATION per default
I think it would be a good default behavior to add it

users with custom i18n JS can still deactive it

### LINKED ISSUE #11212: CSV: log unrendered FacesMessages
So users can see if they did forget a update statement

melloware:
Fix #11211: CLIENT_SIDE_LOCALISATION true by default

### DIFF
--- a/primefaces-showcase/src/main/webapp/WEB-INF/web.xml
+++ b/primefaces-showcase/src/main/webapp/WEB-INF/web.xml
@@ -51,10 +51,6 @@
         <param-value>true</param-value>
     </context-param>
     <context-param>
-        <param-name>primefaces.FONT_AWESOME</param-name>
-        <param-value>true</param-value>
-    </context-param>
-    <context-param>
         <param-name>primefaces.THEME</param-name>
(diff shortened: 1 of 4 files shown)


primefaces-showcase/src/main/webapp/WEB-INF/web.xml
```diff
@@ -50,10 +50,6 @@
         <param-name>primefaces.CLIENT_SIDE_VALIDATION</param-name>
         <param-value>true</param-value>
     </context-param>
-    <context-param>
-        <param-name>primefaces.FONT_AWESOME</param-name>
-        <param-value>true</param-value>
-    </context-param>
```
melloware, line 56: This was a useless prop as its been removed for a while now.

tandraschko:
i think its even used in integration tests and can be removed?

melloware:
You are right i removed it there too.

PR #11216, opened 2024-01-02
Fix #3210: Dialog reimplement minimize

### LINKED ISSUE #3206: Editor: is disabled and data is clear after minimizing the dialog
I have `p:editor` (also try `pe:ckEditor`) inside my dialog. when I minimize and restore the dialog again, the data in editor is gone and editor is disabled. 

I have tried both primefaces editor and primefaces ckEditor but facing same issue on both.

## 1) Environment

- primefaces-6.1

- It does not work on Primeface showcase as well

- primefaces-extensions-6.1.1

- jsf-2.2

## 2) Expected behavior

After restore dialog editor data not clear and editor should not disable.

## 3) Actual behavior

After restore dialog editor data is clear and editor is disable. Instead of editor data display 'true' like below screenshot (Using Firefox)
image

## 4) Steps to reproduce

Open Dialog add some data in editor and minimize dialog then again maximize that dialog.

## 5) Sample XHTML
```
 <p:dialog minimizable="true" modal="false" id="newDlg" widgetVar="dlgNew">
        	<p:editor value="test"/>
 </p:dialog>
```
```
<p:dialog closeOnEscape="true" responsive="true" widgetVar="dlg" id="dlg1" resizable="false" draggable="true" closable="true" dynamic="true" modal="false"  width="97%" minimizable="true" appendTo="@(body)" header="Dialog" onShow="PF('dlg').initPosition()">
   <pe:ckEditor toolbar="#{config['ckEditor.toolBar']}" id="body" widgetVar="editor" readonly="false" width="100%" maxlength="50000" value="#{myBean.textVal}" />
</p:dialog>
```

### LINKED ISSUE #3210: Dialog: Reimplement minimize
Currently minimize moves the DOM and therefore scripts maybe be detached from their DOM elements.
I think it's better to not move the DOM, just hide the dialog and add placeholder to the dock.

https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/522
https://github.com/primefaces/primefaces/issues/3206

melloware:
Fix #3210: Dialog reimplement minimize
Fix #3206

- Clone the dialog and remove everything but the title bar
- `hide()` the actual dialog on minimize and put this clone in the docking area
- On restore destroy the clone and just `show()` the original dialog again

This fixes all issues related to the DOM being detached and re-attached and losing event handlers like Editor etc.

### DIFF
--- a/primefaces-integration-tests/src/main/webapp/dialog/dialog004.xhtml
+++ b/primefaces-integration-tests/src/main/webapp/dialog/dialog004.xhtml
@@ -1,0 +1,36 @@
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml"
+      xmlns:h="http://xmlns.jcp.org/jsf/html"
+      xmlns:f="http://xmlns.jcp.org/jsf/core"
+      xmlns:p="http://primefaces.org/ui">
+
+<f:view contentType="text/html;charset=UTF-8" encoding="UTF-8">
+    <h:head>
+
+    </h:head>
+
+    <h:body>
+
+        <h:form id="form">
+            <p:messages id="msgs" showDetail="true">
+                <p:autoUpdate />
+            </p:messages>
+
+			<p:dialog header="Minimizable" id="dlg" widgetVar="dlg"
+				minHeight="40" width="350" showEffect="fade" minimizable="true"
+				maximizable="true">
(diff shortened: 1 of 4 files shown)


tandraschko:
Very nice! +1 if well tested

melloware:
I added integration tests and I am happy with how this came out!

PR #11217, opened 2024-01-02
Fix #10689: Organigram drop tolerance

### LINKED ISSUE #10689: Organigram: Drag´n´Drop-issues with zoom-out
### Describe the bug

When using zoom out selecting the target node does not work correct.

### Reproducer

1. Go to https://primefaces.org/showcase/ui/data/organigram.xhtm
2. Activate zoom
3. "Zoom out" a few times (eg 4 times)
4. Move eg "Lisa Boehm" over somewhere between "Backend" and "Android"
image
5. You see both ("Backend" and "Android") get highlighted. 
Also two drag´n´drop-AJAX events are triggered. One for each target.
image

### Expected behavior

Same behaviour as with 100% (default) zoom. Just select one target node, fire one drag´n´drop- AJAX-event.

### PrimeFaces edition

Community

### PrimeFaces version

13.0.1

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

17

### Browser(s)

_No response_

melloware:
Fix #10689: Organigram drop tolerance

@christophs78 this fixes it for me and allows only 1 item to be dropped on even if zoomed.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/jquery/extensions.txt
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/jquery/extensions.txt
@@ -32,3 +32,4 @@ jquery.ui.pfextensions:
 - Datepicker style fix (https://github.com/primefaces/primefaces/issues/5374)
 - Slider step min max fix (https://github.com/primefaces/primefaces/issues/8619)
 - Tooltip removal fix (https://github.com/primefaces/primefaces/issues/9941)
+- Droppable zoom factor handling (https://github.com/primefaces/primefaces/issues/10689)
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.js
@@ -119,3 +119,56 @@ $.widget( "ui.sortable", $.ui.sortable, {
         return orig.apply(this, arguments);
     }
 })();
+
+
+// GitHub #10689 allow drag and drop to deal with zoom factor
+(function() {
+    $.ui.ddmanager.prepareOffsets = function(t, event) {
+        var i, j,
+            m = $.ui.ddmanager.droppables[t.options.scope] || [],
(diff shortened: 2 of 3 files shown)


christophs78:
Sorry to say, but the issue is still here.
image

You tried probably exactly the same thing i tried a few months ago.

As described in the issue both (target-)boxes light and an 2 ajax-call´s are triggered. (Just in the showcase-example nothing bad happens and the second one wins. Is a bit more complicated in our real-world-usecase.)

melloware:
Hmmm I can't reproduce.  Only 1 element whichever the mouse pointer is over receives the droppable?

The default "intersect" is broken because when zoomed out its easy for the droppable to take 50% of each square?  But I can't seem to replicate it with "pointer"?

christophs78:
Hm. Issue happens on my side with Edge, Chrome and Firefox.

Did a short video of this:
https://github.com/primefaces/primefaces/assets/10461942/8a212e34-d9b6-495c-8f28-5d83b5c61e4f

melloware:
Jquery Bug: https://bugs.jqueryui.com/ticket/6844/

Jquery Bug: https://bugs.jquery.com/ticket/11114/

christophs78:
Yep. A (very) hard issue. Maybe we even would need to reimplement this component without jQuery or ...
Or implement zoom in another way. (Maybe change fontsizes or ... instead of CSS-scale.)
From my current feeling maybe out of reach with a limited time-budget.

melloware:
I think I found a solution trying it out now.

christophs78:
I am excited :-)

melloware:
OK give this PR a shot locally.  I found multiple stack overflow posts that gave me a hint I needed that we needed to plug into the DDManager which is where that hit logic is being calculated.

christophs78:
Works.

christophs78 (approved):
Still excited you found a way to solve this. Thanks for your work!

melloware:
I have to pat myself on the back for finding a way to inject the zoomfactor per droppable.  Stack Overflow the guy was using a global windows `window.zoomFactor` but that would mean all components would have to use the same zoomFactor!

PR #11223, opened 2024-01-03
Fix #9798: CellEditEvent access whole row/node edited

### LINKED ISSUE #9798: CellEdit : rowKey not set without selectionMode
### Describe the bug

if selectionMode isn't set, rowKey is null in the celledit event. rowKey attribute is set in dataTable tag.

### Reproducer

add a cellEditor with an inputText to a dateable without selectionMode and without a selection checkbox Column. Editor shows up, cellEdit is called with newValue and oldValue, but rowKey is null. 

### Expected behavior

rowKey with correct value, i.E. #{ entry.id }

### PrimeFaces edition

Community

### PrimeFaces version

12.0.0-RC3

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

11

### Browser(s)

_No response_

melloware:
Fix #9798: CellEditEvent access whole row/node edited

This PR solves the issue by allowing access to the whole row "lazily" from the CellEdit event if you want to get say the primary key of the row being edited etc.

### DIFF
--- a/docs/14_0_0/gettingstarted/whatsnew.md
+++ b/docs/14_0_0/gettingstarted/whatsnew.md
@@ -28,6 +28,7 @@ Look into [migration guide](https://primefaces.github.io/primefaces/14_0_0/#/../
     * JPALazyDataModel now supports wildcard filters with `setWildcardSupport(true);` so you can use `*`, `%`, `_` or `?` in filter
     * JPALazyDataModel now supports builder pattern for constructor.
     * Added `filterPlaceholder` for `Column` and `Columns`
+    * Added `rowData` to `CellEditEvent` which contains the entire row from the cell being edited.
 
 * Messages
     * Added `clearMessages` widget method to clear all current messages.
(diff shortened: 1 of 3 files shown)


Rapster:
I'm not a big fan of having these lazy resolve methods in events, events should be immutable and frozen in time. Why not creating these events with the rowData (I believe that's why i mentionned rowKey should be depracated here) ? Also, whenever you're manipulating the rowIndex, you ideally should always reset to either the previous rowindex or -1 (I believe in this case it should be -1), we encoutered few issues in the past because of this

melloware:
@Rapster i was following the `newValue` example in that only evaluate it if someone asks for it.  But i could easily do it like `oldValue` is done on creation of the event.

Really its no different than what was being done in the workaround.

```java
public void onCellEdit(CellEditEvent event) {
    DataTable dataTable = (DataTable) event.getComponent();
    Entity entity = (Entity) dataTable.getRowData();
    ...
}
```

Instead of having this be undocumented feature its a first class citizen on the event.  But I can change it to not be lazy if you want?

PR #11224, opened 2024-01-03
Fixes #7390 - FileUpload: pass files count in FileUploadEvent

### LINKED ISSUE #7390: FileUpload: pass files count in FileUploadEvent
**Description**
If both advanced mode and multiple is set for FileUpload component, every file is uploaded separately via handleFileUpload(FileUploadEvent event)
However, it is not easy to determine which one is the last to trigger some subsequent action.

Currently I utilize remoteCommand, which passes the number of uploaded files to backend via FleUploaded onstart event.

```
<h:form id="upload" enctype="multipart/form-data">

    <p:remoteCommand name="setFilesCount" action="#{bean.setFilesCount}" />

    <p:fileUpload
                  mode="advanced"
                  multiple="true"
                  listener="#{bean.handleFileUpload}"
                  widgetVar="fileUploader"
                  onstart="setFilesCount([{name:'filesCount', value: PF('fileUploader').files.length}]);"/>
```
Then in the backend I check if the count matches and if so, I perform some processing
```
    public void handleFileUpload(FileUploadEvent event) throws IOException {

        UploadedFile file = event.getFile();
        fileStreamMap.put(file.getFileName(), file.getInputStream());
        
        if (fileStreamMap.size() == filesCount) {
```

I can imagine that total count could be passed via that FileUploadEvent, so instead of helper filesCount variable set via remoteCommand I could simply use:
 ```
    public void handleFileUpload(FileUploadEvent event) throws IOException {

        UploadedFile file = event.getFile();
        fileStreamMap.put(file.getFileName(), file.getInputStream());
        
        if (fileStreamMap.size() == event.getFilesCount()) {
 ```

tandraschko:
Fixes #7390


melloware (approved):
Nice!

PR #11225, opened 2024-01-03
Fix #8475: Allow SHIFT range checkbox selection

### LINKED ISSUE #8475: DataTable: selectionMode="multiple" - Selection with shift doesn't work when clicking into a checkbox
Description:
Multi row selection with the shift key only selects the last checkbox, when clicking the checkbox of a row.

Environment:
PF ShowCase - DataTable: Selection
https://www.primefaces.org/showcase/ui/data/datatable/selection.xhtml?jfwid=f4c98

1. In the ShowCase for the  dataTable selection go to the checkbox.
2. Click on the checkbox of the first regular data row.
3. Hold down Shift and click on the checkbox of the last datarow.
4. Only the  first and the last checkbox are checked. Text in row(1) - row(n-1) is marked as selected.

Expected behavior:
All rows and checkboxes are selected/checked

melloware:
Fix #8475: Allow SHIFT range checkbox selection


tandraschko:
+1 if well tested

PR #11226, opened 2024-01-03
Fix #4725: Column accessible header text

### LINKED ISSUE #4725: Datatable: empty `TH` tags should be `TD`
Hello,

For accessibility purpose (WCAG2.1 rule 1.3.1), the HTML tables should have a not empty `th` tags to make the column headers visible for screen reader.

PrimeFaces generates nicely the `th` but there is an exception. If a column contains only technical stuff (s.t. icons), its headerText could be empty. In that case, the following code is generated by `<p:column>` with PrimeFaces 7.0 :
```
<th id="myForm:table:j_idt102" class="ui-state-default" role="columnheader" scope="col">
        <span class="ui-column-title"></span>
</th>
```

As said previously, the `th` can't be empty. So when the `headerText/f:facet header` is blank, is it possible for PrimeFaces to generate a `td` instead of the `th` (only for that column) ?

Please note that this solution is also the one suggested by WAVE : 

> If the table cell is a header, provide text within the cell that describes the column or row. If the cell is not a header or must remain empty (such as the top-left cell in a data table), make the cell a `td `rather than a `th`.

An alternative solution exists : the developer can provide an accessibility-only title to the column such that `<span class="hidden">Action icons column</span>`. Personally I think that the choice between both solutions should be let to the developer because the second one is not always adapted.

melloware:
Fix #4725: Column accessible header text

We already had a column attrbribute for `ariaHeaderText` to displaying screen reader text on columns that do not have a header.

For example...

```xml
<p:column style="width:2rem" ariaHeaderText="Row Toggler">
   <p:rowToggler/>
</p:column>
```

Generates this HTML with `hidden-accessible` so we don't have empty `TH` tags.

```xml
<th id="j_id_7j:j_id_7k:j_id_7m" class="ui-state-default" aria-label="Row Toggler" scope="col" style="width:2rem">
   <span class="ui-column-title ui-helper-hidden-accessible">Row Toggler</span>
</th>
```

### DIFF
--- a/primefaces-showcase/src/main/webapp/ui/data/datatable/crud.xhtml
+++ b/primefaces-showcase/src/main/webapp/ui/data/datatable/crud.xhtml
@@ -86,7 +86,7 @@
                         <h:outputText styleClass="product-badge status-#{product.inventoryStatus.name().toLowerCase()}"
                             value="#{product.inventoryStatus.text}" />
                     </p:column>
-                    <p:column exportable="false">
+                    <p:column exportable="false" ariaHeaderText="Actions">
                         <p:commandButton icon="pi pi-pencil" update=":dialogs:manage-product-content"
                             oncomplete="PF('manageProductDialog').show()"
                             styleClass="edit-button rounded-button ui-button-success" process="@this">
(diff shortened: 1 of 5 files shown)


tandraschko:
+1 if well tested

PR #11227, opened 2024-01-03
Fix #11203 - IT: Tree flickering test

### LINKED ISSUE #11203: IT: Tree flickering test
This test fails intermittently on the nightly IT.

variants:
```
Tree003Test.tabbing:72->AbstractPrimePageTest.assertDisplayed:88 Element should be displayed!
Tree003Test.tabbing:70->AbstractPrimePageTest.assertDisplayed:88 Element should be displayed!
Tree003Test.tabbing:70 expected: <true> but was: <false>
```

christophs78:


christophs78:
Think we have a (simple) solution. Test is passing even with `@RepeatedTest(200)`.
So i´ll remove `@RepeatedTest(200)` and merge this into master.
I´ll try to come back the next few days to move some of the stuff tried in the other PR into a generic state. (taking screenshot´s when our IT´s fail)

PR #11228, opened 2024-01-03
Fix #99: ConfirmDialog allow ajax=false submission

### LINKED ISSUE #99: Non-Ajax submit support for Global ConfirmDialog

melloware:
Fix #99: ConfirmDialog allow ajax=false submission

- [x] Integration Tests Added
- [x] Showcase example added
- [x] Basically if the function passed to the confirm dialog is EMPTY we know this is AJAX=false so remove the existing `onclick` event which brough up the dialog and then `click` the button.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/dialog/dialog.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/dialog/dialog.js
@@ -939,8 +939,16 @@ PrimeFaces.widget.ConfirmDialog = PrimeFaces.widget.Dialog.extend({
                 if(el.hasClass('ui-confirmdialog-yes') && PrimeFaces.confirmSource) {
                     var id = PrimeFaces.confirmSource.get(0);
                     var js = PrimeFaces.confirmSource.data('pfconfirmcommand');
-
-                    PrimeFaces.csp.executeEvent(id, js, e);
+                    
+                    // Test if the function matches the pattern
+                    if (PrimeFaces.ajax.Utils.isAjaxRequest(js)) {
+                        // command is ajax=true
+                        PrimeFaces.csp.executeEvent(id, js, e);
+                    }
+                    else {
+                        // command is ajax=false
+                        $(id).removeAttr("data-pfconfirmcommand").removeAttr("onclick").click();
+                    }
 
                     PrimeFaces.confirmDialog.hide();
                     PrimeFaces.confirmSource = null;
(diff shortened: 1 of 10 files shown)


melloware:
Another weird one.  This test is passing locally for me.

primefaces/src/main/resources/META-INF/resources/primefaces/dialog/dialog.js
```diff
@@ -939,8 +939,18 @@ PrimeFaces.widget.ConfirmDialog = PrimeFaces.widget.Dialog.extend({
                 if(el.hasClass('ui-confirmdialog-yes') && PrimeFaces.confirmSource) {
                     var id = PrimeFaces.confirmSource.get(0);
                     var js = PrimeFaces.confirmSource.data('pfconfirmcommand');
-
-                    PrimeFaces.csp.executeEvent(id, js, e);
+                    
+                    // Define the pattern to match an empty function like: function(event){}
+                    const isEmptyFunctionPattern = /function\s*\(\s*event\s*\)\s*{\s*}/;
+                    // Test if the function matches the pattern
+                    if (isEmptyFunctionPattern.test(js)) {
```
tandraschko: i wonder if it wouldnt be better to check js.contains(pf.ab || jsf.ajax)?

tandraschko: see PF selenium:         return script.contains("PrimeFaces.ab(") || script.contains("pf.ab(")
                || script.contains("mojarra.ab(")
                || script.contains("myfaces.ab(")
                || script.contains("jsf.ajax.request") || script.contains("faces.ajax.request");

jepsar:
I think I'll wait for this to be merged to update my PR. I'll probably won't have time to work on it until this weekend anyway 😁

primefaces/src/main/resources/META-INF/resources/primefaces/dialog/dialog.js
melloware: Not a bad idea let me try that!

melloware:
@tandraschko here is what the `js` variable contains in AJAX vs NON-AJAX mode.

AJAX: 
```js
PrimeFaces.bcn(this,event,[function(event){PrimeFaces.ab({s:"j_id_7j:j_id_7m",f:"j_id_7j",u:"j_id_7j:message"});return false;}]);
```

NON-AJAX:
```js
PrimeFaces.bcn(this,event,[function(event){}]);
```

That is why above I decided to use a regex to look for an empty function `function(event){}`.  Thoughts?

tandraschko:
no idea TBH, i just thought that checking for AJAX is more safe then a empty-function regex, which could mean anything?

melloware:
Nevermind it works like a charm.  Integration tests running now

PR #11240, opened 2024-01-04
Fix #2478: ColumnTogglerEvent giving access to UIColumn

### LINKED ISSUE #2478: ColumnToggler: Pass specific column attribute in toggle event
If a new column attribute like `toggleId` is available it could be passed to the column toggler event.

It would make it more developer friendly in stateful mode since no visibility lookups would be required using the index. Keeping track of column-index is hard to maintain in big tables and results in problems in combination with _columnGroup_ and reorder mode.

A  new attribute `toggleId` would basically group the following two _year_ columns.

	<p:columnGroup>
		<p:row>
			<p:column headerText="Date" colspan="2" visible="#{bean.visibleMap.get('year') or bean.visibleMap.get('month') }"/>
		</p:row>
		<p:row>
			<p:column headerText="Year" toggleId="year" visible="#{bean.visibleMap.get('year')}"/>
		</p:row>
	</p:columnGroup>

	
	<p:column toggleId="year" visible="#{bean.visibleMap.get('year')}">
		#{car.year}
	</p:column>

	public void onToggle(final ToggleEvent e)
	{
		String toggleId = e.getToggleId(); // year
		visibleMap.put(toggleId , e.getVisibility() == Visibility.VISIBLE);
	}

melloware:
Fix #2478: ColumnTogglerEvent giving access to UIColumn

This gives access to the `UIColumn` that was toggled so you can do...

```java
    public void onToggle(ColumnToggleEvent e) {
        Integer index = (Integer) e.getData();
        UIColumn column = e.getColumn();
        Visibility visibility = e.getVisibility();
        String header = column.getAriaHeaderText() != null ? column.getAriaHeaderText() : column.getHeaderText();
        FacesMessage msg = new FacesMessage(FacesMessage.SEVERITY_INFO, "Column " + index + " toggled: " + header + " " + visibility, null);
        FacesContext.getCurrentInstance().addMessage(null, msg);
    }
```

Backwards compatible since it extends `ToggleEvent`.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/columntoggler/ColumnToggler.java
+++ b/primefaces/src/main/java/org/primefaces/component/columntoggler/ColumnToggler.java
@@ -34,6 +34,9 @@ import javax.faces.event.AjaxBehaviorEvent;
 import javax.faces.event.BehaviorEvent;
 import javax.faces.event.FacesEvent;
 
+import org.primefaces.component.api.UIColumn;
+import org.primefaces.component.datatable.DataTable;
+import org.primefaces.event.ColumnToggleEvent;
 import org.primefaces.event.ToggleCloseEvent;
 import org.primefaces.event.ToggleEvent;
 import org.primefaces.expression.SearchExpressionUtils;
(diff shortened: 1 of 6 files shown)


primefaces/src/main/java/org/primefaces/component/columntoggler/ColumnToggler.java
```diff
@@ -87,7 +90,8 @@ public void queueEvent(FacesEvent event) {
             Visibility visibility = Visibility.valueOf(params.get(clientId + "_visibility"));
             int index = Integer.parseInt(params.get(clientId + "_index"));
 
-            super.queueEvent(new ToggleEvent(this, ((AjaxBehaviorEvent) event).getBehavior(), visibility, index));
+            UIColumn column = ((DataTable) this.getDataSourceComponent()).getColumns().get(index);
```
Rapster: Is `this` needed here?

Rapster (changes requested):
One coment, otherwise looks good

primefaces/src/main/java/org/primefaces/component/columntoggler/ColumnToggler.java
melloware: Fixed!

PR #11246, opened 2024-01-05
#11245 - Preserve jQuery.remove behavior

### LINKED ISSUE #11245: jQuery.remove(selector) behaviour / DataTable: RowReorder - Dragged row is shown as empty
### Describe the bug

Hi,

I found a that bug while using DataTable.
After research it appears that it is connected to jQuery.remove method that changes its behaviour.

```javascript
//(from datatable.js)
                for(var i = 0; i < helperCells.length; i++) {
                    var helperCell = helperCells.eq(i);
                    helperCell.width(cells.eq(i).width());
                    // #5584 reflow must remove column title span
                    helperCell.children().remove('.ui-column-title');
                }
```
That line `helperCell.children().remove('.ui-column-title');` removes all cell children, skips filtering by selector.
It could be changed to `helperCell.children().find('.ui-column-title').remove(;` and it will work fine, but still the jQuery.remove function behaviour is different. Before using PrimeFaces components it works as expected.

Real cause is inside file `jquery/jquery.ui.pfextensions.js`.
So importing jquery/jquery-plugins.js by any PrimeFaces component, for example p:chips or p:dataTable, cause remove function to skip argument.

```javascript
//(from jquery/jquery.ui.pfextensions.js)
// GitHub #9941 tooltip removal fix
(function() {
    var ev = new $.Event('remove'),
        orig = $.fn.remove;
    $.fn.remove = function() {        
        $(this).trigger(ev);
        return orig.apply(this, arguments);
    };
})();
```

Triggering remove event on element removes it, without checking the selector provided in arguments.
So as a fix we could check if arguments is empty, and only then trigger event.

### Reproducer

**Reproducer of jQuery.remove problem**
1. Create simple .xhtml page with content as follows:
```html
    <button onclick="$('.parentDiv').children().remove('.elementToRemove')"></button>

    <div class="parentDiv">
        <div class="elementToRemove">
            Element to remove
        </div>
        
        <div class="element">
            Element 1 to NOT REMOVE
        </div>
        
        <div class="element">
            Element 2 to NOT REMOVE
        </div>
    </div>
```

2. Click on button. Only one element has been removed. Thats correct.

3. Change xhtml page. Add any PF component that imports jquery-plugins.js (p:chips or p:dataTable or another one).

```html
<p:chips></p:chips>
```
5. Click on button. ALL elements have been removed! Thats incorrect!

**Reproducer of DataTable problem**:

1. Create simple DataTable with rowReorder enabled, and with columns with some elements inside them (span, div).          
```html
            <p:dataTable id="table" value="#{simpleView.rows}" var="row" 
                         draggableRows="true" 
                         selectionMode="multiple"
                         rowKey="#{row}">
                <p:column>
                        THIS TEXT WILL BE VISIBLE
                    <div>
                        #{row}
                        THIS TEXT WILL BE INVISIBLE!
                    </div>
                </p:column>
            </p:dataTable>
```

2. Initialize DataTable with some rows.
```java

@Named
@ViewScoped
public class SimpleView implements Serializable {

    private List<String> rows = new ArrayList<>();

    @PostConstruct
    public void init() {
        for (int i = 0; i < 1000; i++) {
            rows.add("Row #" + i);
        }
    }

        return rows;
    }

}
```

3. Drag any row (only part of column is visible).

### Expected behavior

**JQuery.remove problem**
I expected that jQuery.remove behaviour will be not changed.

**DataTable problem**
I expected to see whole row content that is being dragged, even if it consists of some other elements (div, spans or others).
On PF 13.0.4 only direct content is preserved.

### PrimeFaces edition

Community

### PrimeFaces version

13.0.4

### Theme

Saga (not connected to theme)

### JSF implementation

Mojarra

### JSF version

2.2 / 2.3

### Java version

1.8

### Browser(s)

_No response_

Krzysztof-13:
Fix #11245

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.js
@@ -116,7 +116,7 @@ $.widget( "ui.sortable", $.ui.sortable, {
         orig = $.fn.remove;
     $.fn.remove = function() {
         // Don't change JQuery.remove(selector) behavior, so trigger event only when remove is called without arguments.
-        if (arguments.length === 0) {
+        if (arguments && arguments.length === 0) {
             $(this).trigger(ev);
         }
         
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.js
@@ -116,7 +116,7 @@ $.widget( "ui.sortable", $.ui.sortable, {
         orig = $.fn.remove;
     $.fn.remove = function() {
         // Don't change JQuery.remove(selector) behavior, so trigger event only when remove is called without arguments.
-        if (arguments && arguments.length === 0) {
+        if (!arguments || arguments.length === 0) {
             $(this).trigger(ev);
         }
         
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.js
@@ -115,7 +115,11 @@ $.widget( "ui.sortable", $.ui.sortable, {
     var ev = new $.Event('remove'),
         orig = $.fn.remove;
     $.fn.remove = function() {
-        $(this).trigger(ev);
+        // Don't change JQuery.remove(selector) behavior, so trigger event only when remove is called without arguments.
+        if (arguments.length === 0) {
+            $(this).trigger(ev);
+        }
+        
         return orig.apply(this, arguments);
     }
 })();


primefaces/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.js
```diff
@@ -115,7 +115,11 @@ $.widget( "ui.sortable", $.ui.sortable, {
     var ev = new $.Event('remove'),
         orig = $.fn.remove;
     $.fn.remove = function() {
-        $(this).trigger(ev);
+        // Don't change JQuery.remove(selector) behavior, so trigger event only when remove is called without arguments.
+        if (arguments.length === 0) {
```
melloware:
```suggestion
        if (arguments && arguments.length === 0) {
```

melloware (changes requested):
Small change just to be safe

melloware (approved):

melloware:
Itegration tests running now.

PR #11291, opened 2024-01-16
Fix #10870: Chart.js new component

### LINKED ISSUE #5665: ChartsJS: TimePoint value type

### LINKED ISSUE #10849: Charts: unified and consistent data interface classes

### LINKED ISSUE #10870: ChartJS: new approach

melloware:
Fix #10870: Chart.js new component
Fix #5665 ChartJS Time Values
Fix #10849 Unify Charts

- [x] Documentation Created
- [x] All new Showcase examples showing Raw JSON and XDEV Model
- [x] Recreated all showcase examples using XDEV
- [x] Custom example so you can paste any JSON in you want
- [x] Facet use `<f:facet name="value>` for simpler charts
- [x] Add zoom plugin https://www.chartjs.org/chartjs-plugin-zoom/latest/guide/
- [x] Compress charts.js


melloware:
I created a Custom example in the Showcase where you can take any Chart.js example code and paste it in and it renders the chart.

image

primefaces/pom.xml
```diff
@@ -607,6 +606,7 @@
                         <!-- uncompressed widgets -->
                         <resourcesSet>
                             <includes>
+                                <include>chart/**</include>
```
Rapster, line 610: does it get minified then?

docs/14_0_0/components/chart.md
```diff
@@ -0,0 +1,270 @@
...
+
+    public String getJson() {
+        return json;
+    }
+}
+```
+
+### Getting Started with XDEV Java Model
+
+You can use [XDEV Chart.js Java Model](https://github.com/xdev-software/chartjs-java-model) to render a server side model of your Chart.
```
Rapster, line 206: should we indicate the XDEV version compatible with PF?

docs/14_0_0/components/chart.md
```diff
@@ -0,0 +1,270 @@
...
+```
+
+```java
+public class Bean {
+    
+    private String json;
+
+    @PostConstruct
+    public void init() {
+        json = "{\r\n"
```
Rapster: can't we use a text block for this (e.g """) ?

primefaces/pom.xml
melloware: No I moved this to the unminified section so it gets combined but not minified.

docs/14_0_0/components/chart.md
melloware: Let me add it to the Dependencies section of the docs where we have OpenLibre and other optional libraries.

docs/14_0_0/components/chart.md
melloware: I would love to but I think that is Java 17.

docs/14_0_0/gettingstarted/whatsnew.md
```diff
@@ -13,6 +13,9 @@ Look into [migration guide](https://primefaces.github.io/primefaces/14_0_0/#/../
     * OS settings for `prefers-reduced-motion: reduce` is now respected and PF disables all animations.
     * Minimum inline Ajax load animation duration via `PrimeFaces.ajax.minLoadAnimation`.
 
+* Chart
+   * A new more flexible version of Chart.js can be used `<p:chart>` allowing raw JSON or XDEV Java Model
+
```
Rapster, line 18: should we mention that the current impl is deprecated and will be remov in later version?

docs/14_0_0/components/chart.md
Rapster: ah my bad, I had in mind it was java 11...

docs/14_0_0/gettingstarted/whatsnew.md
Rapster: should we mention that the current impl is deprecated and will be remov in later version?

melloware: Yes let me add a migration guide warning

melloware:
I also added the `<f:facet name="value">` per @tandraschko suggestion and I like it for simple charts where you don't even need Java code.

primefaces/pom.xml
Rapster: why not minify it?

melloware: So the 2 Chart JS libs the CHartJS itself and the moment plugin are already minified.   If you try and minify it again the Resource Minifier throws an error about the level of Javascript.   So the only thing not being minified is the Widget itself its just being appended on so its small.   That is how the current CHarts.js stuff is happening too because of this issue it was always unminified just combined to 1 file

primefaces/src/main/java/org/primefaces/component/chart/ChartRenderer.java
```diff
@@ -0,0 +1,111 @@
...
+            facet.encodeAll(context);
+
+            // restore the original writer
+            context.setResponseWriter(originalWriter);
+            return fsw.toString();
+        }
+        else {
+            return chart.getValue();
+        }
+    }
```
Rapster, line 109: you had no issues with escaping? I remember doing something very similar but had escaping issues...

melloware: Not that I have seen so far. There may be an issue but I have not run into it yet.

primefaces/pom.xml
Rapster: I assume you're refering to this: https://github.com/primefaces/primefaces/pull/10871#issuecomment-1774091993 and my PR contained the workaround. Ideally everything should be minified (whether it's small or big),

melloware: Yes so I saw what you did and it does not work.  You tried to compress `2-chartjs-widget.js` and then later combine them into `charts.js` with the 2 minified files and `2-widget.js` but that won't work.  It Just combines them but the widget is still uncompressed.

astappiev:
Could it be possible to include [chartjs-plugin-zoom](https://www.chartjs.org/chartjs-plugin-zoom/latest/) as well?

primefaces/pom.xml
Rapster: [chart.js.txt](https://github.com/primefaces/primefaces/files/13975438/chart.js.txt)

I can see 3 files in here and everything look compressed, am I missing something?

melloware:
I don't see why not @astappiev as multiple people have asked for this before.

primefaces/pom.xml
melloware: Let me try again but that is not the behavior I am seeing inside the final JAR.  Maybe I didn't do a `clean`

melloware:
@Rapster compression is working fine thanks!

@astappiev added Zoom plugin

Rapster:
Great job! 💯

melloware:
Final size compressed is 241 KB!

melloware:
@Rapster @astappiev i was also able to accomplish this now!

1. Old and new Charts can co-exist now. I removed the old chart.js file and now they all share the same CHart.js.  All of the demos are working properly and now charts can co-exist!

2. I marked the old ones deprecated.

image

docs/14_0_0/components/chart.md
Rapster: For the sake of readibility, and as it is only documentation, here using text blocks would be better

docs/14_0_0/gettingstarted/dependencies.md
```diff
@@ -18,6 +18,7 @@ any 3rd part work incorporated are compatible with the PrimeFaces Licenses.
 | barcode4j-light | org.primefaces.extensions | barcode4j-light | 2.3.0 | Optional | Barcode |
 | qrcodegen | io.nayuki | qrcodegen | 1.8.0 | Optional | QR Code support for Barcode |
 | owasp-java-html-sanitizer | com.googlecode.owasp-java-html-sanitizer | owasp-java-html-sanitizer |  20220608.1 | Optional | TextEditor |
+| XDev chartjs-java-model | software.xdev | chartjs-java-model |  1.1.2 | Optional | Chart |
```
Rapster, line 21: `XDev chartjs-java-model`  is not even an optional dependency, don't think it should be mention here

docs/14_0_0/components/chart.md
melloware: Good idea

melloware:
@Rapster you can submit any follow PR's if you find I have missed anything.

Rapster (commented):
forgot to submit the review...

docs/14_0_0/gettingstarted/dependencies.md
melloware: I will remove

PR #11300, opened 2024-01-17
Fix #5665  Chartjs: fixes to allow timeseries bar chart

### LINKED ISSUE #5665: ChartsJS: TimePoint value type

astappiev:
Fix #5665 ChartJS Time Values

1. `BarChartDataSet` should allow `NumericPoint` to be used as data, useful to create timeseries bar charts.

2. `CartesianTime` probably was mistakenly left as abstract. It can be used in `CartesianLinearAxes.time` but can't be created without extending (which is redundant).

### DIFF
--- a/primefaces-showcase/src/main/java/org/primefaces/showcase/view/chartjs/ChartJsView.java
+++ b/primefaces-showcase/src/main/java/org/primefaces/showcase/view/chartjs/ChartJsView.java
@@ -348,7 +348,7 @@ public class ChartJsView implements Serializable {
         BarChartDataSet barDataSet = new BarChartDataSet();
         barDataSet.setLabel("My First Dataset");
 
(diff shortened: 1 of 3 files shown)


astappiev:
I just saw #11291, that would be cool to have in next major release. However, it would be nice to include my fixes in the next 13 releases.

melloware:
Yep this will handle it because we automatically include moment.js convesion so it understands long 
epoch.

https://jsfiddle.net/escay/y9j3x5cf/6/

melloware:
Looks like the build is failing.

melloware (approved):

PR #11305, opened 2024-01-18
Fix #11301: DefaultCommand CSP with ajax=false

### LINKED ISSUE #11301: CSP: Default command pointing ajax="false" does nothing
### Describe the bug

Hi, 
in our app we use p:defaultCommand quite a lot, sometimes pointing to ajax="false" commandButton and sometimes to ajax="true".

We found out that when CSP is turned ON, the ajax="false" commandButton is not invoked when user presses ENTER in input field. 

Please see attached reproducer .zip.

I tried PF versions 13.0.2 and 13.0.4, both not working.

### Reproducer

See attached .zip file with primefaces-test reproducer: 
[defaultcommand_vs_ajaxfalse.zip](https://github.com/primefaces/primefaces/files/13975272/defaultcommand_vs_ajaxfalse.zip)

Snippet from reproducer:
```
<h:form id="frmTest">
            value:
            <br/>

            <p:inputText id="inputValueId" value="#{testView.inputValue}"/>

            <br/><br/>

            <p:commandButton id="cb_next_ajax_true" value="next with attribute ajax true" action="step2" ajax="true"/>
            <br/>
            <p:commandButton id="cb_next_ajax_false" value="next with attribute ajax false" action="#{testView.next()}"
                             process="@form" update="@form" ajax="false"/>

          <!-- With ajax it works great WITH and WITHOUT CSP -->
<!--            <p:defaultCommand target="cb_next_ajax_true"/>-->

          <!-- NON AJAX works only when CSP disabled -->
            <p:defaultCommand target="cb_next_ajax_false"/>
            
        </h:form>
```

### Expected behavior

I expect that the  defaultCommand pointing to ajax="false" commandButton would invoke the button both when CSP ON and OFF.

### PrimeFaces edition

Elite

### PrimeFaces version

13.0.2

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3.21

### Java version

17

### Browser(s)

Chrome

melloware:
Fix #11301: DefaultCommand CSP with ajax=false

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.csp.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.csp.js
@@ -59,7 +59,7 @@ if (!PrimeFaces.csp) {
          */
         isFacesForm: function(form) {
             if (form.method === 'post') {
-                for (child in form.children) {
+                for (let child of form.children) {
                     if (child instanceof HTMLInputElement && child.name && child.name.includes(PrimeFaces.VIEW_STATE)) {
                         return true;
                     }
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.csp.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.csp.js
@@ -78,7 +78,8 @@ if (!PrimeFaces.csp) {
             if (event) {
                 var shortenedEvent = event.substring(2, event.length),
                     element = document.getElementById(id),
-                    jqEvent = shortenedEvent + '.' + id;
+                    jqEvent = shortenedEvent + '.' + id,
+                    isAjaxified = PrimeFaces.ajax.Utils.isAjaxRequest(js.toString());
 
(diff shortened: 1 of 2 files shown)


melloware:
@tandraschko i am researching another 14.0.0 weirdness that works in 13.0.5-SNAPSHOT.  Using the reproducer for this ticket the differences when pressing ENTER with DefaultCommand.   it is missing `primeFaces.NONCE` in the request

13.0.5-SNAPSHOT (correct):
image

14.0.0 (incorrect):
image

melloware:
@tandraschko i can't explain it but i fixed it by chnaging

```js
for (child in form.children) {
```

to

```js
for (let child of form.children) {
```

Which according to Mozilla is how to correctly loop over HTMLCollection: https://www.geeksforgeeks.org/htmlcollection-for-loop/

But I have no idea why the same code works in 13.x

tandraschko:
weird but ok :D  im always confused with  in/of in JS

PR #11340, opened 2024-01-26
#11036: CSV: enable (command-)button only after client side validation passes (ui-state-csv-valid + ui-state-csv-invalid)

### LINKED ISSUE #11036: CSV: enable (command-)button only after client side validation passes

christophs78:
# with project specific styling

Fix #11036 

## State after loading
(only ui-state-csv-valid + ui-state-csv-invalid with project-specific styling)
image

## State after some inputs (ui-state-error kicks in like in previous PF versions)
image

# without project specific styling

## State after loading
image

## State after some inputs
image


tandraschko:
Let me review next week again

Should we add this to our sass theme then?

tandraschko:
And is it possible to configure immediate or not?
Afair the most forms on the web does not immediate validate and mark red

christophs78:
> And is it possible to configure immediate or not? Afair the most forms on the web does not immediate validate and mark red

Hm. Technically we need to run CSV for this feature after each full page load and after each AJAX-update. (Because we IMO can´t make assumptions which values are already correct prefilled,.. So we have to look what we have on the client, run CSV and enable/disable buttons based on this afterwards.)
I´ll mark the code-snippet where we already differentiate between calls triggered by full page load or AJAX-update and usual CSV triggered by e.g. `p:clientValidator`.

When we look at https://angular.io/guide/form-validation#control-status-css-classes we see CSS-classes (~states) like ng-dirty or ng-touched. IMO there are good reasons for this. (Like you implied with your response.)
But this IMO would require more work to keep this state after eg AJAX-updates. (The "price" of JSF because state is splittet between server and client.)

IMO it would be a good step to check out this PR and play around with it. I´m open to almost everthing (including adding some styles to sass-theme) We "only" have to keep an eye on what takes how much effort and what are the milestones to get to our target via maybe multiple releases.

We may also do a short video-conference in the evening to discuss ideas and there feasibility.

primefaces/src/main/resources/META-INF/resources/primefaces/validation/validation.highlighters.js
```diff
@@ -34,6 +34,23 @@ if (window.PrimeFaces) {
             }
         },
 
+        /**
+         * Applies ui-state-XXX - css-classes to an element (component).
+         * @param {JQuery} element Element to which apply the css-classes.
+         * @param {boolean} valid Is the input of the element valid?
+         * @param {boolean} csvStateOnly Only set ui-state-csv-XXX - classes but nothing else like ui-state-error.
+         */
+        applyStateCssClasses: function(element, valid, csvStateOnly) {
```
christophs78: Here we may play around ...

tandraschko:
> Technically we need to run CSV for this feature after each full page load.... Because we IMO can´t make assumptions which values are already correct prefilled

I think thats a general question. From a user perspective, i like our old approach much more. And as long it is something additional, its ok for me.

From a user perspective:
If you open a view, where e.g. the form with CSV is only one possible action, why should it be validated?
If you start typing in a related input -> validate
If you press the button -> validate

Of course then we can't disable the button.

Maybe just "Modern" isnt the right name for this - maybe something like "immediate" or "instant" validation?

christophs78:
OK, i choosed "immediate".

With some design- and UX-things it´s like with fashion. Decisions are to some degree driven by "Zeitgeist".

The pro-argument is why let people click a button and tell them afterwards why it does not work (yet). The con-argument is users sometimes need to figure out what´s necessary to get the button enabled.

primefaces/src/main/resources/META-INF/resources/primefaces/validation/validation.common.js
```diff
@@ -441,15 +482,13 @@ if (window.PrimeFaces) {
                 element.attr('aria-invalid', false);
             }
             else {
-                if (highlight) {
-                    highlighter.highlight(element);
-                }
+                highlighter.highlight(element, !highlight);
```
tandraschko: .highlight is called with the second new param, but .unhighlight not, right?

christophs78: Yep. "unhighlight" does essentially the same work as before. When we call highlight from the new codepaths we don´t want  to show e.g. messages, so we need the additional parameter.

tandraschko: should we just pass "false" then? everything else looks weird :D

christophs78: ok, i´ll add it in den evening.

christophs78:
Even github does this trick when opening new issues. "Submit issue" is only enabled after all required fields are filled.

christophs78:
Maybe stupid question because i´m crosschecking nothing breaks ...

There is following code:
```javascript
PrimeFaces.validator.Highlighter = {

	/**
	 * When an element is invalid due to a validation error, the user needs to be informed. This method highlights
	 * the label for the given element by adding an appropriate CSS class.
	 * @param {JQuery} forElement Element with a label to highlight.
	 */
	highlightLabel: function(forElement) {
		var label = $("label[for='" + forElement.attr('id') + "']");
		if (label.hasClass('ui-outputlabel')) {
			label.addClass('ui-state-error');
		}
	},
```
a) this only works for p:outputLabel?
b) there is no default styling for this - it´s also some additional css-class you may use to do some project-specific styling of the label?

--> I´m thinking about dropping csvStateOnly from validation.highlighters.js. I probably introduced this without real need and added unnecessary complexity.

tandraschko:
1) seems so
2) no idea, can't remember 

why did you introduce it?

christophs78:
cleanup done - can we merge this PR to get into v14?

melloware:
Yep I will merge so it can make 14-RC1.  @tandraschko we can always continue to refactor.

PR #11341, opened 2024-01-27
#11322 DatePicker: enabledDates - allows the user to select only dates from the provided list

### LINKED ISSUE #11322: DatePicker: enabledDates - allows the user to select only dates from the provided list
### Description

In the old p:calendar component, we had a feature that allowed the user to select a date from just a few dates in the history (imagine the use case that you want to allow the user to select the days when a diary is written) - for this we used the `beforeShowDate` function with additional javascript function.

The new p:datePicker component doesn't have this feature, but it does have a function for specifying `disabledDates` - which is unfortunately unusable for this use case - you can't insert all but a few dates in the history into the List.

What do you think about a similar feature - the `enabledDates` DatePicker property, which could be used to specify only those dates that can be selected in the component - the others will be disabled.

After a quick look at the code, I think it's not that hard to implement - just create the opposite function to `disableDates`. If you think this function could be useful, let me know - I could prepare a PR for you.  If not, just close the issue :)

Thanks 
Milan

### Additional context

_No response_

milansie:
.. ability to select only the dates, that could be selected from datePicker ..

### DIFF
--- a/docs/14_0_0/components/datepicker.md
+++ b/docs/14_0_0/components/datepicker.md
@@ -40,6 +40,7 @@ ajax selection and more.
 | disabled | false | Boolean | Disables input field
 | disabledDates | null | List<java.time.LocalDate>, List<java.util.Date> (deprecated) | List of dates that should be disabled.
 | disabledDays | null | List&lt;Integer&gt; | List of week day indexes that should be disabled.
+| enabledDates | null | List<java.time.LocalDate>, List<java.util.Date> (deprecated) | List of dates that should be enabled.
 | flex | false | Boolean | Use modern PrimeFlex-Grid instead of classic Grid CSS. (primeflex.css must be included into the template.xhtml)
 | focusOnSelect | false | Boolean | When enabled, input receives focus after a value is picked.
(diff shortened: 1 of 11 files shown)


jepsar:
Can you please add this to the lazy model as well?

melloware (approved):
This looks great.  I will merge once you add the lazy model example!

jepsar:
What happened to the lazy model? It looks like it wasn't added there.

milansie:
> What happened to the lazy model? It looks like it wasn't added there.

..just working on that - i will provide (new) PR in few minutes / hours ...

CLuv557:
I have tried every Primefaces version from 14.0 to 14.0.4 within a maven project and Netbeans 12.5 still says "The attribute enabledDates is not defined in the component datePicker". I deleted the Netbeans cache folder and did a clean and build multiple times. Was this committed in 14.0?

melloware:
let me look i think its missing from the Taglib but it definitely works according to the ShowcasE: https://www.primefaces.org/showcase/ui/input/datepicker/datePickerJava8.xhtml

let me investigate.

melloware:
Submitted fix: https://github.com/primefaces/primefaces/pull/12542

PR #11374, opened 2024-02-01
Fix #11360: FileUploadUtils code coverage and javadoc

### LINKED ISSUE #11360: FileUpload/ImageCropper: path validation incomplete / not well documented / erroneous tests
Discussed in #11352

both tests and code are not well documentation, what actually should be validated

i disabled one test (it fails in windows) as it fails by design: it tries to validate that a path cant be absolute, but it doesnt fail
also it doesnt make sense to fail it as its completely valid to use `AbstractUploadedFile#write(D:/myFile.txt)`

See:
https://github.com/primefaces/primefaces/issues/3268
https://github.com/primefaces/primefaces/pull/9376
https://github.com/primefaces/primefaces/issues/11134

melloware:
Fix #11360: FileUploadUtils code coverage and javadoc

@tandraschko i did the following.

- [x] Wrote more unit tests.
- [x] Added Javadoc comments on what and why code is doing what it is
- [x] Increased Code Coverage on requireValidFileName to 100%
- [x] Un-disabled the test you had disabled because now its fixed properly
- [x] Removed Windows specific tests as now the code properly does absolute and canonical paths by doing

```java
 // standardize Windows paths from d:/test to d:\\test
filePath = filePath.replace("/", "\\");
```
image

### DIFF
--- a/primefaces/src/test/java/org/primefaces/util/FileUploadUtilsTest.java
+++ b/primefaces/src/test/java/org/primefaces/util/FileUploadUtilsTest.java
@@ -25,7 +25,6 @@ package org.primefaces.util;
 
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.mockito.MockedStatic;
 import org.mockito.Mockito;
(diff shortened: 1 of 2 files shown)


primefaces/src/test/java/org/primefaces/util/FileUploadUtilsTest.java
```diff
@@ -327,6 +358,63 @@ void requireValidFilename_Windows() {
...
+        ValidatorException ex = assertThrows(ValidatorException.class, () -> FileUploadUtils.requireValidFilename(invalidFileName));
+
+        // Assert
+        if (os.contains("win")) {
+            assertEquals("Invalid filename: (file:*) contains invalid character: :", ex.getFacesMessage().getDetail());
+        }
+        else {
+            assertEquals("Invalid Linux filename: file:*", ex.getFacesMessage().getDetail());
+        }
+    }
```
Rapster, line 416: why not using  `Mockito#mockStatic`?

primefaces/src/test/java/org/primefaces/util/FileUploadUtilsTest.java
```diff
@@ -232,6 +211,48 @@ void requireValidFilePath_PathTraversal() {
...
+        ValidatorException ex = assertThrows(ValidatorException.class, () -> FileUploadUtils.requireValidFilePath(invalidFileName, false));
+
+        // Assert
+        if (os.contains("win")) {
+            assertEquals("Invalid path: (invalid\\path*) contains invalid character: *", ex.getFacesMessage().getDetail());
+        }
+        else {
+            assertEquals("Invalid directory, \"invalid\\path*\" is not valid.", ex.getFacesMessage().getDetail());
+        }
+    }
```
Rapster, line 254: why not using Mockito#mockStatic?

primefaces/src/test/java/org/primefaces/util/FileUploadUtilsTest.java
Rapster: ok basically, everywhere where `os.contains("win")` is used

melloware: Great idea.  I will update the tests to use mock static.

melloware:
@Rapster i have to leave this as is because the new code uses `Paths.get(string)` to check the path is valid and that uses the underlying OS's FileSystem to determine that.  So there is nothing for me to mock to fake it out without ruining the real test.

PR #11376, opened 2024-02-02
Fix #10902/#11045: DataTable livescroll/virtualscroll fix

### LINKED ISSUE #10902: Datatable: command buttons inside datatable not called when virtualScroll is enabled
### Describe the bug

I have a datatable with virtualScrolling and lazy loading enabled.
Each row displays a label and has a commandButton which turns the label to upperCase and then updates the datatable.
This works for the first rows but not for the ones loaded when scrolling down: the commandButton action is never executed on the backing bean. 

### Reproducer

Reproducer in [https://github.com/YotillaPablo/primefaces-test](https://github.com/YotillaPablo/primefaces-test)

Go to http://localhost:8080/primefaces-test/
Click the 'toUpperCase' button on the first row, labelled 'Item 1': this will turn it to 'ITEM 1' and a 'updating ITEM 1' message will appear on System.out.
Click the 'toUpperCase' button on the row 57, labelled 'Item 57': this won't have any effect and nothing will be printed to System.out

### Expected behavior

The commandButton works on all datatable rows.

### PrimeFaces edition

Community

### PrimeFaces version

13.0.2

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.2

### Java version

17

### Browser(s)

_No response_

### LINKED ISSUE #11045: DataTable: Ajax update on LiveScroll DataTable causes the current page to become the first page
### Describe the bug

If I have a DataTable with lazy="true" and liveScroll="true" and I scroll down until the second page is loaded. 

Whenever an Ajax update is done on the table, then the first page disappears and my second page becomes the "first" page of the Table so when I scroll down, then the third page is loaded. Same logic applies if I have loaded the 5th page, then after an Ajax the 5th page becomes the first page of the table and so on.

### Reproducer

(https://github.com/Vikkear/primefaces-test-datatable-ajax-bug/tree/jpa-datatable-lazy-livescroll-ajax-bug)
1. Scroll down until second or third page is loaded
2. Press the button to start an Ajax update on the table
3. Now page 2 or 3 is the first loaded page in the DataTable

### Expected behavior

Whenever the Ajax update is done, then the first page should still be loaded in the datatable.

### PrimeFaces edition

Elite

### PrimeFaces version

13.0.3

### Theme

nova-light

### JSF implementation

MyFaces

### JSF version

5.1.2

### Java version

17.0.6

### Browser(s)

Version 119.0.6045.160 (Official Build) (64-bit)

melloware:
Fix #10902 DataTable livescroll/vritualscroll fix
Fix #11045: DataTable livescroll/vritualscroll fix

I have confirmed this fixes both tickets.

cc @wlhUser

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java
+++ b/primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java
@@ -489,7 +489,12 @@ public class DataTable extends DataTableBase {
             offset += rows;
         }
 
-        setFirst(offset);
+        if (isVirtualScroll() || isLiveScroll()) {
+            setFirst(0);
+        }
+        else {
+            setFirst(offset);
+        }
 
         if (calculateFirst()) {
             offset = getFirst();


Rapster:
Cool fix! Looks simple enough. But the problem I see then, if the offset is bypassed (no matter what offset has been passed in this method, it means, the passed offset is wrong) Are you fixing the issue at the right place? IMO, we should understand why the offset is wrong in the first place, and if not 0 in lazyScrollData for virtualscroll (a log should indicate the unexpected value)

Sorry, lately I have no time for PF...

melloware:
Agreed but since this is a regression  of LiveScroll/VirtualScroll I think we should implement it and in the future investigate more.  The current behavior is definitely incorrrect and from your refactoring in #9772 i am not sure what the correct outcome is.

Rapster:
regression?

melloware:
Also here: https://github.com/primefaces/primefaces/pull/9809/files#diff-4701970a7aabff8b155f0f3a7e3c2a54291a39b17a4176684c82ed83cb8fe53e

Issue #5710, opened 2020-03-19
OverlayPanel: Race condition with embedded components on AJAX-update


fanste:
## 1) Environment
- PrimeFaces version: 8.0

## 2) Expected behavior
Components (e.g. `p:selectOneMenu`) inside a dynamic overlay (e.g. `p:overlayPanel`) with `appendTo` should not double the found entries when the `jq` property of the component widget is resolved

## 3) Actual behavior
Due to the `appendTo` the DOM of the dynamic overlay may survive the AJAX update which effectively doubles the DOM for a short time. The first instance is at the original position (before it is moved to its `appendTo` destination) which was added during the AJAX update and the second instance is the old one which may be outside the updated area.

The SelectOneMenu is refreshed before the OverlayPanel. But the old DOM of the OverlayPanel is removed during the refresh of the OverlayPanel ([PrimeFaces.utils.registerDynamicOverlay > widget.addRefreshListener](https://github.com/primefaces/primefaces/blob/master/src/main/resources/META-INF/resources/primefaces/core/core.utils.js#L225)). Therefore the SelectOneMenu finds two DOM representation of its widget for `SelectOneMenu#jqId`.

## 4) Steps to reproduce
 - Press button `Alert state ofSelectOneMenu`
    Output should be size = 1 and label = 1
 - Press the button `Update OP`
 - Press button `Alert state ofSelectOneMenu`
    Output should be size = 2 and label = 11 which is wrong

## 5) Sample XHTML
```
<h:form>
    <p:commandButton value="Update OP" process="@this pnlTest" update="pnlTest" />
    <h:panelGroup id="pnlTest">
        <p:commandButton id="btnTest" value="Open Overlay" type="button" />
        <p:overlayPanel id="opTest" for="btnTest" appendTo="@(body)">
            <p:selectOneMenu value="#{testBean.selectedTest}" widgetVar="somTest">
                <f:selectItem itemLabel="1" itemValue="1" />
            </p:selectOneMenu>
        </p:overlayPanel>
    </h:panelGroup>

    <p:commandButton value="Alert state ofSelectOneMenu" type="button" onclick="alert('Size of somTest#jq = ' + PF('somTest').jq.length + '\nSelected Label = ' + PF('somTest').getSelectedLabel())" />
</h:form>
```

## 6) Sample bean
```
@ViewScoped
@Named
public class TestBean implements Serializable {
    private Integer _selectedTest;

    public Integer getSelectedTest() {
        return _selectedTest;
[...]

PR #11413, opened 2024-02-09
Fix #11409: OverlayPanel add `destroyWithTarget` prop

### LINKED ISSUE #11409: OverlayPanel won't show when attached to a CommandButton
### Describe the bug

The fix in #11244 causes problems with an OverlayPanel attached to a CommandButton. The panel will be destroyed and isn't rendered any more, resulting in a JS error like: `Widget for 'xxxx' not available!`

The fix in #11387 does not fix this behavior.

### Reproducer

Not working examples:
```xml
<p:commandButton id="btn" value="Open" />
<p:overlayPanel for="btn" dynamic="true">
  <i>Hint</i>
</p:overlayPanel>
```
```xml
<h:panelGroup id="grp">
  <p:commandButton id="btn" value="Open" />
  <p:overlayPanel for="grp" dynamic="true">
    <i>Hint</i>
  </p:overlayPanel>
</h:panelGroup>
```
Partially working example:
```xml
<h:panelGroup id="grp">
  Working on Text
  <p:commandButton id="btn" value="Not working on button" />
  <p:overlayPanel for="grp" dynamic="true">
    <i>Hint</i>
  </p:overlayPanel>
</h:panelGroup>
```

Workaround, updating the wrapper:
```xml
<h:panelGroup id="wrp">
  <p:commandButton id="btn" value="Open" update="wrp"/>
  <p:overlayPanel for="btn" dynamic="true">
    <i>Hint</i>
  </p:overlayPanel>
</h:panelGroup>
```

### Expected behavior

The minimal example was working in 13.0.4 and should work again.
```xml
<p:commandButton id="btn" value="Open" />
<p:overlayPanel for="btn" dynamic="true">
  <i>Hint</i>
</p:overlayPanel>
```

### PrimeFaces edition

Community

### PrimeFaces version

13.0.5

### Theme

_No response_

### JSF implementation

All

### JSF version

2.3.21

### Java version

17.0.10

### Browser(s)

_No response_

melloware:
Fix #11409: OverlayPanel add `destroyWithTarget` prop

@tandraschko @jepsar @Rapster this was the best name I could come up with but I am open to suggestions...

### DIFF
--- a/docs/13_0_0/components/overlaypanel.md
+++ b/docs/13_0_0/components/overlaypanel.md
@@ -28,6 +28,7 @@ at | left bottom | String | Position of the target relative to the panel.
 autoHide | true | boolean | Whether to hide overlay when hovering over overlay content when using custom show/hide events.
 cache | true | Boolean | Only relevant for dynamic="true": Defines if activating the panel should load the contents from server again. For cache="true" (default) the panel content is only loaded once.
 collision | flip | String | When the positioned element overflows the window in some direction, move it to an alternative position. Similar to my and at, this accepts a single value or a pair for horizontal/vertical, e.g., "flip", "fit", "fit flip", "fit none".
+destroyWithTarget | false | Boolean | Destroys the overlay panel if the target 'for' element is removed from the DOM, default is `false`.
 dismissable | true | Boolean | When set true, clicking outside of the panel hides the overlay.
 dynamic | false | Boolean | Defines if dynamic loading is enabled for the element's panel. If the value is "true", the overlay is not rendered on page load to improve performance. Default is false.
 for | null | String | Target component to display panel next to.
--- a/primefaces-integration-tests/src/main/webapp/overlaypanel/overlayPanel002.xhtml
+++ b/primefaces-integration-tests/src/main/webapp/overlaypanel/overlayPanel002.xhtml
@@ -19,9 +19,9 @@
                 <p:outputPanel id="dialogPanel">
                     <p:commandButton id="btnShow" value="Show" type="button"/>
                     <p:commandButton id="btnUpdate" value="Update" update="dialogPanel"/>
-                    <p:commandButton id="btnDestroy" value="Destroy" actionListener="#{overlayPanel002.destroy}" update="dialogPanel"/>
+                    <p:commandButton id="btnDestroy" value="Destroy" actionListener="#{overlayPanel002.destroy}" update="dialogPanel" onstart="PF('wgtOverlaypanel').destroy();"/>
                     <p:commandButton id="btnAppend" value="Append" actionListener="#{overlayPanel002.appendToSomething}" update="dialogPanel"/>
-                    <p:overlayPanel id="overlaypanel" for="btnShow" rendered="#{overlayPanel002.rendered}" appendTo="#{overlayPanel002.appendTo}">
+                    <p:overlayPanel id="overlaypanel" widgetVar="wgtOverlaypanel" for="btnShow" rendered="#{overlayPanel002.rendered}" appendTo="#{overlayPanel002.appendTo}">
                         <p>Content of overlay panel.</p>
                     </p:overlayPanel>
                 </p:outputPanel>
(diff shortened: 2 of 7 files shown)


melloware:
Sure.  So the original bug: https://github.com/primefaces/primefaces/issues/5710

Basically if the OverlayPanel was in the `body` and it is updated in an `update` statement it creates two overlay panels because the old one is still in the DOM and never gets removed.  I had added `remove` subscription on the `for` element so then the `for` element is removed from the DOM it will also remove the `OverlayPanel` as well.  This works great and fixes #5710 however in this new ticket there are scenario's where that doesn't work so I made it an option.

Although now it has me wondering if I can check if the `appendTo` is `body` automatically apply this?

tandraschko:
> Basically if the OverlayPanel was in the body and it is updated in an update statement it creates two overlay panels because the old one is still in the DOM and never gets removed.

really? shouldnt the AJAX core just replace by ID?

melloware:
Maybe but when we do `appendTo="@body"` is it really replacing the old one by ID or just appending this new element into the DOM in the body?

tandraschko:
the appendTo logic should remove the old one actually

melloware:
here is a great reproducer showing the issue: 
[pf-5710.zip](https://github.com/primefaces/primefaces/files/14229338/pf-5710.zip)

Click "Alert State" and it will show 1 in the DOM.  Then click "Update OP" and then click "Alert State" it shows now 2 instance in the DOM.

tandraschko:
where are the 2 instances? can't see it in firefox

melloware:
I will double check but his jquery code reports 2 instances right?

tandraschko:
yeah but it checks for selectOneMenu, not overlay panel instances?

maybe its a selectOneMenu problem because it has appendTo and is placed inside the overlayPanel, which also has appendTo logic?

tandraschko:
yeah
1 overlayPanel DOM
2 selectOneMenu DOM

melloware:
Sorry it's not the Dom it's the overlay widget is left hanging around

tandraschko:
what do you mean exactly?

melloware:
Yeah it's definitely that the overlay panel is not registered to "destroy" itself.

tandraschko:
i cant follow
there is 1 overlayPanel DOM and 1 overlayPanel widget instance?

melloware:
Sorry no it looks like 1 OverlayPanel and a 2nd SelectOneMenu.

image

tandraschko:
yeah, thats what i said

we need to really dig deep here and analyze #5710 before
everything else is hacking around a workaround propably

i removed your commit [from ](https://github.com/primefaces/primefaces/pull/11244) in my local code and i still have no idea

if you press update+alert, it says that there are 2 "frm:som" but its not really there in the HTML

might be a timing problem

melloware:
OK I have reverted #5710 in main and re-opened the ticket.

tandraschko:
i think its a general problem with nested appendTo components
not sure if i have enough time to check it

would be great if someone could dig deeper (debug both OP and SOM init,refresh,destroy methods and appendToLogic)

melloware:
i will close this for now as its not solving the underlying problem

fanste:
I just stumpled on this issue on my list. Let me explain, why it has two SelectOneMenu (SOM) but only single OverlayPanel (OP).

The page is opend the first time. You get a fresh DOM where appendTo of the OP will transfer the DOM of the OP(containing the SOM) to the correct position. This is triggert by the JS code of the widget, that is always contained in a script tag below the component DOM.

In short, this looks like this:
```
<div id="OP">
    <div id="SOM"></div>
    <script id="SOM_s">...</script>
</div>
<script id="OP_s">...</script>
```

The AJAX triggers a re-render of the OP (and therefore also SOM, as it is contained in OP). But not directly the OP (then is should be no problem) but by updating the parent of OP.

Now you have two times the upper shown DOM. At the original/updated position of the parent of OP and at the appendTo-moved position. As the DOM is processed top-down, it will first run the script of SOM, which will then find the DOM of SOM at two positions (same ID) and so SOM#jq contains two instances. Next it will run the script of OP. The logic of OP is smart enough  to detect the old DOM and removes it first.

But the SOM is already miss-initialized.

melloware:
@fanste yep we removed this PR from 14.0.0 as it did not fix the actual issue.   We are open to suggestions.

PR #11417, opened 2024-02-10
Add & Fix Indonesian Transalations

ganjarsetia:
This is request from https://github.com/primefaces/primelocale/pull/150#issuecomment-1923769579

Keep in mind that I haven't test it. I'm not running Primefaces locally.

I have question:

1. If someone add more translation in `locale-en.js`. How to automatically notify the owner of the repository that other language should be updated? Maybe using CI/CD?
2. Every time I googled Primereact/Primevue component, most question coming from Primefaces forum. Primefaces is more popular rather than Primereact, etc?


melloware:
This looks great!   And No PF JSF is not more popular than PrimeReact it has just been around since 2009 and was PrimeTek's first library so there is a lot of Google Searches to it.

PR #11453, opened 2024-02-17
Fix #11447: Dashboard var attribute

### LINKED ISSUE #11447: Dashboard: with dynamic panels
### Description

I would like to have a dynamic p:dashboard.

Without `p:dashboard` I have this code:
  ```
 <ui:repeat
            value="#{dashboardWidgetByDashboardNotLazyController.myList}"
             var="dashboardWidget" >

                          <p:panel header="#{dashboardWidget.headline}" closable="true" styleClass="#{dashboardWidget.styleClass}" rendered="# 
                                  {dashboardWidget.uniqueName == 'test'}">
                               // Content
                         </panel>
      
                        <p:panel header="#{dashboardWidget.headline}" closable="true" styleClass="{dashboardWidget.styleClass}" rendered="# 
                              {dashboardWidget.uniqueName == 'test2'}">
                            // Content
                       </panel>
      </ui:repeat>
```

At the end I got a nice dashboard with custom headline, custom color etc.
But I have not the possibility like in `p:dashboard` to drag and drop the widget etc.

I´m not sure if this is the right approach to do this. My idea would be:

Add `var` property to `DashboardWidget` like:
`private Object object;`

Afterwards you can use like this:

```
 <p:dashboard
            value="#{myBean.myDashboardModel}"
             var="dashboardWidget" >
```

### Additional context

_No response_

melloware:
Fix #11447: Dashboard var attribute

### DIFF
--- a/docs/14_0_0/components/dashboard.md
+++ b/docs/14_0_0/components/dashboard.md
@@ -29,6 +29,7 @@ Dashboard provides a portal like layout with drag&drop based reorder capabilitie
 | responsive | false | Boolean | In responsive mode, allows use of PrimeFlex CSS to control widget sizes and responsiveness.
 | style | null | String | Inline style of the dashboard container
 | styleClass | null | String | Style class of the dashboard container
+| var | null | String | Name of collection iterator which will be the value of DashboardWidget.getValue().
 
 ## Getting started with Dashboard
 Dashboard is backed by a DashboardModel and consists of panel components.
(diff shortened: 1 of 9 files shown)


melloware:
@tandraschko I love your `ComponentUtils.executeInRequestScope` i wonder if we should refactor more code in 15 to use it?  We do this request scope variable code in a LOT of places.  I an create a backlog ticket for 15.0.0 if you think its worth it?

tandraschko:
+1

i like the method too but the exception handling isnt nice
so i would not do a mass-refactor here

PR #11457, opened 2024-02-18
Fix #11452 - DataTable: LazyDataModel.load gets invoked multiple time…

### LINKED ISSUE #11452: DataTable: LazyDataModel.load gets invoked multiple times when datatable is lazy and has headerRow
### Describe the bug

DataTable component invokes LazyDataModel::load multiple times (one for each item in the current page of the datatable) when the datatable is lazy and has a headerRow.

Please, check the attached zip that reproduces the problem and shows one workaround to the issue.
To execute the reproducer:
```
mvn jetty:run
```
And open the following page in the browser
```
http://localhost:8080/primefaces-test
```

[primefaces-test.zip](https://github.com/primefaces/primefaces/files/14315550/primefaces-test.zip)

### Reproducer

The ProblematicLazyDataModel::load method below will be executed multiple times (one for each returned item in the pagination). With datatable configured with `rows=5`, this is the server output (showing that querying a real database would result in 5 queries being executed with weird pagination parameters):
```
ProblematicLazyDataModel.load(first=0, size=5) executed in the fake database
ProblematicLazyDataModel.load(first=0, size=1) executed in the fake database
ProblematicLazyDataModel.load(first=1, size=2) executed in the fake database
ProblematicLazyDataModel.load(first=2, size=3) executed in the fake database
ProblematicLazyDataModel.load(first=3, size=4) executed in the fake database
```

Please, check provided uploaded primefaces-test.zip for a working and complete example.

Sample xhtml:
```xhtml
<!DOCTYPE html>

    <h:head>
    </h:head>
    <h:body>

        <h1>Problem</h1>

                <p:dataTable value="#{problematicLazyDataModel}" var="obj"
                                         lazy="true" paginator="true"
                                         paginatorPosition="bottom" rows="5">

                        <p:headerRow field="artist" expandable="false">
                                <p:column colspan="3">#{obj.artist}</p:column>
                        </p:headerRow>

                        <p:column headerText="ID">
                                <h:outputText value="#{obj.id}" />
                        </p:column>

                        <p:column headerText="Song">
                                <h:outputText value="#{obj.name}" />
                        </p:column>

                        <p:column headerText="Released">
                                <h:outputText value="#{obj.released}" />
                        </p:column>

                </p:dataTable>

    </h:body>
</html>
```

Sample problematic datamodel:
```java
package org.primefaces.test;

import java.util.List;
import java.util.Map;

import javax.inject.Named;

import org.primefaces.model.FilterMeta;
import org.primefaces.model.LazyDataModel;
import org.primefaces.model.SortMeta;

@Named
@ViewScoped
public class ProblematicLazyDataModel extends LazyDataModel<TestObject> {

        @Override
                return FakeDB.count();
        }

        @Override
        public List<TestObject> load(int first, int pageSize, Map<String, SortMeta> sortBy, Map<String, FilterMeta> filterBy) {
                System.out.println(String.format("%s.load(first=%d, size=%d) executed in the fake database", this.getClass().getSimpleName(), first, pageSize));
                return FakeDB.find(first, pageSize);
        }

}
```

### Expected behavior

DataTable should invoke the LazyDataModel::load method once, like que WorkaroundLazyDataModel example below.
The output:
```
WorkaroundLazyDataModel.load(first=0, size=5) executed in the fake database
```

WorkaroundLazyDataModel code, with overriden getRowData method (the workaround):
```java
package org.primefaces.test;

import java.util.List;
import java.util.Map;

import javax.inject.Named;

import org.primefaces.model.FilterMeta;
import org.primefaces.model.LazyDataModel;
import org.primefaces.model.SortMeta;

@Named
@ViewScoped
public class WorkaroundLazyDataModel extends LazyDataModel<TestObject> {

        @Override
                return FakeDB.count();
        }

        @Override
        public List<TestObject> load(int first, int pageSize, Map<String, SortMeta> sortBy, Map<String, FilterMeta> filterBy) {
                System.out.println(String.format("%s.load(first=%d, size=%d) executed in the fake database", this.getClass().getSimpleName(), first, pageSize));
                return FakeDB.find(first, pageSize);
        }

        /*
         * WORKAROUND
         *
         * */
        @Override
        public TestObject getRowData(int rowIndex, Map<String, SortMeta> sortBy, Map<String, FilterMeta> filterBy) {

                if (rowIndex < 0 || getPageSize() == 0) {
                        return null;
                }

                int pageRowIndex = (rowIndex % getPageSize());
                List<TestObject> wrappedData = getWrappedData();
                if (wrappedData != null && wrappedData.size() > pageRowIndex) {
                        return wrappedData.get(pageRowIndex);
                }

                // couldnt load from cache, trying the old way
                return super.getRowData(rowIndex, sortBy, filterBy);
        }

}
```

### PrimeFaces edition

Community

### PrimeFaces version

13.0.5

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3.21

### Java version

1.8.0_361

### Browser(s)

_No response_

Rapster:
…s when datatable is lazy and has header

Fix #11452

### DIFF
--- a/docs/migrationguide/14_0_0.md
+++ b/docs/migrationguide/14_0_0.md
@@ -57,6 +57,7 @@
   * Using Column custom `sortFunction` signature change now requires a third parameter `SortMeta` like `public int sortByModel(Object car1, Object car2, SortMeta sortMeta)`
   * `JpaLazyDataModel` renamed to `JPALazyDataModel`
   * `JPALazyDataModel`: deprecated constructors, use `JPALazyDataModel.builder()` instead
+  * 'LazyDataModel': 'getRowData()` overload has been renamed `loadRow()`
 
 ### DataTable Selection
 
--- a/docs/migrationguide/14_0_0.md
+++ b/docs/migrationguide/14_0_0.md
@@ -57,7 +57,7 @@
   * Using Column custom `sortFunction` signature change now requires a third parameter `SortMeta` like `public int sortByModel(Object car1, Object car2, SortMeta sortMeta)`
   * `JpaLazyDataModel` renamed to `JPALazyDataModel`
   * `JPALazyDataModel`: deprecated constructors, use `JPALazyDataModel.builder()` instead
-  * 'LazyDataModel': 'getRowData()` overload has been renamed `loadRow()`
+  * 'LazyDataModel': 'getRowData()` overload has been renamed `loadOne()` to load a single row
 
 ### DataTable Selection
 
--- a/primefaces/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
@@ -1085,7 +1085,6 @@ public class DataTableRenderer extends DataRenderer {
     protected void encodeRows(FacesContext context, DataTable table, int first, int last, int columnStart, int columnEnd) throws IOException {
         List<SummaryRow> summaryRows = table.getSummaryRows();
         HeaderRow headerRow = table.getHeaderRow();
-        ELContext elContext = context.getELContext();
 
         SortMeta sort = table.getHighestPriorityActiveSortMeta();
         boolean encodeHeaderRow = headerRow != null && headerRow.isEnabled() && sort != null;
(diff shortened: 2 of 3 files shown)


melloware:
Thanks for fixing this!

tandraschko:
i assume we have a testcase for the initial bug, where we need to load this "single items"?

melloware:
I will look but I am pretty sure an integration test was added when the original fix was made

Rapster:
We do have: https://github.com/primefaces/primefaces/commit/86113f3f11570eb1494d91555579a2635beda2ed

tandraschko:
i dont have a problem to rename the method but we need a hint in the migration guide then, it's a public API

melloware:
I added the migration guide note and changed it to `loadRow` from `loadOne` to be more consistent. I also JavaDoc'ed the method.

Rapster:
Considering we have a load mehod and not a loadRows method, I prefer loadOne which is more explicit and more consistent.

melloware:
Hmmm but `loadOne` what?  Load One Cell?  One Table?  One Row?

Rapster:
You're in LazyDataModel, what else could it be? You don't see JPARepository from Spring for instance having a `findRow` method. I don't think there is any confusion here... It loads one element, simple as that...

melloware:
OK I reverted I don't feel like strongly about it i just thought it would be more clear to downstream users.

melloware (approved):

PR #11474, opened 2024-02-20
Cleanup documentation

tuerker:
Typo fixes and consistent usage of e.g. („exempli gratia“) followed by a comma

Consolidate the two "File Upload" subsections in the PF14 migration document


jepsar:
I believe the jsdocs are generated from JS sources.

melloware:
They are the JS docs should be fixed right in the component as they are then generated.

melloware (approved):

melloware:
Its OK they will get overwritten when I generate the 14.0.0 docs so its fine.

PR #11510, opened 2024-02-22
Fix #11505: Fileupload refactor messageTemplate

### LINKED ISSUE #11505: FileUpload: attribute messageTemplate has no effect anymore
### Describe the bug

Using the attribute messageTemplate in the fileUpload component does not change the way the added file message is formatted.

For instance, using 

```
    <p:fileUpload mode="advanced"
                  multiple="false"
                  messageTemplate="__{size} {name}__"
                  update="messages"
                  listener="#{testView.handleFileUpload}" />

```

i would expect that the resulting message prints the size and the name of the uploaded file enclosed by two underscores. Currently, is has no effect:

image

### Reproducer

Here is the repoducer:  [p-fileUpload.zip](https://github.com/primefaces/primefaces/files/14376252/p-fileUpload.zip)

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

14.0.0-RC1

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

3

### Java version

11

### Browser(s)

_No response_

melloware:
Fix #11505: Deprecate FileUpload messageTemplate

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/fileupload/2-fileupload.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/fileupload/2-fileupload.js
@@ -115,7 +115,7 @@ PrimeFaces.widget.FileUpload = PrimeFaces.widget.BaseWidget.extend({
      * Regular expression that matches image files for which a preview can be shown.
      * @type {RegExp}
      */
-    IMAGE_TYPES: /(\.|\/)(gif|jpe?g|png)$/i,
+    IMAGE_TYPES: /(\.|\/)(gif|jpeg|jpg|png)$/i,
 
     /**
      * @override
--- a/docs/10_0_0/components/fileupload.md
+++ b/docs/10_0_0/components/fileupload.md
@@ -22,7 +22,7 @@ powered rich solution with graceful degradation for legacy browsers.
 | --- | --- | --- | --- |
 | id | null | String | Unique identifier of the component.
 | accept | null | String | Filters files in native file browser dialog.
-| allowTypes | null | String | Regular expression for accepted file types, e.g., /(\\.\|\\/)(gif\|jpe?g\|png)$/
+| allowTypes | null | String | Regular expression for accepted file types, e.g., /(\\.\|\\/)(gif\|jpeg|jpg\|png)$/
 | auto | false | Boolean | When set to true, selecting a file starts the upload process implicitly.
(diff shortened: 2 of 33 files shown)


melloware:
@tandraschko just double checking but i think the only reason the tests are now failing is it was counting on the widget holding these messsage but they were never being used and all of the display happens in the messages component now.  I ran the showcase with this code and all validations and errors appear to be happening.

primefaces/src/main/resources/META-INF/resources/primefaces/fileupload/2-fileupload.js
```diff
@@ -195,26 +186,8 @@ PrimeFaces.widget.FileUpload = PrimeFaces.widget.BaseWidget.extend({
             add: function(e, data) {
                 $this.chooseButton.removeClass('ui-state-hover ui-state-focus');
 
-                if($this.fileAddIndex === 0) {
-                    $this.clearMessages();
-                }
-
-                // we need to fake the filelimit as the jquery-fileupload input always only contains 1 file
-                var fileLimit = data.fileInput.data('p-filelimit');
```
tandraschko: Did you test filelimit with advanced fileupload?

melloware: Yep and it it stopped it at 3 when I tried to upload 4 files and the message was correct.

melloware: image

melloware:
OK let me test some more today with simple mode.

melloware:
OK it is use in Simple mode so I have done the following:

- Update the docs to reflect what `messageTemplate` actually does
- Only send `messageTemplate` in Simple mode now
- Updated MD docs to mark props deprecated with ~~strikethrough~~
- Update Taglib XML docs for those props

Next I think I need to fix the tests as the widget is no longer storing local messages.

melloware:
OK @tandraschko @tuerker I got to the bottom of it.  `messageTemplate` is used in Simple mode to display the name and in Advanced mode only in a few validation scenarios which is why the Integration tests were failing.

So I have updated the docs accordingly.

melloware:
Basically this whole PR is just a documentation update.

melloware:
Agreed @tandraschko let me know when that other PR is ready so I can fix my PR after that one is merged?

PR #11516, opened 2024-02-23
#11506 LazyDataModel: no need to count results if first==0 && lazyData.size < pageSize

### LINKED ISSUE #11506: LazyDataModel: no need to count if first == 0 and lazyData.size < pageSize
### Description

Regarding DataView with lazy=true: 
If first == 0 there is no need to count at the beginning and we can skip the calculateFirst method. 
After the load, if the size of the loaded data is less than the pageSize, there is no need to count.

This is the case for DataView, but I believe its the same case for DataTable, DataList and DataGrid.

Count can be an expensive operation with LazyDataModel, especially when no filters are applied (first load).

It can be achieved playing with the count attribute, but it's awkward and it sometimes breaks on upgrades.

WDYT?

If first == 0 and loaded data size < pageSize no count() operation should be performed on the lazyDataModel

### Additional context

_No response_

cocorossello:
This is the main idea, I have to see how to effectively test it

Fix #11506

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/autocomplete/AutoComplete.java
+++ b/primefaces/src/main/java/org/primefaces/component/autocomplete/AutoComplete.java
@@ -169,7 +169,13 @@ public class AutoComplete extends AutoCompleteBase {
                             .order(SortOrder.ASCENDING)
                             .build());
             suggestions = lazyModel.load(0, getMaxResults(), sortBy, searchFilter);
-            suggestionsCount = lazyModel.count(searchFilter);
+            List suggestionsList = (List) suggestions;
+            if (suggestionsList.size() < getMaxResults()) {
+                suggestionsCount = suggestionsList.size();
+            }
+            else {
+                suggestionsCount = lazyModel.count(searchFilter);
+            }
         }
         else {
             FacesContext facesContext = getFacesContext();
--- a/primefaces/src/main/java/org/primefaces/component/autocomplete/AutoComplete.java
+++ b/primefaces/src/main/java/org/primefaces/component/autocomplete/AutoComplete.java
@@ -168,9 +168,10 @@ public class AutoComplete extends AutoCompleteBase {
                             .field(field)
                             .order(SortOrder.ASCENDING)
                             .build());
-            suggestions = lazyModel.load(0, getMaxResults(), sortBy, searchFilter);
+            int maxResults = getMaxResults();
+            suggestions = lazyModel.load(0, maxResults, sortBy, searchFilter);
             List suggestionsList = (List) suggestions;
-            if (suggestionsList.size() < getMaxResults()) {
+            if (suggestionsList.size() < maxResults) {
                 suggestionsCount = suggestionsList.size();
             }
             else {
--- a/primefaces/src/main/java/org/primefaces/component/datagrid/DataGrid.java
+++ b/primefaces/src/main/java/org/primefaces/component/datagrid/DataGrid.java
@@ -92,7 +92,7 @@ public class DataGrid extends DataGridBase {
             }
             List<?> data = lazyModel.load(getFirst(), getRows(), Collections.emptyMap(), Collections.emptyMap());
             if (data.size() < getRows() && getFirst() == 0) {
-                setRows(data.size());
+                lazyModel.setRowCount(data.size());
             }
             else if (getFirst() == 0) {
                 lazyModel.setRowCount(lazyModel.count(Collections.emptyMap()));
--- a/primefaces/src/main/java/org/primefaces/component/datagrid/DataGrid.java
+++ b/primefaces/src/main/java/org/primefaces/component/datagrid/DataGrid.java
@@ -86,10 +86,17 @@ public class DataGrid extends DataGridBase {
         if (model instanceof LazyDataModel) {
             LazyDataModel lazyModel = (LazyDataModel) model;
 
-            lazyModel.setRowCount(lazyModel.count(Collections.emptyMap()));
-            calculateFirst();
-
+            if (getFirst() > 0) {
+                lazyModel.setRowCount(lazyModel.count(Collections.emptyMap()));
+                calculateFirst();
+            }
             List<?> data = lazyModel.load(getFirst(), getRows(), Collections.emptyMap(), Collections.emptyMap());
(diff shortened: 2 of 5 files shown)


melloware:
OK it looks like 1 integration test failed.

```
Error:  Tests run: 10, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 22.03 s <<< FAILURE! -- in org.primefaces.integrationtests.datatable.DataTable006Test
Error:  org.primefaces.integrationtests.datatable.DataTable006Test.lazySelectionWithFiltering(Page) -- Time elapsed: 3.238 s <<< ERROR!
java.lang.IndexOutOfBoundsException: Index 1 out of bounds for length 0
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:248)
	at java.base/java.util.Objects.checkIndex(Objects.java:374)
	at java.base/java.util.ArrayList.get(ArrayList.java:459)
	at org.primefaces.selenium.component.DataTable.getRow(DataTable.java:63)
	at org.primefaces.selenium.component.base.AbstractTable.getCell(AbstractTable.java:57)
	at org.primefaces.integrationtests.datatable.DataTable006Test.lazySelectionWithFiltering(DataTable006Test.java:329)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at java.base/java.util.concurrent.RecursiveAction.exec(RecursiveAction.java:189)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183)
```

primefaces/src/main/java/org/primefaces/component/autocomplete/AutoComplete.java
```diff
@@ -169,7 +169,13 @@ public void broadcast(javax.faces.event.FacesEvent event) throws javax.faces.eve
                             .order(SortOrder.ASCENDING)
                             .build());
             suggestions = lazyModel.load(0, getMaxResults(), sortBy, searchFilter);
-            suggestionsCount = lazyModel.count(searchFilter);
+            List suggestionsList = (List) suggestions;
+            if (suggestionsList.size() < getMaxResults()) {
```
melloware: We should put getMaxResults into a local variable so the getter is not called multiple times.

primefaces/src/main/java/org/primefaces/component/datagrid/DataGrid.java
```diff
@@ -86,10 +86,17 @@ public void loadLazyData() {
         if (model instanceof LazyDataModel) {
             LazyDataModel lazyModel = (LazyDataModel) model;
 
-            lazyModel.setRowCount(lazyModel.count(Collections.emptyMap()));
-            calculateFirst();
-
+            if (getFirst() > 0) {
```
melloware, line 89: We should put `getFirst()` into a local variable so the getter is not called multiple times.

primefaces/src/main/java/org/primefaces/component/datalist/DataList.java
```diff
@@ -117,10 +117,18 @@ public void loadLazyData() {
         if (model instanceof LazyDataModel) {
             LazyDataModel<?> lazyModel = (LazyDataModel) model;
 
-            lazyModel.setRowCount(lazyModel.count(Collections.emptyMap()));
-            calculateFirst();
+            if (getFirst() > 0) {
```
melloware, line 120: Same extract getFirst() into a local var

primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java
```diff
@@ -496,6 +495,9 @@ public void loadLazyScrollData(int offset, int rows) {
             setFirst(offset);
         }
 
+        if (getFirst() > 0) {
```
melloware, line 498: Same

primefaces/src/main/java/org/primefaces/component/dataview/DataView.java
```diff
@@ -149,10 +149,18 @@ public void loadLazyData() {
         if (model instanceof LazyDataModel) {
             LazyDataModel lazyModel = (LazyDataModel) model;
 
-            lazyModel.setRowCount(lazyModel.count(Collections.emptyMap()));
-            calculateFirst();
+            if (getFirst() > 0) {
```
melloware, line 152: Same

melloware (changes requested):

primefaces/src/main/java/org/primefaces/component/datagrid/DataGrid.java
cocorossello: It's a bit tricky because first will change in calculateFirst. Maybe calculateFirst should return first and in the only case that the boolean value is checked (the DataTable one), store the old first and compare it. 

WDYT?

melloware: No you can leave as is

melloware (approved):

Rapster:
Not sure yet... Anyway, at best this should go in 14.0.1

tandraschko:
why 14.0.1? current release is RC1, next will be either RC2 or final

Rapster:
Because it's not ideal to introduce enhancement during major release specially when it comes to DataTable (and this PR can definitely wait don't you think?)

tandraschko:
IMO its not good to put such enhancements in a bugfix release, better 15.0.0 then

Rapster:
Yeah whatever, as long it's not in the current release I'm cool with that

Rapster:
PR was failing with recent changes, it has been reverted. Needs some investigation as of why it's been failing in 15

PR #11543, opened 2024-02-29
Fix #11540: Picklist trigger input changes on transfer

### LINKED ISSUE #11540: PickList: fire 'change' event
### Description

`<p:pickList>` doesn't fire any change event when an item is transferred.

One way would be to add to `fireTransferEvent` function in `picklist.js`:
```
this.targetInput.trigger("change");
// and probably also this.sourceInput.trigger("change")?
```

### Additional context

This way we can safely reuse existing global form/input change event listeners on them.

melloware:
Fix #11540: Picklist trigger input changes on transfer

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/picklist/picklist.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/picklist/picklist.js
@@ -1171,7 +1171,7 @@ PrimeFaces.widget.PickList = PrimeFaces.widget.BaseWidget.extend({
 
         // #11540: trigger change events on inputs for form handling
         this.targetInput.trigger("change");
-        this.sourceInput.trigger("change")?
+        this.sourceInput.trigger("change");
     },
 
     /**
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/picklist/picklist.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/picklist/picklist.js
@@ -1169,9 +1169,7 @@ PrimeFaces.widget.PickList = PrimeFaces.widget.BaseWidget.extend({
             $this.updateButtonsState();
         }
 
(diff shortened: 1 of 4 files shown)


FlipWarthog (approved):

primefaces/src/main/resources/META-INF/resources/primefaces/picklist/picklist.js
```diff
@@ -1168,6 +1168,10 @@ PrimeFaces.widget.PickList = PrimeFaces.widget.BaseWidget.extend({
             $($this.jqId + ' ul').sortable('enable');
             $this.updateButtonsState();
         }
+
+        // #11540: trigger change events on inputs for form handling
+        this.targetInput.trigger("change");
+        this.sourceInput.trigger("change")?
```
BalusC: Question mark should be a semicolon.

jepsar:
```suggestion
        this.sourceInput.trigger("change");
```

melloware:
Thanks!

jepsar:
Shouldn't this be added to the docs?

BalusC (approved):

tandraschko:
if "transfer" triggers onchange, "reorder" should to
so it means if a user needs different listeners he uses transfer+reordern, change catches both

BalusC:
You could trigger a namespaced event.

jepsar:
I was thinking of https://github.com/primefaces/primefaces/blob/master/docs/14_0_0/components/picklist.md#ajax-behavior-events not sure if we should mention this there.

melloware:
Hmm this just triggers the client side change events on the input but doesn't send an AJAX `change` event.  However I did realize that no `defaultEvent` was defined on this component and there should be so I made it `transfer`

Issue #11569, opened 2024-03-05
Provide a JS util to overwrite a local specific setting like firstDayOfWeek for all locales on the current view


tandraschko:
one of my customers have a production line in a specific country and they have the requiremenent to set the firstDayOfWeek for all locales, indepedent of the current browser locale

in theory they could overwrite all via:
```
PrimeFaces.locales["en"].firstDayOfWeek = 0
PrimeFaces.locales["de"].firstDayOfWeek = 0
PrimeFaces.locales["fr"].firstDayOfWeek = 0
....
```

but thats just to much overhead

i will introduce a util method like:
```
PrimeFaces.overwriteAllLocalesValue('firstDayOfWeek', 0);
```

this will loop all PrimeFaces.locales and overwrite the value in the current view.

any idea for a better method name?

melloware:
Sounds good to me the only other name u can think of if "forceLocaleValue" but not sure it's any better

jepsar:
`setGlobalLocaleValue`?

PR #11573, opened 2024-03-05
Fix #10762 - Chronoline: generate unique client ids per iteration

### LINKED ISSUE #10762: Chronoline: Inherit UIData
### Describe the bug

In the following example:
```xhtml
<p:chronoline value="#{searchChatBean.conversations}" var="conv">
    <p:commandLink action="#{searchChatBean.switchConversation(conv)}" update=":chat" process="@this">
          <h6 class="mb-0">#{conv.title}</h6>
    </p:commandLink>
</p:chronoline>
```

When `searchChatBean.switchConversation(conv)`  called `conv` is always `null`. However, replacing `p:chronoline` with `ui:repeat` fixes the issue.

### Reproducer

_No response_

### Expected behavior

Item iteration works in the same way as in `ui:repeat`.

### PrimeFaces edition

Community

### PrimeFaces version

13.0.1

### Theme

_No response_

### JSF implementation

MyFaces

### JSF version

4.0.1

### Java version

17

### Browser(s)

_No response_

Rapster:
Fix #10762

### DIFF
--- a/primefaces-showcase/src/main/java/org/primefaces/showcase/view/data/ChronolineView.java
+++ b/primefaces-showcase/src/main/java/org/primefaces/showcase/view/data/ChronolineView.java
@@ -25,6 +25,8 @@ package org.primefaces.showcase.view.data;
 
 import javax.annotation.PostConstruct;
 import javax.enterprise.context.RequestScoped;
+import javax.faces.application.FacesMessage;
+import javax.faces.context.FacesContext;
 import javax.inject.Named;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


melloware (approved):
Nice!  I would like to put this in 14.0.0

Rapster:
inheriting from UIData is overkill (UIData is simply a custom UIRepeat which supports row/column and has special options) which is not needed here.

tandraschko:
are you sure about that?
AFAIR MyFaces UIRepeat is just a copy of UIData and vice-versa
both have almost the same LOC (2100 vs 2200)

Rapster:
It would make sense at least UIData inherits UIRepeat... Anyway, I switch to PF UIData (the idea would be to get rid of UITabPanel)

melloware (approved):

tandraschko:
looks good to me
@melloware if you can test the showcase, +1 to put it into 14.0.0-RC2

i will work on #11569 today, i'm fine if you could release a RC2 on weekend and do the final in ~2 weeks
did we forget to rename milestones after releasing RC1?

melloware:
Great plan @tandraschko i will do an RC2 this weekend after your PR for the Locale change.

PR #11583, opened 2024-03-06
Fix #11582: getResourceScriptURI() iterate until found

### LINKED ISSUE #11582: Core: getResourceScriptURI returning incorrect URI
### Describe the bug

I wrote this code and introduced this bug never thinking it through.  Discovered at a client using PFE Document Viewer.

Open F12 client in the showcase and type `PrimeFaces.resources.getResourceScriptURI();` it returns:

```
'/showcase/javax.faces.resource/jquery/jquery.js.xhtml?ln=primefaces&v=13.0.5'
```

This is correct and we then parse the URL looking for `v=`.  This works for any PrimeFaces or OmniFaces URL which always has a `v=` version attached.  However in my app the code was returning this...

```
`/weather/javax.faces.resource/jsf.js.jsf?ln=javax.faces&amp;stage=Development`
```

The first URL was the jsf.js URL which does not have a `v=` version identifier on it.  The bug is because my code was doing `first()` instead of looping for `v=` to find the first proper URI.

### PrimeFaces edition

Community

### PrimeFaces version

13.0.7

### Theme

ALL

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

11

### Browser(s)

_No response_

melloware:
Fix #11582: getResourceScriptURI() iterate until found

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.resources.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.resources.js
@@ -112,16 +112,26 @@ if (!PrimeFaces.resources) {
            * @return {string} The first JavasScript resource URI.
            */
           getResourceScriptURI : function() {
-             if (!PrimeFaces.resources.SCRIPT_URI) {
-                PrimeFaces.resources.SCRIPT_URI =
-                   $('script[src*="/' + PrimeFaces.RESOURCE_IDENTIFIER + '/"]').first().attr('src');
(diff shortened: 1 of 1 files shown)


tandraschko:
this function would really need some unit tests :/

melloware:
Agreed in fact this whole class can use some unit tests as every time I stare at it I have to relearn what it does.   Let me see if I can get Jest working maybe?

tandraschko:
that would be great!

PR #11597, opened 2024-03-07
Fix #11595: CSP wrap custom FacesException

### LINKED ISSUE #11595: CSP: create a subclass of FacesException?
Premise: one of our clients run some sort of monitoring tools on one of our application, which is based on Faces 4/PrimeFaces 13 with CSP enabled.
I don't know exactly what this tools does, basically it makes web requests, the problem is that some those requests are cluttering the logs, the reason is:

- CSP is enabled in the application
- there's a stateless login page (`<f:view transient="true">`)
- the tools makes postback requests without sending back the `primefaces.nonce` param
- an exception is thrown: https://github.com/primefaces/primefaces/blob/d678b4ce99a6a8f67b340297ac81b1b40c1fb2bf/primefaces/src/main/java/org/primefaces/csp/CspState.java#L86
- the exception is logged by both Mojarra and PrimeExceptionHandler

Double logging by Mojarra has been fixed (https://github.com/eclipse-ee4j/mojarra/pull/5411), I would like PrimeExceptionHandler not to log those exceptions too, so my idea is creating a new subclass of `FacesException` (`CspException` or something like that) and use it from `CspState` so we can [exclude it from logging](https://primefaces.github.io/primefaces/13_0_0/#/core/errorhandling?id=exception-logging).

TL;DR
I want to exclude CSP-related exceptions from logging, so my idea is making CspState throw a new subclass of FacesException.

I see no downsides in this, so if you guys are ok with it a can prepare a PR targeting 14.x

melloware:
Fix #11595: CSP wrap custom FacesException

### DIFF
--- a/primefaces/src/main/java/org/primefaces/csp/CspException.java
+++ b/primefaces/src/main/java/org/primefaces/csp/CspException.java
@@ -1,0 +1,47 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2024 PrimeTek Informatics
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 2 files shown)


NicolaIsotta (approved):

tandraschko:
+1

PR #11649, opened 2024-03-18
Fix #11640 - Core: introduce PrimeUIData (replacement of UIData)

### LINKED ISSUE #11640: UIData: split in 2 classes and rename as PrimeUIData
### Description

Current `UIData` is a mix bag of Mojarra, MyFaces and PF impl. It gets tough to follow how PF impl is different from JSF impl and why... 

Split UIData into 2 classes and rename UIData as PrimeUIData

### Additional context

Also this might help to port fixes from Mojarra

Rapster:
Fix #11640


Rapster:
WDYT about the name UIDataFacesPatch...? Maybe UIDataPatch sounds simpler and better?

melloware:
Either  `UIDataOverride` or `UIDataPatch` is good with me

melloware (approved):

primefaces/src/main/java/org/primefaces/component/api/PrimeUIData.java
```diff
@@ -0,0 +1,521 @@
...
+            }
+        }
+    }
+
+    protected boolean shouldSkipChildren(FacesContext context) {
+        return false;
+    }
+
+    @Override
+    public boolean invokeOnComponent(FacesContext context, String clientId, ContextCallback callback)
```
tandraschko: this could acutally inside UIDataPatch - or even removed if its finally fixed in all Mojarra versions, it implemented correctly in MyFaces

tandraschko:
left one small comment, everything else is fine and great work

melloware:
@Rapster one small conflict and then i think this is ready to merge as it has the approval of @tandraschko

melloware (approved):

PR #11661, opened 2024-03-20
Fix #11660: Spinner round properly in case of string value

### LINKED ISSUE #11660: Spinner: MAX value dont work up to 1000 when ROUND is true
### Describe the bug

When I use **Spinner** with **round="true"**      the value of **max** becomes always 1000
And if you pulse increment (+) beyond 1000, the value becomes 1, 2, 3, and then blocks the component

### Reproducer

 <p:spinner min="0" max="3000"  round="true"/>

Dont work beyond 1000

### Expected behavior

 <p:spinner min="0" max="3000"  round="true"/>

Dont work beyond 1000

### PrimeFaces edition

Community

### PrimeFaces version

13.0.1

### Theme

saga

### JSF implementation

Mojarra

### JSF version

10.0.0

### Java version

20.0.1

### Browser(s)

_No response_

melloware:
Fix #11660: Spinner round properly in case of string value

### DIFF
--- a/primefaces-integration-tests/src/main/webapp/spinner/spinner005.xhtml
+++ b/primefaces-integration-tests/src/main/webapp/spinner/spinner005.xhtml
@@ -14,7 +14,7 @@
 
         <h:form id="form">
             <!-- GitHub #6521: new 'round' attribute -->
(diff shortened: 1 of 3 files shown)


melloware:
Added integration test

Issue #11521, opened 2024-02-27
DataTable: Paginator After upgrade to 13.0.6  always uses first=0 in .load() method


AndrewG10i:
### Describe the bug

Upgraded PrimeFaces from 13.0.5 to 13.0.6 and noticed that DataTable with LazyDataModel stopped paginating properly. Investigation revealed that `.load()` method always called with `first=0` on pagination to other pages. Reverting back to 13.0.5 resolves the issue without any other code changes.

I am running Payara 6.2024.2 and PF with Jakarta classifier.

### PrimeFaces edition

None

### PrimeFaces version

13.0.6

### JSF implementation

Mojarra

### JSF version

3.0.2

### Java version

17

melloware:
See: https://github.com/primefaces/primefaces/pull/11377

It's only for virtual scroll and live scroll. So it should not affect paginator at all.

AndrewG10i:
This is the code (DataTable declaration) I have:
```
<p:dataTable id="usersTable" value="#{dirCntroller.usersLazyDS}" lazy="true" var="rowData" rowIndexVar="rowIndex"
             paginator="true" rows="25" rowsPerPageTemplate="25,50,100" paginatorPosition="top" paginatorAlwaysVisible="true"
             paginatorTemplate="{CurrentPageReport} {JumpToPageDropdown} {FirstPageLink} {PreviousPageLink} {PageLinks} {NextPageLink} {LastPageLink} {RowsPerPageDropdown}"
             reflow="true" emptyMessage="#{labels['general.dataTable.noRecordsFound']}" filterDelay="800" styleClass="usersTable" widgetVar="usersTableWV">
```

I will try to prepare reproducer. Meanwhile decided to report the issue so in case someone else have it they can confirm as well.

melloware:
We have tons of integrations tests that test pagination and datatable that all passed so if this is a bug it means there is a gap in the integration tests.

AndrewG10i:
Okay, so far cannot reproduce this issue, reproducer works fine. Will investigate further what exactly triggers issue in my app and in case issue found in the PF lib - will reopen this ticket again. But for now closing it. Thanks!

melloware:
Thanks yes please let me know what you find out!

AndrewG10i:
Okay, after spending several hours on this, I finally found root cause. ;))
[...]

PR #11402, opened 2024-02-07
Fix #11401: calculate first row even if rowCount=0

jungm:
Fixes #11401

Issue #11401, opened 2024-02-07
DataTable: DefaultLazyDataModel throws IllegalArgumentException with external filter


jungm:
### Describe the bug

`DefaultLazyDataModel#load()` can cause an IllegalArgumentException when the user is on a page > 1 and `DefaultLazyDataModel#filter` filters out _all_ elements

### Reproducer

Bean:
```java
@Named
@ViewScoped
public class DataTable044 implements Serializable {
    @Getter @Setter private String programmingLanguageFilter;
    @Getter private LazyDataModel<ProgrammingLanguage> lazyDataModel;

    @Inject private ProgrammingLanguageService service;

    @PostConstruct
    public void init() {
        lazyDataModel = DefaultLazyDataModel.<ProgrammingLanguage>builder()
                .valueSupplier(service::getLangs)
                .filter((ctx, value, filter, locale) -> {
                    if (programmingLanguageFilter == null) {
                        return true;
                    }

                    return ((ProgrammingLanguage) value).getName().contains(programmingLanguageFilter);
                }).build();
    }
}
```

XHTML:
```xhtml
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml"
      xmlns:h="http://xmlns.jcp.org/jsf/html"
      xmlns:f="http://xmlns.jcp.org/jsf/core"
      xmlns:p="http://primefaces.org/ui">

<f:view contentType="text/html;charset=UTF-8" encoding="UTF-8">
    <h:head>

    </h:head>

    <h:body>

        <h:form id="form">
            <p:messages id="msgs" showDetail="true">
                <p:autoUpdate/>
            </p:messages>

            <p:inputText id="filterInput" value="#{dataTable044.programmingLanguageFilter}">
                <p:ajax process="@this" update="datatable"/>
            </p:inputText>

            <p:dataTable id="datatable" value="#{dataTable044.lazyDataModel}" paginator="true" rows="3" var="lang" >
                <p:column field="id"/>
                <p:column field="name"/>
                <p:column field="firstAppeared"/>
                <p:column field="type"/>
            </p:dataTable>
        </h:form>

    </h:body>
</f:view>

</html>
```

Test:
```java
public class DataTable044Test extends AbstractDataTableTest {
    @Test
    @DisplayName("first index is properly reset when table is filtered externally")
    public void test(Page page) throws InterruptedException {
        // Arrange
        page.dataTable.selectPage(2);

        // Act
[...]

PR #11672, opened 2024-03-24
Fix #11671 - DefaultLazyDataModel: make it compatible with other lazy-able components

### LINKED ISSUE #11671: DefaultLazyDataModel: make it compatible with other lazy-able components
### Description

`DefaultLazyDataModel` should be compatible with other components like `DataScroller`

see description

### Additional context

_No response_

Rapster:
Fix #11671

### DIFF
--- a/primefaces/src/main/java/org/primefaces/model/DefaultLazyDataModel.java
+++ b/primefaces/src/main/java/org/primefaces/model/DefaultLazyDataModel.java
@@ -68,19 +68,7 @@ public class DefaultLazyDataModel<T> extends LazyDataModel<T> {
 
         sort(filteredValues);
 
-        if (pageSize == 0) {
-            return filteredValues;
-        }
-
-        int last = first + pageSize;
-        if (last > filteredValues.size()) {
-            last = filteredValues.size();
-        }
-        if (first > last) {
-            first = 0;
-        }
-        // new ArrayList as #subList is not serializable
-        return new ArrayList<>(filteredValues.subList(first, last));
+        return filteredValues.stream().skip(first).limit(pageSize).collect(Collectors.toList());
     }
 
     protected void sort(List<T> values) {
@@ -89,7 +77,10 @@ public class DefaultLazyDataModel<T> extends LazyDataModel<T> {
         }
 
         FacesContext context = FacesContext.getCurrentInstance();
-        values.sort(SortTableComparator.comparingField(context, (UITable) UIComponent.getCurrentComponent(context)));
+        UIComponent source = UIComponent.getCurrentComponent(context);
+        if (source instanceof UITable) {
+            values.sort(SortTableComparator.comparingField(context, (UITable<?>) source));
+        }
 
         if (sorter != null) {
             values.sort(sorter);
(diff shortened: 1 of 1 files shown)


primefaces/src/main/java/org/primefaces/model/DefaultLazyDataModel.java
```diff
@@ -68,19 +68,7 @@ public List<T> load(int first, int pageSize, Map<String, SortMeta> sortBy, Map<S
 
         sort(filteredValues);
 
-        if (pageSize == 0) {
-            return filteredValues;
-        }
```
Rapster: Removed as seems like a strange decision at first glance (if explanation given I can put it back, but I don't think that's how lazymodel has been working so far, we don't do that in our ITs)

primefaces/src/main/java/org/primefaces/model/DefaultLazyDataModel.java
```diff
@@ -68,19 +68,7 @@ public List<T> load(int first, int pageSize, Map<String, SortMeta> sortBy, Map<S
...
-
-        int last = first + pageSize;
-        if (last > filteredValues.size()) {
-            last = filteredValues.size();
-        }
-        if (first > last) {
-            first = 0;
-        }
```
Rapster, line 81: Not sure when that can happen (first > last ?)

Rapster (commented):
I left a few feedbacks regarding the existing code

primefaces/src/main/java/org/primefaces/model/DefaultLazyDataModel.java
Rapster: Ah must be the case, if rows = 0 component display everything, right?

primefaces/src/main/java/org/primefaces/model/DefaultLazyDataModel.java
tandraschko: If data is deleted in meantime?

primefaces/src/main/java/org/primefaces/model/DefaultLazyDataModel.java
tandraschko: Yep, i think so

primefaces/src/main/java/org/primefaces/model/DefaultLazyDataModel.java
Rapster: Yeah but that case is already handled by UIPageableData#calculateFirst in DataTable#loadLazyScrollData

tandraschko: Maybe already fixed in meantime
@jungm maybe you can retest

melloware: @Rapster that was a fix with jungm see: Fix #11521: Datatable first value fix Which was originally driven by the lazy and virtual scroll issue: https://github.com/primefaces/primefaces/pull/11377

jungm: This was my original issue that started all of this: https://github.com/primefaces/primefaces/issues/11401

Fixed it in https://github.com/primefaces/primefaces/pull/11402 in a slightly different way. I also added an IT for this exact issue when I fixed it, if it still passes everything should be good for me :slightly_smiling_face:

#11521 moved this to another place because it caused issues for other users

Rapster: ok, the IT is passing anyway and if it's not failing it's certainly because skip/limit can handle the job

melloware (approved):

PR #11673, opened 2024-03-24
Macro-optimization for DefaultResolver#get

Rapster:


tandraschko:
You can skip containsKey and use get+null check?
Even faster

tandraschko:
+1 now

In MyFaces i have optimized the whole thing even more via a LambdaPropertyDescriptor
But i think the benefit is quite small, it only makes a big difference with very very very much falls to the getter

Rapster:
Yes, hence the "Macro-optimization" description

PR #11695, opened 2024-03-29
Fix #11694: AJAX event handler leak

### LINKED ISSUE #11694: Memory: InputText with simple AJAX event leaks EventListener
Using this simple example of an AJAX `keyup` event on an InputText.  Then simply flip `rendered="false"` to remove the Input DOM element and add it to the `detachedWidgets` for cleanup.  This leaves the `keyUp` EventListener leaked handle holding on to memory and the Input element.

### Example Code

```xml
<h:form id="frmSimple">
        <p:selectBooleanCheckbox value="#{simpleView.rendered}" itemLabel="Show/Hide">
            <p:ajax update="inputs" />
        </p:selectBooleanCheckbox>
        <p></p>
        <div jsf:id="inputs">
            <p:inputText id="input" widgetVar="wgtInput" value="#{simpleView.text}" rendered="#{simpleView.rendered}">
                <p:ajax event="keyup" update="display2" process="@this" />
            </p:inputText>
            <p></p>
            <h:outputText id="display2" value="#{simpleView.text}" rendered="#{simpleView.rendered}" />
        </div>
</h:form>
```

### Heap Dump

image

melloware:
Fix #11694: AJAX event handler leak

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
@@ -1351,17 +1351,6 @@ if (!PrimeFaces.ajax) {
                         try {
                             widget.destroy();
 
-                            // remove any event listeners on jQuery objects
-                            for (var key in widget) {
-                                var jq = widget[key];
-                                if (jq instanceof jQuery) {
-                                    // remove events on all descendants
-                                    jq.children().off();
-                                    // remove events from element
-                                    jq.off();
-                                }
-                            }
-
                             // remove the widget from the global collection
                             delete PrimeFaces.widgets[widgetVar];
                             widget = null;
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
@@ -1351,7 +1351,7 @@ if (!PrimeFaces.ajax) {
                         try {
                             widget.destroy();
 
-                            // null out all values to de-reference any JQ objects
+                            // remove any event listeners on jQuery objects
                             for (var key in widget) {
                                 var jq = widget[key];
                                 if (jq instanceof jQuery) {
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
@@ -1350,8 +1350,6 @@ if (!PrimeFaces.ajax) {
                     if (widget && widget.isDetached() === true) {
                         try {
                             widget.destroy();
-
-                            // remove the widget from the global collection
                             delete PrimeFaces.widgets[widgetVar];
                             widget = null;
                         } catch (e) { PrimeFaces.warn("Error destroying widget: " + widgetVar) }
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
@@ -1348,12 +1348,24 @@ if (!PrimeFaces.ajax) {
 
                     var widget = PF(widgetVar);
                     if (widget && widget.isDetached() === true) {
-                        widget.destroy();
-
                         try {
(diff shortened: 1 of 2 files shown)


tandraschko:
Will do a detailed review in some days
But i think it should go into BaseWidget#destroy, not in the Ajax stuff

melloware:
Yep I can remove this one now. I fixed it with the other fix.

melloware:
sorry this one is still needed just tested at my client and without this code severely leaking handles.

melloware:
Moved to BaseWidget destroy.

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
```diff
@@ -1348,12 +1348,11 @@ if (!PrimeFaces.ajax) {
 
                     var widget = PF(widgetVar);
                     if (widget && widget.isDetached() === true) {
-                        widget.destroy();
-
                         try {
+                            widget.destroy();
                             delete PrimeFaces.widgets[widgetVar];
-                            delete widget;
```
tandraschko, line 1355: why removing the "delete widget"?

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.widget.js
```diff
@@ -393,6 +394,17 @@ if (!PrimeFaces.widget) {
                 }
             }
             this.destroyListeners = [];
+
+            // remove any event listeners on jQuery objects
```
tandraschko: please a bit more docu here
which contains that we loop all stored vars of the current widget, and unbding all event listeners to avoid mem leaks

also shouldnt we do addtionally:
this[key] = null 
just to be sure?

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
melloware: I changed it from `delete widget` which is no longer valid in modern browsers to `widget = null` to null out the variable.  delete should only be used on JSON object keys not on an object itself.

tandraschko: but delete PrimeFaces.widgets[widgetVar]; is still valid?

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.widget.js
melloware: I will update the docs to be more clear.  

I was originally doing `this[key] =null` here as well but some components failed in integration tests and my client had multiple issues with components where if there was a "delay" it was trying to use the `jq` and blowing up.  So I removed it and so far it has not affected any garbage collection.

For example here is a patch i had to add that was failing with `this[key]=null;` here so i had to defensively check the jq but it was not worth it to find it all over the codebase.  In my client that has the severe memory leak everything is garbage collecting fine without `this[key] =null` so I have chosen to leave it out and continue testing!

```js
/**
 * TieredMenu fix is a defensive NPE check on jq after our memory cleanup in destroyDetachedWidgets.
 */
if (PrimeFaces.widget.TieredMenu) {
  PrimeFaces.widget.TieredMenu.prototype.reset = function () {
    var $this = this;
    this.active = false;

    if (this.jq) {
      this.jq.find("li.ui-menuitem-active").each(function () {
        $this.deactivate($(this), true);
      });
    }
  };
}
```

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
melloware: yes you can delete JSON keys with `delete this[key]` but you can't delete a JSON object with `delete obj` it used to be an old IE hack but now all browsers want you to do `obj = null`

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.widget.js
tandraschko: Then pls add a doc here

melloware: comments and docs added

PR #11699, opened 2024-03-29
Fix #11698: InputMask destroy

### LINKED ISSUE #11698: InputMask: Destroy mask if widget is destroyed.
When an InputMask is removed from the DOM we must detach the InputMask wrapper.

melloware:
Fix #11698: InputMask destroy


tandraschko:
+1 here

PR #11701, opened 2024-03-29
Fix #11700: Widgets with trigger/target bind for removal

### LINKED ISSUE #11700: Memory: Widgets with "target" or "trigger" leak
This was a major cause of the leak at my client.   Any element that has a `trigger/target` element should remove its widget when its `target/trigger` is removed from the DOM.

- [ ] TieredMenu (any kind that uses trigger)
- [ ] Sticky (target)
- [ ] OverlayPanel (target)
- [ ] BlockUI (target)
- [ ] KeyFilter (target)
- [ ] Spotlight (target)
- [ ] Tooltip (target)
- [ ] Resizable (jqTarget)
- [ ] DragDrop (jqTarget)
- [ ] ContextMenu (jqTarget)
- [ ] ColumnToggler (trigger and its datatable)

## Reproducer 
This is easy to replicate using the showcase:

1. Navigate to Sticky example: https://www.primefaces.org/showcase/ui/misc/sticky.xhtml
2. Open F12 console and lets remove the Toolbar from the DOM with `$('#tb').remove();`
3. Now take a Memory Dump and you will see the Toolbar was removed from DOM but the `p:sticky` still has a JS reference to it in its `target` property which causes the garbage collector to never collect it.
4. Run `PrimeFaces.destoyDetachedWidgets()`

The 6 buttons, all their DIVS and everything inside the toolbar is now being held in memory as "Detached DOM" elements.

image

melloware:
Fix #11700: Widgets with trigger/target bind for removal

Wanted to put this out for review and get ideas on name and whether this is the best way to handle it.  It feels like it is the cleanest but I am open to names for `bindDomRemovalEvent` etc.


tandraschko:
Will review stuff in ~1 week, currently in holiday

tandraschko:
im not sure about this one
detachedWidgets is currently for the AJAX case, so the widget will be destroyed after the next AJAX request and not instant

i think we should destroy it immediately

melloware:
OK so just to be sure if I have a `p:sticky target="pnl"` and then I do an AJAX update where I no longer render the `pnl` the p:sticky holds the references to the panel in detached elements.

however if the Sticky registers the target for removal and I update the `pnl` to no longer be rendered it adds both the `pnl` and the `sticky` to the detached widgets and they both get cleaned up in the same AJAX right?

Or are you saying when its removed from the DOM the Sticky should be detroyed immediately and not added to `detachedWidgets` ?

tandraschko:
ahh
so you mean its still removed in the current ajax request (not in the next) as the detachedWidget cleanup is done after all elements where removed?

melloware:
Correct!

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.widget.js
```diff
@@ -587,6 +581,25 @@ if (!PrimeFaces.widget) {
...
+        },
+
+        /**
+         * Registers a DOM element such that its removal from the DOM results in the widget becoming "detached,"
+         * subsequently leading to its removal upon completion of the AJAX call.
+         *
+         * @param {JQuery | HTMLElement} watchElement The HTML element that if removed from the DOM will detach this widget.
+         * @since 14.0.0
+         */
+        bindDomRemovalEvent: function(watchElement) {
```
tandraschko: +1 in general for this PR
i think we should just align the method name a bit - its not clear the widget will be destroyed when the passed dom element is removed

melloware: Yep I didn't love the name either.  How about `registerDestroyListener`, or `registerRemovalWatcher`?  I am open to any names...

tandraschko: we already have registerDestroyListener

maybe something like
bindLifetimeOnElement
destroyOnElementRemoval

melloware: i like `destroyOnElementRemoval`

tandraschko:
TBH i wonder if this could destroy some usecases

e.g. a OverlayPanel binded to a button
i know that most of this cases require to update both components
but i can remember that we had some components, that worked if you only update 1 of those 2 married components
this case would be broken with this PR

melloware:
You are right I will only add it to the components I have seen cause a problem at my client. Namely Sticky and a couple of others.

melloware:
Understood this one is risky because of the scenarios you mentioned.

Is there some way i can make this an option??

  The scenario with that SPA app is each time they open a new Tab.... the tab has a Sticky Menu so if they scroll down the page the menu sticks.  They close the tab and the Sticky is now holding memory because its target is not in the DOM but it still has a reference to the `target`.  If they open and close 10 tabs now 10 sticky's are holding memory and it never gets cleaned up.

Even if they re-open a tab they had open that tab creates a new Sticky and new Menu but the original one is holding on to memory.  So if they do this for an 8 hour shift the browser eventually crashes.

tandraschko:
will review next days

tandraschko:
i cant speak for sticky for now (never used it) but e.g. take keyfilter

inputText and keyFilter are propably on same level in DOM
you update the parent container, both are getting destroyed

if have it in different container or and you dont update both, you get a memory leak - but its a user error, not a primefaces bug

maybe we just add some logging here

melloware:
OK i stripped it down to just Tooltip and Sticky (tooltip already had this behavior for the most part).

tandraschko:
can you also exclude unrelated changes here and create a seperate PR? resizable and dragdop doenst use destroyOnElementRemoval

melloware:
Done... Minimal changes now.

tandraschko:
TBH sticky looks very similar to the keyfilter case i mentioned above?
can you post the example code for it?

tandraschko:
(im only talking about the destroyOnElementRemoval, overwriting #destroy makes sense anyway and should fix the leak if the user correctly updates both)

melloware:
This is easy to replicate using the showcase:

1. Navigate to Sticky example: https://www.primefaces.org/showcase/ui/misc/sticky.xhtml
2. Open F12 console and lets remove the Toolbar from the DOM with `$('#tb').remove();`
3. Now take a Memory Dump and you will see the Toolbar was removed from DOM but the `p:sticky` still has a JS reference to it in its `target` property which causes the garbage collector to never collect it.
4. Run `PrimeFaces.destoyDetachedWidgets()`
5. Now take a Memory Dump

The 6 buttons, all their DIVS and everything inside the toolbar is now being held in memory as "Detached DOM" elements.

image

tandraschko:
yeah but thats not a "valid case"
we not talking about that users are removing elements with JS and therefore we are getting leaks?
for me a valid case is like this:
```
 <h:form>
      <div jsf:id="someContainer" rendered="#{el....}>
          <p:toolbar id="tb"/>
          <p:sticky />
      </div>
 </h:form>
```

if you update form, both are rendered/not-rendered and in the same scope

if the scope is different, then you will get a mem leak; exactly the same like with keyfilter.

tandraschko:
yeah, please ask them, im not sure if my thoughts cover all cases but:
```
 <h:form>
      <div jsf:id="someContainer" rendered="#{el....}>
          <p:toolbar id="tb"/>
      </div>
      <p:sticky />
 </h:form>
```
if you take this example and update=someContainer, sticky also wont be automatically reattach to the toolbar
i think the whole thing is very very tricky

should a component auto reattach to the target component? how are the relations between components, parents, child and so on
this really really needs some detailed specs

tandraschko:
nevertheless, maybe its better to provide a own PR with destroy implementations (also keyfilter, dragdrog and so on first)
and really a issue for the releation stuff

melloware:
OK i will put this one back to draft for now as you are right components should be updated together and finding out why the Sticky is not being destroyed when the Tab is detroyed is really the core issue.

PR #11704, opened 2024-04-01
Fix #11703: FileUpload clean up on destroy

### LINKED ISSUE #11703: FileUpload: add destroy method
The file upload needs to clean up after itself.

melloware:
Fix #11703: FileUpload clean up on destroy

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/fileupload/2-fileupload.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/fileupload/2-fileupload.js
@@ -386,6 +386,20 @@ PrimeFaces.widget.FileUpload = PrimeFaces.widget.BaseWidget.extend({
 
         this.jq.fileupload(this.ucfg);
     },
+    
+    /**
+     * @override
+     * @inheritdoc
+     */
+    destroy: function() {
+        try {
+            this.jq.fileupload("destroy");
+        } catch (err) {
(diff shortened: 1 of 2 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/fileupload/2-fileupload.js
```diff
@@ -386,6 +386,21 @@ PrimeFaces.widget.FileUpload = PrimeFaces.widget.BaseWidget.extend({
...
         this.jq.fileupload(this.ucfg);
     },
+    
+    /**
+     * @override
+     * @inheritdoc
+     */
+    destroy: function() {
+        try {
+            $(document).off("mouseover.fileupload mouseout.fileupload mousedown.fileupload mouseup.fileupload focus.fileupload blur.fileupload click.fileupload ", this.rowCancelActionSelector);
```
tandraschko: may we also need the id here?

melloware: agreed.

melloware: I added the ID and also switch the `off()` to just use the name space which jQuery allows you to do and it removes all events with that namespace.  

```js
// Remove all events for a specific namespace
$(document).off('.yourNamespace');
```

PR #11706, opened 2024-04-01
Fix #11705: IdleMonitor destroy widget properly

### LINKED ISSUE #11705: IdleMonitor: Never registered for destruction
The IdleMonitor has multiple issues...

1. Empty `jq` element so its not registered for widget removal if the idle monitor is updated.
2. `destroy` should stop and unregister the idle monitor.

melloware:
Fix #11705: IdleMonitor destroy widget properly

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/idlemonitor/1-idlemonitor.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/idlemonitor/1-idlemonitor.js
@@ -101,9 +101,10 @@ PrimeFaces.widget.IdleMonitor = PrimeFaces.widget.BaseWidget.extend({
         if (this.cfg.multiWindowSupport) {
             clearInterval(this.timer);
         }
-        $(document).off('idle.idleTimer active.idleTimer');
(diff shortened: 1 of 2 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/idlemonitor/1-idlemonitor.js
```diff
@@ -98,11 +98,13 @@ PrimeFaces.widget.IdleMonitor = PrimeFaces.widget.BaseWidget.extend({
      * @inheritdoc
      */
     destroy: function() {
-        this._super();
-
         if (this.cfg.multiWindowSupport) {
             clearInterval(this.timer);
         }
+        $(document).off('idle.idleTimer active.idleTimer');
```
tandraschko: what happens here with multiple timers on the view?

melloware: Great catch I need to be using the ID.  PR updated.

PR #11708, opened 2024-04-01
Fix #11702: DOM leaks on AJAX update

### LINKED ISSUE #11694: Memory: InputText with simple AJAX event leaks EventListener

### LINKED ISSUE #11702: Memory: update="@form" causes DOM Leak
This was actually been around since 2013 I found this forum post: https://forum.primefaces.org/viewtopic.php?p=112616&sid=bc050d696d9491eda81c886c272f705b#p112616

**Simple example:**
```xml
    <h:form id="frmSimple">
        <p:selectBooleanCheckbox value="#{simpleView.rendered}" itemLabel="Show/Hide">
            <p:ajax update="@form" />
        </p:selectBooleanCheckbox>
        <p></p>
        <div jsf:id="inputs">
            <p:inputText id="input" widgetVar="wgtInput" value="#{simpleView.text}" rendered="#{simpleView.rendered}">
                <p:ajax event="keyup" update="display2" process="@this" />
            </p:inputText>
            <p></p>
            <h:outputText id="display2" value="#{simpleView.text}" rendered="#{simpleView.rendered}" />
        </div>
    </h:form>
```

Every time you AJAX update `@form` or `:formSimple` or any variation you will see leaked DOM elements because the old form has jQuery handlers on it.

image

melloware:
Fix #11702: DOM leaks on AJAX update
Fix #11694

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
@@ -1370,7 +1370,7 @@ if (!PrimeFaces.ajax) {
 
                         try {
                             delete PrimeFaces.widgets[widgetVar];
-                            widget = null;
+                            delete widget;
                         } catch (e) { }
                     }
                 }
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
@@ -326,7 +326,9 @@ if (!PrimeFaces.ajax) {
                         PrimeFaces.warn("DOM element with id '" + id + "' cant be found; skip update...");
                     }
                     else {
-                        target.replaceWith(content);
+                        var removedContent = target.replaceWith(content);
+                        // detach all handlers and data to clean up DOM
+                        PrimeFaces.utils.cleanseDomElement(removedContent);
                     }
                 }
             }
@@ -1352,7 +1354,7 @@ if (!PrimeFaces.ajax) {
 
                         try {
                             delete PrimeFaces.widgets[widgetVar];
-                            delete widget;
+                            widget = null;
                         } catch (e) { }
                     }
                 }
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
@@ -1098,6 +1098,46 @@ if (!PrimeFaces.utils) {
             // Decrease zIndex by 1 and return
             return zIndex - 1;
         },
+
+        /**
+         * Deletes all events, 'on' attributes, data, and the element itself in a recursive manner, 
+         * ensuring that the garbage collector does not retain any references to this element or its children.
+         *
+         * @param {JQuery | undefined} jq jQuery object to cleanse
+         * @param {boolean} [clearData] flag to clear data off elements (default to true)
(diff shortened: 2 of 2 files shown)


tandraschko:
I wonder why this is needed, same like the new logic in the Ajax destroy

I wonder If jquery changed something here?
Some years ago i worked on mem leaks and introduced the detach/destroy mechanism, everything was good with most widgets.

melloware:
I think it was a jquery change back in 1.12 where they changed how `replaceWith` works.

But basically the old thing we just replaced is holding on to all its references and never letting the garbage collector work.  By removing all events, data, etc then it all gets garbage collected.

tandraschko:
Can you post the ticket here? Dont they have a replacement?

melloware:
Reproducer:

Do the following steps...

1. Put a JS breakpoint on the `target.replaceWith` step.

image

2. Run the Heap Snapshot to check detached objects before you start (some of them are always from jQuery).

image


3. Now press the "Update @form" checkbox to update the page and remove the rendered Input but pause on the breakpoint and take a Heap Snapshot.

image

4. Now step over the `replaceWith` step and take another heap snapshot.

image

5. Now here is the same step with my code being used...  You can see it reclaims the DIV, Input, and the V8EventHandler that the AJAX `keyup` event was holding on to.

```js
 var removed = target.replaceWith(content);
      cleanseDom(removed);
```

image

melloware:
@tandraschko and just to be clear this is what is holding references `onkeyup="PrimeFaces.ab({s:&quot;frmSimple:input&quot;,e:&quot;keyup&quot;,f:&quot;frmSimple&quot;,p:&quot;frmSimple:input&quot;,u:&quot;frmSimple:display&quot;});"`

That is why I need to do this and remove all "on" handlers.

```js
// Remove inline event attributes
            var attributes = jq[0].attributes;
            for (var i = 0; i < attributes.length; i++) {
                var attributeName = attributes[i].name;
                if (attributeName.startsWith("on")) {
                    jq.removeAttr(attributeName);
                }
            }
```

I also remove jquery `removeData()` because most sites while I was doing my research said a lot of times references are left when someone does `data("value", jsCode)` and the jsCode contains refs. We use `data` all over the PF codebase so this is another defensive safe thing to remove all references.

tandraschko:
why should this script reference trigger any leaks? it doesnt reference any DOM element or anything else?

melloware:
I don't know what Chrome or Edge do under the covers but you can see right here this is what is holding a reference.

image

or if I use the checkbox to update="@form" the checkbox's onChange event holds the handles.

image

melloware:
its the `onchange` and `onkeyup` on the element attribute.  At least in this case I am positive.

PR #11713, opened 2024-04-02
Fix #11712: Detach document/window and events

### LINKED ISSUE #11712: Memory: Never store `$(window)` or `$(document)`
Using jQuery if you do something like this...

```js
var win = $(window);

PrimeFaces.utils.registerScrollHandler(this, 'scroll.' + this.id + '_align', function() {
            var scrollTop = win.scrollTop(),
}
```

Then use it in a callback it leaks and the garbage collector will never collect anything related to this callback.

melloware:
Fix #11712: Detach document/window and events

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/sticky/sticky.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/sticky/sticky.js
@@ -62,11 +62,10 @@ PrimeFaces.widget.Sticky = PrimeFaces.widget.BaseWidget.extend({
      * @private
      */
     bindEvents: function() {
-        var $this = this,
-            win = $(window);
+        var $this = this;
 
         PrimeFaces.utils.registerScrollHandler(this, 'scroll.' + this.id + '_align', function() {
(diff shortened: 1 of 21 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/sticky/sticky.js
```diff
@@ -98,7 +96,7 @@ PrimeFaces.widget.Sticky = PrimeFaces.widget.BaseWidget.extend({
             this.ghost = $('<div class="ui-sticky-ghost"></div>').height(this.target.outerHeight()).insertBefore(this.target);
             this.target.width(this.ghost.outerWidth() - (this.target.outerWidth() - this.target.width()));
             this.fixed = true;
-            win.scrollTop(winScrollTop);
+            $(window).scrollTop(winScrollTop);
```
tandraschko, line 99: for such cases, its actually not required but you did it for uniformity?

melloware: Correct I just did it for uniformity.

PR #11724, opened 2024-04-04
Fix #11723: TabView memory cleanup on refresh/destroy

### LINKED ISSUE #11723: TabView: Implement proper cleanup
### Description
For Tabview using `scrollable="true"` the tabview is not destroying itself properly allowing DOM elements to become detached.

This was a tricky one for me to figure out!

### Reproducer

```xml
<p:tabView id="tabView" scrollable="true">
```

Make sure to have Scrollable Tabview then update it using `update="tabview" or "@form"` and then you will see this in the Heap Dump

image

### Solution

We need to unbind the resize listener on `refresh` because this TabView has a `refreshlistener` that dynamically creates and replaces tabs.

melloware:
Fix #11723: TabView memory cleanup on refresh/destroy


primefaces/src/main/resources/META-INF/resources/primefaces/tabview/tabview.js
```diff
@@ -139,11 +140,22 @@ PrimeFaces.widget.TabView = PrimeFaces.widget.DeferredWidget.extend({
 
             var $this = this;
 
-            PrimeFaces.utils.registerResizeHandler(this, 'resize.' + this.id + '_align', null, function() {
+            this.resizeHandler = PrimeFaces.utils.registerResizeHandler(this, 'resize.' + this.id + '_align', this.jq, function() {
```
tandraschko: i wonder if it makes sense to deregister it automatically on refresh/destroy inside PrimeFaces.utils.registerResizeHandler

melloware: I was thinking the same thing.  I think for registerScrollHandler and resizeHandler we should.

melloware:
OK register for scroll and resize now both add refresh listeners!

tandraschko:
conflicts

melloware:
Conflicts resolved

PR #11746, opened 2024-04-11
ExceptionHandler: null check session, use ClientWindow whenever possible

### LINKED ISSUE #11737: Exception Handler: do not try to create sessions
As discussed on discord (https://discord.com/channels/787967399105134612/787967662293909524/1225465731215392838)

Currently PrimeExceptionHandler and PrimeExceptionHandlerELResolver use ExternalContext#getSessionMap to retrieve the session map. This triggers session creation if it doesn't exist, and it's bad for two reasons:

- if the response is already commited, trying to create a session causes an exception
- if the user want stateless mode (`f:view transient="true"`)

Those classes should check session existence before using the session map.
Some alternatives I thought about:

- using ClientWindow as suggested by @tandraschko -> still needs session if i'm not wrong
- using Flash -> will fail if the response is already commited
- sending exception info as encoded url params -> is it safe?
- saving exception infos in the application map using a random id and send the id as a url param -> how and when the map will be cleaned?

Any thoughts?

NicolaIsotta:
fixes #11737 
may be backported to 13.X

### DIFF
--- a/primefaces/src/main/java/org/primefaces/application/exceptionhandler/PrimeExceptionHandler.java
+++ b/primefaces/src/main/java/org/primefaces/application/exceptionhandler/PrimeExceptionHandler.java
@@ -33,6 +33,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+
 import javax.el.ELException;
 import javax.faces.FacesException;
 import javax.faces.application.ProjectStage;
@@ -48,7 +49,9 @@ import javax.faces.context.FacesContext;
 import javax.faces.context.PartialResponseWriter;
 import javax.faces.event.ExceptionQueuedEvent;
 import javax.faces.event.PhaseId;
+import javax.faces.lifecycle.ClientWindow;
 import javax.faces.view.ViewDeclarationLanguage;
+
 import org.primefaces.component.ajaxexceptionhandler.AjaxExceptionHandler;
 import org.primefaces.component.ajaxexceptionhandler.AjaxExceptionHandlerVisitCallback;
 import org.primefaces.config.PrimeConfiguration;
@@ -60,6 +63,7 @@ import org.primefaces.util.ComponentUtils;
 import org.primefaces.util.LangUtils;
 import org.primefaces.util.EscapeUtils;
 import org.primefaces.util.Lazy;
+import org.primefaces.util.LimitedSizeHashMap;
 
 public class PrimeExceptionHandler extends ExceptionHandlerWrapper {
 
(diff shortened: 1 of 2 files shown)


melloware (changes requested):
Checkstyle issues

NicolaIsotta:
> Checkstyle issues

Gosh I built with quick and I didn't notice. Fixing it now.

primefaces/src/main/java/org/primefaces/application/exceptionhandler/PrimeExceptionHandler.java
```diff
@@ -343,7 +346,17 @@ protected Throwable buildView(FacesContext context, Throwable throwable, Throwab
 
     protected void handleRedirect(FacesContext context, Throwable rootCause, ExceptionInfo info, boolean isResponseReset) throws IOException {
         ExternalContext externalContext = context.getExternalContext();
-        externalContext.getSessionMap().put(ExceptionInfo.ATTRIBUTE_NAME, info);
+        if (externalContext.getSession(false) != null) {
+            ClientWindow clientWindow = externalContext.getClientWindow();
+            if (clientWindow != null && LangUtils.isNotBlank(clientWindow.getId())) {
+                Map<String, ExceptionInfo> windowsMap = (Map<String, ExceptionInfo>)
+                        externalContext.getSessionMap().computeIfAbsent(ExceptionInfo.ATTRIBUTE_NAME + "_map", k -> new HashMap<>());
```
tandraschko: better ConcurrentHashMap, could be accessed in theory by multiple requests

NicolaIsotta: Done

melloware (approved):

primefaces/src/main/java/org/primefaces/application/exceptionhandler/PrimeExceptionHandler.java
```diff
@@ -343,7 +346,17 @@ protected Throwable buildView(FacesContext context, Throwable throwable, Throwab
 
     protected void handleRedirect(FacesContext context, Throwable rootCause, ExceptionInfo info, boolean isResponseReset) throws IOException {
         ExternalContext externalContext = context.getExternalContext();
-        externalContext.getSessionMap().put(ExceptionInfo.ATTRIBUTE_NAME, info);
+        if (externalContext.getSession(false) != null) {
+            ClientWindow clientWindow = externalContext.getClientWindow();
+            if (clientWindow != null && LangUtils.isNotBlank(clientWindow.getId())) {
+                Map<String, ExceptionInfo> windowsMap = (Map<String, ExceptionInfo>)
+                        externalContext.getSessionMap().computeIfAbsent(ExceptionInfo.ATTRIBUTE_NAME + "_map", k -> new ConcurrentHashMap<>());
```
tandraschko: we should better use a LimitedSizeHashMap (we already have this in PF) maybe limit up to 5 or 10

in theory we should introduce a LimitedSizeConcurrentHashMap but its unlikely that you parallel access 2 views anyway and this would need to throw a exception in the same millisecond

NicolaIsotta: done

melloware (approved):

PR #11753, opened 2024-04-13
Fix #11748 - DataExporter: deprecate ExporterUtils

### LINKED ISSUE #11748: DataExporter: Make the ExporterUtils.getColumnValue method overridable
### Description

I do not like the fact, that calls to static methods like `ExporterUtils.getComponentValue()` and friends are still hardwired in the exporter framework and not overwritable by the user.

We have columns containing components that are not in the covered list of that implementation (e.g. pe:timeAgo). So instead of overriding a method there, we have to clone `TableExporter.addCells()`, `TableExporter.addColumnFacets()`, `TableExporter.addColumnGroupFacets()` and `TableExporter.addTableFacets()` in each of our three custom exporters.

In addition the `cellJoinComponents` field in the `TableExporter` class should be accessible via a protected getter method to derived classes, as this is used heavily in the beforementioned addXXX methods.

The user should be able to overwrite the implementation of ExporterUtils methods.

### Additional context

_No response_

Rapster:


Rapster:
cc @stolp

melloware:
@Rapster looks good but i labeled this breaking change for 14.0.0 because users can no longer do `exporter="#{dataExporterView.textExporter}" which I am fine with i just want to make sure the release notes generated mark this as a breaking change

primefaces-showcase/src/main/java/org/primefaces/showcase/PFShowcaseSystemEventListener.java
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.showcase;
```
melloware: This is minor but all the other showcase supporting stuff is in `.util` and maybe we should just call it `ShowcaseSystemEventListener` since PF is implied here?

melloware (changes requested):

Rapster:
Here I'm not really happy about moving everything into TableExporter (since methods are not bound to the exporter instance, that's why I moved these methods into ExporterUtils, and I fear to end-up with anti-pattern like we already have in CoreRenderer where we have plenty of methods just for practical reasons). A solution, would be to do like `DataExporters#register` where a user can register his own utils, WDYT?

melloware:
So I go back and forth.  I like your moving this back into the TablExporter because they are really TableExporter utilities.  If we were to say make an exporter for `DataGrid` for some reason (not that we would) its NOT a table nor does it have `UIColumn` so all those utils in ExporterUtils wouldn't make sense.   So having them contained and overridable in TableExporter make sense to.

Plus we have to balance the fact that we don't want to make it too cumbersome for someone to add a custom exporter by now having to register their own utils too?

Rapster:
I agree with your statement, but I still think it's a very poor design. I'll leave it like this then for simplicity

melloware (approved):

stolp:
Moving ExporterUtils to its own registered component adds the possibility to share them between several exports, so I also think it would be the more elegant solution.

In my code I can certainly live with the implemented state, as we already implemented our own Utility class for this purpose.

stolp (approved):

PR #11756, opened 2024-04-14
Move IOBiConsumer into org.primefaces.util

Rapster:
Here I think we should have something like FailableConsumer from commons-lang


melloware:
Was this the only class in the functional package?  If so I agree with this move.

Rapster:
Yes it was

melloware (approved):

PR #11761, opened 2024-04-16
Fix #11759: InterceptingResolver add convertToType

### LINKED ISSUE #11759: InterceptingResolver does not delegate convertToType
### Describe the bug

Boolean EL expressions including `not`, `and`, `or` resolve to null if already resolved. This is caused by missing delegation of `convertToType` method in [ `org.primefaces.el.InterceptingResolver`](https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/java/org/primefaces/el/InterceptingResolver.java).

Used versions:
Primefaces: 13.0.8
Tomcat-Embed-EL: 10.1.19 (via Springboot 3.2.4)
Mojarra: 4.0.6

### Reproducer

_No response_

### Expected behavior

Using the delegate for proper resolving.

### PrimeFaces edition

None

### PrimeFaces version

13.0.8

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0.0

### Java version

17

### Browser(s)

_No response_

melloware:
Fix #11759: InterceptingResolver add convertToType

### DIFF
--- a/primefaces/src/main/java/org/primefaces/el/InterceptingResolver.java
+++ b/primefaces/src/main/java/org/primefaces/el/InterceptingResolver.java
@@ -82,8 +82,6 @@ public class InterceptingResolver extends ELResolver {
         return delegate.getCommonPropertyType(context, base);
     }
 
-    // @Override
-    // < EL 2.2 compatibility
     @Override
     public Object invoke(ELContext context,
             Object base,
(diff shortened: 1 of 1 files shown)


tandraschko:
do we support something older then 3.0? if yes -> please comment `@Override`

melloware:
nevermind found it: https://stackoverflow.com/a/7237216/502366

primefaces/src/main/java/org/primefaces/el/InterceptingResolver.java
```diff
@@ -92,4 +92,11 @@ public Object invoke(ELContext context,
             Object[] params) {
         return delegate.invoke(context, base, method, paramTypes, params);
     }
+
+    // @Override
+    // < EL 3.0 compatibility
```
tandraschko: why those comments? just remove it, its not true anyway

melloware: I was just following the pattern from the method above it about 2.2 compatibility?

PR #11768, opened 2024-04-17
Fix #11767 - JPALazyDataModel: add resultEnricher callback

### LINKED ISSUE #1167: autocomplete with cache
Selected object became null when cache=true
Try this sequence.
1. Search for a text which is not present, the result rightly shows "No Record" 
2. Then search for a valid text and select that record.
3. If you debug  onAutoSelect(SelectEvent event)  event.getObject() is null.

Note : It works fine when cache=false

```
public void onAutoSelect(SelectEvent event) {
   Object o =  event.getObject();
  }

 <p:autoComplete value="#{bookView.author}" completeMethod="#{bookView.authorAuto}"  var="e" itemValue="#{e}" itemLabel="#{e.name}" emptyMessage="No Record" dropdown="true" cache="true" queryDelay="500" maxResults="500" converterMessage="Invalid Selection" autocomplete="false" forceSelection="true" > 
          <o:converter list="#{bookView.authorFilterList}" converterId="omnifaces.ListConverter"  />
          <p:ajax event="itemSelect" listener="#{bookView.onAutoSelect}" />
        </p:autoComplete>

```

PF 5.3.7, Tomcat 8, Java 8

### LINKED ISSUE #11767: JPALazyDataModel: add resultEnricher callback
### Description

Since the Constructor has been removed in PF14 R3 (https://github.com/primefaces/primefaces/pull/11754), the Post Process is not working anymore.

Can you add the possibility to add e.g. a function which allows to "manipulate" the results from the Lazy Loading result afterwards (e.g. fill some addiotional columns etc.)

Please see also discussion in Discord:
https://discord.com/channels/787967399105134612/787967662293909524/1230050675757547562

To have a function for Post Progress...

### Additional context

_No response_

Rapster:
cc @rider87

### DIFF
--- a/primefaces/src/main/java/org/primefaces/model/JPALazyDataModel.java
+++ b/primefaces/src/main/java/org/primefaces/model/JPALazyDataModel.java
@@ -62,7 +62,7 @@ public class JPALazyDataModel<T> extends LazyDataModel<T> implements Serializabl
     protected SortEnricher<T> sortEnricher;
     protected SerializableSupplier<EntityManager> entityManager;
     protected SerializableFunction<T, Object> rowKeyProvider;
-    protected SerializableConsumer<List<T>> postLoadEnricher;
+    protected SerializableConsumer<List<T>> resultEnricher;
 
     /**
      * For serialization only
@@ -107,8 +107,8 @@ public class JPALazyDataModel<T> extends LazyDataModel<T> implements Serializabl
         }
 
         List<T> result = query.getResultList();
-        if (postLoadEnricher != null) {
-            postLoadEnricher.accept(result);
+        if (resultEnricher != null) {
+            resultEnricher.accept(result);
         }
         return result;
     }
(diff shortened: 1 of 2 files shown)


primefaces/src/main/java/org/primefaces/model/JPALazyDataModel.java
```diff
@@ -401,6 +406,11 @@ public Builder<T> sortEnricher(SortEnricher<T> sortEnricher) {
             return this;
         }
 
+        public Builder<T> postLoadEnricher(SerializableConsumer<List<T>> postLoadEnricher) {
```
tandraschko: also resultEnricher

Rapster: yup just saw it...

tandraschko:
looks good, thanks!

PR #11784, opened 2024-04-19
Fix #11780: Fix resource handling to allow raw JS code

### LINKED ISSUE #11780: Cannot upgrade to 13.0.8. Some strange behaviour with omnifaces
### Describe the bug

This example works in PF 13.0.7 but does not work in 13.0.8
```
      xmlns:o="http://omnifaces.org/ui"
      xmlns:f="http://xmlns.jcp.org/jsf/core">
<f:view>
    <h:head></h:head>
    <h:body>
        <c:if test="true">

        </c:if>
        <o:onloadScript>
            console.log('hi');
        </o:onloadScript>
        <h:form>
            <h1>
                This example works in PF 13.0.7 but does not work in PF 13.0.8
            </h1>
            <p:autoComplete id="acSimple" value="#{autocompleteViewBean.value}"
                            completeMethod="#{autocompleteViewBean.complete}" scrollHeight="250"/>
        </h:form>
    </h:body>
</f:view>
</html>
```

In order for this to fail it has to have both a <c:if> and a <o:onloadScript>, so I'm not sure where is the bug, but it was working in previous versions. 

I provided a minimal reproducer in https://github.com/cocorossello/primefaces-test

To run it:
`mvn clean jetty:run -Pmyfaces23next`

Then open http://localhost:8080/primefaces-test/ and check that the autocomplete doesn't work (the call is performed and I can see the results in the xhr response but the update is lost somehow).

Change PF version to 13.0.7 and it works again

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

13.0.8

### Theme

_No response_

### JSF implementation

MyFaces

### JSF version

2.3-next-M7

### Java version

17

### Browser(s)

_No response_

melloware:
Fix #11780: Fix resource handling to allow raw JS code

@tandraschko what happens is for some reason MyFaces sends this and Mojarra doesn't.

```xml
<update id="javax.faces.Resource"><![CDATA[
            console.log('hi');
        ]]></update>
```

When you try and convert that to Jquery `var $content = $(content);` it throws an error because that is not valid Jquery content.  So I had to add a `catch` and say add the raw `content` to the head if an error is thrown.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
@@ -315,23 +315,7 @@ if (!PrimeFaces.ajax) {
                     PrimeFaces.ajax.Utils.updateBody(content);
                 }
                 else if (id === PrimeFaces.ajax.RESOURCE) {
-                    // #11714 Iterate through each script and stylesheet tag in the content
-                    // checking if resource is already attached to the head and adding it if not
-                    $(content)
-                        .filter("link[href], script[src]")
-                        .each(function() {
-                            var $resource = $(this);
-                            var $head = $("head");
-                            var src = $resource.attr("href") || $resource.attr("src");
-                            var type = this.tagName.toLowerCase();
-                            var $resources = $head.find(type + '[src="' + src + '"], ' + type + '[href="' + src + '"]');
-
-                            // Check if script or stylesheet already exists and add it to head if it does not
-                            if ($resources.length === 0) {
-                                PrimeFaces.debug("Appending " + type + " to head: " + src);
-                                $head.append($resource);
-                            }
-                        });
+                    $('head').append(content);
                 }
                 else if (id === $('head')[0].id) {
                     PrimeFaces.ajax.Utils.updateHead(content);


tandraschko:
dont like that try/catch TBH
i would check for a src attr

melloware:
I am struggling with how though as `content` is just a string value?  And when you do `$(content)` JQuery throws an error.

melloware:
I am checking for sources and links here...

```js
var $content = $(content);
                        var filteredContent = $content.length > 0 ? $content.filter("link[href], script[src]") : $();

                        if (filteredContent.length === 0) {
                            PrimeFaces.debug("Adding content to the head because it lacks any JavaScript or CSS links...");
                            $head.append(content);
                        }
                        ```

But the catch is necessary for something that is not even a HTML element like `console.log()`.

**Comment by tandraschko** [comment]
maybe we should pass the node then and access via XML api

**Comment by melloware** [comment]
Let me poke around with that idea of using the node.

**Comment by tandraschko** [comment]
Do you have a reproducer for the original problem? Maybe we should just revert and fix the root problem

**Comment by melloware** [comment]
The original issue I can definitely watch happen in real time but having trouble isolating why.  But this user from 6 years ago has the exact same issue: https://forum.primefaces.org/viewtopic.php?t=71210

And he couldn't figure out why but for some reason JSF thinks it needs to send the JS and CSS for a component again after its been loaded, unloaded, and reloaded all while on the same page.

**Comment by melloware** [comment]
i will revert for now and re-open that ticket.

**Comment by melloware** [comment]
Also I would like to understand why MyFaces duplicates the script in head and eval.

```xml
<partial-response id="j_id__v_0">
	<changes>
		<update id="javax.faces.Resource"><![CDATA[
            console.log('hi');
        ]]></update>
		<update id="j_id_9:acSimple"><![CDATA[<ul class="ui-autocomplete-items ui-autocomplete-list ui-widget-content ui-widget ui-corner-all ui-helper-reset" role="listbox"><li id="j_id_9:acSimple_item_0" class="ui-autocomplete-item ui-autocomplete-list-item ui-corner-all" data-item-label="Result 1" data-item-value="Result 1" data-item-class="">Result 1</li><li id="j_id_9:acSimple_item_1" class="ui-autocomplete-item ui-autocomplete-list-item ui-corner-all" data-item-label="Result 2" data-item-value="Result 2" data-item-class="">Result 2</li><li id="j_id_9:acSimple_item_2" class="ui-autocomplete-item ui-autocomplete-list-item ui-corner-all" data-item-label="Result 3" data-item-value="Result 3" data-item-class="">Result 3</li></ul>]]></update>
		<update id="j_id__v_0:javax.faces.ViewState:1"><![CDATA[YTFlMDk1NzRjYzBmMTlmYjAwMDAwMDAx]]></update>
		<eval><![CDATA[console.log('hi');]]></eval>
	</changes>
</partial-response>
```

tandraschko:
Both are propably bugs in the impls, we should create issues there

melloware:
ok let me create issues there.

melloware:
Actually I started with OmniFaces his the `<o:onloadScript>` is what triggered this to begin with: https://github.com/omnifaces/omnifaces/issues/804

PR #11787, opened 2024-04-19
PDF export minor refactor

melloware:
@Rapster this minor refactor make my life easier when getting Quarkus Native mode working as it allows me to override just this tinier method similar to how I have to override add blank lines.

See this for an example of what I need to do: https://github.com/quarkiverse/quarkus-primefaces/blob/main/quarkus-primefaces/runtime/src/main/java/io/quarkus/primefaces/runtime/graal/SubstituteDataTablePdfExporter.java


primefaces/src/main/java/org/primefaces/component/datatable/export/DataTablePDFExporter.java
```diff
@@ -161,6 +161,12 @@ protected void addEmptyLine(Paragraph paragraph, int number) {
         }
     }
 
+    protected void addCell(PdfPTable table, PdfPCell cell) {
+        if (table != null && cell != null) {
+            table.addCell(cell);
+        }
```
Rapster: Those null check are really necessary? if they're null there is a truly a problem

melloware: Removed.  I was just being defensive they will never be null.

PR #11805, opened 2024-04-24
Fix #11421: Ignore ColumnRenderer helper in components using columns

### LINKED ISSUE #11421: SelectCheckboxMenu: Custom rendering broken when used as `DataTable` filter in a `ColumnGroup`
### Describe the bug

When a `SelectCheckboxMenu` with custom rendering is used as a `DataTable` filter, and the header of the table is specified within a `ColumnGroup`, then the rendering of the `SelectCheckboxMenu` is broken. The column with the selection labels is empty, and an additional column, which contains the value of the header facet, is rendered for each selection item.

SelectCheckboxMenu-broken

### Reproducer

[primefaces-test-master.zip](https://github.com/primefaces/primefaces/files/14233689/primefaces-test-master.zip)

### Expected behavior

There should be just one column with the selection labels.

### PrimeFaces edition

Community

### PrimeFaces version

13.0.5

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.2

### Java version

1.8.0_202-b08

### Browser(s)

Chromium 121.0.6167.85, Firefox 115.7.0esr

melloware:
Fix #11421: Ignore ColumnRenderer helper in components using columns

This PR does it the "better" way with a utility to wrap and restored the helper.  Note we need to do this on all components that have custom "Column Rendering".


melloware:
also i didn't know if "Stateless" was the right term since this is removing a value from the State map...

tandraschko:
i like this more actually but it wraps IOException into FacesException, which can have drawbacks on exception handling propably
but not sure

Rapster:
Not sure about the name, but the try/catch and the return null doesn't look nice... Use a `Runnable` instead, something like `FailableRunnable` for nicer exception handling

melloware:
@Rapster can you give an example of how that would work?  I have only use Runnable's for threading not just for a lambda wrapper?

Rapster:
You can use Runnable just like any other functional interfaces

melloware:
OK @Rapster i think I got it if you want to re-review

primefaces/src/main/java/org/primefaces/component/autocomplete/AutoCompleteRenderer.java
```diff
@@ -506,7 +508,10 @@ protected void encodeSuggestions(FacesContext context, AutoComplete ac, Object i
         Converter converter = ComponentUtils.getConverter(context, ac);
 
         if (customContent) {
-            encodeSuggestionsAsTable(context, ac, items, converter);
+            FailableRunnable<IOException> runnable = () -> {
```
Rapster: instead of declaring it, I don't mind if it's inline

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
```diff
@@ -698,7 +710,9 @@ public static <T> T executeInRequestScope(FacesContext context, String var, Obje
 
         Object oldValue = requestMap.remove(var);
         try {
-            requestMap.put(var, value);
+            if (value != null) {
+                requestMap.put(var, value);
+            }
 
             return callback.get();
```
Rapster: Simpler this way no?

```java
Object oldValue = requestMap.put(var, value);
try {
    return callback.get();
}
finally {
    if (oldValue != null) {
        requestMap.put(var, oldValue);
    }
}
```

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
```diff
@@ -710,6 +724,35 @@ public static <T> T executeInRequestScope(FacesContext context, String var, Obje
...
+
+        Object oldValue = stateMap.remove(var);
+        try {
+            runnable.run();
+        }
+        finally {
+            if (oldValue != null) {
+                stateMap.put(var, oldValue);
+            }
+        }
```
Rapster, line 751: same here

primefaces/src/main/java/org/primefaces/util/FailableRunnable.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+package org.primefaces.util;
+
+/**
+ * A functional interface like {@link Runnable} that declares a kind of {@link Throwable}.
+ *
+ * @param <E> The kind of {@link Throwable}.
+ */
+@FunctionalInterface
+public interface FailableRunnable<E extends Throwable> {
```
Rapster: We could acually shaded few of these, I'm thinking about IOBiConsumer

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
melloware: I think you are mixing up the two methods. The execute in Request Scope requires a return value while executeStateless does not its a void.  One is putting a temp variable on the request scope an removing it when done.  This one is removing a State variable and putting it back when done.

Rapster (changes requested):

primefaces/src/main/java/org/primefaces/component/autocomplete/AutoCompleteRenderer.java
melloware: Inlined...

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
melloware: For example i need to remove `Constants.HELPER_RENDERER` from the state map, run some code, and then re-instate the `Constants.HELPER_RENDERER` value.

Rapster: You sure? Don't think I'm mixing anything here, also why having this null check before `requestMap.put(var, value);`?

Rapster: My coment is on executeInRequestScope not executeInStatelessScope 😄

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
Rapster: Ok got it, but I think the name doesn't say much about what the method really does

melloware: Yeah I couldn't think of a better name so I am open to suggestions.  Its basically removing a state variable, running, and restoring a state variable.

primefaces/src/main/java/org/primefaces/util/FailableRunnable.java
melloware: I don't have thoughts other than with like our COmmons IO issue its sometimes easier not to shade?  But I could be convinced either way 😄

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
melloware: Resolved

primefaces/src/main/java/org/primefaces/util/FailableRunnable.java
Rapster: Yeah... The difference I see here is that, as they are functional interfaces, the chances for them to change is very low?

Rapster: Another option would be to do like in omnifaces: https://github.com/omnifaces/omnifaces/blob/4.x/src/main/java/org/omnifaces/util/Callback.java

So we don't end up with too many java files?

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
Rapster: Not easy, I wish I could come up with something better than "pop and run" 😅

primefaces/src/main/java/org/primefaces/util/FailableRunnable.java
melloware: Oohhh I like that let me update and see if you like it.

melloware: OK submitted that to a new class called `Callbacks`.

Rapster: Looks nice, at this point, you can:
- rename IOBiConsumer into FailableBiConsumer
- move SerializableXXX classes into Callbacks
- rename ConsumerThree into TriConsumer and move it also

melloware: Done!

Rapster: Yup, looks good! WDYT?

primefaces/src/main/java/org/primefaces/util/Callbacks.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.Objects;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
+
+/**
+ * Collection of callback interfaces. Useful in (mini) visitor and strategy patterns.
+ */
+public final class Callbacks {
```
Rapster, line 37: private ctor

primefaces/src/main/java/org/primefaces/util/FailableRunnable.java
melloware: I like it. I think you are right I need a private CTOR

tandraschko:
why not just a callbacks package? IMO the more common way

Rapster:
As they're just extension of java.util, it makes sense to keep them in org.primefaces.util. I don't like having a deep and wide package structure, it always end up having a mess: the flatter the better.

Callbacks at this point is just a container, but could become what `Failable` is (from commons-lang), we'll most probably introduce new utils in the future, and no need to create another utils

One may argue how cumbersome it is to do someting like `Callbacks.FailableConsumer` but you can use static import and that should solve the "problem"

tandraschko:
all right, thats ok for me

static imports are even more mess in the most times, so lets better stay with it

melloware:
so I am OK to merge this I am very happy with this PR with @Rapster help where we landed on this!

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
Rapster: We can stick with this name for now

Rapster: I have a preference to call it "runInStatelessScope" to make it less verbose, same as the other one

melloware:
@Rapster how about `runWithoutStateVariable` since that is really what its doing?

Rapster:
i'm fine with that

tandraschko:
not sure about the name
its not a "state" var actually - "state "in JSF is always viewstate
better would be something like request attribute

melloware:
I changed it to `runWithoutFacesContextVar`

Rapster (approved):

PR #11816, opened 2024-04-25
Fix #11806: FileUpload switch error to dev warning

### LINKED ISSUE #11806: FileUpload: Decoding FileUpload requires contentType "multipart/form-data"
### Describe the bug

In FileUploadRenderer I see the following method:

 ```
        if (!fileUpload.isDisabled()) {
                LOGGER.severe(() -> "Decoding FileUpload requires contentType \"multipart/form-data\"." +
                        " Skipping clientId: " + component.getClientId(context));
                return;
            }

            PrimeApplicationContext applicationContext = PrimeApplicationContext.getCurrentInstance(context);

            FileUploadDecoder decoder = applicationContext.getFileUploadDecoder();
            decoder.decode(context, fileUpload);
        }
    }

```

On one of my pages, I get the above severe error message. I did set a breakpoint there and find that the request in question is indeed `content-type = application/x-www-form-urlencoded;` but it also is `faces-request = partial/ajax`.

This is just during rendering of the p:fileUpload. The upload itself is working perfectly fine.

Is the logging condition correct as it is, or shouldn't it exclude partial ajax requests here?

### Reproducer

    <h:head>
    </h:head>
    <h:body>

        <h:form id="frmTest">

          <p:fileUpload 
            auto="true" dragDropSupport="true" mode="advanced" skinSimple="false"
            update="@form" multiple="true" sequential="true" />

          <p:commandButton value="Refresh" partialSubmit="true" update="@form" />
        </h:form>

    </h:body>
</html>

### Expected behavior

The message should not be logged for partial updates.

### PrimeFaces edition

None

### PrimeFaces version

14.0.0-RC3

### Theme

_No response_

### JSF implementation

MyFaces

### JSF version

2.3

### Java version

21

### Browser(s)

_No response_

### PR DESCRIPTION
**Pull request by melloware**
Fix #11806: FileUpload switch error to dev warning

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/fileupload/FileUploadRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/fileupload/FileUploadRenderer.java
@@ -38,27 +38,24 @@ import javax.faces.context.ResponseWriter;
 import javax.faces.convert.ConverterException;
 import java.io.IOException;
 import java.util.Arrays;
-import java.util.logging.Logger;
 
 public class FileUploadRenderer extends CoreRenderer {
-
-    private static final Logger LOGGER = Logger.getLogger(FileUploadRenderer.class.getName());
 
     @Override
     public void decode(FacesContext context, UIComponent component) {
         FileUpload fileUpload = (FileUpload) component;
-        if (!fileUpload.isDisabled()) {
-            if (!context.getExternalContext().getRequestContentType().toLowerCase().startsWith("multipart/")) {
-                LOGGER.severe(() -> "Decoding FileUpload requires contentType \"multipart/form-data\"." +
-                        " Skipping clientId: " + component.getClientId(context));
-                return;
-            }
-
-            PrimeApplicationContext applicationContext = PrimeApplicationContext.getCurrentInstance(context);
-
-            FileUploadDecoder decoder = applicationContext.getFileUploadDecoder();
-            decoder.decode(context, fileUpload);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by tandraschko** [comment]
We can so remove the "return" to not break something

**Comment by melloware** [comment]
OK I removed the `return;` and I also short circuit the `isDisabled` check rather than having the nested IF.

File: primefaces/src/main/java/org/primefaces/component/fileupload/FileUploadRenderer.java [hunk 1]
```diff
@@ -38,27 +38,24 @@
...
-                LOGGER.severe(() -> "Decoding FileUpload requires contentType \"multipart/form-data\"." +
-                        " Skipping clientId: " + component.getClientId(context));
-                return;
-            }
+        if (fileUpload.isDisabled()) {
+            return;
+        }
+        if (!context.getExternalContext().getRequestContentType().toLowerCase().startsWith("multipart/")) {
+            logDevelopmentWarning(context,
+                    "Decoding FileUpload requires contentType \"multipart/form-data\". Skipping clientId: " + component.getClientId(context));
```
tandraschko: remove the word "Skipping", we dont skip anymore

melloware: Removed skipping!  new message is `"Decoding FileUpload requires contentType \"multipart/form-data\" for clientId: " + component.getClientId(context)`

PR #11825, opened 2024-04-26
Fix #11824: AjaxStatus always hide in oncomplete

### LINKED ISSUE #11824: AjaxStatus: Not returning to default when located outside the ajax partial submit update scope
### Describe the bug

I found this as a regression from PF 12.0.7:

When the `p:ajaxStatus` is not contained in the form that is refreshed, completing the action does not reset the status field to the default state.

The `p:ajaxStatus` should be updated in any case.

### Reproducer

```
<!DOCTYPE html>

    <h:head>
    </h:head>

    <h:body>

      <p:ajaxStatus id="busyStatus" style="display:block; padding: 0">
           
        <f:facet name="default">
        </f:facet>
  
        <f:facet name="start">
          <i class="pi pi-spin pi-spinner"
            style="font-size: 18px; margin-left: 2px margin-top: 2px"></i>
        </f:facet>
     
      </p:ajaxStatus>

      <h:form id="frmTest">

        <p:commandButton value="Submit" action="#{testView.action}" partialSubmit="true" update="@form" />
      </h:form>

    </h:body>
</html>

```

```
package org.primefaces.test;

import java.io.Serializable;

import javax.inject.Named;

import lombok.Data;

@Data
@Named
@ViewScoped

    public void action() {

        try {
            Thread.sleep(2000);
        } catch (InterruptedException e) {
            e.printStackTrace();
        }
    }
}

```

### Expected behavior

_No response_

### PrimeFaces edition

Community

### PrimeFaces version

14.0.0-RC3

### Theme

_No response_

### JSF implementation

All

### JSF version

2.3

### Java version

21

### Browser(s)

_No response_

melloware:
Fix #11824: AjaxStatus always hide in oncomplete

This answers the question asked previously...

> Could anyone explain this to me?

```js
if (event !== 'complete' || this.jq.children().filter(this.toFacetId('complete')).length) {
            this.jq.children().hide().filter(this.toFacetId(event)).show();
        }
```

Basically upon `oncomplete` we must call `facets.hide()` because we are done.

cc @stolp 

The code is easier to read below than in the change:

```js
    trigger: function(event, args) {
        var callback = this.cfg[event];
        if (callback) {
            callback.apply(document, args);
        }

        // Get the facet based on the event
        var facets = this.jq.children();
        var facet = facets.filter(this.toFacetId(event));

        // We have the following events:
        // 1) start
        // 2) success or error
        // 3) complete
        switch (event) {
            case 'start':
                // always hide other facets on start
                facets.hide();
                
                if (facet.length > 0) {
                    facet.show();
                }
                break;

            case 'success':
            case 'error':
                // we now expect that either a complete or success/error facet is defined
                // if no success/error is defined, lets just rely upon the complete-facet
                if (facet.length > 0) {
                    facets.hide();
                    facet.show();
                }
                break;

            case 'complete':
                var pfArgs = args[2];
                // if the current request leads in a redirect, skip hiding the previous facet (in best case this is the start-facet)
                // when a sucess/error-facet is defined, this wont work as expected as the 'redirect' information is not available before
                if (!pfArgs || pfArgs.redirect) {
                    return;
                }
                
                // #11824 we must hide all facets and show either complete or default
                facets.hide();

                // Show complete-facet if defined, else show default facet
                if (facet.length > 0) {
                    facet.show();
                }
                break;
        }
    }
 ```
 
 the only important change is moving the hide in the `complete`
 
 ```js
                // #11824 we must hide all facets and show either complete or default
                facets.hide();
```

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/ajaxstatus/AjaxStatus001.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/ajaxstatus/AjaxStatus001.java
@@ -1,0 +1,48 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2024 PrimeTek Informatics
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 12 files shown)


tandraschko:
we really need to some tests now
i fear breaking all the other cases now

melloware:
I will add some tests.

melloware:
@tandraschko added a battery of Integration tests that HELPED ME FIND A BUG! 

- tests all facets including similuting an `error` facet since that is hard to do
- a specific test for this actual use case from @stolp

tandraschko:
if you tested the linked issues and they work -> +1

PR #11832, opened 2024-04-29
Fix #11830: Spinner prevent value outside min/max range

### LINKED ISSUE #3223: Spinner: add input validation
## 1) Environment
- Affected browsers: all

## 2) Expected behavior
It must be guaranteed, that the value of p:spinner is between min and max. Just doing client side validation is insufficient. The server also has to validate the submitted value.

## 3) Actual behavior
Range checks are performed on client side via JS only. So it's easily possible to bypass validation at all.

## 4) Steps to reproduce
In developer tools (F12) update the spinner's value in the DOM to an invalid, say negative value:
`$('.ui-spinner-input').val("-1")` Then submit the form by pressing the button.

## 5) Sample XHTML
```xml
	<h:head>
		<title>Spinner test</title>
	</h:head>
	<h:form>
	    <p:spinner value="#{bean.value}" min="0" max="10" />
	</h:form>
</html>  
```

## 6) Sample bean
```java
package de.test.primefaces;

import java.io.Serializable;
import javax.inject.Named;

@Named("bean")
@SessionScoped

	private int value;
	
	public int getValue() {
		return value;
	}

	public void setValue(int value) {
		this.value = value;
	}
	
}
```

## 7) Mitigation
Just repeat the range checking in SpinnerRenderer. Reset the value to min if value<min, reset it to max if value>max. See https://www.owasp.org/index.php/Input_Validation_Cheat_Sheet#Client_Side_vs_Server_Side_Validation for further information and general guidelines.

### LINKED ISSUE #11830: Spinner: serverside input validation
### Describe the bug

The value of p:spinner is not validated on the serverside in relation to the min/max attributes.
I found the exact same issue here https://github.com/primefaces/primefaces/issues/3223 from 2018.
The issue was resolved but then reverted and never openend again.
I created a repoducer to show this is still an open issue in the newest version:
[primefaces-test-jakarta.zip](https://github.com/primefaces/primefaces/files/15150681/primefaces-test-jakarta.zip)

p:inputNumber for example does a serverside check in the Renderer java class and the propsed fix in the issue from 2018 worked analog to that. Was there a reason why this was reverted and never fixed?

### Reproducer

In developer tools (F12) update the spinner's value in the DOM to an invalid value (out of min/max range)
`$('.ui-spinner-input.ui-inputfield').val("10000")`
Then submit the form by pressing the button.
The value will now be shown as 1000 in "Value is:" altough max is set to 4.

### Expected behavior

The value of p:spinner must be validated in relation to min/max on the server side.

### PrimeFaces edition

None

### PrimeFaces version

13.0.8

### Theme

_No response_

### JSF implementation

MyFaces

### JSF version

4.0.2

### Java version

17

### Browser(s)

_No response_

melloware:
Fix #11830: Spinner prevent value outside min/max range
Fix #3223

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/spinner/SpinnerBase.java
+++ b/primefaces/src/main/java/org/primefaces/component/spinner/SpinnerBase.java
@@ -95,7 +95,7 @@ public abstract class SpinnerBase extends AbstractPrimeHtmlInputText implements 
     }
 
     public double getMin() {
-        return (Double) getStateHelper().eval(PropertyKeys.min, Double.MIN_VALUE);
+        return (Double) getStateHelper().eval(PropertyKeys.min, -Double.MAX_VALUE);
     }
 
     public void setMin(double min) {
--- a/primefaces/src/main/java/org/primefaces/component/spinner/SpinnerRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/spinner/SpinnerRenderer.java
@@ -48,7 +48,7 @@ public class SpinnerRenderer extends InputRenderer {
 
         String submittedValue = context.getExternalContext().getRequestParameterMap().get(spinner.getClientId(context) + "_input");
 
-        if (submittedValue != null) {
+        if (LangUtils.isNotEmpty(submittedValue)) {
             String prefix = spinner.getPrefix();
             String suffix = spinner.getSuffix();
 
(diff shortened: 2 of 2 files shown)


melloware:
I just learned something new: https://stackoverflow.com/questions/3884793/why-is-double-min-value-in-not-negative

PR #11837, opened 2024-04-30
Fix #11836 - SONAR: enable quality checks on JS/CSS files

### LINKED ISSUE #11836: SONAR: enable quality checks on JS/CSS files
### Description

For the very same reasons as we do for java files, we should also do it for JS/CSS files

report JS/CSS issues on sonarcloud

### Additional context

_No response_

Rapster:


melloware:
Are we worried about this creating too many issues 😀

Rapster:
Besides the 3.4K of "Unexpected var, use let or const instead", it's about 450 issues (some of them look legit) I don't think it's that much...

melloware:
That is not bad at all.   I guess we would just turn the "use let instead of var" rule off?

Rapster:
Yup I just did (you can either merge it, or wait for my local build to be over toee the final results)

PR #11839, opened 2024-04-30
SONAR: fix issues

Rapster:


tandraschko:
we are short before of 14.0 - thats nothing that should be done after RCs

PR #11851, opened 2024-05-02
Fix #11806: Fileupload ignore invalid content type

### LINKED ISSUE #11806: FileUpload: Decoding FileUpload requires contentType "multipart/form-data"

melloware:
Fix #11806: Fileupload ignore invalid content type


Rapster:
Not a big fan of this method logDevelopementWarning, we lose the source class of where the issue actually comes from. I think we should add an additional parameter

melloware:
That is true. I will update this PR for 15.0.0 but leave the simple change for 14.0.1

tandraschko:
we should add comment here:
// skip further processing as servlet fileupload decoding would throw a exception because of wrong contentType

melloware:
Updated for both your comments.  I removed a few of those log warnings as they are in deprecated methods being removed in another PR.

PR #11853, opened 2024-05-02
Fix #10909: Remove deprecated code

### LINKED ISSUE #10909: Remove Code in 15.0.0 Reminder
- [x] Remove Apache Commons FileUpload stuff (https://github.com/primefaces/primefaces/issues/10751)
- [x] Remove InputSwitch (https://github.com/primefaces/primefaces/issues/10874)
- [x] Remove legacy Chart.JS components and all PF models (https://github.com/primefaces/primefaces/issues/10870)
- [x] Remove `@Deprecated` legacy "table" layouts
- [x] Remove `@Deprecated` FileUpload properties
- [x] Remove `@Deprecated` DataExporter code

melloware:
Fix #10909: Remove deprecated code

@Rapster just two i need you to look at after this PR and its `DataExporter` and `FileUpload` some things marked deprecated looked like they were still in use so I didn't touch them.


primefaces/src/main/java/org/primefaces/component/selectmanycheckbox/SelectManyCheckboxRenderer.java
```diff
@@ -82,9 +82,6 @@ protected void encodeMarkup(FacesContext context, SelectManyCheckbox checkbox) t
             encodeCustomLayout(context, checkbox);
             writer.endElement("span");
         }
-        else if ("grid".equals(layout)) {
-            encodeLegacyTabularLayout(context, checkbox, layout);
```
jepsar, line 86: Maybe keep the else if and throw an exception?

primefaces/src/main/java/org/primefaces/component/selectoneradio/SelectOneRadioRenderer.java
```diff
@@ -69,9 +69,6 @@ protected void encodeMarkup(FacesContext context, SelectOneRadio radio) throws I
         if (custom) {
             encodeCustomLayout(context, radio);
         }
-        else if ("grid".equals(layout)) {
-            encodeLegacyTabularLayout(context, radio, layout);
```
jepsar, line 73: Throw?

primefaces/src/main/java/org/primefaces/component/selectmanycheckbox/SelectManyCheckboxRenderer.java
melloware: good idea

tandraschko:
are we sure we wanna remove tabular support? i can remember a discussion here and that i fixed something regarding :D

melloware:
We kept it in PanelGrid a its SO popular there and we removed the deprecation because so many people use PanelGrid with tabular :)

PR #11858, opened 2024-05-02
Fix #11857: Remove Commons IO shading

### LINKED ISSUE #11857: Core: Do not shade BoundedInputStream
See conversation: https://github.com/primefaces/primefaces/pull/11707

melloware:
Fix #11857: Remove Commons IO shading


tandraschko:
Puh
Tbh, we should not copy filenameutils
Thats just to big and special

PR #11859, opened 2024-05-02
Fix #11857: Remove BoundedInputStream shading

### LINKED ISSUE #11857: Core: Do not shade BoundedInputStream

melloware:
Fix #11857: Remove BoundedInputStream shading

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -335,7 +335,7 @@
             <dependency>
                 <groupId>commons-io</groupId>
                 <artifactId>commons-io</artifactId>
-                <version>2.15.1</version>
+                <version>2.16.0</version>
             </dependency>
 
             <dependency>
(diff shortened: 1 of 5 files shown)


tandraschko:
Looks much better IMO
Wdyt @Rapster?

Rapster:
The whole idea of no longer shading, is with recent upgrades it causes too many issues? That's why we stop shading IOUtils for example, right ?

I personnally prefer shading than copy/pasting even if it means to stick to one specific version of commons-io, now if the upgrade is really necessary and stop us to do some stuff, i'm cool with that

melloware:
That is a good point we could simply lock on 2.15.1 and have dependabot never notify us about commons IO.

I am ok with that if you guys are?

tandraschko:
also ok for now

PR #11867, opened 2024-05-03
Fix #11789: FeedReader switch to new library

### LINKED ISSUE #11789: FeedReader: Switch from Rome to rssreader library
> RSS Reader is a simple Java library for reading RSS and Atom feeds. It has zero 3rd party dependencies, a low memory footprint and can process large feeds. 

https://github.com/w3stling/rssreader

Reasons:

- [x] Zero dependencies unlike ROME which has dependencies on JDOM etc
- [x] Up to date last commit 3 weeks ago vs Rome which seems inactive
- [x] This should be transparent to most users as Rome was used in the component and not exposed.
- [x] Has special handling for Apple Itune Podcast RSS feeds
- [x] Does not XML parse but manually parses the documents to prevent previous XML related security issues like https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-33813

melloware:
Fix #11789: FeedReader switch to new library

- [x] Zero dependencies unlike ROME which has dependencies on JDOM etc
- [x] Up to date last commit 3 weeks ago vs Rome which seems inactive
- [x] This should be transparent to most users as Rome was used in the component and not exposed.
- [x] Has special handling for Apple Itune Podcast RSS feeds
- [x] Does not XML parse but manually parses the documents to prevent previous XML related security issues like https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-33813 
- [x] Quarkus friendly!
- [x] Integration tests updated and passing


primefaces/src/main/java/org/primefaces/component/feedreader/FeedInput.java
```diff
@@ -23,34 +23,86 @@
...
+import com.apptasticsoftware.rssreader.AbstractRssReader;
+import com.apptasticsoftware.rssreader.Item;
+import com.apptasticsoftware.rssreader.RssReader;
+import com.apptasticsoftware.rssreader.module.itunes.ItunesItem;
+import com.apptasticsoftware.rssreader.module.itunes.ItunesRssReader;
 
+/**
+ * Utility class to parse feed input into a list of FeedItem objects.
+ */
 public class FeedInput {
```
Rapster: Utility class, then:

- final?
- private ctor?
- static public methods ?

Rapster: Rename into RSSUtlis ? (FeedInput sounds very much like a POJO)

primefaces/src/main/java/org/primefaces/model/feedreader/FeedItem.java
```diff
@@ -0,0 +1,260 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Class representing a RSS item. A channel may contain any number of items. An item may represent a "story" -- much
+ * like a story in a newspaper or magazine; if so its description is a synopsis of the story, and the link points
+ * to the full story.
+ */
+public class FeedItem implements Serializable {
```
Rapster, line 36: Is this class really necessary? Can't we simply stick to Item from the lib?

primefaces/src/main/java/org/primefaces/component/feedreader/FeedInput.java
melloware: That's how PrimeTek had it before as FeedInput I think to make sure the classes for the library are not in the Renderer causing any classloading issues.  I will clean it up though as it is a bad name.

primefaces/src/main/java/org/primefaces/model/feedreader/FeedItem.java
melloware: I did this because this library wraps EVERY getter in an Optional.  This means in the EL you have to do #{feed.title.orElse("")} which to me is ugly and not very JSF like. So this also makes it Quarkus Native friendly by wrapping it in a normal POJO so it's just #{feed.title} and GraalVM reflection can just discover it as JavaBean compliant.

Rapster: Ok I see, I didn't see the optionals...

primefaces/src/main/java/org/primefaces/component/feedreader/FeedInput.java
Rapster: Not sure I get that point

> I think to make sure the classes for the library are not in the Renderer causing any classloading issues

Even if they are present in the renderer, it shouldn't be an issue

melloware: I am not a classloader expert so @tandraschko @jepsar can correct me but I know in some servers like Payara what happens is the `faces-config.xml` gets scanned at startup and all `Renderer` classes are classloaded.  If you use a class in a Renderer that is NOT present the startup fails.  So if you look at our Optional dependencies here https://primefaces.github.io/primefaces/14_0_0/#/gettingstarted/dependencies?id=dependencies none of those are actually used in a Renderer they are all loaded from a separate class.

So I assume that was PrimeTek's reasoning behind this `FeedInput` class being separate as the only place the `com.rometools` was imported.  

This is ALL just a guess based on past errors i have seen in Payara etc.?

melloware: Refactored into static `RSSUtils.java`

primefaces/src/main/java/org/primefaces/model/feedreader/FeedItem.java
Rapster, line 36: I'm thinking if instead of listing every attributes, to have a ctor with Item and all this class does is to delegate to the item and get rid of the optionables. There is one hiccup, class Item from the lib doesn't seem to be serializable, maybe we could request to make it so

primefaces/src/main/java/org/primefaces/component/feedreader/FeedInput.java
Rapster: > I am not a classloader expert so @tandraschko @jepsar can correct me but I know in some servers like Payara what happens is the `faces-config.xml` gets scanned at startup and all `Renderer` classes are classloaded. If you use a class in a Renderer that is NOT present the startup fails. So if you look at our Optional dependencies here https://primefaces.github.io/primefaces/14_0_0/#/gettingstarted/dependencies?id=dependencies none of those are actually used in a Renderer they are all loaded from a separate class.
> 
> So I assume that was PrimeTek's reasoning behind this `FeedInput` class being separate as the only place the `com.rometools` was imported.
> 
> This is ALL just a guess based on past errors i have seen in Payara etc.?

Not sure again what you're replying to 😅 (sorry...), but anyway RSSUtils looks fine

melloware: I was replying to your comment "Even if they are present in the renderer, it shouldn't be an issue" ?

primefaces/src/main/java/org/primefaces/model/feedreader/FeedItem.java
melloware: I thought of that too but chose this way ONLY to isolate the only place `com.apptasticsoftware.rssreader` would be is in the RSSutils and this FeedItem would just be a Serializable POJO.

primefaces/src/main/java/org/primefaces/component/feedreader/RSSUtils.java
```diff
@@ -0,0 +1,111 @@
...
+        if (isPodcast) {
+            reader = new ItunesRssReader();
+        }
+        else {
+            reader = new RssReader();
+        }
+        if (input instanceof String) {
+            return convert((List<Item>) reader.read((String) input).limit(size).collect(Collectors.toList()));
+        }
+        return convert((List<Item>) reader.read((InputStream) input).limit(size).collect(Collectors.toList()));
```
Rapster, line 71: you can pass size to convert method, and stop the loop once limit reached

melloware: But doesn't the "stream.limit(size)" already limit the size?

Rapster: Yeah, but if you have a huge feed, and you only want the first 10... Making me think, feed reader should be lazy-able (let's wait for someone to ask for it)

Anyway, at the end you should have something like this:

```java
if (input instanceof String) {
    return convert((List<Item>) reader.read((String) input, size);
}
return convert((List<Item>) reader.read((InputStream) input, size);

```

melloware: I will check with the author of the lib he advertises the library handles huge feeds with speed so is he returning a proper stream that only reads as necessary from the upstream feed?

Rapster: Ohhh I see! My bad I thought the whole list was passed and then chunked... Not easy to read all these parenthesis on github review

Rapster (approved):

PR #11873, opened 2024-05-04
#11506 LazyDataModel: no need to count if first == 0 and lazyData.size

### LINKED ISSUE #11506: LazyDataModel: no need to count if first == 0 and lazyData.size < pageSize

cocorossello:
Reopened the PR to see what makes the build fail

Fix #11506


melloware:
OK good just a single failing test @Rapster 

```
DataTable045Test.selectionMultipleWithPaging:53 expected: <true> but was: <false>
```

cocorossello:
The problem is in CustomerLazyDataModelNoCountImpl. 

I don't think this is correct however it is intentional:

```
    @Override
        return 0;
    }

```
Changing to return `datasource.size() `works

tandraschko:
Return 0 is a supported feature If you check the docs.

cocorossello:
ok, then this should do the trick.  The logic should be extracted to a method but I'm not sure on where should I put this this.

Maybe in LazyDataModel, something like

`    private void calculateRowCount(List<?> loadedData, Map<String, FilterMeta> filterBy, int rows, int first) {
        if (getRowCount() == 0) {
            if (data.size() < rows && first == 0) {
                setRowCount(data.size());
            }
            else if (first == 0) {
                setRowCount(count(filterBy));
            }
        }
    }`

tandraschko:
Maybe we should change the return type of count to a nullable int
Makes it more clear

cocorossello:
That is not possible, LazyDataModel extends javax.faces.model.DataModel and has
`
public abstract int getRowCount();

`

tandraschko:
Then we should refactor and docs to -1 instead of 0 maybe

melloware:
It looks like all Integration Tests are passing!

Rapster:
I'll have a look later in the week

melloware:
@Rapster just a reminder on this one.

PR #11878, opened 2024-05-05
Fix#11814: Support DefaultCommand for submit button outside (but linked to) form

### LINKED ISSUE #11814: DefaultCommand: Not working for button outside (but linked to) form
### Describe the bug

The DefaultCommand component only works when it targets CommandButtons that are linked to (using the form-attribute), but placed outside the form. See reproducer.

I'm not entirely sure if this should be classified as a bug or not, but I thought it would make sense if this usage would be supported (it sure would help in my use-case).

I already looked through the code and noticed that it indeed doesn't take into account buttons that are not placed inside the form, it only takes into account the form that is actualy enclosed by the target button: https://github.com/primefaces/primefaces/blob/3c1945e926507fe6831c414b962e31d68de47002/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.defaultcommand.js#L40

If possible I'd like to contribute with a fix for this!

### Reproducer

In the reproducer I created a form with an InputText, CommandButton "Submit inside form" and DefaultCommand. Next to the form there's another CommandButton "Submit outside form" that is linked to the form using the form-attribute. I configured the DefaultCommand to point to button "Submit outside form", but when I press the Enter-key when my focus is in the InputText, the form is submitted using button "Submit inside form".

[PrimeFaces_DefaultCommand_reproducer.zip](https://github.com/primefaces/primefaces/files/15113064/PrimeFaces_DefaultCommand_reproducer.zip)

Relevant portion from reproducer:
```xml
<h:form id="frmTest">
    <p:inputText value="#{testView.inputValue}" />
    <p:commandButton id="buttonInsideForm" value="Submit inside form" action="#{testView.submitActionFromInsideForm}" />
    <p:defaultCommand target="buttonOutsideForm" />
</h:form>

<p:commandButton id="buttonOutsideForm" value="Submit outside form" action="#{testView.submitActionFromOutsideForm}" form="frmTest" />
```

### Expected behavior

When pressing the enter key in the InputText, the form should be submitted by button "Submit outside form", as configured in the DefaultCommand component.

### PrimeFaces edition

Community

### PrimeFaces version

13.0.8

### Theme

_No response_

### JSF implementation

All

### JSF version

2.3

### Java version

17

### Browser(s)

_No response_

BjornJaspers:
Fix #11814 

This is how I thought about fixing this, but of course I'm open for discussion.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/commandlink/CommandLinkRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/commandlink/CommandLinkRenderer.java
@@ -72,7 +72,7 @@ public class CommandLinkRenderer extends CoreRenderer {
         ResponseWriter writer = context.getResponseWriter();
         String clientId = link.getClientId(context);
         Object label = link.getValue();
-        String referencedForm = link.getForm();
+        String form = link.getForm();
 
         String request;
         boolean ajax = link.isAjax();
@@ -102,20 +102,20 @@ public class CommandLinkRenderer extends CoreRenderer {
             writer.writeAttribute("tabindex", "-1", null);
         }
 
-        if (!isValueBlank(referencedForm)) {
-            writer.writeAttribute("data-pf-form", referencedForm, null);
+        if (!isValueBlank(form)) {
+            writer.writeAttribute("data-pf-form", form, null);
         }
 
         if (ajax) {
             request = buildAjaxRequest(context, link);
         }
         else {
-            UIForm form = ComponentTraversalUtils.closestForm(link);
-            if (form == null) {
+            UIForm uiForm = ComponentTraversalUtils.closestForm(link);
+            if (uiForm == null) {
                 throw new FacesException("Commandlink \"" + clientId + "\" must be inside a form component");
             }
 
-            request = buildNonAjaxRequest(context, link, form, clientId, true);
+            request = buildNonAjaxRequest(context, link, uiForm, clientId, true);
         }
 
         if (csvEnabled) {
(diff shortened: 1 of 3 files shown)


melloware (approved):

melloware:
This actually looks really reasonable to me.

tandraschko:
what about normal button, linkbutton or commandlink?

BjornJaspers:
> what about normal button, linkbutton or commandlink?

I updated the implementation to support commandlink. Referencing a form in a normal button and linkbutton doesn't seem to be supported however.

melloware:
Agreed this would only work from CommandXXX types thanks for adding CommandLink

primefaces/src/main/java/org/primefaces/component/commandlink/CommandLinkRenderer.java
```diff
@@ -72,6 +72,7 @@ protected void encodeMarkup(FacesContext context, CommandLink link) throws IOExc
         ResponseWriter writer = context.getResponseWriter();
         String clientId = link.getClientId(context);
         Object label = link.getValue();
+        String referencedForm = link.getForm();
```
tandraschko: i would just call the var "form"

tandraschko:
+1 then

melloware:
@BjornJaspers unfortunately this fix cannot be merged in the 13.X branch that you have it marked for.  Can you resubmit this under branch `master` and leave it up to us to backport to 14.X.   13.X is now under LTS by PrimeTek so only PRO users will received fixes on 13.X

PR #11902, opened 2024-05-09
Fix #11570: Autocomplete allow CTRL+CLICK on multiple selection

### LINKED ISSUE #11570: AutoComplete: Enable ctrl+click/shift+click when using multiple='true'
### Description

When using the autocomplete component in multiple mode, users should be able to select multiple similar suggestions at same time. Currently you have to type the same filter characters several times if you want to select more than one suggestion that match the filter characters.

When suggestions are shown, users should be able to use control+click or shift+click to select multiple suggestions at the same time without having to retype the required number of filter characters. Some components already have this (e.g. SelectManyMenu and PickList).

### Additional context

_No response_

melloware:
Fix #11570: Autocomplete allow CTRL+CLICK on multiple selection

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
@@ -124,18 +124,15 @@ PrimeFaces.widget.AutoComplete = PrimeFaces.widget.BaseWidget.extend({
         this.cfg.minLength = this.cfg.minLength != undefined ? this.cfg.minLength : 1;
         this.cfg.cache = this.cfg.cache || false;
         this.cfg.dropdownMode = this.cfg.dropdownMode || 'blank';
-        this.cfg.autoHighlight = (this.cfg.autoHighlight === undefined) ? true : this.cfg.autoHighlight;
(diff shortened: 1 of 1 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/autocomplete/autocomplete.js
```diff
@@ -128,13 +128,11 @@ PrimeFaces.widget.AutoComplete = PrimeFaces.widget.BaseWidget.extend({
...
-        this.cfg.dynamic = this.cfg.dynamic === true ? true : false;
-        this.cfg.autoSelection = this.cfg.autoSelection === false ? false : true;
-        this.cfg.escape = this.cfg.escape === false ? false : true;
-        this.cfg.hasFooter = this.cfg.hasFooter === true ? true : false;
-        this.cfg.forceSelection = this.cfg.forceSelection === true ? true : false;
+        this.cfg.dynamic = !!this.cfg.dynamic;
+        this.cfg.autoSelection = !!this.cfg.autoSelection;
+        this.cfg.escape = !!this.cfg.escape;
+        this.cfg.hasFooter = !!this.cfg.hasFooter;
+        this.cfg.forceSelection = !!this.cfg.forceSelection;
```
melloware: This fixes Sonar JS warnings while I was in there.

melloware:
I also fixed all the SONAR issues reported in autocomplete.js

PR #11911, opened 2024-05-11
Fix #11885 - DataTable: Select/unselect all with selectionPageOnly="false" does not work properly

### LINKED ISSUE #11885: DataTable: Select/unselect all with selectionPageOnly="false" does not work properly
### Describe the bug

Having a dataTable with selectionPageOnly="false" and paginator so i could select/unselect all items in a table. If i select all items in column header checkbox, it works fine, then i press button which processes whole form. After that i press the same checkbox to unselect all, but only single page is unselected. 

Whole functionality works badly, for example if i select all and unselect all right after, it works.. but if i press the button between those steps or sometimes even just changing page, it does not unselect for all pages.

### Reproducer

[datatableUnselectAll.zip](https://github.com/primefaces/primefaces/files/15234012/datatableUnselectAll.zip)

### Expected behavior

When selectionPageOnly="false" is set, i would expect it to work also for unselecting all items across all pages.

### PrimeFaces edition

Elite

### PrimeFaces version

13.0.9

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0.6

### Java version

17

### Browser(s)

_No response_

Rapster:
Fix #11885

### DIFF
--- a/primefaces-integration-tests/src/main/webapp/datatable/dataTable006.xhtml
+++ b/primefaces-integration-tests/src/main/webapp/datatable/dataTable006.xhtml
@@ -39,7 +39,10 @@
             </p:dataTable>
 
             <p:commandButton id="button" value="Submit" update="@form" action="#{dataTable006.submit}"/>
-            <p:commandButton id="toggleSelectPageOnly" value="Selection Page Only" update="@form" action="#{dataTable006.toggleSelectPageOnly}"/>
+            <p:selectBooleanButton id="toggleSelectPageOnly" value="#{dataTable006.selectionPageOnly}"
+                                   onLabel="Page only" offLabel="All pages">
+                <p:ajax update="@form"/>
+            </p:selectBooleanButton>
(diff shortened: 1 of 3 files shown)


Rapster:
@melloware could it be that simple? Just want to make sure I'm not missing anything

melloware:
Ha let me test!  It could be that simple!

Rapster:
draft for now, I have doubts 😄

christophs78:
This one IMO would deserve a Integrationtest.

melloware:
@christophs78 i think we have some checkAll tests but my guess is we are testing this exact scenario.  Maybe we can modify or enhance an existing test?

christophs78:
Yes, i think so too. I'll look in the evening.

christophs78:
`DataTable006Test ` should be a good base to build upon. Enhance exisiting tests or add some new mutations.

melloware (approved):

PR #11923, opened 2024-05-13
Fix #11914: Allow both URN and URI Taglib

### LINKED ISSUE #11914: Core: Support URN and URI Namespaces?
To catch up Faces itself, migrate XML namespaces as well:

`http://primefaces.org/ui --> primefaces.ui`

### Current:
```xml
```

### To:
```xml
xmlns:p="primefaces"
```

Hopefully there is a clever way to support both?

melloware:
Fix #11914: Allow both URN and URI Taglib

@tandraschko this does similar to what OmniFaces is doing to maintain both for a while it basically copies the taglib and renames the namespace so the JAR has both taglibs.

I also tested this is working with our Jakarta JAR as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ For a full list of the available downloads, please visit the [download page](htt
 	xmlns:pt="jakarta.faces.passthrough"
 	xmlns:jsf="jakarta.faces"
 	xmlns:ui="jakarta.faces.facelets"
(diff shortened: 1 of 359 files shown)


melloware:
You are right I will switch it to primefaces

melloware:
ok switched to `primefaces` as the URN.

tandraschko:
+1

PR #11924, opened 2024-05-13
Fix #11914: 14.0.1 Allow both URN and URI Taglib

### LINKED ISSUE #11914: Core: Support URN and URI Namespaces?

melloware:
Fix #11914: 14.0.1 Allow both URN and URI Taglib


tandraschko:
IMO we should only port bugfixes and not new features in general 
+0 from my side

pizzi80:
it's an addon, 
No one will complain about that ;)

christophs78:
+1

PR #11943, opened 2024-05-15
Fix #11938: CSV switch date parser to moment.js

### LINKED ISSUE #11938: CSV: Date validation requires legacy Calendar
### Describe the bug

The `validation.converter.js` requires the legacy Calendar to be loaded on the page else you get an error.

```js
if(timePattern && datePattern) {
                    return $.datepicker.parseDateTime(datePattern, timePattern, submittedValue, locale, {timeFormat:timePattern});
                }
                else if(timePattern) {
                    return $.datepicker.parseTime(timePattern, submittedValue, locale);
                }
                else {
                    return $.datepicker.parseDate(datePattern, submittedValue, locale);
                }
            }
            catch(exception) {
                var now = $.datepicker.formatDate(pattern, new Date(), locale);
```

Which leads to

image

### Reproducer

_No response_

### Expected behavior

Date CSV should work

### PrimeFaces edition

Community

### PrimeFaces version

14.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0

### Java version

11

### Browser(s)

_No response_

melloware:
Fix #11938: CSV load $.datepicker if not already on page

@tandraschko none of my choices were great here but here is what I came up with.

-  We moved calendar.js to its own file in PF 13 as its "deprecated" and we reducde the size of our jquery-ui-plugins.js by doing this.
- This means its not guaranteed to be on the page when using CSV however those `$datepicker` functions are really valuable.

So my choices were....

1. Copy all that `parse` and `format` code out of $.datePicker into our own utilities but it ended up being a LOT of code.  So I didn't like that.

2. So my easier solution was to check for the existence of `$.datepicker` and if not found dynamically load the script.  This works but I am sure you will hate.


tandraschko:
TBH i dont like that

Isn't there any native js parsing available?

Otherwise maybe:
1) try datepicker
2) try momentjs
3) throw exception?

melloware:
agreed but i wasn't sure of all the insane scenarios these 3 methods actually cover and convert.

```js
                if(timePattern && datePattern) {
                    return $.datepicker.parseDateTime(datePattern, timePattern, submittedValue, locale, {timeFormat:timePattern});
                }
                else if(timePattern) {
                    return $.datepicker.parseTime(timePattern, submittedValue, locale);
                }
                else {
                    return $.datepicker.parseDate(datePattern, submittedValue, locale);
                }
```

Maybe I am overthinking it and moment.js will do it?

melloware:
Switched to Moment.js.

melloware:
I am going to write some integration tests and improve on this implementation.

tandraschko:
shouldnt we split momemtjs + plugin in 2 files and merge? otherwise it could get lost on updating the script

melloware:
Good idea plugins are separated out and joined in the Optimizer.

melloware:
Last step is I am going to add Bean Validation Integration Tests.

tandraschko:
+1

PR #11946, opened 2024-05-15
Fix #11929: ConfirmDialog/Popup yes and no labels and classes

### LINKED ISSUE #11929: Confirm: ability to adjust button label and class of global confirmation
### Description

It would be nice to be able to adjust button labels and classes of a global confirm dialog via `<p:confirm>`. It's currently already possible to adjust the header and message of a global confirm dialog but nothing else than that.

Imaginary code:
```
<p:confirm header="Confirm delete"
           message="Are you sure to delete this?"
           yesButtonLabel="Delete this"
           yesButtonClass="ui-button-danger"
           noButtonLabel="Keep this!" />
```
- `yesButtonLabel`: overrides label of 'Yes' button (and restores it before the global confirm dialog is reused elsewhere)
- `yesButtonClass`: adds given class to 'Yes' button (and removes it before the global confirm dialog is reused elsewhere)
- `noButtonLabel`: overrides label of 'No' button (and restores it before the global confirm dialog is reused elsewhere)
- `noButtonClass`: adds given class to 'No' button (and removes it before the global confirm dialog is reused elsewhere)

### Additional context

Currently I'm working around this with a patched `PrimeFaces.widget.ConfirmDialog` which inspects a custom `type` in `showMessage()` function.

melloware:
Fix #11929: ConfirmDialog/Popup yes and no labels and classes

image

### DIFF
--- a/docs/15_0_0/gettingstarted/whatsnew.md
+++ b/docs/15_0_0/gettingstarted/whatsnew.md
@@ -16,6 +16,12 @@ Look into [migration guide](https://primefaces.github.io/primefaces/15_0_0/#/../
 * Captcha
     * Added [hCaptcha](https://www.hcaptcha.com/) support
     
+* ConfirmDialog/ConfirmPopup
+    * `yesButtonLabel`: overrides label of 'Yes' button (and restores it before the global confirm dialog is reused elsewhere)
+    * `yesButtonClass`: adds given class to 'Yes' button (and removes it before the global confirm dialog is reused elsewhere)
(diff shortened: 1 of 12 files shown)


FlipWarthog:
Can/should this include new tests under ConfirmDialog001Test and ConfirmPopup001Test? Looking at what's there, it probably shouldn't be too terrible to add some that check the overrides and that they get reverted.

melloware:
Good idea I will add some ITs. Should be easy you are right!

melloware:
IT's added

PR #11954, opened 2024-05-16
Fix #11951/#11928: DatePicker accessibility fixes

### LINKED ISSUE #11928: DatePicker: cannot type in datepicker when first displayed (14.0.0)
### Describe the bug

Since PF 14.0.0, datepicker input seems is now writable for some reason (however it isn't readonly).  It is writeable after a date select event. 

### Reproducer

Reproducer in https://github.com/cocorossello/primefaces-test/

Code:

```
      xmlns:f="http://xmlns.jcp.org/jsf/core">
<f:view>
    <h:head></h:head>
    <h:body>
        <p:datePicker/>
    </h:body>
</f:view>
</html>

```

Just click in the input and type some keys. It used to work in 13.0.x

Mojarra/Myfaces does not make a difference

### Expected behavior

Input should be writeable

### PrimeFaces edition

Community

### PrimeFaces version

14.0.0

### Theme

_No response_

### JSF implementation

All

### JSF version

2.3-NEXT

### Java version

17

### Browser(s)

_No response_

### LINKED ISSUE #11951: DatePicker: yearNavigator is no longer rendered as a dropdown
### Describe the bug

Hi!

When setting `yearNavigator="true"` the dropdown is not appearing, instead a spinner is shown. I don't know if the component has been changed deliberately or if it is a bug. Anyway, an option to choose which to use would be loved.
Users report that now they need to do lot of clicks to choose a year, due to the fact that the spinner has some "problems" entering the year manually (I think it's due to some regex applied to the input to let only certain years)
- Arrow keys doesn't work
- Selecting year and then starting to write a new year doesn't work

Also, manual input of the date is complicated to do because getting the field to have focus is very hard. When the field get focus, date popup shows, you cannot write manually because focus is lost. Then, if you click in the input, you still can't write on the field, I think it tries to show the popup again (after some test, with 'esc' you can manage to have the cursor in the input to start writing).
It would be great if, after clicking the field when the popup is oppened, it would be closed and the field would still have focus.

Regards!

[primefaces-test-master-dates.zip](https://github.com/primefaces/primefaces/files/15331487/primefaces-test-master-dates.zip)

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

14.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

-

### Java version

17

### Browser(s)

_No response_

melloware:
Fix #11951: DatePicker time inputs allow arrow keys
Fix #11928: DatePicker allow text input

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
@@ -2124,6 +2124,8 @@
             }
 
             switch (event.key) {
+                case 'ArrowDown':
+                case 'ArrowUp':
                 case 'Tab':
                 case 'Delete':
                 case 'Backspace':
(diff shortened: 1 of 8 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
```diff
@@ -2719,8 +2743,11 @@
             }
 
             var $this = this;
-            $(window).on('resize.' + this.options.id, function() {
-                $this.hideOverlay();
+            $(window).on('resize.' + $this.options.id, function() {
+                // TODO: lazy model for some reason triggers a windows resize
```
melloware, line 2747: @jepsar @christophs78 something to look at in the future but DatePicker012 test was failing and i debugged and dumped JS stacks to figure out for some reason window resize is called but only in that specific 012 Lazy scenario. Its like the redrawing of the lazy panel somehow triggers a window resize i could not figure out but didn't want to hold up this PR as it fixes a bunch of accessibility issues.

PR #11956, opened 2024-05-16
#11952: add expandIcon/collapseIcon attributes to RowToggler

### LINKED ISSUE #11952: DataTable: Allow using custom icons for p:rowToggler
### Description

We'd like to customize the icons being rendered by `p:rowToggler`, as far as I'm concerned after reading the documentation and looking at `RowToggler`/`RowTogglerRenderer` this isn't possible right now in an intuitive way as icons [are hardcoded](https://github.com/primefaces/primefaces/blob/14.0.0/primefaces/src/main/java/org/primefaces/component/rowtoggler/RowToggler.java#L36-L37)

Attributes to define icons being used for expanded/collapsed states like this:
`<p:rowToggler expandIcon="foo" collapseIcon="bar"/> `

### Additional context

_No response_

jungm:
Resolves #11952

### DIFF
--- a/docs/15_0_0/components/rowtoggler.md
+++ b/docs/15_0_0/components/rowtoggler.md
@@ -15,11 +15,15 @@ RowToggler is a helper component for datatable.
 
 ## Attributes
 
-| Name | Default | Type | Description | 
-| --- | --- | --- | --- |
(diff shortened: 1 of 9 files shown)


melloware (approved):
Excellent PR

PR #11964, opened 2024-05-18
#11962: confirmMessage facet

### LINKED ISSUE #11962: Confirm: message facet
### Description

The `p:confirm` `message` attribute is quite limited to express formatted content.
Hence we add support for the following construct:
```xml
<p:commandButton ...>
    <p:confirm header="title"/>
    <f:facet name="confirmMessage">
        fancy content here...
    </f:facet>
</p:commandButton>
```

`p:confirm` uses the `confirmMessage` facet of its parent to yield the message content.

### Additional context

_No response_

fcorneli:
Fix #11962

### DIFF
--- a/primefaces/src/main/java/org/primefaces/behavior/confirm/ConfirmBehavior.java
+++ b/primefaces/src/main/java/org/primefaces/behavior/confirm/ConfirmBehavior.java
@@ -23,14 +23,17 @@
  */
 package org.primefaces.behavior.confirm;
 
+import java.io.IOException;
 import javax.faces.FacesException;
 import javax.faces.component.UIComponent;
 import javax.faces.component.behavior.ClientBehaviorContext;
 import javax.faces.context.FacesContext;
+import javax.faces.context.ResponseWriter;
 import org.primefaces.behavior.base.AbstractBehavior;
 import org.primefaces.behavior.base.BehaviorAttribute;
 import org.primefaces.component.api.Confirmable;
 import org.json.JSONObject;
+import org.primefaces.util.FastStringWriter;
 
 public class ConfirmBehavior extends AbstractBehavior {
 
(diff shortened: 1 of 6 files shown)


primefaces/src/main/java/org/primefaces/behavior/confirm/ConfirmBehavior.java
```diff
@@ -78,7 +81,26 @@ public String getScript(ClientBehaviorContext behaviorContext) {
         String source = getSource();
         String type = JSONObject.quote(getType());
         String headerText = JSONObject.quote(getHeader());
-        String messageText = JSONObject.quote(getMessage());
+
+        String messageText;
+        UIComponent messageFacetComponent = component.getFacet("confirmMessage");
+        if (null != messageFacetComponent) {
```
melloware:
```suggestion
        if (FacetUtils.shouldRenderFacet(messageFacetComponent)) {
```

primefaces/src/main/java/org/primefaces/behavior/confirm/ConfirmBehavior.java
```diff
@@ -78,7 +81,26 @@ public String getScript(ClientBehaviorContext behaviorContext) {
...
+            ResponseWriter originalWriter = context.getResponseWriter();
+            FastStringWriter fsw = new FastStringWriter();
+            ResponseWriter clonedWriter = originalWriter.cloneWithWriter(fsw);
+            context.setResponseWriter(clonedWriter);
+            try {
+                messageFacetComponent.encodeAll(context);
+            }
+            catch (IOException ex) {
+                fsw.write("error encoding confirm message facet");
```
melloware: I think we typically just let these kinds of errors just bubble up?

melloware (changes requested):

primefaces/src/main/java/org/primefaces/behavior/confirm/ConfirmBehavior.java
melloware: Actually use `ComponentUtils.encodeComponent` which wraps a lot of this logic up nicely!

fcorneli: Thanks for the review. Indeed nice utility functions. Will use these.

melloware (approved):

PR #11975, opened 2024-05-21
Fix #11783: Datatable keyboard/screen reader support for cells

### LINKED ISSUE #11783: DataTable: Provide screen reader compliant keyboard navigation
### Description

One can use the keyboard to navigate between the rows by using the up and down 
arrow keys once the focus is moved from the table headers to the table body,
as demonstrated in the showcase here: 

http://www.primefaces.org:8080/showcase/ui/data/datatable/lazy.xhtml

But this approach is insufficient in combination with screan readers (e.g. JAWS) 
and thus is a serious accessibility limitation.

The problem is that the data tables are given the aria role "grid" so that 
screen readers expect the "W3C Pattern Grid Interactive Tabular Data" (see below)
and expect the table to be navigated via the cursor keys from row to row and from cell to cell 
similiar to excel, which is not implemented.
The focus currently stays on top on tbody instead of the "cells" so that the screen 
reader does not read anything from the table content during navigation.

1) It would be nice to have an implementation of table navigation from row to row and cell to cell 
via cursor keys/ home / end / page up / page down according to the example in the WAI Pattern:
https://www.w3.org/WAI/ARIA/apg/patterns/grid/examples/data-grids/

2) Instead of staying on tbody the keyboard focus would need to follow the cursor 
so that the screen readers will read the cells content.

### Additional context

This is the heaviest finding in the acessibility assessment of our primefaces application.

Thank you for your great work on primefaces!

melloware:
Fix #11783: Datatable keyboard/screen reader support for cells

**Financially Sponsored by PRO User ** @Rene-Stoeckel

- [x] Enables keyboard navigation and screen reader support for cells
- [x] new property `cellNavigation` true by default but can be disabled if someone runs into issues
- [x] `cellNavigation` is set to false on Row Selection and Cell Edit because they have their own keyboard handling.

image

datatable-navigation

### DIFF
--- a/docs/14_0_0/components/datatable.md
+++ b/docs/14_0_0/components/datatable.md
@@ -23,6 +23,7 @@ DataTable displays data in tabular format.
 | ariaRowLabel              | null               | String           | Label to read by screen readers on checkbox selection.
 | binding                   | null               | Object           | An el expression that maps to a server side UIComponent instance in a backing bean
 | cellEditMode              | eager              | String           | Defines the cell edit behavior, valid values are "eager" (default) and "lazy".
+| cellNavigation            | true               | Boolean          | Enables cell navigation with the keyboard for WCAG and screen reader compliance.
 | cellSeparator             | null               | String           | Separator text to use in output mode of editable cells with multiple components.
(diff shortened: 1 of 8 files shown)


melloware:
I need to update the docs too.

Issue #11994, opened 2024-05-23
TabView: Memory leak on Tab close


melloware:
### Describe the bug

When closing a tab the closed tab must clean up all its DOM elements to prevent detached elements and DOM memory leak.

### Expected behavior

When closing a tab it should not leave behind any detached DOM elements.

### PrimeFaces edition

Community

### PrimeFaces version

14.0.0

### JSF implementation

All

### JSF version

2.3

### Java version

11

PR #11995, opened 2024-05-23
Fix #11993: OverlayPanel cleanup DOM

### LINKED ISSUE #11993: OverlayPanel: Memory Leak on destroy
### Describe the bug

Currently when an overlay widget is cleaned up it calls `jq.remove()` to remove its contents from the DOM which is good.  

However, if you have say a DataTable inside the overlay which has lots of `.data()` jquery items it leaves it all as detached elements in the DOM when you destroy the overlay.

### Reproducer

_No response_

### Expected behavior

When destroying and overlay panel there should be no detached DOM elements.

### PrimeFaces edition

Community

### PrimeFaces version

14.0.0

### Theme

_No response_

### JSF implementation

All

### JSF version

2.3

### Java version

11

### Browser(s)

_No response_

melloware:
Fix #11993: OverlayPanel cleanup DOM


tandraschko:
should it be required in in BaseWidget _destroy then?

melloware:
I am still investigating. It should be cleaned up by the base widget destroy but it wasn't.  Let me dig deeper.

melloware:
OK the real issue is this one https://github.com/primefaces/primefaces/issues/11994

PR #11998, opened 2024-05-23
Fix #11994: TabView prevent detached DOM elements on TabClose

### LINKED ISSUE #11994: TabView: Memory leak on Tab close

melloware:
Fix #11994: TabView prevent detached DOM elements on TabClose

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
@@ -1134,16 +1134,19 @@ if (!PrimeFaces.utils) {
                     jq.removeAttr(attributeName);
                 }
             }
+
+            // Remove the element from the DOM and trigger onRemove events for widget.destroy.
+            // IMPORTANT: This must occur before jq.off() to ensure the on("remove") events remain registered.
+            if (removeElement) {
+                jq.remove();
+            }
+
             // Remove event listeners
             jq.off();
-            
+
             // Clear data
             if (clearData) {
                 jq.removeData();
-            }
-            // Remove elements itself from memory
-            if (removeElement) {
-                jq.remove();
             }
         },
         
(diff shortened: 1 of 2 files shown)


tandraschko:
why did you refactor like this?

#fireTabCloseEvent should just fire the event, everything else was done before in #remove

melloware:
So i will double check but since `cleanse` removes all elements, events, data off all elements when using `<p:ajax event="tabClose"` it could affect the lifecycle anything that was on that tab expecting the normal widget lifecycle so I chose to do if using `tabClose` to wait until the AJAX is done to clean up the DOM else just clean it up immediately if not using `tabClose`.

It felt like the cleaner / safer way to do it but I could be convinced otherwise?

tandraschko:
both elements were already destroyed before sending the event in the past and also feels better for me
as there is no delay waiting for the ajax response, its just a notifier and we never had any feature request for something here

melloware:
Ok I will test and update.

melloware:
@tandraschko did some testing in my clients real world app.

### Test with just changing to `cleanseElement` instead of `panel.remove()`

Here is what it looks like with this PR where the cleans runs BEFORE the AJAX runs.  Memory recovered back from 16.5MB to 13.3MB.

image


### Current PR
Here is what it looks like with this PR where the cleans runs AFTER the AJAX runs and all the widget.destroy's properly take place.  Memory recovered back to from 16.5MB to 9.5MB and barely any detached elements in memory.

image

melloware:
@tandraschko ok PR updated I got to the bottom of it.  In `cleansDomElement` we must call `jq.remove()` before removing all event listeners with `jq.off()`  This is so the widget `on("remove)` events get fire to register widgets as detached widgets so they clean themselves up.

I have verified with my testing above the memory is now recovered properly in my test scenario.

PR #12002, opened 2024-05-23
Fix #12001: Preserve order of constants

### LINKED ISSUE #12001: ImportConstants/ImportEnum: Preserve order of constants
### Describe the bug

In the `ConstantsHashMap` and `EnumHashMap` classes, a base class of `HashMap` is used, which will not preserve the order of values added to the map. This will result in enum members being returned out of order for example in a `<f:selectItems value="#{SomeEnum}" />`. This can be fixed by changing the base class to `LinkedHashMap`, which is a thin wrapper over `HashMap` that maintains insertion order. This is how OmniFaces handles these values ([source](https://github.com/omnifaces/omnifaces/blob/4.x/src/main/java/org/omnifaces/taghandler/ImportConstants.java#L160)).

(note: I'll be opening a PR for this fix shortly)

### Reproducer

_No response_

### Expected behavior

The order of constants/enum members should be preserved from how they are defined in the corresponding class.

### PrimeFaces edition

None

### PrimeFaces version

14.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

11

### Browser(s)

_No response_

jbrower2:
Fix #12001: Preserve order of constants

Changes the base class of `ConstantsHashMap` and `EnumHashMap` to be `LinkedHashMap`, to preserve the order of constants.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/importconstants/ConstantsHashMap.java
+++ b/primefaces/src/main/java/org/primefaces/component/importconstants/ConstantsHashMap.java
@@ -23,17 +23,17 @@
  */
 package org.primefaces.component.importconstants;
 
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 
 import javax.faces.FacesException;
 
 /**
- * Custom {@link HashMap} which throws an {@link FacesException} if the key/constant does not exist.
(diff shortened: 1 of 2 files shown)


melloware:
This makes sense to me especially if you are getting unexpected ordering and is probably why OmniFaces did the LinkedHashMap.  Thanks!

melloware (approved):

PR #12006, opened 2024-05-24
SelectOneRadio remove deprecated showcase example

melloware:


jepsar:
DIdn't we decide to deprecate "Referenced Custom Layout" as well? I guess we just talked about it. Can't find it in the migration guides.

melloware:
I think we talked about it but didn't decide on it. I think it was in relation to making it ARIA compliant?

PR #12012, opened 2024-05-25
Fix #12004: QR Code log development warning if width/height not set.

### LINKED ISSUE #12004: Barcode: default QR code size
### Description

Like other barcode types, barcode for QR code should have a default scannable size.  Currently its width is 100%. On the showcase page, its width and height are set to 125, which can be its default size. 

change QR default size to 125 x 125.

### Additional context

_No response_

melloware:
Fix #12004: QR Code log development warning if width/height not set.


tandraschko:
Tbh i dont like that
More next week

tandraschko:
IMO we should just do nothing here

its a svg and default in html is 100%, we shouldnt change something here

PR #12013, opened 2024-05-25
Fix 12009: DatePicker012Test flaky on Chrome

### LINKED ISSUE #12009: IT: Nightly Build DatePicker012Test failing Chrome
@christophs78 

DatePicker012 runs for me locally but it looks like in the nightly IT its failing on Chrome only?  I am running locally with Chrome CSP and non-CSP and I can't get it to fail.

https://github.com/primefaces/primefaces/actions/runs/9233425137

I have this TODO in `0-datepicker.js` because in lazy mode i am seeing it trigger a browser resize in the automation tests which closes the panel.

```js
            $(window).on('resize.' + $this.options.id, function() {
                // TODO: lazy model for some reason triggers a windows resize
                if (!$this.options.lazyModel) {
                    $this.hideOverlay();
                }
            });
```

Everything is working when I run manually in the showcase with animations disables so there is something weird that I can figure out going on here.

christophs78:


christophs78:
This fix or call it work-around seems to stabilize the IT on my dev-notebook and on Github actions using `@RepeatedTest(50)`. So I think we´ll try it even when the work-around is not 100% logical.

melloware:
thank you!

PR #12014, opened 2024-05-26
Fix #11980: Auto detect HTML5 compliance

### LINKED ISSUE #11980: Auto detect HTML5_COMPLIANCE
in Faces impls we detect if the current view uses HTML5 doctype

i think PF could do the same

melloware:
Fix #11980: Auto detect HTML5 compliance


tandraschko:
It works like expected?

melloware:
I am still testing but I am wondering instead of caching the Methods if we just cache the Boolean result?

tandraschko:
Dont think its worth
We will just switch to normal method call when we remove javax support

melloware:
Understood.

PR #12025, opened 2024-05-29
Fix #12024: Rating respect required=true

### LINKED ISSUE #12024: Rating: required="true" not working
### Describe the bug

Reported by @jepsar on Discord

`required="true"` not working for Rating.

### Reproducer

```xml
<p:rating id="rating4" value="#{ratingView.rating4}" required="true" requiredMessage="Please rate us!" />
<p:commandButton value="Rate" action="#{ratingView.onSubmit}" update="messages" process="@this rating4" />
```

### Expected behavior

Required message is displayed

### PrimeFaces edition

Community

### PrimeFaces version

14.0.0

### Theme

ALL

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

11

### Browser(s)

_No response_

melloware:
Fix #12024: Rating respect required=true


melloware:
@tandraschko even though `setSubmittedValue(null)` is set and `required="true"` the `isValid()` is always true so this check is necessary i looked at PickList and SelectOne but I can't explain why this is `isValid() == true`.

melloware:
@Rapster helped me figure it out `submittedValue = Constants.EMPTY_STRING` instead of `null`

Rapster (approved):

jepsar (approved):

jepsar:
Yeah, this is working perfectly!

jepsar:
Wait, there is nu ui-state-error!

jepsar (changes requested):
There is no ui-state-error CSS class on errors, and additionally there is no styling for it of that class is added...

In my project I added:
```css
body .ui-rating.ui-state-error .ui-rating-star a {
  color: ...;
}
```

Not ideal though

Screenshot 2024-05-30 at 12 39 11

Screenshot 2024-05-30 at 12 40 08

melloware:
I noticed that too.  Let me try some things.

melloware:
@jepsar i am adding `ui-state-error` and I tried this...

```css
.ui-rating.ui-state-error {
  border: 1px solid red;
}
```

image

if that is too harsh we could do `dotted` for more subtle?

image

But I still think I like yours the best? Thoughts?

jepsar:
Yeah, I think it's the best option as well. We can always improve later. For example, when a star is clicked we could remove the error class from JS, but I'm not sure if we do this on any other component (we shouldn't break consistency).

melloware:
OK i will add that style for now. and we can always improve it.

tandraschko:
I would like to see the border of the stars in red 

Removing with js is not consistent, we have CSV for it that

melloware:
@tandraschko so you prefer how @jepsar screenshot is?

tandraschko:
Yeah
At least the star border, i would try how it looks with normal content colour

melloware:
ok let me try that

melloware:
ok so the start icon is a font icon so the only thing we can do is color the STAR whether its filled or not it gets the whole color of  `font-color`.

tandraschko:
I See
But still better as a Border around everything

melloware:
image

tandraschko:
+1

PR #12036, opened 2024-05-31
Fix #12035 - Columns: sortFunction fails 

### LINKED ISSUE #12035: Columns: sortFunction fails
### Describe the bug

when you use a datatable with dynamic columns `<p:columns `, the parameter sortFunction fails to find the method when you click on the column header .
The ColumnsHandler does not declare the SortMeta that was used since primefaces 14.0.0 but the parameter is added when calling the function.
I'm providing a Pull request from my fork to the master. 
https://github.com/primefaces/primefaces/pull/12036
 I tested locally that it fixes the issue shown in the reproducer I'm adding to this page

### Reproducer

here is a reproducer 
[primefaces-test_jakarta_jetty11.zip](https://github.com/user-attachments/files/15511781/primefaces-test_jakarta_jetty11.zip)
when you launch `mvn clean jetty:run -Pmyfaces40` and open `http://localhost:8080/`
and then click on a columns header to sort, it fails with this error : 
`Caused by: jakarta.el.MethodNotFoundException: Method not found: TestView(string=Welcome to PrimeFaces!!!, integer=null, decimal=null, localDateTime=null, list=[TestObject(id=ad99638d-4b6b-4ed3-909d-614c0beaae41, name=Thriller, artist=Michael Jackson, released=1982), TestObject(id=93e63110-c785-4d61-a864-19abf967da8b, name=Back in Black, artist=AC/DC, released=1980), TestObject(id=3490c722-8bad-4f06-9af7-e0f165ac2541, name=The Bodyguard, artist=Whitney Houston, released=1992), TestObject(id=1fb9ab48-ad24-476a-8a4d-6836fe4a14e1, name=The Dark Side of the Moon, artist=Pink Floyd, released=1973)], columns=[TestView.TestObjectModel(header=the Name, property=name), TestView.TestObjectModel(header=the Artist, property=artist), TestView.TestObjectModel(header=the released date, property=released)]).mySortFunction(java.lang.Object, java.lang.Object)`

Notice I didn't use the last master-SNAPSHOT of primefaces-test
I built this zip with the jakarta branch of primefaces-test, in which I reverted a commit to switch from jetty 11 to jetty 12.
If you don't trust me, change the pom yourself to fit jetty 11 and override only these files 😃  :
- src/main/java/org/primefaces/test/TestView.java
- src/main/webapp/test.xhtml

here is why : There is an other issue in the primefaces-test : contrary to the readme.md, the jetty version of the jakarta branch is not 11 but 12.
We need the jetty 11 or else we can't test with java 11, which is still supported by primefaces 14.0.0

### Expected behavior

when you click on the header of a dynamic column `<p:columns/>` inside a dataTable, the sortFunction method is called or sort the lines.

### PrimeFaces edition

Community

### PrimeFaces version

14.0.0

### Theme

dataTable sortFunction

### JSF implementation

MyFaces

### JSF version

3.0.0

### Java version

11

### Browser(s)

firefox

Del-Tab:
Fix #12035

the dynamic columns sortFunction is loaded the same as the sortFunction from the non-dynamic column, (ie with an additional SortMeta parameter) but the el description wasn't updated to fit this requirement. Here is the fix.


Del-Tab:
here is the issue declared on the original primefaces repository
https://github.com/primefaces/primefaces/issues/12035

tandraschko:
Please provide a clean PR, it contains much more commits

Rapster:
I updated the branch, please confirm nothing is missing

tandraschko:
Looks good to me
Can we make a test for it? Cc @melloware

Del-Tab:
> I updated the branch, please confirm nothing is missing

nothing is missing in my fix, it's tiny, only the declaration of the additional SortMeta.class as parameter and the associated import :) 
Sorry about the mess It's my first PR using a forked  github repository

melloware (approved):

melloware:
LGTM to me to. Looks like DataTable042 tests the current Sort Function for regular `p:column` i should be able to modify a test for `p:columns`

PR #12042, opened 2024-05-31
Update primefaces.taglib.xml -> fix description

Del-Tab:
the description was previously changed from `values are new (default), add none.` to `values are new (default), and none.` but add is a valid value

### DIFF
--- a/docs/14_0_0/components/datatable.md
+++ b/docs/14_0_0/components/datatable.md
@@ -91,7 +91,7 @@ DataTable displays data in tabular format.
 | scrollable                | false              | Boolean          | Makes data scrollable with fixed header.
 | selectAllFilteredOnly     | false              | Boolean          | When enabled, toggle select will only apply on filtered items. Default is false.
 | selection                 | null               | Object           | Reference to the selection data.
-| selectionRowMode          | new                | String           | Defines row selection mode. Valid values are "new", "add" and "none".
+| selectionRowMode          | new                | String           | Defines row select mode for checkbox based selection, valid values are 'new' (default, sets the line as the only selected line), 'add' (toggles to selection) and 'none' (the selection is untouched).
 | selectionMode             | null               | String           | Enables row selection, valid values are "single" and "multiple". Automatically detected based on value-binding to `selection` property. So no need to set this explicit.
 | selectionDisabled         | false              | Boolean          | Disables row selection when true. Example: var="xxx" selectionDisabled="#{xxx.year > 1960}"
 | selectionTextDisabled     | true               | Boolean          | Disables text selection on row click.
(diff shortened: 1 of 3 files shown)


primefaces/src/main/resources/META-INF/primefaces.taglib.xml
```diff
@@ -8833,7 +8833,8 @@
         </attribute>
         <attribute>
             <description>
-                <![CDATA[Defines row select mode for checkbox based selection, valid values are new (default), and none.]]>
+                <![CDATA[Defines row select mode for checkbox based selection, valid values are new (default, sets the line
```
melloware, line 8836:
```suggestion
                <![CDATA[Defines row select mode for checkbox based selection, valid values are 'new' (default, sets the line
```

primefaces/src/main/resources/META-INF/primefaces.taglib.xml
```diff
@@ -8833,7 +8833,8 @@
...
+                as the only selected line), add (adds to selection) and none (the selection is untouched).]]>
```
melloware, line 8837:
```suggestion
                as the only selected line), 'add' (adds to selection) and 'none' (the selection is untouched).]]>
```

melloware (changes requested):

primefaces/src/main/resources/META-INF/primefaces.taglib.xml
Del-Tab: thanks, also 'add' toggles the line, it doesn't 'add'

melloware (approved):

melloware:
Assigned to @Rapster so he can verify your descriptions.

docs/14_0_0/components/datatable.md
```diff
@@ -91,7 +91,7 @@ DataTable displays data in tabular format.
 | scrollable                | false              | Boolean          | Makes data scrollable with fixed header.
 | selectAllFilteredOnly     | false              | Boolean          | When enabled, toggle select will only apply on filtered items. Default is false.
 | selection                 | null               | Object           | Reference to the selection data.
-| selectionRowMode          | new                | String           | Defines row selection mode. Valid values are "new", "add" and "none".
+| selectionRowMode          | new                | String           | Defines row select mode for checkbox based selection, valid values are 'new' (default, sets the line as the only selected line), 'add' (toggles to selection) and 'none' (the selection is untouched).
```
Rapster: I don't remember well how this option works... what do you mean by it 'toggles' the selection? If that's what it is, why did we call it 'add' ?

Rapster: That's what I see from the js doc:

```
* @typedef {"new" | "add" | "none"} PrimeFaces.widget.DataTable.SelectionRowMode Indicates how rows of a DataTable
 * may be selected, when clicking on the row itself (not the checkbox / radiobutton from p:column).
 * `new` always unselects other rows, `add` preserves the currently selected rows, and `none` disables row selection.
``` 

I like this description to be honest

Rapster (changes requested):
update description with js doc

melloware:
I fixed the PR with the JS doc descriptions

PR #12047, opened 2024-05-31
Fix #12045: SameSite=Lax default for JSF 4.0

### LINKED ISSUE #12045: monitorDownload: not working in Chrome with ajax=false and non-secure Cookies, due to SameSite=None
### Describe the bug

Since PrimeFaces 14, the `PrimeFaces.monitorDownload()` function does not complete after the download, if the download is performed without AJAX and without secure Cookies.

Apparently the sending of Cookies has changed in the following PR: #11774
Now, the Cookie property "SameSite=None" is always set, when a Cookie is NOT secure (and JSF >= 4.0 is used):
```
        if (isSecure) {
            properties.put("secure", true);
            if (isJsf40OrHigher) {
                properties.put("SameSite", applicationContext.getConfig().getCookiesSameSite());
            }
        }
        else if (isJsf40OrHigher) {
            properties.put("SameSite", "None");
        }
```

This is however against the Spec (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#none), which states:

> SameSite=None
> Means that the browser sends the cookie with both cross-site and same-site requests. 
> **The Secure attribute must also be set when setting this value, like so SameSite=None; Secure. If Secure is missing an error will be logged**:
>
> Cookie "myCookie" rejected because it has the "SameSite=None" attribute but is missing the "secure" attribute.
> 
> This Set-Cookie was blocked because it had the "SameSite=None" attribute but did not have the "Secure" attribute, which is 
> required in order to use "SameSite=None".

In my current Firefox Version, the Cookie is still accepted, but there is a warning printed in the console (the same warning, which was originally reported in #11774 and should be mitigated by the change).

In Chrome, the cookie is rejected alltogether due to this "SameSite=None" attribute, which breaks at least the  `PrimeFaces.monitorDownload()` function, and perhaps any other Cookie-related functionality.

Please consider removing the "SameSite=None" attribute again for Non-Secure Cookies, or setting it to "Lax" instead (which seems to be the default behaviour, when the attribute is not provided, according to the documentation).

### Reproducer

See reproducer: https://github.com/metaxmx/primefaces-test/tree/jakarta (Commit 9dadf0e92acaaef795310d3972927da329a6dd23)

```
<script type="text/javascript">
    function showStatus() {
        PF('statusDialog').show();
    }
    function hideStatus() {
        PF('statusDialog').hide();
    }
</script>

<p:dialog modal="true" widgetVar="statusDialog" header="Status" draggable="false" closable="false">
   Download Running. Please Wait ...
</p:dialog>

<p:commandButton value="Download" onclick="PrimeFaces.monitorDownload(showStatus, hideStatus)" ajax="false">
    <p:fileDownload value="#{fileBean.file}" contentDisposition="attachment"/>
</p:commandButton>
```

Step to reproduce:
1. run the example locally (with Mojarra 4.0 profile)
2. open the page with the newest Chrome (stable) version
3. click the download button
4. the Status-Dialog opens after the download started
6. when the download finished, **the dialog does NOT close again**

### Expected behavior

The status dialog should close, after the download finished

### PrimeFaces edition

Elite

### PrimeFaces version

14.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0

### Java version

21

### Browser(s)

Chrome 125.0.6422.142 (Windows 11)

melloware:
Fix #12045: SameSite=Lax default for JSF 4.0

### DIFF
--- a/primefaces/src/main/java/org/primefaces/util/ResourceUtils.java
+++ b/primefaces/src/main/java/org/primefaces/util/ResourceUtils.java
@@ -23,18 +23,6 @@
  */
 package org.primefaces.util;
 
-import org.primefaces.context.PrimeApplicationContext;
-import org.primefaces.context.PrimeRequestContext;
-
-import javax.el.ValueExpression;
-import javax.faces.FacesException;
-import javax.faces.application.Application;
-import javax.faces.application.Resource;
-import javax.faces.application.ResourceHandler;
-import javax.faces.component.UIComponent;
-import javax.faces.component.UIOutput;
-import javax.faces.context.ExternalContext;
-import javax.faces.context.FacesContext;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
@@ -46,6 +34,19 @@ import java.util.UUID;
 import java.util.function.Consumer;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+
+import javax.el.ValueExpression;
+import javax.faces.FacesException;
+import javax.faces.application.Application;
+import javax.faces.application.Resource;
+import javax.faces.application.ResourceHandler;
+import javax.faces.component.UIComponent;
+import javax.faces.component.UIOutput;
+import javax.faces.context.ExternalContext;
+import javax.faces.context.FacesContext;
+
+import org.primefaces.context.PrimeApplicationContext;
+import org.primefaces.context.PrimeRequestContext;
 
 public class ResourceUtils {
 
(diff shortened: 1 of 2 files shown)


tandraschko:
I dont understand it anyway TBH

Shouldnt we always use applicationContext.getConfig().getCookiesSameSite()?

Just to trigger a discussion here ;)

melloware:
Its a good question so we only use that property for when you are using "secure" to secure your cookies but usually localhost you wouldn't want "Strict" or whatever value they set in `applicationContext.getConfig().getCookiesSameSite()` or if they had no value set it would still trigger that browser warning about a cookie with NO samesite value.

So really all we are checking is in non-secure environments we default it to "Lax" rather than the SameSite value be missing entirely and triggering the browser warning about this being removed in the future.

melloware:
But I could be convinced otherwise @metaxmx ?

metaxmx:
Of course, setting SameSite to either Strict or Lax should be a valid choice in most scenarios.

Only setting it to None is a problem (as I lined out in #12045), and also leaving it out completely could lead to problems, once browsers start enforcing it.
So, as long as there is a valid default value, I think it is also fine to use the configured COOKIES_SAME_SITE value, as long as it is not "None" ...

melloware:
OK new PR is much cleaner and will use `applicationContext.getConfig().getCookiesSameSite()` but if it is NOT secure and "None" will switch to "Lax" or if its blank it will be "Lax"

metaxmx:
Ich checked your updated PR again, in Chrome and Firefox, and it (still) works.
If no configuration for COOKIES_SAME_SITE is given, a value of "Strict" (Default) is used, and if I manually set it to "None" (and use no Secure connection), it reverts to "Lax".

melloware:
Hmmm but that does have me wondering whether when running locally the default of "Strict" makes sense.

PR #12055, opened 2024-06-02
Fix #12054: Taglib don't allow VAR as EL expression

### LINKED ISSUE #12054: Taglib: Remove type from `var` fields
Based on this Faces issue: https://github.com/jakartaee/faces/issues/1906

Mojarra PR: https://github.com/eclipse-ee4j/mojarra/pull/5448

Should we do the same for PrimeFaces taglib?

melloware:
Fix #12054: Taglib don't allow VAR as EL expression

Draft PR for review

### DIFF
--- a/primefaces/src/main/resources/META-INF/primefaces.taglib.xml
+++ b/primefaces/src/main/resources/META-INF/primefaces.taglib.xml
@@ -643,7 +643,6 @@
             </description>
             <name>var</name>
             <required>false</required>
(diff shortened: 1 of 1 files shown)


tandraschko:
i dont understand
what happens in JSF internal when you remove the type? please check with MF

melloware:
basically VAR is not a ValueExpressions but it shows up as one in VDL like this

image

PR #12092, opened 2024-06-08
#12085 Core: Ajax performance (cleanseDomElement): No need to cleanse options elements

### LINKED ISSUE #12085: Core: Ajax performance (cleanseDomElement)
### Description

Since #11722 I can see a noticeable performance degradation in client side when there is an ajax and there are many inputs or DOM elements.

I can trace the problem down to the last remove in this function:

```
        cleanseDomElement: function(jq, clearData = true, removeElement = true) {
            if (!jq || !jq.length) {
                return;
            }
            // Recursively remove events from children elements
            jq.children().each(function() {
                PrimeFaces.utils.cleanseDomElement($(this), clearData, removeElement);
            });

            // Remove inline event attributes
            var attributes = jq[0].attributes;
            for (var i = 0; i < attributes.length; i++) {
                var attributeName = attributes[i].name;
                if (attributeName.startsWith("on")) {
                    jq.removeAttr(attributeName);
                }
            }
            // Remove event listeners
            jq.off();
            
            // Clear data
            if (clearData) {
                jq.removeData();
            }
            // Remove elements itself from memory
            if (removeElement) {
                jq.remove();
            }
        },
```

If I replace the last 
       `jq.remove();`
with
      `  jq.each(function() {this.remove()});`

I can see that performance improves very much.  In my test I have an ajax that takes about 1.3seconds scripting and with the change it only takes 180ms.

Looks like a safe change, since children are removed before. 
Do you think it's worthy to integrate this in PF?

CleanseDomElement should be faster. Of course another solution is to shrink the DOM tree to update, but there is room to improve cleanseDomElement

### Additional context

_No response_

cocorossello:
Fix #12085 In my test, with the performance measurement script provided in https://github.com/primefaces/primefaces/issues/12085

Without the patch:

```
AJAX updateElement : 104.026123046875 ms
common.js:751 Removed 269 DOM elements
common.js:750 AJAX updateElement : 1.4541015625 ms
common.js:751 Removed 1 DOM elements
common.js:750 AJAX updateElement : 1908.331787109375 ms
common.js:751 Removed 33193 DOM elements
```
With the patch:

```
AJAX updateElement : 99.87109375 ms
common.js:41 Removed 22 DOM elements
common.js:40 AJAX updateElement : 2.444091796875 ms
common.js:41 Removed 1 DOM elements
common.js:40 AJAX updateElement : 390.40087890625 ms
common.js:41 Removed 2413 DOM elements
```
Numbers are still way worse than the ones @melloware posted in https://github.com/primefaces/primefaces/issues/12085#issuecomment-2155018334, but this is an improvement

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
@@ -1122,7 +1122,7 @@ if (!PrimeFaces.utils) {
                 return;
             }
 
-            if (jq.prop("tagName") !== 'SELECT') { //No need to cleanse <option>s tag
+            if (!jq.is("select")) { //No need to cleanse <option> tags
                 // Recursively remove events from children elements
                 jq.children().each(function() {
                     PrimeFaces.utils.cleanseDomElement($(this), clearData, removeElement);
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
@@ -1121,11 +1121,13 @@ if (!PrimeFaces.utils) {
             if (!jq || !jq.length) {
                 return;
             }
-            // Recursively remove events from children elements
-            jq.children().each(function() {
-                PrimeFaces.utils.cleanseDomElement($(this), clearData, removeElement);
-            });
 
+            if (jq.prop("tagName") !== 'SELECT') { //No need to cleanse <option>s tag
+                // Recursively remove events from children elements
+                jq.children().each(function() {
+                    PrimeFaces.utils.cleanseDomElement($(this), clearData, removeElement);
+                });
+            }
             // Remove inline event attributes
             var attributes = jq[0].attributes;
             for (var i = 0; i < attributes.length; i++) {
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
@@ -1122,7 +1122,8 @@ if (!PrimeFaces.utils) {
                 return;
             }
 
-            if (!jq.is("select")) { //No need to cleanse <option> tags
+            //Skip cleanse of select and svg tags, it can impact performance if a lot of tags are present. They don't have PF listeners attached, so cleanse it's unnecesary.
+            if (!jq.is("select, svg, svg *")) {
                 // Recursively remove events from children elements
                 jq.children().each(function() {
                     PrimeFaces.utils.cleanseDomElement($(this), clearData, removeElement);
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
@@ -1138,7 +1138,7 @@ if (!PrimeFaces.utils) {
                 }
             }
 
-            // Trigger onRemove events for widget.destroy and remove the element from the DOM and
+            // Trigger onRemove events for widget.destroy and remove the element from the DOM
             // IMPORTANT: This must occur before jq.off() to ensure the on("remove") events remain registered.
             if (removeElement) {
                 jq.triggerHandler("remove");
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
@@ -1140,7 +1140,8 @@ if (!PrimeFaces.utils) {
             // Remove the element from the DOM and trigger onRemove events for widget.destroy.
             // IMPORTANT: This must occur before jq.off() to ensure the on("remove") events remain registered.
             if (removeElement) {
-                jq.remove();
+                jq.triggerHandler("remove");
+                jq.get(0).remove()
             }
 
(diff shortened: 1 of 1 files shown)


cocorossello:
I can see that still with the patch the performance is related to options size. If I don't render the options is much faster (scripting time)

I think it is because the event bubbles up (maybe?). In any case, if you need the remove event it can be fired with:

```
    // Remove the element from the DOM and trigger onRemove events for widget.destroy.
    // IMPORTANT: This must occur before jq.off() to ensure the on("remove") events remain registered.
    if (removeElement) {
        jq.triggerHandler("remove");
        jq.get(0).remove()
    }
```

This way performance numbers are good:

```
AJAX updateElement : 100.050048828125 ms
common.js:41 Removed 22 DOM elements
common.js:40 AJAX updateElement : 1.617919921875 ms
common.js:41 Removed 1 DOM elements
common.js:40 AJAX updateElement : 44.89892578125 ms
common.js:41 Removed 2808 DOM elements
```
Can you test if this is enough for primefaces to cleanup?

melloware:
@cocorossello yes let me test this in my real world scenarios

melloware:
OK i did the same behavior in the app with current code and your patch.

```
BEFORE CHANGE:
VM43:121 AJAX updateElement: 75.097900390625 ms
VM43:122 Removed 138 DOM elements
VM43:121 AJAX updateElement: 211.8740234375 ms
VM43:122 Removed 199 DOM elements
VM43:121 AJAX updateElement: 302.350830078125 ms
VM43:122 Removed 11642 DOM elements
VM43:121 AJAX updateElement: 259.909912109375 ms
VM43:122 Removed 12050 DOM elements
VM43:553 TabView: 4.22998046875 ms
VM43:554 Removed 247 DOM elements
VM43:121 AJAX updateElement: 234.94091796875 ms
VM43:122 Removed 12020 DOM elements
VM43:553 TabView: 3.346923828125 ms
VM43:554 Removed 424 DOM elements
VM43:121 AJAX updateElement: 223.243896484375 ms
VM43:122 Removed 11596 DOM elements
VM43:553 TabView: 53.662109375 ms
VM43:554 Removed 11468 DOM elements
VM43:121 AJAX updateElement: 26.52294921875 ms
VM43:122 Removed 128 DOM elements


AFTER CHANGE:
VM43:122 AJAX updateElement: 73.114013671875 ms
VM43:123 Removed 138 DOM elements
VM43:122 AJAX updateElement: 212.7138671875 ms
VM43:123 Removed 199 DOM elements
VM43:122 AJAX updateElement: 281.281005859375 ms
VM43:123 Removed 11642 DOM elements
VM43:122 AJAX updateElement: 257.079833984375 ms
VM43:123 Removed 12050 DOM elements
VM43:554 TabView: 4.556884765625 ms
VM43:555 Removed 247 DOM elements
VM43:122 AJAX updateElement: 248.955078125 ms
VM43:123 Removed 12020 DOM elements
VM43:554 TabView: 6.14501953125 ms
VM43:555 Removed 424 DOM elements
VM43:122 AJAX updateElement: 211.81298828125 ms
VM43:123 Removed 11596 DOM elements
VM43:554 TabView: 54.3701171875 ms
VM43:555 Removed 11468 DOM elements
VM43:122 AJAX updateElement: 20.333984375 ms
VM43:123 Removed 128 DOM elements
```

in my testing it seems your change performs about the same but i can confirm all the memory is still being cleaned up.

cocorossello:
Ok, I updated the PR with this findings. It still makes sense to skip the removal of select's children. (44ms skipping vs 130 without skipping)

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
```diff
@@ -1121,11 +1121,13 @@ if (!PrimeFaces.utils) {
             if (!jq || !jq.length) {
                 return;
             }
-            // Recursively remove events from children elements
-            jq.children().each(function() {
-                PrimeFaces.utils.cleanseDomElement($(this), clearData, removeElement);
-            });
 
+            if (jq.prop("tagName") !== 'SELECT') { //No need to cleanse <option>s tag
```
melloware:
```suggestion
            if (!jq.is("select")) { //No need to cleanse <option>s tag
```

melloware (changes requested):

tandraschko:
looks good to me
are the other childs to skip?

melloware:
I would think head would only be there if update="@all" but i have never used "@all" so i am not sure if head is in there.

melloware (approved):

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
```diff
@@ -1121,11 +1121,13 @@ if (!PrimeFaces.utils) {
             if (!jq || !jq.length) {
                 return;
             }
-            // Recursively remove events from children elements
-            jq.children().each(function() {
-                PrimeFaces.utils.cleanseDomElement($(this), clearData, removeElement);
-            });
 
+            if (!jq.is("select")) { //No need to cleanse <option> tags
```
melloware:
```suggestion
            if (!jq.is("select, svg")) { //No need to cleanse SVG or <option> tags
```
```suggestion
            if (!jq.is("select")) { //No need to cleanse <option> tags
```

melloware (changes requested):

tandraschko:
+1 to merge when you added svg exclude

melloware (approved):

PR #12109, opened 2024-06-13
Fix #12099: p:columns rendered in p:columnGroup not fully working

### LINKED ISSUE #12099: DataTable: Handling of Columns in column group differs from body
### Describe the bug

```xhtml
<p:dataTable>
    <p:columnGroup type="header">
        <p:row>
            <p:columns value="#{myBean.columnList}" var="c" rendered="#{c ne null}" headerText="#{c.name}" />
        </p:row>
    </p:columnGroup>


    <p:columns value="#{myBean.columnList}" var="c" rendered="#{c ne null}">
        <h:outputText value="#{c.name}" />
    </p:columns>
</p:dataTable>
```

The above code snippet demonstrates, that `p:columns#rendered` does not work inside the `p:columnGroup` but it does work in the header. Reason is the specialized logic to render the column group which first checks, if the [`p:row`'s children is renderable](https://github.com/primefaces/primefaces/blob/8a659235f064a27b306888c1c295d248ff36a377/primefaces/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java#L957C29-L957C63) and then applies the state to `p:columns`. While the body first applies the state and then [checks for renderable](https://github.com/primefaces/primefaces/blob/8a659235f064a27b306888c1c295d248ff36a377/primefaces/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java#L1231).

This should be aligned to be the same logic - first apply, then check.

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

Community

### PrimeFaces version

11.0.0, 15.0.0-SNAPSHOT (master)

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.2.19

### Java version

8u3xx

### Browser(s)

_No response_

fanste:
Fix #12099 `p:columns#var` was not available in `p:columns#rendered`

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
@@ -954,20 +954,22 @@ public class DataTableRenderer extends DataRenderer {
                         }
 
                         for (UIComponent headerRowChild : headerRow.getChildren()) {
-                            if (headerRowChild.isRendered()) {
-                                if (headerRowChild instanceof Column) {
+                            if (headerRowChild instanceof Column) {
+                                if (headerRowChild.isRendered()) {
                                     encodeColumnHeader(context, table, (Column) headerRowChild);
                                 }
-                                else if (headerRowChild instanceof Columns) {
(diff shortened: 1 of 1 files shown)


melloware (approved):

Rapster (approved):

tandraschko:
+1

PR #12112, opened 2024-06-13
Fix #12098: Tooltip autoHide="false" hidden when not crossing arrow

### LINKED ISSUE #12098: Tooltip: autoHide="false" does not work
### Describe the bug

Hello all.

Tooltip's autoHide="false" is not working. If I move the mouse over the tooltip's content, it disappears whether autoHide is set as "true" or "false". Last version I remember it worked was 13.

This was tested on Edge, Chrome and Firefox.

### Reproducer

This happens even in the [showcase](https://www.primefaces.org/showcase/ui/overlay/tooltip/tooltipOptions.xhtml).

### Expected behavior

Tooltip should stay on place when hovered on so I can select it's contents.

### PrimeFaces edition

Community

### PrimeFaces version

14.0.0

### Theme

Saga

### JSF implementation

Mojarra

### JSF version

4.0

### Java version

21

### Browser(s)

_No response_

redddcyclone:
Complements #12102 on 
Fix #12098

This fixes the check for allowHide by checking if the previous element's (the one that the mouse is leaving from) "aria-describedby" value is the same as the target element's (tooltip) or any of its parents' id (because if there is any child elements on the tooltip, the event may refer to it as the target instead of the tooltip itself). This includes the "ui-tooltip-arrow" element.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/tooltip/tooltip.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/tooltip/tooltip.js
@@ -301,7 +301,7 @@ PrimeFaces.widget.Tooltip = PrimeFaces.widget.BaseWidget.extend({
         else {
             var mouseTarget = $(e.relatedTarget);
             var previousElement = this.target;
-            this.allowHide = !(mouseTarget.is(this.target) ||
+            this.allowHide = !(mouseTarget.is(previousElement) ||
                                previousElement.attr('aria-describedby') === mouseTarget.closest('#' + this.id).attr('id'));
         }
         if (this.allowHide) {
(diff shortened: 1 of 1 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/tooltip/tooltip.js
```diff
@@ -300,10 +300,9 @@ PrimeFaces.widget.Tooltip = PrimeFaces.widget.BaseWidget.extend({
         }
         else {
             var mouseTarget = $(e.relatedTarget);
-            this.allowHide = !(mouseTarget.is(this.target) || 
-                               mouseTarget.hasClass('ui-tooltip-arrow') ||
-                               mouseTarget.attr('aria-describedby') === this.id || 
-                               mouseTarget.parent().attr('aria-describedby') === this.id);
+            var previousElement = this.target;
+            this.allowHide = !(mouseTarget.is(this.target) ||
```
melloware:
```suggestion
            this.allowHide = !(mouseTarget.is(previousElement) ||
```
```suggestion
            this.allowHide = !(mouseTarget.is(this.target) ||
```

melloware (changes requested):

melloware (approved):

PR #12119, opened 2024-06-14
Fix #12118: Tooltip: escape selector for mouseTarget.closest

### LINKED ISSUE #12118: Tooltip: autoHide="false" never gets hidden
### Describe the bug

Hello

I detected a new bug on 14.0.1 related to issue #12098 and its corresponding fix #12112. The tooltip gets stuck if the component's or any of its parents' ID have a colon or any other characters that should be escaped in a selector.

I'll be submitting a PR soon for this.

### Reproducer

1. Have a tooltip with autoHide="false" set
2. Show the tooltip.
3. It is never hidden because an exception is caught

### Expected behavior

Tooltip should hide when mouse is moved out of its trigger component and the tooltip itself

### PrimeFaces edition

Community

### PrimeFaces version

14.0.1

### Theme

Saga

### JSF implementation

Mojarra

### JSF version

4.0

### Java version

21

### Browser(s)

_No response_

redddcyclone:
Fix #12118

This PR fixes an issue that happens when the tooltip's ID has characters that should be escaped, like the colon, something that happens frequently on Faces.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/tooltip/tooltip.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/tooltip/tooltip.js
@@ -302,7 +302,7 @@ PrimeFaces.widget.Tooltip = PrimeFaces.widget.BaseWidget.extend({
             var mouseTarget = $(e.relatedTarget);
             var previousElement = this.target;
             this.allowHide = !(mouseTarget.is(previousElement) ||
-                               previousElement.attr('aria-describedby') === mouseTarget.closest('#' + this.id).attr('id'));
+                               previousElement.attr('aria-describedby') === mouseTarget.closest('#' + $.escapeSelector(this.id)).attr('id'));
         }
         if (this.allowHide) {
             this.hide();
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/tooltip/tooltip.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/tooltip/tooltip.js
@@ -302,7 +302,9 @@ PrimeFaces.widget.Tooltip = PrimeFaces.widget.BaseWidget.extend({
             var mouseTarget = $(e.relatedTarget);
             var previousElement = this.target;
             this.allowHide = !(mouseTarget.is(previousElement) ||
-                               previousElement.attr('aria-describedby') === mouseTarget.closest('#' + $.escapeSelector(this.id)).attr('id'));
+                               previousElement.attr('aria-describedby') === mouseTarget.closest(PrimeFaces.escapeClientId(this.id)).attr('id') ||
+                               mouseTarget.attr('aria-describedby') === this.id ||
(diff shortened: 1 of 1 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/tooltip/tooltip.js
```diff
@@ -302,7 +302,7 @@ PrimeFaces.widget.Tooltip = PrimeFaces.widget.BaseWidget.extend({
             var mouseTarget = $(e.relatedTarget);
             var previousElement = this.target;
             this.allowHide = !(mouseTarget.is(previousElement) ||
-                               previousElement.attr('aria-describedby') === mouseTarget.closest('#' + this.id).attr('id'));
+                               previousElement.attr('aria-describedby') === mouseTarget.closest('#' + $.escapeSelector(this.id)).attr('id'));
```
melloware:
```suggestion
                               previousElement.attr('aria-describedby') === mouseTarget.closest(PrimeFaces.escapeClientId((this.id)).attr('id'));
```

melloware (changes requested):

melloware:
I found another issue.  The original code had it where if you were inside the tooltip and moved the mouse back into the calling button the tooltip would stay open.  I think this is why it had the code for 

```
mouseTarget.attr('aria-describedby') === this.id || 
                               mouseTarget.parent().attr('aria-describedby') === this.id)
```

basically if navigating back into the tooltips owner don't close the tooltip

primefaces/src/main/resources/META-INF/resources/primefaces/tooltip/tooltip.js
redddcyclone: I don't know how ```PrimeFaces.escapeClientId``` works, but wouldn't it remove the required # character?

EDIT: Never mind, I tried it and it works correctly.

melloware: No it adds the `#` and escapes the `:`

```js
/**
         * Creates an ID to a CSS ID selector that matches elements with that ID. For example:
         * ```
         * PrimeFaces.escapeClientId("form:input"); // => "#form\:input"
         * PrimeFaces.escapeClientId("form#input"); // => "#form#input"
         * ```
         *
         * __Please note that this method does not escape all characters that need to be escaped and will not work with arbitrary IDs__
         * @param {string} id ID to convert.
         * @return {string} A CSS ID selector for the given ID.
         */
        escapeClientId : function(id) {
            return "#" + id.replace(/:/g,"\\:");
        },
```

redddcyclone:
@melloware you're right, I didn't notice this because I use tooltips linked to text, and this seem to happen with buttons (and other components maybe). I restored this functionality on the last commit and replaced ```$.escapeSelector``` with ```PrimeFaces.escapeClientId```

melloware (approved):

melloware:
Thanks for the PR!

PR #12134, opened 2024-06-17
Fix #12130: CommandLink restore TabIndex properly

### LINKED ISSUE #12130: CommandLink: Tabindex gets removed
### Describe the bug

If `tabindex` is defined on the CommandLink component, the tabindex gets removed when clicking on the link.

Reproducer: [primefaces-test-jakarta-m6ppdbjay7rrtxktxnunxemhoh.zip](https://github.com/user-attachments/files/15866014/primefaces-test-jakarta-m6ppdbjay7rrtxktxnunxemhoh.zip)

### Reproducer

1. Go to http://localhost:8080/test.xhtml
2. Click on the link "Link Tabindex 5"
3. Tabindex is not set anymore

### Expected behavior

If a tabindex is defined, the tabindex should be preserved and not removed.

### PrimeFaces edition

Community

### PrimeFaces version

14.0.1

### Theme

_No response_

### JSF implementation

MyFaces

### JSF version

4.0

### Java version

21

### Browser(s)

_No response_

melloware:
Fix #12130: CommandLink restore TabIndex properly

cc @pfumi

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/commandlink/CommandLinkRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/commandlink/CommandLinkRenderer.java
@@ -26,6 +26,7 @@ package org.primefaces.component.commandlink;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+
 import javax.faces.FacesException;
 import javax.faces.component.UIComponent;
 import javax.faces.component.UIForm;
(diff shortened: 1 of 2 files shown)


primefaces/src/main/java/org/primefaces/component/commandlink/CommandLinkRenderer.java
```diff
@@ -139,6 +137,10 @@ protected void encodeMarkup(FacesContext context, CommandLink link) throws IOExc
             renderPassThruAttributes(context, link, HTML.LINK_ATTRS);
         }
 
+        if (link.isDisabled()) {
+            writer.writeAttribute("tabindex", "-1", null);
+        }
+
```
melloware, line 143: I moved this down because this needs to come AFTER `renderPassThroughAttributes` as the last `tabIndex` will win

PR #12135, opened 2024-06-17
Fix #12131: TextEditor only fire change on API insert

### LINKED ISSUE #12131: TextEditor doesn't detect change of contents
Reported by PRO User;
> The goal : If a machine.domain.tld is inserted into the TextEditor without a "https://" prefix, automatically insert one.
After having used the Quill editor API to modify a Delta containing the text, the <a> tag's href shows the updated value. However, if the page is saved, the valueChangeListener can be used to view the TextEditor's before and after values. They are the same.
The attached .xhtml has a JS function replaceLink( url ) the updates the editor's Deltas.
Set a breakpoint in the attached controller's onHTMLChanged( ValueChangeEvent vce ) method. ...

melloware:
Fix #12131: TextEditor only fire change on API insert


tandraschko:
+1 to merge it
currently ITs are broken

PR #12142, opened 2024-06-19
Fix #12138: Taglib only include URN version in Jakarta JAR

### LINKED ISSUE #12138: JavaEE JAR contains JakartaEE URN taglib
i think it should only be in the jakarta shade

melloware:
Fix #12138: Taglib only include URN version in Jakarta JAR

@tandraschko this has to strip the file from the JAR post process so I only do it when installing/deploying the JAR.   I can't do it any earlier or else it gets stripped from the Jakarta JAR too.

### DIFF
--- a/primefaces/pom.xml
+++ b/primefaces/pom.xml
@@ -1229,8 +1229,7 @@
                         <phase>process-resources</phase>
                         <configuration>
                             <target>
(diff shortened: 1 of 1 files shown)


tandraschko:
if it works, its looks good for me

PR #12171, opened 2024-06-23
Update to OpenPDF 2.0.2

andreasrosdalw:
https://github.com/LibrePDF/OpenPDF/releases/tag/2.0.2

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [11, 21]
+        java: [17, 21]
     steps:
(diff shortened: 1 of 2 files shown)


melloware:
Looks like the build is failing.

melloware (changes requested):
We can support this as 2.x is JDK 17 only

> Java 17 or later is required for OpenPDF 2.0.0. This will allow using modern Java 17+ compatible syntax and libraries. Futhermore, it will ease maintenance effort because we don't support older Java versions.

tandraschko:
We decided to stay on java11, at least until migrate to Jakarta for 100%

PR #12215, opened 2024-06-28
Fix #12214: RowExpansion rendered fix

### LINKED ISSUE #12214: DataTable: Row Expansion and expandedRow rendered issue
### Describe the bug

I am having an issue with a datatable that uses rowExpansion in PrimeFaces. I have configured expandedRow="true" on the datatable, but it fails when it retrieves records that do not have content to display in rowExpansion, as there is not always something to show.

### Reproducer
[primefaces-test-master-row-expansion.zip](https://github.com/user-attachments/files/16033522/primefaces-test-master-row-expansion.zip)

  - Change expandedRow:
    - If it is true: it causes an exception.
    - If it is false: it works correctly, except that rowExpansion does not expand unless I remove the rendered attribute.
    - In previous versions of PrimeFaces, this worked correctly.

```html
<p:dataTable value="#{testView.list}"
             var="testLine"
             rowKey="#{testLine.id}"
             expandedRow="true"
             lazy="false">
    <f:facet name="header">
        Expand rows to see detailed information
    </f:facet>
    <p:column style="width:2rem" ariaHeaderText="Row Toggler">
        <p:rowToggler rendered="#{testLine.testChild.size() > 0}"/>
    </p:column>

    <p:column headerText="Name">
        <h:outputText value="#{testLine.name}"/>
    </p:column>

    <p:column headerText="Artist">
        <h:outputText value="#{testLine.artist}"/>
    </p:column>

    <p:column headerText="Released">
        <h:outputText value="#{testLine.released}"/>
    </p:column>

    <p:rowExpansion rendered="#{testLine.testChild.size() > 0}">
        <h2>Data Expanded</h2>
        <p:dataTable value="#{testLine.testChild}" var="testChildLine" rowKey="#{testChildLine.id}" lazy="false">
            <p:column headerText="Name">
                <h:outputText value="#{testChildLine.name}"/>
            </p:column>

            <p:column headerText="Artist">
                <h:outputText value="#{testChildLine.artist}"/>
            </p:column>

            <p:column headerText="Released">
                <h:outputText value="#{testChildLine.released}"/>
            </p:column>
        </p:dataTable>
    </p:rowExpansion>
</p:dataTable>

```

### Expected behavior

_No response_

### PrimeFaces edition

Community

### PrimeFaces version

14.0.1

### Theme

Freya

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

21

### Browser(s)

_No response_

melloware:
Fix #12214: RowExpansion rendered fix

- [x] It was determing whether row expansion was enabled or not based on if row expansion rendered but really we just want to know if the table has a `<p:rowExpansion>` for determining that as the `rendered` is evaluated inside the feature
- [x] Switched to use our Style Builder

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java
+++ b/primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java
@@ -28,6 +28,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
+
 import javax.el.ELContext;
 import javax.el.ValueExpression;
 import javax.faces.FacesException;
@@ -548,7 +549,7 @@ public class DataTable extends DataTableBase {
     }
 
     public RowExpansion getRowExpansion() {
-        return ComponentTraversalUtils.firstChildRendered(RowExpansion.class, this);
+        return ComponentTraversalUtils.firstChild(RowExpansion.class, this);
     }
 
     @Override
--- a/primefaces/src/main/java/org/primefaces/component/datatable/feature/RowExpandFeature.java
+++ b/primefaces/src/main/java/org/primefaces/component/datatable/feature/RowExpandFeature.java
@@ -27,6 +27,7 @@ import java.io.IOException;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
+
 import javax.faces.context.FacesContext;
 import javax.faces.context.ResponseWriter;
 
@@ -34,6 +35,7 @@ import org.primefaces.component.datatable.DataTable;
 import org.primefaces.component.datatable.DataTableRenderer;
 import org.primefaces.component.datatable.DataTableState;
 import org.primefaces.component.rowexpansion.RowExpansion;
+import org.primefaces.context.PrimeRequestContext;
 import org.primefaces.util.LangUtils;
 
 public class RowExpandFeature implements DataTableFeature {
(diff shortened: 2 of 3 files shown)


Rapster:
> It was determing whether row expansion was enabled or not

Where is that ? I'm trying to understand how the new code is different from the old one ?

melloware:
OK there were two issues let me go through them both and how the Test Reproducer helped me figure it out.

1. To determine whether row Expansion was being used on the table it was using 

```java
return ComponentTraversalUtils.firstChildRendered(RowExpansion.class, this);
```

That was returning NULL and thus causing the NPE in RowExpandFeature. So in RowExpandFeature i changed it from..

```java
if (rowExpansion.isRendered()) {
```

to

```java
if (rowExpansion != null && rowExpansion.isRendered()) {
```

That led me to problem 2

2. Now in the Renderer the code to tell the widget whether Row Expansion is enabled or not was this.

```java
        if (table.getRowExpansion() != null) {
            wb.attr("expansion", true).attr("rowExpandMode", table.getRowExpandMode());
        }
```

That of course was still NULL so it was not rendering expansion at all.   So that is why the fix from `firstChildRendered` to `firstChild` fixes it. We don't care if the expansion is rendered or not we just care if the table has an Expansion.  The `rendered` is evaluated later in RowFeature.


I wonder if SubTable will have the same issue as it also does `firstChildRendered`

Rapster:
Ok I see, basically whether it's rendered or not, the js widget still need to know if it can potentially be used. Regarding subTable, I don't think subTable is used in js, it's only an util for rendering

Rapster (approved):

melloware:
Yep you got it.

PR #12218, opened 2024-06-29
Sonar JS fixes for Core

### LINKED ISSUE #11863: SONAR: Fix JS issues
### Description

see https://sonarcloud.io/project/issues?languages=js&resolved=false&id=org.primefaces%3Aprimefaces

less issues

### Additional context

_No response_

melloware:
Fix https://github.com/primefaces/primefaces/issues/11863

Addresses the Sonar issues on in the `/core` package only


tandraschko:
Need to think about it, im not that happy with it right now

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.js
```diff
@@ -1201,7 +1193,7 @@
 
             // try and strip specific language from nl_BE to just nl
             if (!locale) {
-                var localeKey = cfgLocale ? cfgLocale : PrimeFaces.settings.locale;
+                var localeKey = cfgLocale || PrimeFaces.settings.locale;
```
Rapster: this localKey var could be renamed since it's a boolean

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
```diff
@@ -451,23 +449,25 @@ if (!PrimeFaces.utils) {
...
+
+                $scrollParent.off(scrollNamespace, scrollHandler).on(scrollNamespace, scrollHandler);
+            });
 
             return {
                 unbind: function() {
-                    for (var i = 0; i < scrollableParents.length; i++) {
-                        $(scrollableParents[i]).off(scrollNamespace);
-                    }
+                    scrollableParents.forEach(scrollParent => {
```
Rapster: why not use for of ?

melloware: Switched to for-of

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.js
melloware: This actually is the value "en_US" either using the local config or global locale value before it gets split to "en".

Rapster: ohh

tandraschko:
what about external JS? are the filtered out somehow?
https://sonarcloud.io/project/issues?impactSoftwareQualities=RELIABILITY&issueStatuses=OPEN%2CCONFIRMED&id=org.primefaces%3Aprimefaces&open=AY8u9KHvLK6QkKtQMXzo

i would like to not scan it

Rapster:
I just added the exclusion (there might be still a few left)

Rapster (approved):

PR #12223, opened 2024-07-01
Fix #12222: Datatable ignore expanded row content when autosizing

### LINKED ISSUE #12222: DataTable: column width autoAdjust feature should ignore expanded Rows
### Description

Currently cell navigation is not active together with row expansion, but In order to provide a more general support for the cellNavigation in the dataTable in long term the feature #11918 should ignore expanded rows. 

Currently it would consider the width of the expanded rows (class= ui-expanded-row-content) in the selector cellsInSameColumn in autosizeColumnWidth() so that the new calculated width is not as expected when there are tables under the expanded rows.

image

autosizeColumnWidth() should ignore expanded row content rows when calculating the new column width for column width auto adjustment.

### Additional context

_No response_

melloware:
Fix #12222: Datatable ignore expanded row content when autosizing


Rene-Stoeckel:
Oh, that was quick. :-) Please note my comment on #12222.

melloware:
Yep made the change!

Rene-Stoeckel:
Looks good.

PR #12252, opened 2024-07-04
Fix #12241: Restore AJAX resetValues behavior

### LINKED ISSUE #12241: CommandButton: Resetting Inputs is broken for resetValues="true"
### Discussed in https://github.com/orgs/primefaces/discussions/2200

<sup>Originally posted by **hangrybear666** June 25, 2024</sup>
In my entire Application, `resetValues="true"` Flags on `<p:commandButton/>` Elements have no effect since upgrading from PrimeFaces 10. 
I have to manually call `PrimeFaces.current().resetInputs(resetInputId);` in my models and pass the IDs I want to update to the model as well to achieve the desired result.

The `resetValues` Flag in the following code has no effect:
```
<p:commandButton global="false" class="generic-btn-cancel"
        value="#{commonRes.common_cancel}" icon="pi pi-undo"
        actionListener="#{exampleModel.cancelEditOrInsert}"
        update="resetInputId"
        immediate="true" resetValues="true"/>
```

@tandraschko I've created a primefaces-test project that is self-explanatory.

**MP4 Screen Capture displaying workflow:**
https://github.com/primefaces/community/assets/85507965/a8045e9c-3993-4a89-9fab-da53c5fb9ab4

**Reproducer Primfaces v14 Test Archive**
[primefaces-test-commandbutton-resetValues.zip](https://github.com/user-attachments/files/16071343/primefaces-test-commandbutton-resetValues.zip)

P.S. ignore the src\main\webapp\effectiveDataTable.xhtml file, it's for another test project I'm making</div>

melloware:
Fix #12241: Restore AJAX resetValues behavior

### DIFF
--- a/primefaces/src/main/java/org/primefaces/context/PrimePartialViewContext.java
+++ b/primefaces/src/main/java/org/primefaces/context/PrimePartialViewContext.java
@@ -32,6 +32,7 @@ import javax.faces.context.PartialViewContextWrapper;
 import javax.faces.event.PhaseId;
 
 import org.primefaces.config.PrimeConfiguration;
+import org.primefaces.config.PrimeEnvironment;
 import org.primefaces.csp.CspPartialResponseWriter;
 import org.primefaces.util.ComponentUtils;
 import org.primefaces.util.Constants;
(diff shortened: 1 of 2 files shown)


tandraschko:
actually not a good fix, we need to wait for the response of the spec ticket if we fix it for old versions or not

resetValues will be invoked 2 times now, one in PF, one party in JSF

melloware:
OK but Mojarra has already stated they aren't fixing the 2.3.X branch so we need to decide if and how to restore the proper functionality to PF 14.

melloware:
Added check to only run this code if < 4.0

tandraschko:
@melloware i will merge this for now
could you please create a issue in mojarra itself, if not already done?

PR #12260, opened 2024-07-05
Fix #12259: DataTable multi-select SHIFT on page > 1

### LINKED ISSUE #12259: Datatable: Multiple selection with SHIFT key does not work past page 1
### Describe the bug

If I have a datatable with multiple pages and multiple selection enabled, I can click a row then hold shift then click another row to select a range of rows.  This only works on page 1 starting with PF 14.0.0.RC1 and tested up to 14.0.2.  It will not work on any page of results beyond 1.

### Reproducer

[primefaces-test.zip](https://github.com/user-attachments/files/16111330/primefaces-test.zip)

### Expected behavior

_No response_

### PrimeFaces edition

Community

### PrimeFaces version

14.0.0.RC1 through 14.0.2

### Theme

_No response_

### JSF implementation

All

### JSF version

2.3 and 4.x (Jakarta)

### Java version

17/21

### Browser(s)

_No response_

melloware:
Fix #12259: DataTable multi-select SHIFT on page > 1

### DIFF
--- a/primefaces-integration-tests/src/main/webapp/datatable/dataTable005Paging.xhtml
+++ b/primefaces-integration-tests/src/main/webapp/datatable/dataTable005Paging.xhtml
@@ -1,0 +1,36 @@
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml"
+      xmlns:h="http://xmlns.jcp.org/jsf/html"
+      xmlns:f="http://xmlns.jcp.org/jsf/core"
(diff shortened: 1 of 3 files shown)


melloware:
Looks like DataTable005 test does but it does NOT use paging so looks like i have to add another test that pages to page 2 and multiselects.

tandraschko:
that would be great

melloware:
IT added

PR #12265, opened 2024-07-07
Fix #12264: XML minification

### LINKED ISSUE #12264: Core: Minify XML files in JAR
Currently our Taglib.xml is quite large and should be whitespace minified in the JAR file.

melloware:
Fix #12264: XML minification

### Before:
faces-config.xml  87 KB
primefaces.taglib.xml   1,352 KB

### After:
faces-config.xml  79 KB
primefaces.taglib.xml   979 KB


tandraschko:
For me its a -0.5 to merge such things 14.0.x, it should only contain fixxes
It saves only some KB on the Webserver

melloware:
OK I  will make it 15.0.0 only.

PR #12287, opened 2024-07-15
Fix #12283: Handle ValueExpression attributes on p:validateFile

### LINKED ISSUE #12283: ValidateFile: with variable in sizeLimit not working
### Describe the bug

If i use p:valiadteFile with a variable in sizeLimit (fileUploadView.size) it doesn't consider it at runtime.

```xml
 <h:form id="frmTest">

			<p:growl id="growl" sticky="true" showDetail="true"/>

			<p:inputText value="#{fileUploadView.size}" >
				<p:ajax update="xx,input20" />
			</p:inputText>
			<h:outputText id="xx" value="#{fileUploadView.size}" />

			<p:fileUpload id="input20" update="growl" 
					  listener="#{fileUploadView.handleFileUpload}"
					  mode="advanced" dragDropSupport="true" auto="true" >
				<p:validateFile sizeLimit="#{fileUploadView.getSizeLimit(fileUploadView.size)}" allowTypes="/(\.|\/)(gif|jpeg|jpg|png)$/" />
			</p:fileUpload>

        </h:form>
```

```java
@Named
@RequestScoped
public class FileUploadView {

	private UploadedFile file;
	private String size;

	public void upload() {
		if (file != null) {
			FacesMessage message = new FacesMessage("Successful", file.getFileName() + " is uploaded.");
			FacesContext.getCurrentInstance().addMessage(null, message);
		}
	}

	public void handleFileUpload(FileUploadEvent event) {
		FacesMessage message = new FacesMessage("Successful", event.getFile().getFileName() + " is uploaded.");
		FacesContext.getCurrentInstance().addMessage(null, message);
	}

	public UploadedFile getFile() {
		return file;
	}

	public void setFile(UploadedFile file) {
		this.file = file;
	}


	public String getSizeLimit(String limit) {
		if (limit != null) {
			return limit;
		}
		return "100000";
	}

	@PostConstruct
	public void init() {
		setSize("70000");
	}

	public String getSize() {
		return size;
	}

	public void setSize(String size) {
		this.size = size;
	}

}
```

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

14.0.2

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0

### Java version

17

### Browser(s)

_No response_

jungm:
Fixes #12283, also refactored `AbstractBehaviour` to make behaviours and validators share the same code to handle ValueExpressions


tandraschko:
hmmm not sure if i like the delegation, which is 1 more object instance each beahvior/validator and so on

maybe create a interface for all the methods + default impls and init 
```
@Getter protected Map<String, Object> literals;
@Getter protected Map<String, ValueExpression> bindings;
```

in the Validator/Converter/BehaviorBase?

tandraschko:
discussed with markus that we align everything with Component and StateHelper

tandraschko:
discussed with markus
he will do some minor adjustments, after that we can merge

primefaces/src/main/java/org/primefaces/behavior/validate/ClientValidator.java
```diff
@@ -78,18 +78,18 @@ protected BehaviorAttribute[] getAllAttributes() {
     }
 
     public String getEvent() {
-        return eval(PropertyKeys.event, null);
```
melloware, line 81: @tandraschko for consistency with other code and Behaviors shouldn't this technically be called `ClientValidatorBehavior`?

melloware (changes requested):

primefaces/src/main/java/org/primefaces/behavior/validate/ClientValidator.java
tandraschko: yes -> new issue pls

melloware: OK i will create one and fix it.

PR #12299, opened 2024-07-17
Fix #12297: Move validation.js to core.js

### LINKED ISSUE #12297: Core: Add validation.js to core.js since its always included now
### Describe the bug

I'm trying to migrate to PF 14 but a new `validation.js` is always added and I don't need it...

Also If I defer the loading with OmniFaces `o:deferredScript` it will be added twice!

I've tried to disable the CSV from web.xml but `validation.js` is always added

```
<context-param>
</context-param>

<context-param>
    <param-name>primefaces.CLIENT_SIDE_LOCALISATION</param-name>
</context-param>

<context-param>
    <param-name>primefaces.INTERPOLATE_CLIENT_SIDE_VALIDATION_MESSAGES</param-name>
</context-param>
```

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

Community

### PrimeFaces version

14.0.2

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4

### Java version

21

### Browser(s)

_No response_

melloware:
Fix #12297: Move validation.js to core.js

### DIFF
--- a/primefaces/pom.xml
+++ b/primefaces/pom.xml
@@ -719,6 +719,10 @@
                                 <include>core/core.clientwindow.js</include>
                                 <include>ajaxstatus/ajaxstatus.js</include>
                                 <include>poll/poll.js</include>
+                                <include>validation/validation.common.js</include>
+                                <include>validation/validation.converters.js</include>
+                                <include>validation/validation.validators.js</include>
+                                <include>validation/validation.highlighters.js</include>
                             </includes>
                             <aggregations>
                                 <aggregation>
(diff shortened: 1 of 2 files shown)


tandraschko:
shouldnt we move BV and moment, too?

it currently might even a bug that we also need moment for normal validation?

melloware:
OK looked into it and there is no way to use client side validation without the server side flag set.  So I think all of the current code is proper.

I also have this defensive check in the JS code if someone does try it.

```js
            if (typeof moment === 'undefined') {
                PrimeFaces.error("Moment.js is not loaded! Please enable 'primefaces.CLIENT_SIDE_VALIDATION' in web.xml!");
            }
```

PR #12308, opened 2024-07-17
Fix #12307: DatePicker Now respect timeZone if set

### LINKED ISSUE #12307: DatePicker: Now/Today button always uses browser timezone not component timezone
### Describe the bug

Using the DatePicker with `showButtonBar` the Now/Today button always uses `new Date()` which is the browser timezone date.  It should be the Now of the Datepicker `timezone` property.

### Reproducer

```xml
<p:datePicker id="buttonbar" value="#{calendarJava8View.date10}" showButtonBar="true" timezone="Etc/UTC"/>
```

### Expected behavior

Now should put the time in the timezone of the DatePicker not the browser.  In the example above NOW should be UTC time and not my current browser's timezone.

### PrimeFaces edition

Community

### PrimeFaces version

14.0.2

### Theme

ALL

### JSF implementation

All

### JSF version

2.3

### Java version

11

### Browser(s)

ALL

melloware:
Fix #12307: DatePicker Now respect timeZone if set

### DIFF
--- a/primefaces-showcase/src/main/webapp/ui/input/datepicker/datePickerJava8.xhtml
+++ b/primefaces-showcase/src/main/webapp/ui/input/datepicker/datePickerJava8.xhtml
@@ -1,9 +1,5 @@
-<ui:composition xmlns="http://www.w3.org/1999/xhtml"
-                xmlns:ui="jakarta.faces.facelets"
-                xmlns:h="jakarta.faces.html"
-                xmlns:f="jakarta.faces.core"
-                xmlns:p="primefaces"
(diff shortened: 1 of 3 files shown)


tandraschko:
+1 if well tested

christophs78 (approved):
Code-additions/modifications look good to me.
From a professional view - is this what we think users expect? (For the usecases i personally know this should not do harm. So it´s Ok for me.)

melloware:
@christophs78 good question but i think this doesn't violate the law of least surprise (i hope).

PR #12347, opened 2024-07-29
Fix #12346: Spinner respect Min/Decimal annotations

### LINKED ISSUE #12346: Spinner: Not respecting @Min or @DecimalMin correctly
### Describe the bug

Spinner is not correctly finding `@Min(0)` or `@DecimalMin`

```xml
<p:spinner id="basic" value="#{spinnerView.number1}" required="true">
```

```java
    @Min(0)
    @Max(100)
    private Integer number1;
```

### Reproducer

[pf-spinner.zip](https://github.com/user-attachments/files/16418953/pf-spinner.zip)

### Expected behavior

Min should set the component minimum.

### PrimeFaces edition

Community

### PrimeFaces version

14.0.3

### Theme

ALL

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

11

### Browser(s)

_No response_

melloware:
Fix #12346: Spinner respect Min/Decimal annotations

### DIFF
--- a/primefaces/src/main/java/org/primefaces/metadata/transformer/AbstractInputMetadataTransformer.java
+++ b/primefaces/src/main/java/org/primefaces/metadata/transformer/AbstractInputMetadataTransformer.java
@@ -72,7 +72,7 @@ public abstract class AbstractInputMetadataTransformer implements MetadataTransf
         return Integer.MIN_VALUE;
     }
 
-    protected boolean isMaxlenghtSet(UIInput input) {
+    protected boolean isMaxlengthSet(UIInput input) {
         return getMaxlength(input) != Integer.MIN_VALUE;
     }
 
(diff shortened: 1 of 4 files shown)


primefaces/src/main/java/org/primefaces/metadata/transformer/AbstractInputMetadataTransformer.java
```diff
@@ -72,7 +72,7 @@ else if (input instanceof InputTextarea) {
         return Integer.MIN_VALUE;
     }
 
-    protected boolean isMaxlenghtSet(UIInput input) {
+    protected boolean isMaxLengthSet(UIInput input) {
```
tandraschko: It was by design, see the getter or HTML attr name

melloware: OK i changed the L to lowercase but the word "length" was misspelled "lenght"

tandraschko:
Maybe we should create getter on the spinner for the Default value and reuse there and in the StateHelper
Ofherwise it will get out of sync again

melloware:
good idea.

Issue #11681, opened 2024-03-27
Themes: Style select elements


jepsar:
They are used in the rows per page selector of the DataTable and the navigator of the DatePicker for example. They could do with some mild styling.

<img width="550" alt="Screenshot 2024-03-27 at 17 38 13" src="https://github.com/primefaces/primefaces/assets/7500178/48ccc974-38f3-4de8-aa58-8480f07490c2">

<img width="457" alt="Screenshot 2024-03-27 at 17 41 46" src="https://github.com/primefaces/primefaces/assets/7500178/462711d9-ffed-4a1d-b180-67177a5bc17f">

melloware:
Nice. When I looked years ago the styling of the native select one was limited but the web has come a long way since the last time I looked.

jepsar:
I did a little bit of styling in our project, and the select now looks like:

<img width="418" alt="Screenshot 2024-04-08 at 17 13 35" src="https://github.com/primefaces/primefaces/assets/7500178/e19223bd-5e47-4beb-8f5b-d89bdcab34f9">

melloware:
Definitely an improvement!

melloware:
i have this in the theme and its looking real good.

```css
    select, input {
      color: $inputTextColor;
      background: $inputBg;
      padding: $inputPadding;
      border: $inputBorder;
      outline: 0 none;
      font-size: $inputTextFontSize;
      border-radius: $borderRadius;
      transition: $transition;
      -webkit-appearance: none;
      appearance: none;
    
      &:focus {
        @include focused-input();
      }

      &:hover {
        border-color: $inputHoverBorderColor;
      }
    }
 ```

![image](https://github.com/primefaces/primefaces/assets/4399574/8ddd209f-9316-4c10-8e6c-3299cc9c5307)

However now it throws the paginator centering off which I need to figure out.

The new input styling pushing the icons so everything is not centered.

![image](https://github.com/primefaces/primefaces/assets/4399574/f9d79ebb-66a2-468a-9f85-757379e3d853)

melloware:
Agreed let me see what is removing that arrow.

melloware:
Check all `ui-inputfield` instances such as Spinner to make sure highlighting is working.

PR #12355, opened 2024-07-30
Fix #12352: Spinner allow min+required to be NULL

### LINKED ISSUE #12352: Spinner: required+min does not allow a NULL start value
### Describe the bug

Reported by one of my customers.  They need to be able to specify `min="0" max="100" required="true"` and FORCE the user to pick a value.  Right now that is not possible.

here is the scenario:

```xml
<p:spinner id="allowsNull" value="#{spinnerView.number1}" required="true" max="100" />

<p:spinner id="notAllowsNull" value="#{spinnerView.number1}" required="true" min="0" max="100"/>
```

Renders like this:
image

The reason is this code in the spinner.js widget has been there for years but to me this is not correct behavior.

```js
if(isNaN(parsedValue)) {
            if(PrimeFaces.trim(value) === '' && this.cfg.min !== undefined && this.cfg.required) {
                parsedValue = this.cfg.min;
            } else {
                parsedValue = null;
            }
        }
```

Basically it defaults to the `min` if `required="true"` the required becomes moot and a user can submit the page without having to select a value.

### Reproducer

[pf-spinner-required.zip](https://github.com/user-attachments/files/16428388/pf-spinner-required.zip)

### Expected behavior

The spinner should be allowed to be NULL if using a `min` and `required`

### PrimeFaces edition

Community

### PrimeFaces version

14.0.3

### Theme

ALL

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

11

### Browser(s)

_No response_

melloware:
Fix #12352: Spinner allow min+required to be NULL

Neeed to update SASS theme to make #11681 more specific

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/spinner/Spinner007.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/spinner/Spinner007.java
@@ -1,0 +1,54 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2024 PrimeTek Informatics
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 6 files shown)


jepsar (approved):

tandraschko:
what are those theme changes?

melloware:
@tandraschko we added this in 15.0: https://github.com/primefaces/primefaces/issues/11681

Well that theme is not specific enough so the Spinner "required" red highlighting was not showing up because that other style was "more specific".  I am going to do more testing today.

tandraschko:
shouldnt we make a new issue then for better tracking?

melloware:
OK i think i will reopen https://github.com/primefaces/primefaces/issues/11681 since its related to that 15.0.0 ticket.

tandraschko:
code looks good, +1 if well tested

PR #12363, opened 2024-08-01
#12362 - DefaultLazyDataModel: better implementation of count

### LINKED ISSUE #12362: DefaultLazyDataModel: better implementation of count
Currently count and load both invokes the valueSupplier, but this is not necessary as mentioned by https://primefaces.github.io/primefaces/14_0_0/#/components/datatable?id=lazy-loading

we can reduce it to 1 call

tandraschko:
Fix #12362 - DefaultLazyDataModel: better implementation of count


melloware:
Yep no problem.

melloware (approved):

PR #12368, opened 2024-08-01
DefaultLazyDataModel: add ability for mixed filterBy in valueSupplier method and also automatic reflection based

### LINKED ISSUE #12366: DefaultLazyDataModel: add ability for mixed filterBy in valueSupplier method and also automatic reflection based
i currently work on a legacy backend and there is a API for receiving objects, which will be shown in the datatable, which accepts some filter params. Other filter params should be applied automatically by our reflection logic.

So imagine something like this:
```
            <p:dataTable id="list" value="#{controller.model}" var="item" widgetVar="myList">
                <p:column field="customer.name" />
                <p:column field="status" />
                <p:column field="created">
                    <f:facet name="filter">
                        <p:datePicker onchange="PF('myList').filter()" converter="#{myConverter}" />
                    </f:facet>
                </p:column>
            </p:dataTable>
```

the API e.g. contains a filter param for the created date but not e.g. for status.

i would like to see something like this:

```
    @PostConstruct
    public void init() {
        model = DefaultLazyDataModel.<MyObject>builder()
                .rowKeyProvider(MyObject::getID)
                .ignoreFilter((filterMeta) -> {
                      return filterMeta.getField().equals("created") })
                .valueSupplier((filterBy) -> {
                    return someAPI.getAll(filterBy.get("created"));
                }).build();
```

tandraschko:
Fix #12366

### DIFF
--- a/docs/15_0_0/components/datatable.md
+++ b/docs/15_0_0/components/datatable.md
@@ -190,7 +190,7 @@ public class CarBean {
     @PostConstruct
     public void init() {
         cars = DefaultLazyDataModel.<Car>builder()
-                .valueSupplier(() -> carService.getCars())
+                .valueSupplier((filterBy) -> carService.getCars())
                 .rowKeyProvider(Car::getId) // required for selection
                 .build();
     }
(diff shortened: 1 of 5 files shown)


tandraschko:
@melloware if you reviewed it and its ok, could we port it to 14.x?

melloware (approved):

melloware:
Yep on it.

PR #12381, opened 2024-08-01
Fix #12379 14.0.4 DataTable: dont try to write filterValue back to EL if its read-only

### LINKED ISSUE #12379: DataTable: dont try to write filterValue back to EL, if its read-only
```
<p:column field="myField" " filterValue="#{controller.myFieldFilterDefaultValue}">
```

the EL only is a getter, no setter

melloware:
Fix #12379 14.0.4 DataTable: dont try to write filterValue back to EL if its read-only


tandraschko:
thanks man!

tandraschko:
im still wondered how much still can be improved for certain, not common, usecases

melloware:
Agreed it sounds like you have found some good ones just implementing this for your customer that we have not hit yet.

PR #12392, opened 2024-08-02
Fix #11735: Added rowsPerPage in the PageEvent

### LINKED ISSUE #11735: Paginator: PageEvent add rowsPerPage property
### Description

You currently cannot capture rows per page changes (from the `rowsPerPageTemplate` attribute) through the `PageEvent` using the `page` event of a `DataTable`.

Add a `rowsPerPage` property so you can.

### Additional context

As a workaround you can use the request parameter
```java
  public void onPaginate() {
    final DataTable dataTable = (DataTable) Components.getCurrentComponent();
    final String rppParam = Faces.getRequestParameter(dataTable.getClientId() + "_rppDD");
    System.out.println("rppParam: " + rppParam);
  }
```

Also, in the current documentation is not clear you can use the `page` event to capture rows per page changes.

melloware:
Fix #11735: Added rowsPerPage in the PageEvent

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java
+++ b/primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java
@@ -355,8 +355,10 @@ public class DataTable extends DataTableBase {
                 int rows = getRowsToRender();
                 int first = Integer.parseInt(params.get(clientId + "_first"));
                 int page = rows > 0 ? (first / rows) : 0;
+                String rowsPerPageParam = params.get(clientId + "_rows");
+                Integer rowsPerPage = LangUtils.isNotBlank(rowsPerPageParam) ? Integer.parseInt(rowsPerPageParam) : null;
 
-                wrapperEvent = new PageEvent(this, behaviorEvent.getBehavior(), page);
+                wrapperEvent = new PageEvent(this, behaviorEvent.getBehavior(), page, rowsPerPage);
             }
             else if ("sort".equals(eventName)) {
                 wrapperEvent = new SortEvent(this, behaviorEvent.getBehavior(), getSortByAsMap());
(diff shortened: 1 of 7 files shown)


primefaces/src/main/java/org/primefaces/component/datatable/DataTable.java
```diff
@@ -354,9 +354,10 @@ else if ("rowUnselect".equals(eventName) || "rowUnselectCheckbox".equals(eventNa
             else if ("page".equals(eventName) || "virtualScroll".equals(eventName) || "liveScroll".equals(eventName)) {
                 int rows = getRowsToRender();
                 int first = Integer.parseInt(params.get(clientId + "_first"));
+                int rowsPerPage = Integer.parseInt(Objects.toString(params.get(clientId + "_rows"), "-1"));
```
tandraschko: should we make it nullable instead of -1?

melloware: The only reason i left it lower `int` with a value was to be consistent with the other fields...and since for `virtuslScroll` and `LiveScroll` its a PageEvent with no rows i figured just make it -1 and leave it an `int` vs making it an `Integer`?

```java
                int rows = getRowsToRender();
                int first = Integer.parseInt(params.get(clientId + "_first"));
                int rowsPerPage = Integer.parseInt(params.get(clientId + "_rows"));
                int page = rows > 0 ? (int) (first / rows) : 0;

                PageEvent pageEvent = new PageEvent(this, behaviorEvent.getBehavior(), page, rowsPerPage);
```

tandraschko: i see
i would prefer nullable, its more clear

tandraschko: actually we should rethink other fields, too then

melloware: OK I am going to refactor this to be `Integer` and allow NULL.

jepsar (approved):

PR #12395, opened 2024-08-03
#12394 - Include SASS themes in project folder 

### LINKED ISSUE #12394: Include SASS themes in project folder

tandraschko:
Fixes #12394


tandraschko:
2 things to mention
1) we have more sass themen then themes in PF.jar, i skip them for now
2) some themes.css already have changes, maybe we missed some commits before

melloware (approved):

melloware:
Hmmm I have been keeping the themes up to date and usually copy them over when a change it made?  It could just be a line ending difference with the DART compressor?

tandraschko:
we may also add the other themes to something like primefaces-extra-themes in future, not sure

melloware:
I like this way better.

tandraschko:
i mean that we have like 10 themes in primefaces-themes but only 5 in primefaces.jar

melloware:
Should we just add them all?

tandraschko:
dont think so, 1 is about 200kb

tandraschko:
they may be just legacy, no idea

melloware:
OK then i think what you have is good.  Those other themes might be used in premium themes or they are legacy like you said.

melloware:
want me to squash and merge it?

tandraschko:
yeah

jepsar:
What happened to the build scripts?

melloware:
you don't need them now its done as part of the Maven build in the pom.xml it copies the files to the correct place.

PR #12400, opened 2024-08-05
Fix #12399: CSP improve NULL checking

### LINKED ISSUE #12399: Problem with CSP for Ajax-Requests
### Discussed in https://github.com/orgs/primefaces/discussions/2432

<sup>Originally posted by **pneumcke** August  1, 2024</sup>
Since 14.0.2 probably caused by https://github.com/primefaces/primefaces/commit/2b0b11e490f2110a1170db6cd056d9343a4c7ceb all Ajax requests fail in our application due to CSP violations.

The previous implementation in CSPState had two ways of accessing the nonce:
  this.nonce = (String)viewMap.get("primefaces.nonce");
and
  this.nonce = (String)this.context.getExternalContext().getRequestParameterMap().get("primefaces.nonce");

For the Ajax requests, the first access returns "null" - which was OK since the nonce was later accessed via the request parameter. 

Since 14.0.2 this behaviour does not work anymore. The view root exists and has 16 parameters (WELD%ManagedBean, omnifaces.Socket, ...) - but no "primefaces.nonce".

The JSF implementation is myfaces 4.0.1

Do you have any ideas? Are we doing something wrong?

</div>

melloware:
Fix #12399: CSP improve NULL checking

@tandraschko this prevents in case the map has NULL so `Objects.equals(NULL, "")` would be false we should always be comparing a String to a String with a default of Constants.EMPTY_STRING so `Objects.equals("", "")`

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/core/csp/CoreCsp001.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/core/csp/CoreCsp001.java
@@ -1,0 +1,54 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2024 PrimeTek Informatics
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.integrationtests.core.csp;
+
+import java.io.Serializable;
+
+import javax.faces.application.FacesMessage;
+import javax.faces.context.FacesContext;
+import javax.faces.view.ViewScoped;
+import javax.inject.Named;
+
+import lombok.Data;
+
+@Named
+@ViewScoped
+@Data
+public class CoreCsp001 implements Serializable {
+
+    private static final long serialVersionUID = 8797995450720503195L;
+
+    private String firstname;
+    private String lastname;
+
+    public String openInNewPage() {
+        return "coreCsp001-page2.xhtml";
+    }
+
+    public void save() {
+        FacesContext.getCurrentInstance().addMessage(null,
+                new FacesMessage("Welcome " + firstname + " " + lastname));
+    }
+
+}
--- a/primefaces-integration-tests/src/main/webapp/core/csp/coreCsp001-page2.xhtml
+++ b/primefaces-integration-tests/src/main/webapp/core/csp/coreCsp001-page2.xhtml
@@ -1,0 +1,21 @@
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml"
+      xmlns:h="http://xmlns.jcp.org/jsf/html"
+      xmlns:f="http://xmlns.jcp.org/jsf/core"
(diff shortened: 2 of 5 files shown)


tandraschko:
why Constants.EMPTY_STRING as default? its undefined, null is more correct

melloware:
I changed it to nulls but the reason i went with EMPTY_STRING is right now it only works if `NULL == NULL` not if `NULL == ""` i was trying to defend against any combination of NULL or EMPTY_STRING as equals.

tandraschko:
does context.getExternalContext().getRequestParameterMap() return null or "" with a empty param?

tandraschko:
otherwise we can change again to "" but with comment

tandraschko:
can you manually try a request e.g. myView.xhtml?blub=

and see what the map returns?
the docs are not clear enough

melloware:
let me test both Mojarra and MyFaces and see what they do.

melloware:
@tandraschko here is the screenshot of our scenario we can only produce in our real world Mojarra 2.3 app.

image

the ViewState is NULL but the Request is there?   This is when we open a new page with `<h:commandLink ajax="false"`

tandraschko:
is it postback or ajax request?

tandraschko:
if its a POST, it should contain the noonce actually, independing if AJAX or not
maybe its a none-jsf form?

melloware:
@FlipWarthog is out of the office on holiday this week but here is the exact link causing the issue.

1. Confirmed it is inside an `<h:form>`
2. This link is in a Datatable row
3. 

```xml
<h:commandLink action="#{graphController.graphClicked()}" 
title="Graphs" 
target="_blank" 
rendered="#{row.votesExist}" >
	<f:setPropertyActionListener value="#{row}" target="#{meetingMaintenanceForm.questionSelected}"/>
	<span class="ui-button-icon-left ui-icon ui-c ui-icon-suitcase"/>
	<span class="ui-button-text ui-c">ui-button</span>
</h:commandLink>
```

FlipWarthog:
Specifically, the commandLink opens a new tab and the _existing tab_ now throws CSP errors on any other Ajax button.

tandraschko:
It think we should add a IT:
- Ajax + Non-Ajax Post
- with and without redirect
- also with Forward

Then we should cover all cases

tandraschko:
Im almost sure that it can also bei caused by a Non-Ajax forward

melloware:
i think the first step is we need to reproduce this with a PF Test reproducer so we can prove the issue first. Then add IT tests.

melloware:
OK reproducer:
[pf-12399.zip](https://github.com/user-attachments/files/16499255/pf-12399.zip)

1. Go to http://localhost:8080/
2. Click `Save` to see Validation error.
3. Click the `"Open In New Page"` link.
4. Go back to page 1 and press `"Save"` again you see the error.


```
SEVERE: Error Rendering View[/test.xhtml]
org.primefaces.csp.CspException: CSP nonce mismatch
        at org.primefaces.csp.CspState.getNonce(CspState.java:60)
        at org.primefaces.csp.CspResponseWriter.listenOnEndAttribute(CspResponseWriter.java:192)
        at org.primefaces.csp.CspResponseWriter.write(CspResponseWriter.java:173)
        at org.primefaces.csp.CspPartialResponseWriter.write(CspPartialResponseWriter.java:177)
        at java.base/java.io.Writer.write(Writer.java:290)
        at java.base/java.io.Writer.write(Writer.java:249)
        at org.primefaces.util.WidgetBuilder.init(WidgetBuilder.java:64)
        at org.primefaces.util.WidgetBuilder.init(WidgetBuilder.java:94)
        at org.primefaces.util.WidgetBuilder.init(WidgetBuilder.java:78)
        at org.primefaces.component.messages.MessagesRenderer.encodeScript(MessagesRenderer.java:111)
        at org.primefaces.component.messages.MessagesRenderer.encodeEnd(MessagesRenderer.java:46)
        at javax.faces.component.UIComponentBase.encodeEnd(UIComponentBase.java:600)
        at javax.faces.component.UIComponent.encodeAll(UIComponent.java:1655)
        at javax.faces.render.Renderer.encodeChildren(Renderer.java:152)
        at javax.faces.component.UIComponentBase.encodeChildren(UIComponentBase.java:571)
        at javax.faces.component.UIComponent.encodeAll(UIComponent.java:1648)
        at com.sun.faces.context.PartialViewContextImpl$PhaseAwareVisitCallback.visit(PartialViewContextImpl.java:629)
        at com.sun.faces.component.visit.PartialVisitContext.invokeVisitCallback(PartialVisitContext.java:159)
        at javax.faces.component.UIForm.visitTree(UIForm.java:366)
```

melloware:
I have confirmed this now fixes the issue.

```
if (context.isPostback() && LangUtils.isNotBlank(nonceViewState) && !Objects.equals(nonceViewState, nonceRequest)) {
```

tandraschko:
Can you add a IT?

melloware:
Yep let me add an IT now.

tandraschko:
@melloware 
AFAICS its a forward right?
we can then even improve the impl to ignore the viewstate nonce in a forward

because otherwise its actually valid

melloware:
OK IT test added that fails.

tandraschko:
i see the exception but the test still runs green

melloware:
really it fails for me locally with a StaleElementExcepiton?

melloware:
image

image

tandraschko:
still working on my machine :D

however i understand the case now, its just a new JSF view in the same request

melloware:
Werid so in mine as soon as that AJAX goes the PF exception Handler kicks in and displays a new page with the stack trace so all of the Selenium elements are gone at that point.  Not sure why that is not happening for you.

PR #12417, opened 2024-08-06
Fix #12416: Core.js use crypto.randomUUID if available

### LINKED ISSUE #12416: Core JS: use crypto randomUUID
[uuid: function() {](https://sonarcloud.io/code?id=org.primefaces%3Aprimefaces&selected=org.primefaces%3Aprimefaces%3Asrc%2Fmain%2Fresources%2FMETA-INF%2Fresources%2Fprimefaces%2Fcore%2Fcore.js&line=1312)

instead of manually using math.random:
https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

melloware:
Fix #12416: Core.js use crypto.randomUUID if available

According to MDN there are contexts where `crypto` may not be available for the browser so i left the fallback in.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.js
@@ -1305,21 +1305,21 @@
         /**
          * Generate a RFC-4122 compliant UUID to be used as a unique identifier.
          *
+         * Uses crypto.randomUUID() if available, otherwise falls back to a custom implementation.
+         *
          * See https://www.ietf.org/rfc/rfc4122.txt
          *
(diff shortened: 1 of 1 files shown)


melloware:
Yes the fallback method in crypto has been around for years and even IE11 supports it.  Random UUId is newer and I tested chrome and Firefox it is there but I just don't know about all the mobile browser options out there

jepsar:
https://caniuse.com/?search=randomUUID

melloware:
Yep but this one has been around way longer: https://caniuse.com/?search=getRandomValues

So randomUUID should work for 90% of the circumstances and the fallback covers the last 10%.

PR #12453, opened 2024-08-11
Add Bangla locale

melloware:


riyadomf:
@melloware is there a way to localize the days and year? I may pass a function which will convert each number to my own locale.
image

melloware:
So this is something that would need to be implemented in all 4 Prime Libs.  I think I saw an open ticket on NG or Vue that asked for the same thing.

PR #12466, opened 2024-08-13
Fix #12463: Signature allow typed text, improve accessibility

### LINKED ISSUE #12463: Signature: Allow typing of signature along with drawing
### Description

Request from my client.  They don't like always using the mouse to draw a signature as its really hard to draw a good one.  They wanted an option for visually impaired or people who just don't like to draw to be able to type their signature.

signature

Ability to type the signature and control font and fontsize.  Default to "cursive" font provided by the browser.

### Additional context

_No response_

melloware:
Fix #12463: Signature allow typed text, improve accessibility

- [x] ARIA and WCAG accessibility
- [x] `ESC` to clear signature
- [x] Updated to standards we use in other components
- [x] Allow typing of signature

signature


tandraschko:
+1 if well tested

PR #12478, opened 2024-08-16
Sonar cleanup possible security DOS warnings

melloware:


melloware:
OK i pulled those and just fixed the two issues that could actually be fixed.

PR #12538, opened 2024-08-30
Fix Sonar Quality Gate failures

melloware:


tandraschko:
should sonar really scan unittests?

melloware:
@Rapster good question?

Rapster:
I want to say yes, why not?

PR #12539, opened 2024-08-30
Fix #12526: SelectOne ARIA/WCAG for groups

### LINKED ISSUE #12526: SelectOneMenu: Accessibility enhancement render grouped with <optgroup> tags
### Description

I see that the "Grouping" dropdown on the [SelectOneMenu page](https://www.primefaces.org/showcase/ui/input/oneMenu.xhtml?jfwid=33b32) in the Primefaces showcase ends up being rendered as a `<ul>` with `<li>` elements. The group label items are not selectable but have a special "ui-selectonemenu-item-group" class. 

The problem I have is, when reading these with a screenreader, such as NVDA, it still includes these groupings in the count of items, and does not reference the group at all. Despite the fact that there are only 10 valid options to select from, arrowing through the list yields the following reading:

"Select One 1 of 12
Germany 3 of 12
...
Mexico 12 of 12"

Ideally, this type of dropdown could work like the example listed [here in MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup), using the `<optgroup label="">` tag.

When using NVDA on this page, it correctly identifies that there are 6 selectable options available, and reads the groups as well. Arrowing through _this_ list yields the following reading, which is the behavior I would expect:

"Theropods grouping
Tyrannosaurus 1 of 6
Velociraptor 2 of 6
...
Sauropods grouping
Diplodocus 4 of 6"

melloware:
Fix #12526: SelectOne ARIA/WCAG for groups

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectonemenu.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectonemenu.js
@@ -183,11 +183,11 @@ PrimeFaces.widget.SelectOneMenu = PrimeFaces.widget.DeferredWidget.extend({
     initContents: function() {
         this.itemsContainer = this.itemsWrapper.children('.ui-selectonemenu-items');
         this.items = this.itemsContainer.find('.ui-selectonemenu-item');
-        this.optGroupsSize = this.itemsContainer.children('li.ui-selectonemenu-item-group').length;
+        this.optGroupsSize = this.itemsContainer.find('li.ui-selectonemenu-item-group').length;
 
         var $this = this,
         selectedOption = this.options.filter(':selected'),
-        highlightedItem = this.items.eq(this.options.index(selectedOption));
+        highlightedItem = this.selectItemFromOption(selectedOption);
 
         //disable options
         $this.items.filter('[disabled]').addClass('ui-state-disabled');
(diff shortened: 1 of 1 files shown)


melloware:
@christophs78 i know you and your coworker did a lot of work on this so it would be great if you guys could retest this. 😄 

Leaving as draft for now

melloware:
I will fix the integration tests but i want to verify this actually is better ARIA/WCAG compliant before doing so

christophs78:
Thanks for working on this! I´ll be back at office at 9th of september and check this with my coworker this week.

melloware:
@christophs78 my local dev is also flaky its like the Chrome Driver keeps getting stuck so its making it impossible for me to debug these test failures. It seems i have introduced an "off by one error" somewhere now that there are groups but I am having trouble debugging. I could use an assist.

christophs78:
Regarding random Chrome(Driver) 128 regressions vs 127 we discussed on Discord i also hope a bit for Chrome(Driver) 129. The current best thing to run integration test (local) stable is to run them one Firefox. 

With Firefox i get following results for this PR (SelectOneMenu test´s only).
image

Same for Chrome ignoring some sporadic failures due to Chrome(Driver) 128.

melloware:
OK that helps i fixed one of the issues now checking the other one.

melloware:
All fixed!  The Integration tests caught some bugs thankfully and I have been able to resolve them all.

PR #12549, opened 2024-09-01
Fix #12511: Datatable/TreeTable filterNormalize

### LINKED ISSUE #12511: DataTable/TreeTable: New `filterNormalize` attribute
### Description

Firstly I really appreciate the new `filterNormalize="true"` attribute, available in components like `<p:selectCheckboxMenu/>`
 
 **I was wondering if it is possible to have the same behaviour in _<p:datatable columns_ /> as well.** E.g.:
`<p:column   filterBy="#{op.name}"   filterMatchMode="contains"  **filterNormalize="true"**   .../> `

BTW, I'm actually using a custom filter method in Bean and using it like that: `filterFunction="#{utilBean.filterContainsWithoutAccents}"` and it works fine when user filter using the Column Header Filter Box, but fails when user use the main global filter box like:
` <h:inputText id="globalFilter"  onkeydown="if (event.keyCode == 13) { return  PF('tableWdg').filter(); }"        />` which not calls ServerSide Bean method` public boolean filterContainsWithoutAccents(Object value, Object filter, Locale locale) {.`.. anymore.

Regards 

For me, the cleanest solution would be  to have  `filterNormalize=""` attribute like we have now in components like `<p:selectCheckboxMenu/>` to filtering for column data, escaping accented characters.

As a workaround I'm using a server side custom filter method called in: `filterFunction="#{utilBean.filterContainsWithoutAccents}"`  but apparently it just work   when called by the 'Column Header Filter Box', and it's not   called   by global filter `PF('tableWdg').filter();`,  there is a way to call all declared columns `filterFunction=""` in that context as well?

Thanks in advance.

### Additional context

_No response_

melloware:
Fix #12511: Datatable/TreeTable filterNormalize

@jepsar this is what I was thinking because its easier to capture here on the server side no matter what filters or global filters in play.


primefaces/src/main/java/org/primefaces/component/datatable/feature/FilterFeature.java
```diff
@@ -138,7 +141,7 @@ public void filter(FacesContext context, DataTable table) {
 
                 if (globalFilter != null && globalFilter.isActive() && !globalMatch.get() && !hasGlobalFilterFunction) {
                     FilterConstraint constraint = globalFilter.getConstraint();
-                    Object filterValue = globalFilter.getFilterValue();
+                    Object filterValue = LangUtils.normalize(globalFilter.getFilterValue(), table.isFilterNormalize());
                     globalMatch.set(constraint.isMatching(context, columnValue, filterValue, filterLocale));
```
jepsar: We should normalize the columnValue as well, right?

melloware: Good point. We probably should for comparison.

Rapster:
Shouldn't we handle this in FilterMeta instead?

melloware:
yes in theory we could i have to see where those FilterMeta are constructed and would they have access to the `table.isFilterNormalize()` new setting for both DT and TT?

Rapster:
You can either pass UITable or add a boolean to `of` method

Rapster:
Missing code in setFilterValue and getLocalValue ?

melloware:
All i did was when `FilterMeta.of` is called to create filter meta is passes the DT or TT value for `isFilterNormalize()`.  And then we apply the normalize on the value.  Is this wrong?

jepsar:
If the dataTable contain "café" but you search on "cafe", I believe you now only normalize "cafe".

melloware:
so was my previous PR correct?  where i did everything last minute in FilterFeature?

primefaces/src/main/java/org/primefaces/component/datatable/feature/FilterFeature.java
jepsar: @melloware basically this was the same issue I believe.

melloware:
So we need to also Normalize the other side of the filtering where its checking the value.  I have to figure out where that happens 😄

melloware:
OK i see what @Rapster meant needed to add it to `SetFilterValue` and `getLocalValue` and now its comparing the filter to the table value properly.

tandraschko:
i would personally prefer to not put logic into setters or getters as we loose the actual filterValue (also if not needed currently) and create a method like: 'resolveFilterValue', where we can put logic into 

it would be just a bit nicer IMO but the PR is also ok for me.

melloware:
@Rapster any thoughts on the resolveFilterValue?

Rapster:
Code is fine IMO

PR #12553, opened 2024-09-03
Fix #12551: Validator allowTypes regex with flags

### LINKED ISSUE #12551: ValidateFile: RegEx for allowTypes, modifiers no longer supported
### Discussed in https://github.com/orgs/primefaces/discussions/2619

<sup>Originally posted by **cyanotyp** September  2, 2024</sup>
Hi all,
I have a question if in PF14 the modifiers for **RegEx** on e.g. `<p:validatefile allowTypes` are no longer supported or if this is a bug .
- My simple useCase was in PF13  `allowTypes="/(\.|\/)(xlsx)$/i"` 
- the **/i** modifier was added since some users preferred to use file extension in capitals e.g myfile.XLSX
- This worked fine in PF13, but in 14 a file with caps extension now shows the "unsupported file type" message.

I took a little time to check the PF source code and found in
[https://github.com/primefaces/primefaces/blob/14.X/primefaces/src/main/resources/META-INF/resources/primefaces/validation/validation.validators.js](https://github.com/primefaces/primefaces/blob/14.X/primefaces/src/main/resources/META-INF/resources/primefaces/validation/validation.validators.js)
```
 // but as we parse the data-attribute from string to RegEx object, we must remove leading and ending slashes
     var transformedAllowtypes = allowtypes.substring(1, allowtypes.length - 1);
```
So a RegEx always is always stripped by 1 char at both ends, so my  **/i** modifier is stripped.
in PF 13 it was 
```
  if(!this.regex.test(value)) {
        throw vc.getMessage(this.TYPE_MESSAGE_ID, vc.getLabel(element));
   }
```

A "work-around" would be to change my RegEx to allowTypes="/(\.|\/)([cC][sS][vV]|[tT][xX][tT])$/" , but this is not very nice 

### Reproducer

create a <p:fileupload and attach a <p:validatefile allowTypes="/(\.|\/)(xlsx)$/i" as described in https://www.primefaces.org/showcase/ui/file/upload/single.xhtml?jfwid=39a38
```
<p:fileUpload id="myUpload" listener="#{myBean.upload}" displayFilename="true" mode="advanced" >
     <p:validateFile allowTypes="/(\.|\/)(xlsx)$/i" fileLimit="1"/>
</p:fileUpload>
```
Try to upload a file named myfile.TXT (extension in caps)
Upload shows "Unsupported FileType" message

### Expected behavior

with the usecase in "Reproducer", Try to upload a file named myfile.TXT (extension in caps) shall work and not show "Unsupported FileType" message

### PrimeFaces edition

Community

### PrimeFaces version

14.0.4-jakarta

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

3.0

### Java version

17

### Browser(s)

Firefox 128 ESR

melloware:
Fix #12551: Validator allowTypes regex with flags

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/validation/validation.validators.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/validation/validation.validators.js
@@ -123,8 +123,10 @@ if (window.PrimeFaces) {
                 if (allowtypes) {
                     // normally a regex is a object like /(\.|\/)(csv)$/
                     // but as we parse the data-attribute from string to RegEx object, we must remove leading and ending slashes
-                    var transformedAllowtypes = allowtypes.substring(1, allowtypes.length - 1);
-                    allowtypesRegExp = new RegExp(transformedAllowtypes);
+                    var regexParts = allowtypes.match(/^\/(.*)\/([a-z]*)$/);
+                    var transformedAllowtypes = regexParts[1];
+                    var flags = regexParts[2];
+                    allowtypesRegExp = new RegExp(transformedAllowtypes, flags);
                 }
 
                 if (filelimit && value.length > filelimit) {


cyanotyp (commented):
I have to admin I'm not a javascript expert! 
just wondering: if there are NO flags "var flags = regexParts[2]; " in java would throw an indexoutofbounds exception - how about javascript?

cyanotyp (approved):
Verified your changes work with AND without modifier 👍

PR #12556, opened 2024-09-04
XDev 2.2.0 change Color to RGBAColor

melloware:
https://github.com/xdev-software/chartjs-java-model/blob/develop/CHANGELOG.md#220


jepsar (commented):
It's in the docs as well https://primefaces.github.io/primefaces/14_0_0/#/components/chart

melloware:
good catch!

PR #12583, opened 2024-09-11
Fix #12577: Screenreader improve hidden accessible

### LINKED ISSUE #11577: SelectBooleanCheckbox: in Firefox, NVDA just reads "blank" when checkbox is focused with tab
### Describe the bug

When you use tab to select a checkbox, NVDA simply reads "blank." You then need to press the arrow down key to hear the itemLabel of the checkbox. In both cases, you can press space to check or uncheck the checkbox, so it's unclear where the focus lies the first time the screen reader reads "blank." When you click the checkbox with the mouse, it always says "blank," regardless of the checkbox's state.

This issue seems to only arise when using Firefox with NVDA:

- NVDA and selectBooleanCheckbox function correctly with other browsers like Chrome or Edge.
- Other screen readers, such as Microsoft Narrator, function properly with selectBooleanCheckbox in Firefox.
- Other checkboxes, like those from [Ant Design](https://ant.design/components/checkbox), work with NVDA and Firefox.

This problem might stem from Firefox or NVDA rather than PrimeFaces, so I'd like to apologize in advance if this is not the appropriate place to address it. However, since I'm unsure where the issue lies, any help would be greatly appreciated.

### Reproducer

Can be reproduced in the [Primefaces Showcase](https://www.primefaces.org/showcase/ui/input/booleanCheckbox.xhtml?jfwid=918e0) using Firefox (123.0.1) and NVDA (2023.3.4).

### Expected behavior

NVDA should behave the same in Firefox as in Chrome or Edge

### PrimeFaces edition

Community

### PrimeFaces version

whatever the showcase uses

### Theme

_No response_

### JSF implementation

All

### JSF version

-

### Java version

-

### Browser(s)

Firefox 123.0.1 (64-Bit)

### LINKED ISSUE #11623: SelectOneListbox: NVDA screenreader issue with Firefox
### Describe the bug

Similar to [this](https://github.com/primefaces/primefaces/issues/11577) issue i found, the NVDA screenreader doesn't work correctly for the SelectOneListbox in Firefox when navigating with the keyboard. The screenreader only reads out "selected" or "not selected" instead of the elements. This issue doesn't exist in other browsers like Chrome.
It does work with another similar listbox I've tried though: https://primeng.org/listbox
But I'm not sure if its even a PrimeFaces issue of if the cause lies elsewhere. As you mentioned in the linked issue, the problem doesn't seem to be that troublesome but I just wanted to let you know it exists.

### Reproducer

Navigate through the SelectOneListbox with the keyboard (arrow keys) in the [Showcase](http://www.primefaces.org/showcase/ui/input/listbox.xhtml?jfwid=4576d) with NVDA and Firefox

### Expected behavior

NVDA should read out the selected elements

### PrimeFaces edition

None

### PrimeFaces version

Showcase version

### Theme

_No response_

### JSF implementation

All

### JSF version

/

### Java version

/

### Browser(s)

Firefox 123.0.1 (64-Bit)

### LINKED ISSUE #12577: SelectBooleanCheckbox, ToggleSwitch: NVDA (Screen reader) reads "empty" when focusing
### Describe the bug

The NVDA screen reader doesn't recognize the labels for the checkbox and the toggle switch

### Reproducer

1. Activate NVDA
2. go to [the checkbox demo page](http://www.primefaces.org/showcase/ui/input/booleanCheckbox.xhtml) or the [toggle switch demo page](http://www.primefaces.org/showcase/ui/input/toggleSwitch.xhtml?jfwid=a29d8)
3. when you focus on the checkbox or the toggle switch, NVDA just announces "empty" ("leer" in German).

### Expected behavior

The labels of both will be announced

### PrimeFaces edition

Community

### PrimeFaces version

14

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

any

### Java version

any

### Browser(s)

Firefox and Chrome (any version) under Windows 10 and NVDA 2024_3

### Our Workaround

    html .ui-helper-hidden-accessible {
      border: unset;
      clip: unset;
      height: unset;
      margin: unset;
      overflow: unset;
      padding: unset;
      width: unset;
      pointer-events: none;
      opacity: 0;
    }
    
    .ui-helper-hidden-accessible input {
      transform: unset !important;
    }

### Related

https://github.com/primefaces/primefaces/issues/6561

melloware:
Fix #12577: Screenreader improve hidden accessible
Fix https://github.com/primefaces/primefaces/issues/11623 NVDA issues with ui-helper-hidden-accessible
Fix https://github.com/primefaces/primefaces/issues/11577: NVDA issues with ui-helper-hidden-accessible

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.css
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.css
@@ -1,10 +1,3 @@
-.ui-helper-hidden-accessible input,
-.ui-helper-hidden-accessible select {
-    -webkit-transform: scale(0); 
-    -moz-transform: scale(0); 
-    transform: scale(0);
-}
-
 .ui-shadow {
     -moz-box-shadow: 0px 5px 10px rgba(0,0,0,0.8);
     -webkit-box-shadow: 0px 5px 10px rgba(0,0,0,0.8);
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.css
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.css
@@ -3,22 +3,22 @@
     padding: 0.3em;
 }
 
-.ui-icon { 
-    display: block; 
-    vertical-align: baseline;
+.ui-icon {
+    display: block;
     margin-top: 0px;
 }
 
-/* https://stackoverflow.com/a/62109988/502366 */
+/* Modern override for all browsers and screenreaders */
 .ui-helper-hidden-accessible {
-    border: 0 !important;
-    padding: 0 !important;
-    margin: -1px !important;
+    border: unset;
+    clip: unset;
+    height: unset;
+    margin: unset;
+    overflow: unset;
+    padding: unset;
+    width: unset;
     position: absolute !important;
-    height: 1px !important;
-    width: 1px !important;
-    overflow: hidden !important;
-    clip: rect(1px, 1px, 1px, 1px) !important; /* Legacy Browsers */
-    clip-path: inset(50%) !important; /* Modern Browsers */
+    pointer-events: none !important;
+    opacity: 0 !important;
     white-space: nowrap !important;
 }


primefaces/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.css
```diff
@@ -3,22 +3,21 @@
...
-    display: block; 
-    vertical-align: baseline;
+.ui-icon {
+    display: block;
     margin-top: 0px;
 }
 
-/* https://stackoverflow.com/a/62109988/502366 */
+/* Modern override for all browsers and screenreaders */
 .ui-helper-hidden-accessible {
```
gipoezcan, line 12: Thank you for the quick PR!

As long as they aren't actually added by jQuery UI (I'm not sure if they are or not), I don't think you need to "unset" the values for border, clip, height, margin, overflow, padding and width. Also, I do think that you need to keep the position at "absolute" otherwise in our case at least, the layout was a little broken (they keep occupying space).

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.css
```diff
@@ -1,8 +1,6 @@
 .ui-helper-hidden-accessible input,
 .ui-helper-hidden-accessible select {
-    -webkit-transform: scale(0); 
-    -moz-transform: scale(0); 
-    transform: scale(0);
+    transform: unset;
```
gipoezcan: also here, IMHO you can delete the rule completely instead of unsetting it

melloware: I agree here.

primefaces/src/main/resources/META-INF/resources/primefaces/jquery/jquery.ui.pfextensions.css
melloware: Yes these are set by jQuery UI so we need to unset them as you did.

melloware:
@gipoezcan updated if you would like to re-review.

gipoezcan (approved):
Everything looks good to me! Thank you very much!

PR #12610, opened 2024-09-18
Fixed mismatched parentheses in JPALazyDataModel example

SkylerMime:
I believe the extra right parentheses after builder() are all incorrect, since every .method() should immediately follow.


melloware:
Thanks can you fix the same doc in the 15.0.0 folder too please?

FlipWarthog:
> Thanks can you fix the same doc in the 15.0.0 folder too please?

I already got those: https://github.com/primefaces/primefaces/commit/c304ebb6122c9a8f49e3858613fa9c58d8c94868

PR #12613, opened 2024-09-20
Fix #12157: DatePicker add default hour,minute,second,ms

### LINKED ISSUE #12157: DatePicker: Add `defaults` to match legacy Calendar
### Description

I am unable to migrate component **Calendar** to **DatePicker**. 

In **Calendar** I have the attributes **defaultHour, defaultMinute, defaultSecond, defaultMillise** that in case I don't have any value it sets the default time according to its values. 

In **DatePicker** these attributes do not exist and it sets the current time.

Add **defaultHour, defaultMinute, defaultSecond, defaultMillisec** attributes to **DatePicker** to perform the same functionality as the **Calendar** component.

### Additional context

_No response_

melloware:
Fix #12157: DatePicker add default hour,minute,second,ms


tandraschko:
shouldnt we make it `defaultMillisecond`? i dont like to have `defaultSecond` but `defaultMillisec`

melloware:
Agreed that is a better name, that is what the legacy Calendar was using but I will update it and the migration guide in this PR.

melloware:
Updated.

PR #12615, opened 2024-09-21
Fixes primefaces/primefaces#12614, include allSuffix in cache key for importEnum tag

### LINKED ISSUE #12614: ImportEnum: Fails when used with different allSufix
### Describe the bug

`<p:importEnum/>` fails when used with the same enum class, but different values for `allSuffix`. 

This happens because the [ImportEnumTagHandler](https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/java/org/primefaces/component/importenum/ImportEnumTagHandler.java#L118) only uses the enum class as a cache key, but not the configured `allSuffix`. So when multiple different suffixes are used, the cached map won't be correct.

We had this issue when different developers used different suffixes in different pages of our app. I since changed it so that it always uses the same `allSuffix`, but it should still work and not break.

### Reproducer

Replace the XHTML in https://www.primefaces.org/showcase/ui/misc/importEnum.xhtml with the following, then try opening the page again. It will fail with an error.

```xhtml
<div class="card">
    <p:importEnum type="jakarta.faces.application.ProjectStage" var="JsfProjectStages"
                  allSuffix="ALL_ENUM_VALUES"/>
    <p:importEnum type="jakarta.faces.application.ProjectStage" var="Stages"
                  allSuffix="ALL_STAGES"/>

    <h5 class="mt-0">Development</h5>
    <h:outputText value="#{JsfProjectStages.Development}" />

    <h5>ALL</h5>
    <ui:repeat var="current" value="#{JsfProjectStages.ALL_ENUM_VALUES}">
        <h:outputText value="#{current}" styleClass="mb-2 block"/>
    </ui:repeat>

    <h5>ALL_STAGES</h5>
    <p>You can use the same enum with different variable names and suffix for all enums, if necessary.</p>
    <ui:repeat var="current" value="#{Stages.ALL_STAGES}">
        <h:outputText value="#{current}" styleClass="mb-2 block"/>
    </ui:repeat>
</div>
```

### Expected behavior

No error is produced. The XHTML page output all enum values when using the configured `allSuffix`.

### PrimeFaces edition

None

### PrimeFaces version

15.0.0-SNAPSHOT

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4

### Java version

17

### Browser(s)

_No response_

blutorange:
Fixes #12614. As mentioned there, this includes the `allSuffix` in the cache key when getting the map with the enum constant for the `importEnum` tag.

I added another `importEnum` tag to the show case with another variable name and allSuffix. Without this fix, this will result in an error if the faces stages is not set to development.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,7 +29,7 @@ What is the first step to be taken?
 -----------------------------------
 
 First of all, **Discuss with the [project members](http://forum.primefaces.org/)** (a new thread should do) about
-your ideas: new features, fixes, documentation... whatever you would like to contribute to the project. Let we
+your ideas: new features, fixes, documentation... whatever you would like to contribute to the project. Let us
 discuss the possibilities with you so that we make sure your contribution goes in the right direction and aligns
 with the project's standards, intentions and roadmap.
 
--- a/docs/15_0_0/components/importenum.md
+++ b/docs/15_0_0/components/importenum.md
@@ -2,12 +2,11 @@
 
 In older EL versions (< 3.0), it's not possible to use enum constants or any other static
 fields/methods in an EL expression. As it is not really a good practice to create beans with
-getter/setter for each constants class, we provide an utils tag which allows to import enum values in
+getter/setter for each constants class, we provide a utils tag which allows to import enum values in
 a page.
 
 The enum values can be accessed via the name of the class (default setting) or via a custom name
-(var attribute). It also possible to get all enum values of the class with the "ALL_VALUES" suffix
-or a custom prefix via the "allSuffix" attribute.
+(var attribute). It is also possible to get all enum values of the class with the "ALL_VALUES" key.
 
 ## Info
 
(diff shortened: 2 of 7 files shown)


blutorange:
The integration test failures are `StaleElementReference` in various unrelated components such as `DatePicker` and `SelectOneMenu`; so I don't think they are related to these changes. Seems to be a general stability issue with the tests, since different tests fail in different runs and also seems to happen in different PRs.

tandraschko:
Why should we cache the allSuffix?
I think the allSuffix should never be in the cache, also not the default value. Just type and all values.

blutorange:
@tandraschko Currently, the all suffix is in the cache, causing the error. But if I understand your suggestion correctly, we only store the enum values in the cache. Then we get those from the cache and create a new map which includes the `allSuffix` and `DEFAULT_ALL_SUFFIX`? Something along the lines of:

```java
Map<Class<?>, Map<String, Object>> cache = 
  PrimeApplicationContext.getCurrentInstance(FacesContext.getCurrentInstance()).getEnumCacheMap();
Map<String, Object> enums = cache.get(cacheKey);

// Potentially replace with a slightly optimized implementation
// of the Map interface that does not require creating a new HashMap
Map<String, Object> allMap = new HashMap<>(enums);
allMap.put(DEFAULT_ALL_SUFFIX, enums.values());
allMap.put(allSuffix, enums.values());
return allMap;
```

The downside is of course having to create an additional map. I don't know what exactly the purpose of the cache is, but I guess it's because `Class#getEnumConstants` needs to create a new array each time it is called?

melloware:
As a sidenote the Integration test are randomly failing due to a bug in Chrome 128. We are waiting on Chrome 129 to hopefully fix it.

tandraschko:
The cache is just micro optimization.

We should even think about deprecate the allSuffix tag or say: if you dont use the default, we create a new map each time.

blutorange:
> We should even think about deprecate the allSuffix tag or say: if you dont use the default, we create a new map each time.

I like that idea. We can also do both: deprecate the `allSuffix` attribute and, if still used, skip the cache. Even now, `ALL_VALUES` is always set anyways, whether or not an `allSuffix` is set.

The only case I can think of where `allSuffix` would be necessary is when the enum contains an enum constant with the 
name `ALL_VALUES`. But that is rare; and since EL now supports method calls, one could always just use `Map#values` in that case:

```xml
<p:importEnum type="some.enum.MyEnum" var="MyEnum" />
<ui:repeat var="current" value="#{MyEnum.values()}" />
```

blutorange:
So anyways, I updated the PR so that the allSuffix is now deprecated. When used, a new hash map is created so that it works as expected. I also updated the docs regarding this.

tandraschko:
great PR!

PR #12616, opened 2024-09-22
CI - Java 23 (instead of 22)

christophs78:


christophs78:
There seem to be Lombok-issues with Java 23: https://github.com/projectlombok/lombok/issues/3752
(Does not affect PF (Core) but our ITs.)

melloware (approved):

PR #12628, opened 2024-09-25
Fix #12625: Improve poor performance of paginated treeTable with larg…

### LINKED ISSUE #12625: TreeTable: poor performance in paginated treeTable with large number of nodes
### Describe the bug

The **treeTable** component experiences significant performance degradation with large datasets when pagination is enabled. This is caused by an unbounded iteration of root nodes in the `processNode` method, which processes all nodes instead of restricting to nodes on the current page.

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

13.0

### Theme

_No response_

### JSF implementation

MyFaces

### JSF version

4.0

### Java version

21

### Browser(s)

_No response_

skybber:
I've tested it on v13.0.9

Fix #12625

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/treetable/TreeTable.java
+++ b/primefaces/src/main/java/org/primefaces/component/treetable/TreeTable.java
@@ -588,4 +588,33 @@ public class TreeTable extends TreeTableBase {
         return getFacesContext().getCurrentPhaseId() == PhaseId.RENDER_RESPONSE
                 && (!ComponentUtils.isNestedWithinIterator(this) || columns.stream().noneMatch(DynamicColumn.class::isInstance));
     }
+
+    @Override
+    protected void processNode(FacesContext context, PhaseId phaseId, TreeNode root, TreeNode treeNode, String rowKey) {
+        if (!isPaginator() || root != treeNode) {
+            super.processNode(context, phaseId, root, treeNode, rowKey);
+        }
+        else {
+            if (treeNode != null && shouldVisitNode(treeNode) && treeNode.getChildCount() > 0) {
+                int first = getFirst();
+                int rows = getRows() == 0 ? getRowCount() : getRows();
+
+                processColumnChildren(context, phaseId, root, rowKey);
+
+                List<TreeNode> children = root.getChildren();
+                int childCount = root.getChildCount();
+                int last = (first + rows);
+                if (last > childCount) {
+                    last = childCount;
+                }
+
+                for (int i = first; i < last; i++) {
+                    TreeNode child = children.get(i);
+                    String childRowKey = rowKey == null ? String.valueOf(i) : rowKey + SEPARATOR + i;
+                    processNode(context, phaseId, root, child, childRowKey);
+                }
+            }
+        }
+    }
+
 }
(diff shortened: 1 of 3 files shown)


melloware (approved):

melloware:
This looks good to me but want @Rapster to review also.

skybber:
Perhaps it would be better to have that iteration in UITree, where isPaginator() would always return false. This way, the patch code is dependent on the implementation in the superclass.

melloware:
Well I dont think UITree can ever be paginated its only for TreeTable so i think its OK where it is?

skybber:
If UITree is never going to be paginated, it would be better to leave it as it is and keep the patch in its current form, as you are writing.

primefaces/src/main/java/org/primefaces/component/treetable/TreeTableBase.java
```diff
@@ -526,4 +530,32 @@ public String getExportTag() {
     public void setExportTag(String exportTag) {
         getStateHelper().put(PropertyKeys.exportTag, exportTag);
     }
-}
+
+    protected void processNode(FacesContext context, PhaseId phaseId, TreeNode root, TreeNode treeNode, String rowKey) {
```
melloware:
```suggestion
    @Override
    protected void processNode(FacesContext context, PhaseId phaseId, TreeNode root, TreeNode treeNode, String rowKey) {
```

melloware (changes requested):

melloware (approved):

Rapster:
Should be in TreeTable

skybber:
@melloware @Rapster , moved to TreeTable

primefaces/src/main/java/org/primefaces/component/treetable/TreeTableBase.java
```diff
@@ -25,6 +25,10 @@
 
 import javax.el.MethodExpression;
 import javax.faces.component.behavior.ClientBehaviorHolder;
+import javax.faces.context.FacesContext;
+import javax.faces.event.PhaseId;
+
+import java.util.List;
```
Rapster: To revert

skybber:
@Rapster done

melloware (approved):

primefaces/src/main/java/org/primefaces/component/treetable/TreeTable.java
```diff
@@ -588,4 +588,33 @@ protected boolean isCacheableColumns(List<UIColumn> columns) {
...
+                List<TreeNode> children = root.getChildren();
+                int childCount = root.getChildCount();
+                int last = (first + rows);
+                if (last > childCount) {
+                    last = childCount;
+                }
+
+                for (int i = first; i < last; i++) {
+                    TreeNode child = children.get(i);
+                    String childRowKey = rowKey == null ? String.valueOf(i) : rowKey + SEPARATOR + i;
```
Rapster: Can we introduce an util method to create the childRowKey, I saw it several times in UITree already

melloware: I assume you want this util method in the UITree itself?

Rapster: Yes

melloware (approved):

Rapster (approved):

PR #12652, opened 2024-09-29
Fix #12651: Datatable/TreeTable add scrollBarPosition attribute

### LINKED ISSUE #12651: DataTable/TreeTable: allow horizontal scrollbar to the top
### Description

Request from a client and also found past requests on the PF Forum.

https://stackoverflow.com/questions/33565238/move-scrollbar-to-top-in-datatable-primefaces

A new `scrollBarPosition="top"` attribute to change from the default `"bottom"`

### Additional context

image

melloware:
Fix #12651: Datatable/TreeTable add scrollBarPosition attribute

### DIFF
--- a/docs/15_0_0/components/datatable.md
+++ b/docs/15_0_0/components/datatable.md
@@ -90,6 +90,7 @@ DataTable displays data in tabular format.
 | scrollRows                | 0                  | Integer          | Number of rows to load on live scroll.
 | scrollWidth               | null               | Integer          | Scroll viewport width.
 | scrollable                | false              | Boolean          | Makes data scrollable with fixed header.
+| scrollBarPosition         | bottom             | String           | Position of the horizontal scrollbar, valid values are "bottom" (default) and "top".
 | selectAllFilteredOnly     | false              | Boolean          | When enabled, toggle select will only apply on filtered items. Default is false.
 | selection                 | null               | Object           | Reference to the selection data.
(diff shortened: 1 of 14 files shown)


tandraschko:
is it really only the small css change?

i wonder if this must be confirugable on component level or if we should just add it to the docs like:
if you wanna have the scrollbar on top, please add this css.

i suggest this because the normal UI/UX is to have it at the bottom and this is something very special.

melloware:
Yes it is CSS but similar to how we offer `rtl="true"` to flip the scrollbar to the far left instead of right i found both on Stack Overflow and the old PF forums this feature has been asked for many times. and I recently had a client ask for this behavior so it felt right to have this controlled by PF with an attribute?  Plus if the PF Datatable structure changes for any reason in the future we have the CSS to update.  Those were my thoughts.

tandraschko:
IMO RTL was bad decision to control this on component behavior
you dont want to say: component 1 rtl, component 2 rtl
thats a mechanism to be configured for the whole view, or the whole application

and thats the same for this feature
just because we mad the mistake once in the past, we can still develop further

how we wanna configure such stuff, which is not on component level, more on view or application?

melloware:
I don't disagree.  This feature came out of my client can possibly have a large number of rows on the screen at once and they don't like scrolling all the way to the bottom of the screen to scroll to the right. 😄

tandraschko:
maybe we can collect some such configs, which only makes sense on app/view level and check their commonalities first?

tandraschko:
closing this for now; can you make a discussion issue about such configs and collect similar stuff?

PR #12670, opened 2024-10-02
Fix #12666: Datatable: filtering with BigDecimal

### LINKED ISSUE #12666: Datatable: filtering with BigDecimal
### Describe the bug

If  a column in a dataTable is of type BigDecimal, you can't filter it easily because you have to type the exact number of decimals. One should use `compareTo` instead of `equals` with BigDecimal.

### Reproducer

To test it you can use the showcase:

Add a property to `org.primefaces.showcase.domain.Customer` such as:
`private BigDecimal price = new BigDecimal("4.00")`.

In `ui/data/datatable/filter.xhtml` add the column:
```
<p:column field="price" filterMatchMode="exact">
     <h:outputText value="#{customer.price}" >
           <f:convertNumber minFractionDigits="0" maxFractionDigits="2"/>
      </h:outputText>
</p:column>
```

### Expected behavior

If I type 4, the rows with price 4 should appear.

### PrimeFaces edition

None

### PrimeFaces version

15.0.0-SNAPSHOT

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

17

### Browser(s)

_No response_

FlipWarthog:
Fix #12666

As mentioned in https://github.com/primefaces/primefaces/issues/12666#issuecomment-2389177712 I think the implementation of this filter constraint was not entirely correct. With this rework, it now properly uses `Object.equals` rather than a `String` conversion comparison, like `ExactFilterConstraint` should. It also makes use of `compareTo`, where available, to address the original issue with BigDecimal (and any other similar cases) by setting the filter with:

```xml
<p:column field="price" filterMatchMode="equals" converter="jakarta.faces.BigDecimal">
    <h:outputText value="#{customer.price}" >
        <f:convertNumber minFractionDigits="0" maxFractionDigits="2"/>
    </h:outputText>
</p:column>
```

@melloware @Rapster take a look and let me know what you think.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/model/filter/EqualsFilterConstraint.java
+++ b/primefaces/src/main/java/org/primefaces/model/filter/EqualsFilterConstraint.java
@@ -25,12 +25,12 @@ package org.primefaces.model.filter;
 
 import java.util.function.BiPredicate;
 
-public class EqualsFilterConstraint extends StringFilterConstraint {
+public class EqualsFilterConstraint extends ComparableFilterConstraint {
 
     private static final long serialVersionUID = 1L;
 
     @Override
-    protected BiPredicate<String, String> getPredicate() {
-        return String::equals;
+    protected BiPredicate<Comparable, Comparable> getPredicate() {
+        return (o1, o2) -> o1.compareTo(o2) == 0;
     }
 }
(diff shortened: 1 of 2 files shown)


melloware:
Ok we will have to look into the integration test failure

melloware (approved):

melloware:
@Rapster i was planning on merging this unless you object?

PR #12693, opened 2024-10-08
Fix typo in commandLink.html

SkylerMime:


melloware:
Ahhh this is generated. IT needs to be fixed also in `primefaces.taglib.xml`

melloware:
Ahh nevermind its already fixed I just haven't updated the VDL doc.

PR #12696, opened 2024-10-08
Fix #12694: AjaxStatus handle Faces completion

### LINKED ISSUE #12694: SelectOneMenu: ajaxStatus does not go back to status "complete"
### Describe the bug

Hi, 
Within a form containing a selectOneMenu with a listener on ajax event "change", when an item is selected in the select menu,  component p:ajaxStatus switches to state "start" but never finishes to the state "complete".
In my case, the spinner displayed by p:ajaxStatus does not disappear, even though it should after an item is selected from selectOneMenu 

### Reproducer

```
<!DOCTYPE html>
      xmlns:h="jakarta.faces.html"
      xmlns:f="jakarta.faces.core"
<h:head/>

<h:body styleClass="main-body">
        <p:ajaxStatus styleClass="block mb-4">
            <f:facet name="default">
                <h:outputText value="Status: StandBy"/>
            </f:facet>

            <f:facet name="start">
                <i class="pi pi-spin pi-spinner loading-icon" aria-hidden="true"></i>
            </f:facet>

            <f:facet name="complete">
                <h:outputText value="Status: Completed"/>
            </f:facet>
        </p:ajaxStatus>

        <h:form id="simpleForm">

            <p:selectOneMenu value="#{myForm.valueHolder}">
                <f:selectItem itemLabel="label1" itemValue="value1"/>
                <f:selectItem itemLabel="label2" itemValue="value2"/>
                <f:ajax event="itemSelect"
                        listener="#{myForm.onChange()}"
                        render=":simpleForm"/>
            </p:selectOneMenu>
        </h:form>
</h:body>
</html>

```

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

14.0.5

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

21

### Browser(s)

_No response_

melloware:
Fix #12694: AjaxStatus handle Faces completion

Faces has 1 arg while PF has 3 args.

The code to shortcircuit on redirect was assuming always 3 args.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/ajaxstatus/ajaxstatus.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/ajaxstatus/ajaxstatus.js
@@ -122,23 +122,19 @@ PrimeFaces.widget.AjaxStatus = PrimeFaces.widget.BaseWidget.extend({
                     }, $this.cfg.delay);
                 }
                 else if(data.status === 'complete') {
-
+                    // ignore PF complete event when JSF success/error event is fired right after
                 }
                 else if(data.status === 'success') {
-                    if($this.timeout) {
-                        $this.deleteTimeout();
-                    }
+                    $this.deleteTimeout();
                     $this.trigger('success', arguments);
-                    $this.trigger('complete', arguments);
+                    $this.trigger('facesComplete', arguments);
                 }
             });
 
             jsf.ajax.addOnError(function(data) {
-                if($this.timeout) {
-                    $this.deleteTimeout();
-                }
+                $this.deleteTimeout();
                 $this.trigger('error', arguments);
-                $this.trigger('complete', arguments);
+                $this.trigger('facesComplete', arguments);
             });
         }
     },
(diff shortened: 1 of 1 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/ajaxstatus/ajaxstatus.js
```diff
@@ -189,10 +189,11 @@ PrimeFaces.widget.AjaxStatus = PrimeFaces.widget.BaseWidget.extend({
                 break;
 
             case 'complete':
+                var isFaces = args.length === 1;
```
tandraschko: Please add a comment here 

Maybe there is a more safer check possible?

melloware: Let me investigate since they both have at least 1 arg I can look for something in arg[0] to detect JSF from PF

tandraschko:
maybe its easier to check if its a PF request and then do `!isPF` as mojarra and myfaces might be different

melloware:
OK updated how does that look?

tandraschko:
hmm, for it still feels hacky
maybe we can introduce 2 different events? `.trigger('complete')` and `.trigger('facesComplete')` 

then its quite clear

melloware:
OK that would be more clear.

melloware:
OK updated to have `facesComplete` event

tandraschko:
looks good for me!

melloware:
@tandraschko i think my original solution while hacky solved all issues  Now the `complete` is closing the facet but not firing the `oncomplete` JS code.

PR #12700, opened 2024-10-10
Update POI

tandraschko:


tandraschko:
the problem is that we declare our shaded dependencies as dependencyManagement and overwrite therfore version from POI and so on

tandraschko:
it may also be that we dont need to workaround anymore in showcase/integration-tests for intellij as i get:

grafik

melloware (approved):

melloware:
Excellent!

PR #12702, opened 2024-10-10
14.0.6 POI 5.3.0 and Proper shading of classes

melloware:
@tandraschko i backported this to 14 only because I am running into this same conflict with Quarkus POI now

https://github.com/quarkiverse/quarkus-primefaces/actions/runs/11268755147/job/31336077122

I left the POI version in 14 at 5.2.5 since that is what it was this just fixes the shading issues.  Thoughts?


melloware:
I am going to upgrade to 5.3.0 and update docs for 14 and 15.

Issue #6868, opened 2021-01-20
TimeLine: Using CSP breaks JS


melloware:
**Describe the defect**
Using timeline with CSP=true is failing Integration Tests and this was proven locally.

![image](https://user-images.githubusercontent.com/4399574/105234036-a3d6ed80-5b38-11eb-9564-c05dab9a425c.png)

**Environment:**
 - PF Version: _10.0.0-rc1_

This is because the "standalone" vis-timeline build must be used.

See: https://github.com/visjs/vis-timeline#standalone-build

melloware:
CC @sqores

melloware:
For **charlietran1986**

Download and unzip this:
[pf-csp.zip](https://github.com/primefaces/primefaces/files/5852029/pf-csp.zip)

Run `mvn clean jetty:run` and go to http://localhost:8080/primefaces-test/test.xhtml

charlie1986:
I have created new Enable command button to enable back input text, but the browser console still throws such error.
[pf-csp.zip](https://github.com/primefaces/primefaces/files/5862549/pf-csp.zip)

melloware:
Just tried your reproducer. Its because there is no "enable" script in test.js.  So there is no function to call unless I am not seeing it?

melloware:
Ahh you are disabling too much. You can't disable hidden inputs like ViewState or you break JSP. I tweaked your example so now it works...
[pf-csp-working.zip](https://github.com/primefaces/primefaces/files/5862608/pf-csp-working.zip)

PR #12717, opened 2024-10-13
Convert widgets to classes + switch to esbuild / ESM + load packages from NPM

### LINKED ISSUE #11539: TypeDoc Updates
@blutorange can you take a pass at updating the typedoc `package.json`?  I am getting warnings from libraries that don't support Node 20.  I am getting a warning from Coloris.  I would like to update all this before the 14.0.0 release.

## Install Warnings
```
PS C:\dev\primetek\primefaces\primefaces\src\main\type-definitions> npm install
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: '@es-joy/jsdoccomment@0.10.8',
npm WARN EBADENGINE   required: { node: '^12 || ^14 || ^16' },
npm WARN EBADENGINE   current: { node: 'v20.9.0', npm: '10.1.0' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'eslint-plugin-jsdoc@36.1.0',
npm WARN EBADENGINE   required: { node: '^12 || ^14 || ^16' },
npm WARN EBADENGINE   current: { node: 'v20.9.0', npm: '10.1.0' }
npm WARN EBADENGINE }

```

## Build Warnings

```
[INFO] Warning: Specifying multiple comments at the start of a file to use the first comment as the comment for the module has been deprecated. Use @module or @packageDocumentation instead.
```

> [!NOTE]  
> This is caused by AutoNumeric index.d.ts having two different comments at the top of the file: https://github.com/autoNumeric/autoNumeric/blob/next/index.d.ts

### LINKED ISSUE #12500: Reworking JavaScript documentation & TypeScript declarations
We document the JavaScript code with doc comments, then generate TypeScript declarations from that, and finally use the declarations to generate an API documentation.

That works well, but the current setup is sub-optimal and needs to be improved.

Why?

* We currently use a custom script to extract TypeScript declarations from the JavaScript works -- but it's complex. It also contains a non-trivial amount of custom, non-standard JSDoc syntax.
* Take a look at the readme and the source code here: https://github.com/primefaces/primefaces/tree/master/primefaces/src/main/type-definitions
* It uses TypeScript in a way that may break when TypeScript changes their API, and that may be hard to fix. See also #11539. I tried
  updating to the latest TypeScript version, which proved to be harder than expected due to TypeScript API changes.

The custom script was necessary to work with the existing JavaScript code structure. I evaluated some tools to generated TypeScript declarations from the existing JavaScript code structure, but none were up to the task. But now it's 2024, IE support is gone, so I think we should take a look at better approaches.

While I think we can mostly keep everything backwards-compatible, there are bound to be a few edge cases. So it should only be done as part of a new major version.

I haven't worked out all details yet, but the general approach would be to completely remove the custom documentation script; make some adjustment to the JavaScript files and then use TypeScript to generate the type declaration files. Finally, use the declaration files to generate the documentation, the same we are doing it now.

What do you think about it?

See below for details. **My two main questions:**

* Are we OK with shipping JavaScript code that contains some ES6 features, such as e.g. classes? Or do we need to support browsers that don't support these (are there even any such browsers anymore)?
* Should we keep the code in the PrimeFaces repo or use NPM to fetch the dependencies?

## Fix the JavaScript code structure

The current structure of the JavaScript code is hard to understand for tools. In particular:

* Using functions as classes and mutating prototypes
  * In fact, we use a custom class "polyfill" called [Simple JavaScript Inheritance by By John Resig](https://github.com/primefaces/primefaces/blob/2e7aad32e0a01de041976ecdec0ae907f2228ef4/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.widget.js#L6-L63) That's even harder to analyze.
* Assigning to the global scope everywhere. Most tools understand modules better than a bunch of global modifying script files. 

The first point is the most important one. All widgets are intended to be classes, but it's hard for tools like TypeScript to infer the structure of the intended class from functions and prototypes and the custom class polyfill we currently use.

Instead, we should just use a class directly, JavaScript supports those for almost 10 years now. That raises my first question:

* **Question** Are we OK with shipping JavaScript code that contains ES6 classes? If not, we need to add a production build step that downlevels JavaScript back to function prototypes. There are tools like [esbuild](https://esbuild.github.io/) that can do this extremly quickly; and the PrimeFaces resources optimizer plugin uses closure compiler (I believe), which should also be able to do this.

Secondly, I would use ES modules (`import` / `export`) to make dependencies explicit. I'll have to experiment with that a bit, especially how it interacts with quick dev builds and starting a dev server. For development, all browsers support ES modules natively, but the Jakarta Faces resource servlet mechanism does not play nicely with ES modules.

For example, [forms.inputtext.js](https://github.com/primefaces/primefaces/blob/2e7aad32e0a01de041976ecdec0ae907f2228ef4/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.inputtext.js) would be changed to something like this. Note that the actual code remains the same -- only the structure is changed (the code should be changed if we want TypeScript to understand it for type checking, but as I said, that can be done later, if we want):

```js
// forms.inputtext.js

// @ts-nocheck

/** @import {PartialWidgetCfg} from "../core/core.widget.js"; */
/** @import {InputTextCfg} from "./forms.inputtext.types.js"; */

import * as PrimeFaces from "../core/core.js"
import * as utils from "../core/core.utils.js"
import { BaseWidget } from "../core/core.widget.js"

/**
 * __PrimeFaces InputText Widget__
 *
 * InputText is an extension to standard inputText with skinning capabilities.
 *
 * @template {InputTextCfg} TCfg
 */
export class InputText extends BaseWidget {
  /**
   * @param {PartialWidgetCfg<TCfg>} cfg
   */
  constructor(cfg) {
    super(cfg);
    PrimeFaces.skinInput(this.jq);

    /**
     * The DOM element for the counter that informs the user about the number of
     * characters they can still enter before they reach the limit.
     * @type {JQuery}
     */
    this.counter;

    //Counter
    if (this.cfg.counter) {
      this.counter = this.cfg.counter ? $(PrimeFaces.escapeClientId(this.cfg.counter)) : null;
      this.cfg.counterTemplate = this.cfg.counterTemplate || '{0}';
      this.updateCounter();

      if (this.counter) {
        var $this = this;
        this.jq.on('input.inputtext-counter', function (e) {
          $this.updateCounter();
        });
      }
    }
  }

  /**
   * Disables this input so that the user cannot enter a value anymore.
   */
  disable() {
    utils.disableInputWidget(this.jq);
  }

  /**
   * Enables this input so that the user can enter a value.
   */
  enable() {
    utils.enableInputWidget(this.jq);
  }

  /**
   * Updates the counter value that keeps count of how many more characters the user can enter before they reach the
   * limit.
   * @private
   */
  updateCounter() {
    var value = this.normalizeNewlines(this.jq.val()),
      length = this.cfg.countBytesAsChars ? PrimeFaces.utils.countBytes(value) : value.length;

    if (this.counter && this.cfg.maxlength) {
      var remaining = this.cfg.maxlength - length;
      if (remaining < 0) {
        remaining = 0;
      }

      var counterText = this.cfg.counterTemplate
        .replace('{0}', remaining)
        .replace('{1}', length)
        .replace('{2}', this.cfg.maxlength);

      this.counter.text(counterText);
    }
  }

  /**
   * Replaces all line breaks with a Window-style line break (carriage return + line feed).
   * @private
   * @param {string} text Text to normalize.
   * @return {string} The given text, with all line breaks replaced with carriage return + line feed.
   */
  normalizeNewlines(text) {
    return text.replace(/(\r\n|\r|\n)/g, '\r\n');
  }
}
```

In addition, we'd have a separate `*.d.ts` file with types need by the widget. It does not have any effect on the runtime. Using a separate file
gets rid of [having to pollute the header with long doc comments.](https://github.com/primefaces/primefaces/blob/2e7aad32e0a01de041976ecdec0ae907f2228ef4/primefaces/src/main/resources/META-INF/resources/primefaces/dialog/dialog.js#L7-L88)

```ts
// forms.inputtext.types.d.ts

/**
 * The configuration for the {@link  InputText| InputText widget}.
 * You can access this configuration via
 * {@link PrimeFaces.widget.BaseWidget.cfg|BaseWidget.cfg}. Please note that this
 * configuration is usually meant to be read-only and should not be modified.
 */
export interface InputTextCfg {
    /**
     * ID of the label component to display remaining and entered characters.
     */
    counter: string;
    /**
     * Template text to display in counter, default value is `{0}`.
     */
    counterTemplate: string;
}
```

## Type annotations via JSDocs

This is not so much a change, but rather a by-product we'd get for free.

TypeScript allows two ways how JavaScript code can be annotated with types and be type checked:

* The usual TypeScript *.ts files, which extend the grammar / syntax of Javascript. Type annotations are added inline after a colon `:`.
* [Type annotation](https://www.puruvj.dev/blog/get-to-know-typescript--using-typescript-without-typescript) [via JSDocs](https://alexharri.com/blog/jsdoc-as-an-alternative-typescript-syntax). This is basically what we already have -- types inside doc comments, e.g. see [forms.button.js](https://github.com/primefaces/primefaces/blob/2e7aad32e0a01de041976ecdec0ae907f2228ef4/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.button.js#L18)

Recently, [there were some](https://www.codestackr.com/blog/typescript-vs-jsdoc-a-shift-in-the-development-landscape) major project that switched from TypeScript files back to plain JavaScript files, but are still type-checked via type annotation in JSDocs (some news articles called that "removing TypeScript", but [that's misleading at best](https://www.reddit.com/r/sveltejs/comments/12cyady/are_you_leaving_svelte_cause_they_are_ditching/)).  A major example would be [Svelte](https://github.com/sveltejs/svelte/pull/8569). One major [advantage](https://github.com/sveltejs/kit/discussions/4429#discussioncomment-2423814) of this approach is that it removes the required compile step from TypeScript to JavaScript. I also experimented with this at work and it works pretty well.

In theory, this means that we'd bascially just have to set a few flags when running TypeScript (to generate the type declarations for the documentation); and would have fully type checked code.

In practice, I'm certain there will be many type errors. Fortunately, you can just add `// @ts-nocheck` to the top of a file, which tells TypeScript not to fail on type errors in that file. This way, we can add type checking on a gradual basis, e.g. when editing a file anyways.

See also TypeScript's [JSDoc reference](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#constructor). One major drawback was that
importing types from other files was quite painful to write, but this was also recently remediated by support [for @import via a doc comment.](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#the-jsdoc-import-tag)

A quick example how it looks like (again, note that this is what we already have!)

```js
/** @import {Integer} from "./helper-types" */

/**
 * @param {Integer} x 
 * @param {Integer} y 
 * @returns {Integer}
 */
export function sum(x, y) {
    return x + y;
}

/**
 * @param {Integer} x 
 * @param {Integer} y 
 * @returns {Integer}
 */
export function multiply(x,y ) {
    return x* y;
}
```

## Reconsider whether we want to inline 3rd party code

This is optional, but I though I'd bring it up and ask for your opinion.

Currently, all 3rd party code is added directly to the PrimeFaces repo, see e.g. [signature](https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/resources/META-INF/resources/primefaces/signature/0-signature.js).

We can keep it this way, but the alternative would be to use NPM, i.e. just declare the needed 3rd party libraries in a `package.json`. For development, there are several options such as building a "vendor bundle" that contains all code from these 3rd party libraries. You would only have to build it once, to speed up development.

* **Question** What do you think, should we keep the code in the PrimeFaces repo or use NPM to fetch the dependencies?

blutorange:
Draft, see below, don't merge yet.

As we discussed in #12500, it's probably better to take some small steps first. Here I 

* convert all widgets to native classes
* remove the custom build script that generated the type declarations and TypeDocs. It won't work with classes as is, and when we're done, it wouldn't be needed anymore anyway. The previously generated TypeDocs are still available, and I don't think a new release is schedule soon?
* (edit: ) change the build system from closure compiler to esbuild

---

I did come across one issue with closure compiler: Some JavaDoc comments currently include an `@interface` tag that closure compiler appears to parse, resulting in wrong non-working output when using classes (it doesn't call `super` in the constructor).

There are probably ways around that, but I wanted to bring up another point. Eventually, we're going to convert (some or all) widgets to TypeScript files. Closure compiler can't work with TypeScript, so we'll need to run another tool. Also, TypeScript always checks types and you might not want or need that for a quick dev build.

So I made an experiment in this branch: [switch-to-esbuild](https://github.com/blutorange/primefaces/tree/switch-to-esbuild/primefaces/src/main/resources/META-INF/resources/test-esbuild).

See also the link above with a readme and more details, but in short: [esbuild](https://esbuild.github.io/) is _really_ fast (less than a second on my machine for all JS+CSS files including minification), works well with TypeScript files,  and there's not different in bundle sizes compared to closure compiler / yui-compressor.

What are you thoughts?

Fix #12500
Fix #11539


melloware:
Looks good!  A couple of things and this is just observations not anything to change.

1. This will break any existing MonkeyPatches. I can live with this as long as we provide a guide on how to update MonkeyPatches.
2. The Integration Tests are failing (which was probably to be expected.

blutorange:
Thanks for the feedback! Do you (or anybody else) have any thoughts on using `esbuild` to process CSS/JS (and soon TS) files? See the post above and also the [readme here](https://github.com/blutorange/primefaces/tree/switch-to-esbuild/primefaces/src/main/resources/META-INF/resources/test-esbuild), it seems quite promising to me. Supports TypeScript, much faster than closure compiler, and same or smaller file size when minified.

---

> This will break any existing MonkeyPatches. I can live with this as long as we provide a guide on how to update MonkeyPatches.

You mean something like `PrimeFaces.widget.SelectCheckboxMenu.prototype = function() {...}` ? This should still work (to some extent), e.g.:

```js
class Foo {
    foo() {return 9}
}
Foo.prototype.foo = function(){return 12}
```

It's possible to configure transpilers to downlevel class syntax to function syntax, but perhaps it is time to move on.

> The Integration Tests are failing (which was probably to be expected.

Interestingly, only some variations of Java / JSF are failing, but I'll take a look at that later.

melloware:
My +1 on esbuild if its faster and smaller.

tandraschko:
looks good for me

blutorange:
Ok, then I'll take a look if the idea with esbuild works out and update the PR. That would solve the current issue that closure compiler outputs the wrong code due to JSDoc comments.

blutorange:
I'll need to do some more manual testing and take a look at the automated tests, but it's basically done now.

A quick summary of the additional changes:

* The build now uses [ESBuild](https://esbuild.github.io/) to bundle and minify the files. Takes around 1s on my machine, regardless of whether minifiy is enabled.
* We can now use [ESM](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules), i.e. `import / export`. This is used to import modules from NPM. The code structure could be refactored more, but that might be part of another PR -- I do not want to do too much in a single PR.
* Resources such as fonts and images in CSS files can now be referenced normally via relative paths (`./images/loader.gif`) in CSS flies. I added an ESBuild plugin that converts the URLs to EL resource expression (`#{resource['primefaces:fileupload/images/loader.gif']}`).
* All libraries that are available on NPM are now loaded via NPM. This:
  * reduces the amount of code in our repo
  * makes it easier to know which version we are using
  * allows dependabot to bump versions
  * makes it easy to see at a glacne  which libraries we are using
  * could be used to generate a `thirdpartylibs` file
* For most libraries, we did not make any changes (which I checked  via checksum / diff tool). Where we did make changes, there's now a folder [src/main/frontend/.yarn/patches](https://github.com/blutorange/primefaces/tree/esbuild/primefaces/src/main/frontend/.yarn/patches) with a patch file for each such package. This makes it obvious which modifications we made, and makes updating easier. The patch files are applied automatically during package install.
* I moved all frontend source files to `src/main/frontend`, which now contains the `package.json`. During the build, the bundled files written to `src/main/generated-resources`. Personally, I don't like putting files in `src/main/resources` that aren't actually supposed to be included as resource files in the JAR. The actual resources are generated, and that's what `src/main/generated-resources` is for.

The goal is to convert the files to TypeScript and generate documentation from that. This will need some more PRs. My general idea currently is:

* refactor the code a bit more to take advantage of `import` / `export`, to make it easier for tools like [typedoc](https://typedoc.org/). Especially the `core/*` files
* Move all type annotation from comments into the code, thereby turning the files into TypeScript files. I'm hoping this can be done automatically with a script. There will be type error, but we can ignore those and fix them later.
* Take a look at publishing the package and using typedoc to generate documentation.

melloware:
Ok I will start reviewing.

melloware:
Would it be possible to remove all #sourceMapping URLs from all files?  I typically do this as it's just noise in the final about not find the source etc

Also I usually fix comment so they get removed from the final build even if they have  bang ! In the comment. It also reduces a lot of the size

blutorange:
Thanks for taking a look!

> Would it be possible to remove all #sourceMapping URLs from all files? I typically do this as it's just noise in the final about not find the source etc

It should dot it. Did you see any source mapping URLs? When you build with `-P minify`, the bundler is set not to output any source mapping URLs. It only does it for the development build, where you want that for debugging. In fact, in development mode it adds inline source maps everywhere, which makes debugging easier.

> Also I usually fix comment so they get removed from the final build even if they have bang ! In the comment. It also reduces a lot of the size

I've also set the bundler not to output any legal comments in production mode (`-P minify`). So we don't have to do anything manually anymore. (It currently outputs a separare `*.js.LEGAL.txt` file, but that could also be disabled).

blutorange:
For reference and as a note to myself, there is one more issue I need to solve before this can get merged:

* Assume you have a very simple page with e.g. only `<p:outputPanel>`. Now, only the `components.js` is loaded.
* However, `components.js` imports e.g. the package `autosize`.
* `autosize` is provided by `jquery-plugins.js`, which exposes it to the global scope and `components.js` takes it from the global scope.
* Since the binding of imports happens when a file is loaded, `components.js` get an empty object for `autosize`.
* When the page now makes an AJAX update and adds the `<p:inputTextArea>` component, `jquery-plugins.js` gets loaded as a `ResourceDependency` of `InputTextArea`.
* That's too late however, the newly added `autosize` won't be available to the already loaded `components.js`

melloware:
@blutorange 
1. That is great about sourcemapping and comments that is exactly what I want!  I didn't test this PR i was just thinking out loud.
2. As for `<p:outputPanel>` I guess I am not quite clear as it has this...


```java
@ResourceDependency(library = "primefaces", name = "components.css")
@ResourceDependency(library = "primefaces", name = "jquery/jquery.js")
@ResourceDependency(library = "primefaces", name = "jquery/jquery-plugins.js")
@ResourceDependency(library = "primefaces", name = "core.js")
@ResourceDependency(library = "primefaces", name = "components.js")
public class OutputPanel extends OutputPanelBase {
```

So why is `jquery-plugin.js` not loaded even if I have a single `<p:outputPanel>` on the screen?

blutorange:
> That is great about sourcemapping and comments that is exactly what I want! I didn't test this PR i was just thinking out loud.
>
> a separate *.js.LEGAL.txt is great leave it in!

: )

> As for <p:outputPanel> I guess I am not quite clear as it has this...

That was more a todo for myself.

But basically, I wanted to be able to use import syntax. So the InputTextArea file now looks something like this:

```js
// forms.inputtextarea.js
import autosize from "autosize";

class InputTextArea {
  render() {
    autosize(...);
  }
}
```

The textarea component is in `components.js` and `autosize` is in `jquery-plugins.js`. So these two files need to be linked together somehow.

The way esbuild bundles is that `jquery-plugins.js` adds autosize to the global scope:

> window.autosize = ...

And the inputtextarea reads it from there, but like this:

```js
// It accesses the import when loaded, not when it first uses the import
const autosize  = window.autosize;

class InputTextArea {
  render() {
    autosize(...);
  }
}
```

And that fails, because `window.autosize` isn't defined when `components.js` is loaded.

---

One way to solve it is of course to not use `import` and just manually add autosize to a global variable; and the access that variable when the text area is rendered. I'm just thinking if I can find a better way.

That's kinda the hardest part though, dependencies between different files are very implicit right now. We need to manually define which final JavaScript should contain which libraries, and which final JavaScript files should simply try and access those libraries from `window`.

melloware:
OK now I understand

melloware:
should all depenedencies in `package.json` not use the hat `^` technique and declare exact versions.  I think it will make things clearer and also make dependabot checks clearer and easier.  With a project this large the `^` notation scares me.

blutorange:
> should all depenedencies in package.json not use the hat ^ technique and declare exact versions. I think it will make things clearer and also make dependabot checks clearer and easier. With a project this large the ^ notation scares me.

That's a good point, actually. It's not quite as simple as that, but basically, the versions are now fixed. Details if you're interested:

* One difference between Java/Maven and Node/NPM is that Java can only load one instance of a class, whereas there is no such restriction for NPM. So Maven must resolve everything to one exact version. NPM can give different packages different versions of a dependency. For example, library A could request `jquery@3.7.1` and another library B could request `jquery@3.5.2`.
* Having too many versions has some disadvantages such as increasing download times and the bundle size in general. It's also problematic for packages that are not side-effect free (e.g. jquery, which registers itself on the global scope).
* This is why packages should declare a range of versions for their dependencies, such as `~1.0.0`, `^1.0.0`, or `1.0.0 - 1.10.10`.
* Each package manager has a lock file that records the actual versions being used so that every install uses the same version.

What I did:

* Add a build-time check that checks for duplicate packages and fails if there are any.
* Use a fixed version for all dependencies where possible (and it was possible for all but jquery)
* For jquery, we want `4.0.0-beta.2`, but most packages still only accept `3.x`. I added a `resolutions` field to the `package.json` which pins the version to `4.0.0-beta.2`.

blutorange:
> everything is implicit because we use the global scope for 15 years.

I have been thinking about what a modern solution would look like. It would be hard / a lot of work to do right now, but I'd be interested of what you think of the general idea.

I assume you have heard of code splitting? We could create just a single entry point `index.js` that imports all the required modules. For those parts that should be loaded only when needed, we can use dynamic import expressions `await import("./path/to/module.js")`. Bundlers can then split that code into multiple JavaScript files.

For PrimeFaces, what should be loaded dynamically are mainly the various widget that need external dependencies. [Here's an example](https://github.com/blutorange/primefaces/blob/demo-code-splitting/primefaces/src/main/frontend/src/core/core.widget.registry.ts#L618-L744) of how that could look like -- [widgets are loaded](https://github.com/blutorange/primefaces/blob/demo-code-splitting/primefaces/src/main/frontend/src/core/core.js#L735) only when needed.

Once you run the bundler, the output would look like this: [src/main/frontend/dist](https://github.com/blutorange/primefaces/tree/demo-code-splitting/primefaces/src/main/frontend/dist) (disregard the additional folders). There's a main `index.js` corresponding to `core.js`, which imports the other files when required.

The **advantages** of that approach are:

* No more manual dependency management. The `@ResourceDependency` are not needed anymore (except perhaps for CSS files). Only a single script file `index.js` needs to be loaded manually.
* It always works. Code splitting is an optimization, but no matter how or where it is split, the resulting bundle works. Instead of having to define bundles manually, bundles are a result of the code itself.
* The bundler can analyse the bundle and split the code optimally.

But, why I think it's **hard to do for now**:

* Browser support. This needs native browser support for ES modules (`import` / `export`)
  * According to [mozilla](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script), browser support is pretty good though. Supported by Chrome since 2018 and by Firefox since 2020.
* Needs relative links to work. If file `a.js` has an `import "./b.js"`, resolving `./b.js` against the URL of `a.js` needs to result in a working URL. That's not compatible with the default Faces resource loading mechanism (`.xhtml` extension, additional query parameters such as `ln`).
  * Could be solved though by post-processing the resulting files and adjusting the URLs.
* The resulting JavaScript files are different (and can change between versions). Currently, the JAR contains files such as `jquery/jquery.js` or `core.js` that wouldn't be present anymore. That is a breaking change for users that manually use those files via `h:outputScript`.
  * Can be solved partially by creating wrapper files such as `jquery/jquery.js` that simply forward to the actual files with the implementation. Still needs `<script type=module>` or it won't work, though.
* `<script type=module>` are always `<script defer>`. That is, they won't execute immediately, but after the document was parsed and before the document is ready (`DOMContentLoaded`). This may affect users who try to execute script such as `PrimeFaces.info("Hello world")` before the document is ready.
  * Could be solved partially by loading `jquery` synchronously or providing a `$` shim so users can at least register on-document-ready callbacks.
* Requires code structure changes. For code splitting to work well, the code base needs to be migrated so that is uses `import / export` everywhere. That's not as bad as it sounds, but still quite some work to do.

melloware:
I definitely hear what you are saying about code splitting but i just worry this gets us so far away from JSF basics and the @resourcedependency stuff and it obfuscates with index.js for debugging and trying to tell what the heck is going on. That is just me thinking out loud about support and I worry even further this will break every MonkeyPatch in production.

blutorange:
> I definitely hear what you are saying about code splitting but i just worry this gets us so far away from JSF basics and the @resourcedependency stuff and it obfuscates with index.js for debugging and trying to tell what the heck is going on. That is just me thinking out loud about support and I worry even further this will break every MonkeyPatch in production.

Yeah, that was also kind just me thinking about it. When I did try it in a separate branch, I noticed lots of issues with that approach. Since you share my doubts, let's put that off the table for now.

But for my understanding, how would that break any monkey patch? Unless people are replacing our JAR resources with their own files?

melloware:
well i would hvae to test it but it just feels like the MonkeyPatching would be broken because of the same "`import`" issue you mentioned with TextArea and Autosize. If they patch in some functionality but that `import`?

Also wouldn't this affect:
1. Our CSP calculation of hashcode as it has to go through ResourceDependency?
2. OmniFaces CombinedResourceHandler and SRI calculation of HASH code of scripts?

blutorange:
Just a quick status update: The tests and everything are passing now. But before we merge this, I wanted to make sure I knew exactly how the final result with ESBuild + TypeScript + TypeDoc would look like (even though this PR won't do all of that). Took me some time, but I think I now have a plan for how it can work out. 

https://github.com/blutorange/primefaces-js-refactor-test/tree/demo-modules-export-2

The hardest part was getting TypeDoc to output pretty, readable HTML files. I had to raise some issues over at TypeDoc, and I think it's now looking good.

I also reworked the project structure slightly so that we don't have the issue from [12717](https://github.com/primefaces/primefaces/pull/12717#issuecomment-2466188620):

* One folder for each bundle, e.g. `core/*` or `components/*`.
* All files in such a folder/bundle can use `export` / `import` internally.
* But they use the global scope for referencing code from other folders/bundles
* Each folder has an `index.ts` file that exposes everything we want to expose to the global scope.

By using the global scope as before, everything should work as before, minimizing compatibility issues.

Also, it means that within a folder/bundle, we can create a separate file with a helper function; without automatically exposing it to the global scope. Only if we need to use that helper functions in other bundles do we need to expose it to the global scope.

Again, I won't add TypeScript / TypeDoc in this commit. There are few adjustments I need to make regarding the project structure, then this is ready to be merged.

image

melloware:
Great progress!

blutorange:
@melloware Another thing I just noticed: PrimeFaces extensions still creates widget classes via `BaseWidget.extend(...)` That won't work anymore.

* It might be possible to transpile the classes down to functions+prototypes and add back some of the old class workaround code.
* Or we just update the PF extensions code to use classes as well. That should hopefully be possible without any of the other refactorings from this (and following) PRs.

I'd tend more towards the latter option. Adjusting our code should not be an issue. There might be users who have custom widgets extending from PrimeFaces widgets. But considering how much effort it is to create a custom Faces component, I don't think it's unreasonable to ask those users to update to the class syntax (which by itself is just removing a bunch of `: function` and commas and adjusting super calls if any). Also, some users might already be using classes -- we have a few custom widgets using the class syntax and it's working fine.

https://github.com/primefaces-extensions/primefaces-extensions/blob/3bf0f494e50fef1a3cf79fe39d9662687b8e6770/core/src/main/resources/META-INF/resources/primefaces-extensions/counter/1-counter-widget.js#L6

melloware:
Yep we will have to update PFE in sync with PF after we merge this PR.

blutorange:
One question regarding moment locales: How / when are these loaded?

* [moment/moment.js[(https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/resources/META-INF/resources/primefaces/moment/moment.js) doesn't include any locales, so users need to load these themselves?
* But [timeline/0-vis-timeline.js](https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/resources/META-INF/resources/primefaces/timeline/0-vis-timeline.js#L5842) contains some locales (de, es, fr,  it, ja, nl, pl, ru, uk). But that file also includes a copy of the entire `moment` library, so perhaps that's not intentional?
* Schedule ( [schedule/fullcalendar.min.js](https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/resources/META-INF/resources/primefaces/schedule/fullcalendar.min.js) ) does not seem to include any moment locales

melloware:
@blutorange Timeline and Schedule have always bothered me. Here is what we document for Timeline: https://primefaces.github.io/primefaces/15_0_0/#/components/timeline?id=localization

I think we need a better way for both.

blutorange:
@melloware Hmm, I see. The ideal way imo would be

* We include the locale data in our `locales/de.js`
* When the timeline widget is *initialized*, we define the moment locale (if not yet defined) via `moment.defineLocale("de", PrimeFaces.Locales.de.moment)`

Then users wouldn't have to do anything. But that doesn't work -- it's unfortunate that moment doesn't ship just the locale data, it always calls `moment.defineLocale` itself (https://github.com/moment/moment/blob/develop/src/locale/de.js)

melloware:
Agreed it's always been a little weird and i have answered that Timeline issue on Stack Overflow a few times for people that would report "timeline localization doesn't work" so i finally added it to the docs.

melloware:
On thing we could do is modify the HeadRenderer to load the moment-locale when we load the PF locale if we detect Moment needs to be loaded.

https://github.com/primefaces/primefaces/blob/54573a7b90772bd9d058fffa4b073de6ab1befd2/primefaces/src/main/java/org/primefaces/renderkit/HeadRenderer.java#L131-L152

But that is only on page init

blutorange:
Actually, I just had an idea that we could use a function with a dynamic imports like this in our locale:

```js
// locales/locale-es.js
PrimeFaces.locales['es'] = {
  ...,
  loadMomentLocale: () => import("moment/dist/locale/de.js"),
}
```

And then call `PrimeFaces.locales['es'].loadMomentLocale()` when the timeline widget is initialized. I'll try this out later if it works out.

tandraschko:
+1 to import it in headrenderer for now, to have it the same like our own locale scripts

we can still refactor later

tandraschko:
we will release RC1 the next days; do you think we can finish this work for RC2 or should we target PF16?

blutorange:
@tandraschko I'd vote for PF16. Especially considering that this PR is only the first part (I'll want to make separate PRs for converting JSDoc type annotation to TypeScript annotations and generating the type declarations + doc pages from that).

blutorange:
The one failing test is due to the rate limit:

> Error:  Failed to execute goal io.github.cleydyr:dart-sass-maven-plugin:1.4.0:compile-sass (generate-themes) on project primefaces-themes: Execution generate-themes of goal io.github.cleydyr:dart-sass-maven-plugin:1.4.0:compile-sass failed: Failed to retrieve https://api.github.com/repos/sass/dart-sass/releases: {"message":"API rate limit exceeded for 20.172.0.32. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}: GitHub rate limit exceeded. -> [Help 1]

melloware:
so locally Dart Caches itself but on the GitHub runners i think its new everytime not sure how to tell it to cache there?

blutorange:
> [Actually, I just had an idea that we could use a function with a dynamic imports like this in our locale:](https://github.com/primefaces/primefaces/pull/12717#issuecomment-2510215996)

For reference, I gave this a quick shot and it would work, see [this branch](https://github.com/blutorange/primefaces/commit/294d1ceb14c2593ccfba7711c11fd63a0425e75a). As long as the corresponding PrimeFaces locale was loaded, it just works. I'll make a separate PR where we can discuss this once this is merged.

melloware:
Excellent that would be 1 more headache solved!

blutorange:
This PR is ready now (but as discussed above, it's better to merge this into PF16)

Just to confirm: since the master branch contains the current PF15 release candidate, we need to wait until the release of PF15 before it gets updated to the 16.0.0-SNAPSHOT, where this can get merged into?

As I said, the next part is making more code refactoring to prepare for the TypeScript conversion, but I want to wait for this to get merged to prevent merge conflicts.

melloware:
@blutorange yeah for now i think the plan is we should release 15.0.0 by EOM.  I don't think we really need an RC2 as since we can do 15.0.1 15.0.2 etc the need for multiple RC's isn't as important as it was in years past.  Then I will make a 15.X branch and master will be on 16.0.0-SNAPSHOT

blutorange:
End of this month sounds good :+1:

Also note to myself, I should test with CSP enabled (if not already enabled by default) due to issues such as #6868

blutorange:
I've seen there's a RC.2 now, things never go as planned ;) But no problem.

On another note, we have some singleton such as `PrimeFaces.clientwindow` that can be rewritten for improved static code analysis either with a class or without a class by simply exporting what's needed. Which style would you prefer?

https://gist.github.com/blutorange/f71b0c68491176bcb5e3720f13f7fbc7

(`core.clientwindow_esm.js` has the advantage that you can make certain constants, variables or functions truly private by not exporting them)

melloware:
I think I like style 1?

blutorange:
For reference, I had some time and started working on converting the `core` script to pure TypeScript. That will be a separate PR after we get this merged, but you can take a look if you're interested:

https://github.com/blutorange/primefaces/tree/refactor-js-classes-wip-core-classes/primefaces/src/main/frontend/packages/core/src

tandraschko:
are we ready here?
we should focus on refactoring parts first now after 15.0 has been released

same as the jakarta migration (i will do that part)

blutorange:
Great, 15 was released! Yeah, this PR is ready, but let me do one final manual test with the showcase.

melloware:
Sorry just created one more conflict for you.

melloware:
@blutorange is this safe to merge?

melloware:
@blutorange build fails on windows with


```
[INFO] --- frontend:1.15.1:yarn (build) @ primefaces ---
[INFO] Running 'yarn run build:bundle' in C:\dev\primetek\primefaces\primefaces\src\main\frontend
[INFO] Building PrimeFaces resources with mode development...
[INFO] Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
[INFO]   at throwIfUnsupportedURLScheme (node:internal/modules/esm/load:227:11)
[INFO]   at defaultLoad (node:internal/modules/esm/load:109:3)
[INFO]   at nextLoad (node:internal/modules/esm/hooks:748:28)
[INFO]   at load$1 (file:///C:/dev/primetek/primefaces/primefaces/src/main/frontend/.pnp.loader.mjs:1521:12)
[INFO]   at nextLoad (node:internal/modules/esm/hooks:748:28)
[INFO]   at Hooks.load (node:internal/modules/esm/hooks:385:26)
[INFO]   at MessagePort.handleMessage (node:internal/modules/esm/worker:199:24)
[INFO]   at [nodejs.internal.kHybridDispatch] (node:internal/event_target:816:20)
[INFO]   at MessagePort.<anonymous> (node:internal/per_context/messageport:23:28)
```

blutorange:
I was just clicking through the showcase to check for issues, but seems to be fine.

Hmm, that's interesting, considering the github actions build also builds on Windows? But that certainly looks like an issue with file paths

melloware:
no Actions is all Ubuntu Linux

melloware:
also i thought we didn't need the type-definitions folder anymore?

image

melloware:
honestly @blutorange i appreciate all this work but I am looking at this is an insanely complex build system using many "patches" and beta features and i can't even debug why its broken.  Not sure we want to do this long term?

i almost feel like we should roll this back.

blutorange:
Regarding the complex build system, let me write my point of view.  It's really just using a package manager (yarn) to install dependencies. Then it runs esbuild to bundle it. That's  it. 

The patches: Previously we had lots of 3rd party libs included directly in the repo that were modified somehow. I just used the NPM dependencies directly and added  a patch with our changes. Imo that makes it easier to track what we changed

But I guess if you fell uncomfortable we can roll it back for now and discuss it. I'm trying to get ahold of a  Windows machine to take a look at it. Issues with path differences Linux/Windows are pretty common unfortunately

melloware:
Maybe I just need to study it more and document it because it seems like a lot.

blutorange:
Though it's kinda unfortunate we only notice this after the merge. I can understand that's frustrating.

It's probably just some initial issue I guess, since I don't normally have a windows machine. At work, we use this kind of setup and it works for both linux and windows users.

But I'm open to changes. Perhaps we can make some things easier / simpler.

Either way, I got a windows device running now and can reproduce. Let me take a look.

blutorange:
See #13362 regarding the windows build issue

PR #12733, opened 2024-10-17
Fix #12732: If the menuitem has an icon defined use the icon instead of the index number

### LINKED ISSUE #12732: Steps: Display icon of menuItem instead of index number if defined
### Description

At the moment the steps component always renders the index of the current step inside the bubble. 
An easy improvement would be to also add the option to display an icon as alternative since it sometimes can provide more meaning.

If <p:menuitem ... icon="pi pi-bars"> has an icon defined the icon should be displayed inside the steps bubble instead of the index number.

### Additional context

_No response_

gako:
Fixes #12732

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/steps/StepsRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/steps/StepsRenderer.java
@@ -150,7 +150,14 @@ public class StepsRenderer extends BaseMenuRenderer {
 
         writer.startElement("span", steps);
         writer.writeAttribute("class", Steps.STEP_NUMBER_CLASS, null);
-        writer.writeText((index + 1), null);
+        if (menuitem.getIcon() != null) {
+            writer.startElement("span", null);
+            writer.writeAttribute("class", Steps.STEP_NUMBER_ICON_CLASS+" "+ menuitem.getIcon(), null);
+            writer.endElement("span");
+        }
+        else {
+            writer.writeText((index + 1), null);
+        }
         writer.endElement("span");
 
         Object value = menuitem.getValue();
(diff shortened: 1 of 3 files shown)


primefaces/src/main/java/org/primefaces/component/steps/StepsRenderer.java
```diff
@@ -150,7 +150,14 @@ protected void encodeMenuItem(FacesContext context, Steps steps, MenuItem menuit
 
         writer.startElement("span", steps);
         writer.writeAttribute("class", Steps.STEP_NUMBER_CLASS, null);
-        writer.writeText((index + 1), null);
+        if (menuitem.getIcon() != null) {
```
jepsar: It's better to use LangUtils.isEmpty here

jepsar:
Don't we need additional styling here? Can you please add a screenshot with an example of mixed items (so with and without an icon) and one of icons only?

gako:
The nice thing is it works without any additional styling. At least with PrimeIcons and Font Awesome. I havn't checked any other libraries.

The _ui-input-number-icon_ class is only there for convenience if somebody wants to add additional styling.

Here is a example of the showcase with icons instead of numbers (mixing is also possible)
grafik

primefaces/src/main/java/org/primefaces/component/steps/StepsRenderer.java
gako: Right, that way empty icon="" attributes are also handled correctly.  thx

gako:
Here is the mixed version.
For mixed mode the issue how the steps should be counted is open for discussion. The third bubble may also have the index number 2. But to keep things simple I did't change any logic there. Mixing is anyway probably only an edge case.

grafik

jepsar:
> For mixed mode the issue how the steps should be counted is open for discussion.

I think it's okay like it is currently.

jepsar:
@melloware should we promote/show this in the showcase?

melloware (approved):

melloware:
@gako yeah  showcase example would be great!

melloware (approved):

Issue #12519, opened 2024-08-24
Core: clean up PF utils


Rapster:
### Description

Based on https://github.com/primefaces/primefaces/pull/12512

Too many utils, not clear as to where to look when necessary

### Describe the solution you would like

IMO, we should stick to 3 utils at best:

- [ ] ComponentUtils _(no CompositeUtils, no TraversalUtils, no FacetUtils, the differences are very slight)_
- [ ] RendererUtils _(move a few CoreRenderer methods there)_
- [ ] ConverterUtils

A few methods can move elsewhere, for instance isFlex / isTouchable etc. can go in their respective classes

Also, 

- [ ] ~~MapBuilder should be  replaced by Map.of or Map.ofEntries~~ (this is tricky because Map.ofEntries does not allow NULL values)
- [x] PF class is useless, therefore should be deleted
- [x] Unused NestedSelectItem class
- [ ] ~LimitedSizeHashMap, still necessary?~ yes in use to defend against memory

tandraschko:
i wont give input here the next ~5-6 weeks, will be in a big japan holiday soon and refactoring isnt high prio for me

melloware:
@tandraschko have a great trip!  @Rapster i wouldn't mind waiting to do this until 16.0.X and would rather address some of the open bugs for 15.0.0 if we are putting our capacity somewhere.

Rapster:
Should open bugs be fixed in 14.x and 15.x ? Don't see why waiting for 16 when 15 is far from being released

melloware:
Yes they should I am just saying refactoring for 15.x for me is lower priority that addressing defects is all I am saying.

pizzi80:
imho code cleanup and refactors are always good to ease bug fixing and improvements ... 

in java, with a decent IDE, refactors are easy and almost bug free

btw which is the release date of PF 15.0 ?

`LimitedSizeHashMap` could at least be renamed in `LRUMap`

melloware:
No release date yet for 15.0.0

Rapster:
> LimitedSizeHashMap could at least be renamed in LRUMap

HashMap is no LRUMap

> btw which is the release date of PF 15.0 ?

Wrong place to ask that question (open a discussion instead)

pizzi80:
> > LimitedSizeHashMap could at least be renamed in LRUMap
> 
> HashMap is no LRUMap

the current PrimeFaces implementation it's backed by a `LinkedHashMap` 
[...]

PR #12747, opened 2024-10-19
Fix #12744 - Datatable: supports multiple editableValueHolder in filters

### LINKED ISSUE #12744: Datatable: Filtering dynamic columns from a composite component
Hi all,

I'm trying to filter dynamic columns (<p:columns ...), but filtering doesn't work if it's called from a composite component.

I made a standalone project (from primefaces-test) to try this (link is on bottom of this message).

If i move, for example, <p:inputText from file.xhtml to test.xhtml (just after <sf:file .....), then filtering will work properly.

Am i missing something or this is a bug ?

I tried it with PF 14.0.6.

Thank you !

[primefaces-test (1).zip](https://github.com/user-attachments/files/17437123/primefaces-test.1.zip)

Rapster:
I always felt that getting only the first item was some sort of trick but now it feels more compliant with the JSF API

### DIFF
--- a/primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
+++ b/primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
@@ -762,13 +762,6 @@ public class ComponentUtils {
         return component.getClass().getName().endsWith("UIRepeat");
     }
 
-    public static boolean isVisitable(VisitContext context, UIComponent component) {
-        // Copy from MyFaces
-        Set<VisitHint> hints = context.getHints();
-        return (!hints.contains(VisitHint.SKIP_UNRENDERED) || component.isRendered())
-                && (!hints.contains(VisitHint.SKIP_TRANSIENT) || !component.isTransient());
-    }
-
     public static Object convertToType(Object value, Class<?> valueType, Logger logger) {
         // skip null
         if (value == null) {
--- a/primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
+++ b/primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
@@ -34,7 +34,6 @@ import org.primefaces.context.PrimeRequestContext;
 import org.primefaces.csp.CspResponseWriter;
 import org.primefaces.renderkit.RendererUtils;
 
-import java.beans.BeanInfo;
 import java.io.IOException;
 import java.io.Serializable;
 import java.io.UncheckedIOException;
@@ -60,7 +59,6 @@ import javax.faces.FacesException;
 import javax.faces.FacesWrapper;
 import javax.faces.application.ConfigurableNavigationHandler;
 import javax.faces.application.NavigationCase;
-import javax.faces.component.ContextCallback;
 import javax.faces.component.EditableValueHolder;
 import javax.faces.component.StateHelper;
 import javax.faces.component.UIComponent;
@@ -71,18 +69,14 @@ import javax.faces.component.ValueHolder;
 import javax.faces.component.behavior.ClientBehavior;
 import javax.faces.component.behavior.ClientBehaviorHolder;
 import javax.faces.component.html.HtmlOutputFormat;
-import javax.faces.component.visit.VisitCallback;
 import javax.faces.component.visit.VisitContext;
 import javax.faces.component.visit.VisitHint;
-import javax.faces.component.visit.VisitResult;
 import javax.faces.context.ExternalContext;
 import javax.faces.context.FacesContext;
 import javax.faces.context.ResponseWriter;
 import javax.faces.convert.Converter;
 import javax.faces.convert.ConverterException;
 import javax.faces.render.Renderer;
-import javax.faces.view.AttachedObjectTarget;
-import javax.faces.view.EditableValueHolderAttachedObjectTarget;
 
 public class ComponentUtils {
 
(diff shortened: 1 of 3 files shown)


melloware (approved):
LGTM

melloware:
It looks like only on Mojarra we have two IT test failures.

```
OutputLabel001Test.composite2:140->assertLabel:222 expected: <true> but was: <false>
OutputLabel001Test.composite3:151->assertLabel:222 expected: <true> but was: <false>
```

Rapster:
> It looks like only on Mojarra we have two IT test failures.
> 
> ```
> OutputLabel001Test.composite2:140->assertLabel:222 expected: <true> but was: <false>
> OutputLabel001Test.composite3:151->assertLabel:222 expected: <true> but was: <false>
> ```

Yeah I think we had to tweak it for Mojarra (~8 years ago). But no tweak anymore, I managed to get rid of that code

melloware:
@tandraschko after you review let me know if its OK to squash and merge and backport to 14.0.7.

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
```diff
@@ -812,4 +818,59 @@ public static Object convertToType(Object value, Class<?> valueType, Logger logg
             return value;
         }
     }
+
+    /**
+     * Invoke callback on the first rendered {@link EditableValueHolder} component
+     */
+    public static void invokeOnDeepestEditableValueHolder(FacesContext context, UIComponent composite, ContextCallback callback) {
```
Rapster: ~@tandraschko I think it could be renamed, turns out to be the closest as it takes the first rendered component WDYT?~

tandraschko: the closest means the "closest outer" one?
in this case we go into all nested composites and look for the deepest nested EditableValueHolder
therefore IMO its "deepest" and belongs only the composites, therefor it was in CompositeUtils.

tandraschko: just to make sure:

if you do something like
```
p:outputLabel for=@next
myCC:wrappedWrappedInput
```

and wrappedWrappedInput has a cc:editableValueHolder for another CC called wrappedInput, which has a cc:editableValueHolder for a p:inputText

OutputLabelRenderer will then call the invokeOnDeepestEditableValueHolder and will be finally invoked for the real p:inputText

Rapster:
> the closest means the "closest outer" one?

Deepest turns out to be fine, it's not accurate but it's fine. We actually don't look for the deepest, but the "first" rendered EditableValueHolder to be precise

> in this case we go into all nested composites and look for the deepest nested EditableValueHolder
therefore IMO its "deepest" and belongs only the composites, therefor it was in CompositeUtils.

It was indeed introduced because of composites, but again see #12519. CompositeUtils if it was only meant for composite, it's not the case, it's being used from simple component, to UIPanel component and for composite component.

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
Rapster: > the closest means the "closest outer" one?

Deepest turns out to be fine, it's not accurate but it's fine. We actually don't look for the deepest, but the "first" rendered EditableValueHolder to be precise

> in this case we go into all nested composites and look for the deepest nested EditableValueHolder
therefore IMO its "deepest" and belongs only the composites, therefor it was in CompositeUtils.

It was indeed introduced because of composites, but again see #12519. CompositeUtils if it was only meant for composite, it's not the case, it's being used from simple component, to UIPanel component and for composite component.

tandraschko: it was actually only meant only for composites; im also not happy with ComponentUtils but this belongs to the linked issue and this bugfix
i just dont like to mix refactorings and bugfixes

Rapster: I don't mind putting it back in CompositeUtils, and move it once https://github.com/primefaces/primefaces/issues/12519 gets solved

melloware: That is a good idea @Rapster we can refactor with that other ticket.

primefaces/src/main/java/org/primefaces/util/CompositeUtils.java
```diff
@@ -44,50 +47,55 @@ public static boolean isComposite(UIComponent component) {
...
-                    }
+        @Override
+        public VisitResult visit(VisitContext context, UIComponent target) {
+            if (target instanceof EditableValueHolder) {
+                callback.invokeContextCallback(context.getFacesContext(), target);
+                return VisitResult.COMPLETE;
+            }
+            else if (UIComponent.isCompositeComponent(target)) {
+                visitEditableValueHolderTargets(context, target);
+                return VisitResult.COMPLETE;
```
Rapster: Don't merge I think I need to make quick fix here

primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
tandraschko: sounds good! +1

Rapster:
Ok I think I'm done with this one, I'm quite happy about the result 😄

melloware (approved):

tandraschko:
looks good!

PR #12749, opened 2024-10-19
Fix #12743 Do not render empty command containers (Menubar, Toolbar, MenuButton,...) [Draft]

### LINKED ISSUE #12743: AbstractMenu: "renderEmptyContainer" option to prevent empty menus from being rendered
### Description

I often encounter cases were I use some kind of menu/container component (MenuButton,  Menubar, Toolbar, ...) with a list of menu items that are either rendered or not depending on the current rights of the user. In cases where the current user has none of the required rights the menuButton is still rendered and can be clicked but an empty popup is shown.

E.g "Action" MenuButton with Menu Items for "Edit", "Delete", "Print" each having its own rendered definition
`
<p:menuButton value="actions">
   <p:menuitem value="Edit" rendered="#{checkRequirementsRights('edit')}">
   <p:menuitem value="Delete" rendered="#{checkRequirementsRights('delete')}">
   <p:menuitem value="Print" rendered="#{checkRequirementsRights('print')}">
</p:menuButton>
`

I could also add a rendered to the menuButton and combine any checks with OR but depending on the number of items this soon becomes anoying. For my toolbar this is not possible because here I define the toolbar itself in a pageTemplate and each subPage defines it's unique actions for it.

Adding a "renderEmptyContainer" (default true for backward compatibility) with a check wether any children are rendered and if not not rendering the whole component could solve this issue.

Is this something that could be added as a feature or can anyone come up with a different solution?
If yes I could provide a PR for some components where I already fixed it via an extension.

Components that act as some kind of container for other items (Toolbar, MenuButton, Menubar, ...) should have an option to not being rendered at all if no real children ( e.g ignore separaors) are present/rendered.

An alternative solution could be to render the component as usual but add some kind of empty marker via a css style class to the container ("ui-widget-empty" or something like that). If somebody want's to hide these components a simple .ui-widget-empty {display:none} rule would suffice. For this solution no additional attribute would be necessary since always adding the marker class should do no harm if no css rule is defined.

### Additional context

_No response_

gako:
Initial Draft PR regarding not (always) rendering containers that contain no rendered children. 
Fix #12743

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/splitbutton/SplitButtonRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/splitbutton/SplitButtonRenderer.java
@@ -23,6 +23,7 @@
  */
 package org.primefaces.component.splitbutton;
 
+import org.primefaces.component.api.MenuItemAware;
 import org.primefaces.component.menu.AbstractMenu;
 import org.primefaces.component.menu.Menu;
 import org.primefaces.component.menubutton.MenuButton;
@@ -100,10 +101,12 @@ public class SplitButtonRenderer extends MenuItemAwareRenderer {
         String styleClass = button.getStyleClass();
         styleClass = styleClass == null ? SplitButton.STYLE_CLASS : SplitButton.STYLE_CLASS + " " + styleClass;
 
+        boolean hasOverlay = shouldBeRendered(context, button);
+
         writer.startElement("div", button);
         writer.writeAttribute("id", clientId, "id");
         writer.writeAttribute("class", styleClass, "id");
-        writer.writeAttribute(HTML.ARIA_HASPOPUP, "true", null);
+        writer.writeAttribute(HTML.ARIA_HASPOPUP, Boolean.toString(hasOverlay), null);
         writer.writeAttribute(HTML.ARIA_CONTROLS, menuId, null);
         writer.writeAttribute(HTML.ARIA_EXPANDED, "false", null);
         if (button.getStyle() != null) {
@@ -114,7 +117,9 @@ public class SplitButtonRenderer extends MenuItemAwareRenderer {
         OverlayPanel customOverlay = button.getCustomOverlay();
         if (customOverlay != null || button.getElementsCount() > 0) {
             encodeMenuIcon(context, button, menuButtonId);
-            encodeMenu(context, button, menuId);
+            if (hasOverlay) {
+                encodeMenu(context, button, menuId);
+            }
         }
 
         writer.endElement("div");
@@ -441,5 +446,13 @@ public class SplitButtonRenderer extends MenuItemAwareRenderer {
     @Override
     public boolean getRendersChildren() {
         return true;
+    }
+
+    @Override
+    protected boolean shouldBeRendered(FacesContext context, MenuItemAware container) {
+        SplitButton button = (SplitButton) container;
+        boolean  rendered = button.getCustomOverlay() != null;
+        rendered = rendered || super.shouldBeRendered(context, button);
+        return rendered;
     }
 }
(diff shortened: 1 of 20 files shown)


tandraschko:
i like this more as introducing a new attribute
Some things we need to consider
1) is the #isRendered always working? EL like #{component} is not available without using pop/pushComponent
2) if we add pop/pushComponent stuff, does it influence performance?
3) is there a case where non-rendered menuItems does still make sense? their is e.g. in splitButton. We should check this and add integrationtests to not break something here.

gako:
ad 1:  You're right, didn't think of that. I added it to the stack during the rendered check. Is it enough to add it just before the call of isRendered?

ad 2:
I just went through most components and the following may need a special treatment:

- SplitButton: 
If no menu items are rendered, the chevron currently has no function. This means either we do not render the chevron at all, basically converting the SplitButton to a regular button or render the chevron but disable it.
- Speed Dial
If there is a badge on the speed dial rendering it without any menu items may be necessary. Button probably should also be disabled if no action can be taken with it.
- Dock
Empty dock may as well be part of the visual design, with its static position at the screen border. (Keep old behaviour)
- Stack
Same as dock, empty basket should probably be rendered as before. (Keep old behaviour)

primefaces/src/main/java/org/primefaces/component/splitbutton/SplitButtonRenderer.java
```diff
@@ -87,6 +88,8 @@ public void encodeEnd(FacesContext context, UIComponent component) throws IOExce
             model.generateUniqueIds();
         }
 
+        boolean hasOverlay = shouldBeRendered(context, button);
```
tandraschko: can be removed

tandraschko:
added a comment
and we also have a conflict
also i would like to see a entry in the migration guide

otherwise it looks good for me

any other reviewers?

primefaces/src/main/java/org/primefaces/component/toolbar/ToolbarRenderer.java
```diff
@@ -115,4 +119,19 @@ public void encodeChildren(FacesContext facesContext, UIComponent component) thr
     public boolean getRendersChildren() {
         return true;
     }
+
+    protected boolean shouldBeRendered(FacesContext facesContext, Toolbar toolbar) {
+        if (toolbar.getChildCount() > 0) {
+            for (UIComponent child : toolbar.getChildren()) {
+                if (child.isRendered() && child instanceof ToolbarGroup) {
+                    ToolbarGroup toolbarGroup = (ToolbarGroup) child;
+                    return toolbarGroup.getChildren().stream().filter(me -> !(me instanceof Separator)).anyMatch(UIComponent::isRendered);
```
tandraschko: shouldnt it use
`shouldBeRendered(FacesContext context, MenuElement menuElement)`
because push/pop?

gako: Theoretically yes, but since Toolbar is the only component that is no real Menu, it does not extend BaseMenuRenderer and does not use MenuItems to define it's children, I added a similiar method that's based around UIComponent to ToolbarRenderer itself.

gako:
By the way: I also changed a minor thing in FileUploadUtilsTest.requireValidFilePath_AbsoluteFileWithWindowsSpaces 

Because I got ~"Device Not Read" Exceptions in the method getCanonicalFile on my local machine due to the fact that in my case D: is a CD-ROM/DVD-Drive which is empty and windows/java seems to throw an error in that case.
I changed the path to device C: since that should always be available.

tandraschko:
can you rebase your PR? (259 files changed now)

gako:
Strange I did a rebase onto the latest master, the changed files still showed up.
No I reset the PR to master and reapplied the changes. The push history is lost, but at least the changed files are now correct.

melloware:
This LGTM to me too can you add a note in the /docs/migration_guide/15.0.0.md about this change please?

melloware (approved):

melloware:
That is a good point because if there is nothing at all in the tree nothing can be updated...

tandraschko:
looks good to me

PR #12774, opened 2024-10-24
Document animations

tandraschko:


docs/15_0_0/core/animations.md
```diff
@@ -0,0 +1,28 @@
...
+
+## disable / enable animations 
+
+Its also possible to override the `prefers-reduced-motion` setting via JS:
+
+```js
+PrimeFaces.utils.disableAnimations();
+PrimeFaces.utils.enableAnimations();
+```
+
```
melloware, line 27: Maybe we should move the blurb from `\core\ajaxJavascriptApi.md` here too about `Inline load animation`?

melloware (changes requested):

melloware (approved):

PR #12775, opened 2024-10-24
Fixes #12771 14.0.7 AJAX: resolving clientIds of currently unrendered should not throw a exception

### LINKED ISSUE #12771: AJAX: resolving clientIds of currently unrendered components should not throw a exception
Currently when you have:

```
            <p:commandButton id="btn" actionListener="#{testView.buttonAction}" update="@widgetVar(testInputText)" />
            <p:panel id="testPanel" widgetVar="testPanel">
                <p:inputText id="testInputText" widgetVar="testInputText" value="#{testView.string}" rendered="false"/>
            </p:panel>
```

our AJAX core and `@widgetVar` resolver will skip unrendered components.
this is actually bad as rendered can be changed dynamically and the update would work then.
Throwing an ComponentNotFoundException isnt the right way as the component is there, just not yet updateable.
"component not found" vs "component not yet visible"

i will reintroduce the old behavior and collect also the unredendered clientIds

melloware:
Fix #12771 14.0.7 AJAX: resolving clientIds of currently unrendered should not throw a exception

### DIFF
--- a/docs/PrimeFaces.d.ts
+++ b/docs/PrimeFaces.d.ts
@@ -15845,23 +15845,7 @@ declare namespace PrimeFaces.widget {
     }
 }
 declare namespace PrimeFaces.widget.FileUpload {
-    /**
-     * A error message for a file upload widget.
-     */
-    export interface UploadMessage {
-        /**
-         * The name of the uploaded file.
-         */
-        filename: string;
-        /**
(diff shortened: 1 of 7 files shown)


melloware:
Ahhh let me see its tough to cherry pick a single commit but let me try.

PR #12794, opened 2024-10-28
Fixes #9254 - Treetable: clearFilters() API make table breaks table when filtering is enabled

### LINKED ISSUE #8073: TreeTable: after filtering and changing value during ajax old filtered data is still shown

### LINKED ISSUE #9254: Treetable: clearFilters() API make table breaks table when filtering is enabled
### Describe the bug

After I use the clearFilters(), the table is frozen. Although the value is change on the Java side, the tree table does not show any changes.

### Reproducer

1. Go to left corner to choose a tree root and load the tree root (default is root1) as you like
2. Click on "clear filters" button (no matter is anything in the filter side)
3. Go back to left corner to select other tree root to load
4. See the **table would not change**. But in the debugger, the **value has changed to selected tree node**
5. No errors on console
[primefaces-test-master.zip](https://github.com/primefaces/primefaces/files/9689575/primefaces-test-master.zip)

### Expected behavior

The expected result is:
After click the "clear filter" button, when user change the root and load it, the table should be changed correspondingly.

### PrimeFaces edition

Community

### PrimeFaces version

11.0.0

### Theme

_No response_

### JSF implementation

All

### JSF version

2.2

### Java version

1.8

### Browser(s)

_No response_

tandraschko:
I think it should also fix #8073

i aligned it with DataTable now, to not store the filtered stuff memory and just refilter each request

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/timeline/DefaultTimelineUpdater.java
+++ b/primefaces/src/main/java/org/primefaces/component/timeline/DefaultTimelineUpdater.java
@@ -135,121 +135,124 @@ public class DefaultTimelineUpdater extends TimelineUpdater implements PhaseList
         if (crudOperationDatas == null) {
             return;
         }
-        StringBuilder sb = new StringBuilder();
 
-        Timeline timeline = (Timeline) context.getViewRoot().findComponent(id);
-        TimelineRenderer timelineRenderer = ComponentUtils.getUnwrappedRenderer(
-                context,
-                Timeline.COMPONENT_FAMILY,
-                Timeline.DEFAULT_RENDERER);
+        context.getViewRoot().invokeOnComponent(context, id, (ctx, component) -> {
+            StringBuilder sb = new StringBuilder();
 
-        TimelineModel<Object, Object> model = timeline.getValue();
-        List<TimelineGroup<Object>> groups = timelineRenderer.calculateGroupsFromModel(model);
-        UIComponent groupFacet = timeline.getFacet("group");
-        // buffer for groups' content
-        Map<String, String> groupsContent = new HashMap<>();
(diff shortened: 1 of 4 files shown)


tandraschko:
seems we already had this in the past?
https://github.com/christophs78/primefaces/commit/4328cf64fb32458ed5b208f71e147f41c96c4c42

lets see what the IT says

tandraschko:
```
Error:  Failures: 
Error:    TreeTable001Test.filter:196 expected: <3> but was: <1>
Error:    TreeTable001Test.globalFilter:250 expected: <3> but was: <1>
Error:  Errors: 
Error:    TreeTable006Test.rowEdit:78->AbstractTreeTableTest.assertMessage:116 » NullPointer
Error:    TreeTable007Test.cellEdit:73->AbstractTreeTableTest.assertMessage:116 » NullPointer
```

the 2 upper errors are because the expanded state is lost as the nodes are cloned.
This actually display the whole misconception of TreeTable filtering.
There is no easy fix:
1) syncing
    currenlty value is overwritten from filteredValue; Therefore all work is done now on the filteredValue cloned objects which will be lost next request. we would need to sync value changes... Just not possible.
2) just see filtering as a additional info, to remeber filtered rowKeys and not-render them again. That means that filteredValue will be removed and value always contains all nodes. We just filter them out on rendering. This is a breaking chance.

tandraschko:
closing in favor of #12795

Issue #12802, opened 2024-10-29
DataTable: When inside a AccordionPanel sticky header stays in wrong position


ra-ke:
### Describe the bug

When a data table with `stickyHeader="true"` is placed within the second tab of an accordion panel (with the `multiple` option enabled), collapsing the first tab causes the sticky header in the second tab to become misaligned.

a demonstrator can be found [here](https://github.com/ra-ke/primefaces-datatable-accordion-demo)

### Reproducer

1. Expand the second tab of the accordion panel
2. Collapse the first tab of the accordion panel
3. see the sticky header stay in ints position

### Expected behavior

the sticky header moves upwards with the rest of the table

### PrimeFaces edition

Community

### PrimeFaces version

14.0.6

### JSF implementation

All

### JSF version

2.3

### Java version

17

### Browser(s)

Chrome and Firefox

melloware:
@ra-ke I don't ever recommend doing what you are doing with a Sticky inside another moving container I was able to "fix" your reproducer.
[pf-12802.zip](https://github.com/user-attachments/files/17556653/pf-12802.zip)

Basically after opening or closing tabs you have to tell the browser it "scrolled" so the sticky moves.

```xml
<p:ajax event="tabClose" process="@this" oncomplete="PrimeFaces.queueTask(function() {window.dispatchEvent(new Event('scroll'))}, 400)" />
<p:ajax event="tabChange" process="@this" oncomplete="PrimeFaces.queueTask(function() {window.dispatchEvent(new Event('scroll'))}, 400)" />
```

This is not something PrimeFaces will fix as we can't know every scenario and like I said I don't recommend this UI/UX paradigm anyway.

PR #12803, opened 2024-10-29
Execute filterEnricher before columnFilters allows multi-field filtering from outside dataTable

riyadomf:
Execute filterEnricher before columnFilters (where applyGlobalFilters previously was). It will now be possible to create custom filterMeta without referencing specific field; handle the filterMeta in enricher, create predicates accordingly, and remove the filterMeta

### DIFF
--- a/primefaces/src/main/java/org/primefaces/model/JPALazyDataModel.java
+++ b/primefaces/src/main/java/org/primefaces/model/JPALazyDataModel.java
@@ -142,6 +142,10 @@ public class JPALazyDataModel<T> extends LazyDataModel<T> implements Serializabl
 
         List<Predicate> predicates = new ArrayList<>();
 
+        if (filterEnricher != null) {
+            filterEnricher.enrich(filterBy, cb, cq, root, predicates);
+        }
+
         if (filterBy != null) {
             FacesContext context = FacesContext.getCurrentInstance();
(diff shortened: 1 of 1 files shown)


riyadomf:
solves #12802

tandraschko:
please create a new issue with exact information, what you try to solve here

we posted you a solution for your globalFilter, not sure what exact issue of this big discussion tries to solve this.

tandraschko:
sorry, we need a real github-issue with a problem you are solving
in your other issue, it seems you are hacking around datatable internals instead of trying the official way

riyadomf:
I understand, but the official way doesn't let me use multi field filter outside dataTable.
Shifting up the filterEnricher will at least let me hack around it.

tandraschko:
why it doesnt?
as we said, you can add outside filters from beans in filterEnricher as much as you like

you manipulate datatable internals in 2 corners instead of just use how it meant to be

melloware:
@riyadomf i already proved your use case will work just fine with my example code for some reason you would rather hack around than just refactor your code and use the component correctly?

riyadomf:
Ok, got your point. Thank you!

tandraschko:
if you wanna hack arround, i would suggest inherit JPALazyDataModel and add something like a filterMetaManipulator callback.
i will close this for now

PR #12818, opened 2024-10-31
Fixes #12816 - Unify TreeNodes / Cleanup duplicate code

### LINKED ISSUE #12816: Unify TreeNodes / Cleanup duplicate code
as i looked into #12795 recently

currently `CheckboxTreeNode` and `DefaultTreeNode` are 95% the same code
`CheckBoxTreeNode` can just inherit `DefaultTreeNode` like `LazyDefaultTreeNode` does

also TreeNode defines children as `List<TreeNode>` - but we also have `TreeNodeList`
i think we should deprecate `TreeNodeLis`t - using `List<TreeNode>` is enough

tandraschko:
Fixes #12816 - Cleanup TreeNodes

### DIFF
--- a/docs/migrationguide/15_0_0.md
+++ b/docs/migrationguide/15_0_0.md
@@ -46,3 +46,7 @@
 ## Showcase
   * Was migrated to Jakarta EE 10. Requires e.g. Jetty 12 and Java 17 to run stand-alone.
 
+## Tree / TreeTable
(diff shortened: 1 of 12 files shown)


melloware (changes requested):
Migration Guide notes in case anyone has customized their own tree nodes?

melloware (approved):

PR #12823, opened 2024-11-01
Refactor Barcode to use a single modern dependency

melloware:
For discussion. Upgrade the barcode to use a modern library that is full featured.  Switch to https://github.com/woo-j/OkapiBarcode which supports every feature.

1. Barcode4J has not received an update since 2012 (it dead)
2. Naryuki QrCode is only used for QR Codes
3. Okapi supports both in 1 library including PNG  and SVG support
4. Okapi supports more barcodes than other other library out there
5. Okapi is kept up to date and still receiving commits
6. Okapi has no 3rd party dependencies
7. As you can see from this PR it reduces the amount of PrimeFaces code as well without changing the XHTML at all!

**TODO:**

- [x] Support orientation https://github.com/woo-j/OkapiBarcode/issues/114
- [x] Add new property for `magnification` so you can make a barcode 2x, 3x or Nx bigger!
- [x] Move `BarcodeHandler` to the `.component.barcode` package?


tandraschko:
in general +1
didnt do a detailed code review

primefaces/src/main/java/org/primefaces/application/resource/BarcodeHandler.java
```diff
@@ -21,51 +21,62 @@
...
+        generators.put("code39", new Code3Of9());
+        generators.put("code128", new Code128());
+        generators.put("ean8", new Ean(Ean.Mode.EAN8));
+        generators.put("ean13", new Ean(Ean.Mode.EAN13));
+        generators.put("upca", new Upc(Mode.UPCA));
+        generators.put("upce", new Upc(Mode.UPCE));
+        generators.put("postnet", new Postnet());
+        generators.put("pdf417", new Pdf417());
+        generators.put("datamatrix", new DataMatrix());
+        generators.put("qr", new QrCode());
```
Rapster: These `Symbol` seems to be stateful (it was certainly the same before) but ideally the state of these objects shouldn't be shared within the same HTTP request (and different requests). This map should act as a factory, so instead of having symbol instance, something like `Pdf417::new`. This way you ensure you have a clean object every time it's requested

Rapster: And +1 to move BarcodeHandler to the .component.barcode package

melloware: OK yeah i will do the `::new` i saw another library doing that recently so good feedback!

melloware: Switched to `private final Map<String, Supplier<Symbol>> generators;` so it returns a new instance each time,

primefaces/src/main/java/org/primefaces/component/barcode/BarcodeHandler.java
```diff
@@ -21,51 +21,64 @@
...
+        generators.put("code39", () -> new Code3Of9());
+        generators.put("code128", () -> new Code128());
+        generators.put("ean8", () -> new Ean(Ean.Mode.EAN8));
+        generators.put("ean13", () -> new Ean(Ean.Mode.EAN13));
+        generators.put("upca", () -> new Upc(Mode.UPCA));
+        generators.put("upce", () -> new Upc(Mode.UPCE));
+        generators.put("postnet", () -> new Postnet());
+        generators.put("pdf417", () -> new Pdf417());
+        generators.put("datamatrix", () -> new DataMatrix());
+        generators.put("qr", () -> new QrCode());
```
Rapster: wherever it's possible, prefer XXX::new (some IDEs complains otherwise)

Rapster: Also, you can "Inline initialized" this map by using `Map.ofEntries` and make it static and have one single instance

melloware: Changes made!

primefaces/src/main/java/org/primefaces/component/barcode/BarcodeHandler.java
```diff
@@ -21,52 +21,61 @@
...
-        generators.put("code128", new Code128Generator());
-        generators.put("ean8", new EAN8Generator());
-        generators.put("ean13", new EAN13Generator());
-        generators.put("upca", new UPCAGenerator());
-        generators.put("upce", new UPCEGenerator());
-        generators.put("postnet", new PostnetGenerator());
-        generators.put("pdf417", new PDF417Generator());
-        generators.put("datamatrix", new DataMatrixGenerator());
-    }
+    private final Map<String, Supplier<Symbol>> generators = Map.ofEntries(
```
Rapster: static final :smile:

melloware: Done

primefaces/src/main/java/org/primefaces/component/barcode/BarcodeHandler.java
```diff
@@ -85,13 +94,20 @@ public void handle(FacesContext context) throws IOException {
...
             String hrp = params.get("hrp");
             int orientation = Integer.parseInt(params.get("ori"));
+            double magnification = Double.parseDouble(params.get("mag"));
             boolean cache = Boolean.parseBoolean(params.get(Constants.DYNAMIC_CONTENT_CACHE_PARAM));
 
-            generator.getBarcodeBean().setMsgPosition(HumanReadablePlacement.byName(hrp));
+            generator.setHumanReadableLocation(HumanReadableLocation.valueOf(hrp.toUpperCase(Locale.ROOT)));
+            generator.setContent(value);
+
+            if ("qr".equals(type)) {
```
Rapster: To avoid unsafe cast warning, it might be "better" to use `instanceof QrCode`

melloware: Done!

melloware:
Okapi released 0.4.9 with the rotation fix and its updated and now 100% backwards compatible with the old Barcode4J!

PR #12826, opened 2024-11-04
Fix #12825: Make Dashboard scope configurable

### LINKED ISSUE #12825: Dashboard: Scope of jquery drag and drop should be configurable
### Description

At the moment the scope of draggable droppable is hardcoded to "dashboard" which is fine as long as only one dashboard is used per page. I have a scenario where I have three different sections each containing a dashboard with items. To prevent the items from one section to be draggable to a different section i need a different scope in the jquery config.

_Scope_ should be configurable in the dashboard tag and then passed forward to the jquery draggable/droppable part.

### Additional context

_No response_

gako:
Fix #12825

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/dashboard/DashboardBase.java
+++ b/primefaces/src/main/java/org/primefaces/component/dashboard/DashboardBase.java
@@ -44,7 +44,8 @@ public abstract class DashboardBase extends UIPanel implements Widget, ClientBeh
         style,
         styleClass,
         responsive,
-        var
+        var,
+        scope
     }
 
     public DashboardBase() {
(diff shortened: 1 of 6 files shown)


primefaces/src/main/java/org/primefaces/component/dashboard/DashboardBase.java
```diff
@@ -44,7 +44,8 @@ public enum PropertyKeys {
         style,
         styleClass,
         responsive,
-        var
+        var,
+        scope
```
melloware: You can move scope up to be in alpha order.  I am trying to do this lately 😄

melloware (changes requested):
Don't forget a note in 15.0.0 whatsnew.md and to add the new prop in dashboard.md in 15.0.0 folder?

primefaces/src/main/java/org/primefaces/component/dashboard/DashboardBase.java
gako: having a half-sorted list felt strange, so i've taken the liberty of sorting the whole list. :-)

melloware (approved):

PR #12842, opened 2024-11-05
Fix #12841: Delayed Hover Effect for TieredMenus

### LINKED ISSUE #12841: TieredMenu: ShowDelay should use delayed hover effect if toggleEvent is hover
### Description

At the moment the showDelay only delays the showing of the submenu by X ms, but it has no effect at whether the event is fired at all or not. Meaning the submenu of a e.g menubar or tieredmenu will be shown even if I hover of the item for less than showDelay ms.

To reproduce this behavior set a relative high showDelay (1sec) and quickly move over the items of a menubar. The submenu will popup long after you left the item.

Also if I leave a TieredMenu the last active submenu stays open if hideDelay = 0, I think it should be closed immediatly.

A submenu should should only be shown if the mouse hovers over the desired element for longer than `showDelay`.
That way accidental hovers won't trigger a submenu display.

### Additional context

_No response_

gako:
Fix #12841

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/menu/menu.tieredmenu.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/menu/menu.tieredmenu.js
@@ -116,6 +116,16 @@ PrimeFaces.widget.TieredMenu = PrimeFaces.widget.Menu.extend({
             }
         });
 
+        // clear timeout of possible delayed show event if mouseleave is fired before showDelay was over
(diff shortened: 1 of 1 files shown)


melloware:
i am testing this now.

melloware:
Testing locally looks good

melloware:
@gako i think this actually broke one subtle feature of MenuBar: https://github.com/orgs/primefaces/discussions/3539

Issue #12845, opened 2024-11-07
TabMenu: Scrollbar below TabMenu on Google Chrome


gako:
### Describe the bug

If you open http://www.primefaces.org/showcase/ui/menu/tabMenu.xhtml in latest google chrome version you will always see a scrollbar below tabmenu due to `overflow-x:scroll` rule on `.ui-tabmenu-nav`.  `auto` would probably be the better behaviour.

### Reproducer

Open http://www.primefaces.org/showcase/ui/menu/tabMenu.xhtml in Google Chrome

### Expected behavior

Scrollbar should only be visible if number of tabs exceeds screenwidth

### PrimeFaces edition

Community

### PrimeFaces version

14.0.5

### Theme

Saga,Arya, all based on ThemeBase

### JSF implementation

Mojarra

### JSF version

2.2

### Java version

17

### Browser(s)

Chrome Version 130.0.6723.117 (Offizieller Build) (64-Bit)

tandraschko:
@melloware if you wanna merge to 14.x, feel free
but remember to copy all themes.css to pf and showcase

melloware:
got it

PR #12846, opened 2024-11-07
Fix 12845: Tab menu chrome scrollbar

gako:
see #12845


tandraschko:
please next time also push generated themes.css
i will commit it now

gako:
you right I forgot they are also checked in, my bad.

PR #12848, opened 2024-11-07
Fix #12844: SelectCheckBoxMenu allow empty label

### LINKED ISSUE #12844: SelectCheckBoxMenu: Broken rendering for empty labels
### Describe the bug

When a user provides the empty string for a SelectItem in a SelectCheckBoxMenu it renders the id instead of the blank string in the drop-down list.

In addition, when selected, the chips area contains the empty string which leads to a render error, rendering a non breaking space would fix that here.

### Reproducer

https://github.com/stolp/primefaces-test/tree/pf-12844

### Expected behavior

Render an empty label and add a non breaking space to the chips entry.

### PrimeFaces edition

Community

### PrimeFaces version

14.0.6

### Theme

Default theme

### JSF implementation

Mojarra

### JSF version

4.0

### Java version

11

### Browser(s)

all

melloware:
Fix #12844: SelectCheckBoxMenu allow empty label

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/selectcheckboxmenu/SelectChecboxMenu002.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/selectcheckboxmenu/SelectChecboxMenu002.java
@@ -1,0 +1,90 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2024 PrimeTek Informatics
+ *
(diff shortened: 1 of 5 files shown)


melloware:
@christophs78 @tandraschko it looks like this was "half" working but the issue is when we moved the item rendering to the Client Side this use case was not fully implemented

tandraschko:
can we have a test for it?

melloware:
Yep let me add one today.

melloware:
IT added

PR #12857, opened 2024-11-08
Fixes #12856 - Timeline: Adding events dynamically not working anymore in when passing a search expression to the TimelineUpdater

### LINKED ISSUE #12856: Timeline: Adding events dynamically not working anymore in when passing a search expression to the TimelineUpdater
PrimeFaces version: 15.0.0-SNAPSHOT

**Issue**

While doing some testing for #12717, I noticed that adding events to the timeline does not work anymore. First I thought it might be due to the refactoring, but I tried with the current master branch and it doesn't work there either.

Start the showcase and go to e.g. `http://localhost:8080/showcase/ui/data/timeline/dragdrop.xhtml`. Try adding an event to the timeline, the added event won't show up.

The same issues also occurs for `http://localhost:8080/showcase/ui/data/timeline/lazy.xhtml`.

[Screencast from 2024-11-08 13-53-18.webm](https://github.com/user-attachments/assets/d90a1131-c226-445b-947e-2d7ce178aa5b)

When I compare the behavior with 14.x, the `partial-response` to the AJAX request that adds the event seems to be the missing an `<eval>` change with `PF('widget_form_timeline').addEvent({...})`

14.x
image

15.x
image

**Cause**

At  first glance, the issue seems to be that the `DefaultTimelineUpdate` tries to look for timeline components via `context.getViewRoot().invokeOnComponent(FacesContext, String id, callback)`, but with an ID such as `:form:timeline`. Note the column at the begin, that's a search expression, not a valid ID.

That "ID" `:form:timeline` itself [is set in the DndTimelineView](https://github.com/primefaces/primefaces/blob/0d7b7b84ad436cf91cb97e7fa7560e88ae3170c3/primefaces-showcase/src/main/java/org/primefaces/showcase/view/data/timeline/DndTimelineView.java#L85) class from the showcase. That code is from 2021 however and was not changed since, so I guess it's fair to assume that people already rely on this working.

The root cause is that the part `context.getViewRoot().findComponent(id)` in the `DefaultTimelineRenderer` was changed to `context.getViewRoot().invokeOnComponent(context, id, ...)` here:

https://github.com/primefaces/primefaces/commit/12915c1d2a9b860ad9cb3a982220c0f9f8cf65c8#diff-91ae8ac887739485b870229ee9bd52bcfa89ab30b6b0471b26c2608c1ccd8b53R139

As per the JavaDoc, `findComponent` accepts a search expression, but `invokeOnComponent` only a client ID.

---

image

**Possible solutions**

We can't just switch back to `findComponent`, since that was the issue from #6944 / #12792

The variable name certainly suggests that only IDs were intended to be used, not search expressions. That would be a breaking change though, since I'd bet people rely on the current behavior.  But perhaps that's acceptable and should be easy for users to fix?

tandraschko:
Fixes #12856 - Timeline: Adding events dynamically not working anymore in when passing a search expression to the TimelineUpdater


melloware:
Looks like some TreeTable changes in here as well?

blutorange:
Here's the PR without the extraneous commits #12858, this one should be closed.

PR #12863, opened 2024-11-11
Fix #12855: HeadRenderer check for already loaded resource

### LINKED ISSUE #12855: moment.js gets loaded twice when client side validation is enabled
PrimeFaces version: 15.0.0-SNAPSHOT

**Issue**

`moment.js` gets loaded twice when client side validation is enabled.

Start the showcase and go to e.g. `http://localhost:8080/showcase/ui/chart/bubble.xhtml`. Check the network tab, `moment.js` is loaded twice. 

This happens since #11943.  `moment.js` is included as a `ResourceDependency` on the `Chart` component, and again in the `HeadRenderer`:

https://github.com/primefaces/primefaces/blob/0d7b7b84ad436cf91cb97e7fa7560e88ae3170c3/primefaces/src/main/java/org/primefaces/renderkit/HeadRenderer.java#L141

The downsides are that (a) more code than necessary is loaded, and (b) there are now 2 instances of `moment.js` co-existing and any setup like `moment.defineLocale` won't affect the other instance. Especially with the refactor from #12717, the first instance of `moment.js`  can be captured by some code before the other instance is loaded.

**Possible solution**

The `HeadRenderer` always inserts `<script>` element with moment.js, without regard for whether that resource is already included. [OmniFaces has a addScriptResource](https://omnifaces.org/docs/javadoc/current/org.omnifaces/org/omnifaces/util/Components.html#addScriptResource(java.lang.String,java.lang.String)) utility function that takes care of this. We don't use OmniFaces, but I saw there's a `org.primefaces.util.ResourceUtils` with a `addComponentResource` method, though that might need some additions such as the [ResourceHandler.isResourceRendered check from OmniFaces](https://github.com/omnifaces/omnifaces/blob/372a78d66ec8b460066ca7e6d56c96c1bccc6dd0/src/main/java/org/omnifaces/util/ComponentsLocal.java#L316) .

Perhaps just adding the `ResourceHandler.isResourceRendered` is also enough already.

image

melloware:
Fix #12855: HeadRenderer check for already loaded resource

@blutorange can you test if this is enough?

### DIFF
--- a/primefaces/src/main/java/org/primefaces/renderkit/HeadRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/renderkit/HeadRenderer.java
@@ -44,6 +44,7 @@ import javax.el.ValueExpression;
 import javax.faces.FacesException;
 import javax.faces.application.ProjectStage;
 import javax.faces.application.Resource;
+import javax.faces.application.ResourceHandler;
 import javax.faces.component.UIComponent;
 import javax.faces.component.UIViewRoot;
 import javax.faces.context.ExternalContext;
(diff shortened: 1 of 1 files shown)


blutorange:
Yeah, the check seems to be enough for the initial page load. 

If I understand it correctly, `@ResourceDependency` from components in the view tree are always included, regardless of whether the component is `rendered=false/true`.

But you can still have something like this with JSTL tags

```xml
<h:form>
  <c:if test="#{bean.show}">
    <p:timeline />
  </c:if>
  <p:commandButton actionListener=#{bean.doShow()}  update="@form" />
</h:form>
```

That still loads moment.js twice when I try it. Not sure if we need to support that, but there are probably people out there doing something like that.

Example using the showcase where it still loads it twice:

[primefaces/primefaces-showcase/src/main/webapp/ui/data/timeline/basic.xhtml](https://github.com/user-attachments/files/17703991/basic.txt)

melloware:
Yeah the different lifecycle of JSTL vs JSF that is a tougher one.

PR #12873, opened 2024-11-11
Fix #12870: OverlayPanel allow showEvent 'none'

### LINKED ISSUE #12870: OverlayPanel: will be visible even if showEvent is set to none
### Describe the bug

p:overlayPanel is visible even its showEvent is set to none.
This happen cause i hit enter inside an Input in this case an p:autoComplete.

i would like to show the OverlayPanel only if a condition is met with primefaces.executeScript("PF('overlay').show()").

### Reproducer

Just type something inside the autocomplete and press enter on your keyboard.
This will make the overlayPanel  visible.

[primefaces-test-master.zip](https://github.com/user-attachments/files/17707029/primefaces-test-master.zip)

### Expected behavior

If the showEvent of an overlayPanel is set to none i would expect that it will be never visible.
Or only, if i call via JS PF('overlayPanel').show();

### PrimeFaces edition

None

### PrimeFaces version

14.0.0

### Theme

_No response_

### JSF implementation

All

### JSF version

latest

### Java version

17

### Browser(s)

_No response_

melloware:
Fix #12870: OverlayPanel allow showEvent 'none'

### DIFF
--- a/docs/15_0_0/components/overlaypanel.md
+++ b/docs/15_0_0/components/overlaypanel.md
@@ -38,7 +38,7 @@ onHide | null | String | Client side callback to execute when panel is hidden.
 onShow | null | String | Client side callback to execute when panel is shown.
 showCloseIcon | false | Boolean | Displays a close icon to hide the overlay, default is false.
 showDelay | 0 | Integer | Delay time to show overlay in milliseconds.
-showEvent | click | String | Event on target to show the panel.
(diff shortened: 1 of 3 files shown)


tandraschko:
There are many space changes 

Our default always was "function()" without space between

melloware:
ahhh that is a mistake let me fix.

PR #12878, opened 2024-11-12
Fix #12484: Picklist allow custom labels per component

### LINKED ISSUE #12484: pickList: Customization of labels for the buttons not working any more
### Describe the bug

I use pickList in my application. In version 13.0.7 I could overwrite the labels as described in https://stackoverflow.com/questions/70086529/customize-the-text-of-the-picklist-buttons-of-primefaces :

<p:pickList ...
            addLabel="Voeg toe"
            addAllLabel="Voeg alles toe"
            removeLabel="Verwijder"
            removeAllLabel="Verwijder alles"
            labelDisplay="inline"/>
            
This does not work any more with primefaces version 14.0.3

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

Community

### PrimeFaces version

14.0.3

### Theme

Saga

### JSF implementation

Mojarra

### JSF version

4.0.1

### Java version

17

### Browser(s)

Chrome 127.0.6533.120 (Official Build) (64-bit)

melloware:
Fix #12484: Picklist allow custom labels per component

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.widget.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.widget.js
@@ -608,6 +608,32 @@ if (!PrimeFaces.widget) {
             }
             
             return this.cfg.formId;
+        },
+
+        /**
+         * Gets a localized label text for this widget.
+         * @param {string} label The label key to look up
+         * @returns {string} The label text - either from the widget configuration if specified,
+         * or from the PrimeFaces global labels
+         */
+        getLabel: function(label) {
+            if (this.cfg.labels && this.cfg.labels[label]) {
+                return this.cfg.labels[label];
+            }
+            return PrimeFaces.getLocaleLabel(label);
+        },
+        
+        /**
+         * Creates an ARIA label for an element.
+         * @param {string} label The label key to look up
+         * @returns {string} The ARIA label text - either from the widget configuration if specified, 
+         * or from the PrimeFaces global ARIA labels
+         */
+        getAriaLabel: function(label) {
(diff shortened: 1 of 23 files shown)


tandraschko:
Let me think about it
I would like to see a more common solution, useable by other components

tandraschko:
i think about that each widget could have:

```
cfg.labels = {
....
}
cfg.label.aria = {
....
}
```

and you move getLabel/getAriaLabel to the BaseWidget

so the mechanism is the same, just a global pattern

melloware:
I LOVE IT!!!  Let me work on that.

melloware:
@tandraschko if this looks good let me know i will fix all the other widgets that get labels or aria labels.

I chose in the preconstruct to initialize the `labels ={}` and `aria={}` so if someone does this ..

```xml
<f:attribute name="widgetPreConstruct" value="cfg.labels.aria.moveUp = 'Slide Up';" />
```

they would get an NPE on labels or aria

primefaces/src/main/resources/META-INF/resources/primefaces/core/core.widget.js
```diff
@@ -608,6 +608,32 @@ if (!PrimeFaces.widget) {
...
             return this.cfg.formId;
+        },
+
+        /**
+         * Gets a localized label text for this widget.
+         * @param {string} label The label key to look up
+         * @returns {string} The label text - either from the widget configuration if specified,
+         * or from the PrimeFaces global labels
+         */
+        getLocaleLabel: function(label) {
```
tandraschko: should we just call it getLabel?

both (non-aria and aria) are locale-labels :D just a different category

tandraschko:
see my comment; otherwise it looks good

melloware:
All changes made and all IT's passing!

PR #12881, opened 2024-11-13
Fix #12853: Datatable filter toggler via js

### LINKED ISSUE #12853: Datatable: Toggler to show/hide filter input components
### Description

During my migration of our primefaces project I'm going through our extensions/enhancements and try to mainstream/baseline them as much as possible. One of our extensions is a javascript toggler to show/hide (fade in/ fade out) filter components in a datatable.
For pages where filtering is the exception and not the regular case we tend to keep the initial ui as clean and simple as possible.
Meaning all the filter fields are hidden, but can be shown by clicking on a toggle.

The whole logic happens only via js/css. So the fields will be rendered as usually. Default filters are still taken into account, they are just not visible.

See video for actual behaviour:
https://github.com/user-attachments/assets/4f74caa9-eebb-43c6-9cd9-d784caf5382f

The toggle buttons can either be bound via a trigger attribute in the datatable itself or  called manually on any component via a Javascript Widget method.

Is this something you might want to add to PF? If yes I can provide a PR with the changes.

### Additional context

_No response_

gako:
Fix #12853


primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
```diff
@@ -5751,6 +5751,29 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
         this.bodyTable.css('top', '0px');
         this.scrollBody.scrollTop(0);
         this.clearScrollState();
+    },
+
+    /**
+     * In case of a toggleable filter datatable toggle all filter input components
+     * @param {string | null} speed speed of fade animation. (see jquery fadeToggle method)
+     * @param {function | null} callback will be executed after animation is finished. (see jquery fadeToggle method)
```
melloware:
```suggestion
     * @param {() => void | null} callback will be executed after animation is finished. (see jquery fadeToggle method)
```

primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
```diff
@@ -5751,6 +5751,29 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
         this.bodyTable.css('top', '0px');
         this.scrollBody.scrollTop(0);
         this.clearScrollState();
+    },
+
+    /**
+     * In case of a toggleable filter datatable toggle all filter input components
+     * @param {string | null} speed speed of fade animation. (see jquery fadeToggle method)
+     * @param {() => void | null} callback will be executed after animation is finished. (see jquery fadeToggle method)
```
melloware:
```suggestion
     * @param {(() => void) | null} callback will be executed after animation is finished. (see jquery fadeToggle method)
```

primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.js
```diff
@@ -5751,6 +5751,29 @@ PrimeFaces.widget.DataTable = PrimeFaces.widget.DeferredWidget.extend({
...
+    toggleFilter: function(speed, callback) {
+        this.jq.find(".ui-column-filter, .ui-column-customfilter").fadeToggle(speed || 0, callback);
+    },
+
+    /**
+     * Bind click event to given jq element to show/hide filter components. Toggles between pi-filter and pi-filter-slash icon if used.
+     * @param {JQuery} element jquery object to bind click events to
+     * */
+    bindFilterToggleEvents: function(element) {
+        var _self = this;
```
melloware: We typically us `$this` instead of `_self`

melloware (changes requested):

tandraschko:
there should be a JS API section in the datatable.md

melloware (approved):

tandraschko:
TBH i would just remove bindFilterToggleEvents and only provide 1 binding method
providing 2 similar APIs is just not required

tandraschko:
and IMO the manual call is the better one as binding requires some init code, which will break when the DT is updated and so on. toggleFilter is more safe and intuitive

primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.css
```diff
@@ -68,6 +68,11 @@
     margin: .5rem auto auto auto;
 }
 
+.ui-datatable.ui-filter-togglable thead .ui-column-customfilter,
+.ui-datatable.ui-filter-togglable thead .ui-column-filter {
+    display:none;
+}
```
melloware, line 75: Doesn't this style need to be applied somewhere?

melloware (changes requested):

primefaces/src/main/resources/META-INF/resources/primefaces/datatable/datatable.css
gako: yes, if you want the column filter to be hidden initially on page load you have to set the stylev 'ui-filter-togglable' in the datatable itself.

melloware (approved):

melloware:
Thanks @gako this is great!

melloware:
@gako you can see it in action here: https://quarkus-faces-melloware-8a6a34c1.koyeb.app/ui/data/datatable/filter.xhtml

nice work.

PR #12900, opened 2024-11-18
Fix #12887: Widget callBehavior fall back to DOM `on` if no behaviors

### LINKED ISSUE #12887: InputText: Missing client behaviors
### Describe the bug

Reported on Stack Overflow: https://stackoverflow.com/questions/79189454/no-behavior-available-in-primefaces-widget

I want to fill an Primefaces inputText field by Javascript and trigger then the change event. In my understanding of the documentation of the inputText widget the [callBehavior](https://primefaces.github.io/primefaces/jsdocs/classes/src_PrimeFaces.PrimeFaces.widget.InputText.html#callBehavior) function would be the right way.

So my field looks like this:

```xml
<p:inputText id="myInput" value="#{myBean.value}">
   <p:ajax event="change" update="@this" onstart="console.log('Change triggered')" />
</p:inputText>
```

My JS function looks like this

```js
function myFun() {
   const theId = // Have to get the id dynamically
   const inputWdgt = PrimeFaces.getWidgetById(theId);
   inputWdgt.jq.val('My new value');
   inputWdgt.callBehavior('change');
}
```
But in result neither the log in the onstart is printed nor the update is executed.

In my further investigation I saw, that the cfg object in the browser doesn't contain any behaviors. Hence it is obvious, that the callBehavior has no effect (since the function relates to these). So the question would be, why are there no behaviors defined for the widget?

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

Community

### PrimeFaces version

14.0.7

### Theme

ALL

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

11

### Browser(s)

ALL

melloware:
Fix #12887: Widget callBehavior fall back to DOM `on` if no behaviors

@tandraschko this works if NO behaviors found we check the `this.jq` for an `on` event and call it if found.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/inputtext/InputTextRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/inputtext/InputTextRenderer.java
@@ -84,8 +84,6 @@ public class InputTextRenderer extends InputRenderer {
             }
         }
 
-        encodeClientBehaviors(context, inputText);
-
         wb.finish();
     }
 
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.widget.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.widget.js
@@ -505,6 +505,10 @@ if (!PrimeFaces.widget) {
             if(this.hasBehavior(event)) {
                 this.cfg.behaviors[event].call(this, ext);
             }
+            else if (this.cfg.behaviors === undefined && this.jq.length > 0) {
+                // #12887 if no behavior is defined, try to call the DOM event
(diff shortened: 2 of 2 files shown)


tandraschko:
is this.cfg.behaviors really undefined per default?
that would be great

melloware:
Yep that works too!

tandraschko:
perfect
did you also check my first comment?

melloware:
Switched to `else if (this.cfg.behaviors === undefined && this.jq.length > 0) {`

fcorneli:
Nice!

PR #12905, opened 2024-11-18
Fix #12904: MessageFactory pass in FacesContext

### LINKED ISSUE #12904: MessageFactory: Pass FacesContext into all methods
### Description

Right now MessageFactory does `FacesContext.currentInstance()` and it should really take FacesContext as a parameter.

pass the FacesContext

### Additional context

_No response_

melloware:
Fix #12904: MessageFactory pass in FacesContext


tandraschko:
it should be also to remove your static mock stuff now

melloware:
Just removed it.

PR #12935, opened 2024-11-22
Fix #12933: FileUpload rename directoryPath and fix path check

### LINKED ISSUE #12933: UploadedFile#write broken
the param is currently called filePath, but its actually the directory. If yes we should return the whole filepath then and rename the param to directory.
we should also check docs

melloware:
Fix #12933: FileUpload rename directoryPath and fix path check


tandraschko:
Should we return the File instead of void?

melloware:
probably a good idea!

melloware:
Updated

tandraschko:
Thanks!

PR #12946, opened 2024-11-25
Fix #12941: 14.0.8 Simple FileUpload: support dropZone

### LINKED ISSUE #12941: Simple FileUpload: support dropZone

melloware:
Fix #12941: 14.0.8 Simple FileUpload: support dropZone

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/fileupload/FileUploadRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/fileupload/FileUploadRenderer.java
@@ -32,12 +32,13 @@ import org.primefaces.util.StyleClassBuilder;
 import org.primefaces.util.WidgetBuilder;
 import org.primefaces.validate.FileValidator;
 
+import java.io.IOException;
+import java.util.Arrays;
+
 import javax.faces.component.UIComponent;
 import javax.faces.context.FacesContext;
 import javax.faces.context.ResponseWriter;
 import javax.faces.convert.ConverterException;
-import java.io.IOException;
-import java.util.Arrays;
 
 public class FileUploadRenderer extends CoreRenderer {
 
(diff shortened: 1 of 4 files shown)


primefaces/src/main/java/org/primefaces/component/fileupload/FileUploadRenderer.java
```diff
@@ -64,12 +65,6 @@ public void decode(FacesContext context, UIComponent component) {
     public void encodeEnd(FacesContext context, UIComponent component) throws IOException {
         FileUpload fileUpload = (FileUpload) component;
 
-        if (fileUpload.getDropZone() != null) {
```
jepsar, line 67: We are back porting a breaking change which is not even documented?

melloware: @tandraschko is this a breaking change?  I thought this was only ADDING functionality to SimpleFileUpload?

melloware: Oh I see previously if you had a DROPZONE it assumed you wanted Advanced mode.

jepsar: Yes, there is a note on it in the migration guide in Thomas' PR. I think we can leave this block in.

melloware: OK in Main we would need to add the same note in the 14.0.0 Migration Guide because adding it here won't do any good the docs are only built from Main!

tandraschko: We need to add this to a 14_x migration guide

melloware: Yep i will take care of it

PR #12950, opened 2024-11-25
Fix #12949: HeadRenderer use resources not links

### LINKED ISSUE #12949: HeadRenderer: Prevent duplicates
### Description

HeadRenderer has been in PF since the beginning and it actually quite clever to support `facets` to allow loading order.  Its also vital to ensure scripts are loaded properly in order so PrimeFaces does not break.

**Issues with current implementation:**
- https://github.com/primefaces/primefaces/issues/12855 scripts are loaded as `links` and not `resources` so duplicates are being added to the DOM in some scenarios
- Because they are not resources they are not picked up by OmniFaces CombinedResourceHandler.  BalusC wrote about this years ago: https://github.com/omnifaces/omnifaces/wiki/Combine-hardcoded-PrimeFaces-resources-using-CombinedResourceHandler
- PFE added the code based on BalusC article to make it work but its a HACK!: https://www.primefaces.org/showcase-ext/sections/utils/combinedResourceHandler.jsf

cc @stolp @blutorange 

Currently here is the ordered output of PrimeFaces and has been for years with a mixture of `link` and `resource` dependencies.  The links were used to ensure the proper order in the `<head>` .

**DatePicker Example from Showcase:**
```
link: theme.css
link: primeicons/primeicons.css
resource: css/primeflex3.css
resource: css/flags/flags.css
resource: components.css
resource: jquery/jquery.js
resource: jquery/jquery-plugins.js
resource: core.js
resource: components.js
resource: inputmask/inputmask.js
resource: datepicker/datepicker.js
resource: touch/touchswipe.js
resource: css/layout/layout.css
link: moment/moment.js
link: validation/validation.bv.js
link: locales/locale-en.js
```

I would like the HeadRenderer to render `resource` not `links` to allow OmniFaces to work properly as well as follow Faces standards.

**Expected Outcome:**
```
resource: theme.css
resource: primeicons/primeicons.css
resource: css/primeflex3.css
resource: css/flags/flags.css
resource: components.css
resource: jquery/jquery.js
resource: jquery/jquery-plugins.js
resource: core.js
resource: components.js
resource: inputmask/inputmask.js
resource: datepicker/datepicker.js
resource: touch/touchswipe.js
resource: css/layout/layout.css
resource: moment/moment.js
resource: validation/validation.bv.js
resource: locales/locale-en.js
```

### Additional context

_No response_

melloware:
Fix #12949: HeadRenderer use resources not links

Opened this as draft for discussion:

1. Replace links with resources.
2. Now that resources are pulled we need to ensure the order so use an ArrayList so we can sort things to top or later bottom if we need to before applying.   
3. Currently adding the CSS needs to be moved back to the top of the list
4. Adding the JS adds it properly to the bottom of the list so I just had to move the code up before processing
5. Took ideas from `OmniFaces` how it adds resources to `head` or `body` etc and its got some nice features like uniquely giving ID's to components if missing and ensuring no duplicates are added.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/renderkit/HeadRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/renderkit/HeadRenderer.java
@@ -30,8 +30,10 @@ import org.primefaces.context.PrimeRequestContext;
 import org.primefaces.util.FacetUtils;
 import org.primefaces.util.LocaleUtils;
 import org.primefaces.util.MapBuilder;
+import org.primefaces.util.ResourceUtils;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -43,11 +45,8 @@ import javax.el.ExpressionFactory;
 import javax.el.ValueExpression;
 import javax.faces.FacesException;
 import javax.faces.application.ProjectStage;
-import javax.faces.application.Resource;
-import javax.faces.application.ResourceHandler;
 import javax.faces.component.UIComponent;
 import javax.faces.component.UIViewRoot;
-import javax.faces.context.ExternalContext;
 import javax.faces.context.FacesContext;
 import javax.faces.context.ResponseWriter;
 import javax.faces.lifecycle.ClientWindow;
@@ -59,7 +58,7 @@ import javax.servlet.http.HttpServletResponse;
  * Renders head content based on the following order
  * - First Facet
  * - Theme CSS
- * - FontAwesome
+ * - PrimeIcons CSS
  * - Middle Facet
  * - Registered Resources
  * - Client Validation Scripts
@@ -129,14 +128,6 @@ public class HeadRenderer extends Renderer {
             middle.encodeAll(context);
         }
 
-        //Registered Resources
-        UIViewRoot viewRoot = context.getViewRoot();
-        List<UIComponent> resources = viewRoot.getComponentResources(context, "head");
-        for (int i = 0; i < resources.size(); i++) {
-            UIComponent resource = resources.get(i);
-            resource.encodeAll(context);
-        }
-
         if (applicationContext.getConfig().isClientSideValidationEnabled()) {
             // moment is needed for Date validation
             encodeJS(context, LIBRARY, "moment/moment.js");
@@ -160,6 +151,17 @@ public class HeadRenderer extends Renderer {
             }
         }
 
+        //Registered Resources
+        UIViewRoot viewRoot = context.getViewRoot();
+        List<UIComponent> resources = new ArrayList<>(viewRoot.getComponentResources(context, "head"));
+        moveResourceToTop(resources, "primeicons/primeicons.css");
+        moveResourceToTop(resources, "theme.css");
+        for (int i = 0; i < resources.size(); i++) {
+            UIComponent resource = resources.get(i);
+            LOGGER.log(Level.FINE, () -> "HeadRenderer resource: " + resource.getAttributes().get("name"));
+            resource.encodeAll(context);
+        }
+
         encodeSettingScripts(context, applicationContext, requestContext, writer);
 
         // encode initialization scripts
@@ -180,47 +182,11 @@ public class HeadRenderer extends Renderer {
     }
 
     protected void encodeCSS(FacesContext context, String library, String resource) throws IOException {
-        ResourceHandler resourceHandler = context.getApplication().getResourceHandler();
-        if (resourceHandler.isResourceRendered(context, resource, library)) {
-            // resource already rendered, skip
-            return;
-        }
-
-        ResponseWriter writer = context.getResponseWriter();
-        ExternalContext externalContext = context.getExternalContext();
-
-        Resource cssResource = resourceHandler.createResource(resource, library);
-        if (cssResource == null) {
-            throw new FacesException("Error loading CSS, cannot find \"" + resource + "\" resource of \"" + library + "\" library");
-        }
-        else {
-            writer.startElement("link", null);
-            writer.writeAttribute("type", "text/css", null);
-            writer.writeAttribute("rel", "stylesheet", null);
-            writer.writeAttribute("href", externalContext.encodeResourceURL(cssResource.getRequestPath()), null);
-            writer.endElement("link");
-        }
+        ResourceUtils.addStyleSheetResource(context, library, resource);
     }
 
     protected void encodeJS(FacesContext context, String library, String script) throws IOException {
-        ResourceHandler resourceHandler = context.getApplication().getResourceHandler();
-        if (resourceHandler.isResourceRendered(context, script, library)) {
-            // resource already rendered, skip
-            return;
-        }
-
-        ResponseWriter writer = context.getResponseWriter();
-        ExternalContext externalContext = context.getExternalContext();
-        Resource resource = resourceHandler.createResource(script, library);
-
-        if (resource == null) {
-            throw new FacesException("Error loading JavaScript, cannot find \"" + script + "\" resource of \"" + library + "\" library");
-        }
-        else {
-            writer.startElement("script", null);
-            writer.writeAttribute("src", externalContext.encodeResourceURL(resource.getRequestPath()), null);
-            writer.endElement("script");
-        }
+        ResourceUtils.addJavascriptResource(context, library, script);
     }
 
     protected void encodeSettingScripts(FacesContext context, PrimeApplicationContext applicationContext, PrimeRequestContext requestContext,
@@ -312,5 +278,22 @@ public class HeadRenderer extends Renderer {
 
             writer.endElement("script");
         }
+    }
+
+    /**
+     * Moves a resource component to the top of the resources list based on its name.
+     * This is used to ensure certain resources like theme.css and primeicons.css are loaded first.
+     *
+     * @param resources The list of UIComponent resources to sort
+     * @param resource The resource name suffix to move to the top
+     */
+    protected void moveResourceToTop(List<UIComponent> resources, String resource) {
+        resources.sort((a, b) -> {
+            String nameA = (String) a.getAttributes().get("name");
+            String nameB = (String) b.getAttributes().get("name");
+            if (nameA != null && nameA.endsWith(resource)) return -1;
+            if (nameB != null && nameB.endsWith(resource)) return 1;
+            return 0;
+        });
     }
 }
--- a/primefaces/src/main/java/org/primefaces/renderkit/HeadRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/renderkit/HeadRenderer.java
@@ -29,13 +29,16 @@ import org.primefaces.context.PrimeApplicationContext;
 import org.primefaces.context.PrimeRequestContext;
 import org.primefaces.util.FacetUtils;
 import org.primefaces.util.LocaleUtils;
+import org.primefaces.util.ResourceUtils;
 
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import javax.faces.FacesException;
 import javax.faces.application.ProjectStage;
 import javax.faces.component.UIComponent;
 import javax.faces.component.UIViewRoot;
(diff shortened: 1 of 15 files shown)


primefaces/src/main/java/org/primefaces/util/ResourceUtils.java
```diff
@@ -335,4 +320,197 @@ public static Resource evaluateResourceExpression(FacesContext facesContext, Val
...
+    }
+
+    /**
+     * Sets a unique ID for a component resource, handling MyFaces-specific ID generation. MyFaces requires a special flag to be set during ID generation for
+     * resources.
+     *
+     * @param context The FacesContext
+     * @param resource The component resource needing a unique ID
+     */
+    public static void setComponentResourceUniqueId(FacesContext context, UIComponent resource) {
```
tandraschko: why is this necessary at all?
in the past we didnt set a ID in this methods and it worked fine?

primefaces/src/main/java/org/primefaces/renderkit/HeadRenderer.java
```diff
@@ -180,47 +182,11 @@ public void encodeEnd(FacesContext context, UIComponent component) throws IOExce
...
-            throw new FacesException("Error loading CSS, cannot find \"" + resource + "\" resource of \"" + library + "\" library");
-        }
-        else {
-            writer.startElement("link", null);
-            writer.writeAttribute("type", "text/css", null);
-            writer.writeAttribute("rel", "stylesheet", null);
-            writer.writeAttribute("href", externalContext.encodeResourceURL(cssResource.getRequestPath()), null);
-            writer.endElement("link");
-        }
+        ResourceUtils.addStyleSheetResource(context, library, resource);
```
tandraschko: IMO adding real JSF resources during RENDER_RESPONSE sounds a bit late
why not system listener or phase listener?

primefaces/src/main/java/org/primefaces/util/ResourceUtils.java
tandraschko: in general such impl specific hacks are very dangerous and we would need to check it in myfaces first

primefaces/src/main/java/org/primefaces/util/ResourceUtils.java
```diff
@@ -335,4 +320,197 @@ public static Resource evaluateResourceExpression(FacesContext facesContext, Val
...
+     * @param type The renderer type
+     * @param libraryName The library name
+     * @param resourceName The resource name
+     * @param target The target location ("head" or "body")
+     * @return The added or existing UIComponent
+     */
+    public static UIComponent addScriptResourceToTarget(FacesContext context, String type, String libraryName, String resourceName, String target) {
+        String id = (libraryName != null ? libraryName.replaceAll("\\W+", "_") + "_" : "") + resourceName.replaceAll("\\W+", "_");
+
+        // Check for existing resource with same ID
```
tandraschko: i would really like to check JSF impl internals, if we really add a resource a second time there, if its already in the list
i dont think so, otherwise for me its a impl bug

primefaces/src/main/java/org/primefaces/util/ResourceUtils.java
melloware: Agreed I grabbed this directly from OmniFaces as this is how BalusC was doing it. Also a note our PF Extensions code had this as well..

```java
        // https://github.com/primefaces-extensions/primefaces-extensions/issues/517
        // MyFaces needs ID set to prevent duplicate ID check in Dev mode, Mojarra does not
        final PrimeEnvironment environment = PrimeApplicationContext.getCurrentInstance(context).getEnvironment();
        if (!environment.isMojarra()) {
            js.setId("primfaces-script-processor");
        }
```

I worked with @volosied in MyFaces trying to resolve this a a couple of years ago.  We had a long back and forth here about it: https://github.com/primefaces-extensions/primefaces-extensions/issues/517

primefaces/src/main/java/org/primefaces/renderkit/HeadRenderer.java
melloware: I tried both a SystemEventListener and a PhaseListener and the order was never correct.   Depending on where things were added locale could be added before `core.js` thus not loading the locales because `window.PrimeFaces` did not exist yet.  

Updating the HeadRenderer this way felt the "safest" and most backwards compatible to ensure if people are using the `<f:facet name="first" />` stuff it would all work as it does now it just no longer uses `<link>`

primefaces/src/main/java/org/primefaces/util/ResourceUtils.java
melloware: This was also in OmniFaces and it may not be necessary anymore but it didn't seem to hurt?

primefaces/src/main/java/org/primefaces/renderkit/HeadRenderer.java
melloware: That being said I was using `UiViewRoot` with `PostAddToView` so not sure if that was the right SystemComponentListener place to do it?

primefaces/src/main/java/org/primefaces/util/ResourceUtils.java
tandraschko: im just not happy to move a hack from one place to another place

IMO we should check first aht both impls are NOT adding the component again when we call:
context.getViewRoot().addComponentResource(context, resource, target);
without given the component a ID

melloware: I will test it.

tandraschko: We can add Integration Test for it, it runs on different versions

melloware:
OK Integration Test is passing where I check the order of all head elements and I specifically added duplicates and ensured no duplicates are created. Now I am going to remove all the hacks.

melloware:
All tests passing for both MyFaces and Mojarra with all hacks removed!

tandraschko:
i dont think the test cover this:
moment JS is added by a component, added by the user manually, added by the HeadRenderer

is really only 1 moment.js in the final markup?

primefaces/src/main/java/org/primefaces/renderkit/HeadRenderer.java
```diff
@@ -160,6 +151,17 @@ public void encodeBegin(FacesContext context, UIComponent component) throws IOEx
...
         }
 
+        //Registered Resources
+        UIViewRoot viewRoot = context.getViewRoot();
+        List<UIComponent> resources = new ArrayList<>(viewRoot.getComponentResources(context, "head"));
+        moveResourceToTop(resources, "primeicons/primeicons.css");
+        moveResourceToTop(resources, "theme.css");
+        for (int i = 0; i < resources.size(); i++) {
+            UIComponent resource = resources.get(i);
+            LOGGER.log(Level.FINE, "HeadRenderer resource: " + resource.getAttributes().get("name"));
```
tandraschko: user lambda or remove, currently the string is always concatinated

primefaces/src/main/java/org/primefaces/renderkit/HeadRenderer.java
```diff
@@ -59,7 +58,7 @@
  * Renders head content based on the following order
  * - First Facet
  * - Theme CSS
- * - FontAwesome
+ * - PrimeIcons CSS
```
tandraschko, line 56: is the whole list and order still correct?

melloware:
I will add the moment.js test today

primefaces/src/main/java/org/primefaces/renderkit/HeadRenderer.java
melloware: Yes the order is still correct

primefaces/src/main/java/org/primefaces/renderkit/HeadRenderer.java
melloware: Changed to Lambda

melloware:
@tandraschko added two new tests...

1. 002 tests the scripts and order on page load to esnure no duplicates including moment.js with Chart.js component
2. 003 tests a dynamic load of a component after page load `c:if` and checks the scripts BEFORE and AFTER and ensures the only new script that is added to the page was Charts.js as Moment.js was already loaded.

**Note:**
BalusC is clever I had to put back this block..

```java
String id = (libraryName != null ? libraryName.replaceAll("\\W+", "_") + "_" : "") + resourceName.replaceAll("\\W+", "_");

        // Check for existing resource with same ID which may be located in the head or body
        // MOVE_SCRIPTS_TO_BOTTOM may have moved a resource out of the head so we need to check both locations
        for (UIComponent existingResource : context.getViewRoot().getComponentResources(context)) {
            if (id.equals(existingResource.getId())) {
                return existingResource;
            }
        }
```

This is because in Test 003 it was duplicating moment.js.  Why?  So the check for moment.js `getResourceHandler().isResourceRendered(context, resourceName, libraryName)` is reporting `false` and adding moment.js again. I am not sure why it is doing it in both MyFaces and Mojarra.

So BalusC code scans both the HEAD and the BODY and because our MOVE_SCRIPTS_TO_BOTTOM had moment.js in the body it finds it and determines NOT to add it again.

So now the tests prove everything is working but we will have to ask @BalusC why `getResourceHandler().isResourceRendered(context, resourceName, libraryName)` reports false in IT 003.

tandraschko:
then the question is: why does addComponentResource add it a second time?
i think we need to dig deeper

tandraschko:
dont get me wrong
i just dont like to add imp specific hacks in PF, without any direct benefit here in PF

either we do it correct, or better leave the hack e.g. in PFE

melloware:
well its not impl specific both MyFaces and Mojarra fail without that check?

melloware:
could it be a gap or JSTL related?

tandraschko:
I mean
You have also multiple components with jquery dependency and you only have one rendered
That works already fine

melloware:
ok let me add some more debug and dump out what's going on.

tandraschko:
also see
https://github.com/apache/myfaces/blob/535b157a906e452f6de57a998495beeef1314ac3/api/src/main/java/jakarta/faces/component/UIViewRoot.java#L225

we auto set a ID here

melloware:
OK I removed the hack and setting the ID and I think everything is working now?   here is the output.

**HeadRenderer runs once:**
```
ResourceUtils.addResource - type: javax.faces.resource.Stylesheet, library: primefaces-saga-blue, resource: theme.css, isRendered: false
ResourceUtils.addResource - type: javax.faces.resource.Stylesheet, library: primefaces, resource: primeicons/primeicons.css, isRendered: false
ResourceUtils.addResource - type: javax.faces.resource.Script, library: primefaces, resource: moment/moment.js, isRendered: false
ResourceUtils.addResource - type: javax.faces.resource.Script, library: primefaces, resource: validation/validation.bv.js, isRendered: false
ResourceUtils.addResource - type: javax.faces.resource.Script, library: primefaces, resource: locales/locale-en.js, isRendered: false
Nov 27, 2024 11:27:42 AM org.apache.myfaces.util.ExternalSpecifications lambda$static$3
INFO: MyFaces Core Servlet 4.0 support enabled
```

**PAge Load:**
```
Head Script sources: 0
Body Script sources: 7
/javax.faces.resource/jquery/jquery.js.xhtml;jsessionid=2A6372897600A673E7A9080E65A5F529?ln=primefaces&v=15.0.0-SNAPSHOT&v=15.0.0-SNAPSHOT
/javax.faces.resource/jquery/jquery-plugins.js.xhtml;jsessionid=2A6372897600A673E7A9080E65A5F529?ln=primefaces&v=15.0.0-SNAPSHOT&v=15.0.0-SNAPSHOT
/javax.faces.resource/core.js.xhtml;jsessionid=2A6372897600A673E7A9080E65A5F529?ln=primefaces&v=15.0.0-SNAPSHOT&v=15.0.0-SNAPSHOT
/javax.faces.resource/components.js.xhtml;jsessionid=2A6372897600A673E7A9080E65A5F529?ln=primefaces&v=15.0.0-SNAPSHOT&v=15.0.0-SNAPSHOT
/javax.faces.resource/moment/moment.js.xhtml;jsessionid=2A6372897600A673E7A9080E65A5F529?ln=primefaces&v=15.0.0-SNAPSHOT&v=15.0.0-SNAPSHOT
/javax.faces.resource/validation/validation.bv.js.xhtml;jsessionid=2A6372897600A673E7A9080E65A5F529?ln=primefaces&v=15.0.0-SNAPSHOT&v=15.0.0-SNAPSHOT
/javax.faces.resource/locales/locale-en.js.xhtml;jsessionid=2A6372897600A673E7A9080E65A5F529?ln=primefaces&v=15.0.0-SNAPSHOT&v=15.0.0-SNAPSHOT
```

**After** Component is displayed only Chart.JS is added correctly.

```
Head Script sources: 1
/javax.faces.resource/chart/chart.js.xhtml?ln=primefaces&v=15.0.0-SNAPSHOT&v=15.0.0-SNAPSHOT
Body Script sources: 7
/javax.faces.resource/jquery/jquery.js.xhtml;jsessionid=2A6372897600A673E7A9080E65A5F529?ln=primefaces&v=15.0.0-SNAPSHOT&v=15.0.0-SNAPSHOT
/javax.faces.resource/jquery/jquery-plugins.js.xhtml;jsessionid=2A6372897600A673E7A9080E65A5F529?ln=primefaces&v=15.0.0-SNAPSHOT&v=15.0.0-SNAPSHOT
/javax.faces.resource/core.js.xhtml;jsessionid=2A6372897600A673E7A9080E65A5F529?ln=primefaces&v=15.0.0-SNAPSHOT&v=15.0.0-SNAPSHOT
/javax.faces.resource/components.js.xhtml;jsessionid=2A6372897600A673E7A9080E65A5F529?ln=primefaces&v=15.0.0-SNAPSHOT&v=15.0.0-SNAPSHOT
/javax.faces.resource/moment/moment.js.xhtml;jsessionid=2A6372897600A673E7A9080E65A5F529?ln=primefaces&v=15.0.0-SNAPSHOT&v=15.0.0-SNAPSHOT
/javax.faces.resource/validation/validation.bv.js.xhtml;jsessionid=2A6372897600A673E7A9080E65A5F529?ln=primefaces&v=15.0.0-SNAPSHOT&v=15.0.0-SNAPSHOT
/javax.faces.resource/locales/locale-en.js.xhtml;jsessionid=2A6372897600A673E7A9080E65A5F529?ln=primefaces&v=15.0.0-SNAPSHOT&v=15.0.0-SNAPSHOT
```

melloware:
all IT's passing with expected results with all hacks removed?   I have a feel while running the IT's and rebuilding PF jar over and over i must have had an inconsistent state which threw me off.

melloware:
Testing this with OmniFaces and found one issue I am working on.

PR #12954, opened 2024-11-25
Fix #12953: SelectOneRadio proper identification by label

### LINKED ISSUE #12953: SelectOneRadio: aria-labelledby attribute is using text instead of the ID.
### Describe the bug

According to the aria-labelledby specification, its value should be:

"The id or space-separated list of element ids that describe the current element."
(Source: [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby)).

But the SelectOneRadio is using the text instead of the element ID.

However, this issue cannot be reproduced in the showcase because it uses `<h5>` instead of `<p:outputLabel>`.

Other option is use just the aria-label instead of aria-labelledby. Remembering that "aria-labelledby takes precedence over all other methods of providing an accessible name, including aria-label, `<label>`, and the element's inner text."

### Reproducer

```
<p:outputLabel value="Label Text" for="idRadio"/>
<p:selectOneRadio id="idRadio"
				value="#{controller.value}">
				<f:selectItem itemLabel="Yes" itemValue="0" />
				<f:selectItem itemLabel="No" itemValue="1" />
</p:selectOneRadio>
```

**Actual Output**
The generated HTML is as follows:
`<div id="form:idRadio" role="radiogroup" aria-labelledby="Label Text" ...>
`

### Expected behavior

The expected HTML should be:
`<div id="form:idRadio" role="radiogroup" aria-labelledby="idRadio" ...>
`

### PrimeFaces edition

Community

### PrimeFaces version

14.0.8

### Theme

custom

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

21

### Browser(s)

_No response_

melloware:
Fix #12953: SelectOneRadio proper identification by label

This was incorrectly just pointing to text instead of the ID of the `p:outputLabel`.  That is what `InputHolder` is for so `p:outputLabel` can set the corrected `LabelledBy` element.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/selectoneradio/SelectOneRadio.java
+++ b/primefaces/src/main/java/org/primefaces/component/selectoneradio/SelectOneRadio.java
@@ -47,4 +47,24 @@ public class SelectOneRadio extends SelectOneRadioBase {
     public String getRadioButtonId(FacesContext context, int index) {
         return this.getClientId(context) + UINamingContainer.getSeparatorChar(context) + index;
     }
+
+    @Override
+    public String getInputClientId() {
+        return getRadioButtonId(getFacesContext(), 0);
+    }
+
+    @Override
+    public String getValidatableInputClientId() {
+        return getRadioButtonId(getFacesContext(), 0);
+    }
+
+    @Override
+    public String getLabelledBy() {
+        return (String) getStateHelper().get("labelledby");
+    }
+
+    @Override
+    public void setLabelledBy(String labelledBy) {
+        getStateHelper().put("labelledby", labelledBy);
+    }
 }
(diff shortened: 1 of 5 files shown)


melloware:
Added Integration Test verifying the fix

primefaces/src/main/java/org/primefaces/component/selectoneradio/SelectOneRadio.java
```diff
@@ -47,4 +47,24 @@ public class SelectOneRadio extends SelectOneRadioBase {
     public String getRadioButtonId(FacesContext context, int index) {
         return this.getClientId(context) + UINamingContainer.getSeparatorChar(context) + index;
     }
+
+    @Override
+    public String getInputClientId() {
+        return getRadioButtonId(FacesContext.getCurrentInstance(), 0);
```
tandraschko: getFacesContext() from UIComponentBase
in MF its cached one vs ThreadLocal lookup

melloware: Good catch!  Fixed

PR #12960, opened 2024-11-27
Fix #12958 : ColorPicker popup not working in Sidebar

### LINKED ISSUE #12958: ColorPicker: is not working with p:sidebar
### Describe the bug

Hi, 

as mentioned in the Discord channel, p:sidebar is not working in p:colorPicker.
The Pop Up is not showing correctly.

```
<p:sidebar>
                <p:colorPicker value="#{myBean.color} />
</p:sidebar>
```

The <p:colorPicker is behind the <p:sidebar:

image

It should work like p:dialog

### Reproducer

_No response_

### Expected behavior

Pop Up from color picker is working and showing as expected

### PrimeFaces edition

Community

### PrimeFaces version

14.0.8

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.4

### Java version

11

### Browser(s)

_No response_

melloware:
Fix #12958 : ColorPicker popup not working in Sidebar

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/colorpicker/1-colorpicker.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/colorpicker/1-colorpicker.js
@@ -274,7 +274,7 @@ PrimeFaces.widget.ColorPicker = PrimeFaces.widget.BaseWidget.extend({
      * @private
      */
     setupDialogSupport: function() {
-        var dialog = this.input[0].closest('.ui-dialog');
+        var dialog = this.input[0].closest('.ui-dialog, .ui-sidebar');
         if (dialog) {
             this.cfg.parent = PrimeFaces.escapeClientId(dialog.id);
         }


tandraschko:
@melloware dont we have something like ui-dynamic-overlay? not sure

melloware:
hmmm i wonder if this does need to be refactored as I see this in about 5 components...

```js
/**
     * Performs some setup required to make this overlay menu work with dialogs.
     * @protected
     */
    setupDialogSupport: function() {
        var dialog = this.trigger.parents('.ui-dialog:first');

        if (dialog.length == 1 && dialog.css('position') === 'fixed') {
            this.getMenuElement().css('position', 'fixed');
        }
    },
```

tandraschko:
yeah
its actually overlay support
overlaypanel e.g. too

melloware:
I will try and mess with it since Colorpicker need a `parent` set when used inside any kind of overlay.  
https://github.com/melloware/Coloris

```js
// The default behavior is to append the color picker's dialog to the end of the document's
  // body. but it is possible to append it to a custom parent instead. This is especially useful
  // when the color fields are in a scrollable container and you want the color picker's dialog
  // to remain anchored to them. You will need to set the CSS position of the desired container
  // to "relative" or "absolute".
  // Note: This should be a scrollable container with enough space to display the picker.
  parent: '.container',

```

PR #12964, opened 2024-11-28
Fixes #12942 - Advanced FileUpload: highlight when drag/drop files

### LINKED ISSUE #12942: Advanced FileUpload: highlight when drag/drop files
When you drag over the advanced fileupload, there is no highlighting 
http://www.primefaces.org/showcase/ui/file/upload/dnd.xhtml?jfwid=f6694

only in our styled custom dropZone in the showcase

tandraschko:
Fixes #12942 - Advanced FileUpload: highlight when drag/drop files


primefaces-themes/theme-base/_components.scss
```diff
@@ -3,6 +3,7 @@
 @import './_mixins';
 @import './_common';
 @import './_colors';
+@import './_keyframes';
```
melloware, line 6: I am wondering if we should call keyframes as `animations`?  I am also ok leaving it keyframes.

tandraschko: doesnt matter for me
keyframes or animations is the same, its just called keyframes in css

melloware (approved):

PR #12986, opened 2024-12-03
Fix #12985: DatePicker prevent ESC from bubbling up when panel open

### LINKED ISSUE #12985: DatePicker: escape key handling in combination with Dialog
**1 Environment**

PrimeFaces version: primefaces 14.0.8
Application server: jetty 12.0.15 + version: mojarra40

**2 Expected behavior**

In primefaces 14 when you activate a datePicker field it opens the "visual date selector/dialog" because of WCAG.
You can leave the "visual date selector/dialog" by using the escape key or clicking on the associate field.

This works fine, but when adding a datePicker field to a Dialog window with closeOnEscape="true" then a problem arose.
Users get very confused because the escape key meant for the closing of "visual date selector/dialog" closes the complete dialog.
It is even more confusing when the datePicker is the first field in the Dialog, because this opens the "visual date selector/dialog" by default.

The user expects when a "visual date selector/dialog" is open that it closes with the escape key and the Dialog is stil open.

**3 Actual behavior**

When a "visual date selector/dialog" is open the escape key closes the "visual date selector/dialog" including the Dialog where the datePicker is in.

**4 Steps to reproduce**

See sample project. Click button "Show Dialog" then press escape key.

**5 Sample XHTML / Sample bean**

See attached project:
[primefaces-test-esc.zip](https://github.com/user-attachments/files/17992507/primefaces-test-esc.zip)

run with:
mvn clean jetty:run -Pmojarra40

melloware:
Fix #12985: DatePicker prevent ESC from bubbling up when panel open

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
@@ -2338,6 +2338,8 @@
 
         onPanelKeyDown: function(event) {
             if (event.key === 'Escape') {
+                event.preventDefault();
+                event.stopPropagation();
                 this.onEscapeKey(event);
             }
         },


tandraschko:
+1

PR #12990, opened 2024-12-03
Fix #12961: Remove legacy ThemeRoller ui-shadow

### LINKED ISSUE #12961: Remove unused JQuery ui-corner-* classes
found the following old themroller stuff:

unused:

- [x] ui-corner-* #12983 

only a few usages in our CSS but almost rendered for every component; we need to check:
- [ ] ui-widget
- [ ] ui-widget-content
- [ ] ui-shadow

melloware:
Fix #12961: Remove legacy ThemeRoller ui-shadow

@tandraschko this one was easier to manage and I removed all spacing.


tandraschko:
i dont think this is complete
ui-shadow is still used in some cases and we need some theme css

melloware:
Ahh yep I see it in some themes.

tandraschko:
also ui-shadow has an effect
we have to align those components in theme.css which used ui-shadow in the past

melloware:
IT used in `_menu.scss` so I reverted removing it from all the Menu components

tandraschko:
i will take care of it, dont have much to explain now

PR #12999, opened 2024-12-04
15.0.0-RC1

melloware:


tandraschko:
we should also adjust the version table in the readme.md

PR #13004, opened 2024-12-05
Fix #13003: Themes add sources.jar

### LINKED ISSUE #13003: Build: primefaces-themes does not create sources jar
We've built a custom primefaces theme based on the sass theme internally (just copy everything and modify as we need) and I would like to easily be able to update the base theme, e.g. when we want to use new features/components. I think I could build a somewhat elegant solution if there was a sources .jar published as well.

Right now the source plugin is just getting skipped because `primefaces-themes` is not using a standard maven layout (`src/main/resources`) nor defining a source resources directory:
```
[INFO] --- source:3.3.1:jar-no-fork (attach-sources) @ primefaces-themes ---
[INFO] No sources in project. Archive not created
```

melloware:
Fix #13003: Themes add sources.jar


tandraschko:
can you post a screenshot from the archive content?

melloware:
Sure:
image

I kept the directory structure in the JAR exactly as it is on disk

PR #13006, opened 2024-12-05
Fix #13005: DataView NPE in encodeHeader

### LINKED ISSUE #13005: DataView: NPE in encodeHeader
### Describe the bug

```
Dec 05, 2024 9:26:39 AM org.primefaces.application.exceptionhandler.PrimeExceptionHandler logException
SEVERE: Cannot invoke "jakarta.faces.component.UIComponent.encodeAll(jakarta.faces.context.FacesContext)" because "fHeader" is null
java.lang.NullPointerException: Cannot invoke "jakarta.faces.component.UIComponent.encodeAll(jakarta.faces.context.FacesContext)" because "fHeader" is null
        at org.primefaces.component.dataview.DataViewRenderer.encodeHeader(DataViewRenderer.java:148)
        at org.primefaces.component.dataview.DataViewRenderer.encodeMarkup(DataViewRenderer.java:123)
        at org.primefaces.component.dataview.DataViewRenderer.encodeEnd(DataViewRenderer.java:88)
        at jakarta.faces.component.UIComponentBase.encodeEnd(UIComponentBase.java:586)
```

### Reproducer

Showcase in 15.0

### Expected behavior

no NPE

### PrimeFaces edition

Community

### PrimeFaces version

15.0.0-RC1

### Theme

ALL

### JSF implementation

Mojarra

### JSF version

4.0.9

### Java version

11

### Browser(s)

ALL

melloware:
Fix #13005: DataView NPE in encodeHeader

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/dataview/DataViewRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/dataview/DataViewRenderer.java
@@ -140,12 +140,15 @@ public class DataViewRenderer extends DataRenderer {
     protected void encodeHeader(FacesContext context, DataView dataview) throws IOException {
         ResponseWriter writer = context.getResponseWriter();
         UIComponent fHeader = dataview.getFacet("header");
+        boolean isRenderFacet = FacetUtils.shouldRenderFacet(fHeader);
 
-        if (FacetUtils.shouldRenderFacet(fHeader) || hasLayoutOptions(context, dataview)) {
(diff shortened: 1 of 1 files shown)


tandraschko:
does layout-options make sense without header?
otherwise we can add a null check our encodeAll

melloware:
you are right it does make sense just updated the check.  It doesn't want to render the DIV at all if no facet or Layout, but if either exists render so we need the check again around the facet.encodeAll

PR #13039, opened 2024-12-10
ValueExpressionStateHelper: implement Faces 4.0 eval

### LINKED ISSUE #13038: ValueExpressionStateHelper does not implement Faces 4.0 eval
`StateHelper#eval(Serializable, Supplier)` was introduced in Faces 4.0: https://jakarta.ee/specifications/faces/4.0/apidocs/jakarta/faces/component/statehelper#eval(java.io.Serializable,java.util.function.Supplier)

This exception is thrown when this method is invoked
`java.lang.AbstractMethodError: Receiver class org.primefaces.el.ValueExpressionStateHelper does not define or inherit an implementation of the resolved method 'abstract java.lang.Object eval(java.io.Serializable, java.util.function.Supplier)' of interface jakarta.faces.component.StateHelper.`

PrimeFaces cannot @Override this right now for obvious reasons, but it should be possible to just implement this IMO

jungm:
Fixes #13038

### DIFF
--- a/primefaces/src/main/java/org/primefaces/el/ValueExpressionStateHelper.java
+++ b/primefaces/src/main/java/org/primefaces/el/ValueExpressionStateHelper.java
@@ -111,6 +111,7 @@ public class ValueExpressionStateHelper implements StateHelper {
         return ve.getValue(FacesContext.getCurrentInstance().getELContext());
     }
 
+    // @Override (we can't actually override right now because this is a new method in Faces 4.0)
     public Object eval(Serializable key, Supplier<Object> defaultValueSupplier) {
         Object result = eval(key, UNKNOWN_VALUE);
         if (result == UNKNOWN_VALUE) {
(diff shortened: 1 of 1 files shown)


primefaces/src/main/java/org/primefaces/el/ValueExpressionStateHelper.java
```diff
@@ -108,6 +111,19 @@ public Object eval(Serializable key, Object defaultValue) {
         return ve.getValue(FacesContext.getCurrentInstance().getELContext());
     }
 
+    public Object eval(Serializable key, Supplier<Object> defaultValueSupplier) {
```
tandraschko, line 115: please add ``// @Override`

jungm: done

PR #13064, opened 2024-12-12
Fixes #13063 - added sample

### LINKED ISSUE #13063: p:media not working when using StreamedContent
### Describe the bug

I am trying to use Primefaces p:media component in Version 13.0.5 to display pdf´s loaded from database.

This is my BackingBean-getter and it´s definetly being invoked and building the StreamedContent with the provided byte[] from the database:

public StreamedContent getMedia() {
    StreamedContent result;
    if (this.selectedRow == null) {
        return null;
    }
    result = DefaultStreamedContent.builder()
            .contentType("application/pdf")
            .stream(() -> {
                try {
                    return new ByteArrayInputStream(this.getSelectedDokument().getDocumentAsByte());
                } catch (Exception e) {
                    e.printStackTrace();
                    return null;
                }
            })
            .build();
    return result;
}

XHML:

         <p:media value="#{controller.getMedia()}" width="100%" height="590px" id="media" cache="false"
           player="pdf"/>

Bean is SessionScoped.

When using a static path to the pdf-File it´s working fine.
Any ideas?

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

Community

### PrimeFaces version

13.0.5

### Theme

Avalon

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

17

### Browser(s)

All

tandraschko:
Fixes #13063 - added sample


Koeppes:
Hi Thomas,

I found out, why it´s not working in my case.
I am building a reusable Diatog, that shows the PDF Viewer.
The files will be provided by different ManageBeans, so the root for the p:media tag has to be variable.

See the following code:

"printerBean" is the variable ManagedBean, that provides the StreamedContent.

                xmlns:composite="http://java.sun.com/jsf/composite"
                xmlns:c="http://java.sun.com/jsp/jstl/core"
    <f:metadata>
    </f:metadata>
    <composite:interface>
        <composite:attribute name="printerBean" required="true"
        />
        <composite:attribute name="dokument" required="true"
        />
        <composite:attribute name="widgetVar" required="true"
        />
    </composite:interface>
    <composite:implementation>
        <c:set var="printer" value="#{printerBean}"/>
        <p:dialog header="#{msg['label.app.dialog.titel.drucken']}"
                  id="printDlg" widgetVar="#{cc.attrs.widgetVar}" modal="true"
                  closable="false" dynamic="true" closeOnEscape="false" width="75%"
                  height="75%" position="center">
            <p:outputLabel value="#{printerBean.getDocumentName()}"></p:outputLabel>
            <div class="p-field p-col-12 p-md-12">
                <p:media value="#{printerBean.getMedia()}" width="100%" height="590px" id="media" cache="false"
                         player="pdf"/>

            </div>
            <f:facet name="footer">
                <h:panelGroup layout="block" styleClass="w100p cb-div-element">
                    <p:commandButton id="CloseBtn" immediate="true"
                                     value="#{msg['app.dialog.btnAbbrechen.label']}"
                                     class="ui-button-secondary"
                                     action="#{controller.cancelPrintDialog()}">
                    </p:commandButton>
                </h:panelGroup>
            </f:facet>
        </p:dialog>
    </composite:implementation>
</ui:composition>

The outputLabel is working fine (Name of the File is shown above the MediaPlayer as expected).
But Media component stays Empty.

When replacing the 

                <p:media value="#{printerBean.getMedia()}"

with

                <p:media value="#{actualNameOfTheBean.getMedia()}"

it´s working fine.

Tested this behaviour with you getGraphicText()-Method and it´s also not working.

Regards
Jürgen

tandraschko:
The problem isnt the composite, method expressions are not supported in all cases

Koeppes:
I am using method expressions in both cases, am I not? It´s only a scope issue in relation of the getMedia()-method, as I see it.

tandraschko:
yeah
try .media

Koeppes:
this has no effect. it´s even getting worse. with getMedia() the method is invoked, but pdf-Viewer is being displayed in the dialog, bit not showing anything.

with ".media" the method won´t be invoked and the media-player is not being displayed in the dialog.

tandraschko:
if you provide a primefaces-test project, i can have a look

Koeppes:
Hi,

I pushed a demo project: https://github.com/Koeppes/tag

After start please click on „Calendar“.
You´ll see two buttons.

One with dynamic dialog, which reproduces my problem.
To onyl difference to my real project: it´s throwing an exception, I don´t get.

Regards
Juergen

melloware:
@Koeppes your link gives me a 404.  Are you sure its not private?

PR #13077, opened 2024-12-16
Fixes #13073

### LINKED ISSUE #13073: AjaxExceptionHandler: switch to client-side handling, so we can even support default Faces ExceptionHandler
Currently our PrimeFacesExceptionHandler looks for AjaxExceptionHandler components in the view, resolve updates, directly render the component in the response and so on.

We can simply reimplement this in client-side `#doError` method

tandraschko:
Fixes #13073


melloware (approved):
This is fantastic!

PR #13107, opened 2024-12-23
Fix #12663: ConfirmDialog/Popup restore original values

### LINKED ISSUE #12663: ConfirmPopup / ConfirmDialog: Default icon not displayed if not provied by p:confirm
### Describe the bug

`p:confirmDialog` and `p:confirmPopup` do provide an attribut named `icon`. But the default icon is not shown, if no icon is provided by `p:confirm`.

- `p:confirmDialog`: Icon is set to `null`, if `p:confirm` provides no icon (misses correct check for no icon). And if `p:confirm` provides an icon, the default one is removed permanently, see [here](https://github.com/primefaces/primefaces/blob/60fa909/primefaces/src/main/resources/META-INF/resources/primefaces/dialog/dialog.js#L1052).

- `p:confirmPopup`: Default Icon is always permanently removed and the one from `p:confirm` is set, if provded, see [here](https://github.com/primefaces/primefaces/blob/380a939/primefaces/src/main/resources/META-INF/resources/primefaces/confirmpopup/confirmpopup.js#L343).

The same somehow applies to `header` and `message` attributes. Once overwritten by `p:confirm` specific value, the default is lost.

Maybe the handling of `p:confirm` specific value vs default values needs discussion, but default icon is clearly not working as designed, IMHO.

### Reproducer

_No response_

### Expected behavior

Default icon should be visible, if no icon is provied by `p:confirm`

### PrimeFaces edition

Community

### PrimeFaces version

10, 15.0.0-SNAPSHOT

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.2

### Java version

8

### Browser(s)

_No response_

melloware:
Fix #12663: ConfirmDialog/Popup restore original values

Similar to what was already being done for Yes and No buttons and icons must store original value in `.data` and restore it upon hide.

Added integration tests

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/confirmpopup/confirmpopup.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/confirmpopup/confirmpopup.js
@@ -57,6 +57,9 @@ PrimeFaces.widget.ConfirmPopup = PrimeFaces.widget.DynamicOverlayWidget.extend({
         this.message = this.content.children('.ui-confirm-popup-message');
         this.icon = this.content.children('.ui-confirm-popup-icon');
         if (this.cfg.global) {
+            this.content.data('p-text', this.content.text());
+            this.message.data('p-text', this.message.text());
+            this.icon.data('p-icon', this.icon.removeClass('ui-confirm-popup-icon').attr('class'));
             this.yesButton = this.jq.find('.ui-confirm-popup-yes');
             this.noButton = this.jq.find('.ui-confirm-popup-no');
             this.yesButton.data('p-text', this.yesButton.children('.ui-button-text').text());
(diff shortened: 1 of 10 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/confirmpopup/confirmpopup.js
```diff
@@ -251,6 +254,15 @@ PrimeFaces.widget.ConfirmPopup = PrimeFaces.widget.DynamicOverlayWidget.extend({
     restoreButtons: function() {
         var $this = this;
         if ($this.cfg.global) {
+            if ($this.content.data('p-text')) {
```
tandraschko, line 257: as alternative we could store all the attributes in the root element
but not sure whats better or not

melloware:
I will just merge this for now we can always refactor if we feel its better in the future.

PR #13108, opened 2024-12-23
Fix #11974: Autocomplete multiple mode support float label

### LINKED ISSUE #11974: AutoComplete: Float label and multiple="true" not working
### Discussed in https://github.com/orgs/primefaces/discussions/1987

<sup>Originally posted by **Cprusinowski** May 21, 2024</sup>
Hi,

When I try to use code below then float label doesn't work.
Cause of this is multiple="true".
Does anyone have a solution?

```
<span class="ui-float-label">
    <p:autoComplete multiple="true" />
    <p:outputLabel for="@previous" value="AutoComplete" />
</span>
```

### Reproducer

Here is a runnable reproducer showing the issue: 
[discussion-1987.zip](https://github.com/primefaces/community/files/15389742/discussion-1987.zip)

melloware:
Fix #11974: Autocomplete multiple mode support float label


tandraschko:
+1 if well tested

PR #13116, opened 2024-12-28
#13115 Sonar: add test-coverage via JaCoCo

### LINKED ISSUE #13115: SonarQube: add test-coverage via JaCoCo
### Description

We currently introduce SonarQube in our organization where i do my day-work. As part of this we configure JaCoCo - in our scenario via Gradle - to get metrics regarding test-coverage.

Hope this is also kind of a quick-win on the Maven-side of things.
https://docs.sonarsource.com/sonarqube-cloud/enriching/test-coverage/java-test-coverage/#add-coverage-in-a-single-module-maven-project

### Additional context

_No response_

christophs78:


christophs78:
Hm. Maybe it´s a bit more complicated because we have IT´s in a separate module. OOTB we only get coverage for unit- and integration-test located within PrimeFaces(-core)-module.

christophs78:
I´ll have to look into https://www.baeldung.com/maven-jacoco-multi-module-project.

melloware:
Yep I think you will need to use the multmodule so it has the coverage. It is still not perfect but its "better".

GedMarc:
If you want to aggregate results, best with a separate module all together 👍 
But it really is better to scan individually, with a project like this with mass extentions available,
Then use the reporting module to test integrations between the modules, 
In my experience for projects like these 👍

christophs78:
Note for my own usage:
https://docs.sonarsource.com/sonarqube-server/latest/analyzing-source-code/scanners/sonarscanner-for-maven/

.github/workflows/build.yml
```diff
@@ -26,9 +26,12 @@ jobs:
           cache: 'maven'
       - name: Build
         run: mvn clean install -T1C -Pjsdoc,minify --batch-mode --show-version
+      - name: Sonar Cloud - simulate (without sonar:sonar)
+        if: matrix.java == 21
+        run: mvn clean install -Pquick -pl !primefaces-integration-tests-jakarta
       - name: Sonar Cloud
         if: github.repository == 'primefaces/primefaces' && github.ref == 'refs/heads/master' && matrix.java == 21
-        run: mvn clean org.jacoco:jacoco-maven-plugin:prepare-agent install sonar:sonar -Pquick -Dsonar.organization=primefaces -Dsonar.host.url=https://sonarcloud.io -Dsonar.token=${{secrets.SONAR_TOKEN}}  -fprimefaces/pom.xml
+        run: mvn clean install sonar:sonar -Pquick -pl !primefaces-integration-tests-jakarta -Dsonar.organization=primefaces -Dsonar.host.url=https://sonarcloud.io -Dsonar.token=${{secrets.SONAR_TOKEN}}
```
christophs78: change #1 - we run `mvn clean install sonar:sonar ...` for the whole project instead of only for Primefaces(-Core)
This is necessary to get code-coverage generated via "primefaces-integration-tests" included.

.github/workflows/build.yml
```diff
@@ -26,9 +26,12 @@ jobs:
           cache: 'maven'
       - name: Build
         run: mvn clean install -T1C -Pjsdoc,minify --batch-mode --show-version
+      - name: Sonar Cloud - simulate (without sonar:sonar)
```
christophs78: I´ll remove this step once validated.

primefaces-cli/pom.xml
```diff
@@ -16,6 +16,10 @@
 
     <name>PrimeFaces CLI</name>
 
+    <properties>
+        <sonar.skip>true</sonar.skip>
```
christophs78, line 20: For all modules except primefaces(-core) and primefaces-coverage we set sonar.skip=true because we probably do not want them to get included into Sonar.

primefaces-coverage/pom.xml
```diff
@@ -0,0 +1,70 @@
...
+            </plugin>
+            <plugin>
+                <groupId>org.jacoco</groupId>
+                <artifactId>jacoco-maven-plugin</artifactId>
+                <version>${jacoco.version}</version>
+                <executions>
+                    <execution>
+                        <phase>verify</phase>
+                        <goals>
+                            <goal>report-aggregate</goal>
```
christophs78, line 54: We aggregate JaCoCo-coverage provided by primefaces-(core) (mostly unit-tests) and primfaces-integration-tests (e2e-integration-tests).

christophs78:
I´ll give it a shot. Worst case is i have to add some more commits or rollback this squashed PR.

christophs78:
Had to rollback because
> Failed to execute goal org.sonarsource.scanner.maven:sonar-maven-plugin:4.0.0.4121:sonar (default-cli) on project primefaces-parent: Could not find a default branch for project with key 'org.primefaces:primefaces-parent'. Make sure project exists. -> [Help 1]

Setting `sonar.skip=true` in the parent-pom.xml does not work either because this results in skipping the whole project. Setting `sonar.skip=false` for modules does not compensate for this

PR #13121, opened 2024-12-30
Fix #13120: primefaces.password.INVALID_MATCH_detail quote fields

### LINKED ISSUE #13120: Password: better INVALID_MATCH message
Currently its:

```
primefaces.password.INVALID_MATCH_detail = {0} should match with {1}.
```

if one input has the label "Password", the other "repeat Password", the msg will be:

```
Password should match with repeat Password.
```

its just weird, we just should add quotation marks around

melloware:
Fix #13120: primefaces.password.INVALID_MATCH_detail quote fields


tandraschko:
+1

PR #13132, opened 2024-12-31
Fix #13131: Picklist disable drag/drop on touch devices

### LINKED ISSUE #13131: PickList: add flag to disable dragdrop support
### Description

Would be nice if PrimeFaces [`<p:pickList>`](https://www.primefaces.org/showcase/ui/data/pickList.xhtml) allows disabling dragdrop support (among others to improve user experience on touchscreen devices).

Note: this is already implemented in PrimeNG [`<p-picklist>`](https://primeng.org/picklist).

Add a new boolean attribute like `dragdrop` in PrimeNG one.

### Additional context

_No response_

melloware:
Fix #13131: Picklist disable drag/drop on touch devices

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/picklist/picklist.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/picklist/picklist.js
@@ -132,60 +132,7 @@ PrimeFaces.widget.PickList = PrimeFaces.widget.BaseWidget.extend({
             $(this.jqId + ' .ui-picklist-filter-container').addClass('ui-state-disabled').children('input').attr('disabled', 'disabled');
         }
         else {
-            var $this = this,
-                reordered = true;
-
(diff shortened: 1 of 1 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/picklist/picklist.js
```diff
@@ -351,6 +362,8 @@ PrimeFaces.widget.PickList = PrimeFaces.widget.BaseWidget.extend({
                 $this.focusedItem = list.children('.ui-picklist-item:visible:first');
             }
 
+            list.children('.ui-picklist-outline').removeClass('ui-picklist-outline');
```
melloware, line 365: This is because I noticed a bug when selecting a new item by MOUSE it was leaving the old selection still outlined.

PR #13139, opened 2025-01-05
#11364 CSV - complex/multi-value: add showcase-excample

### LINKED ISSUE #11364: CSV: add integrationtest (and maybe showcase-example) for Multi-Field / Complex validation
### Description

Integration-test(s) and idealy showcase-exmple for
https://primefaces.github.io/primefaces/13_0_0/#/core/csv?id=multi-field-complex-validation

christophs78:
Fix #11364

There was already a - not yet linked - showcase-example provided in 2021 by @tandraschko.
But it is broken. Propably not linked because of this. 

After we figured out how it work´s and/or fixed the implementation we will have to fix the documentation too. (https://primefaces.github.io/primefaces/15_0_0/#/core/csv?id=multi-field-complex-validation)

### DIFF
--- a/primefaces-showcase/src/main/webapp/ui/csv/complex.xhtml
+++ b/primefaces-showcase/src/main/webapp/ui/csv/complex.xhtml
@@ -33,7 +33,7 @@
 
 
     <ui:define name="title">
-        Client Side Validation <span class="subitem">Complex</span>
+        Client Side Validation <span class="subitem">Complex / Multi-Field</span>
     </ui:define>
 
     <ui:define name="description">
@@ -46,7 +46,7 @@
     <ui:define name="implementation">
         <div class="card">
             <h:form id="form">
-                <p:panel header="Validate" pt:data-p-val="MyComplexValidator">
+                <p:outputPanel pt:data-p-val="MyComplexValidator">
                     <p:messages showDetail="true">
                         <p:autoUpdate/>
                     </p:messages>
@@ -72,7 +72,7 @@
                     <p:commandButton value="Ajax" id="btnAjax" update="grid" icon="pi pi-check" validateClient="true"
                                      style="margin-right:10px"/>
                     <p:commandButton value="Disabled" ajax="false" icon="pi pi-check" style="margin-right:10px"/>
-                </p:panel>
+                </p:outputPanel>
             </h:form>
         </div>
     </ui:define>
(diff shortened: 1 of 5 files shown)


christophs78:
It´s working. But we may discuss the way.
And afterwards fix the documentation.

melloware:
@tandraschko please review.

primefaces-showcase/src/main/webapp/ui/csv/complex.xhtml
```diff
@@ -12,18 +12,22 @@
...
+                    let nameRequiredPf = PF('nameRequired');
+                    let namePf = PF('name');
 
-                        throw {
-                            summary: 'Validation Error',
-                            detail: 'A name is required!'
-                        };
+                    if (nameRequiredPf.isChecked() && namePf.jq.val().length === 0) {
+                        PrimeFaces.validator.Highlighter.types['default'].highlight(namePf.jq);
+                        PrimeFaces.validation.ValidationContext.addMessage(namePf.jq, { summary: 'Validation Error', detail: 'A name is required!' });
```
tandraschko, line 28: it it "better" / more intuitiv for users to call addMessage manually or was it broken?

tandraschko: built-in validators uses:
`throw vc.getMessage(this.MINIMUM_MESSAGE_ID, min, vc.getLabel(element));`

which actually is:

```
            throw {
                summary: summary,
                detail: detail
            };
```

tandraschko:
also i would like to add:

```
        highlight: function(widget) {
            var element = widget.jq;
            var highlighterType = element.data('p-hl') || 'default';
            var highlighter = PrimeFaces.validator.Highlighter.types[highlighterType];

            highlighter.highlight(element);
            element.attr('aria-invalid', true);
        },

        unhighlight: function(widget) {
            var element = widget.jq;
            var highlighterType = element.data('p-hl') || 'default';
            var highlighter = PrimeFaces.validator.Highlighter.types[highlighterType];

            highlighter.unhighlight(element);
            element.attr('aria-invalid', false);
        },
```

to the ValidationContext

so you can just do 

`PrimeFaces.validation.ValidationContext.highlight(namePf);`

primefaces-showcase/src/main/webapp/ui/csv/complex.xhtml
christophs78:
```javascript
                        throw {
                            summary: 'Validation Error',
                            detail: 'A name is required!'
                        };
```

Is (for this usecase) IMO broken. At least i could not get it working.

tandraschko:
wanna work further on it or should i? i can also check the throw-case then

christophs78:
Would appreciate when you can jump in and work on this. Promised from my side to work on the showcase-example without anticipating there are probable issues deeper inside.

PR #13153, opened 2025-01-08
Fix #13152: SelectOneMenu scrollbar issues with long option label

### LINKED ISSUE #13152: SelectOneMenu: Long option label cause scrollbar issues on Firefox
### Describe the bug

Mostly a cosmetic issue.

If the overlay panel has a vertical scrollbar and some option label being longer than the input, then the scrollbar obscures item space and in the worst case creates (e.g. with default padding) an additional horizontal scrollbar.

pf-select-scrollbar-issue

Only Firefox related, since their scrollbar behaviour seems to differ from others, i.e. the scrollbar width is included in the overflowing container's width and not added on top like in chrome. Issue report: https://bugzilla.mozilla.org/show_bug.cgi?id=764076.

This issue is similar to #5490, which was fixed in b6768c69 by setting a scrollbar via CSS during overlay width calculation, so that the scrollbar width is considered in the overlay width. However, it seems this fix has been lost in the meantime.

After a short investigation, it seems that the previous fix in combination with an additional overflow check during `alignPanelWidth` in `forms.selectonemenu.js` will work fine and results in a similar behaviour like Chrome. Tested it with multiple SelectOneMenu's setups.

What do you think? I can submit a PR.

### Reproducer

Reproducer enclosed with multiple SelectOneMenus having different width and number of options.
It includes also the proposed fix as commented-out monkey patch.
[primefaces-test-select-scrollbar.zip](https://github.com/user-attachments/files/18351890/primefaces-test-select-scrollbar.zip)

### Expected behavior

The scrollbar should be added on top of the width of the overlay panel to not break spacing/layout.

### PrimeFaces edition

Community

### PrimeFaces version

14.0.9

### Theme

Saga

### JSF implementation

All

### JSF version

4.1.

### Java version

17

### Browser(s)

Firefox

guewin:
Fix #13152: SelectOneMenu scrollbar issues with long option label

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectonemenu.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/forms/forms.selectonemenu.js
@@ -280,12 +280,21 @@ PrimeFaces.widget.SelectOneMenu = PrimeFaces.widget.DeferredWidget.extend({
     alignPanelWidth: function() {
         if(!this.panelWidthAdjusted) {
             var jqWidth = this.jq.outerWidth();
+
+            /* Ensures that the width of a potential scrollbar is considered in the following panel width calculation */
+            let hasVerticalScrollbar = this.itemsWrapper[0].scrollHeight > this.itemsWrapper[0].clientHeight;
+            if (hasVerticalScrollbar) {
+                this.itemsWrapper.css({'overflow': 'scroll'});
+            }
+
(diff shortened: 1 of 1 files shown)


melloware (approved):
LGTM

guewin:
Rebased with latest changes from master.

guewin:
@melloware I dont have write access, do you mind merging?

PR #13176, opened 2025-01-12
Fix #13169: TabView handle dynamic datamodel tabs

### LINKED ISSUE #13169: Tabview: Dynamic visitTree behaviour
### Description

I'm analysing a JSF lifecycle. I'm using a dynamic TabView. Each Tab has a custom component.
I load all tabs (click on them) than click on a button on the last tab. I'm debugging visitTree. I don't understand why for the phase RESTORE_VIEW visitTree is called just for the custom component in the first tab. Analysing a little bit I see that the VisitHint.SKIP_ITERATION is set, casuing UiTabPanel.visitTree to return after calling super.visitTree. Can you help to unterstand the behaviour?

I guess that, when the RESTORE_VIEW phase executes, VisitHint.SKIP_ITERATION is set because only the currently active tab's components need to be restored. But, from what I see, the visit is than delegated to UIComponent,  and there the first Tab is visited.

I'm not sure if it is a Bug or I just don't unterstand the flow. I'm using primefaces 13.0.3, but I don't see any relevant changes since then.

image

Visit all Tabs?

### Additional context

_No response_

melloware:
Fix #13169: TabView handle dynamic datamodel tabs

This draft adds the same Dynamic Tab loaded check to the DataModel scenario as it does in the non repeating scenario.

Not sure if this is correct but it seems like it should be there

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/api/UITabPanel.java
+++ b/primefaces/src/main/java/org/primefaces/component/api/UITabPanel.java
@@ -973,8 +973,8 @@ public class UITabPanel extends UIPanel implements NamingContainer {
         else {
             // override the behavior from UIComponent to visit
             // all children once per "row"
-
-            if (ComponentUtils.isSkipIteration(context, context.getFacesContext())) {
+            // #13169 do not skip if tabs are dynamic each one needs to be visited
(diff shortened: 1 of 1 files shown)


tandraschko:
Im in holiday 

But we need tests to make sure, loaded tabs are Not visited again

PR #13179, opened 2025-01-12
#13178 PF Selenium: Selenium deprecated WebStorage - abstraction

### LINKED ISSUE #13178: PF Selenium: Selenium deprecated WebStorage - abstraction
https://github.com/SeleniumHQ/selenium/issues/10397

-->Remove it form org.primefaces.selenium.AbstractPrimePage and org.primefaces.selenium.PrimeSelenium
--> check whether it breaks something
--> if it breaks something maybe do some JS and as an option provide our own abstraction

christophs78:


melloware (approved):

christophs78:
For me it looks like we can remove it. Selenium itself does not provide a good explaination why they removed it. Maybe due to issues they did not want to tackle. The alternative from my feeling is to write JS-code to access e.g. WebStorage.

PR #13194, opened 2025-01-17
Fix #13191: ComponentUtils allow flex="false"

### LINKED ISSUE #13191: SelectOneRadio: Context param overrides component attribute
### Describe the bug

In my app, I have this `context-param` set globally:
```xml
<context-param>
	<param-name>primefaces.FLEX</param-name>
</context-param>
```
but that does not allow me to override `flex="false"` for an individual component. I believe this is because of this line in [`ComponentUtils`](https://github.com/primefaces/primefaces/blob/14.0.10/primefaces/src/main/java/org/primefaces/util/ComponentUtils.java#L288):
```java
return component.isFlex() || PrimeRequestContext.getCurrentInstance(context).isFlex();
```
I think a reasonable solution would be to make `FlexAware.isFlex` return a boxed `Boolean` rather than primitive `boolean`, so that `null` is a valid return value, and make `ComponentUtils.isFlex` only use `PrimeRequestContext` if `null` was returned.

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

14.0.9

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3.14

### Java version

OpenJDK 11.0.23+9

### Browser(s)

_No response_

jbrower2:
Changes the signature of `FlexAware.isFlex()` to return null instead of defaulting to `false`.

Fix #13191: ComponentUtils allow flex="false"

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/api/FlexAware.java
+++ b/primefaces/src/main/java/org/primefaces/component/api/FlexAware.java
@@ -33,5 +33,5 @@ public interface FlexAware {
      *
      * @return true to use PrimeFlex instead of Grid CSS.
      */
-    boolean isFlex();
+    Boolean getFlex();
 }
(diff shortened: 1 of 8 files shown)


jbrower2:
@melloware should I make the corresponding `setFlex` methods take in `Boolean` as well?

tandraschko:
yes
get/set must match

melloware (approved):

PR #13199, opened 2025-01-17
Fix #13189: PrimeFaces.debounce utility function

### LINKED ISSUE #13189: Core: Add `PrimeFaces.debounce` method for filter delays
### Description

Based on Discord discussion: https://discord.com/channels/787967399105134612/787967662293909524/1329804397496107090

<@788103814817120305> I have a question and would love to hear what you think 🤓. We have the following scenario:
TreeTable with a globalFilterFunction:
```xml
<p:treeTable id="..."
             styleClass="..."
             value="#{...}"
             widgetVar="..."
             var="item"
             nodeVar="node"
             selectionMode="single"
             skipChildren="true"
             filteredValue="#{...}"
             globalFilterFunction="#{...}"
             selection="#{...}">
```
A custom InputText to call the global filter function:
```xml
<f:facet name="header">
    <div class="text-right">
        <p:inputText id="globalFilter"
                     value="#{...}"
                     onkeyup="PF('...').filter();"
                     placeholder="#{...}"/>
    </div>
</f:facet>
```
What happens now, obviously, is that the filter function triggers on every key entered, which is not optimal for performance, as all such requests are, of course, queued. So what we would need is some kind of delay like RemoteCommand has.

Do you have any suggestions for the optimal implementation? Do we have to write our own JS function using timeouts? Or is there another solution we are not thinking of?
Thank you very much in advance!

Add `PrimeFaces.debounce(() -> PF('...').filter(), 400)` with a default value of 400 that is OPTIONAL so if left out will be 400.

### Additional context

_No response_

melloware:
Fix #13189: PrimeFaces.debounce utility function

### DIFF
--- a/docs/15_0_0/components/datatable.md
+++ b/docs/15_0_0/components/datatable.md
@@ -374,7 +374,7 @@ Following is a basic filtering datatable with these options demonstrated:
 
     <f:facet name="header">
         <h:outputText value="Search all fields:" />
-        <h:inputText id="globalFilter" onkeyup="PF('carsTable').filter()" />
+        <h:inputText id="globalFilter" onkeyup="PrimeFaces.debounce(() => PF('carsTable').filter())" />
     </f:facet>
 
     <p:column field="model" headerText="Model" filterMatchMode="contains" />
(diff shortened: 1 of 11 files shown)


tandraschko:
+1 if well tested, code looks very good
should we add it to datatable docs and filter showcase?

melloware:
OK docs and showcase updated!

PR #13200, opened 2025-01-19
Dependency Manage Jackson

melloware:
It looks like WebDriverManager uses `com.github.docker-java` which I am pretty sure we don't even use and it brings in vulnerable jackson.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -438,6 +438,15 @@
                 </exclusions>
             </dependency>
 
+            <!-- Jackson for Integration Tests / Selenium -->
+            <dependency>
+                <groupId>com.fasterxml.jackson</groupId>
+                <artifactId>jackson-bom</artifactId>
+                <version>2.18.2</version>
+                <scope>import</scope>
+                <type>pom</type>
+            </dependency>   
+
             <!-- JUnit -->
             <dependency>
                 <groupId>org.junit.jupiter</groupId>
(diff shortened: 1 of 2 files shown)


tandraschko:
Im not sure If its a good Idea
If anyone importing our pf selenium, Jackson will BE Missing and exception be thrown?

melloware:
ok let me try it with dependency managament then in our parent pom.xml

tandraschko:
I dont think we should do anything here 
We need to ping the other libs to update

melloware:
i think they have updated just no release in 8 months.

melloware:
I updated to use the Jackson BOM so across all our projects we are using the same Jackson version.

tandraschko:
i dont think it fixes the problem
the user got still the old jackson version when importing pf selenium?

melloware:
Yes it just controls OUR PF build but does not affect the user.

tandraschko:
maybe we should just get rid of jackson it in other PF modules, it always brings such problems

currently we use it:
- directly for resteasy in showcase; runtime dependency
- directly for resteasy in IT; runtime dependency
- transistive dependency for webdriver stuff; how is it used there anyway?

we could use jsonb instead of jackson for resteasy also

melloware:
The transitive WebDriver stuff is because it uses `com.github.docker-java` to do some communication with Docker i guess in some circumstances where people run Selenium inside Docker containers or maybe for BrowserStack or something like that.

tandraschko:
can you check this?
maybe we can rly just exclude jackson, if its just not our intended use

melloware:
yeah it was passing with my original commit excluding Docker and Jackson.

PR #13201, opened 2025-01-19
Fix #13182: DatePicker time input separate keydown/keyup

### LINKED ISSUE #13182: DatePicker: selected text of time input is not overwritten
### Describe the bug

- Have a `<p:datePicker>` with `timeInput="true"`.
- Open the date picker.
- Select one or two characters of either hour or minute field of time input.
- Enter a digit.

Expectation: the entered digit comes in place of the selection.
Actual: nothing happens.

I've already monkeypatched it, you may find it useful to improve next PrimeFaces version herewith :)

```js
/**
 * https://github.com/primefaces/primefaces/blob/14.0.9/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
 */
function patchDatePicker(datePicker) {
    const originalOnTimeInputKeyDown = datePicker.onTimeInputKeyDown;
    datePicker.onTimeInputKeyDown = function(event) {
        switch (event.key) {
            case 'ArrowLeft':
            case 'ArrowRight':
            case 'Home':
            case 'End':
            case 'Shift':
                return true; // Allow these keys as well.
        }

        const selection = document.getSelection();

        if (selection && selection.rangeCount == 1) {
            const selectionEnd = selection.toString().length;

            if (selectionEnd > 0) {
                let selectionStart = 0;

                if (selectionEnd < 2) { 
                    // input type="number" doesn't support element.selectionStart/End, hence this trick, see also https://stackoverflow.com/a/32855335
                    selection.modify('extend', 'forward', 'character');
                    let forwards = true;

                    if (selection.toString().length == 0) {
                        // it was selected backwards (right to left) instead of forwards (left to right)
                        forwards = false;
                        selection.modify('extend', 'backward', 'character');
                        selection.modify('extend', 'backward', 'character');
                    }

                    if (selection.toString().length == (forwards ? 1 : 2)) {
                        selectionStart = 1;
                    }
                }

                const input = event.currentTarget;
                input.value = input.value.slice(0, selectionStart) + input.value.slice(selectionStart + selectionEnd);

                if (0 <= event.key && event.key <= 9) {
                    if (selectionStart == 0) {
                        input.value = event.key + input.value; // because default impl does input.value + event.key
                        event.preventDefault();
                        event.stopPropagation();
                        return;
                    }
                }
            }
        }

        originalOnTimeInputKeyDown.call(datePicker, event);
    }
}
```

Note that I also adjusted it to allow other keys within the time inputs as well, it are exactly those keys which allows you to be able to select text by keyboard only.

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

14.0.9

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3.21

### Java version

17

### Browser(s)

_No response_

melloware:
Fix #13182: DatePicker time input separate keydown/keyup

@BalusC this separates the keydown and keyup and is much better user experience and selection just works as does Cut Copy Paste.

keydown - only prevents invalid keys from moving on
keyup - performs the numeric validations

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/datepicker/0-datepicker.js
@@ -1920,9 +1920,10 @@
                 yearNavigatorSelector = '.ui-datepicker-header > .ui-datepicker-title > .ui-datepicker-year';
             this.panel.off('change.datePicker-monthNav', monthNavigatorSelector)
                 .on('change.datePicker-monthNav', monthNavigatorSelector, null, this.onMonthDropdownChange.bind($this));
-            this.panel.off('change.datePicker-yearnav keydown.datePicker-yearnav', yearNavigatorSelector)
+            this.panel.off('change.datePicker-yearnav keydown.datePicker-yearnav keyup.datePicker-yearnav', yearNavigatorSelector)
                 .on('change.datePicker-yearnav', yearNavigatorSelector, null, this.onYearInputChange.bind($this))
-                .on('keydown.datePicker-yearnav', yearNavigatorSelector, null, function(event) {$this.onTimeInputKeyDown(event);});
+                .on('keydown.datePicker-yearnav', yearNavigatorSelector, null, function(event) {$this.onTimeInputKeyDown(event);})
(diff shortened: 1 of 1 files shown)


BalusC:
Perfect. Works for me. Tyvm!

PR #13247, opened 2025-01-24
Fixes #13244

### LINKED ISSUE #13244: TriStateCheckbox: switch from string bindings to boolean
currently is mapped as:
2 = Boolean.False
1 = Boolean.True
0 = null

why dont we use Boolean Object direclty?

@melloware WDYT?

tandraschko:
Fixes #13244

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/tristatecheckbox/tristatecheckbox.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/tristatecheckbox/tristatecheckbox.js
@@ -30,7 +30,7 @@ PrimeFaces.widget.TriStateCheckbox = PrimeFaces.widget.BaseWidget.extend({
      * @inheritdoc
      * @param {PrimeFaces.PartialWidgetCfg<TCfg>} cfg
      */
-    init:function (cfg) {
+    init: function(cfg) {
         this._super(cfg);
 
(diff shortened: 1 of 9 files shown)


primefaces/src/main/resources/META-INF/resources/primefaces/tristatecheckbox/tristatecheckbox.js
```diff
@@ -108,7 +108,7 @@ PrimeFaces.widget.TriStateCheckbox = PrimeFaces.widget.BaseWidget.extend({
      * Toggles this tri state checkbox in the given direction. Moves between unchecked, half-checked, and fully-checked.
      * @param {-1 | 1} direction `-1` to move backwards through the states, `+1` to move forward through the states 
      */
-    toggle:function (direction) {
+    toggle: function (direction) {
```
melloware: This space looks like an errant change?  Other than that looks great!

melloware (changes requested):

primefaces/src/main/resources/META-INF/resources/primefaces/tristatecheckbox/tristatecheckbox.js
tandraschko: just aligned to our defaults while looking through code and found another place

PR #13279, opened 2025-01-31
Fix #13278: InputTextArea handle UTF-8 and emojis

### LINKED ISSUE #13278: InputTextarea: countBytesAsChars fails
### 1 Environment

PrimeFaces version: primefaces 14.0.11
Application server: jetty 12.0.16 + version: mojarra40

### 2 Expected behavior

The InputTextarea has an attribute countBytesAsChars. This attribute is intended to change the count behaviour of the textarea counter: Should bytes instead of characters be counted. When countBytesAsChars is true the counterTemplate should count on characters.

Related discussion: https://github.com/orgs/primefaces/discussions/3483

### 3 Actual behavior

When the attribute countBytesAsChars is set to true the counterTemplate "{0} characters remaining." is not displayed/working due to a javascript error. The error is: jQuery.Deferred exception: URI malformed URIError: URI malformed.

### 4 Steps to reproduce

See sample project. Open browser javascript console (F12) to see the error.

### 5 Sample XHTML / Sample bean

See attached project: [primefaces-test-texarea.zip](https://github.com/user-attachments/files/18616193/primefaces-test-texarea.zip)

run with:
mvn clean jetty:run -Pmojarra40

melloware:
Fix #13278: InputTextArea handle UTF-8 and emojis

@OBreidenbach @jepsar @christophs78 you guys worked on this original function but as you can see it doesn't work for Emoji's and other UTF-8 chars.

Switching to https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder makes it work with anything and I also have to update the `applyMaxLength` to respect the byte values.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.utils.js
@@ -16,6 +16,14 @@ if (!PrimeFaces.utils) {
     PrimeFaces.utils = {
 
         /**
+         * TextEncoder instance used for string encoding operations.
+         * Initialized as null and typically set to a TextEncoder instance when needed.
+         * @type {TextEncoder|null}
+         * @private
+         */
+        TEXT_ENCODER: null,
+
+        /**
          * Finds the element to which the overlay panel should be appended. If none is specified explicitly, append the
          * panel to the body.
          * @param {PrimeFaces.widget.DynamicOverlayWidget} widget A widget that has a panel to be appended.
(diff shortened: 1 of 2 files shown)


jepsar:
I can't remember that I've worked on this. I know we have had issues with bytes, but that was back in the day when we were using Apache Trinidad in combination with an Oracle database (where column lengths were in bytes). We had custom validators for that, not a JS count.

melloware:
For reference the discussion was here: https://github.com/primefaces/primefaces/pull/7254

christophs78:
To my best knowledge and looking at git-commit-history i never touched this.
From a short look your changes seem to make sense.
It may be an option to add some more test´s to InputTextArea003Test to cover countBytesAsChars = true and add some Emoji´s and other UTF-8-characters.

PR #13293, opened 2025-02-03
Fix PrimeFaces.resources.isExtensionMapping

blutorange:
Noticed this while refactoring to classes. Not sure if this method is ever used, but `PrimeFaces.resources.isExtensionMapping` always returned `undefined`. The method sets`PrimeFaces.resources.IS_EXTENSION_MAPPING`, but erroneously returns `PrimeFaces.IS_EXTENSION_MAPPING`, which is never set and thus always `undefined`.


melloware (approved):

melloware:
Yes I think this is used by PFE for a few widgets to resolve their URLs. I know it was used heavily in CKEditor but I will check if used by others!

PR #13301, opened 2025-02-04
Fix #13300: AjaxStatus handle f:ajax and p:ajax

### LINKED ISSUE #13300: AjaxStatus: complete event does not fire.
### Describe the bug

After updating from PF 13.0.10 to 14.0.11, we noticed the behavior of p:ajaxStatus had changed. The 'oncomplete' event used to fire after an ajax request completes. Now it is never called and only 'onstart' and 'onsuccess' are called.

Normally the sequence is:
- Start
- Success
- Complete

### Reproducer

PF14 Reproducer: https://github.com/koan00/primefaces-test/tree/ft/AjaxStatus
PF13 behavior: https://github.com/koan00/primefaces-test/tree/ft/AjaxStatusPF13

Click on the on the check box. This should create an ajax request. Review the javascript console on the browser. Note that you only get a 'start' and 'success' message, but no 'complete' message.

### Expected behavior

A 'complete' event is fired after the ajax request completes.

### PrimeFaces edition

Community

### PrimeFaces version

14.0.11

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0

### Java version

17

### Browser(s)

Chrome, edge.

melloware:
Fix #13300: AjaxStatus handle f:ajax and p:ajax

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/ajaxstatus/ajaxstatus.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/ajaxstatus/ajaxstatus.js
@@ -85,23 +85,46 @@ PrimeFaces.widget.AjaxStatus = PrimeFaces.widget.BaseWidget.extend({
     },
 
     /**
-     * Listen to the relevant events on the document element.
+     * Binds event handlers to AJAX events on the document element.
+     * Handles both PrimeFaces AJAX events and native JSF AJAX events.
+     * 
+     * For PrimeFaces AJAX events:
+     * - pfAjaxStart: Triggered when AJAX request starts. After configured delay, triggers 'start' event.
+     * - pfAjaxError: Triggered when AJAX request fails. Triggers 'error' event with xhr, settings, error details.
+     * - pfAjaxSuccess: Triggered when AJAX request succeeds. Triggers 'success' event with xhr, settings.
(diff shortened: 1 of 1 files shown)


tandraschko:
i would really like to have some code docs

melloware:
I just documented everything and the differences between PF and Faces and why we are doing what we do because the Faces Ordering is different than PF ordering.

primefaces/src/main/resources/META-INF/resources/primefaces/ajaxstatus/ajaxstatus.js
```diff
@@ -112,29 +135,33 @@ PrimeFaces.widget.AjaxStatus = PrimeFaces.widget.BaseWidget.extend({
...
                 if(data.status === 'begin') {
+
                     $this.timeout = PrimeFaces.queueTask(function() {
                         $this.trigger('start', arguments);
                     }, $this.cfg.delay);
                 }
                 else if(data.status === 'complete') {
-                    // ignore PF complete event when JSF success/error event is fired right after
+                    // Ignore JSF complete event since it fires before success/error. We'll trigger complete manually after success/error to match PrimeFaces event order
+                    PrimeFaces.debug('Faces complete event fired');
```
tandraschko: do we need this?

melloware: Nope let me remove

PR #13308, opened 2025-02-09
Remove PrimeFaces.setCaretToEnd JS method

blutorange:
`PrimeFaces.setCaretToEnd` is broken. It takes a parameter `element`. Then it calls `element.trigger(...)`, which only works if `element` is a JQuery instance. Afterwards, it calls `element.setSelectionRange(...)`, which only works if `element` is an `HTMLInputElement` or `HTMLTextAreaElement`. No matter what you pass the method, one of these is going to fail.

Fortunately, that method does not seem to be used, at least I did not find any usages.

The intention of the method seems to have been to take a raw `HTMLInputElement` or `HTMLTextAreaElement`, I suppose. I changed that to use `element.focus()` instead, and removed `element.createTextRange` since that is an Internet Explorer-only method and we don't support IE anymore.

https://github.com/primefaces/primefaces/blob/081c590fa19efceddcb2ec98aaba0f9da60cd35e/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.js#L634-L652

### DIFF
--- a/docs/migrationguide/15_0_0.md
+++ b/docs/migrationguide/15_0_0.md
@@ -10,6 +10,7 @@
   * Deleted `autocomplete="off"` from any hidden inputs as is recommended by w3c spec
   * Deleted long time deprecated `primefaces.LEGACY_WIDGET_NAMESPACE`
   * Removed additional themes from primefaces.jar (only saga is still there), you now need to include `primefaces-themes.jar`. Please check the docs. 
+  * Deleted legacy client-side JavaScript method `PrimeFaces.setCaretToEnd`, which did not work and was unused. Use `element.setSelectionRange(0, element.value.length)` on an `HTMLInputElement` or `HTMLTextAreaElement` if you need that functionality.
 
 ## Deprecated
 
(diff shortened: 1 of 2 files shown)


melloware:
Interesting I don't see this used anywhere in the code base?  Is this just a utility function for use outside PF?

melloware:
I wonder if we should remove it? I looked and it was created 13 years ago by Optimus and it has had no real changes since?

blutorange:
Yeah, I found commit `e7dafbd90ace0c358a5d223983a4ec21f4ef04c8` where it was actually used in the PrimeFaces code. I guess it's probably also safe to remove, especially since a major version is about to be released. And nobody really could have used that method, since it didn't work. I updated the PR, remove the function and added a note to the docs.

melloware (approved):

blutorange:
While we're on the topic, `PrimeFaces.hasSelection` also seems broken, it checks `getSelection().length > 0`, but `getSelection()` returns a `Selection` object which doesn't have a `length` property, and so the function always returns `false`. I also wonder if perhaps `getSelection` was meant to return the string text, not a `Selection` object? Do you know if these functions are used, search for `PrimeFaces.getSelection` / `PrimeFaces.hasSelection` didn't return any result.

image

https://github.com/primefaces/primefaces/blob/081c590fa19efceddcb2ec98aaba0f9da60cd35e/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.js#L757-L759

melloware:
I have run into this recently. Apparently Chrome and Firefox work differently with getSelection()  let me find where I just hit that.

melloware:
I don't think either of these methods is used either!

melloware:
I say remove them.

blutorange:
Alright, I removed them.

melloware (approved):

tandraschko:
did you check PF extensions?

melloware:
Confirmed none of these are used in PFE

PR #13309, opened 2025-02-09
#13278: Textarea, countBytesAsChars, Integrationtests

christophs78:


melloware (approved):

christophs78:
> @christophs78 this Selenium Test just started failing regularly and it was not failing previously.
> 
> ```
> SelectCheckboxMenu002Test.submitEmptyLabel:56 » ElementClickIntercepted element click intercepted: Element <button id="form:button" name="form:button" class="ui-button ui-widget ui-state-default ui-button-text-only" type="submit" data-ajax="true">...</button> is not clickable at point (151, 145). Other element would receive the click: <div class="ui-widget-header ui-selectcheckboxmenu-header ui-helper-clearfix">...</div>
> ```

Just plain coincidence. Not related.
See https://github.com/primefaces/primefaces/issues/13330

PR #13324, opened 2025-02-11
Fix prefix check in PrimeFaces.ajax.Request.addParams

blutorange:
Fix prefix check in PrimeFaces.ajax.Request.addParams

See #13313

The reason why I wanted to ask is that right now, the prefix never got added; and somehow nobody noticed that something wasn't working.


tandraschko:
Its only used in liferay, maybe not that much Users...

melloware (approved):

PR #13333, opened 2025-02-14
#13330 fix SelectCheckboxMenu002Test#submitEmptyLabel for Chrome 133

### LINKED ISSUE #13330: Integration Test Failure
Recent test failures on both 14 and 15 branches but nothing out of the ordinary as far as commits and can't quite tell which commit it started with?

```
SelectCheckboxMenu002Test.submitEmptyLabel:56 » 
ElementClickIntercepted element click intercepted: Element <button id="form:button" name="form:button" class="ui-button ui-widget ui-state-default ui-button-text-only" type="submit" data-ajax="true">...</button> 
is not clickable at point (151, 145). 
Other element would receive the click: <div class="ui-widget-header ui-selectcheckboxmenu-header ui-helper-clearfix">...</div>
```

christophs78:
Fix #13330


melloware:
Thanks !  I will backport to 14.X too

PR #13362, opened 2025-02-20
Fix windows build for frontend project

blutorange:
@melloware 

This should fix the Window build. It's pretty late here, need to sleep now : /

Also: I'm open to revert the merge if anybody does not feel comfortable with the build system.  We need to all feel ok and be on the same page.

The issue was that I was trying to read a JSON file via JavaScript's `import`, and that doesn't work with absolute paths. That was just node.js that complained about it. I changed it to reading the JSON file directly via node's `fs` module. The build on he Window device I was using now works.

> study it more and document it

For what it's worth, I wrote a little bit in the readme here https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/frontend/README.md

Very brief summary: 

>The PrimeFaces JAR contains multiple different bundles, such as core.js, components.js, or terminal/terminal.js
>
> For each bundle, there's a folder in `packages/`.
>
> The build script build/bin/bundle.mjs lists all folders and invokes esbuild on each folder, to produce the final
bundled file (core.js, components.js etc.).

### DIFF
--- a/primefaces/src/main/frontend/build/common/build-settings.mjs
+++ b/primefaces/src/main/frontend/build/common/build-settings.mjs
@@ -1,3 +1,5 @@
+import fs from "node:fs/promises";
+
 /**
  * Allows a sub folder to affect the build settings when building the
  * frontend project (via ESBuild).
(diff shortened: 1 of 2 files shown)


melloware:
OK next issue all the icons are messed up

image

when I run the showcase with `mvn -Ddev`

melloware:
The icon URLS are munged and getting 404's

image

awa-xima:
:/ Probably an issue with `\` vs `/`. I'll take a look.

PR #13379, opened 2025-02-21
Revert hammerjs and core-emitter

melloware:
@tandraschko @blutorange these two i had to roll back their versions because they are locked to vis-timeline versions.

I wish we could add comments in JSON but unforuntately that is not in the JSON spec.


blutorange:
There's a JSON with comments standard (jsonc), but not many tools support it.

If it works the downgrade seems fine to me.

blutorange:
Also I guess you are referring to these lines from the build output? 

Seems good we should downgrade, but I don't think that fixes the actual build error regarding that it can't find `jqueryui`.

Tthe build with Windows is working for you now?

> [INFO] ➤ YN0060: │ component-emitter is listed by your project with version 2.0.0 (p42a13), which doesn't satisfy what vis-timeline and other dependencies request (^1.3.0).
> [INFO] ➤ YN0060: │ propagating-hammerjs is listed by your project with version 3.0.0 (p8ab86), which doesn't satisfy what vis-timeline requests (^1.4.0 || ^2.0.0).

melloware:
The build on windows was working before this just showing those YARN warnings. I wonder if @tandraschko somehow it was picking it up as a failure?    My OS and JDK

```
Apache Maven 3.9.9 (8e8579a9e76f7d015ee5ec7bfcdc97d260186937)
Maven home: C:\tools\apache-maven-3.9.9
Java version: 21.0.4, vendor: Eclipse Adoptium, runtime: C:\tools\jdk-21.0.4+7
Default locale: en_US, platform encoding: UTF-8
OS name: "windows 11", version: "10.0", arch: "amd64", family: "windows"
```

melloware:
@blutorange i was wrong this just fixes the warnings but NOT the error's he is seeing.  But I am not seeing those errors on my Windows machine.

blutorange:
Included in #13386, so this can be closed.

melloware:
Closed!

PR #13385, opened 2025-02-21
Fix #13346: Schedule events must be called after the viewChange event is fired

### LINKED ISSUE #13346: Schedule: viewName Updates Too Late in loadEvents()
### Describe the bug

The issue you're encountering in PrimeFaces' <p:schedule> component is due to the order of event processing in JSF and PrimeFaces' AJAX behavior. Specifically, the viewChange event updates the viewName property after the schedule component requests events. This causes viewName to still hold the previous value when loadEvents() is executed.

Why is this happening?
The viewChange event is triggered after the schedule component attempts to load events.
When PrimeFaces initializes the schedule, it invokes loadEvents() using the old value of viewName because the change event hasn't been processed yet.
The onViewChange method updates viewName, but it is already too late—the method loadEvents() has already been executed with the outdated viewName.

You worked around the issue by using selectors to target buttons: fc-dayGridMonth-button and fc-timeGridWeek-button. Before executing the action, you added an event that triggers an update of the viewName value.
This ensures that viewName is updated before PrimeFaces loads the events, preventing outdated values from being used in loadEvents().

### Reproducer

Change view, and watch console, there is sout: "I want here to see correct viewName".... 
Reproducer with example: https://github.com/kamilkrzywanski/primefaces-test/tree/newTerribleIssue

### Expected behavior

Correct viewName on events load

### PrimeFaces edition

None

### PrimeFaces version

13,14

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4

### Java version

17,21

### Browser(s)

_No response_

melloware:
Fix #13346: Schedule events must be called after the viewChange event is fired

### DIFF
--- a/primefaces/src/main/frontend/packages/schedule/schedule/src/schedule.widget.js
+++ b/primefaces/src/main/frontend/packages/schedule/schedule/src/schedule.widget.js
@@ -318,29 +318,31 @@ PrimeFaces.widget.Schedule = class Schedule extends PrimeFaces.widget.DeferredWi
         var $this = this;
 
         this.cfg.options.events = function(fetchInfo, successCallback) {
-            var options = {
-                source: $this.id,
-                process: $this.id,
-                update: $this.id,
-                formId: $this.getParentFormId(),
-                params: [
-                    {name: $this.id + '_event', value: true},
-                    {name: $this.id + '_start', value: fetchInfo.start.toISOString()},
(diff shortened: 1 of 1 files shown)


blutorange:
As I said, looks good to me and I had also tried it locally, seemed to have fixed the issue :+1:

PR #13387, opened 2025-02-23
Fix client side email validator for 15.x

blutorange:
Noticed this while working on [refactor-part-2-convert-core-to-typescript](https://github.com/blutorange/primefaces/tree/refactor-part-2-convert-core-to-typescript). The client-side Email bean validator does not work. Violations are caught only by the server when the form is submitted.

Can be seen in the showcase when you add input field for emails: Enter an invalid email, then click on `save`. It will show validation errors for other fields that are required and empty, but not for the email input field. 

As per the Validator interface, implementations must not return anything, but throw a validation message when constraint is violated.

Perhaps a candidate for the next 15.0.X release.

```ts
        /**
         * A validator for checking whether the value of an element confirms to certain restrictions.
         */
        export interface Validator {
            /**
             * Validates the given element. If it is not valid, an error message
             * of type {@link BaseFacesMessage} (or an array thereof) should be
             * thrown (via a `throw ...` statement).
             * @param element Element to validate
             * @param value Current value of the element
             * @throws The error message when the element with its current value
             * is not valid. Can be either a {@link BaseFacesMessage} or an
             * array thereof.
             */
            validate(element: JQuery, value?: unknown): void;
        }
```

### DIFF
--- a/primefaces-showcase/src/main/java/org/primefaces/showcase/view/csv/BeanValidationView.java
+++ b/primefaces-showcase/src/main/java/org/primefaces/showcase/view/csv/BeanValidationView.java
@@ -30,6 +30,7 @@ import jakarta.inject.Named;
 import jakarta.validation.constraints.AssertTrue;
 import jakarta.validation.constraints.DecimalMax;
 import jakarta.validation.constraints.Digits;
+import jakarta.validation.constraints.Email;
 import jakarta.validation.constraints.Future;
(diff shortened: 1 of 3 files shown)


blutorange:
Also just wondering, is it normal  that validation messages are not shown next to the input field anymore in the 15.x showcase?

Showcase 14.0.6
image

Showcase 15.x
image

PR #13395, opened 2025-02-24
Ensure cache folder on same drive with Windows #13372

blutorange:
I added a Maven profile that gets activated on Windows. It uses antrun to check the drive letters and if they don't match, changes the cache folder to the `<DRIVE_LETTER>:\.yarn\berry\cache`.

Worked for me with Windows and Linux. But please also test on your machine.

This way, the Maven build should always work. Only when you want to install node.js and yarn manually and run `yarn run build` on the `primefaces/src/main/frontend` folder manually, you may have to set an environment variable (see readme).

If this doesn't work out, then we'll just set `enableGlobalCache: false`. There's already an environment variable for turning it on/off.


tandraschko:
enableGlobalCache just deactivates a yarn-shared-cache over multiple projects, right?

TBH i would prefer to disable this cache in favor of adding such antrun hacks

blutorange:
Yes, `enableGlobalCache: false` will create a folder in the current git repo with all `*.zip` files from the NPM dependencies we use.

`enableGlobalCache: true` is like Maven's `~/.m2/repository`, `enableGlobalCache: false` is like copying all needed JAR files to each project (though that happens manually and is gitignored).

But I'm open to either option -- I can also just disable the global cache.

blutorange:
> Actually let me think about this. If I am doing builds constantly with mvn -Pquick i don't want it to re-run and NPM install for all the dependecies each time?

If we put the folder int the gitignore, but not in the maven clean plugin, that shouldn't be an issue. The folder with the *.zip files just stays there.

melloware:
OK

blutorange:
Sorry it took a bit. Reverted the antrun configuration and disabled the global cache.

melloware (approved):

PR #13440, opened 2025-02-26
Fix #13438: IdleMonitor stack overflow

### LINKED ISSUE #13348: DatePicker: when using mask and CSP, not working on mobile
### Describe the bug

A datePicker with mask doesn't call the listener associated when a date is changed (or removed) manually in the input.
Calendar of datePicker doesn't allow to select the date.

It doesn't work only if CSP is enabled and MOVE_SCRIPTS_TO_BOTTOM is enabled AND page is resized as on mobile devices
primefaces.CSP true primefaces.MOVE_SCRIPTS_TO_BOTTOM true 

### Reproducer

[pf.zip](https://github.com/user-attachments/files/18842720/pf.zip)

### Expected behavior

Listeners should be called and date should be seleted from calendar

### PrimeFaces edition

Community

### PrimeFaces version

14.0.12

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0.9

### Java version

17

### Browser(s)

_No response_

### LINKED ISSUE #13438: IdleMonitor causing Too much recursion error
### Describe the bug

When using the idlemonitor just simply printing out a console message I'm getting an `Too Much Recursion Error`

This is in Firefox and Safari on MacOS 15.2 running PF 15

```
 <!-- Monitor user inactivity; triggers when the calculated timeout is reached -->
        <p:idleMonitor timeout="5000"
                       onidle="console.log('IDLE')">
        </p:idleMonitor>
```

Image

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

15.0

### Theme

_No response_

### JSF implementation

MyFaces

### JSF version

4.1

### Java version

17

### Browser(s)

_No response_

melloware:
Fix #13438: IdleMonitor stack overflow

### DIFF
--- a/primefaces/src/main/frontend/packages/idlemonitor/idlemonitor/src/idlemonitor.widget.js
+++ b/primefaces/src/main/frontend/packages/idlemonitor/idlemonitor/src/idlemonitor.widget.js
@@ -44,14 +44,14 @@ PrimeFaces.widget.IdleMonitor = class IdleMonitor extends PrimeFaces.widget.Base
                 $this.cfg.onidle.call($this);
             }
 
-            $this.callBehavior('idle');
+            $this.callBehavior('idle', undefined, false);
         })
         .on("active.idleTimer" + this.cfg.id, function(){
             if($this.cfg.onactive) {
                 $this.cfg.onactive.call($this);
             }
 
-            $this.callBehavior('active');
+            $this.callBehavior('active', undefined, false);
         });
 
         $.idleTimer(this.cfg.timeout, document, this.cfg.id);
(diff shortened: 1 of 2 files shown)


primefaces/src/main/frontend/packages/idlemonitor/idlemonitor/src/idlemonitor.widget.js
```diff
@@ -44,14 +44,18 @@ PrimeFaces.widget.IdleMonitor = class IdleMonitor extends PrimeFaces.widget.Base
                 $this.cfg.onidle.call($this);
             }
 
-            $this.callBehavior('idle');
+            if($this.hasBehavior('idle')) {
```
tandraschko: IMO callBehavior should check "hasBehavior" by itself

i introduced hasBehavior for other cases, were a manual AJAX request needs to be fired

tandraschko: ahh, i see
maybe we should add a param "fallbackToDomEvent" with default true, and set this to false here

melloware: OK i thought about that i like that idea!

tandraschko:
+1

PR #13445, opened 2025-02-28
resolve deprecation warnings

tandraschko:


melloware (approved):

tandraschko:
nice
seems that 4.1 needs a newer java version at all
and mojarra 4.1 is not compatible with older CDI versions

tandraschko:
@melloware can we configure a second build matrix like:
Java11 = 4.0
Java17 = 4.1
Java23 = 4.0 + 4.1?

melloware:
yeah i think we can want me to tackle that?

tandraschko:
yeah plz

melloware:
OK will take care of it!

tandraschko:
otherwise im fine with one 4.1 impl for now as well

melloware:
yep will look into both!

PR #13446, opened 2025-02-28
Fix build matrix

melloware:
Trying this...


tandraschko:
TBH i think we should not change showcase, it should be run on 4.0 as minimum

if we use EE11 API, we cant verify it works on 4.0

melloware:
OK all passing all permutations!

PR #13454, opened 2025-02-28
Fix #13453: LazyModel doc update and migration guide

### LINKED ISSUE #13453: DataTable: With custom lazy model not tracking rowCount
### Describe the bug

Original Report: https://github.com/orgs/primefaces/discussions/3645

@tandraschko with this fix: https://github.com/primefaces/primefaces/issues/12362

Now a user with a custom lazy model needs to know to update rowCount in the `load` method.

When using code from here https://github.com/primefaces/primefaces/blob/master/primefaces-showcase/src/main/java/org/primefaces/showcase/view/data/datatable/LazyCustomerDataModel.java
on PF v14.0.12 - datatable Paginator Template is updating completely correct on Filter.

Once moved exact same code to 15.0.0 - datatable Paginator Template is NOT updating at all.

### Reproducer

_No response_

### Expected behavior

Paginator should update.

### PrimeFaces edition

Community

### PrimeFaces version

15.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0

### Java version

11

### Browser(s)

_No response_

melloware:
Fix #13453: LazyModel doc update and migration guide

### DIFF
--- a/docs/15_0_0/components/datatable.md
+++ b/docs/15_0_0/components/datatable.md
@@ -897,7 +897,8 @@ public class CarBean {
 
             @Override
             public List<Car> load(int first, int pageSize, Map<String, SortMeta> sortBy, Map<String, FilterMeta> filterBy) {
-                // load physical data
+                // load physical data and filter it
+                setRowCount(rows); // set new number of rows after filtering.
                 return requestedResultPage;
             }
         };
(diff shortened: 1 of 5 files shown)


tandraschko:
Can you are a test first?
It should not be needed If #count is implemented

melloware:
Count doesn't work because of the optimization you made in the ticket.

melloware:
This ticket I think is the one https://github.com/primefaces/primefaces/issues/12362

But not sure

melloware:
Could also be this one: https://github.com/primefaces/primefaces/issues/11506

tandraschko:
My issue should only change DefaultLazyDataModel

We need a test

PR #13456, opened 2025-03-01
Convert core to TypeScript + fix type errors + some related widgets

blutorange:
As promised, I had started working on part 2 of the frontend refactor:

Here I convert the core scripts to TypeScript (the easy part, renamed to *.ts and inline types from JSDocs) and fix type errors (the hard part).

__Please take a look at the general code structure before merging (and wait for automated tests)__ E.g. `primefaces/src/main/frontend/packages/core/` and let me know what you think.

The basic idea is that each package (e.g. core) uses `import / export` internally. And exposes some stuff to the global scope, which can be used by other packages (e.g. components).

Only  the `index.ts` should "do something", i.e. register event listeners or add new functions to `PrimeFaces...`. All other files should simply export classes, functions and constants.

This PR is basically done, I need to wait for automated tests and fix some issues, if any.

---

* Convert core to TypeScript + fix type errors, as well as some related widgets.
* Running TypeScript now reports no errors. (There are still some widgets to convert to TypeScript, though).
> cd primefaces/src/main/frontend && yarn tsc -b
* Dont log spurious error messge in destroyAttachedWidgets when widget does not exist anymore.
* Fix bean validation keys (there was an *n* missing in `jakarta.validatio.`)
* Add frontend dist folder to maven clean
* Slightly more log output for bundle script
* Add simple hook mechanism to core that lets external code hook into functionality. Needed for proper typing. Previously, core.js was e.g. trying to access PrimeFaces.widget.Dialog and its methods, which does not type check, as Dialog is declared much later in components.js.
* Use more types from DefinitelyTyped (@typed/...), some of which I added.


blutorange:
For the record, there was also a type on the bean validation messages that I fixed

image

melloware:
Looking at this I see you moved some things like BlueImp to devDependencies?  But isn't BlueImP bundled in so isn't it a compile dependency?

melloware:
OK looks like `Csv001Test` and a bunch of DatePicker tests failed.

blutorange:
> Looking at this I see you moved some things like BlueImp to devDependencies? But isn't BlueImP bundled in so isn't it a compile dependency?

The JavaScript ecosystem is different from Maven (Maven makes sense, JS doesn't ;) devDependencies are what's needed for building. dependencies are what other people that install our package need to install.

And since we are bundling everything, people don't need to install `downloadjs`. But they need to install typings such as `@types/downloadjs` (which we don't bundle).

Some packages include their types directly in their package, so these packages need to remain in `dependencies` (e.g. `@fullcalendar`)

melloware:
Thanks for the explanation.  It does always trip me up.

blutorange:
> OK looks like Csv001Test and a bunch of DatePicker tests failed.

I would have been very surprised if nothing failed. I already took a look at the showcase and tested it, looked good. But it's awesome there are now automated tests for more cases :+1: Seems like a bunch of failures causes by the same issue, I'll take a look

blutorange:
Turns out it was a simple issue, `jquery.ui.prime.datepicker.cjs` isn't type checked yet, so of course there was a lingering `undefined` error. Tests are now passing.

I need to take a few weeks off to do some other life related things. After that, my plan is to take a look at the remaining widgets and convert those to TypeScript as well, probably several small PRs. (If anybody wants to take a stab at converting a widget, feel free.) Once that's all done, we only need to adjust the build so that it type checks via TypeScript and generates documentation via typedoc.

I think we can easily get that done before the 16.x release -- but just to make sure: PF 16 won't come out for some months at least?

melloware:
Yes you have plenty of time 16.X not coming out for a few months

blutorange:
Anyways, since all tests are passing now, this PR is ready now as far as I'm concerned.

tandraschko:
@blutorange is it normal that intellij shows errors? i would like to have a working IDE :D

melloware:
@tandraschko i was seeing the same thing but I did a full clean and rebuild of my workspace and now I don't see it.

image

blutorange:
I normally use VSCode for frontend code. But when I opened it in IntelliJ:

* It worked immediately when I opened the `primefaces/src/main/frontend` folder
* It didn't work when I opened the root folder. What I noticed was that it asked me to run `npm install`, but it needs to be `yarn install`.
* I went to `settings` -> `Languages & Frameworks` -> `Node.js` -> `Package manager` and selected `yarn`. After a restart, it worked and right clicking on the `package.json` offered the option to run `yarn install`.

melloware:
we need to add an IntelliJ section here @blutorange https://github.com/primefaces/primefaces/wiki/IDE-Configuration

Also feel free to review my VSCode section for any additions or improvements!

PR #13474, opened 2025-03-03
Fix Firefox test

melloware:
From stack overflow: https://stackoverflow.com/a/32882849/502366


melloware:
cc @blutorange

blutorange:
The interesting part is that [mozilla's documentation for TouchEvent](https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent) marks Firefox as supporting that event -- but I guess it's only available on firefox on mobile devices? Strange, but seems fine to check.

melloware:
Yep whacky world of browsers!

PR #13479, opened 2025-03-04
Fix #13543: Fix getRowCount for lazy

### LINKED ISSUE #13453: DataTable: With custom lazy model not tracking rowCount

melloware:
Fix #13453: Failing integration test

When filtering notice the paginator still thinks there are 100 rows even though `count()` is correct.

image

Once you click on any page then it corrects itself.

image


tandraschko:
also the ProgrammingLanguageLazyDataModel has a #count impl but still calls setRowCount, we should also fix this one

melloware:
OK i removed it from `ProgrammingLanguageLazyDataModel` so more tests i would think will fail.  Then I will try and revert that fix

melloware:
@tandraschko i reverted that and all passing now!

melloware:
I just reverted the DataTable one not the other components.

melloware:
I looked at his original ticket and his intent was to fix DataView and assumed it was a problem for DataTable.   But it was not.

So I pinged @cocorossello and asked him to chime in.

melloware:
Will do

PR #13487, opened 2025-03-05
Fix #12519: Remove NestedSelectItem/PF

### LINKED ISSUE #12519: Core: clean up PF utils

melloware:
Fix #12519: Remove NestedSelectItem/PF


melloware:
@tandraschko wasnt sure if you wanted to keep the PF class. I am fine with it if you want.

PR #13503, opened 2025-03-07
Fix #13211: FileUpload multiple mode only discard invalid files

### LINKED ISSUE #13211: FileUpload: CSV Multiple fileupload doesn't upload files, if one of them is not valid
### Describe the bug

Hi,

There is a change in the fileUpload reimplementation which is (in my opinion) not correct. Until PF13, when a user used multiple fileupload and tried to upload multiple files, if one of them did not meet the criteria, the others were successfully uploaded.
Since PF14, there has been a change which prevents uploading any file in such cases.

In the previous implementation, this code was in the add method in 2-fileupload.js:
```
            add: function(e, data) {
...
   var validMsg = $this.validate(file);

// do some stuff on the base of validMsg

```
Each file was handled separately.

The current implementation is:
```
            add: function(e, data) {
...
                    var validationResult = PrimeFaces.validation.validate($this.jq, dataFileInput, update, true, true, true, true);
                    if (!validationResult.valid) {

```

This validate method is called each time any file is added, but PrimeFaces.validator['primefaces.File'] validates the entire set of files, and if any of them is not valid, it returns false.

Therefore, there are two possible bugs (improvements) that should be considered for fixing (if possible):

- Prevent multiple validation (whilst the validation of files is fast because it only validates name and size, it validates all files n² times)
- Restore the possibility to upload files which are correct - and report only invalid files

I haven't tested it with server-side validation, just client-side. It might work correctly with server-side validation; however, in my opinion, server-side validation for file upload is not the right approach.

### Reproducer

On the PF showcase (https://www.primefaces.org/showcase/ui/file/upload/multiple.xhtml), try to upload JPG and XLS files; none of them are uploaded. The JPG file(s) should be uploaded.

### Expected behavior

Files that meets validation criterias should be uploaded

### PrimeFaces edition

Community

### PrimeFaces version

14.0+

### Theme

_No response_

### JSF implementation

All

### JSF version

all

### Java version

all

### Browser(s)

all

melloware:
Fix #13211: FileUpload multiple mode only discard invalid files

@tandraschko what this does is it looks for the `file.name` in the error message for a match.  If found it knows to only exclude this file and continue processing others.   If for some reason the error messages don't match the behavior is NO DIFFERENT than it is now.

But this restores the behavior of PF13 and before.

### DIFF
--- a/primefaces/src/main/frontend/packages/fileupload/fileupload/src/fileupload.widget.ts
+++ b/primefaces/src/main/frontend/packages/fileupload/fileupload/src/fileupload.widget.ts
@@ -412,11 +412,25 @@ export class FileUpload<Cfg extends FileUploadCfg> extends PrimeFaces.widget.Bas
                 var file = data.files ? data.files[0] : null;
                 if (file) {
                     this.clearMessages();
-
-                    // we need to pass the real invisible input, which contains the file list
-                    var validationResult = PrimeFaces.validation.validate(this.jq, dataFileInput, update, true, true, true, true, false);
+ 
+                    // #13211 Create a new DataTransfer object to hold just the current file
+                    const dataTransfer = new DataTransfer();
+                    dataTransfer.items.add(file);
+                    // Create a temporary file input element
+                    const singleFileInput = $<HTMLInputElement>('<input type="file" />');
+                    // Set the files on the temporary input
+                    (singleFileInput[0] as HTMLInputElement).files = dataTransfer.files;
+                    // Copy validation metadata from original input to temporary one
+                    singleFileInput.data(dataFileInput.data());
+ 
+                    // pass the single file input being processed
+                    const validationResult = PrimeFaces.validation.validate(this.jq, singleFileInput, update, true, true, true, true, false);
+ 
+                    // clean up the temporary file input
+                    singleFileInput.remove();
+ 
+                    // if the file is invalid, show the message
                     if (!validationResult.valid) {
-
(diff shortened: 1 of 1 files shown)


melloware:
yeah the issue is that because `PrimeFaces.validation.validate(this.jq, dataFileInput, update, true, true, true, true, false);` validates all the files. So if you add 10 files and even though you are processing file 1 it validates all 10. Then in file 2 it validates all 10.

So this simply scans the results and says it only cares about if the file.name matches.  Without a major overhaul of `PrimeFaces.validation.validate` this was the least invasive fix to make it work like PF13 did.

tandraschko:
im just not sure if that all is correct TBH

just take:
```var file = data.files ? data.files[0] : null;```

why does it take the first file in a file list? is that really the newest added file?

and so on

i just say that this method needs propably some more love and a hacky/easy solution should not be our main goal

melloware:
see data.files is always an array of 1 file in this mode.

image

tandraschko:
So invalid files are not removed from the Input?

melloware:
So in this video I am going to select 3 files. 2 are valid 1 is invalid.

fileupload

tandraschko:
The video is with this PR applied?

melloware:
Yes video is with PR applied without it none of the files are accepted.

melloware:
OK PR updated your suggestion works. I create a temporary single file input.  Copy all the `data-val` fields from the main input so it gets processed. Validate it then clean it up.

tandraschko:
nice solution

Issue #13488, opened 2025-03-06
Dialog: On mouse scrolling, the dialog moves back to its original position


mulia-dhamma:
### Describe the bug

Dialog : Is it possible to let the dialog position stay at its last position ?

https://github.com/user-attachments/assets/ac8af577-5874-43e5-a4d7-e026d29095c2

### Expected behavior

The dialog position should remain fixed at the last position, even as the user scrolls the page.

### PrimeFaces edition

Community

### PrimeFaces version

14.0.6

### JSF implementation

Mojarra

### JSF version

2.0

### Java version

11

### Browser(s)

Google Chrome

melloware:
@mulia-dhamma yep turn `responsive="false"`

See: https://primefaces.github.io/primefaces/vdldoc/p/dialog.html#responsive

PR #13516, opened 2025-03-09
Improve dialog scrollability

BalusC:
Consider a dialog with "really long content".
```xml
<p:dialog header="Dialog header" modal="true" width="500">
    <div style="min-height: 1500px">
        Some really long content
    </div>
</p:dialog>
```
By default, this is not scrollable. Only those served by 'Dialog Framework' are scrollable but even then it's with a fixed header which doesn't give nice visual user experience.

Setting CSS `position:absolute` at `.ui-dialog` comes close but it still doesn't yield nice visual user experience. It basically scrolls *together* with the body instead of *over* it. The ultimate goal is to achieve same markup and user experience as seen with Bootstrap modal here at section "[Scrolling long content](https://getbootstrap.com/docs/5.0/components/modal/#scrolling-long-content)". It's possible to achieve this with pure CSS, but the `<p:dialog>` is missing a real "container" element.

This PR makes the necessary adjustments to `DialogRenderer`, `dialog.js` and `dialog.css` to achieve this. Basically, it wraps all children of `<div class="ui-dialog>` in a new `<div class="ui-dialog-container>` element which in turn represents the 'real' dialog. The `<div class="ui-dialog>` basically becomes the scrollable overlay.

Unfortunately, this is not fully backwards compatible with possibly existing custom CSS on `.ui-dialog`. It could probably be improved by swapping `ui-dialog` with `ui-dialog-container` elements, but you would then still need to swap along the `ui-widget` related classes to satisfy jQuery UI rules and this would still possibly break existing custom CSS on `.ui-dialog.ui-widget` selectors. So this backwards incompatibility wrt custom CSS probably needs to be documented and this whole thing probably needs to be moved to PrimeFaces 16.X due to the breaking nature and 15.0.0 already being released.

Nonetheless, let me know if this markup is OK or if I should swap it or if I should rename `ui-dialog-container` to something else. Also note that there's a constant `Dialog.CONTAINER_CLASS` which is ultimately set on `ui-dialog` element. It could in long term probably cause confusion and this should probably also be renamed or swapped for clarity.

Oh, we should probably also review relevance of `blockScroll` and `positionType` attributes. These could be removed now. Only the `fitViewport` is fine, it basically makes the `ui-dialog-content` scrollable which continues to work fine.


melloware:
@BalusC we have some users that do want the dialog to scroll with the browser scroller like this scenario: https://github.com/primefaces/primefaces/issues/13488

Would this still work with these changes?

BalusC:
Ok, understood. I removed implicit `ui-overflow-hidden` as I figured out `blockScroll` already does exactly that.

So, the original code
```xml
<p:dialog header="Dialog header" modal="true" width="500">
    <div style="min-height: 1500px">
        Some really long content
    </div>
</p:dialog>
```
will still create a scrollable dialog on long content, but the body will also scroll along when dialog hits its scoll limit.

And to achieve the original intent, simply add `blockScroll="true"`
```xml
<p:dialog header="Dialog header" modal="true" blockScroll="true" width="500">
    <div style="min-height: 1500px">
        Some really long content
    </div>
</p:dialog>
```

melloware:
Awesome let me confirm a few more scenarios tomorrow we have had through the years

melloware:
Thank you!

BalusC:
I tested all possible permutations of blockScroll=true, fitViewport=true and positionType=absolute with short/long content and made sure they open OK and are draggable and resizable just fine.

Noted should be that combining `positionType="absolute"` with `blockScroll="true"` doesn't make sense.

primefaces/src/main/resources/META-INF/resources/primefaces/dialog/dialog.js
```diff
@@ -97,8 +97,9 @@ PrimeFaces.widget.Dialog = PrimeFaces.widget.DynamicOverlayWidget.extend({
     init: function(cfg) {
         this._super(cfg);
 
-        this.content = this.jq.children('.ui-dialog-content');
-        this.titlebar = this.jq.children('.ui-dialog-titlebar');
+        this.container = this.jq.find('.ui-dialog-container');
```
melloware, line 100: Need JSDoc for `container` and update the description of `content` up top for our TypeScript check to pass.

primefaces/src/main/resources/META-INF/resources/primefaces/dialog/dialog.js
```diff
@@ -108,6 +109,7 @@ PrimeFaces.widget.Dialog = PrimeFaces.widget.DynamicOverlayWidget.extend({
         this.jqEl = this.jq[0];
 
         this.positionInitialized = false;
+        this.containerMargin = 20;
```
melloware, line 112: Need JSDoc comment in class header for this item

primefaces/src/main/java/org/primefaces/component/dialog/DialogRenderer.java
```diff
@@ -116,6 +116,9 @@ protected void encodeMarkup(FacesContext context, Dialog dialog) throws IOExcept
             writer.writeAttribute("style", style, null);
         }
 
+        writer.startElement("div", null);
+        writer.writeAttribute("class", Dialog.CONTAINER_CLASS, null);
```
melloware, line 120: Looks like ConfirmDialog selenium tests are failing it might be related to this new element.

melloware (changes requested):

pizzi80:
Very good

A nice add on could be a fixed at bottom footer, 
specially on mobile screen size:

When you have a very long form, even if it's scrollable, 
if you want to edit some inputs and then click "save"
you have to scroll down a lot ...

This is the trickiest part to create a "perfect" modal...

BalusC:
will address comments this weekend

BalusC:
I'll recreate the PR as well, there are for unclear reason 44 conflicts on mostly unrelated files when I try to merge master in order to retarget for 16.x.

BalusC:
Hmm looks like project structure has changed? I can't find dialog.js anywhere now.

BalusC:
Ok it appears to be moved here https://github.com/primefaces/primefaces/tree/master/primefaces/src/main/frontend/packages/components/src/dialog

melloware:
yep we just refactored for TypeScript!

BalusC:
And PrimeFaces 16 appears to have a hard dependency on commons IO now:
```
Cause: Class 'org.primefaces.component.imagecropper.ImageCropperRenderer' is missing a runtime dependency: java.lang.NoClassDefFoundError: org/apache/commons/io/input/BoundedInputStream
```
Is this intentional?

BalusC:
And during render of `<p:button>` in test page it appears to have a hard dependency on OWASP encoders as well:
```
java.lang.NoClassDefFoundError: org/owasp/encoder/Encode
	at org.primefaces.util.EscapeUtils.forJavaScript(EscapeUtils.java:138)
	at org.primefaces.component.button.ButtonRenderer.buildOnclick(ButtonRenderer.java:110)
```

melloware:
it gets auto shaded in `mvn clean install` we shade a few classes from Commons IO and OWASP Encoder.

melloware:
on Primefaces i usually run `mvn -Pquick` that does a quick clean install of what you need.

BalusC:
Ok it was just the Eclipse workspace resolution which didn't run the shade job.

BalusC:
New PR here: https://github.com/primefaces/primefaces/pull/13575

Note I renamed ui-dialog-container to ui-dialog-box over there for clarity and also added some improvements/fixes to issues I found related to dragging/resizing, especially with positionType=absolute.

PR #13523, opened 2025-03-11
Showcase switch primeflex and prismjs to webjars

melloware:
This fixes a CVE in prism.js and will also lets us know with Dependabot when there are new versions


tandraschko:
is primeflex also in the docs and should we update?

melloware:
Oh yeah let me update the docs too.

PR #13537, opened 2025-03-12
Fix #13532: Captcha respect defaults

### LINKED ISSUE #13532: Captcha: NPE on validation
### Describe the bug

After upgrading to v15, we are getting errors on form submit with captcha.

```
java.net.MalformedURLException: Cannot invoke "String.length()" because "spec" is null
	at java.base/java.net.URL.<init>(URL.java:806) ~[?:?]
	at java.base/java.net.URL.<init>(URL.java:654) ~[?:?]
	at java.base/java.net.URL.<init>(URL.java:590) ~[?:?]
	at org.primefaces.component.captcha.Captcha.validateValue(Captcha.java:74) ~[primefaces-15.0.0-jakarta.jar:15.0.0]
	at jakarta.faces.component.UIInput.validate(UIInput.java:717) ~[myfaces-api-4.1.0.jar:4.1.0]
	at jakarta.faces.component.UIInput.processValidators(UIInput.java:297) ~[myfaces-api-4.1.0.jar:4.1.0]
	at jakarta.faces.component.UIComponentBase.processValidators(UIComponentBase.java:1379) ~[myfaces-api-4.1.0.jar:4.1.0]
	at jakarta.faces.component.UIForm.processValidators(UIForm.java:200) ~[myfaces-api-4.1.0.jar:4.1.0]
	at org.omnifaces.component.input.Form.processValidators(Form.java:145) ~[omnifaces-5.0-M2.jar:5.0-M2]
	at jakarta.faces.component.UIComponentBase.processValidators(UIComponentBase.java:1379) ~[myfaces-api-4.1.0.jar:4.1.0]
	at jakarta.faces.component.UIComponentBase.processValidators(UIComponentBase.java:1379) ~[myfaces-api-4.1.0.jar:4.1.0]
	at jakarta.faces.component.UIComponentBase.processValidators(UIComponentBase.java:1379) ~[myfaces-api-4.1.0.jar:4.1.0]
	at jakarta.faces.component.UIViewRoot._processValidatorsDefault(UIViewRoot.java:1760) ~[myfaces-api-4.1.0.jar:4.1.0]
	at jakarta.faces.component.UIViewRoot$ProcessValidatorPhaseProcessor.process(UIViewRoot.java:1868) ~[myfaces-api-4.1.0.jar:4.1.0]
	at jakarta.faces.component.UIViewRoot._process(UIViewRoot.java:1716) ~[myfaces-api-4.1.0.jar:4.1.0]
	at jakarta.faces.component.UIViewRoot.processValidators(UIViewRoot.java:975) ~[myfaces-api-4.1.0.jar:4.1.0]
```

Tried ajax and non-ajax form submission, the result is the same.
And it doesn't work for both old g-recaptcha and new h-captcha.

I'm not sure how exactly it is expected to work, as an assignment of `verifyUrl` only happens on component's render and it's not called on validation.

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

Community

### PrimeFaces version

15.0.0-jakarta

### Theme

_No response_

### JSF implementation

MyFaces

### JSF version

4.1.0

### Java version

21

### Browser(s)

_No response_

melloware:
Fix #13532: Captcha respect defaults

@tandraschko need some advice here.  I want to set these defaults on the component and typically we do that like this...

```java
 public String getTheme() {
        return (String) getStateHelper().eval(PropertyKeys.theme, "auto");
    }
```

I can't do that here because the defaults are dynamic based on the type either `Google` or `HCaptcha`.

I was setting it on `encodeEnd` but it looks like the form submission then sets them back to NULL because they are not set as true defaults on the statehelper.  

**Is there a single spot where a component can be initialized with defaults?**

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/captcha/CaptchaBase.java
+++ b/primefaces/src/main/java/org/primefaces/component/captcha/CaptchaBase.java
@@ -122,7 +122,17 @@ public abstract class CaptchaBase extends UIInput implements Widget {
     }
 
     public String getExecutor() {
-        return (String) getStateHelper().eval(PropertyKeys.executor, null);
+        return (String) getStateHelper().eval(PropertyKeys.executor, () -> {
+            String type = this.getType();
+            switch (type) {
+                case Captcha.RECAPTCHA:
+                    return "grecaptcha";
+                case Captcha.HCAPTCHA:
+                    return "hcaptcha";
+                default:
+                    return null;
+            }
+        });
     }
 
     public void setExecutor(String executor) {
@@ -130,7 +140,17 @@ public abstract class CaptchaBase extends UIInput implements Widget {
     }
 
     public String getSourceUrl() {
-        return (String) getStateHelper().eval(PropertyKeys.sourceUrl, null);
+        return (String) getStateHelper().eval(PropertyKeys.sourceUrl, () -> {
+            String type = this.getType();
+            switch (type) {
+                case Captcha.RECAPTCHA:
+                    return "https://www.google.com/recaptcha/api.js";
+                case Captcha.HCAPTCHA:
+                    return "https://js.hcaptcha.com/1/api.js";
+                default:
+                    return null;
+            }
+        });
     }
 
     public void setSourceUrl(String sourceUrl) {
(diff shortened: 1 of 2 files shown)


tandraschko:
there is no single spot but i would use the supplier in eval and do the IF there

tandraschko:
setting the values just blows up the state, dynamic eval is the better way

melloware:
awesome I think I see what you mean and did it right can you check?  Things seem to be working.

tandraschko:
yep

melloware:
I love this.

PR #13555, opened 2025-03-14
Fix #13521: Tooltip allow multiple events and ESC to close

### LINKED ISSUE #13521: Tooltip: for accessibility provide multiple showEvent as default mouseover, focus and touchstart and ability to close by Esc keypress
### Description

If, in addition to hover, you also want to have the tooltip for the focus, for example, you have to define a second p:tooltip with the identical text for the same target. Disadvantage: the aria-describedby attribute of the target element then already contains two IDs for both target elements and reads out the same tooltip text twice. The same phenomenon occurs if you create a third p:tooltip for showEvent=“touchstart”. The IDs for the aria-describedby attribute of the target element are collected with each additional p:tooltip for the same target element and all texts with identical content are read out repeatedly.

In addition, the tooltip cannot be closed by pressing the Esc key. Both aspects needed for accessibility reasons.

- default showEvent: mouseover focus touchstart (currently only default mouseover event displaying the tooltip)
  -> no need to define multiple p:tooltip for different showEvent having indentical value to be shown readed repeated by screenreader
- dito default hideEvent: mouseout blur touchend
- Esc keypress should close the tooltip for the target element by default

### Additional context

_No response_

melloware:
Fix #13521: Tooltip allow multiple events and ESC to close

### DIFF
--- a/primefaces/src/main/frontend/packages/components/src/tooltip/tooltip.widget.js
+++ b/primefaces/src/main/frontend/packages/components/src/tooltip/tooltip.widget.js
@@ -64,8 +64,8 @@ PrimeFaces.widget.Tooltip = class Tooltip extends PrimeFaces.widget.BaseWidget {
     init(cfg) {
         super.init(cfg);
 
-        this.cfg.showEvent = this.cfg.showEvent ? this.cfg.showEvent + '.tooltip' : 'mouseenter.tooltip';
-        this.cfg.hideEvent = this.cfg.hideEvent ? this.cfg.hideEvent + '.tooltip' : 'mouseleave.tooltip';
+        this.cfg.showEvent = this.cfg.showEvent ? this.cfg.showEvent.split(' ').map(e => e + '.tooltip').join(' ') : 'mouseenter.tooltip';
+        this.cfg.hideEvent = this.cfg.hideEvent ? this.cfg.hideEvent.split(' ').map(e => e + '.tooltip').join(' ') : 'mouseleave.tooltip';
         this.cfg.showEffect = this.cfg.showEffect ? this.cfg.showEffect : 'fade';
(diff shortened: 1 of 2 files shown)


primefaces/src/main/frontend/packages/components/src/tooltip/tooltip.widget.js
```diff
@@ -64,8 +64,8 @@ PrimeFaces.widget.Tooltip = class Tooltip extends PrimeFaces.widget.BaseWidget {
     init(cfg) {
         super.init(cfg);
 
-        this.cfg.showEvent = this.cfg.showEvent ? this.cfg.showEvent + '.tooltip' : 'mouseenter.tooltip';
-        this.cfg.hideEvent = this.cfg.hideEvent ? this.cfg.hideEvent + '.tooltip' : 'mouseleave.tooltip';
+        this.cfg.showEvent = this.cfg.showEvent ? this.cfg.showEvent.split(' ').map(e => e + '.tooltip').join(' ') : 'mouseover.tooltip';
```
tandraschko: why to switch from enter/leave to over/out?

melloware: From this MDN tip: https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseenter_event

>With deep hierarchies, the number of mouseenter events sent can be quite huge and cause significant performance problems. In such cases, it is better to listen for mouseover events.

melloware: They have example showing the difference here: https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseover_event#examples

melloware: actually now looking at that example mouseenter and leave does seem more correct for this tooltip.  Let me revert.

PR #13562, opened 2025-03-14
Fix #13538: 15.0.2 SplitButton aria menu button fixes

### LINKED ISSUE #13538: SplitButton: Insufficient accessibility suitability, operation with screen reader not possible
### Description

During our accessibility assessment of our application the SplitButtons were criticised as a last problem.
Although normal usability via keyboard is completely fine, it is insufficient when screenreaders (JAWS or NVDA)
are used: 
It is possible to open the split button with the screenreader navigation, but it is impossible 
to navigate to the subelement buttons. Also the screenreaders do not properly read the elements and subelements of the split button.

The testers analyzed the showcase:
https://www.primefaces.org/showcase/ui/button/splitButton.xhtml?jfwid=61a87
and gave some details and hints with a screenshot, which I attached.

Image

The attribute „aria-haspopup“ is on the wrong html level. It should be moved to the affected interactive element, in
this case the second, highlighted <button> element in the screenshot, because this has the popup. It's value should not be "true" but "menu" for this situation.

The dropdown button has an improper name "ui-button". For usage with screenreader it should be adjustable, e.g. to contain a text like "more options", or "more actions."

The testers did not give us more recommendations how to finally fix the split button,
instead they pointed to an example how such a split button should be 
implemented to be accessibility compliant:
https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/examples/menu-button-links/

### Additional context

We have a workaround: 
We used standard buttons for the actions as we only grouped 2 or 3 buttons with the split button and we had enough space in our tool bar.

melloware:
Fix #13538: 15.0.2 SplitButton aria menu button fixes

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/splitbutton/SplitButtonRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/splitbutton/SplitButtonRenderer.java
@@ -106,9 +106,6 @@ public class SplitButtonRenderer extends MenuItemAwareRenderer {
         writer.startElement("div", button);
         writer.writeAttribute("id", clientId, "id");
         writer.writeAttribute("class", styleClass, "id");
-        writer.writeAttribute(HTML.ARIA_HASPOPUP, Boolean.toString(hasOverlay), null);
-        writer.writeAttribute(HTML.ARIA_CONTROLS, menuId, null);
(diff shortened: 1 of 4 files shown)


tandraschko:
the code looks good
if your screenreader accepts it, +1

PR #13572, opened 2025-03-16
Fix #13551: Accordion dynamic content load send MVS state

### LINKED ISSUE #13551: AccordionPanel: multi view state not persisted in some cases
### Describe the bug

Based on the showcase implementations, it appears that accordionPanel is not required to be inside a form to operate correctly. However, in practice we noticed that some options do not work unless the accordionPanel is inside a form. Specifically, if both multiViewState=true and multiple=true are set, the panel doesn't always restore its state after a page refresh. Behavior gets more odd from there. If you replicate the exact same accordionPanel and wrap it in a form, and then place this on the same page before the non-wrapped panel, both panels operate correctly even though one is still outside the form.

### Reproducer

Test Case here: https://github.com/koan00/primefaces-test/tree/ft/AccordionStatePersistence

1) Expand multiple columns
2) refresh page with browser controls
3) Actual result: Default state of only panel 0 being open

Discovered on PF14 and confirmed on PF15 via primefaces-test.

### Expected behavior

1) Expand multiple columns
2) refresh page with browser controls
3) Expected result: Same panels are open after refresh

### PrimeFaces edition

Community

### PrimeFaces version

15.0.1 14.0.12

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0

### Java version

17

### Browser(s)

Chrome

melloware:
Fix #13551: Accordion dynamic content load send MVS state

### DIFF
--- a/primefaces/src/main/frontend/packages/components/src/accordion/accordion.widget.js
+++ b/primefaces/src/main/frontend/packages/components/src/accordion/accordion.widget.js
@@ -367,7 +367,8 @@ PrimeFaces.widget.AccordionPanel = class AccordionPanel extends PrimeFaces.widge
                 params: [
                     { name: this.id + '_contentLoad', value: true },
                     { name: this.id + '_currentTab', value: panel.attr('id') },
-                    { name: this.id + '_tabindex', value: parseInt(panel.index() / 2) }
+                    { name: this.id + '_tabindex', value: parseInt(panel.index() / 2) },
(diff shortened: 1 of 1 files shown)


tandraschko:
Why does it fix it? Was it also Broken with parent Form?

tandraschko:
then please add a comment that this is actually commited through a hidden input and a ref to the ticket

melloware:
Comment added to the code

PR #13575, opened 2025-03-16
improve modal dialog scrollability

BalusC:
Consider a dialog with "really long content".
```xml
<p:dialog header="Dialog header" modal="true" width="500">
    <div style="min-height: 1500px">
        Some really long content
    </div>
</p:dialog>
```
By default, this is not scrollable. Only those served by 'Dialog Framework' are scrollable but even then it's with a fixed header which doesn't give nice visual user experience.

Setting CSS `position:absolute` at `.ui-dialog` comes close but it still doesn't yield nice visual user experience. It basically scrolls *together* with the body instead of *over* it. The ultimate goal is to achieve same markup and user experience as seen with Bootstrap modal here at section "[Scrolling long content](https://getbootstrap.com/docs/5.0/components/modal/#scrolling-long-content)". It's possible to achieve this with pure CSS, but the `<p:dialog>` is missing a real "container" element.

This PR makes the necessary adjustments to `DialogRenderer`, `dialog.widget.js` and `_dialog.scss` to achieve this. Basically, it wraps all children of `<div class="ui-dialog>` in a new `<div class="ui-dialog-box>` element which in turn represents the 'real' dialog. The `<div class="ui-dialog>` basically becomes the scrollable overlay. So the Bootstrap modal goal can finally be achieved as follows:

```xml
<p:dialog header="Dialog header" modal="true" blockScroll="true" width="500">
    <div style="min-height: 1500px">
        Some really long content
    </div>
</p:dialog>
```

Unfortunately, this is not fully backwards compatible with possibly existing custom CSS on `.ui-dialog`. So this backwards incompatibility wrt custom CSS needs to be documented in migration guide.

I also took the opportunity to let the `DialogRenderer` throw an `IllegalArgumentException` when `blockScroll="true"` is used without `modal="true"` (as it has no effect then) and also when `blockScroll="true"` is used together with `positionType="absolute"` (as these are essentially conflicting). Also dragging/resizing/reopening of dialog with `positionType="absolute"` has been improved.

This was the test file I used to test all possible combinations. Also dragging and resizing has been tested on each of them.
```xml
<!DOCTYPE html>
<html lang="en"
    xmlns:h="jakarta.faces.html"
    xmlns:f="jakarta.faces.core"
    xmlns:c="jakarta.tags.core"
    xmlns:p="primefaces"
>
    <h:head>
        <title>test</title>
    </h:head>
    <h:body>
        <c:set var="attributeCombinations" value="#{[
            [],
            ['modal'],
            ['modal', 'blockScroll'],
            ['fitViewport'],
            ['modal', 'fitViewport'],
            ['modal', 'blockScroll', 'fitViewport'],
            ['positionType'],
            ['modal', 'positionType'],
            ['positionType', 'fitViewport'],
            ['modal', 'positionType', 'fitViewport']
        ]}" />

        <c:forEach items="#{attributeCombinations}" var="attributes">
            <c:set var="name" value="#{empty attributes ? 'none' : attributes.stream().reduce((l,r) -> l += '_' += r).get()}" />

            <c:if test="#{not attributes.contains('fitViewport')}">
                <h3>small content #{name}</h3>
                <p><p:button value="show" onclick="PF('small_#{name}').show();return false" /></p>
                <p:dialog widgetVar="small_#{name}" header="small content">
                    <c:forEach items="#{attributes}" var="attribute">
                        <f:attribute name="#{attribute}" value="#{attribute eq 'positionType' ? 'absolute' : true}" />
                    </c:forEach>
                    <div>small content #{name}</div>
                </p:dialog>
                <hr/>
            </c:if>
    
            <h3>long content #{name}</h3>
            <p><p:button value="show" onclick="PF('long_#{name}').show();return false" /></p>
            <p:dialog widgetVar="long_#{name}" header="long content">
                <c:forEach items="#{attributes}" var="attribute">
                    <f:attribute name="#{attribute}" value="#{attribute eq 'positionType' ? 'absolute' : true}" />
                </c:forEach>
                <div style="min-height: 1500px">long content #{name}</div>
            </p:dialog>
            <hr/>
        </c:forEach>
    </h:body>
</html>
```


BalusC:
BTW: I wonder if these theme.css really belong in version control as they're autogenerated no?

melloware:
It looks like there are merge conflicts?  

In theory we could not check in the CSS I think we have always done it just to visually differentiate what has changed

BalusC:
Yeah I'll regenerate these other day. Not behind pc atm. Also still need to verify if dialog framework and confirm/message dialogs continue to work fine.

melloware:
@BalusC sorry i made a merge conflict.  Also can you add a note in `\migrationguide\16_0_0.md`

BalusC:
I've further tested/polished all this and added a migration guide entry.

PR #13586, opened 2025-03-18
CI - Java 24 instead of 23

christophs78:
Maybe we have to wait a few days until Github Actions provides Java 24. Let´s see. :-)


christophs78:
Temurin 24 is now available.
But we have to wait for a lombok - update. See https://github.com/projectlombok/lombok/issues/3772 / https://github.com/projectlombok/lombok/issues/3772.

melloware:
I just rebased this with Lombok that supports 24

PR #13603, opened 2025-03-20
Fix #13597: InputNumber use default range for numeric type

### LINKED ISSUE #13597: InputNumber: use by default a range of numbers for numeric type
### Description

Now inputNumber by default (without defined maxValue/minValue attributes) allows input of numbers [in diapason -10000000000000 .. 10000000000000](https://github.com/primefaces/primefaces/blob/65141dcddba6c35972ae1c181f554de211d48b33/primefaces/src/main/java/org/primefaces/component/inputnumber/InputNumberRenderer.java#L51-L52). But in most cases this is incorrect:

- for Long values this diapason is too short and does not allow big Long numbers
- for Integer/Short/Byte values this diapason is too long and allow user to make an error by input non-convertable value.

Such default behaviour makes an easy passibility for beginner/lazy programmer to make an error, that will be obvious only in runtime, when website user want input an exceeding value. In this case website user have no such possibility (for Long values) or get a conversion error (for Integer/Short/Byte values).

Proposal is to make this default behaviour more safe: allow input only values, that could be converted without defining converter explicitly. Type of number value (and hence min/max values) in most cases can be determined on runtime (by value or by EL-expression).

### Additional context

_No response_

melloware:
Fix #13597: InputNumber use default range for numeric type

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/inputnumber/InputNumberRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/inputnumber/InputNumberRenderer.java
@@ -87,7 +87,7 @@ public class InputNumberRenderer extends InputRenderer<InputNumber> {
             else {
                 // Coerce submittedValue to (effective) range of [minValue, maxValue]
                 BigDecimal value = new BigDecimal(submittedValue);
-                submittedValue = coerceValueInRange(value, component).toString();
+                submittedValue = coerceValueInRange(context, value, component).toString();
             }
         }
         catch (NumberFormatException ex) {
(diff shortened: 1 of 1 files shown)


bvfalcon:
Please do not merge yet: this change is required, but not enough.
Methods getEffectiveMinValue/getEffectiveMaxValue are used to coerce a number value, that were input by user. It is necessary also modify (in the same manner) methods getMinimum/getMaximum -- they prevent incorrect input and conversion error.
@melloware thanks a lot! i am surprised, how rapid changes are done. Honestly say, i was going to make today/tomorrow the same changes and commit them.

melloware:
Nice!  @bvfalcon feel free to submit a PR since you are testing.  I will close this one!

Issue #10730, opened 2023-09-28
DataTable: filterMatchMode in does not work correctly with enums


tandraschko:
Currently checking DataTable002 because of #8395, a bit modified but just using a `List<ProgrammingLanguage>` and filtering by type doesnt work.

See:

```
                <p:column field="type" filterMatchMode="exact">
                    <f:facet name="filter">
                        <p:selectCheckboxMenu id="filterType" onchange="PF('datatable').filter()" multiple="true" updateLabel="true">
                            <f:selectItems value="#{dataTable002.types}"/>
                        </p:selectCheckboxMenu>
                    </f:facet>
                </p:column>
```

@Rapster WDYT?

melloware:
we use it in our apps let me get an example.

melloware:
```xml
<p:column id="colTemplateKey" headerText="#{appmsg['column.emailtemplate.templateKey']}"
						sortBy="#{row.emailTemplateKey.text}" filterBy="#{row.templateKey.text}" filterMatchMode="in"
						styleClass="col-left" width="180" sortPriority="1" sortOrder="asc">
	<f:facet name="filter">
		<p:selectCheckboxMenu id="chkTemplateKeyFilter" label="#{appmsg['label.all']}" filter="true"
								filterMatchMode="startsWith" updateLabel="true">
			<p:ajax event="change" process="@this" oncomplete="PF('tableEntity').filter();" />
			<p:ajax event="toggleSelect" process="@this" oncomplete="PF('tableEntity').filter();" />
			<f:selectItems value="#{emailTemplateKeyFilterOptions}" />
		</p:selectCheckboxMenu>
	</f:facet>
	<h:outputText value="#{row.emailTemplateKey.displayText}" />
</p:column>
```

tandraschko:
yep, i was debugging currently
seems like filterMatchMode must be "in", so your test is wrong

tandraschko:
still the same problem
InFilterConstraint receives the string value as value but and the filter array contains the value as enum

FlipWarthog:
> still the same problem InFilterConstraint receives the value as enum but and the filter array contains the value as string
 
@melloware I think we ran into this and in what you posted it's being handled by `filterBy="#{row.templateKey.text}"`. Our Enums have a `getText()` method:
```java
   public String getText() {
      return this.name();
   }
```
And the `SelectItem`s value is also the enum's `getText()`:
```java
[...]

PR #13606, opened 2025-03-20
Refactor InFilterConstraint to leverage EqualsFilterConstraint

FlipWarthog:
There have been a few issues identified and fixed between InFilterConstraint and EqualsFilterConstraint (e.g., #12666 , #10730, #8106), but they have been fixed separately/inconsistently. Since `in` is effectively `equals` over a collection, I thought it might make sense to leverage one for the other, so any future fixes/enhancements to one always applies to the other. The initial driver for this was to apply the fix for #12666 to `filterMatchMode="in"`.

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datatable/DataTable048.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datatable/DataTable048.java
@@ -25,32 +25,43 @@ package org.primefaces.integrationtests.datatable;
 
 import java.io.Serializable;
 import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.List;
 
-import lombok.AllArgsConstructor;
+import jakarta.annotation.PostConstruct;
+import jakarta.faces.model.SelectItem;
+import jakarta.faces.view.ViewScoped;
+import jakarta.inject.Inject;
+import jakarta.inject.Named;
+
 import lombok.Data;
-import lombok.NoArgsConstructor;
 
+@Named
+@ViewScoped
 @Data
-@NoArgsConstructor
-@AllArgsConstructor
-public class ProgrammingLanguage implements Serializable {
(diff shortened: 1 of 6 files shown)


FlipWarthog:
@melloware I added some integration tests for this and it works but #8106 is confusing me.. `null` select item value for IN means "filter to only records where the value is null" but for EQUALS it means ALL. I guess that has been the case for a while and it's still true here, but can you confirm that is correct?

melloware:
Yeah from that it looks like they want to filter the `value` of NULL?

FlipWarthog:
Right, that's how I read it and that's how it works. So the tests I added confirm it and show the difference between IN and EQUALS with how they handle `null`. I'm going to switch this out of draft. Also if @tandraschko and @christophs78 want to give it a once over, I'd appreciate it. One question I had was on `InFilterConstraint extends EqualsFilterConstraint` and if that's okay or if I should refactor out a static comparison method and just use that in both Contsraints, which is a pattern in other ones.

christophs78:
> Right, that's how I read it and that's how it works. So the tests I added confirm it and show the difference between IN and EQUALS with how they handle `null`. I'm going to switch this out of draft. Also if @tandraschko and @christophs78 want to give it a once over, I'd appreciate it. One question I had was on `InFilterConstraint extends EqualsFilterConstraint` and if that's okay or if I should refactor out a static comparison method and just use that in both Contsraints, which is a pattern in other ones.

In my opinion this is a good PR. It improves our codebase and also improves integration-test-coverage.
DataTable026Test (and a few more) already provide a good base-coverage so i hope we don´t break things.
Extending and adding integration-test´s IMO realy helps to stabilze things. DataTable and all it´s useage-szenarios are way to complicated to overlook them and think about all possible regressions of a change.
Regard extending vs static comparison method i don´t have a strong opionon. I´m perfectly fine with both ways you suggested.

melloware (approved):

melloware:
Yes thank you for the integration tests as well!

PR #13617, opened 2025-03-24
Fix spelling

FlipWarthog:
I noticed `SelectCheckboxMenu#togglPanel()` in PrimeFaces Selenium had a typo, so I fixed it and did a once-over of most/all of that submodule. I also saw that PrimeFaces uses "toggleable" consistently except in one place where "togglable" is used, so I updated it, despite both being acceptable spellings. Since both of these changes affect end-users I updated the v16 What's New.

### DIFF
--- a/docs/16_0_0/gettingstarted/whatsnew.md
+++ b/docs/16_0_0/gettingstarted/whatsnew.md
@@ -2,88 +2,16 @@
 
 This page contains a list of big features. Please check the GitHub issues for all changes.
 
-## 15.0.0
+## 16.0.0
 
 ### PrimeFaces
 
-Look into [migration guide](https://primefaces.github.io/primefaces/15_0_0/#/../migrationguide/15_0_0) for more enhancements and changes.
-
-* Core
-    * Added `widgetPreConstruct` attribute to support custom logic before widget creation. See [widgets](../core/widgets.md) for more information.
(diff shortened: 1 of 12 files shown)


docs/16_0_0/gettingstarted/whatsnew.md
```diff
@@ -2,88 +2,16 @@
...
-    * Added `toggleFilter()` JS Widget function to show/hide filter components (see showcase example for usage).
-
-* DatePicker
-    * Added `defaultHour`, `defaultMinute`, `defaultSecond`, `defaultMillisec` attributes to match legacy `Calendar` component
-    * Added ability to pick weeks by `view="week"`
-    * Added `showLongMonthNames` attribute to display long month names instead of short names in month picker and month navigator
-
-* FeedReader
-    * Added `podcast="true"` property if [Apple Itunes Podcast](https://help.apple.com/itc/podcasts_connect/#/itcb54353390) parsing and specific tags 
+    * Rename `ui-filter-togglable` to `ui-filter-toggleable`
```
tandraschko: i dont think it should go in "whats new", better put it in migration guide

FlipWarthog: Ah you're right. Moved them both.

PR #13629, opened 2025-03-26
Fix #13620: DataTable: Global select checkbox gets automatically selected when all visible table rows on the current page are selected

### LINKED ISSUE #13620: DataTable: Global select checkbox gets automatically selected when all visible table rows on the current page are selected
### Describe the bug

The Global select checkbox should only get activated when *all* rows are selected.

Yet in this scenario it is sufficient to select all *visible* rows when using a paged data table..

### Reproducer

This can easily be reproduced on the current showcase at https://www.primefaces.org/showcase/ui/data/datatable/selection.xhtml under Checkbox:

1. Individually select all visible rows in the table
2. The Global select checkbox gets activated
3. When switching to the second page, the rows shown are not selected, but the global select checkbox is.
 

### Expected behavior

The Global select checkbox should only get activated when *all* rows are selected.

### PrimeFaces edition

Community

### PrimeFaces version

15.0.2

### Theme

_No response_

### JSF implementation

All

### JSF version

2.3

### Java version

21

### Browser(s)

_No response_

FlipWarthog:
@melloware Quick option: just disallow it for `selectionPageOnly="false"`. This works and the other issue referenced is still resolved. But I think the more comprehensive solution would be to refactor selection to pass the selectable row keys back and forth, along with the selection itself, so the JS can better check/do things in the non-page only cases...

Fix #13620

### DIFF
--- a/primefaces/src/main/frontend/packages/components/src/datatable/datatable.widget.js
+++ b/primefaces/src/main/frontend/packages/components/src/datatable/datatable.widget.js
@@ -5146,10 +5146,12 @@ PrimeFaces.widget.DataTable = class DataTable extends PrimeFaces.widget.Deferred
             }
 
             var totalEnabled = enabledCheckboxes.length;
-            if(totalEnabled && totalEnabled === selectedCheckboxes.length)
-               this.checkHeaderCheckbox();
-            else
-               this.uncheckHeaderCheckbox();
+            if (this.cfg.selectionPageOnly) {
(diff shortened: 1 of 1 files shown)


melloware (approved):

melloware:
I like this fix. In non selectionPageOnly scenarios it will be difficult to determine if all pages are checked

melloware:
You can open another PR off the 15.x branch or I can cherry pick it tomorrow.

PR #13679, opened 2025-04-05
Fix #13678: 15.0.4 Exporter fix for LazyModel

### LINKED ISSUE #13678: DataExporter: data is not exported (Excel format) if DataTable is lazy
### Describe the bug

From Primefaces version 14 (and 15) DataExporter cannot export data in Excel format. It only exports header.
Exporter's target is a lazy loaded DataTable.

### Reproducer

Reproducer: https://github.com/dmainardi/primefaces-test

Try to export in XLS or XLSX, it does not work (blank results, only header).
Page only XLS or XLSX exports do work.

In 'ProductLazyDataModel' file I used the _To avoid doing a separate count-statement against your datasource you may implement it like this_ trick: row count is set in *load* method rather than in the *count* one.
Doing some debugging I noticed that before the export the *count* method is called and, since the result is ZERO, no data is exported.

In Primefaces 13 (and before) the exporting worked without problems.

### Expected behavior

If you lower the PrimeFaces version to 13.0.10 it works, data can be seen.

### PrimeFaces edition

Community

### PrimeFaces version

14.0.0 and 15.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3 and 4.0

### Java version

21

### Browser(s)

Firefox

melloware:
Fix #13678: 15.0.4 Exporter fix for LazyModel

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
+++ b/primefaces/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
@@ -102,7 +102,7 @@ public abstract class DataTableExporter<P, O extends ExporterOptions> extends Ta
             LazyDataModel<Object> lazyDataModel = (LazyDataModel<Object>) table.getValue();
             Integer bufferSize = exportConfiguration.getBufferSize();
             boolean bufferized = bufferSize != null;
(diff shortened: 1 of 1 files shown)


tandraschko:
Hmm
Thinking about it
How was it before all the refactorings?

We wanna count all rows actually, so the current code was correct in general
Maybe your old Change was more correct

PR #13680, opened 2025-04-05
Fix #13678: Exporter fix for LazyModel

### LINKED ISSUE #13678: DataExporter: data is not exported (Excel format) if DataTable is lazy

melloware:
Fix #13678: Exporter fix for LazyModel

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
+++ b/primefaces/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
@@ -102,7 +102,7 @@ public abstract class DataTableExporter<P, O extends ExporterOptions> extends Ta
             LazyDataModel<Object> lazyDataModel = (LazyDataModel<Object>) table.getValue();
             Integer bufferSize = exportConfiguration.getBufferSize();
             boolean bufferized = bufferSize != null;
-            int batchSize = Objects.requireNonNullElseGet(bufferSize, () -> table.getRows());
+            int batchSize = Objects.requireNonNullElseGet(bufferSize, () -> lazyDataModel.getRowCount());
 
             if (batchSize > 0) {
                 List<?> wrappedData = lazyDataModel.getWrappedData();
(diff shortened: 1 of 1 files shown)


tandraschko:
i dont think that is a very good fix

we just get the current, almost random, rowCount to calculate the batchSize

either we pass the dataTable pageSize

or

either we call we leave it as it as and document it: if you use exporter, without batchSize, you need to implement getRowCount

melloware:
OK sounds like a documentation issue?

tandraschko:
dont know
IMO pageSize is a ok fix

melloware:
hmmm i am not sure what you mean sorry for the confusion!

tandraschko:
int batchSize = Objects.requireNonNullElseGet(bufferSize, () -> dataTable.getRows());

melloware:
OK updated

melloware:
OK i think my original fix is the right fix.

PR #13691, opened 2025-04-09
Fix #13689: @Positive use min value = 0

### LINKED ISSUE #13689: InputNumber: Cannot clear if value is marked @Positive
### Describe the bug

If a number is validated with `jakarta.validation.constraints.Positive`, you can't clear InputNumber.

### Reproducer

In `org.primefaces.showcase.view.input.InputNumberView` add the constraint @Positive to input2
```
    @Positive
    private Double input2 = Double.valueOf(0);
```
Go to /showcase/ui/input/inputNumber.xhtml
- Change the value of the field "Currency". For example, 3
- hit tab
- go back to the field "Currency"
- Try to clear the field

### Expected behavior

The constraint `@Positive` says that "null elements are considered valid."
As a consequence one should be able to clear InputNumber

### PrimeFaces edition

None

### PrimeFaces version

14.0.4

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

17

### Browser(s)

_No response_

melloware:
Fix #13689: @Positive use min value = 0

### DIFF
--- a/primefaces/src/main/java/org/primefaces/validate/bean/PositiveClientValidationConstraint.java
+++ b/primefaces/src/main/java/org/primefaces/validate/bean/PositiveClientValidationConstraint.java
@@ -33,7 +33,7 @@ public class PositiveClientValidationConstraint extends AbstractClientValidation
     public static final String MESSAGE_METADATA = "data-p-positive-msg";
     public static final String CONSTRAINT_CLASS_NAME = String.format("%s.%s", jakarta.validation.constraints.NotNull.class.getPackage().getName(), CONSTRAINT_ID);
     public static final String MESSAGE_ID =  String.format("{%s.%s}", CONSTRAINT_CLASS_NAME, ATTR_MESSAGE);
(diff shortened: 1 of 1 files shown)


tandraschko:
you know that this will break the metadata transform?
either we remove it for inputnumber or the transformer must set 0000000001

melloware:
not sure what you mean the MetaDataTransform uses the constant?

```java
if (annotationClassName.equals(PositiveClientValidationConstraint.CONSTRAINT_ID)) {
                    inputNumber.setMinValue(PositiveClientValidationConstraint.MIN_VALUE);
                }
``` 
?

melloware:
I am confused minValue=0 is what `@PositiveOrZero` does?

tandraschko:
oh, or i think the other way around :D

if it works its ok fo rme

melloware:
I actually think I will leave it like it is we have a battery of Integration Tests and documentation for this.

melloware:
OK let me update the Integration tests

PR #13702, opened 2025-04-12
Switch to Faces annotations

melloware:
Since we are Faces 4.0 only now we don't have to worry about backward compatibility and this thins out the faces-config.xml nicely.  I am not sure if what's left in the faces-config.xml can be replaced with annotations?

- [x] Jetty / with and without org.apache.myfacs.annotation.USE_CDI_FOR_ANNOTATION_SCANNING
- [x] Quarkus
- [x] Payara
- [x] Wildfly
- [x] TomEE / with and without org.apache.myfacs.annotation.USE_CDI_FOR_ANNOTATION_SCANNING
- [ ] OpenLiberty

we might still need a beans.xml for scanning and USE_CDI_FOR_ANNOTATION_SCANNING


lprimak:
Awesome! Modernization for the win!

fcorneli:
The big cleanup, nice!

tandraschko:
As i already said to you, IMO the effort is it not worth and some containers might break.
But you already did it 😄
This will also break propably the myfaces feature mentioned here: https://github.com/jakartaee/faces/issues/1584


So we have to try all containers and create issues there if not working.

melloware:
Ok I didn't switch any of the converters I left them all in faces-config.xml. 

But I agree we should test this in more than just our integration tests. Payara, Wildfly, tomee etc

tandraschko:
Openliberty and quarkus might be important

melloware:
I tested Payara and Quarkus and both are working fine.  I added a checklist up top for other servers if anyone wants to help.

tandraschko:
When you use TomEE or MF on jetty, please also try o.a.m.annotation.USE_CDI_FOR_ANNOTATION_SCANNING
We might need a beans.xml

fcorneli:
About including a `beans.xml` within the PF JAR. Be careful here with regards to multiple WAR within EAR deployments. The reason why I split Enterprise JSF into an `ejsf-taglib` JAR and a CDI-enabled (i.e., with `beans.xml`) `ejsf-api` JAR is to make it work under such deployments. So `ejsf-taglib` can occur within multiple WARs, but `ejsf-api` can be included only once at the EAR `lib` level. Putting `beans.xml` within the PF JAR would limit deployment to single WAR EARs.

fcorneli:
@melloware For OpenLiberty testing, check out:
https://github.com/e-Contract/enterprise-jsf/blob/master/ejsf-demo-ee10/src/main/liberty/config/server.xml

melloware:
I thought for faces 4.0 Java ee 10 that CDI was already required by all containers?

lprimak:
> Putting beans.xml within the PF JAR would limit deployment to single WAR EARs.

This is not per the Spec. There is a long-standing bug in Payara and GlassFish that prevent this from happening.
This works in WildFly and I believe OpenLiberty as well.
In Payara, I am working on https://github.com/payara/Payara/pull/7212 to fix this issue, and it's close to completion

tandraschko:
TBH i dont care about that crazy bugs in containers anymore. Payara didnt upgrade a Mojarra since over 1 years: [657](https://github.com/payara/Payara/issues/6578)
for much such projects just feels dead.

there is nothing per spec that prevents us adding a beans.xml

melloware:
Looks like Wildfly works as well.

melloware:
I just ran all the integration tests with 

```xml
    </context-param>
	<context-param>
        <param-name>org.apache.myfacs.annotation.USE_CDI_FOR_ANNOTATION_SCANNING</param-name>
    </context-param>
```

and all the integration tests passed!

tandraschko:
and that even we dont have a beans.xml in PF.jar, right? weird

lagar84:
Is there a typo in 
`<param-name>org.apache.myfacs.annotation.USE_CDI_FOR_ANNOTATION_SCANNING</param-name>`?
should it be `myfaces` instead of `myfacs`?

tandraschko:
yup

tandraschko:
https://github.com/apache/myfaces/blob/15cda2def00f6ef8cefa53e2d669616b0ff45419/impl/src/main/java/org/apache/myfaces/config/webparameters/MyfacesConfig.java#L371

melloware:
HA trying again...

tandraschko:
yep

and then please retest quarkus

melloware:
Quarkus works it just prints out this message

```
2025-04-17 15:35:14,793 WARN  [io.qua.arc.dep.BeanArchiveProcessor] (build-30) Detected bean archive with bean discovery mode of 'all', this is not portable in CDI Lite and is treated as 'annotated' in Quarkus! Path to beans.xml: org.primefaces:primefaces:16.0.0-SNAPSHOT:/META-INF/beans.xml
```

Which I will fix in the PF Extension just like I did for OmniFaces:
https://github.com/quarkiverse/quarkus-omnifaces/blob/1cd2d12609525c10fdeb27d907e77163aceabbde/deployment/src/main/java/io/quarkiverse/omnifaces/deployment/OmnifacesProcessor.java#L148-L152

melloware:
@BalusC i am looking at OmniFaces and your `beans.xml` has `bean-discovery-mode="annotated"` but according to CDI spec that is only supposed to scan for CDI beans but `@FacesComponent` are not CDI beans?

When I try this with MyFaces CDI only it won't find this components in `bean-discovery-mode="annotated"` i had to use ``bean-discovery-mode="all"` which is not recommended?  I am SO confused.

```
</context-param>
	<context-param>
        <param-name>org.apache.myfaces.annotation.USE_CDI_FOR_ANNOTATION_SCANNING</param-name>
    </context-param>

**Comment by mkomko** [comment]
@melloware It was a whole thing for us as well: https://github.com/omnifaces/omnifaces/issues/780

File: primefaces-integration-tests/src/main/webapp/WEB-INF/web.xml [hunk 1]
```diff
@@ -50,6 +50,10 @@
         <param-name>primefaces.CLIENT_SIDE_VALIDATION</param-name>
         <param-value>true</param-value>
     </context-param>
+	<context-param>
```
mkomko: Incorrect indentation? 😉

melloware: Fixed

primefaces/src/main/java/org/primefaces/component/dataview/DataViewGridItem.java
```diff
@@ -23,6 +23,8 @@
  */
 package org.primefaces.component.dataview;
 
+import jakarta.faces.component.FacesComponent;
+@FacesComponent(value = DataViewGridItem.COMPONENT_TYPE, namespace = DataViewGridItem.COMPONENT_FAMILY)
```
FlipWarthog, line 28: Looks like there's some formatting quirks in some of these files that didn't already have any imports.

melloware: yeah some space look missing!

tandraschko:
so, whats left for this PR - just testing?

it would be great if we can close this soon, dont like open PRs :D

melloware:
I think we should merge this I have tested everything I can except OpenLiberty but I am fairly confident it will work based on all the other testing.

PR #13805, opened 2025-05-16
Fix #13804: 15.0.5 Lithuanian translation

### LINKED ISSUE #13804: Lithuanian texts in locale-lt.js are broken
Encoding of texts in `locale-lt.js` broke somehow. For example, `monthNames` were incorrectly changed from:
`["Sausis", "Vasaris", "Kovas", "Balandis", "Gegužė", "Birželis", "Liepa", "Rugpjūtis", "Rugsėjis", "Spalis", "Lapkritis", "Gruodis"]`
to:
`["Sausis", "Vasaris", "Kovas", "Balandis", "Geguéƒá?", "Biréƒelis", "Liepa", "Rugpjé¨tis", "Rugsá?jis", "Spalis", "Lapkritis", "Gruodis"]`

_Originally posted by @gino0631 in https://github.com/primefaces/primefaces/issues/12860#issuecomment-2887155106_

melloware:
Fix #13804  15.0.5 Lithuanian translation


primefaces/src/main/resources/META-INF/resources/primefaces/locales/locale-lt.js
```diff
@@ -2,129 +2,192 @@ if (window.PrimeFaces) {
...
     "dateIs": "Data yra",
-    "dateIsNot": "Data ná?ra",
-    "dayNames": ["Sekmadienis", "Pirmadienis", "Antradienis", "Treá?iadienis", "Ketvirtadienis", "Penktadienis", "é eéÀtadienis"],
-    "dayNamesMin": ["S", "P", "A", "T", "K", "Pn", "é "],
-    "dayNamesShort": ["S", "P", "A", "T", "K", "Pn", "é "],
-    "emptyFilterMessage": "Rezultaté° nerasta",
-    "emptyMessage": "Ná?ra pasirinkimo",
-    "emptySearchMessage": "áÛraéÀé° nerasta",
+    "dateIsNot": "Data nėra",
+    "dayNames": [
```
gino0631: Maybe it should be in a single line?

melloware:
It doesn't have to be it's just a JS array.  It gets parsed the same

gino0631:
Technically, yes, but I wasn't sure regarding the coding style. It was just to draw an attention in case such formatting wasn't intentional.

melloware:
Can you check the true and false labels?  Do they look translated correctly?

gino0631:
They are not translated; same with "weak": "Weak".

melloware:
Yep can you translate and I can get PrimeLocale updated as well?

melloware:
I took a shot with Google Translate

gino0631:
Translations depend on context - do you have any information where these labels are used?

melloware:
Weak, Medium, Strong are to indicate Password strength as you are typing a password

gino0631:
Thank you, these look good.
I also reviewed the entire file. It looks good, maybe just a few small things:
1. "dateFormat": "yyyy-mm-dd" (if I understand the purpose of the string correctly - i.e. dates are usually written like 2025-05-18)
2. false/true, depending on context, may sound more natural translated as no/yes - do you have any information where these are used?
3. "move(All)ToSource/Target" are not translated - again, would be better to suggest knowing the use case.

melloware:
Let me found out where true/false is used but those move to target are used in the PickList component buttons if you go to the showcase for PickList!

The date format you have correct it's not actually used in PrimeFaces but is used in PrimeReact/NG/Vue

melloware:
OK trueLabel and falseLabel are used here: https://primereact.org/tristatecheckbox/#accessibility

So PF should probably use them as well if PrimeReact is using them for TriStateCheckbox.

Picklist: https://www.primefaces.org/showcase/ui/data/pickList.xhtml

has the Move To Target buttons etc.

gino0631:
Thank you, I have checked the links.

In general, the translations look acceptable. To improve,

move* could be shorter (without "element"), and slightly different:
"moveAllToSource": "Perkelti visus į pradinį sąrašą",
"moveAllToTarget": "Perkelti visus į tikslinį sąrašą",
"moveToSource": "Perkelti į pradinį sąrašą",
"moveToTarget": "Perkelti į tikslinį sąrašą",

true/false in this case could look more naturally if translated as yes/no, as we are not referring to Boolean logic here:
"falseLabel": "Ne",
"trueLabel": "Taip",

melloware:
thanks i will fix PrimeLocale too.

gino0631:
@melloware good job - you fixed the reported issue, and even made some improvements, thanks!

melloware:
Thanks for the assist it will be in 15.0.5 which should be out in the next week or two.

PR #13806, opened 2025-05-17
Fix #13802: Schedule ignoreAutoUpdate on core fetch events

### LINKED ISSUE #13802: Schedule: set ignoreAutoUpdate for the load event
Currently the schedule loads all the stuff via ajax and also updates e.g. growl components

The problem is that there is no "load" behavior where i could add:

```
<p:ajax event="load" ignoreAutoUpdate="true" />
```

melloware:
Fix #13802: Schedule ignoreAutoUpdate on core fetch events


tandraschko:
What about ln 151?

tandraschko:
Not Sure but yes, lets leave it for now

PR #13822, opened 2025-05-19
Add 'Unsupported features' to JPALazyDataModel docs

FlipWarthog:
A coworker got spun around the axel trying to implement `filterFunction` for our `JPALazyDataModel` impl. On the surface (not understanding the details of how `JPALazyDataModel` is implemented) it was a valid assumption that the feature should work and how it acts when it does not is very hard to trace. Given that, and I think it's something that actually can't be implemented, I opted to add a new section to the docs to begin listing features that are not implemented so future devs don't run into the same situation.

Ex. how `filterFunction` "fails silently"
[jpalazydatamodel-filterfunction.zip](https://github.com/user-attachments/files/20302274/jpalazydatamodel-filterfunction.zip)

cc @melloware @tandraschko

### DIFF
--- a/docs/15_0_0/components/datatable.md
+++ b/docs/15_0_0/components/datatable.md
@@ -1083,6 +1083,10 @@ Map<String, FilterMeta> filterMeta = Collections.emptyMap();
 List<MyEntity> result = lazyDataModel.load(0, lazyDataModel.count(filterMeta), sortMeta, filterMeta);
 ```
 
+#### Unsupported features
+
+Due to its implementation based on Criteria API, certain features of `DataTable` simply cannot work with `JPALazyDataModel` but may "fail silently". To aid implementors in avoiding running into issues, we've compiled (and will continue to add to) a list of features in this category:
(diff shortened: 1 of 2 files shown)


melloware:
@FlipWarthog can you add this same blurb to the 16.0.0 MD as well

FlipWarthog:
> @FlipWarthog can you add this same blurb to the 16.0.0 MD as well

Done, forgot the docs were versioned

melloware (approved):

melloware:
@tandraschko  is on holiday for 2 weeks but this change looks good to me for the docs.  I am sure we will come up with others as we go too.

lprimak:
Can he try https://docs.flowlogix.com/#section-jpa-lazymodel

Not sure what exactly was missing but maybe that will work better. 
Or something we both can learn from

melloware:
@lprimak its the same issue from looking at your docs you ignore `filterFunction` property on the DataTable.

FlipWarthog:
@lprimak Thanks for the suggestion. We did already solve it with how we hook our custom filtering API into the table, but it took him a couple hours of spinning before he realized filterFunction wasn't the way to do it and asked for help (and I even had to double check and make sure it wasn't our custom code not implementing it)

PR #13840, opened 2025-05-24
Fix #13836: Release workflow and Sonatype publisher upgrade

### LINKED ISSUE #13836: Build: Move to Sonatype Central Publisher
Greetings OSSRH Publisher,

As you may have heard, OSSRH is reaching end of life on June 30, 2025. OSSRH users need to migrate their namespaces to the Central Portal as soon as possible.

Instructions for self migration are located here: https://central.sonatype.org/faq/what-is-different-between-central-portal-and-legacy-ossrh/#self-service-migration

To make the transition smoother we will be automatically migrating publishers that have not used oss.sonatype.org or s01.sonatype.org to publish artifacts in some time starting with the oldest and working our way forward. To avoid disruption to your publishing processes we strongly encourage migrating before the June 30, 2025 deadline.

Thank you for your assistance,

The Central Team

@tandraschko  i will take care of this

melloware:
Fix #13836: Release workflow and Sonatype publisher upgrade

image


melloware:
Yep I will DM you I need you to add 5 secrets for the GpG keys and sonatype user and token. 

After that it all just will magically bump the version, cut the release and bump it back to SNAPSHOT

primefaces/pom.xml
```diff
@@ -568,7 +568,7 @@
                     <createDependencyReducedPom>false</createDependencyReducedPom>
                     <promoteTransitiveDependencies>true</promoteTransitiveDependencies>
                     <minimizeJar>true</minimizeJar>
-                    <createSourcesJar>true</createSourcesJar>
```
tandraschko, line 571: why false?

pom.xml
```diff
@@ -109,11 +98,64 @@
 
     <build>
         <directory>${build.output}</directory>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
```
tandraschko, line 104: why all this reconfiguration?
i would like to leave the PR as small as possible

melloware: 4 of those are required by the JReleaser Plugin see https://jreleaser.org/guide/latest/examples/maven/staging-artifacts.html as recommended best practice.

primefaces/pom.xml
melloware: this break the new maven central publisher or Jreleaser I forget which but it states "source jar is created twice" error

melloware:
I have to fix some merge conflicts

melloware:
OK ready to go

PR #13842, opened 2025-05-24
#13841 SonarQube: add PR-decoration

### LINKED ISSUE #13841: SonarQube: add PR-decoration
### Description

https://docs.sonarsource.com/sonarqube-cloud/improving/pull-request-analysis/

We do this on my day-job in the combination of Bitbucket + Jenkins + SonarQube Server.
In theory it should be easier with GitHub + SonarQube Cloud.

### Additional context

_No response_

christophs78:


christophs78:
@Rapster  wrote following last summer on our primefaces-team-channel:

> "The issue I faced might be because the token is a "repository secret" instead of being an "environnment secret". I would like to try that... Can I request to whoever has access to the settings to set the SONAR_TOKEN as a environnement var please?"

melloware:
There are also Organization Secrets also.  Not sure the difference in the 3!

image

melloware:
@AndreasIgel i fixed the merge conflict and new build is running now

AndreasIgel:
> @AndreasIgel i fixed the merge conflict and new build is running now

I might see the issue. But I wout try it in a fork first! Is there any rule in this repository to not put sonar properties into maven-properties in pom.xml?

melloware:
we have typically avoided the Sonar config in Maven because its a GitHub only feature other users would never need.

AndreasIgel:
This could not work in that way, because the organziation is wrong in the maven-command. It is org.primefaces and not primefaces. Furthermore I would change the following:
- not putting the token in command line (sonarqube plugin is taking it from environment directly)
- not putting the token into a documentation part, not sure if github hiding it correctly even there or exposing it in some condition by that
- adding sonar to plugin-management to fix up to a concrete version
- adding specific sonar-properties in pom.xml and not just in the execution command
- excluding dependabot because it does not have access to secrets

I tested the changes with my sonarcloud account and a fork: https://github.com/java-helpers/primefaces-fork/pull/1, the resulting pull-request with these updates is https://github.com/primefaces/primefaces/pull/14619. Maybe that already solves the problems here or primetek really needs to change something in their sonarcloud account.

AndreasIgel:
You were right about the organization key. As you previously mentioned, the issue seems to be that the SonarCloud token is either expired or lacks permissions for pull request workflows.

tandraschko:
it seems that currently my personal account token is used

it seems that OS plan does not allow custom project plans

i cant even define any permissions for my personal token

tandraschko:
will close it in favor of the other PR

PR #13856, opened 2025-05-27
Fix #13846: DatePicker improve accessibility of date cells

### LINKED ISSUE #13846: DatePicker: Accessibility issues
### Describe the bug

axe-core raises accessiblity issues for the DatePicker component.

**Issue 1**
Image

**Issue 2**
Image

### Reproducer

1. Run [primefaces-test-issue13846.zip](https://github.com/user-attachments/files/20443887/primefaces-test-issue13846.zip)
2. Click into the datepicker field
3. Issues arise mentioned in the "Describe the bug" section

### Expected behavior

Issue 1: "aria-selected" should be used in a correct manner. "aria-selected" is not permitted on a link element.
A solution could be to set role="gridcell" on the `<td>`, and set the "aria-selected" on the `<td>` instead of the `<a>`.
An other solution is to change the `<a>` to `<button>` and use "aria-pressed".

Issue 2: The cell is only disabled styling wise, it should also set an ARIA attribute like "aria-disabled". E.g. Primevue does that -> https://primevue.org/datepicker/

### PrimeFaces edition

Community

### PrimeFaces version

15.0.3

### Theme

_No response_

### JSF implementation

All

### JSF version

4.0

### Java version

11

### Browser(s)

_No response_

melloware:
Fix #13846: DatePicker improve accessibility of date cells

@pfumi now i remove aria from `<a>` and moved to `<td>`. Added `role="gridcell"` and only render `aria-selected` and `aria-disabled` if they are `true`

### DIFF
--- a/primefaces/src/main/frontend/packages/datepicker/datepicker/src/jquery.ui.prime.datepicker.cjs
+++ b/primefaces/src/main/frontend/packages/datepicker/datepicker/src/jquery.ui.prime.datepicker.cjs
@@ -1415,7 +1415,7 @@ $.widget("prime.datePicker", {
             'ui-state-disabled': disabled
         });
 
-        return '<a tabindex="0" class="ui-monthpicker-month' + monthClass + '">' + content + '</a>';
+        return '<a tabindex="0" class="ui-monthpicker-month' + monthClass + '" aria-disabled="' + disabled + '">' + content + '</a>';
     },
 
     renderMonthViewMonths: function() {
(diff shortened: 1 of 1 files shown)


primefaces/src/main/frontend/packages/datepicker/datepicker/src/jquery.ui.prime.datepicker.cjs
```diff
@@ -1625,7 +1625,7 @@ $.widget("prime.datePicker", {
                 lastDate = weekDates[6],
                 cellClass = firstDate.otherMonth && lastDate.otherMonth && !this.options.showOtherMonths ? ' ui-datepicker-other-month-hidden' : '';
 
-            weekHtml += '<td class="ui-datepicker-weeknumber' + cellClass + '"><span class="ui-state-disabled">' +
+            weekHtml += '<td class="ui-datepicker-weeknumber' + cellClass + '"><span class="ui-state-disabled" aria-disabled="true">' +
```
pfumi: Probably on the span the aria-disabled is not necessary, because it has no meaning to the screen reader

melloware:
@pfumi did you forget to push Finish Review?  I don't see any comments?

PR #13867, opened 2025-05-29
Fix #13854: 15.0.5 Table render with ARIA role="group"

### LINKED ISSUE #13854: AutoComplete: Accessibility issues with `<p:column>` custom rendering
### Describe the bug

Issue 2&3 from #13849

Image
Image

### Reproducer

1. Run [primefaces-test.zip](https://github.com/user-attachments/files/20444283/primefaces-test.zip)
2. Click into the field "autocomplete tabular" and type "val1"
3. Problems described in "Describe the bug" arise

### Expected behavior

The problem is the `<tbody>` tag. Elements with `role="option"` have to be direct children of an element with `role="listbox"`.

### PrimeFaces edition

Community

### PrimeFaces version

15.0.3

### Theme

_No response_

### JSF implementation

All

### JSF version

4.0

### Java version

11

### Browser(s)

_No response_

melloware:
Fix #13854: 15.0.5 Table render with ARIA role="group"


pfumi:
The only change which is to be made is, to apply role group for tabular lists on the body, everything else role listbox and aria-multiselectable should be left untouched!

melloware:
Yep please re-review

pfumi (commented):
Looks good

PR #13870, opened 2025-05-29
Refactor: precompile frequently used regex patterns for better performance

kamilkrzywanski:
While working on the codebase, I noticed several regex patterns being compiled repeatedly at runtime. I identified most of these occurrences and refactored the code to use precompiled Pattern instances as static final fields.


melloware (approved):
LGTM

tandraschko:
Im not aure If we should do just optimizations to the showcase

PR #13877, opened 2025-05-30
#13876 DefaultCommand: Infinite recursion when scope is the same as parent form

### LINKED ISSUE #13876: DefaultCommand: Uncaught RangeError: Maximum call stack size exceeded when in scope is passed only form
### Describe the bug

Using <p:defaultCommand> with the scope attribute set to the same element as the parent <h:form> causes infinite recursion and a StackOverflow error when pressing the Enter key in an input field.

Will create a PR

### Reproducer

https://github.com/kamilkrzywanski/primefaces-test/tree/RECURSION_BUG

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

A lot of

### Theme

Any

### JSF implementation

Mojarra

### JSF version

Any

### Java version

Any

### Browser(s)

_No response_

kamilkrzywanski:
Fixes #13876

### DIFF
--- a/primefaces/src/main/frontend/packages/components/src/forms/forms.defaultcommand.widget.js
+++ b/primefaces/src/main/frontend/packages/components/src/forms/forms.defaultcommand.widget.js
@@ -62,7 +62,7 @@ PrimeFaces.widget.DefaultCommand = class DefaultCommand extends PrimeFaces.widge
         });
 
         // Add keydown listener to scope if available
-        if (this.scope) {
+        if (this.scope && this.scope[0] !== closestForm[0]) {
             this.scope.off(scopeKeydown).on(scopeKeydown, function(e) {
                 if (e.key === 'Enter') {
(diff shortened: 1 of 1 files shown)


primefaces/src/main/frontend/packages/components/src/forms/forms.defaultcommand.widget.js
```diff
@@ -62,7 +62,7 @@ PrimeFaces.widget.DefaultCommand = class DefaultCommand extends PrimeFaces.widge
         });
 
         // Add keydown listener to scope if available
```
melloware:
```suggestion
        // Add keydown listener to scope if not the closest form to prevent recursion
```

melloware (approved):

kamilkrzywanski:
@melloware @melloware has been updated; in some edge cases, it was selecting the wrong button in this situation.

PR #13901, opened 2025-06-10
AccordionPanel: key-based active state (#13890)

### LINKED ISSUE #13890: AccordionPanel: key-based state instead of index-based state
### Description

This is more of a proposal, and probably applies to a lot more components than just p:accordionPanel, but it's the one I'm the most interested in 🙂
I'm happy to help implement this, but would really appreciate some thoughts and feedback on this!
https://github.com/primefaces/primefaces/issues/12926 might be related to this as well

The current index-based approach has a huge downside: If MVS is used, the component updates and the amount of rendered tabs changes somehow (e.g. `rendered` attribute on `p:tab`, or `value` is used on `p:accordionPanel`) the indices won't match the accordions the user expects to be active.

Take this example:
```xhtml
            <p:selectBooleanCheckbox value="#{testView.renderSecondTab}" label="render second tab">
                <p:ajax process="@this" update="myAccordionPanel"/>
            </p:selectBooleanCheckbox>
            <p:accordionPanel id="myAccordionPanel" multiple="true" activeIndex="0,2" multiViewState="true">
                <p:tab title="First">
                    should be active
                </p:tab>
                <p:tab title="Second" rendered="#{testView.renderSecondTab}">
                    should be active
                </p:tab>
                <p:tab title="Third">
                    should be collapsed
                </p:tab>
                <p:tab title="Fourth">
                    should be active
                </p:tab>
            </p:accordionPanel>
```

An index based approach is imo not optimal for a component that can dynamically change and "mess up" the indices

I'm imagining something like that:

```xhtml
            <p:accordionPanel id="myAccordionPanel" multiple="true" activeKeys="first,second,fourth" multiViewState="true">
                <p:tab key="first" title="First">
                    should be active
                </p:tab>
                <p:tab key="second" title="Second" rendered="#{testView.renderSecondTab}">
                    should be active
                </p:tab>
                <p:tab key="third" title="Third">
                    should be collapsed
                </p:tab>
                <p:tab key="fourth" title="Fourth">
                    should be active
                </p:tab>
            </p:accordionPanel>
```

### Additional context

I think doing this in 16.x only would be beneficial as this will likely be a breaking change

There are a couple of things to consider, at least what I'm aware of:
- be backwards compatible to still support activeIndex? or just move on and break things?
- what happens when no keys are defined? when only a few are defined? do we auto generate them (like myfaces does with j_id_123)?
- probably other components that would benefit from this as well

jungm:
Fix #13890, `p:tab` now has a key attribute, which can be used to reference it in the `active` attribute (renamed from activeIndex) in `p:accordionPanel`.

Referencing tabs by index is still fully supported

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/accordionpanel/AccordionPanel.java
+++ b/primefaces/src/main/java/org/primefaces/component/accordionpanel/AccordionPanel.java
@@ -159,10 +159,18 @@ public class AccordionPanel extends AccordionPanelBase {
         super.processUpdates(context);
 
         ELContext elContext = getFacesContext().getELContext();
-        ValueExpression expr = ValueExpressionAnalyzer.getExpression(elContext,
+        ValueExpression activeExpr = ValueExpressionAnalyzer.getExpression(elContext,
                 getValueExpression(PropertyKeys.active.toString()), true);
-        if (expr != null && !expr.isReadOnly(elContext)) {
-            expr.setValue(elContext, getActive());
+        if (activeExpr != null && !activeExpr.isReadOnly(elContext)) {
+            activeExpr.setValue(elContext, getActive());
+            resetActive();
+        }
+
+        // Deprecated
+        ValueExpression activeIndexExpr = ValueExpressionAnalyzer.getExpression(elContext,
+                getValueExpression(PropertyKeys.activeIndex.toString()), true);
+        if (activeIndexExpr != null && !activeIndexExpr.isReadOnly(elContext)) {
+            activeIndexExpr.setValue(elContext, getActive());
             resetActive();
         }
     }
--- a/primefaces/src/main/java/org/primefaces/component/accordionpanel/AccordionPanel.java
+++ b/primefaces/src/main/java/org/primefaces/component/accordionpanel/AccordionPanel.java
@@ -161,22 +161,22 @@ public class AccordionPanel extends AccordionPanelBase {
         ELContext elContext = getFacesContext().getELContext();
         ValueExpression activeExpr = ValueExpressionAnalyzer.getExpression(elContext,
                 getValueExpression(PropertyKeys.active.toString()), true);
-        if (activeExpr != null && !activeExpr.isReadOnly(elContext)) {
-            activeExpr.setValue(elContext, getActive());
-            resetActive();
+
+        // Fallback to deprecated activeIndex
+        if (activeExpr == null) {
+            activeExpr = ValueExpressionAnalyzer.getExpression(elContext,
+                    getValueExpression(PropertyKeys.activeIndex.toString()), true);
         }
 
-        // Deprecated
-        ValueExpression activeIndexExpr = ValueExpressionAnalyzer.getExpression(elContext,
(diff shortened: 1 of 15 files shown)


primefaces/src/main/java/org/primefaces/component/accordionpanel/AccordionPanelRenderer.java
```diff
@@ -252,6 +251,9 @@ protected void encodeTab(FacesContext context, AccordionPanel component, Tab tab
         writer.writeAttribute(HTML.ARIA_CONTROLS, clientId, null);
         writer.writeAttribute(HTML.ARIA_LABEL, tab.getAriaLabel(), null);
         writer.writeAttribute("tabindex", tabindex, null);
+        if (tab.getKey() != null) {
+            writer.writeAttribute("tabKey", tab.getKey(), null);
```
tandraschko: as its our own data-attr, i would use something like: data-key

primefaces/src/main/java/org/primefaces/component/accordionpanel/AccordionPanelRenderer.java
```diff
@@ -56,15 +56,15 @@ public void decode(FacesContext context, AccordionPanel component) {
...
+                component.setActive("");
             }
             else {
-                component.setActiveIndex(active);
+                component.setActive(active);
             }
 
             if (component.isMultiViewState()) {
                 AccordionState as = component.getMultiViewState(true);
-                as.setActiveIndex(component.getActiveIndex());
```
tandraschko, line 67: what happens if the user uses the activeIndex attr? is it remapped correctly?

primefaces/src/main/java/org/primefaces/component/accordionpanel/AccordionPanel.java
```diff
@@ -159,16 +159,24 @@ public void processUpdates(FacesContext context) {
...
+        }
+
+        // Deprecated
+        ValueExpression activeIndexExpr = ValueExpressionAnalyzer.getExpression(elContext,
                 getValueExpression(PropertyKeys.activeIndex.toString()), true);
-        if (expr != null && !expr.isReadOnly(elContext)) {
-            expr.setValue(elContext, getActiveIndex());
-            resetActiveIndex();
+        if (activeIndexExpr != null && !activeIndexExpr.isReadOnly(elContext)) {
+            activeIndexExpr.setValue(elContext, getActive());
```
tandraschko: what happens if the user uses the activeIndex attr? is it remapped correctly?

tandraschko:
@melloware wanna review, too?

what about other components? does tabview has mvs?

melloware:
yep let me review this.

melloware (approved):

melloware:
This looks great. @jungm can you look at TabView also?  It's very similar to this component.

jungm:
@melloware will do, but I think this should be a different issue/pr for tabview?

melloware:
yeah agreed separate issue and PR. I will merge this one.

PR #13903, opened 2025-06-10
Tomcat 10.1.42

melloware:


melloware:
@christophs78 interesting just bumping a minor version of Tomcat and the same 4 tests failed across the board!

christophs78:
I´ll try to look at this in the evening - following test´s fail:

```
Error:  Errors: 
Error:    FileUpload011Test.advancedSingleChunkedUpload:126 » Timeout Timeout while waiting for custom guard!
Error:    FileUpload011Test.advancedSingleChunkedUploadFileLimit:194 » Timeout Timeout while waiting for custom guard!
Error:    FileUpload011Test.advancedSingleChunkedUploadTwice:145 » Timeout Timeout while waiting for custom guard!
Error:    FileUpload013Test.dynamicAllowTypes:53 » Timeout Timeout while waiting for custom guard!
```

Hard to believe this is coincidence.

melloware:
Thank you!

christophs78:
Hm. I can´t reproduce this (yet) locally on Windows 11 + Tomcat 10.1.42 + Chrome 137. 
I´ll keep an eye on this. Maybe i can find something within the next few days.

christophs78:
On Github Actions + Ubuntu 24.04 + Chrome 137:
works with Tomcat up to 10.1.41, is broken since 10.1.42

melloware:
interesting!

christophs78:
following code fails (org.primefaces.selenium.component.FileUpload#getAdvancedUploadButton)

 ```java
   /**
     * Gets the Upload button of the widget.
     * The button is only rendered in advanced mode.
     * @return the widget's upload button
     */
    public WebElement getAdvancedUploadButton() {
        WebElement element = findElement(By.cssSelector(".ui-fileupload-buttonbar button.ui-fileupload-upload"));

        WebElement guarded = Guard.custom(
            element,
            200,
            ConfigProvider.getInstance().getTimeoutFileUpload(),
            PrimeExpectedConditions.script("return " + getWidgetByIdScript() + ".files.length === 0;"));

        return guarded;
    }
```

So files.length will never get 0 on Tomcat 10.1.42 on Ubuntu 24.4.

https://tomcat.apache.org/tomcat-10.1-doc/changelog.html doesn´t  IMO shout out loud what they changed that influences PF fileupload. Maybe i can find some more time comming weekend to dig deeper into this.

Best first bet they "played arround" with Commons FileUpload-versions. Maybe Commons FileUpload (used by Tomcat) has a regression or incompatiblity. See https://github.com/apache/tomcat/commits/10.1.x/

**Comment by christophs78** [comment]
Just updated Tomcat to 10.1.41. (the last working version in our usecase)
Maybe we simply wait for one or two another Tomcat 10.1.x-patch releases and hope the issue goes away without much effort on our side. (Not sure yet whether i want to jump into this issue in - depth.)

**Comment by melloware** [comment]
totally agree!

**Comment by melloware** [comment]
lets see if Tomcat fixes it.

**Comment by pizzi80** [comment]
> Best first bet they "played arround" with Commons FileUpload-versions. Maybe Commons FileUpload (used by Tomcat) has a regression or incompatiblity. See https://github.com/apache/tomcat/commits/10.1.x/

Exactly, they have released and upgraded the internal "repackaged" 
Apache Common FileUpload version 1.6 

I've reproduced the bug also with:

- Java 24
- Tomcat 10.1.42
- Mojarra 
- PrimeFaces 13.0.8 

with the following upload (something like...)

```
<p:fileUpload id="upload" widgetVar="uploadWidget" required="true" 
    sequential="true"
    multiple="true" 
    mode="advanced" 
    auto="true" 
    dragDropSupport="false"
    previewWidth="40" 
    accept="image/*" 
    fileLimit="10" 
    listener="#{bean.upload}"
    process="@this" 
    update="@none" 
    immediate="true" 
    global="false"
    oncomplete="if (PF('uploadWidget').files.length === 0) alert('completed')" />
```

**Comment by melloware** [comment]
thanks for confirming @pizzi80

**Comment by pizzi80** [comment]
thanks guys, you saved me from a huge headache in production...

Today I've further investigated and it seems that the problem
is caused by the new Part's limits default values


https://tomcat.apache.org/tomcat-10.1-doc/config/http.html#Common_Attributes


**maxParameterCount**	
The maximum total number of request parameters (including uploaded files) obtained from the query string and, for POST requests, the request body if the content type is application/x-www-form-urlencoded or multipart/form-data. Request parameters beyond this limit will be ignored. A value of less than 0 means no limit. If not specified, a default of 10000 is used. Note that FailedRequestFilter [filter](https://tomcat.apache.org/tomcat-10.1-doc/config/filter.html) can be used to reject requests that exceed the limit.

**maxPartCount**	
The maximum total number of parts permitted in a request where the content type is multipart/form-data. This limit is in addition to maxParameterCount. Requests that exceed this limit will be rejected. A value of less than 0 means no limit. If not specified, a default of 10 is used.

**maxPartHeaderSize**	
The maximum number of header bytes permitted per part in a request where the content type is multipart/form-data. Requests that exceed this limit will be rejected. A value of less than 0 means no limit. If not specified, a default of 512 is used.

With the following setup on Connector inside `server.xml` 
I have no more problem:

```
<Connector ....  maxParameterCount="-1" maxPartCount="-1" maxPartHeaderSize="-1"  ...  />
```

Now the question is: what are some appropriate values ​​
to avoid "DoS attacks" and at the same time always make p:fileUpload work?

**Comment by melloware** [comment]
hmmm so is it maxPartHeaderSize causing the issue?  I am not sure which exactly is making our Integeration tests fail?

**Comment by pizzi80** [comment]
also the first sentence of **maxParameterCount** is not completely clear:

_"The maximum total number of request parameters (including uploaded files) obtained from the query string and, for POST requests"_

how many files I can send? does a `Part` count as 1 parameter?


Also the **maxPartCount** default value is 10 which is very low for common use case of p:fileUpload advanced mode ...

**Comment by pizzi80** [comment]
Playing a bit with the parameters, it seems that the problematic one is **maxPartCount**

The default value is 10, but it is probably not calculated correctly
or the `Part` count is not calculated correctly

I see that PF failing tests are about chuncked input ... maybe is counted twice or N times ... ?
or something related?

If I set `maxPartCount=20` tomcat accepts maximum 2 files from p:fileUpload, so 
it's a 1/10 ... ?

**Comment by pizzi80** [comment]
I've reported the bug to Tomcat

https://bz.apache.org/bugzilla/show_bug.cgi?id=69721

I'm not sure it's well described, but at least it's a start...

**Comment by christophs78** [comment]
With Tomcat 10.1.42 following was changed: https://github.com/apache/tomcat/commit/e34fe96ef8ee782b0e56b64358e8dc57cbe336a6#diff-57d2f0a72170743f6c3687a48997b2aa37d8d209efe200f00a0b9dc51fc7e572

**Comment by christophs78** [comment]
for a simple usecase there are at least 10 parts:
image

**Comment by pizzi80** [comment]
exactly the default value it's 10 and it will break every JSF webapp ...

**Comment by christophs78** [comment]
@pizzi80: Thanks for pointing us on maxPartCount and maxPartHeaderSize!

**Comment by rzo1** [comment]
We ran into issues in our app due to the new default setting, which led to a frustrating debugging session and several TomEE rebuilds (with varying Tomcat versions). Eventually, I discovered that version 10.1.41 was the last one that worked, and that things would be fine again in 10.1.43 (thanks to the increased default maxPartCount or - of course - by overriding  the parameter in the connector). @jungm just pointed me to this PR—wish I had seen it much earlier, it would've saved @mawiesne and me a lot of debugging time! 😄

Thanks for reporting the issue @pizzi80 and everyone involved ;-)

**Comment by melloware** [comment]
yeah @rzo1 i think everyone is a little frustrated at this Tomcat change in a point release.

**Comment by pizzi80** [comment]
thanks for the feedback! 
I'm glad that also my contribution helped someone !
... I know that feeling... :sweat_smile: :expressionless:

To recap, the situation is the following:

1) JSF and PrimeFaces both add other request parameters
    so the multipart request parameter count is higher then expected

2) With JSF the relative Tomcat exception it's not logged.

3) Using ajax what you get is only a generic error 500 in chrome dev tools!

4) This is the official Tomcat ticket about this thing:
https://bz.apache.org/bugzilla/show_bug.cgi?id=69710

5) In Tomcat 10.1.43 the default value of **maxPartCount**
will be `50` and imho will still break some webapps
with big forms or `p:fileUpload` with multiple files 

6)  The configuration and the math to calculate the max amout of memory for multipart request,
is the following. 

for example inside `server.xml`:
```
<Connector ....... maxConnections="1000" maxPartCount="100" maxPartHeaderSize="1024"  />
```

with these values, Tomcat 10.1 and 11 will use:

1000 x 100 x 1KB x 2 =~ 200MB 



cheers ;)

PR #13947, opened 2025-06-27
Fix #13946: ToggleSwitch proper use of aria-checked

### LINKED ISSUE #13946: ToggleSwitch: Invalid aria-checked status
### Describe the bug

Disabled ToggleSwitch Elements has input with aria-checked="aria-checked" which is not a valid value.

### Reproducer

_No response_

### Expected behavior

Override that fixes the behavior 

```
protected void encodeInput(FacesContext context, ToggleSwitch toggleSwitch, String clientId, boolean checked) throws IOException {
        // Get the original writer to write most of the input
        String inputId = clientId + "_input";

        writer.startElement("div", toggleSwitch);
        writer.writeAttribute("class", "ui-helper-hidden-accessible", null);
        writer.startElement("input", null);

        // Unfortunately, the original method is not modular, so we have to copy-paste it and fix the one line.

        String ariaLabel = toggleSwitch.getAriaLabel() != null ? toggleSwitch.getAriaLabel() : toggleSwitch.getLabel();

        writer.writeAttribute("id", inputId, "id");
        writer.writeAttribute("name", inputId, null);
        writer.writeAttribute("type", "checkbox", null);
        writer.writeAttribute("autocomplete", "off", null);

        // **** THIS IS THE FIX ****
        if (checked && !toggleSwitch.isDisabled()) {
            writer.writeAttribute("aria-checked", "true", null);
            writer.writeAttribute("checked", "checked", null);
        }
        // **** END OF FIX ****

        writer.writeAttribute("aria-label", ariaLabel, null);

        /*
            Moved to the fix above from the original renderer:
            writer.writeAttribute("checked", "checked", null);
         */

        renderValidationMetadata(context, toggleSwitch);
        renderAccessibilityAttributes(context, toggleSwitch);
        renderPassThruAttributes(context, toggleSwitch, org.primefaces.util.HTML.TAB_INDEX);
        renderOnchange(context, toggleSwitch);
        renderDomEvents(context, toggleSwitch, org.primefaces.util.HTML.BLUR_FOCUS_EVENTS);

        writer.endElement("input");
        writer.endElement("div");
    }
```

### PrimeFaces edition

None

### PrimeFaces version

14.0.12

### Theme

_No response_

### Java version

17

### Browser(s)

_No response_

melloware:
Fix #13946: ToggleSwitch proper use of aria-checked

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/toggleswitch/ToggleSwitchRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/toggleswitch/ToggleSwitchRenderer.java
@@ -131,10 +131,10 @@ public class ToggleSwitchRenderer extends InputRenderer<ToggleSwitch> {
         writer.writeAttribute("name", inputId, null);
         writer.writeAttribute("type", "checkbox", null);
         writer.writeAttribute("autocomplete", "off", null);
-        writer.writeAttribute(HTML.ARIA_CHECKED, checked, null);
         writer.writeAttribute(HTML.ARIA_LABEL, ariaLabel, null);
 
-        if (checked) {
+        if (checked && !component.isDisabled()) {
+            writer.writeAttribute(HTML.ARIA_CHECKED, "true", null);
             writer.writeAttribute("checked", "checked", null);
         }
 


gipoezcan:
Thank you!

PR #13951, opened 2025-06-28
Fix #13945: Firefox hidden input have autocomplete="off"

### LINKED ISSUE #13945: TabView: Dynamic tabs are broken after refresh in Firefox
### Describe the bug

With dynamic TabView, the last viewed tab is broken (not opening and empty) after refresh in Firefox.

### Reproducer

Steps to reproduce:

1. Create a dynamic TabView with caching and three tabs ("On-Demand Loading" example from the Showcase can be used):
```
<h:form id="form">
    <p:tabView widgetVar="test" dynamic="true" cache="true">
      <p:tab title="Header I">One</p:tab>
      <p:tab title="Header II">Two</p:tab>
      <p:tab title="Header III">Three</p:tab>
    </p:tabView>
  </h:form>
```
2. Open the page, click "Header III".
3. Press F5 (Refresh).
4. Click "Header III", observe the tab is not opening. Execute `PF('test').getActiveIndex()` and observe it returns `2`. Click "Header II", then "Header III", observe it is empty.

This issue is specific to Firefox, and is reproducible in an isolated case (but, for an unknown reason, I was not able to reproduce it in the Showcase).

Reproducer:
https://github.com/gino0631/primefaces-test-tabview

### Expected behavior

TabView should work as in other browsers, e.g. Chrome.

### PrimeFaces edition

Community

### PrimeFaces version

15.0.5

### Theme

_No response_

### JSF implementation

MyFaces

### JSF version

4.0

### Java version

21

### Browser(s)

Firefox 140.0 (64-bit)

melloware:
Fix #13945: Hidden input have autocomplete="off"

Since Firefox still has this issue and `autocomplete="off"` is invalid HTML according to W3C: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete

> If including the autocomplete attribute on [hidden](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/hidden) input elements (<input type="hidden">), its value must be an ordered list of space-separated tokens; the on and off keywords are not allowed.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/renderkit/CoreRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/renderkit/CoreRenderer.java
@@ -30,6 +30,7 @@ import org.primefaces.component.api.RTLAware;
 import org.primefaces.context.PrimeApplicationContext;
 import org.primefaces.context.PrimeRequestContext;
 import org.primefaces.convert.ClientConverter;
+import org.primefaces.util.AgentUtils;
 import org.primefaces.util.AjaxRequestBuilder;
 import org.primefaces.util.ComponentUtils;
 import org.primefaces.util.Constants;
@@ -312,6 +313,10 @@ public abstract class CoreRenderer<T extends UIComponent> extends Renderer<T> {
         }
         if (value != null) {
             writer.writeAttribute("value", value, null);
+        }
+        // autocomplete="off" is invalid HTML but fixes Firefox caching issues
+        if (AgentUtils.isFirefox(context)) {
+            writer.writeAttribute("autocomplete", "off", null);
         }
         writer.endElement("input");
     }


tandraschko:
autocomplete=off seems to be valid? https://www.w3schools.com/tags/att_input_autocomplete.asp

melloware:
@tandraschko Autocomplete `on` and `off` is not valid for `type="hidden"` according to this: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete

image


And the W3C checker reports this:
image

So this legacy FireFox issue is still around 20 years later even though its not valid HTML so that is why I only put it in the DOM if its Firefox browser now!

tandraschko:
oh i see

Issue #5299, opened 2019-11-04
SelectOneMenu: with required=true and f:SelectItem with itemDisabled=true for empty value prevents ajax update with validation error


irieill:
## 1) Environment
- PrimeFaces version: **7.0**
- Does it work on the newest released PrimeFaces version? **No**
- Does it work on the newest sources in GitHub? **Not tested**
- Application server + version: **Wildfly 17.0.1.Final**
- Affected browsers: **Tested with Firefox, IE, Edge**

## 2) Expected behavior
Given a p:SelectOneMenu with required=true, with an initial empty value and a f:SelectItem with itemDisabled=true for empty value, it should be possible to update the value via Ajax (regardless if p:ajax or f:ajax) as it is possible with simple h:SelectOneMenu.

## 3) Actual behavior
Unable to change the value because of a validation error (value required).

Interesting side observation: selecting an option of one of the p:SelectOneMenus first, leads to the error. selecting an option of one of the h:SelectOneMenus right after, all menus are updated with the correct value, except the first selected one.

## 4) Steps to reproduce
Select item from one of the menus.

## 5) Sample XHTML
Note the mix of p:SelectOneMenu, h:SelectOneMenu, p:ajax and f:ajax.
```
<!DOCTYPE html>
<html
  xmlns="http://www.w3.org/1999/xhtml"
  xmlns:f="http://xmlns.jcp.org/jsf/core"
  xmlns:h="http://xmlns.jcp.org/jsf/html"
  xmlns:p="http://primefaces.org/ui"
>
<f:view>
  <h:head></h:head>
  <h:body>
    <p:log></p:log>
    <h:form id="form">
      <p:growl>
        <p:autoUpdate></p:autoUpdate>
      </p:growl>
        <p:outputPanel id="subcontainer">
          <p:selectOneMenu
            id="dropdown_prfa"
            value="#{dropdownController.value}"
            required="true"
          >
            <f:ajax
              execute="dropdown_prfa"
              render="subcontainer"
            >
            </f:ajax>
            <f:selectItem
              itemLabel="?"
              itemValue=""
              noSelectionOption="true"
              itemDisabled="true"
            />
            <f:selectItems
              value="#{dropdownController.options}"
              var="option"
              itemLabel="#{option}"
              itemValue="#{option}"
            />
          </p:selectOneMenu>
          <br />
          <p:selectOneMenu
            id="dropdown_prpa"
            value="#{dropdownController.value}"
            required="true"
          >
[...]

PR #13956, opened 2025-06-30
SelectOneRadio can select disabled items

### LINKED ISSUE #13954: SelectOneRadio: Selecting disabled items
### Describe the bug

It is possible to select a disabled option in the radio button if you modify the value of one of the enabled options to be the value of the disabled option.

Full writeup: https://stackoverflow.com/questions/21328768/primefaces-dialog-framework-not-working

### Reproducer

PrimefacesTest project: https://github.com/EvanKnowles/primefaces-test-radio-select

In the primfaces test project, open the default test.xhtml. There is a radio button with two disabled items. Clicking the button will set the value of the enabled `Two` item to be the `three-disabled` value. Clicking Two afterwards will select the third disabled item.

### Expected behavior

_No response_

### PrimeFaces edition

Community

### PrimeFaces version

15

### Theme

_No response_

### JSF implementation

All

### JSF version

3.0

### Java version

24

### Browser(s)

_No response_

EvanKnowles:
Include SelectOneRadio integration test case showing selection of a disabled item

Fix #13954

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/selectoneradio/SelectOneRadio006.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/selectoneradio/SelectOneRadio006.java
@@ -1,0 +1,48 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2025 PrimeTek Informatics
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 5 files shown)


tandraschko:
i see
so the disabled handling is only implemented in the client side JS and as you set the value via JS directly, its not validted on server side anymore?

EvanKnowles:
Yes, server side doesn't validate so you can get invalid values set on the bean.

tandraschko:
wanna also work to fix it?

this should be the code involved: https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/java/org/primefaces/renderkit/SelectRenderer.java#L402

EvanKnowles:
Reading through the linked issues and the commented out code that you reverted a while back, it seems like making the change there will cause other issues to re-appear, so I've added a check where this is called to ensure that the values are valid before setting the submitted value on the select one radio component.

tandraschko:
FYI: other tests are failing now

melloware:
@tandraschko it looks like you commented that line out 7 years ago to not throw the exception but I couldn't find any explanation of why.

EvanKnowles:
I found a related issue around Ajax posts not working, might have been that? Checking out the other tests, not sure how I missed them on my test run.

tandraschko:
> @tandraschko it looks like you commented that line out 7 years ago to not throw the exception but I couldn't find any explanation of why.

no idea anymore
propably we aligned it with the MF implementation?

melloware:
I just pushed a fix to this branch where I removed `ignore silently for now` comment and reinstated this 

```java
throw new FacesException("Disabled select item has been submitted. ClientId: " + component.getClientId(context));
```

EvanKnowles:
With that exception restored, we probably don't need my other changes to the renderer - should I revert them?

tandraschko:
please double check that all with MF codebase and jira tickets associated

melloware:
@tandraschko i have been researching this as I told @EvanKnowles to open this ticket. It looks like 7 years ago @cnsgithub opened this issue and fixed it. Then a a little while later you refactored with @georgwolf to fix grouping and commented this line out.

image

you just commented `ignore silently for now` as I think you were refactoring and maybe just forgot to restore this after you refactored?

image

EvanKnowles:
Here's a brief comment on uncommenting the lines breaking stuff: https://github.com/primefaces/primefaces/issues/5299#issuecomment-549519925

melloware:
@EvanKnowles yes please revert your change in SelectOneRenderer.java then we can watch the IT's run again with my change

EvanKnowles:
Weird, my integration test is failing now on a strange CSS selector error. Sorry, I'll work out why and fix the test.

tandraschko:
it may be that an exception will be thrown and the HTML element is not visible because of maybe an error page is displayed?

EvanKnowles:
That's exactly what was happening, eventually I noticed the flash of an error page - pushed a fixed test.

melloware:
Looks like it is all passing now!

melloware (approved):

Issue #13246, opened 2025-01-24
DataTable: toggle filter doesnt work when placed in column header


tandraschko:
if you take ui/data/datatable/filter.xhtml and place the toggler e.g. in a action column, it instantly toggles back:

```
                    <p:column>
                        <f:facet name="header">
                            <div class="flex align-items-center justify-content-between">
                                <div align="right">
                                    <p:toggleSwitch id="manualToggle" styleClass="vertical-align-middle ml-3"
                                                    onchange="PF('customersTable4').toggleFilter()" onIcon="pi pi-filter"
                                                    offIcon="pi pi-filter-slash"/>
                                </div>
                            </div>
                        </f:facet>
```

tandraschko:
thats how i want to use it to avoid the big header row:

![Image](https://github.com/user-attachments/assets/76618c3d-7d0e-4e98-b26b-083674a54fc2)

tandraschko:
it seems ToggleSwitch#encodeEnd is invoked twice

there seems to be some weird stuff with column header encoding

however, i dont have that time, i will commit workaround

wlhUser:
Hello , 
I'm facing the same problem as described in the current ticket, with a <p:selectCheckboxMenu/> in the table header. It is also encoded twice like, and shows a weird behavior.
I'm testing with **PF 15.0.5 jakarta**

```
pf.cw("SelectCheckboxMenu","widget_error_list_statusFilter",{id:"error_list:statusFilter"",updateLabel:true,multiple:true,renderPanelContentOnClient:true,appendTo:"@(body)"});
pf.cw("SelectCheckboxMenu","widget_error_list_statusFilter",{id:"error_list:statusFilter"",updateLabel:true,multiple:true,renderPanelContentOnClient:true,appendTo:"@(body)"});
```

According to my debugging, the reason might be #13016.

The first time the header components are encoded, comes from DataTableRenderer line 224. See screenshot.

![Image](https://github.com/user-attachments/assets/60f672ab-0864-46ec-8740-2ed930c2a774)

This comes from #13016.

The second time the header components are encoded comes from DataTableRenderer line 770.

![Image](https://github.com/user-attachments/assets/fee69dad-b52e-4694-9af5-b5e6c22d4f1e)


Maybe I also open a new ticket for this problem, as the current one is already closed.

melloware:
yes open a new ticket and reproducer please.

PR #13959, opened 2025-06-30
Fix #13953: DataRenderer should not encode header facet for ARIA Header

### LINKED ISSUE #13953: DataTable: Some components do not work within Header facet
### Describe the bug

When placing components like <p:selectOneMenu/> or <p:selectCheckboxMenu/> in the header of a dataTable  columns (facet="header"), they do not work as expected. Clicking on them does not open the expected panel.

This ticket kind of duplicates #13246, which was solved by a workaround only.

### Reproducer

[basic.zip](https://github.com/user-attachments/files/20973113/basic.zip)

The reproducer is based on the PF Showcase for PF 15.0.5.
You need to replace '/primefaces-showcase/src/main/webapp/ui/data/datatable/basic.xhtml' by the content of the attached file.
It adds 3 <p:selectCheckboxMenu/> to the showcase:

- One above the table, which works as expected
- One in the 1st column of the table header, which does not work - **this is the one this ticket is about**
- One in the 2nd column of the table header - this one works due to a trick, but shows also where the problem in the code is.
(See attached screenshot)

Image

Working components open the panel with the select-options when clicking them. The not working one does not.

Showcase needs to be run with:
**-Pnon-ee,mojarra-4.0**

When debugging the problem, you can see that the components within a table header columns, get initialized twice, which is already described in #13246.

Image

According to my debugging, the reason might be #13016, which leads to header components (or parts of them) being encoded twice.

The first time the header components are encoded, comes from DataTableRenderer line 224 (in PF15.0.5). See screenshot. This is the problem - the method **resolveColumnAriaHeaderText** does encode components within the header. I suppose it should handle only the header text and nothing else. However, if you define a **headerText** in the dataTable column, this encoding will not be done. Thats why the select int the Col2 of the example works.

Image

The second time the header components are encoded comes from DataTableRenderer line 770.

Image

### Expected behavior

Components should work within dataTable header column in the same way as outside of a table.

### PrimeFaces edition

None

### PrimeFaces version

15.0.5

### Theme

--

### JSF implementation

Mojarra

### JSF version

4 

### Java version

17

### Browser(s)

All

melloware:
Fix #13953: 15.0.6 DataRenderer should not encode header facet for ARIA Header

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/datatable/DataTableRenderer.java
@@ -674,8 +674,10 @@ public class DataTableRenderer extends DataRenderer<DataTable> {
         writer.startElement("th", component);
         writer.writeAttribute("id", clientId, null);
         writer.writeAttribute("class", columnClass, null);
-        writer.writeAttribute(HTML.ARIA_LABEL, ariaHeaderLabel, null);
         writer.writeAttribute("scope", "col", null);
+        if (LangUtils.isNotBlank(column.getAriaHeaderText())) {
+            writer.writeAttribute(HTML.ARIA_LABEL, ariaHeaderLabel, null);
+        }
         if (component != null) {
             renderDynamicPassThruAttributes(context, component);
         }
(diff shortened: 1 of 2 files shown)


melloware:
@tandraschko  i was trying to be too clever and should not have bene encoding the HEADER facet.  If you want ARIA Header Text they should just declare it. This was causing your issue in https://github.com/primefaces/primefaces/issues/13246

PR #13973, opened 2025-07-04
Fix #13858: Extending validateFile with allowMediaTypes

### LINKED ISSUE #13858: FileUpload: Introduce attribute for MIME Type Validation in <p:validateFile>
### Description

**Current Behavior**

- The accept attribute in <p:fileUpload> serves as a client-side hint, guiding users to select appropriate file types. However, it doesn't enforce restrictions, allowing users to bypass it and upload unsupported file types.
- The <p:validateFile> component can enforce file type restrictions (based on filename) using the allowTypes attribute (which utilizes regular expressions). This will be double-checked at server side as well.
- The <p:validateFile> component can enforce content type validation based on java libs like Apache Tika or mime-types by setting contentType="true". The validation is based on the types defined in the accept attribute.

**Identified Issue**
There's no dedicated attribute, such as allowMediaTypes, to explicitly specify allowed MIME types for server-side content type validation. 

**Proposed Enhancement**
Introduce an allowMediaTypes attribute in <p:validateFile> to explicitly define permissible MIME types, enhancing clarity and maintainability.

Optional:
Automatic populate allowTypes when empty and contentType="true" to ensure consistency between client-side hints and server-side validations. When contentType="true" the file content is checked serverside. If the content is invalid a message will appear which is based on allowTypes. This means that allowedTypes is mandatory when using contentType="true", otherwise a incomplete/incorrect 'null' file message will appear.

Example with allowMediaTypes:
```
<p:fileUpload
  accept=".xlsx">
  <p:validateFile
    sizeLimit="4194304"
    allowTypes="/^order.*\.xlsx$/"
    allowMediaTypes="application/vnd.openxmlformats-officedocument.spreadsheetml.sheet,application/zip"
    contentType="true"
  />
</p:fileUpload>
```

Change is backward compatible, this still works:
```
<p:fileUpload
  accept="application/vnd.openxmlformats-officedocument.spreadsheetml.sheet,application/zip">
  <p:validateFile
    sizeLimit="4194304"
    allowTypes="/^order.*\.xlsx$/"
    contentType="true"
  />
</p:fileUpload>
```

### Additional context

Reference: [Discussion #3964](https://github.com/orgs/primefaces/discussions/3964)

### LINKED ISSUE #14932: FileUpload: content type validation not validating files correctly
### Describe the bug

It seems that when using something like:
```xml
<p:validateFile allowTypes="/(.|/)(gif|jpeg|jpg|png)$/" contentType="true">
```
The string in  allowTypes is not whats being used to perform validation of the content type, but instead the p:fileUpload accept="" attribute.
Both attributes are passed in FileUploadUtils#isValidType:
Image

But only "accept" is passed in to FileUploadUtils#isValidFileContent, 

A way to reproduce:
1 ) Use `<p:validateFile allowTypes="/(.|/)(pdf)$/" contentType="true">` (with a content type library in the classpath)
2 ) Upload a zip file with extension renamed to .pdf
3) This causes FileUploadUtils#isValidFileContent to do an early return because allowedContentTypes is empty (since accept="" attribute is being passed in)

If you keep the example but add accept=".pdf", it enters this if that only checks file extension without checking content type:
https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java#L330

### Reproducer

_No response_

### Expected behavior

When a file content type library is provided maybe we should make it so that both file extension AND content type must match instead of just checking file extension. Or maybe this is just a case of wrong attribute being passed in?  Unclear to me, I dont know in which cases allowTypes would start with "."

### PrimeFaces edition

Community

### PrimeFaces version

15.0.15

### Theme

_No response_

### JSF implementation

MyFaces

### JSF version

jakarta

### Java version

25

### Browser(s)

_No response_

DazzlingBob:
This PR introduces enhanced file validation capabilities with the following changes:

- New allowMediaTypes property: Allows specifying MIME types directly on the validator (e.g., image/*, application/pdf) as an alternative to the component's accept attribute
- resolveAccept() method: Determines the appropriate accept value for validation, prioritizing validator's allowMediaTypes over component's accept attribute when content-type validation is enabled
- resolveAllowTypes() method: Intelligently resolves file extension patterns, either from the validator's allowTypes property or by extracting extensions from the component's accept attribute
- extractAllowTypes() method: Converts HTML accept attributes containing file extensions (e.g., .pdf,.doc) into JavaScript-compatible regex patterns, filtering out MIME types and invalid extensions

These changes provide more flexibility in file validation configuration while maintaining backward compatibility. 

Fix #13858 
Fix #14932

### DIFF
--- a/primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
+++ b/primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
@@ -43,6 +43,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
@@ -470,4 +471,27 @@ public class FileUploadUtils {
             return decimalFormat.format((bytes / Math.pow(1024, i))) + " " + sizes[i];
         }
     }
+
+    /**
+     * Extracts file extensions from an HTML accept attribute and converts them to a JavaScript-compatible
+     * regular expression pattern. Only processes dot-prefixed extensions (e.g., ".pdf", ".doc") and
+     * ignores MIME types (e.g., "image/*", "text/plain").
+     * @param componentAccept Raw accept attribute from the component (e.g., ".pdf,.doc,image/*")
+     * @return a regex pattern matching the file extensions (e.g., "/.*\.(pdf|doc)/"), or null if no valid extensions found
+     */
+    public static String extractAllowTypes(String componentAccept) {
+        if (LangUtils.isBlank(componentAccept)) return null;
+        return Arrays.stream(componentAccept.split(","))
+                .map(String::trim)
+                .filter(part -> part.startsWith("."))
+                .map(part -> part.substring(1).trim())
+                .filter(ext -> !ext.isEmpty())
+                .filter(ext -> ext.matches("[a-zA-Z0-9]{1,10}"))
+                .distinct()
+                .collect(Collectors.collectingAndThen(
+                        Collectors.toList(),
+                        list -> list.isEmpty() ? null : "/.*\\.(" + String.join("|", list) + ")/"
+                ));
+    }
+
 }
--- a/primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
+++ b/primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
@@ -476,12 +476,12 @@ public class FileUploadUtils {
      * Extracts file extensions from an HTML accept attribute and converts them to a JavaScript-compatible
      * regular expression pattern. Only processes dot-prefixed extensions (e.g., ".pdf", ".doc") and
      * ignores MIME types (e.g., "image/*", "text/plain").
-     * @param componentAccept Raw accept attribute from the component (e.g., ".pdf,.doc,image/*")
+     * @param accept Raw accept attribute from the component (e.g., ".pdf,.doc,image/*")
      * @return a regex pattern matching the file extensions (e.g., "/.*\.(pdf|doc)/"), or null if no valid extensions found
      */
-    public static String extractAllowTypes(String componentAccept) {
-        if (LangUtils.isBlank(componentAccept)) return null;
-        return Arrays.stream(componentAccept.split(","))
+    public static String extractAllowTypes(String accept) {
+        if (LangUtils.isBlank(accept)) return null;
+        return Arrays.stream(accept.split(","))
                 .map(String::trim)
                 .filter(part -> part.startsWith("."))
                 .map(part -> part.substring(1).trim())
--- a/primefaces/src/main/java/org/primefaces/validate/FileValidator.java
+++ b/primefaces/src/main/java/org/primefaces/validate/FileValidator.java
@@ -35,7 +35,6 @@ import org.primefaces.util.MessageFactory;
 import org.primefaces.validate.base.AbstractPrimeValidator;
 import org.primefaces.virusscan.VirusException;
 
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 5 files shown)


melloware:
Running build now. I will let Thomas review also but from first glance this looks like a nice addition/refactor.

primefaces/src/main/java/org/primefaces/validate/FileValidator.java
```diff
@@ -60,51 +63,119 @@ public enum PropertyKeys {
...
+    }
+
+    /**
+     * Extracts file extensions from an HTML accept attribute and converts them to a JavaScript-compatible
+     * regular expression pattern. Only processes dot-prefixed extensions (e.g., ".pdf", ".doc") and
+     * ignores MIME types (e.g., "image/*", "text/plain").
+     * @param componentAccept Raw accept attribute from the component (e.g., ".pdf,.doc,image/*")
+     * @return a regex pattern matching the file extensions (e.g., "/.*\.(pdf|doc)/"), or null if no valid extensions found
+     */
+    private static String extractAllowTypes(String componentAccept) {
```
tandraschko: can you please add some units tests?
you could also move it to the FileuploadUtils if necessary for better accessibility.

primefaces/src/main/java/org/primefaces/validate/FileValidator.java
```diff
@@ -60,51 +63,119 @@ public enum PropertyKeys {
...
+     */
+    private String resolveAllowTypes(String componentAccept, UIComponent component) {
+        String allowTypes = getAllowTypes();
+        if (LangUtils.isBlank(allowTypes) && Boolean.TRUE.equals(getContentType())) {
+            String extractedAllowTypes = extractAllowTypes(componentAccept);
+            if (LangUtils.isNotBlank(extractedAllowTypes)) {
+                allowTypes = extractedAllowTypes;
+            }
+            else {
+                LOGGER.log(Level.WARNING, "Attribute allowTypes is missing in p:validateFile. ClientId: " + component.getClientId());
```
tandraschko: adjust message that either allowTypes or accept is required

primefaces/src/main/java/org/primefaces/validate/FileValidator.java
DazzlingBob: Moved function extractAllowTypes to FileUploadUtils and added unit tests for extractAllowTypes.

primefaces/src/main/java/org/primefaces/validate/FileValidator.java
DazzlingBob: Updated warning message.

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
```diff
@@ -470,4 +471,27 @@ public static String formatBytes(Long bytes, Locale locale) {
...
     }
+
+    /**
+     * Extracts file extensions from an HTML accept attribute and converts them to a JavaScript-compatible
+     * regular expression pattern. Only processes dot-prefixed extensions (e.g., ".pdf", ".doc") and
+     * ignores MIME types (e.g., "image/*", "text/plain").
+     * @param componentAccept Raw accept attribute from the component (e.g., ".pdf,.doc,image/*")
+     * @return a regex pattern matching the file extensions (e.g., "/.*\.(pdf|doc)/"), or null if no valid extensions found
+     */
+    public static String extractAllowTypes(String componentAccept) {
```
tandraschko: i think in this scope, we can just call it "accept" instead of "componentAccept"

primefaces/src/main/java/org/primefaces/validate/FileValidator.java
```diff
@@ -60,51 +62,99 @@ public enum PropertyKeys {
...
+    }
+
+    /**
+     * Resolves the accept value for validation.
+     * If content-type validation is enabled, returns either the validator's allowMediaTypes
+     * or the component's accept attribute.
+     * @param componentAccept Raw accept attribute from the component
+     * @return the accept value to use for validation, or null if content type validation is disabled
+     */
+    private String resolveAccept(String componentAccept) {
```
tandraschko: same here
in the method scope its just "accept"

primefaces/src/main/java/org/primefaces/validate/FileValidator.java
```diff
@@ -60,51 +62,99 @@ public enum PropertyKeys {
...
+    /**
+     * Resolves the allowed file types pattern for validation.
+     * Returns the validator's allowTypes if configured. If allowTypes is not configured
+     * and content-type validation is enabled, extracts file extensions from the component's
+     * accept attribute and converts them to a regex pattern.
+     * @param componentAccept Raw accept attribute from the component
+     * @param component The UI component being validated
+     * @return the file types pattern to use for validation, or null if none found
+     */
+    private String resolveAllowTypes(String componentAccept, UIComponent component) {
```
tandraschko: same here
in the method scope its just "accept"

primefaces/src/main/java/org/primefaces/util/FileUploadUtils.java
DazzlingBob: Changed.

primefaces/src/main/java/org/primefaces/validate/FileValidator.java
DazzlingBob: Struggled with that naming too, wanted to differentiate the original accept (from the component) and the manipulated accept. Maybe we don't need to do that.

primefaces/src/main/java/org/primefaces/validate/FileValidator.java
DazzlingBob: Changed.

melloware (approved):

tandraschko:
Good PR, thanks for your work!

PR #13997, opened 2025-07-11
Fix #13993: ResourceUtils close inputstream

### LINKED ISSUE #13993: GraphicImage: InputStream not closed
### Describe the bug

When using GraphicImage with dynamic content, the InputStream, passed as StreamedContent in the ValueExpression, seems not to be closed after use, resulting in possible resource leakage.

The InputStream is processed by DynamicContentSrcBuilder and ResourceUtils to convert the stream to a Base64 string. 

This was refactored in v12 and the InputStream was properly closed after use.
See https://github.com/primefaces/primefaces/commit/35fb017806566d02c122d10d03961eaaa51d2558#diff-61f38752c2eac65f3b2cfa01d255cc05b89c0d076eea21622ba5936f3b0b88d2R142 on R142.

However, this was recently removed during a sonar cleanup in https://github.com/primefaces/primefaces/pull/11839/files#diff-61f38752c2eac65f3b2cfa01d255cc05b89c0d076eea21622ba5936f3b0b88d2L155.

If I am not missing anything, IMO the input stream should be properly closed by the ResourceUtil method.

### Reproducer

_No response_

### Expected behavior

The InputStream should be properly closed after use.

### PrimeFaces edition

Community

### PrimeFaces version

>= 15

### Theme

_No response_

### JSF implementation

All

### JSF version

4

### Java version

17

### Browser(s)

-

melloware:
Fix #13993: ResourceUtils close inputstream

### DIFF
--- a/primefaces/src/main/java/org/primefaces/util/DynamicContentSrcBuilder.java
+++ b/primefaces/src/main/java/org/primefaces/util/DynamicContentSrcBuilder.java
@@ -100,7 +100,7 @@ public class DynamicContentSrcBuilder {
                     return ResourceUtils.toBase64(context, is);
                 }
                 catch (IOException e) {
-                    throw new FacesException("Could not open InputStream from InputStream", e);
+                    throw new FacesException("Could not open InputStream from StreamedContent", e);
                 }
             }
         }
--- a/primefaces/src/main/java/org/primefaces/util/DynamicContentSrcBuilder.java
+++ b/primefaces/src/main/java/org/primefaces/util/DynamicContentSrcBuilder.java
@@ -28,6 +28,7 @@ import org.primefaces.context.PrimeApplicationContext;
 import org.primefaces.el.ValueExpressionAnalyzer;
 import org.primefaces.model.StreamedContent;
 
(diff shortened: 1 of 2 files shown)


tandraschko:
i would propably add a try statement around the IS from the caller, not inside the toByteArray

melloware:
OK i moved it out of there and too the calling locations.

primefaces/src/main/java/org/primefaces/util/DynamicContentSrcBuilder.java
```diff
@@ -90,7 +96,12 @@ else if (InputStream.class.isAssignableFrom(type)) {
                 return buildStreaming(context, component, extractedVE, cache);
             }
             else {
-                return ResourceUtils.toBase64(context, (InputStream) value.get());
+                try (InputStream is = (InputStream) value.get()) {
+                    return ResourceUtils.toBase64(context, is);
+                }
+                catch (IOException e) {
+                    throw new FacesException("Could not open InputStream from InputStream", e);
```
tandraschko: from StreamedContent

Issue #10561, opened 2023-08-28
Datepicker with `autoMonthFormat ` is not working as expected


mertsincan:
Reported By PRO User;
> ...Removing the `autoMonthFormat` attribute indeed helps to fix the weird behaviour of the pop-up.
However, it breaks another thing we were dealing with earlier - after the attribute is removed and a value outside allowed min and max is entered from keyboard it is automatically changed to first allowed date; e.g. when the min is 1.8.2023 and I write 1.2.2023 and hit tab, the date is automatically changed to 01.08.2023 without user being notified/warned in any way. The current (correct) behaviour is that the value is formatted to 01.02.2023 but not changed to value inside min-max range and only when the form is submitted, the validator stops the submition and user is notified.
...

Issue #9224, opened 2022-09-21
Add autoMonthFormat attribute to DatePicker


mertsincan:
Whether to format the month. Default is true.

PR #14023, opened 2025-07-23
Fix #14007 Allow disabling autoMonthFormat to prevent the date entered by a user from being set to the min or max date restriction

### LINKED ISSUE #14007: DatePicker: Inconsistent validation month vs year maxDate
### Describe the bug

Using maxDate shows different behaviours in validation.
When the value exceeds the maxDate by a year the validation message shows up.
When the value exceeds the maxDate by just month it will automatically change the value to the max month.

Example: Max Date = 01.09.2025
Entered Date = 01.01.2026 --> Validation shows up
Entered Date = 01.12.2025 --> Value changes to 01.09.2025

### Reproducer

```xml                        
<p:datePicker  value="#{datePickerBean.datum}"
                        yearNavigator="true" yearRange="#{datePickerBean.yearRange}"
                        maxdate="#{datePickerBean.maxDate}"
                        monthNavigator="true">
                            <p:ajax update="msg_datepicker"/>
</p:datePicker>
<p:message id="msg_datepicker" for="@previous" />
```

### Expected behavior

Validation should show up in each case and not change a value automatically. The user might not see the changed value and just continue sending the form.

### PrimeFaces edition

None

### PrimeFaces version

15.0.6 (also tried in 13.0.10)

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.1.0

### Java version

17

### Browser(s)

_No response_

iJMW:
Fix #14007 

After investigating showMinMaxRange and autoMonthFormat some more for issue #14007, which seems to be the same as or similar to #10561, it seems like showMinMaxRange doesn't belong within onInputChange or updateViewDate.

In addition, removing this attribute allows the autoMonthFormat to be set to false and disable the automatic update to the date field in the reproducer to the min or max date. Instead, it provides an error.

I could see this being extended to say that if autoMonthFormat, or renamed to something like autoDateFormat, is true and the input is outside of the mindate or maxdate attributes, it should be adjusted regardless of the year. During testing, I noted that setting the time outside of the maxdate also resulted in a validation error. This attribute seems to be very specific in that it only cares about the month within the same year. The attribute was added as part of #9224.

### DIFF
--- a/primefaces/src/main/frontend/packages/datepicker/datepicker/src/jquery.ui.prime.datepicker.cjs
+++ b/primefaces/src/main/frontend/packages/datepicker/datepicker/src/jquery.ui.prime.datepicker.cjs
@@ -2373,7 +2373,7 @@ $.widget("prime.datePicker", {
     },
 
     onInputChange: function(event) {
-        if ((this.options.autoMonthFormat || !this.inputfield.val() || this.options.showMinMaxRange) && this.options.monthNavigator && this.options.view !== 'month') { 
+        if ((this.options.autoMonthFormat || !this.inputfield.val()) && this.options.monthNavigator && this.options.view !== 'month') { 
             var viewMonth = this.viewDate.getMonth();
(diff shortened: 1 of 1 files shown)


melloware:
You are right not lot of information in #9224 so it must have been added as a PRO support request.

Issue #5014, opened 2019-07-22
Datatable: (scrollable): size column size calculation defect for percentage-width or no width given


regenduft:
Version: Primefaces 6.2.21

## 2) Expected behavior

when no width given, all columns should have same size and be distributed equally over the available scrollWidth

when width="30%", this should be 30% and not 30px!

## 3) Actual behavior

since commit https://github.com/primefaces/primefaces/commit/852741bf1376fc8e966441bffc9a82e80c601b15 to solve issue https://github.com/primefaces/primefaces/issues/4801

column width is NOT calculated correctly anymore, if NOT using box-sizing:border-box

- 30% is interpreted as 30px

- and when no width given, its distributed over the window-width instead of the given scrollWidth

- and columns on the left are small and further right columns get wider

## 4) Steps to reproduce

any datatable with scrollable=true

## 5) Sample XHTML

Primefaces 6.2.21 and 

`<?xml version='1.0' encoding='UTF-8' ?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">

<html lang="en" xmlns="http://www.w3.org/1999/xhtml"
      xmlns:c="http://java.sun.com/jsp/jstl/core"
      xmlns:ui="http://java.sun.com/jsf/facelets"
      xmlns:h="http://java.sun.com/jsf/html"
      xmlns:f="http://java.sun.com/jsf/core"
      xmlns:p="http://primefaces.org/ui" >

	<f:view contentType="text/html; charset=utf-8" >

		<h:head></h:head>

		<h:body>

		   <h3>Vertical</h3>
    <p:dataTable var="car" value="#{testBean.cars}" scrollable="true" scrollWidth="600px" scrollHeight="150">
        <p:column headerText="Id" footerText="Id" width="20%">
            <h:outputText value="#{car.id}" />
        </p:column>
        <p:column headerText="Brand" footerText="Brand" >
            <h:outputText value="#{car.brand}" />
        </p:column>
        <p:column headerText="Year" footerText="Year" >
            <h:outputText value="#{car.year}" />
        </p:column>
        <p:column headerText="Color" footerText="Color" width="30%">
            <h:outputText value="#{car.color}" />
        </p:column>
    </p:dataTable>

		   <h3>Horizontal and Vertical</h3>
[...]

PR #14035, opened 2025-07-29
Fix #14034: Datatable: Columns Not Rendered When Using Mixed 'px' and '%' Widths 

### LINKED ISSUE #14034: Datatable: Columns Not Rendered When Using Mixed 'px' and '%' Widths
### Describe the bug

When rendering a p:dataTable with columns where each column has a defined width, some columns fail to render properly when the width units are mixed (px and %).
Some column headers are missing.

Image

### Reproducer

https://github.com/kamilkrzywanski/primefaces-test/tree/width_bug

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

13,14,15, 16-SNAPSHOT tested

### Theme

Saga

### JSF implementation

Mojarra

### JSF version

Any

### Java version

Any

### Browser(s)

_No response_

kamilkrzywanski:
Fix https://github.com/primefaces/primefaces/issues/14034

### DIFF
--- a/primefaces-integration-tests/pom.xml
+++ b/primefaces-integration-tests/pom.xml
@@ -133,7 +133,7 @@
         <dependency>
             <groupId>org.glassfish.jaxb</groupId>
             <artifactId>jaxb-runtime</artifactId>
-            <version>4.0.5</version>
+            <version>4.0.6</version>
             <scope>runtime</scope>
         </dependency>
 
--- a/primefaces-showcase/pom.xml
+++ b/primefaces-showcase/pom.xml
@@ -190,7 +190,7 @@
         <dependency>
             <groupId>org.glassfish.jaxb</groupId>
             <artifactId>jaxb-runtime</artifactId>
-            <version>4.0.5</version>
+            <version>4.0.6</version>
         </dependency>
     </dependencies>
 


melloware:
I wonder why `box-sizing: content-box;` was there to begin?

melloware:
@jepsar any thoughts on this?  It does seem weird.

FlipWarthog:
I will look into the reproducer and see if I can figure out what's going on. This change, as it is, does affect the rendering of tables. I did a side-by-side comparison of some showcase tables and some render differently, e.g., Selection:

`Before`
image

`After`
image

melloware:
I was thinking scrollable might be broken also

FlipWarthog:
Okay, I dug into it a little more and I don't believe this change is right.

**Background**
If you look at the [MDN docs for box-sizing](https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing), it says:
> `content-box` gives you the **default** CSS box-sizing behavior. If you set an element's width to 100 pixels, then the element's content box will be 100 pixels wide, and the width of any border or padding will be added to the final rendered width, making the element wider than 100px.
> `border-box` tells the browser to account for any border and padding in the values you specify for an element's width and height. If you set an element's width to 100 pixels, that 100 pixels will include any border or padding you added, and the content box will shrink to absorb that extra width. This typically makes it much easier to size elements.


Note that `content-box` is the default behavior, but PrimeFaces specifically (in _common.scss) sets
```css
* {
    box-sizing: border-box;
}
```
So the default for PrimeFaces is now `border-box`, with only a handful of overrides setting it back to the default (like DataTable header/footer columns). Without having the full history of those changes, I would assume there is a good reason for that and removing it for this specific case it probably not correct. Looking at the reproducer in the issue, it does render all the columns at a certain viewport resolution, which makes sense since the %-based widths are being given "what's left" after the fixed-pixel widths are accounted for:
image


@kamilkrzywanski If you are having specific issues after upgrading PrimeFaces to 10+, I think this falls in the category of "theme's/component's change and sometimes you gotta change your code to reflect that". Given the description of the CSS style above, I think the answer is to adjust the problem areas on your side since this is a very sweeping change to make in PF itself.

FlipWarthog:
> I was thinking scrollable might be broken also

@melloware I think scrollable was fine. It looks like any that set a specific width on a column, like all the row select ones, are affected and maybe few others. So if this change is the way to go, I think the PR would need to include updates all the relevant showcase examples to fix the rendering plus a note in the migration guide for the change. Maybe review other places where `box-sizing` is set back to the default?

kamilkrzywanski:
After a long investigation, I believe I've found the root cause.
The last working version of fixColumnWidths was from PrimeFaces 6.1. With that JavaScript, my reproducer behaves correctly, in my opinion.

@melloware @FlipWarthog 

PrimeFaces.widget.DataTable.prototype.fixColumnWidths = function() {
    var d = this;
    if (!this.columnWidthsFixed) {
        if (this.cfg.scrollable) {
            this.scrollHeader.find("> .ui-datatable-scrollable-header-box > table > thead > tr > th").each(function() {
                var h = $(this)
                    , e = h.index()
                    , f = h.width();
                h.width(f);
                if (d.footerCols.length > 0) {
                    var g = d.footerCols.eq(e);
                    g.width(f)
                }
            })
        } else {
            var b = this.jq.find("> .ui-datatable-tablewrapper > table > thead > tr > th")
                , a = b.filter(":visible")
                , c = b.filter(":hidden");
            this.setColumnsWidth(a);
            this.setColumnsWidth(c)
        }
        this.columnWidthsFixed = true
    }
}


To be more precise its commit: 64c70b8c
image

kamilkrzywanski:
In the more modern version, it's probably related to getColumnWidthInfo:
```
    if(!width) {
        width = col[0].style.width;
        isOuterWidth = width && (col.css('box-sizing') === 'border-box');
    }
```

When I negate the condition (col.css('box-sizing') === 'border-box') or simply remove it, everything starts working correctly.

```
PrimeFaces.widget.DataTable.prototype.getColumnWidthInfo = function(col, isIncludeResizeableState) {
    var $this = this;
    var width, isOuterWidth;

    if(isIncludeResizeableState && this.resizableState) {
        width = $this.findColWidthInResizableState(col.attr('id'));
        isOuterWidth = false;
    }

    if(!width) {
        width = col[0].style.width;
        isOuterWidth = width && !(col.css('box-sizing') === 'border-box');
    }

    if(!width) {
        width = col.width();
        isOuterWidth = false;
    }

    return {
        width: width,
        isOuterWidth: isOuterWidth
    };
}
```

kamilkrzywanski:
It might be also instresting :) 
https://github.com/primefaces/primefaces/issues/5014

How about just remove this fragment?

    if(!width) {
        width = col[0].style.width;
        isOuterWidth = width && (col.css('box-sizing') === 'border-box');
    }

FlipWarthog:
Ah I see, I was not looking at the JS. And it looks like some of this relates to resizable columns too. Let me do some more investigation. I think we will need some reproducers to test against using different combinations of 1) fixed widths and percents, 2) resizable columns, and 3) scrollable. Just to be sure what we're fixing is just fixing and not introducing new issues, plus to determine what, if anything, should be included migration notes.

melloware:
Much appreciated @FlipWarthog @kamilkrzywanski

melloware:
OK after discussing with @FlipWarthog  i don't think PF should make any fix at this time. The `border-box` is done for specific scenarios and your scenario is an outlier mixing % and px columns is never recommended.  

@kamilkrzywanski i think it best you patch your local code if you need this but for most users this will break other cases.

PR #14059, opened 2025-08-08
Implement #13700: if oncomplete represents a ValueExpression then run it via executeScript() after action event

### LINKED ISSUE #13700: Ajax: support for post-action EL resolution in oncomplete attribute
### Description

RichFaces supported this construct:
```xml
<a4j:commandButton 
     action="#{bean.modifyProperty}"
     oncomplete="console.log('#{bean.modifiedProperty}');"/>
```
PrimeFaces doesn't. It basically prints the value as it was during initial request.

This could be implemented by replacing the `oncomplete` by basically the same functionality as done by the following pseudo Java code which is invoked in beforePhase of render response:
```java
PrimeFaces.current().executeScript(evaluateExpressionGet(oncomplete));
```
I understand that this is possibly quite a work and is potentially a breaking change, but it has clear long term advantages. More intuitive oncomplete with EL and more terse than all other ways/workarounds.

### Additional context

Stumbled upon this case while migrating RichFaces to PrimeFaces. Work around for now (given that you cannot modify model to use e.g. addCallbackParam):
```xml
<p:commandButton
     action="#{bean.modifyProperty}"
     update="@this"
     pt:data-property="#{bean.modifiedProperty}"
     oncomplete="console.log(document.getElementById('#{component.clientId}').dataset.property);"/>
```

BalusC:
Fix #13700

### DIFF
--- a/primefaces/src/main/frontend/packages/core/src/core/core.ajax.ts
+++ b/primefaces/src/main/frontend/packages/core/src/core/core.ajax.ts
@@ -996,6 +996,7 @@ export class AjaxRequest {
             })
             .always(function(this: PrimeType.ajax.PrimeFacesSettings, data, status, xhr: string | PrimeType.ajax.pfXHR) {
                 const pfArgs = typeof xhr === "string" ? undefined : xhr.pfArgs;
+                const pfScripts = typeof xhr === "string" ? [] : (xhr.pfScripts || []);
 
                 // first call the extension callback (e.g. datatable paging)
                 if (cfg.ext && cfg.ext.oncomplete) {
@@ -1005,6 +1006,11 @@ export class AjaxRequest {
                 // after that, call the end user's callback, which should be called when everything is ready
                 if (cfg.oncomplete) {
                     cfg.oncomplete.call(this, xhr, status, pfArgs, data);
+                }
+
+                // finally, call the programmatic scripts
+                for (const script of pfScripts) {
+                    new Function("xhr", "status", "args", "data", script)(xhr, status, pfArgs, data);
                 }
 
                 if (global) {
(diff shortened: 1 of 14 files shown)


primefaces/src/main/frontend/packages/core/src/core/core.ajax.ts
```diff
@@ -1007,6 +1008,11 @@ export class AjaxRequest {
                     cfg.oncomplete.call(this, xhr, status, pfArgs, data);
                 }
 
+                // finally, call the programmatic scripts
+                for (const script of pfScripts) {
+                    new Function("xhr", "status", "args", "data", script)(xhr, status, pfArgs, data);
+                }
```
BalusC, line 1033: This approach is chosen because we need to retain the function arguments. The executeScript() never supported these.

primefaces/src/main/java/org/primefaces/component/api/DialogReturnAware.java
```diff
@@ -44,7 +44,7 @@ default boolean isDialogReturnEvent(FacesEvent event, FacesContext context) {
             Map<String, String> params = context.getExternalContext().getRequestParameterMap();
             String eventName = params.get(Constants.RequestParams.PARTIAL_BEHAVIOR_EVENT_PARAM);
 
-            return eventName.equals(EVENT_DIALOG_RETURN);
+            return EVENT_DIALOG_RETURN.equals(eventName);
```
BalusC, line 47: This work actually uncovered an unrelated bug here. This line threw NPE when `DynamicOncompleteEvent` was queued but it could as good happen for any other custom AjaxBehaviorEvent.

tandraschko:
Will review on monday

christophs78:
Maybe you looked already into e.g. https://github.com/primefaces/primefaces/actions/runs/16839170687/job/47705934604

> Tests run: 850, Failures: 8, Errors: 369, Skipped: 2

It seems to be related to PrimeFaces CSP implementation. Without CSP your PR seems to run through all integration tests. With CSP enabled a fair amount of them fails. You should be able to reproduce this with e.g. `AjaxStatus001Test`.

BalusC:
Right. I'll take a second look.

primefaces/src/main/frontend/packages/core/src/core/core.ajax.ts
melloware: I have a feeling this new Function is what is causing the CSP issues

melloware: This might have to be changed to use `PrimeFaces.csp.eval(script)` which execute CSP safe dynamic JS.

melloware: Soimething like this might actually work...

```js
for (const script of pfScripts) {
  // Store args in a temp global so the script can use them
  window.__pfTempArgs = {
    xhr: xhr,
    status: status,
    args: pfArgs,
    data: data
  };

  PrimeFaces.csp.eval(
    `
        (function(xhr, status, args, data) {
            ${script}
        })(
            window.__pfTempArgs.xhr,
            window.__pfTempArgs.status,
            window.__pfTempArgs.args,
            window.__pfTempArgs.data
        );
        delete window.__pfTempArgs;
   `);
}
```

BalusC: Indeed, see latest commit.

BalusC:
Added an IT as well.

tandraschko:
looks good for me
+1

PR #14068, opened 2025-08-11
fix(#14067): InputNumber: move autoNumeric updating from blur to change event

### LINKED ISSUE #14067: InputNumber: better support for automation tools
InputNumber is currently hard/impossible to modify with external tools that are not aware of PF/autoNumeric internals. In our case the following code is being ran:
```
let element = document.getElementById('asdf');

element.value = '100 €';
element.dispatchEvent(new Event('change'));
```

This however does not work because in the change event listener autoNumeric is queried for the current numericValue to copy it to the hidden input. autoNumeric has an internal state of the input value that is only updated when invoking `setValue`, however external automation tools are of course not aware of PF internals and how to manipulate an InputNumber

jungm:
resolves https://github.com/primefaces/primefaces/issues/14067

possibly breaks https://github.com/primefaces/primefaces/issues/6447, I wasn't able to reproduce this maybe @melloware you can test this?

### DIFF
--- a/primefaces/src/main/frontend/packages/inputnumber/inputnumber/src/inputnumber.widget.js
+++ b/primefaces/src/main/frontend/packages/inputnumber/inputnumber/src/inputnumber.widget.js
@@ -169,7 +169,7 @@ PrimeFaces.widget.InputNumber = class InputNumber extends PrimeFaces.widget.Base
                 newValue = element.getNumericString();
             } else {
                 // if using decimal places we can use the input value
-                newValue = $this.value;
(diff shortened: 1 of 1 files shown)


melloware:
Well it is sitll an open AutoNumeric issue and I know we have seen this at our clients where the browser remembers the last number you typed: https://github.com/autoNumeric/autoNumeric/issues/536

Its actually quite annoying!  But we will try to reproduce.

melloware:
I think I am good with this fix as I can't seem to reproduce the original issue

PR #14121, opened 2025-09-01
Fix #14118: Prevent unwanted change events in DatePicker with disabled animations

### LINKED ISSUE #14118: DatePicker: change event is triggered with `mask="true"` and animation disabled when focus received
### Describe the bug

When the following conditions are met, the change event is triggered when the DatePicker receives focus, even though no value has changed.
- value is empty
- `mask="true"`
- animation disabled (CSS `prefers-reduced-motion` is `reduce`)

### Reproducer

https://github.com/maruTA-bis5/primefaces-test/commit/765660e33093af202b20b175165b22cec40fc955

### Expected behavior

The change event is not triggered because the value has not changed.

### PrimeFaces edition

Community

### PrimeFaces version

15.0.7

### Theme

_No response_

### JSF implementation

All

### JSF version

Mojarra 4.0.11, Myfaces 4.0.3

### Java version

17

### Browser(s)

_No response_

maruTA-bis5:
Fixes #14118

### DIFF
--- a/primefaces/src/main/frontend/packages/datepicker/datepicker/src/jquery.ui.prime.datepicker.cjs
+++ b/primefaces/src/main/frontend/packages/datepicker/datepicker/src/jquery.ui.prime.datepicker.cjs
@@ -2737,7 +2737,7 @@ $.widget("prime.datePicker", {
             }
         }
         if (focused) {
-            focused.first().trigger('focus');
+            PrimeFaces.queueTask(function() { focused.first().trigger('focus') });
         }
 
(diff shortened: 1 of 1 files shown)


melloware:
Nice did this work locally?  If not you might want to add 1 ms like ` PrimeFaces.queueTask(function() { focused.first().trigger('focus') }, 1);`

that delays it on the UI thread vs a MicroTask.

maruTA-bis5:
It worked fine locally, but I agree that delaying on the UI thread probably makes more sense.

melloware (approved):

PR #14151, opened 2025-09-11
Fix #14150: Apply default script-src 'self' when CSP_REPORT_ONLY_POLICY is incomplete; update docs

### LINKED ISSUE #14150: CSP: documentation/implementation CSP_REPORT_ONLY_POLICY
### Description

The CSP docs file https://github.com/primefaces/primefaces/blob/15.X/docs/15_0_0/core/contentsecuritypolicy.md currently shows an example where primefaces.CSP_REPORT_ONLY_POLICY is set to a report-uri only.
That is no longer correct: primefaces.CSP_REPORT_ONLY_POLICY is now treated as a custom policy and therefore must contain the full policy directives you need — for example script-src 'self'.

The docs should be updated and we should consider changing CspPhaseListener.java to provide a sensible default (auto-include script-src 'self') when script-src and default-src directive are missing.
If a user provided only a report-uri, append a sensible default rather than setting an incomplete header.

Example currently in the docs:
```
<context-param>
    <param-value>reportOnly</param-value>
</context-param>
<context-param>
    <param-name>primefaces.CSP_REPORT_ONLY_POLICY</param-name>
    <param-value>report-uri /my-csp-reports</param-value>
</context-param>

```
This example is incorrect now. It should be:

```
<context-param>
    <param-value>reportOnly</param-value>
</context-param>
<context-param>
    <param-name>primefaces.CSP_REPORT_ONLY_POLICY</param-name>
    <param-value>report-uri /my-csp-reports; script-src 'self' </param-value>
</context-param>

```
with note: primefaces.CSP_REPORT_ONLY_POLICY is a full policy string (like primefaces.CSP_POLICY) — include the directives (e.g. script-src) you require; report-uri by itself is not sufficient as a complete policy.

DazzlingBob:
Fix #14150 

- Updated contentsecuritypolicy.md to explain that CSP_REPORT_ONLY_POLICY
  must either include a full directive (e.g. script-src, default-src, etc.)
  or will fall back to a default "script-src 'self'".
- Modified CspPhaseListener to append a default directive if missing,
  ensuring valid report-only policies.

### DIFF
--- a/docs/16_0_0/core/contentsecuritypolicy.md
+++ b/docs/16_0_0/core/contentsecuritypolicy.md
@@ -30,9 +30,17 @@ Browsers will report CSP violations to a user-defined HTTP endpoint using `repor
 </context-param>
 <context-param>
     <param-name>primefaces.CSP_REPORT_ONLY_POLICY</param-name>
-    <param-value>report-uri /my-csp-reports</param-value>
+    <param-value>report-uri /my-csp-reports; script-src 'self'</param-value>
 </context-param>
 ```
+⚠️ Note: primefaces.CSP_REPORT_ONLY_POLICY works the same way as primefaces.CSP_POLICY:
+you can provide a complete CSP policy string containing the directives you need (for example, script-src, default-src, etc.).
+If you do not provide a full policy, PrimeFaces will automatically apply a default policy of:
+
+```
+script-src 'self'
+```
+This ensures the report-only header always contains a valid directive, even if only report-uri is configured.
 
 ## Policy
 There are many ways to configure CSP for different levels of security. Currently, PrimeFaces has chosen to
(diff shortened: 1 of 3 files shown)


melloware (approved):
LGTM

docs/16_0_0/core/contentsecuritypolicy.md
```diff
@@ -30,9 +30,17 @@ Browsers will report CSP violations to a user-defined HTTP endpoint using `repor
 </context-param>
 <context-param>
     <param-name>primefaces.CSP_REPORT_ONLY_POLICY</param-name>
-    <param-value>report-uri /my-csp-reports</param-value>
+    <param-value>report-uri /my-csp-reports; script-src 'self'</param-value>
```
melloware, line 33: Can you update the 15_0_0 doc as well as I am going to backport this to 15.0.8

melloware (changes requested):

docs/16_0_0/core/contentsecuritypolicy.md
DazzlingBob: Docs are updated!

PR #14155, opened 2025-09-12
exclude classes with 3rd party dependencies from beans.xml

jungm:
TomEE logs a warning when deploying a bean that requires a class that isn't present on the classpath, thus these beans should probably excluded from scanning to avoid these warnings

Maybe could also be worth it to switch to bean discovery mode annoated and give all JSF classes that are deployed via annotations a scope? As far as I know beans.xml was only added to support org.apache.myfaces.annotation.USE_CDI_FOR_ANNOTATION_SCANNING


tandraschko:
As its currently just to support a MyFaces exclusive feature, excludes are fine.
In hope in future JSF will make `@FacesCompoent` a CDI annotation

PR #14185, opened 2025-09-19
Fix #14183: AutoComplete ARIA for moreText

### LINKED ISSUE #14183: AutoComplete: screen readers output the last menu item "MoreText" as undefined.
### Describe the bug

In the autoComplete component, the last item in the menu that indicates that there are more items is rendered as undefined by the screen readers.

### Reproducer

Go to any autoComplete component that has a maxResult attribute set to a value and that has the moreText attribute to have a value, and when read with the screen reader it will be rendered as undefined. There is also an example in the primefaces showcase.

Image

### Expected behavior

The moreText attribute value is read by the screen reader inside the autocomplete component that has an maxResults attribute with it's value set.(ex. by adding aria-label)

### PrimeFaces edition

None

### PrimeFaces version

14.0.12

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0

### Java version

17.0.9

### Browser(s)

_No response_

melloware:
Fix #14183: AutoComplete ARIA for moreText


prusuteodora (commented):
It also needs added the data-item-label in order to be read when navigating with the arrow keys. The provided fix makes the moreText red by the screen reader  only when hovering over it.

PR #14207, opened 2025-09-27
Fix #14183: AutoCompete screen reader for More Text

### LINKED ISSUE #14183: AutoComplete: screen readers output the last menu item "MoreText" as undefined.

melloware:
Fix #14183: AutoCompete screen reader for More Text

### DIFF
--- a/primefaces/src/main/frontend/packages/components/src/autocomplete/autocomplete.widget.js
+++ b/primefaces/src/main/frontend/packages/components/src/autocomplete/autocomplete.widget.js
@@ -1003,7 +1003,7 @@ PrimeFaces.widget.AutoComplete = class AutoComplete extends PrimeFaces.widget.Ba
                 });
                 if (suggestions.moreAvailable && $this.cfg.moreText) {
                     var moreTextEncoded = $("<div>").text($this.cfg.moreText).html();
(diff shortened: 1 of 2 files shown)


tandraschko:
does mit maybe make sense to have both attr?

melloware:
Actually i think i know the real issue will explain in the ticket

melloware:
It only needs the data-item-label becuase we use `active-aria-descendant` to describe which item is currently active!

melloware:
@tandraschko  the real fix was simply the MORE TExt needed an ID field so `aria-activedescendant` could point to it.

PR #14216, opened 2025-09-29
Fix 14108: Support post processors in CSV and XML exporters

### LINKED ISSUE #14108: DataExporter: Unable to post process CSVs
### Describe the bug

Upgrading from Primefaces 13 to 15, we are now unable to post process CSVs with the dataExporter because the backing document object was changed from a StringBuilder to PrintWriter.

I actually changed it from PrintWriter to StringBuilder back in 2017 with this PR: https://github.com/primefaces/primefaces/pull/2976

The use case I have in the comments are still in use by our company:

```
		if(document != null) {
			String content = document.toString();
			String newContent = StringUtils.replace(content, "\"=", "\"'=");
			document.replace(0, document.length(), newContent);
		}
```

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

15.0.6

### Theme

saga

### JSF implementation

Mojarra

### JSF version

4.0

### Java version

21

### Browser(s)

_No response_

firsthour:
Fix #14108: Support post processors in CSV and XML exporters

This brings back building CSV and XML datatable exports using a `StringBuilder` instead of a `PrintWriter`. This allows us to interrogate the file being built during or after processing properly. Currently a CSV or XML post processor is unable to modify the file in any way (other than appending text at the very end) due to it being a stream inside the `PrintWriter`.

Original PR back in the day: https://github.com/primefaces/primefaces/pull/2976

### DIFF
--- a/docs/15_0_0/components/dataexporter.md
+++ b/docs/15_0_0/components/dataexporter.md
@@ -15,7 +15,7 @@ such as excel, pdf, csv and xml.
 
 | Name | Default | Type | Description |
 | --- | --- | --- | --- |
-type | null | String | Export type: "xls", "xlsx", "xlsxstream", "pdf", "csv", "xml".
+type | null | String | Export type: "xls", "xlsx", "xlsxstream", "pdf", "csv", "csvmemory", "xml", "xmlmemory".
 target | null | String | Search expression to resolve one or multiple target components.
 fileName | null | String | Filename of the generated export file, defaults to target component id.
 pageOnly | false | Boolean | Exports only current page instead of whole dataset.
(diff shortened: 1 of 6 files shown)


tandraschko:
wasnt the PrintWriter by design for better performance?
maybe its better to provide both - or you do just your own

melloware:
Hmmm StringBuilder should be as fast as PrintWriter right?  Can we look through commit history and see who changed the code @firsthour  ?  Was it @Rapster maybe we can ask him?

tandraschko:
PrintWriter might directlry write to the output stream? StringBuilder blows up mem

firsthour:
Looks like it was changed in 252a4d8228b37daeb4a05644bf29d9ca82325127 and our CSV post processor has been silently not running since that was released... Only became an actual breaking change when generics (`<PrintWriter>`) were added to the exporters. I definitely understand `StringBuilder` will be memory heavy so I'll leave it up to you all.

More discussion in the PR: https://github.com/primefaces/primefaces/pull/6408#issuecomment-716356783

Double checked the migration guides and it seems the loss of post-processors on CSV and XML were never documented.

tandraschko:
we should definitliy leave the PrintWriter as default

maybe there is a good way to support both? Or just provide a second optional impl

tandraschko:
i think we should leave this as it is for now
but we can propably provide this as second optional impl like
InMemoryDataTableCSVExporter

firsthour:
Updated PR to introduce `InMemoryDataTableCSVExporter` and `InMemoryDataTableXMLExporter` as suggested, the new classes are exactly what my original PR had, and then a change to hookup `<p:dataExporter type="csvmemory or xmlmemory"...`

tandraschko:
can you please add a doc the classes, why there are here, compared to other exporters
and also please adjust the docs

firsthour:
Let me know your thoughts on the documentation.

tandraschko:
looks good for me

can you create the same PR for master?

firsthour:
> looks good for me
> 
> can you create the same PR for master?

https://github.com/primefaces/primefaces/pull/14269

PR #14230, opened 2025-10-05
Add scroll direction CSS classes `ui-scrollable-x` / `ui-scrollable-y` to scrollBody of Treetable and Datatable

### LINKED ISSUE #14225: TreeTable: Add scroll direction indicators (x/y) as CSS classes in updateVerticalScroll
### Description

We are currently applying a monkey patch to the updateVerticalScroll method of the TreeTable component to determine whether the scrollable body is scrollable in the horizontal (x) or vertical (y) direction.

This allows us to dynamically add or remove specific CSS classes on the scrollable body to control visual behavior such as scroll shadows, rounded corners, or padding adjustments, based on the presence of scrollbars.

The default TreeTable implementation does not provide such indicators, so we had to extend the behavior ourselves.

Here’s an anonymized example of our current patch:

```js
addMonkeyPatches: function () {
    PrimeFaces.widget.TreeTable.prototype.updateVerticalScroll = CustomPatch.TreeTable.updateVerticalScroll(
        PrimeFaces.widget.TreeTable.prototype.updateVerticalScroll
    );
},
CustomPatch: {
    TreeTable: {
        updateVerticalScroll: function (originalUpdateVerticalScroll) {
            return function () {
                originalUpdateVerticalScroll.call(this);
                const scrollBodyElement = this.scrollBody[0];
                if (scrollBodyElement.scrollWidth > scrollBodyElement.clientWidth) {
                    scrollBodyElement.classList.add("scrollable-x");
                } else {
                    scrollBodyElement.classList.remove("scrollable-x");
                }
                if (scrollBodyElement.scrollHeight > scrollBodyElement.clientHeight) {
                    scrollBodyElement.classList.add("scrollable-y");
                } else {
                    scrollBodyElement.classList.remove("scrollable-y");
                }
            };
        }
    }
}
```

We propose that the PrimeFaces TreeTable (and optionally DataTable) component:

Adds native scroll detection in the updateVerticalScroll() method.

Dynamically sets the following CSS classes on the scroll body DOM element:

p-scrollable-x if horizontal scroll is active.

p-scrollable-y if vertical scroll is active.

These classes should also be removed when no longer needed (i.e. when content shrinks back).

This would make it much easier for developers to apply custom styles based on scroll presence, without relying on custom patches or polling scroll state in user code.

### Additional context

We currently use the above patch to adjust styling in our UI when horizontal or vertical scrollbars appear. This includes things like suppressing corner rounding, adding gradient overlays, or showing sticky shadows.

Having these scroll state indicators as native classes would eliminate the need for monkey patching and ensure compatibility with future updates.

mrcasio:
Fix #14225

Adds CSS classes ui-scrollable-x and ui-scrollable-y to the scrollable body of TreeTable and DataTable to indicate horizontal and vertical scroll overflow.

These classes allow developers to apply conditional styles (e.g. shadows, padding, corner radius) based on the presence of scrollbars.

TreeTable: Added at the end of updateVerticalScroll()
DataTable: Added at the end of setupScrolling()
Classes are updated via:

```js
el.classList.toggle('ui-scrollable-x', el.scrollWidth > el.clientWidth);
el.classList.toggle('ui-scrollable-y', el.scrollHeight > el.clientHeight);
```

### DIFF
--- a/primefaces/src/main/frontend/packages/components/src/datatable/datatable.widget.js
+++ b/primefaces/src/main/frontend/packages/components/src/datatable/datatable.widget.js
@@ -1736,6 +1736,16 @@ PrimeFaces.widget.DataTable = class DataTable extends PrimeFaces.widget.Deferred
                 $this.adjustScrollWidth();
             }
         });
+
+        // Add or remove CSS classes based on scroll overflow direction
+        const el = this.scrollBody && this.scrollBody[0];
+        if (el) {
+            const hasHorizontalScroll = el.scrollWidth > el.clientWidth;
+            const hasVerticalScroll = el.scrollHeight > el.clientHeight;
+
+            el.classList.toggle('ui-scrollable-x', hasHorizontalScroll);
+            el.classList.toggle('ui-scrollable-y', hasVerticalScroll);
+        }
     }
 
     /**
--- a/primefaces/src/main/frontend/packages/components/src/treetable/treetable.widget.js
+++ b/primefaces/src/main/frontend/packages/components/src/treetable/treetable.widget.js
@@ -2537,9 +2537,10 @@ PrimeFaces.widget.TreeTable = class TreeTable extends PrimeFaces.widget.Deferred
     }
 
     /**
-     * Updates the vertical scroll position and adjusts the margin.
-     * @private
-     */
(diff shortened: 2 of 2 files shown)


melloware (approved):
LGTM to me!

PR #14240, opened 2025-10-08
Fix 14239: CSP_POLICY Expression Language support

### LINKED ISSUE #14239: CSP_POLICY: Expression Language support
### Description

We need to provide the users in our app a way to modify the CSP header.
Since we want the built-in handling of primefaces with the nonce, we activated the 
primefaces.CSP context paramter but found that the primefaces.CSP_POLICY does not support EL values like the primefaces.THEME for example does.

Here is what we would like to work

```
<context-param>
        <param-value>#{CSPPolicyBean.cspPolicy}</param-value>
 </context-param>
```

### Additional context

I have created a commit that I tested within my application that adds this support. 
The main idea was to do the same as in the HeadRenderer for the theme EL support.

Maverick-501:
Fixes #14239

### DIFF
--- a/primefaces/src/main/java/org/primefaces/csp/CspPhaseListener.java
+++ b/primefaces/src/main/java/org/primefaces/csp/CspPhaseListener.java
@@ -23,15 +23,15 @@
  */
 package org.primefaces.csp;
 
-import jakarta.el.ELContext;
-import jakarta.el.ExpressionFactory;
-import jakarta.el.ValueExpression;
 import org.primefaces.PrimeFaces;
 import org.primefaces.context.PrimeApplicationContext;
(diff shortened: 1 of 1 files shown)


melloware (approved):

tandraschko:
I remember there is a easier and better way
There is a method Like evaluateValueExpressionGet in Faces classes

melloware:
@tandraschko i was looking for a better way and noticed we do what @Maverick-501 did above in quite a few places so I assumed that is correct?  Let me look.

tandraschko:
The method im talking about doesnt need to create a Value Expression and is faster If no EL is used

melloware:
Yep found it!   PR coming...

PR #14263, opened 2025-10-15
Fix #14255: Datatable selection stop propagation on checkbox click

### LINKED ISSUE #14255: Datatable: Selection issue on expanded row
### Describe the bug

**Describe the bug:**
In my attached code I have a parent table with multiple rows (rows are expandable) when I expand it and try to select a sub-row, it doesn't get selected, 
Note: both parent table and child table should have **selectionMode="multiple"**

After further digging in the code, when we check a sub-row (row in the child table) the **SelectEvent** of the child table is fired, but the issue is that just after it an **UnselectEvent** is captured by the parent table. which is not something that should happen.

I also want to mention that onRowClick the sub-row is selected, but what we want is to select a row from its checkbox.

**Producer**: 
[primefaces-test-15.0.8.zip](https://github.com/user-attachments/files/22852444/primefaces-test-15.0.8.zip)

### Reproducer

https://github.com/user-attachments/files/22852444/primefaces-test-15.0.8.zip

### Expected behavior

Be able to select a single or multiple rows in the expanded child table.

### PrimeFaces edition

Community

### PrimeFaces version

15.0.8 (the bug is there since the release of the v12.0.0)

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.1

### Java version

17

### Browser(s)

_No response_

melloware:
Fix #14255: Datatable selection stop propagation on checkbox click

### DIFF
--- a/primefaces/src/main/frontend/packages/components/src/datatable/datatable.widget.js
+++ b/primefaces/src/main/frontend/packages/components/src/datatable/datatable.widget.js
@@ -1315,6 +1315,9 @@ PrimeFaces.widget.DataTable = class DataTable extends PrimeFaces.widget.Deferred
                     $this.selectRowWithCheckbox(checkbox, e);
                 else
                     $this.unselectRowWithCheckbox(checkbox, e);
+
+                e.preventDefault();
+                e.stopPropagation();
             });
         }
         else {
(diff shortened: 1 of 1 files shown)


tandraschko:
+1

PR #14275, opened 2025-10-16
Remove responseFlushBuffer call in BarcodeHandler 

### LINKED ISSUE #14274: Barcode: Issues with Omnifaces CompressedResponseFilter
### Describe the bug

I found a similar issue like  https://github.com/omnifaces/omnifaces/issues/336 and https://github.com/primefaces/primefaces/issues/1993 with the Barcode Component.

In case you have the CompressedResponseFilter active inside the web.xml:

	
	<filter>
		<filter-name>compressedResponseFilter</filter-name>
		<filter-class>org.omnifaces.filter.CompressedResponseFilter</filter-class>
	</filter>
	<filter-mapping>
		<filter-name>compressedResponseFilter</filter-name>
		<servlet-name>facesServlet</servlet-name>
		<dispatcher>REQUEST</dispatcher>
		<dispatcher>ERROR</dispatcher>
	</filter-mapping>

You get every time, when you render a Barcode the following message:
[WARNING] handleException /jakarta.faces.resource/dynamiccontent.properties.xhtml java.io.IOException: Stream is already closed.

(A reproducer is attached to the ticket) 

Last time the solution was to remove externalContext.responseFlushBuffer(); from the handle code.

I romved then the call inside org.primefaces.component.barcode.BarcodeHandler.handle(FacesContext context) and the message disappeared in the logs.

PR is also submited.

### Reproducer

[primefaces-test.zip](https://github.com/user-attachments/files/22948778/primefaces-test.zip)

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

15.0.8

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.1.0

### Java version

21

### Browser(s)

_No response_

Teutron-flo:
Fix #14274

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/barcode/BarcodeHandler.java
+++ b/primefaces/src/main/java/org/primefaces/component/barcode/BarcodeHandler.java
@@ -174,7 +174,6 @@ public class BarcodeHandler extends BaseDynamicContentHandler {
             LOGGER.log(Level.SEVERE, "Error in streaming barcode resource. {0}", new Object[]{e.getMessage()});
         }
         finally {
-            externalContext.responseFlushBuffer();
             context.responseComplete();
         }
     }


tandraschko:
we have it in different locations, not sure if we should apply it

melloware:
@Teutron-flo  maybe ask @BalusC I think in our case we should be closing it?

BalusC:
BarcodeHandler with format=svg (default) invokes SvgRenderer#render which closes the stream (close already does flush).
Running externalContext.responseFlushBuffer() afterwards should not be needed.
See also https://github.com/omnifaces/omnifaces/issues/912

Most probably also not needed for ImageIO.write() which is used for format=png but I haven't debugged it.

Alternative is to put this responseFlushBuffer in try with empty catch or a FINE logger.

This existing bug is apparently only visible with CompressedResponseFilter because it explicitly throws exception when flush is called on closed stream in order to make the developer aware. I guess the original/underlying stream didn't do that.

melloware:
OK after reviewing the old Barcode4J did not close the streams but Okapi and ImageIO do. So this PR is correct we do not need to do this.

PR #14281, opened 2025-10-19
Fix #14264: TriStateCheckBox documentation updates

### LINKED ISSUE #14264: TriStateCheckbox: Update Documentation
Reported in discussions: https://github.com/orgs/primefaces/discussions/4392

by @tsotzolas

PF 15 and 16.

it takes a object by docs but the value must lead to a boolean. Currently we support String and bool/Boolean, String in that case is "true" or "false", we parse that automatically.

melloware:
Fix #14264: TriStateCheckBox documentation updates


tandraschko:
TBH i wouldnt document the string support, Boolean should be the official usage only.

melloware:
Got it let me remove.

PR #14282, opened 2025-10-19
Fix #14280: Chips separator character handling

### LINKED ISSUE #14280: Chips: item is not added when typing separator char
### Description

https://showcase.primefaces.org/ui/input/chips.xhtml

Type something in "Paste" example (as it is the only example using `separator` attribute) and then type comma. You'd expect that chip is added. This did not happen.

A new [switch case](https://github.com/primefaces/primefaces/blob/v15.0.8/primefaces/src/main/resources/META-INF/resources/primefaces/chips/chips.js#L88-L110) to `keydown.chips` in associated JS which checks `this.cfg.separator`.

We should probably also document in vdl and validate in renderer that the `separator` attribute can accept only 1 character.

### Additional context

Work around:
```
<p:chips ... separator=" " onkeydown="if(event.key==' '){event.preventDefault();event.target.dispatchEvent(new KeyboardEvent('keydown',{key:'Enter'}));}" />
```

melloware:
Fix #14280: Chips separator character handling


melloware:
@BalusC please review

PR #14311, opened 2025-10-28
Sticky: add new stickyTopAt property

### LINKED ISSUE #14022: Sticky: support stickyTopAt property
### Description

Currently sticky uses the `margin` property to control positioning. Supporting `stickyTopAt` property like DataTable would be a nice addition.

Just like DataTable.

### Additional context

I may take a look at this as soon as I have some time

NicolaIsotta:
fixes #14022 

still draft cause I want to improve resizing behavior

### DIFF
--- a/docs/16_0_0/components/sticky.md
+++ b/docs/16_0_0/components/sticky.md
@@ -23,8 +23,9 @@ viewport during scrolling.
 id | null | String | Unique identifier of the component.
 rendered | true | Boolean | Boolean value to specify the rendering of the component, when set to false component will not be rendered.
 binding | null | Object | An el expression that maps to a server side UIComponent instance in a backing bean.
-target | null | String | Component to make sticky.
 margin | 0 | Integer | Margin to the top of the page during fixed scrolling.
+stickyTopAt | null | String | Selector for elements fixed at the top of the page whose height should be considered when positioning the sticky element.
+target | null | String | Component to make sticky.
 
 ## Getting started with Sticky
 Sticky requires a target to keep in viewport on scroll. Here is a sticky toolbar;
--- a/docs/16_0_0/gettingstarted/whatsnew.md
+++ b/docs/16_0_0/gettingstarted/whatsnew.md
@@ -10,5 +10,7 @@ Look into [migration guide](https://primefaces.github.io/primefaces/16_0_0/#/../
 
 * Spinner
     * Added `dir="rtl"` right to left support.
+* Sticky
+    * Added `stickyTopAt` attribute for elements fixed at the top of the page whose height should be considered when positioning the sticky element.
 * TreeTable
   * Added `resizeMode` attribute to support column resizing using the "expand" mode.
--- a/primefaces-showcase/src/main/webapp/ui/misc/sticky.xhtml
+++ b/primefaces-showcase/src/main/webapp/ui/misc/sticky.xhtml
@@ -14,20 +14,6 @@
         Scroll down to see the sticky toolbar.
     </ui:define>
 
-    <ui:define name="head">
-        <style>
-            .ui-sticky {
-                top: 70px !important;
-            }
-
-            @media (max-width: 960px) {
-                .ui-sticky {
-                    top: 110px !important;
-                }
-            }
-        </style>
-    </ui:define>
-
     <ui:param name="documentationLink" value="/components/sticky"/>
     <ui:param name="widgetLink" value="Sticky-1"/>
 
@@ -47,7 +33,7 @@
 
             <div style="height:800px"></div>
 
-            <p:sticky target="tb"/>
+            <p:sticky target="tb" stickyTopAt=".layout-topbar"/>
         </div>
     </ui:define>
 
--- a/primefaces/src/main/frontend/packages/components/src/sticky/sticky.widget.js
+++ b/primefaces/src/main/frontend/packages/components/src/sticky/sticky.widget.js
@@ -21,6 +21,8 @@
  * @extends {PrimeFaces.widget.BaseWidgetCfg} cfg
  * 
  * @prop {number} cfg.margin Margin to the top of the page during fixed scrolling.
+ * @prop {string} cfg.stickyTopAt Selector for elements fixed at the top of the page whose height should be considered
+ * when positioning the sticky element.
  * @prop {string} cfg.target The client ID of the component to be made sticky.
  */
 PrimeFaces.widget.Sticky = class Sticky extends PrimeFaces.widget.BaseWidget {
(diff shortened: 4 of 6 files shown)


docs/16_0_0/components/sticky.md
```diff
@@ -25,6 +25,7 @@ rendered | true | Boolean | Boolean value to specify the rendering of the compon
 binding | null | Object | An el expression that maps to a server side UIComponent instance in a backing bean.
 target | null | String | Component to make sticky.
 margin | 0 | Integer | Margin to the top of the page during fixed scrolling.
+stickyTopAt | null | String | Selector for elements fixed at the top of the page whose height should be considered when positioning the sticky element. Default is null.
```
melloware: you can move this to be in alpha order next to the other props

melloware (changes requested):

NicolaIsotta:
I've discovered that even the current version is bugged when resizing - harder to fix than I thought, but I'm getting closer I hope

NicolaIsotta:
Okay, now I like it. It seems pretty robust. I may backport the resizing fix to 15.X

melloware:
Thanks — here's a focused code review of the changes in primefaces/src/main/frontend/packages/components/src/sticky/sticky.widget.js.

Summary
- The change adds support for accounting for elements fixed at the top of the page (cfg.stickyTopAt) when computing the sticky offset/margin.
- A new method calculateDynamicMargin() is introduced and the resize handler uses its result to decide whether to recreate the ghost element and re-fix the sticky element.
- A bug is present: init() calls calculateDynamicOffset(), but the new method is named calculateDynamicMargin(), so the function call will fail (ReferenceError) unless calculateDynamicOffset exists elsewhere.

What looks good
- The feature to consider heights of top fixed elements is useful in layouts with multiple fixed headers/toolbars.
- The resize handler attempts to detect margin changes and reapply positioning when needed which is the right idea.
- The method returns a boolean (previousMargin !== this.cfg.margin) so resize logic can react only when needed.

Bugs / correctness issues
1. Name mismatch (critical)
   - In init() the code calls this.calculateDynamicOffset(); but the new function is calculateDynamicMargin(). This will throw at runtime unless calculateDynamicOffset is defined elsewhere. Likely intended to call calculateDynamicMargin().

2. Overwrites configured margin and ignores an explicit base margin
   - calculateDynamicMargin sets this.cfg.margin to the sum of the heights of stickyTopAt elements, replacing any user-provided cfg.margin.
   - If the intent is to add the fixed elements' heights on top of an explicit cfg.margin, the code should preserve the original margin and add the computed top elements total to that base margin (or store an internal baseMargin).

3. Null/empty handling and type expectations
   - The code sets fixedElementsOnTop = this.cfg.stickyTopAt ? $(this.cfg.stickyTopAt) : null; then checks fixedElementsOnTop && fixedElementsOnTop.length. Prefer always returning an empty jQuery collection instead of null (simpler to work with).
   - If stickyTopAt is a selector that matches nothing, current logic leaves cfg.margin unchanged — that may be fine, but should be explicit.

4. Use of var and globals
   - The resize handler uses var marginChanged = $this.calculateDynamicMargin(); — using let/const would be preferred for modern code style consistency. The code otherwise uses var in places.

5. Potential double computation and layout thrashing
   - There are multiple offset/height calls (ghost.offset(), target.offset(), target.width/outerWidth, etc.) inside the resize handler and fix/restore logic. Ensure they are minimized and measurements cached where sensible to avoid layout thrashing on resize.

6. Missing JSDoc and behaviors
   - The new cfg.stickyTopAt is documented, good — consider documenting whether its heights are added to or replace cfg.margin.
   - calculateDynamicMargin JSDoc says it returns boolean but does not document side effects (overwriting cfg.margin).

7. Edge cases
   - If stickyTopAt elements change height but selector is the same, the method will adjust — good.
   - If stickyTopAt elements are hidden/display:none, outerHeight() will be 0; if elements include margins, outerHeight without true parameter does not include margins. If margins matter consider outerHeight(true).

Suggestions / recommended fixes
- Fix the function call in init():
  - Replace this.calculateDynamicOffset(); with this.calculateDynamicMargin();

- Preserve an original/base margin:
  - In init() set this._baseMargin = this.cfg.margin || 0;
  - In calculateDynamicMargin compute totalTopHeight, then set this.cfg.margin = this._baseMargin + totalTopHeight;
  - This preserves any configured margin while adding dynamic heights.

- Use empty jQuery collection instead of null:
  - const fixedElementsOnTop = this.cfg.stickyTopAt ? $(this.cfg.stickyTopAt) : $();

- Use modern let/const and prefer ===:
  - const previousMargin = this.cfg.margin;
  - return previousMargin !== this.cfg.margin; (fine) but use const/let throughout.

- Consider outerHeight(true) if margins should be included:
  - fixedElementsOnTop.toArray().reduce((height, elem) => height + $(elem).outerHeight(true), 0);

- Minor robustness:
  - Guard against exceptions if stickyTopAt is not a selector string (maybe accept element/jQuery directly).
  - If multiple elements may overlap, ensure double-counting isn't an issue (usually not).

- Tests and docs:
  - Add a unit/integration test that uses a page with an existing fixed header, verifies the sticky element positions correctly after resize and scroll.
  - Update the PrimeFaces docs or component configuration docs to explain cfg.stickyTopAt behavior and that it is additive to cfg.margin.

Suggested patch sketch (core parts)
- In init():
  - this.cfg.margin = this.cfg.margin || 0;
  - this.cfg.stickyTopAt = this.cfg.stickyTopAt || null;
  - this._baseMargin = this.cfg.margin; // store original
  - this.calculateDynamicMargin();

- calculateDynamicMargin():
  - const fixed = this.cfg.stickyTopAt ? $(this.cfg.stickyTopAt) : $();
  - const previousMargin = this.cfg.margin;
  - const topHeight = fixed.length ? fixed.toArray().reduce((sum, el) => sum + $(el).outerHeight(true), 0) : 0;
  - this.cfg.margin = this._baseMargin + topHeight;
  - return previousMargin !== this.cfg.margin;

Other notes
- The resize handler logic that removes the ghost and calls fix(true) when the margin changed looks reasonable, but ensure fix(true) signature actually causes re-creation/recompute of the ghost/position. Confirm the implementation of fix() so removing ghost + fix(true) leads to correct DOM state.


Conclusion
- The idea and general approach are good and address a real need. However the init() -> calculateDynamicOffset() call is a blocking bug and must be fixed. Also preserve the configured margin by storing a base margin and adding the computed heights rather than overwriting user config. Finally, prefer const/let, handle empty selector cases consistently, and add tests/documentation.

NicolaIsotta:
Good catch - I forgot to rename the first call of the function

melloware:
Thanks to Copilot!

melloware (approved):

PR #14338, opened 2025-11-07
Fix #13896: Add ariaDescribedBy to input components

### LINKED ISSUE #13896: Accessibility: add attribute for aria-describedby for validate purpose
Hi, 

this case is rather complicated. I hope I can explain this properly.

**first case**:
code from [showcase](http://www.primefaces.org/showcase/ui/input/inputText.xhtml?jfwid=19660)
sorry, I don't know how to properly format the code

```xml
<div class="field">
        <p:outputLabel for="username2" styleClass="block">Username</p:outputLabel>
        <p:inputText id="username2" styleClass="block ui-state-error"/>
        <small class="block ui-state-error">Username is not available.</small>
    </div>
```

Image

this create a simple input field with a validation error message. However, when using keyboard for navigation and tab into the field, the error message will not be read out. This can be achieved by adding `aria-describedby` to the input, and since `aria-describedby` is not supported in inputText, I have to use passthrough:

```xml
<div class="field">
        <p:outputLabel for="username2" styleClass="block">Username</p:outputLabel>
        <p:inputText id="username2" styleClass="block ui-state-error" pt:aria-describedby="errorUsername"/>
        <small id="errorUsername" class="block ui-state-error">Username is not available.</small>
    </div>
```

this works fine. However, with this only usable with passthrough, more problems occur:

**second case**
When the input field is in a form (which they mostly are), referencing to a id become more complicated:

```xml
<h:form id="personalForm">
        <div class="field">
        <p:outputLabel for="username2" styleClass="block">Username</p:outputLabel>
        <p:inputText id="username2" styleClass="block ui-state-error" pt:aria-describedby="errorUsername"/>
        <small id="errorUsername" class="block ui-state-error">Username is not available.</small>
    </div>
    </h:form>
```

thanks to the auto-generated id, `pt:aria-describedby="errorUsername` won't work anymore, you need `pt:aria-describedby="personalForm:errorUsername`. And when form get more complex, the id will get longer.

**third case**
In my case, I have a form in which there is a list of dynamically generated input fields, for this I use <ui:repeat id="names">. In the generated HTML output, the error message container will get the `id="personalForm:names:0:errorUsername"`. So now, I can't reference to this id at all, right? Since pt only pass the whole thing through as string, something like `pt:aria-describedby="personalForm:names:{index}:errorUsername"` will not work.

```xml
<h:form id="personalForm">
        <div class="form-fields">
            <ui:repeat value="#{testView.list}" var="item" id="names">
                <div class="field-container">
                    <span class="ui-float-label">
                        <p:inputText value="#{item.name}"
                                     pt:aria-describedby="personalForm:names:{index}:errorUsername">
                        </p:inputText>
                        <p:outputLabel for="@previous" value="Name" />
                    </span>
                </div>
                <p:fragment id="errorUsername">
                    <small class="block ui-state-error">Username is not available.</small>
                </p:fragment>
            </ui:repeat>
        </div>
    </h:form>
```

So, the "most simple" way will be add a new attribute for `aria-describedby` (I think in primevue this exists already?). I know it is probably not that simple, because inputText is only one example, it should exits in all form control. 
In the meantime, is there any other way to reference to a generated id in a <ui:repeat>?

Thanks for your work.

melloware:
Fix #13896: Add ariaDescribedBy to input components

### DIFF
--- a/docs/16_0_0/components/autocomplete.md
+++ b/docs/16_0_0/components/autocomplete.md
@@ -23,6 +23,7 @@ AutoComplete provides live suggestions while an input is being typed.
 | accesskey | null | String | Access key that when pressed transfers focus to the input element.
 | active | true | Boolean | Defines if autocomplete functionality is enabled.
 | alt | null | String | Alternate textual description of the input field.
+| ariaDescribedBy | null | String | The aria-describedby attribute is used to define a component id that describes the current element for accessibility.
 | appendTo | null | String | Appends the overlay to the element defined by search expression. Defaults to document body.
 | at | left bottom | String | Position of input with respect to panel.
 | autocomplete | null | String | Controls browser autocomplete behavior.
--- a/docs/16_0_0/components/calendar.md
+++ b/docs/16_0_0/components/calendar.md
@@ -26,6 +26,7 @@ ajax selection and more.
 | value | null | java.time.LocalDate, java.time.LocalDateTime, java.time.LocalTime, java.util.Date (deprecated) | Value of the component
 | accesskey | null | String | Access key that when pressed transfers focus to the input element.
 | alt | null | String | Alternate textual description of the input field.
+| ariaDescribedBy | null | String | The aria-describedby attribute is used to define a component id that describes the current element for accessibility.
 | autocomplete | null | String | Controls browser autocomplete behavior.
 | beforeShow | null | String | Callback to execute before displaying calendar, element and calendar instance are passed as parameters
 | beforeShowDay | null | String | Client side callback to execute before displaying a date, used to customize date display.
--- a/docs/16_0_0/components/cascadeselect.md
+++ b/docs/16_0_0/components/cascadeselect.md
@@ -20,25 +20,26 @@ CascadeSelect displays a nested structure of options.
 | Name | Default | Type | Description | 
 | --- | --- | --- | --- |
 id | null | String | Unique identifier of the component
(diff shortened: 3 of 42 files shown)


tandraschko:
this change destroys my applications:

```
2025-11-19 15:56:34 SEVERE (executor-thread-11) Cannot find component for expression "form:crud:form:groups:0:fields:1:j_id_4v" referenced from "form:crud:form:groups:0:fields:1:inputSelectOne".: jakarta.faces.component.search.ComponentNotFoundException: Cannot find component for expression "form:crud:form:groups:0:fields:1:j_id_4v" referenced from "form:crud:form:groups:0:fields:1:inputSelectOne".
	at org.apache.myfaces.component.search.SearchExpressionHandlerImpl.resolveComponent(SearchExpressionHandlerImpl.java:255)
	at org.primefaces.expression.SearchExpressionUtils.contextlessResolveComponent(SearchExpressionUtils.java:91)
	at org.primefaces.expression.SearchExpressionUtils.contextlessResolveComponent(SearchExpressionUtils.java:79)
	at org.primefaces.renderkit.InputRenderer.renderAccessibilityAttributes(InputRenderer.java:149)
	at org.primefaces.renderkit.InputRenderer.renderAccessibilityAttributes(InputRenderer.java:117)```

melloware:
interesting i tested using Showcase and PrimeFaces Test and the Integration Tests all pass?

PR #14355, opened 2025-11-11
Fix #14353: Picklist remove aria-required

### LINKED ISSUE #14353: pickList: required="true" leads to an a11y problem regarding to unallowed aria-required="true" without role assigned for ui-picklist-list-wrapper div
### Describe the bug

If the required attribute is set for the picklist e.g.
`<p:pickList id="pickListRequired" value="#{pickListView.cities}" var="cities" itemLabel="#{cities}" itemValue="#{cities}" required="true"/>`

the rendered result contains aria-required on a div without a role defined.
`<div class="ui-picklist-list-wrapper" aria-required="true">`

Assistive technologies like NVDA screenreader can not valid consume aria-required on the div without a role attribute assigned additionally and it also breaks the a11y accessibility check rule ID `aria-allowed-attr`: "[Elements must only use supported ARIA attributes](https://dequeuniversity.com/rules/axe/4.10/aria-allowed-attr?application=AxeChrome)" 

### Reproducer

`<p:pickList id="pickListRequired" value="#{pickListView.cities}" var="cities" itemLabel="#{cities}" itemValue="#{cities}" required="true"/>`

leads to

`<div class="ui-picklist-list-wrapper" aria-required="true">`
...
whole content for the target picklist:

`<div class="ui-picklist-list-wrapper" aria-required="true"><ul class="ui-widget-content ui-picklist-list ui-picklist-target ui-sortable" tabindex="-1"></ul><select id="j_idt142:pickListRequired_target" name="j_idt142:pickListRequired_target" multiple="multiple" class="ui-helper-hidden"></select></div>`

### Expected behavior

`aria-required="true"` attribute should only be applied to elements that have a role like "listbox" or "menu". A div without an explicit role specification here is not useful for assistive technologies, so aria-required has no effect there.

### PrimeFaces edition

None

### PrimeFaces version

14.0.12, 16.0.0-SNAPSHOT

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0

### Java version

17.0.6

### Browser(s)

_No response_

melloware:
Fix #14353: Picklist remove aria-required

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/picklist/PickListRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/picklist/PickListRenderer.java
@@ -219,11 +219,6 @@ public class PickListRenderer extends InputRenderer<PickList> {
         writer.startElement("div", null);
         writer.writeAttribute("class", PickList.LIST_WRAPPER_CLASS, null);
 
-        // only render required on target list
-        if (!isSource) {
-            renderARIARequired(context, component);
-        }
-
         if (filter) {
             encodeFilter(context, component, listId + "_filter", isSource);
         }


melloware:
@rho66 we added this back in 2019 as part of #4047 but this Picklist should not have gotten this Aria Required

PR #14361, opened 2025-11-12
added CDK

tandraschko:
the CDK contains of some API classes and a taglib generated (via maven) + subsclass generator (AnnotationProcessor)

this shows how to use it for the AjaxBehavior, Badge and Panel

please review ;D

basically it generates all the boilerplate code...

whats open:
- refactor all components to use it
- check out if we need something more (propably validator/convert taglib generation)
- activate taglib generation
- remove some code in PF, which is now unused during this change

IMO this tasks can all be done in new tasks, aftering mergin this

@melloware @christophs78

### DIFF
--- a/primefaces/src/main/java/org/primefaces/behavior/ajax/AjaxBehaviorBase.java
+++ b/primefaces/src/main/java/org/primefaces/behavior/ajax/AjaxBehaviorBase.java
@@ -1,0 +1,92 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2025 PrimeTek Informatics
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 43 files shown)


melloware:
this is awesome!  But Panel IT's are failing?   We can definitely submit PR's to fix one component at a time after this PR!

tandraschko:
wrong event name, fixed it now

melloware (approved):

primefaces/src/main/java/org/primefaces/behavior/ajax/AjaxBehaviorBase.java
```diff
@@ -0,0 +1,92 @@
...
+package org.primefaces.behavior.ajax;
+
+import org.primefaces.cdk.api.Property;
+import org.primefaces.cdk.api.behavior.PrimeClientBehavior;
+
+import jakarta.el.MethodExpression;
+
+public abstract class AjaxBehaviorBase extends PrimeClientBehavior {
+
+    @Property
```
melloware, line 33: Do we need to add descriptions here?  If so I can add them after merging...

tandraschko: sure
i just was to lazy

melloware: No worries you have enough on your plate!

christophs78 (approved):
Great to see you working on this! Should be definitive an improvement.
As a detail - does it make sense to add a short README.md to the new module?

tandraschko:
Sure it does, will do it when all Features are used in PF

tandraschko:
added some basic docs

feel free to merge it if you like it, im quite happy with it for now

melloware:
OK I will merge and try my hand and converting a component!

PR #14373, opened 2025-11-17
CDK: Update Behaviors

melloware:
@tandraschko i updated the CDK to spit out the generated props and facets in alphabetical order.

I then updated the docs for AjaxBehavior and converted the other behaviors to similar pattern


primefaces/src/main/java/org/primefaces/behavior/printer/PrinterBehaviorHandler.java
```diff
@@ -0,0 +1,41 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.behavior.printer;
+
+import org.primefaces.cdk.api.behavior.PrimeClientBehaviorHandler;
+
+import jakarta.faces.view.facelets.BehaviorConfig;
+
```
tandraschko: i dont think we need it as it was not required before?

primefaces/src/main/java/org/primefaces/behavior/validate/ClientValidatorBehaviorHandler.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.primefaces.cdk.api.behavior.PrimeClientBehaviorHandler;
+
+import jakarta.faces.view.facelets.BehaviorConfig;
+
+public class ClientValidatorBehaviorHandler extends PrimeClientBehaviorHandler<ClientValidatorBehavior> {
+
+    public ClientValidatorBehaviorHandler(BehaviorConfig config) {
+        super(config);
+    }
```
tandraschko: same here, actually not necessary

melloware: OK wasn't sure if we should be consistent i will remove.

melloware:
OK i was also able to delete AbstractBehavior and Handler as now they are no longer in use

tandraschko:
+1
lookgs good for me then

PR #14396, opened 2025-11-21
Fix #14390: BooleanAwareResponswriter for HTML5 boolean attributes

### LINKED ISSUE #14390: Video: The `controls` attribute does not produce the expected effect
When I read controls javadoc for <p:video> tag, my understanding is that, it is of type java.lang.Boolean and it is supposed to control whether the audio/video should display controls (like play/pause etc.).  However, whether I set the attribute to true or false, the result is the same, but when I remove it, all the controls disappear. If this attribute is not meant to disable the controls or make them invisible, which other attribute can I use to achieve this goal?
Thanks.

This is the contains of my test.xhtml page
PrimeFaces 15.0.10
Mojarra 4.0.11
jetty:12:0.15 / tomcat 11.0.11

```xml
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:ui="jakarta.faces.facelets" xmlns:f="jakarta.faces.core"
    xmlns:h="jakarta.faces.html" xmlns:p="primefaces">

<h:head>
    <title>PrimeFaces Test - Good</title>
</h:head>
<h:body>

    <h:form id="frmTest">
        <p:outputLabel value="Without controls" />
        <p:video height="320" value="/resources/video/letGraphics.mp4">
            Your browser does not support the video tag.
        </p:video>
        <p:outputLabel value="control=false" />
        <p:video height="320" value="/resources/video/letGraphics.mp4" controls="false">
            Your browser does not support the video tag.
        </p:video>
        <p:outputLabel value="control=true" />
        <p:video height="320" value="/resources/video/letGraphics.mp4" controls="true">
            Your browser does not support the video tag.
        </p:video>
        <p:outputLabel value="controls=false, controlslist=nodownload" />
        <p:video height="320" value="/resources/video/letGraphics.mp4"  controls="false" controlslist="nodownload">
            Your browser does not support the video tag.
        </p:video>
        <p:outputLabel value="controls=true, controlslist=nodownload" />
        <p:video height="320" value="/resources/video/letGraphics.mp4"  controls="true" controlslist="nodownload">
            Your browser does not support the video tag.
        </p:video>
    </h:form>

</h:body>

</html>
```

```xml
<img width="533" height="598" alt="Image" src="https://github.com/user-attachments/assets/dec9275a-aa6c-492d-a402-81406285e5ba" />
```

[primefaces-test.zip](https://github.com/user-attachments/files/23661424/primefaces-test.zip)

melloware:
Fix #14390: BooleanAwareResponswriter for HTML5 boolean attributes

## Overview

`BooleanAwareResponseWriter` is a decorator class that extends `ResponseWriterWrapper` to automatically handle HTML5 boolean attributes according to the HTML5 specification. It ensures that boolean attributes are written in the correct format without requiring manual handling in individual renderers.

## What It Does

The class intercepts calls to `writeAttribute()` and applies special handling for HTML5 boolean attributes:

- **When value is `true`**: Writes the attribute in "naked" form (e.g., `<input disabled>`)
- **When value is `false`**: Omits the attribute entirely (e.g., `<input>` instead of `<input disabled="false">`)
- **When value is `null` or empty**: Treats as `false` and omits the attribute entirely

### Supported Boolean Attributes

The class recognizes 24 HTML5 boolean attributes:
- `allowfullscreen`, `async`, `autofocus`, `autoplay`
- `checked`, `controls`, `default`, `defer`
- `disabled`, `formnovalidate`, `hidden`, `ismap`
- `itemscope`, `loop`, `multiple`, `muted`
- `nomodule`, `novalidate`, `open`, `playsinline`
- `readonly`, `required`, `reversed`, `selected`

## How It Works

1. **Interception**: Overrides `writeAttribute(String name, Object value, String property)` to intercept all attribute writes
2. **Detection**: Checks if the attribute name (case-insensitive) is in the set of known boolean attributes
3. **Value Processing**:
   - If `Boolean.FALSE` → omits the attribute
   - If string `"false"` (case-insensitive) → omits the attribute
   - Otherwise (including `Boolean.TRUE`, `"true"`, `disabled`,`empty string) → writes as `name=""` (valid HTML5 format)
4. **Delegation**: Non-boolean attributes are passed through unchanged to the wrapped `ResponseWriter`

## Integration

The class is automatically applied in `PrimeFacesContext.setResponseWriter()`:
- It wraps the `ResponseWriter` **first**, before other decorators (CSP, MoveScriptsToBottom)
- It checks for existing wrappers to avoid double-wrapping
- It's applied globally to all PrimeFaces rendering operations

### Before BooleanAwareResponseWriter

**Problems:**
1. **Inconsistent Output**: Different renderers handled boolean attributes differently
   - Some wrote `disabled="true"` (invalid HTML5)
   - Some wrote `disabled="disabled"` (XHTML style)
   - Some wrote `disabled=""` (correct but inconsistent)
   - Some forgot to omit attributes when `false`

2. **Code Duplication**: Each renderer had to manually check and format boolean attributes:
   ```java
   // Example of manual handling (repeated in many renderers)
   if (component.isDisabled()) {
       writer.writeAttribute("disabled", "", null);
   }
   // Or worse:
   writer.writeAttribute("disabled", component.isDisabled() ? "disabled" : null, null);
   ```

3. **Maintenance Burden**: Adding new boolean attributes or fixing formatting required changes across multiple renderer classes

4. **HTML5 Compliance**: Output didn't always conform to HTML5 specification, which could cause:
   - Validation errors
   - Unexpected browser behavior
   - Accessibility issues

### After BooleanAwareResponseWriter

**Benefits:**
1. **Centralized Logic**: Single point of control for all boolean attribute handling
2. **Automatic Compliance**: All renderers automatically produce HTML5-compliant output
3. **Simplified Renderers**: Renderers can simply call `writeAttribute()` with boolean values without special handling:
   ```java
   // Now renderers can simply write:
   writer.writeAttribute("disabled", component.isDisabled(), null);
   // The decorator handles the formatting automatically
   ```

4. **Consistency**: All components produce consistent HTML output across the entire framework
5. **Future-Proof**: New boolean attributes can be added to the set without touching individual renderers
6. **Correct Behavior**: 
   - `false` values are properly omitted (not written as `disabled="false"`)
   - `true` values are written in the correct HTML5 format
   - Case-insensitive string handling for flexibility

## Technical Details

- **Implementation Pattern**: Decorator pattern using `ResponseWriterWrapper`
- **Performance**: Uses a `Set<String>` for O(1) lookup of boolean attributes
- **Case Handling**: Case-insensitive attribute name matching
- **Value Handling**: Supports both `Boolean` objects and string representations
- **JSF Integration**: Uses `writeAttribute()` with empty string to ensure proper JSF buffering and avoid timing issues

## Example Output

**Before:**
```html
<input disabled="true" />
<input disabled="false" />
<input disabled="disabled" />
```

**After:**
```html
<input disabled />
<input />
<input disabled />
```

This ensures all boolean attributes follow the HTML5 specification where the presence of the attribute indicates `true` and its absence indicates `false`.

Here is video with `controls="true"` and `controls="false"` now

image

### DIFF
--- a/primefaces-showcase/src/main/webapp/ui/multimedia/video.xhtml
+++ b/primefaces-showcase/src/main/webapp/ui/multimedia/video.xhtml
@@ -36,7 +36,7 @@
         <div class="card">
             <h5>WEBM</h5>
             <p:video value="/resources/demo/media/sample-webm.webm"
-                     controls="true"
+                     controls="false"
                      width="420"
                      height="250"
                      onplay="console.log('WEBM Started Playing')"
--- a/primefaces/src/main/java/org/primefaces/application/resource/BooleanAwareResponseWriter.java
+++ b/primefaces/src/main/java/org/primefaces/application/resource/BooleanAwareResponseWriter.java
@@ -1,0 +1,111 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2025 PrimeTek Informatics
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.application.resource;
+
+import org.primefaces.util.Constants;
+
+import java.io.IOException;
+import java.util.Set;
+
+import jakarta.faces.context.ResponseWriter;
+import jakarta.faces.context.ResponseWriterWrapper;
+
+/**
+ * A ResponseWriter decorator that automatically handles HTML5 boolean attributes.
+ * Boolean attributes are written in naked form (e.g., "disabled") when their value
+ * is "true" or Boolean.TRUE. If the value is "false", they are omitted entirely.
+ */
+public class BooleanAwareResponseWriter extends ResponseWriterWrapper {
+
+    // -------------------------------------------------------------------------
+    //  HTML5 Boolean Attributes (complete, spec-correct)
+    // -------------------------------------------------------------------------
+    private static final Set<String> BOOLEAN_ATTRS = Set.of(
+            "allowfullscreen",
+            "async",
+            "autofocus",
+            "autoplay",
+            "checked",
+            "controls",
+            "default",
+            "defer",
+            "disabled",
+            "formnovalidate",
+            "hidden",
+            "ismap",
+            "itemscope",
+            "loop",
+            "multiple",
+            "muted",
+            "nomodule",
+            "novalidate",
+            "open",
+            "playsinline",
+            "readonly",
+            "required",
+            "reversed",
+            "selected"
+    );
+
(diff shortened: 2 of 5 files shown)


tandraschko:
is
disabled="true" (invalid HTML5)
really invalid?

second thing, we should only apply it if the template is HTML5 markup, not for XHTML

melloware:
Great question so its a browser thing. Most browsers don't care `disabled="disabled"`  but according to the spec it should be `disabled` or `disabled=""`.  For `controls` the browsers seem to be strict about these HTML5 boolean props!

tandraschko:
so controls="" doesnt work?

melloware:
`controls=""` does work and that is what this ResponsWriter generates for all HTML5 boolean props.  So before we were generting

```
disabled="disabled"
controls="true"
```

now gets generated as 

```
disabled=""
controls=""
```

Which the browser is happy with always.

The browser considers `controls=""` and `controls` as the same.

melloware:
A couple of other notes...

1. ~Its broken on MyFaces but not Mojarra i need to look at why.~

2. Instead of writing the following everywhere

```java
if (disabled) {
      writer.writeAttribute("disabled", "disabled", null);
}
```

we can just write

```java
 writer.writeAttribute("disabled", disabled, null);
```

melloware:
@tandraschko i can close this because your CDK now that we define properties is not creatign `controls="false"` for passthrough because i am declaring the property as a CDK property proper boolean

PR #14412, opened 2025-11-23
CDK: Accordion, Carousel, Galleria, Tabview

melloware:
@tandraschko  sorry this is so big i started just doing Accordion but that led to UiTabPanel which lead to Galleria, Carousel, Tabview.

- [x] RTLAware is now similar to Widget interface
- [x] TouchAware is now similar to Widget interface
- [x] Tab converted
- [x] UITabPanel converted
- [x] Accordion
- [x] Carousel
- [x] Galleria
- [x] TabView

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/accordionpanel/AccordionPanel.java
+++ b/primefaces/src/main/java/org/primefaces/component/accordionpanel/AccordionPanel.java
@@ -29,11 +29,8 @@ import org.primefaces.el.ValueExpressionAnalyzer;
 import org.primefaces.event.TabChangeEvent;
 import org.primefaces.event.TabCloseEvent;
 import org.primefaces.event.TabEvent;
-import org.primefaces.util.ComponentUtils;
 import org.primefaces.util.Constants;
-import org.primefaces.util.MapBuilder;
 
-import java.util.Collection;
 import java.util.Map;
 
 import jakarta.el.ELContext;
@@ -46,7 +43,6 @@ import jakarta.faces.component.UIComponent;
 import jakarta.faces.context.FacesContext;
 import jakarta.faces.event.AbortProcessingException;
 import jakarta.faces.event.AjaxBehaviorEvent;
-import jakarta.faces.event.BehaviorEvent;
 import jakarta.faces.event.FacesEvent;
 
 @FacesComponent(value = AccordionPanel.COMPONENT_TYPE, namespace = AccordionPanel.COMPONENT_FAMILY)
@@ -55,7 +51,7 @@ import jakarta.faces.event.FacesEvent;
 @ResourceDependency(library = "primefaces", name = "jquery/jquery-plugins.js")
 @ResourceDependency(library = "primefaces", name = "core.js")
 @ResourceDependency(library = "primefaces", name = "components.js")
-public class AccordionPanel extends AccordionPanelBase {
+public class AccordionPanel extends AccordionPanelBaseImpl {
 
     public static final String COMPONENT_TYPE = "org.primefaces.component.AccordionPanel";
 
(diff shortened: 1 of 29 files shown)


tandraschko:
Will review tomorrow

primefaces/src/main/java/org/primefaces/component/accordionpanel/AccordionPanel.java
```diff
@@ -68,28 +64,9 @@ public class AccordionPanel extends AccordionPanelBase {
...
-        return BEHAVIOR_EVENT_MAPPING;
-    }
-
-    @Override
-    public Collection<String> getEventNames() {
-        return EVENT_NAMES;
-    }
-
     @Override
     public String getDefaultEventName() {
```
tandraschko: should we add a "default=true" to the @FacesBehaviorAnnotation?

primefaces/src/main/java/org/primefaces/component/accordionpanel/AccordionPanelBase.java
```diff
@@ -67,133 +54,42 @@ public String getFamily() {
...
-    public String getStyle() {
-        return (String) getStateHelper().eval(PropertyKeys.style, null);
-    }
+    @Property(description = "Client side callback to execute when a tab is closed.")
+    public abstract String getOnTabClose();
 
-    public void setStyle(String style) {
-        getStateHelper().put(PropertyKeys.style, style);
-    }
+    @Property(defaultValue = "true", description = "When enabled, dynamically loaded tab contents are cached. Default is true.")
```
tandraschko, line 78: i dont like the "Default is..."
should we remove it?
but we have to add a new field then like "implicitDefaultValue" as we manage some defaults in JS directly, not via StateHelper

primefaces/src/main/java/org/primefaces/component/api/TouchAware.java
```diff
@@ -23,24 +23,14 @@
...
  * to turn it off globally or per component that supports touch.
  */
 public interface TouchAware {
 
-    /**
-     * Is this component touch enabled?
-     *
-     * @return false to disable
-     */
+    @Property(description = "Enable touch support if browser detection supports it. Default is NULL because it is globally enabled by default.")
```
tandraschko: boolean isXXX
Boolean getXXX

we should change that to primitive propably?

primefaces/src/main/java/org/primefaces/component/accordionpanel/AccordionPanel.java
melloware: Agreed a `default` on the Events would save more boilerplate!

primefaces/src/main/java/org/primefaces/component/accordionpanel/AccordionPanelBase.java
melloware: Agreed a lot of this I was just taking from the primefaces.taglib.xml descriptions but you are right we should add an `implicitDefaultValue`?

primefaces/src/main/java/org/primefaces/component/api/TouchAware.java
melloware: I was going to but realized this is an override. If the web.xml TOUCH_ENABLED is true that is the global default this is just a per component override. So if this is NULL it uses the global TOUCH_ENABLED if it has a value it would not look at the global. So it has to be true, flase, NULL if I default to true it then would never reach TOUCH_ENABLED in `ComponentUtils.isTouch`

tandraschko: then please call it "getTouchable"

melloware: OK fixed to getTouchable

PR #14419, opened 2025-11-24
Fix #14418: first part to add keyboard navigation to SlideMenu

### LINKED ISSUE #14418: SlideMenu: Missing keyboard navigation
### Describe the bug

Right now there is no possibility to focus the Slide Menu component or navigate through the menu entries only with the keyboard. The reason for the same is that all elements are basicly rendered with tabindex -1 and there is no function to change it during user interaction.

The current state can be tested on the showcase page -> https://showcase.primefaces.org/ui/menu/slideMenu.xhtml

To solve the problem the following things have to be done:

1. All menu entries needs a intial value on the first rendering:

Main menu entries        -> tabindex 0
Sub menu links              -> tabindex 0
Sub menu entries          -> tabindex -1
Sub menu back button -> tabindex -1

2. Add a function to switch tabindex when you move to a sub menu and also when you go back to main menu

3. Add js functions for hover and focus to add the necessary styleclasses

4. Make sub menu back button "clickable"

5. Add an addtional fix for overlay slide menus.

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

15.0.10

### Theme

_No response_

### JSF implementation

All

### JSF version

4.0

### Java version

21

### Browser(s)

_No response_

Teutron-flo:
Fix #14418


primefaces/src/main/frontend/packages/components/src/menu/menu.slidemenu.widget.ts
```diff
@@ -232,4 +250,16 @@ export class SlideMenu<Cfg extends SlideMenuCfg = SlideMenuCfg> extends Menu<Cfg
...
+    * Change the tabindex of the menu elements
+    * @private
+    */
+    private changeTabindex() {
+        this.linksActive = this.jq.find('a.ui-menuitem-link[tabindex=0]');
+        this.linksInactive = this.jq.find('a.ui-menuitem-link[tabindex=-1]');
+
+        this.linksActive.attr('tabindex', -1);
+        this.linksInactive.attr('tabindex', 0);
+    }
```
melloware: MOve this method up above the `override show` so its groupped.

melloware (changes requested):

primefaces/src/main/frontend/packages/components/src/menu/menu.slidemenu.widget.ts
Teutron-flo: Changes are done in the latest commit

melloware (approved):

PR #14420, opened 2025-11-24
CDK: Media, Audio, Video

melloware:


primefaces/src/main/java/org/primefaces/component/api/UIMediaBase.java
```diff
@@ -0,0 +1,185 @@
...
+
+import jakarta.faces.component.UIComponentBase;
+import jakarta.faces.context.FacesContext;
+
+/**
+ * Base class for HTML5 audio and video components.
+ */
+public abstract class UIMediaBase extends UIComponentBase {
+
+    public String resolveSource(FacesContext context, UIMediaBase media) throws IOException {
```
tandraschko: we dont need UIMediaBase media as param, juse use "this"

melloware: Done

primefaces/src/main/java/org/primefaces/component/api/UIMediaBase.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    @Property(defaultValue = "true", description = "Whether to cache the media content")
+    public abstract boolean isCache();
+
+    @Property(description = "Specifies that the media will start playing as soon as it is ready.")
+    public abstract boolean isAutoplay();
+
+    @Property(description = "Specifies that media controls should be displayed (such as a play/pause button etc).")
+    public abstract boolean isControls();
```
melloware, line 57: @tandraschko this is how i figured it out when i did this with the CDK i realized it was all rendering correctly.

PR #14425, opened 2025-11-25
PrimePartialResponseWriter: encode scripts for xml

### LINKED ISSUE #14427: scriptsToExecute not escaped in Ajax response
### Describe the bug

Scripts in PrimeRequestContext#getScriptsToExecute are not escaped and can break parsing the partial response written by PrimeFaces

### Reproducer

The following oncomplete never runs and yields an XML parsing error in Firefox dev tools, Chrome doesn't work either but it doesn't log any errors:
```
            <p:commandButton id="btn1" value="button 1"
                             process="@form" update="@none"
                             oncomplete="#{'alert(true &amp;&amp; false);'}"/>
```

For reference, this xml element is returned in the partial response and cannot be parsed: `<extension ln="primefaces" type="script">alert(true && false);</extension>`

### Expected behavior

Scripts written in the extension elements are properly escaped

### PrimeFaces edition

Community

### PrimeFaces version

16.0.0-SNAPSHOT

### Theme

_No response_

### JSF implementation

All

### JSF version

4.0

### Java version

25

### Browser(s)

_No response_

jungm:
Fixes #14427

### DIFF
--- a/primefaces/src/main/java/org/primefaces/context/PrimePartialResponseWriter.java
+++ b/primefaces/src/main/java/org/primefaces/context/PrimePartialResponseWriter.java
@@ -208,7 +208,7 @@ public class PrimePartialResponseWriter extends PartialResponseWriter {
             startEval();
 
             for (String initScript : initScripts) {
-                getWrapped().write(EscapeUtils.forXml(initScript));
+                getWrapped().write(EscapeUtils.forCDATA(initScript));
                 getWrapped().write(';');
             }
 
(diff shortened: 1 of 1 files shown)


tandraschko:
can you also add a issue for this one?

melloware (approved):

tandraschko:
i cant remember the whole XML response but cant we just render a CDATA around the script?

jungm:
yup should be possible as well

tandraschko:
can you post a whole AJAX response?
i think update content is also wrapped in CDATA, better to do this here too propably

jungm:
@tandraschko that's what it looks like now
```
<?xml version='1.0' encoding='UTF-8'?>
<partial-response>
    <changes>
        <update id="j_id1:jakarta.faces.ViewState:0"><![CDATA[7433092339894341240:8734877916689508986]]></update>
        <extension type="script" ln="primefaces"><![CDATA[alert(true && false);]]></extension>
    </changes>
</partial-response>
```

PR #14426, opened 2025-11-25
ComponentUtils test partial source before queuing DynamicOncompleteEvent

### LINKED ISSUE #14424: AjaxSource: oncomplete is always invoked when using EL
### Describe the bug

When using EL in the `oncomplete` attribute on AjaxSource components (e.g. CommandButton) it gets invoked every time the component is processed, even if the ajax callback as not  triggered by this component

From some quick debugging I saw this gets written as `<extension ln="primefaces" type="script">` elements in the AJAX response, originating from `ComponentUtils#decodeBehaviours`. Probably just some filtering missing there

### Reproducer

```
        <h:form id="frmTest">
            <p:commandButton id="btn1" value="button 1"
                             process="@form" update="@none"
                             oncomplete="#{'alert(1)'}"/>
            <p:commandButton id="btn2" value="button 2"
                             process="@form" update="@none"
                             oncomplete="#{'alert(2)'}"/>
        </h:form>
```

Click one of the two buttons and both 1 and 2 are alerted

### Expected behavior

only one alert is triggered when a button gets clicked

### PrimeFaces edition

Community

### PrimeFaces version

16.0.0-SNAPSHOT

### Theme

_No response_

### JSF implementation

All

### JSF version

4.0

### Java version

25

### Browser(s)

_No response_

jungm:
Fixes #14424

### DIFF
--- a/primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
+++ b/primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
@@ -267,13 +267,7 @@ public class ComponentUtils {
     }
 
     public static void decodeBehaviors(FacesContext context, UIComponent component) {
-        if (!(component instanceof ClientBehaviorHolder)) {
-            return;
-        }
-
-        Map<String, String> params = context.getExternalContext().getRequestParameterMap();
-        String partialSource = params.get(Constants.RequestParams.PARTIAL_SOURCE_PARAM);
-        if (partialSource != null && !component.getClientId().equals(partialSource)) {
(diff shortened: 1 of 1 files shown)


primefaces/src/main/java/org/primefaces/util/ComponentUtils.java
```diff
@@ -271,6 +271,12 @@ public static void decodeBehaviors(FacesContext context, UIComponent component)
             return;
         }
 
+        Map<String, String> params = context.getExternalContext().getRequestParameterMap();
```
tandraschko: you can use ComponentUtils#isRequestSource here

melloware (approved):

PR #14435, opened 2025-11-25
CDK: AjaxStatus

melloware:
@tandraschko found one minor issue here.  The facet is called "default" but i can't make a CDK binding for it because `default` in the enum is a reserved Java keyword.


tandraschko:
should we prefix java keywords with a "_" or somthing?

tandraschko:
added escaping, should work now fine

melloware:
awesome let me update this PR.

melloware:
pushed and worked like a charm i see `_default` in the generated code.

PR #14439, opened 2025-11-25
#14406: Add TreeTable filterPrune descendants mode with pruning logic / tests

### LINKED ISSUE #14406: TreeTable: on filter part, add prune tree feature based on the children status rather than ancestor
### Description

The current filter behavior is keeping all nodes if that node's any ancestors has matched filter value. Even though that node is not match.
 
But we want to add a feature remove all sub-nodes which are not match the filter value.

For example, filtering to get 'true' value with current behavior. The tree will keep the 'B' and 'E' node since one of their own ancestor has 'true' value:

Image

But we want to remove the 'B' and 'E' since they are not in 'true' status and they do not have any descendant with 'true' status.

Image

we want to remove the 'B' and 'E' since they are not in 'true' status and they do not have any descendant with 'true' status.

Image

### Additional context

_No response_

jxmai:
Fix https://github.com/primefaces/primefaces/issues/14406

Proposal: add an API attribute in treeTable called `filterPrune="descendants"`

Behavior when filterPrune="descendants":

- A node stays if it matches the filter itself OR if any descendant matches.

- A node is pruned only when neither it nor any of its descendants match.

- Defaults to current behavior when the attribute is absent.

---

Proposed API:

* Attribute: filterPrune
* Values: none (default, current behavior: keep all children of a matching node), descendants (prune nonmatching children unless they or their descendants match).
* Usage:

```html
  <p:treeTable value="#{bean.root}"
               var="node"
               filterPrune="descendants">
      <p:column headerText="Name" filterBy="#{node.name}" filterMatchMode="contains">
          #{node.name}
      </p:column>
  </p:treeTable>
```

---

Sample Test project:

http://localhost:8080/

[primefaces-test-14406.zip](https://github.com/user-attachments/files/23755337/primefaces-test-14406.zip)

image

### DIFF
--- a/docs/16_0_0/components/treetable.md
+++ b/docs/16_0_0/components/treetable.md
@@ -39,6 +39,7 @@ filterDelay | 300 | Integer | Delay in milliseconds before sending an ajax filte
 filterEvent | keyup | String | Event triggering filter for input filters. If "enter" it will only filter after ENTER key is pressed.
 filteredValue  | null | TreeNode | TreeNode to keep filtered data.
 filterNormalize | false | Boolean | Defines if filtering would be done using normalized values (accents will be removed from characters). Default is false.
+filterPrune | none | String | Controls pruning during filtering. Use `descendants` to prune non-matching children unless they or their descendants match; `none` keeps children of matching nodes. |
 first | 0 | Integer | Index of the first data to display.
 globalFilter | null | String | Value of the global filter to use when filtering by default.
 globalFilterFunction | null | MethodExpression | Custom implementation to globally filter a value against a constraint.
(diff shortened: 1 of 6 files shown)


melloware:
Don't forget the whatsnew.md and the 16-0/docs for treetable.md to add the new property.

jxmai:
Thanks for the reminder! I updated the [docs](https://github.com/jxmai/primefaces/blob/feature/%2314406/docs/16_0_0/components/treetable.md) and [what's new](https://github.com/jxmai/primefaces/blob/feature/%2314406/docs/16_0_0/gettingstarted/whatsnew.md).

jxmai:
@melloware I saw an unrelated error from CI (should be unstable env?), can we re-run the CI, there shouldn't be any error by the changes.

melloware:
yep re-running now. Thanks for adding the docs!

melloware (approved):

PR #14442, opened 2025-11-26
Fix #14390: 15.0.11 Passthrough attribute String "false" should be treated like boolean false

### LINKED ISSUE #14390: Video: The `controls` attribute does not produce the expected effect

melloware:
Fix #14390: 15.0.11 Passthrough attribute String "false" should be treated like boolean false

@tandraschko got to teh bottom of this.  The magic is in `shouldRenderAttribute` if a passthrough attribute is a `String` its not respecting "true" and "false".  If it was a `Boolean` it was.

so I added a check to treat `false` like boolean false to not render the attribute instead of rendering `controls="false"`.

```java
protected boolean shouldRenderAttribute(Object value) {
        if (value == null) {
            return false;
        }

        if (value instanceof Boolean) {
            return (Boolean) value;
        }
        else if (value instanceof Number) {
            Number number = (Number) value;

            if (value instanceof Integer) {
                return number.intValue() != Integer.MIN_VALUE;
            }
            else if (value instanceof Double) {
                return number.doubleValue() != Double.MIN_VALUE;
            }
            else if (value instanceof Long) {
                return number.longValue() != Long.MIN_VALUE;
            }
            else if (value instanceof Byte) {
                return number.byteValue() != Byte.MIN_VALUE;
            }
            else if (value instanceof Float) {
                return number.floatValue() != Float.MIN_VALUE;
            }
            else if (value instanceof Short) {
                return number.shortValue() != Short.MIN_VALUE;
            }
        }
        else if (value instanceof String) {
            // #14390: passthrough attribute with "false" should be treated like boolean false
            return !"false".equalsIgnoreCase((String) value);
        }

        return true;
    }
```

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/video/Video.java
+++ b/primefaces/src/main/java/org/primefaces/component/video/Video.java
@@ -50,14 +50,6 @@ public class Video extends UIMedia {
         return COMPONENT_FAMILY;
     }
 
-    public boolean isControls() {
-        return (boolean) getStateHelper().eval(PropertyKeys.controls, true);
-    }
-
-    public void setControls(boolean controls) {
-        getStateHelper().put(PropertyKeys.controls, controls);
-    }
-
     public String getWidth() {
         return (String) getStateHelper().eval(PropertyKeys.width, null);
     }
--- a/primefaces/src/main/java/org/primefaces/renderkit/CoreRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/renderkit/CoreRenderer.java
@@ -256,13 +256,22 @@ public abstract class CoreRenderer extends Renderer {
         renderPassThroughAttributes(context, component);
     }
 
+    /**
+     * Renders a single attribute on the given UIComponent if the value should be rendered.
+     *
+     * @param context   the current FacesContext instance
+     * @param component the component on which to render the attribute
+     * @param attribute the name of the attribute to render
+     * @param value     the value of the attribute
+     * @throws IOException if an input/output error occurs during rendering
+     */
     protected void renderAttribute(FacesContext context, UIComponent component, String attribute, Object value)
                 throws IOException {
         ResponseWriter writer = context.getResponseWriter();
 
         if (shouldRenderAttribute(value)) {
             String stringValue = value.toString();
-            if (Boolean.parseBoolean(stringValue)) {
+            if ("true".equalsIgnoreCase(stringValue)) {
                 writer.writeAttribute(attribute, true, attribute);
             }
             else {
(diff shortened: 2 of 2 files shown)


tandraschko:
can you still document
https://github.com/primefaces/primefaces/blob/592f8310aa1d6b7c1ba2fc51253df6040df5e8ce/primefaces/src/main/java/org/primefaces/renderkit/CoreRenderer.java#L274

still unsure what it does

melloware:
I just changed the code from `Boolean.parseBoolean` to `"true".equalsIgnoreCase(stringValue)` so its more readable and that is what it was doing.  It was checking if its the word "true" "TRUE" etc all other values return false including NULL

tandraschko:
but where is the code which decides:

controls=true renders controls=controls, where as false removes it completely?

melloware:
its this line of VideoRenderer.java https://github.com/primefaces/primefaces/blob/00ebfd622a0201b49405b1255396d1a2f76d6018/primefaces/src/main/java/org/primefaces/component/video/VideoRenderer.java#L69

melloware:
it treats ALL passthrough attributes as Strings

melloware:
when i added the Boolean proeprty explicitly then the `shouldRender` saw it as a Boolean and said its false don't render this attribute at all.  When its undefined in the  component it defaults to String and shouldrender was saying yes and it gets rendered as `controls="controls"` or `controls="false"`  because its just a string and fell through the `shouldRender` check

tandraschko:
i dont get it

at the end, we call responseWriter.write("attrname", attributeValue), right?

If "shouldRenderAttribute" returns false for "false as bool", thats ok

now, we added "false" string support, still ok

but shouldnt
responseWriter.write("attrname", true)

lead to

controls="true" in the html?

melloware:
OK yes so let me step through the values of `controls`

**String:** (current PrimeFaces 15.0.10)
```
console="true"
console="false"
```

**String:** (current PrimeFaces 15.0.11-SNAPSHOT with fixes above)
```
console="true"
// "false" not rendered
```

**Boolean:** (current PrimeFaces 16.0.0-SNAPSHOT using CDK which has controls as a boolean)
```
console="true"
// false NOT rendered at all
```

tandraschko:
Ahhh
I See 
All clear now 🤞

PR #14444, opened 2025-11-26
CDK: InputTextArea

### LINKED ISSUE #14445: InputTextArea: NPE in AutoComplete example this.items undefined
Image

melloware:
@tandraschko explanation below...

- [x] CDK now scans hiearchy for `@FacesBehaviors` (itemSelect in one class , but inherit all the others from interface)
- [x] CDK added `callSuper=true` to allow documenting a property (like getValue) but having it pass through to it's super class
- [x] `InputAware` interface lets us map and re-use all the behaviors/props from both HTMLInput and InputTextArea so this will be reused when i get to InputText components
- [x] `CountCharactersAware` will be reused in both InputText and InputTextArea
- [x] FacesBehavior in Textarea but still inherits all the ones from `InputAware`!

Fix #14445


primefaces/src/main/frontend/packages/components/src/forms/forms.inputtextarea.widget.js
```diff
@@ -37,6 +37,7 @@ PrimeFaces.widget.InputTextarea = class InputTextarea extends PrimeFaces.widget.
      */
     init(cfg) {
         super.init(cfg);
+        this.items = $();
```
melloware, line 41: was getting an NPE in the showcase when you first try and query the Autocomplete example because this.items is undefined. I need to check PF15 for the same issue

primefaces/src/main/java/org/primefaces/component/inputtext/InputTextBase.java
```diff
@@ -85,10 +84,10 @@ public void setCounterTemplate(String counterTemplate) {
     }
 
     public boolean getCountBytesAsChars() {
-        return (Boolean) this.getStateHelper().eval(InputTextareaBase.PropertyKeys.countBytesAsChars, false);
+        return (Boolean) this.getStateHelper().eval(PropertyKeys.countBytesAsChars, false);
     }
 
     public void setCountBytesAsChars(boolean countBytesAsChars) {
-        this.getStateHelper().put(InputTextareaBase.PropertyKeys.countBytesAsChars, countBytesAsChars);
+        this.getStateHelper().put(PropertyKeys.countBytesAsChars, countBytesAsChars);
```
melloware, line 91: InputText mistakenly had this import i will check PF15 to see if it has the same issue.

primefaces/src/main/java/org/primefaces/component/inputtextarea/InputTextareaRenderer.java
```diff
@@ -121,7 +121,7 @@ protected void encodeScript(FacesContext context, InputTextarea component) throw
         WidgetBuilder wb = getWidgetBuilder(context);
         wb.init("InputTextarea", component)
                 .attr("autoResize", autoResize)
-                .attr("maxlength", component.getMaxlength(), Integer.MAX_VALUE);
+                .attr("maxlength", component.getMaxlength(), Integer.MIN_VALUE);
```
melloware, line 124: this is MinValue to be consistent with InputText and HtmlInputText default

primefaces/src/main/java/org/primefaces/component/api/InputAware.java
```diff
@@ -0,0 +1,240 @@
...
+    @Property(description = "Defines a label for the input element.")
+    String getLabel();
+
+    @Property(description = "Specifies the language of the element's content.")
+    String getLang();
+
+    @Property(description = "Indicates whether the local value of the component has been set.", defaultValue = "false")
+    boolean isLocalValueSet();
+
+    @Property(description = "Defines the maximum number of characters allowed in the input.", defaultValue = "-2147483648")
```
melloware, line 109: Integer.MIN_VALUE has to be expressed as -2147483648 since defaultValue requires a String?  didn't know if I coudl do Integer.MIN_VALUE.toString()?

primefaces/src/main/java/org/primefaces/component/api/InputAware.java
```diff
@@ -0,0 +1,240 @@
...
+ * Interface for components that are aware of typical input behaviors and attributes.
+ * <p>
+ * Implementing this interface allows a component to declare and document standard input-related properties, such as access key,
+ * maximum length, converter, accessibility features, keyboard input modes, and event scripts.
+ * <p>
+ * The presence of this interface indicates that the implementing component behaves similarly to HTML input elements,
+ * providing customizable interaction and validation mechanisms.
+ */
+@FacesBehaviorEvents({
+    @FacesBehaviorEvent(name = "blur", event = AjaxBehaviorEvent.class, description = "Fires when the element loses focus."),
```
melloware, line 43: Wasn't sure if I was supposed to use implicitValue here?

primefaces/src/main/java/org/primefaces/component/api/InputAware.java
melloware: Nope `Integer.MIN_VALUE.toString()` not allowed but no big deal.

primefaces/src/main/frontend/packages/components/src/forms/forms.inputtextarea.widget.js
melloware: Yep PR submitted to fix PF15

primefaces/src/main/java/org/primefaces/component/inputtext/InputTextBase.java
melloware: Yep PF15 same issue fixed in PR.

primefaces/src/main/java/org/primefaces/component/api/InputAware.java
tandraschko: What do you mean exactly?

melloware: I saw the annotation you had some value "implicit" but wasn't sure what it was used for in BehaviorEvent?

tandraschko: This currently not needed

tandraschko:
I wonder If "change" should be defautEvent=true?

melloware:
I was wondering that but when I look in the base class in the Faces impl it has valueChange as the default so I kept it?

tandraschko:
Hmm
Yeah
We just to call the method then in the showcase docs

melloware:
OK I added a showcase example for the default AJAX event

tandraschko:
I dont mean example
I mean in the docs / panel.xhtml / componentDocumentation.xhtml

melloware:
Oh ok added it as rendered="false" like Panel and also added `Default` to the events tab so you can see which one is default.

melloware:
I also fixed Badge and will add the others in another PR. Badge XHTML had a mixture of TAB and SPACE so I make all TAB's SPACES as well

tandraschko:
perfect
thats enough for now

i dont we should add componentDocumentation to all components, just lets stay with panel and badge for testing later

PR #14457, opened 2025-11-30
CDK: Card

melloware:


tandraschko:
Just pass FacetKey instead of String and you can completely remove the Switch

melloware:
Great suggestion I don't know why I didn't think of that!  Fixed.

PR #14461, opened 2025-12-01
CDK: InputText and all subclasses of InputText

melloware:
Sorry this is a big one but building off all the good work from TextArea this was the logical one to do next and you have to do them all together.

### DIFF
--- a/primefaces-cdk/primefaces-cdk-impl/src/main/java/org/primefaces/cdk/impl/subclass/AnnotationProcessor.java
+++ b/primefaces-cdk/primefaces-cdk-impl/src/main/java/org/primefaces/cdk/impl/subclass/AnnotationProcessor.java
@@ -1096,7 +1096,7 @@ public class AnnotationProcessor extends AbstractProcessor {
             w.print("        " + escapeKeyword(event.getName()) + "(\"" + event.getName() + "\", " +
                     event.getEventClass() + ".class, \"" + description + "\", " + event.isImplicit() + ", " + event.isDefaultEvent() + ")");
             w.println(i < events.size() - 1 ? "," : ";");
-            if (event.isDefaultEvent()) {
+            if (event.isDefaultEvent() && defaultEvent == null) {
                 defaultEvent = event;
             }
         }
--- a/primefaces/src/main/java/org/primefaces/component/api/UICalendar.java
+++ b/primefaces/src/main/java/org/primefaces/component/api/UICalendar.java
@@ -23,6 +23,11 @@
  */
 package org.primefaces.component.api;
 
+import org.primefaces.cdk.api.FacesBehaviorEvent;
+import org.primefaces.cdk.api.FacesBehaviorEvents;
+import org.primefaces.cdk.api.FacesComponentBase;
+import org.primefaces.cdk.api.Property;
+import org.primefaces.event.DateViewChangeEvent;
 import org.primefaces.util.CalendarUtils;
 import org.primefaces.util.ComponentUtils;
 import org.primefaces.util.ELUtils;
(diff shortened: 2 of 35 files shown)


primefaces-cdk/primefaces-cdk-impl/src/main/java/org/primefaces/cdk/impl/subclass/AnnotationProcessor.java
```diff
@@ -1096,7 +1096,7 @@ private void writeClientBehaviorEventKeys(PrintWriter w, List<BehaviorEventInfo>
             w.print("        " + escapeKeyword(event.getName()) + "(\"" + event.getName() + "\", " +
                     event.getEventClass() + ".class, \"" + description + "\", " + event.isImplicit() + ", " + event.isDefaultEvent() + ")");
             w.println(i < events.size() - 1 ? "," : ";");
-            if (event.isDefaultEvent()) {
+            if (event.isDefaultEvent() && defaultEvent == null) {
```
melloware, line 1099: We want the first default event found in the order of nearest class so first one found wins. This was discovered in ColorPicker.java

tandraschko:
CalendarUtils.getObjectAsInstant(context, this, getMindate(), "mindate")

Cant we just pass the object value instead of attr name?
We should never ever use attr names as str

melloware:
Agreed I will fix.

melloware:
OK fixed to use the PropertyKeys that was only being used to display the property name in the FacesException which is why it uses `name()`

tandraschko:
Timeonly is also used as String name

melloware:
Good catch. Fixed.

tandraschko:
Can you maybe search for others?
Just on my Phone until end of the week

melloware:
Thanks found a couple of more and Also attempting DatePicker to `List<Object> disabledDates` instead of `List`

tandraschko:
Why is it required, that the getter ist implemented in UICalendar? Its not good that UICalendar references the Impl for the ProperyKeys

melloware:
So UICalendar used to have this helper

```java
public Boolean isTimeOnlyWithoutDefault() {
        return (Boolean) getStateHelper().eval(PropertyKeys.timeOnly);
}
```

Which is `isTimeOnly` but without the default and makes a capital Boolean so it can be NULL.

```java
public boolean isTimeOnly() {
        return (Boolean) getStateHelper().eval(PropertyKeys.timeOnly, false);
    }
```

So to keep this method in UICalendar it has to reference a PropertyKeys which is in UICalendarImpl now since this is a base class used by DatePicker and Calendar.  I can move this method to Calendar and DatePicker but then its no longer referencable by its interface `UICalendar` unless i make it `abtract` which I can do.

tandraschko:
Maybe we can create methods like:
isXySet

Otherwise yeah, make it abstracr

melloware:
OK i refactored it to make it abstract in `UICalendar` and the implementations are in `DatePicker` and `Calendar` so it can use the PropertyKeys correctly.

primefaces/src/main/java/org/primefaces/component/api/UICalendar.java
```diff
@@ -54,141 +64,55 @@ public abstract class UICalendar extends AbstractPrimeHtmlInputText implements I
...
-    public Object getMindate() {
-        return getStateHelper().eval(PropertyKeys.mindate, null);
-    }
+    @Property(description = "When enabled, shows only time picker.", defaultValue = "false")
+    public abstract boolean isTimeOnly();
 
-    public void setMindate(Object mindate) {
-        getStateHelper().put(PropertyKeys.mindate, mindate);
-    }
+    public abstract Boolean isTimeOnlyWithoutDefault();
```
tandraschko: same again
primitive -> is
bool Object -> get

primefaces/src/main/java/org/primefaces/component/api/UICalendar.java
```diff
@@ -400,8 +290,8 @@ public Class<?> getValueType() {
     }
 
     public void validateMinMax(FacesContext context) {
-        Instant minDate = CalendarUtils.getObjectAsInstant(context, this, getMindate(), PropertyKeys.mindate.name());
-        Instant maxDate = CalendarUtils.getObjectAsInstant(context, this, getMaxdate(), PropertyKeys.maxdate.name());
+        Instant minDate = CalendarUtils.getObjectAsInstant(context, this, getMindate(), UICalendarImpl.PropertyKeys.mindate.name());
```
tandraschko: same again
we should not use child class stuff from parent class

primefaces/src/main/java/org/primefaces/component/calendar/BaseCalendarRenderer.java
```diff
@@ -110,7 +111,7 @@ protected void encodeInput(FacesContext context, T component, String id, String
         boolean disabled = false;
 
         if (popup) {
-            String inputStyleClass = createStyleClass(component, UICalendar.PropertyKeys.inputStyleClass.name(), UICalendar.INPUT_STYLE_CLASS);
+            String inputStyleClass = createStyleClass(component, UICalendarImpl.PropertyKeys.inputStyleClass.name(), UICalendar.INPUT_STYLE_CLASS);
```
tandraschko: same here

melloware:
OK renamed to `getTimeOnlyWithoutDefault`

Removed use of `UICalendarBaseImpl` in BaseCalendar and UiCalendar had to go back to using strings instead of Enum.name()

melloware:
Update this PR with `FacesComponentDescription` will do the others

PR #14491, opened 2025-12-05
Fix #14487: DatePicker widget setDate(null) should trigger AJAX

### LINKED ISSUE #14487: DatePicker: setDate(null) Javascript function doesn't trigger Ajax
### Describe the bug

We have Ajax change events attached to some DatePicker elements. We have to use Javascript setDate() method in some of our automated tests. setDate() triggers Ajax, if the parameter is a date string, but no Ajax will be triggered if we use setDate(null) or setDate('') instead, although the input field will be cleared.

### Reproducer

Sorry, no complete code, but any datePicker with <p:ajax event="change" /> tag will work. Call it with PF('<widget_var>').setDate('any date'); to check that Ajax normally works and then call it with PF('<widget_var>').setDate(null); . No Ajax will be triggered.

### Expected behavior

Ajax is triggered in both cases.

### PrimeFaces edition

Community

### PrimeFaces version

15.0.10

### Theme

Modified saga

### JSF implementation

All

### JSF version

4.0

### Java version

21

### Browser(s)

Chrome and Firefox (no other browsers tested)

melloware:
Fix #14487: DatePicker widget setDate(null) should trigger AJAX

### DIFF
--- a/primefaces/src/main/frontend/packages/datepicker/datepicker/src/datepicker.widget.js
+++ b/primefaces/src/main/frontend/packages/datepicker/datepicker/src/datepicker.widget.js
@@ -489,6 +489,11 @@ PrimeFaces.widget.DatePicker = class DatePicker extends PrimeFaces.widget.BaseWi
      */
     setDate(date) {
         this.jq.datePicker('setDate', date);
+
+        // #14487 AJAX is not triggered on setDate(null). Specifically not done in the Prime Datepicker widget.
+        if (!date) {
+            this.fireDateSelectEvent();
+        }
     }
 
     /**


tandraschko:
not sure but this has a potential to break something

melloware:
Yeah i thought so too so i wanted to run all the integration tests first (we have a LOT for DatePicker).  So far its all passing.

tandraschko:
im not sure IT covers multiple requests

primefaces/src/main/frontend/packages/datepicker/datepicker/src/datepicker.widget.js
```diff
@@ -489,6 +489,11 @@ PrimeFaces.widget.DatePicker = class DatePicker extends PrimeFaces.widget.BaseWi
      */
     setDate(date) {
         this.jq.datePicker('setDate', date);
+
+        // #14487 AJAX is not triggered on setDate(null). Specifically not done in the Prime Datepicker widget.
+        if (!date) {
```
tandraschko, line 494: why is the if required? it should be required either with null or real date?

melloware: So this could be done in the core Prime DatePicker here is the setDate

```js
    setDate: function(date) {
        if (!date) {
            this.updateModel(null, null);
            return;
        }

        var newDate = this.parseValue(date);
        var newDateMeta = { day: newDate.getDate(), month: newDate.getMonth(), year: newDate.getFullYear(), selectable: true /*, today: true*/ };

        /* set changes */
        this.value = newDate;
        this.updateViewDate(null, newDate);
        this.onDateSelect(null, newDateMeta);
    }
```

I could change it to this

```js
    setDate: function(date) {
        if (!date) {
            this.updateModel(null, null);
            this.onDateSelect(null, null);  // FIRE AJAX CHANGE ON NULL
            return;
        }

        var newDate = this.parseValue(date);
        var newDateMeta = { day: newDate.getDate(), month: newDate.getMonth(), year: newDate.getFullYear(), selectable: true /*, today: true*/ };

        /* set changes */
        this.value = newDate;
        this.updateViewDate(null, newDate);
        this.onDateSelect(null, newDateMeta);
    }
```

but since this is called internally i was concerned about doing it there where as outside in the widget I know the only people calling `setDate` on the widget are users doing it intentionally.  Thoughts?

tandraschko:
im not sure yet, it juts feels a bit dirty now

melloware:
We can also just mark it "workaround provided" as I gave the user a MonkeyPatch.

melloware:
ok closing this as workaround provided.

PR #14494, opened 2025-12-05
Fix #14474: Picklist dragDrop property to disable transfer/reordering

### LINKED ISSUE #13131: PickList: add flag to disable dragdrop support

### LINKED ISSUE #14474: PickList: Reordering cannot be disabled
### Describe the bug

The documentation states that:

> PickList support reordering of source and target lists, these are enabled by showSourceControls and showTargetControls options.

However, this is incorrect. These options only show or hide the reorder buttons, but they do not enable or disable reordering.

In practice:

- Reordering (drag-and-drop sorting) is always enabled, regardless of showSourceControls or showTargetControls.
- There is no supported API to disable reordering in either list.

This is problematic when an application needs a PickList without allowing users to change the order of items. Since reordering cannot be disabled, the UI suggests functionality that should not be available.

### Reproducer

1. Create a basic pickList
2. Set `showSourceControls="false"`  and `showTargetControls="false"`
3. Run the application (or just open the PrimeFaces Showcase)
4. Try dragging items within the source or target lists

### Expected behavior

Items should not be reorderable. Ideally, it should be possible to configure this behavior with an attribute.

### PrimeFaces edition

None

### PrimeFaces version

15.0.0

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0

### Java version

21

### Browser(s)

_No response_

melloware:
Fix #14474: Picklist dragDrop property to disable transfer/reordering
Fix #13131

- [x] New feature for 16.0.0
- [x] called it `dragDrop` which is what FileUpload calls the property

### DIFF
--- a/docs/16_0_0/components/picklist.md
+++ b/docs/16_0_0/components/picklist.md
@@ -58,6 +58,7 @@ orientation | horizontal | String | Defines layout orientation, valid values are
 responsive | false | Boolean | In responsive mode, picklist adjusts itself based on screen width.
 escape | true | Boolean | Defines if labels of the component are escaped or not.
 tabindex | null | String | Position of the element in the tabbing order.
+dragDrop | true | Boolean | Specifies dragdrop based drag and drop transfers and reordering, default is true and works only on supported browsers.
 filterEvent | keyup | String | Client side event to invoke picklist filtering for input fields.
 filterDelay | 300 | Integer  | Delay to wait in milliseconds before sending each filter query.
 escapeValue | true | Boolean | Defines if values of the component are escaped or not.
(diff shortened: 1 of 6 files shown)


tandraschko:
i still dont see a reason to make disableable
but if primeng has, it im ok with it

PR #14496, opened 2025-12-06
Modernize Browser Detection (MS Edge)

melloware:
- the old jquery browser plugin had not been updated in 11 years
- I refactored it to support MS Edge modern Chromium
- I refactored to include TypeScript in the plugin instead of Definitely Typed

cc @blutorange


tandraschko:
Why we need it anymore? Only for safari?

melloware:
in our code its used in about 10 places in DataTable/TreeTable/AutoComplete and TextEditor.

Then if I remember multiple jQuery plugins we use have it in internally as `$.browser` as well.

For example

```js
 var browser = PrimeFaces.env.browser;
if(browser.webkit === true || browser.mozilla === true)
      this.frozenBody.append('<div style="height:' + scrollBarWidth + ';border:1px solid transparent"></div>');
else
      this.frozenBodyTable.css('margin-bottom', scrollBarWidth);
```

tandraschko:
I need to think about it

I dont like that we depend on such a version, that is managed by you instead its part of PF Code.
Its quite hacky

melloware:
yeah it was a public version 0.1.0 that @blutorange broguht in in the TypeScript rewrite: https://github.com/gabceb/jquery-browser-plugin

before we had this plugin locally in PF code.

melloware:
in 15 we had the code locally: https://github.com/primefaces/primefaces/blob/15.X/primefaces/src/main/resources/META-INF/resources/primefaces/jquery/jquery.browser.js

tandraschko:
We should go back to have it local then IMO, if its not maintained anymore

melloware:
I like that idea I will update this PR

melloware:
Restored this to use a local file we control

PR #14498, opened 2025-12-06
Fix #14479: TabView add new widget selectTabById(id) method

### LINKED ISSUE #14479: TabView: add select tab by id function
### Description

TabView and other tab components have a select function to select the tab by index.

Often its easier to select the tab by its id, for example using URL hash navigation.

Below is the code which currently converts the id of a tab to its index.

If i open the below site like `text.xhtml#tabFoo` it will select tab-foo client side.

It supports full id of a tab `myForm:tabView:tabFoo` or just 'tabFoo' as argument so the caller can decide.

```
<script type="text/javascript">
		
						$(document).ready(function() 
			{
				onUrlHashChangeAccount();
			});
			
			$(window).on('hashchange', onUrlHashChangeAccount);
			
			function onUrlHashChangeAccount()
			{
				// select tab
				var idTabView = "myForm:tabView"

				// #myForm:tabView:tabFoo
				// #tabFoo
				var urlHash = location.hash.slice(1);
				
				tabViewSelectTabById(idTabView, urlHash, true);
			}
		
			
			/**
			 * Selects the tab with the given id of the given tab-view.
			 *
			 * @param idTabView Selector of tab-view
			 * @param idTab ID of the tab to select. Can start with id of tabView.
			 * @param silent Silent param of TabView.select(int, bool) function.
			 */
			function tabViewSelectTabById(idTabView, idTab, silent)
			{
				var idTabViewEscaped = PrimeFaces.escapeClientId(idTabView);
				
				var tabsPanels = $(idTabViewEscaped + ' > .ui-tabs-panels');

				if(idTab != null)
				{
					// support full id tabView:tabFoo
					if (idTab.startsWith(idTabView))
					{
						idTab = idTab.substring(idTabView.length);
						if(idTab.startsWith(":"))
							idTab = idTab.substring(1);
					}
					
					if (idTab != "")
					{
						var tabIndex = $(tabsPanels).find(idTabViewEscaped + "\\:" + idTab).index();
		
						if (tabIndex != -1)
							PrimeFaces.getWidgetById(idTabView).select(tabIndex, silent);
					}
				}
			}
			
		</script>
```

Add function to tab components client widget: `selectById(idTab, silent)`

### Additional context

The tabs also have anchor href tags with the url-hash but somehow it does not gets reflected in the URL. 

Both things in combination would add browser history-back  / forth support to  tab-view  / tab-components.

melloware:
Fix #14479: TabView add new widget selectTabById(id) method

```js
// Get the widget instance
var tabView = PrimeFaces.getWidgetById('myTabView');

// Select by full ID
tabView.selectTabById('#form:myTabView:tab1');
tabView.selectTabById('form:myTabView:tab2');

// Select silently (no events)
tabView.selectTabById(''form:myTabView:tab2', true);
```

### DIFF
--- a/primefaces/src/main/frontend/packages/components/src/tabview/tabview.widget.js
+++ b/primefaces/src/main/frontend/packages/components/src/tabview/tabview.widget.js
@@ -856,4 +856,31 @@ PrimeFaces.widget.TabView = class TabView extends PrimeFaces.widget.DeferredWidg
         PrimeFaces.invokeDeferredRenders(this.id);
     }
 
+    /**
+     * Selects the tab with the given ID. The ID can be a full client ID (e.g., "form:tabView:tabFoo")
+     * or a partial ID relative to this tab view (e.g., "tabFoo" or ":tabFoo").
+     * @param {string | HTMLElement | JQuery | undefined | null} tabId Client ID of the tab to select.
+     * @param {boolean} [silent=false] Controls whether events are triggered. When `true`, the
+     * `onTabChange` callback and AJAX events are not fired.
+     * @return {boolean} `true` if the tab was found and selected, `false` otherwise.
+     */
+    selectTabById(tabId, silent) {
+        if (!tabId) {
+            return false;
+        }
+
+        const tabPanel = PrimeFaces.expressions.SearchExpressionFacade.resolveComponentsAsSelector(this.jq, tabId);
+        if (tabPanel.length === 0) {
+            return false;
+        }
+
+        // Get the index of the panel within the panelContainer
+        const tabIndex = this.panelContainer.children().index(tabPanel);
+        if (tabIndex === -1) {
+            return false;
+        }
+
+        return this.select(tabIndex, silent);
+    }
+
 }
--- a/primefaces/src/main/frontend/packages/core/src/core/core.expressions.ts
+++ b/primefaces/src/main/frontend/packages/core/src/core/core.expressions.ts
@@ -37,7 +37,8 @@ export class SearchExpressionFacade {
 
                 // just a id
                 if (expression.indexOf("@") == -1) {
-                    const element = document.getElementById(expression);
+                    const cleanedExpression = expression.startsWith("#") ? expression.substring(1) : expression;
+                    const element = document.getElementById(cleanedExpression);
                     if (element !== null) {
                         elements = elements.add($(element));
                     }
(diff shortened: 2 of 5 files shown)


tandraschko:
Will review next days

tandraschko:
im not sure in generel
this code doesnt use our expressions, not jquery. so its a third way of resolving a component somehow.

also the 
`tabView.selectTabById(':tab1');`
reuses a pattern from the server side, which is actually absolut, not relative. we must not support this.

i really suggest to rely on search expressions (SearchExpressionFacade) here, which means:
`
tabView.selectTabById('#form:myTabView:tab1');
tabView.selectTabById('form:myTabView:tab2');
`
should work and
`tabView.selectTabById('@this:tab1');`

but it also means, we have to enhance the client side SearchExpressionFacade, as it does not support nesting yet like: @this:subId

melloware:
OK for now should I simplify and only allow the full ID like `tabView.selectTabById('#form:myTabView:tab1'); tabView.selectTabById('form:myTabView:tab2');` ?

tandraschko:
Yeah
Only full ID and reuse SearchExpressionFacade

melloware:
OK i did make 1 fix in SearchExpressionFacade so users can do either of these.

```js
tabView.selectTabById('#form:myTabView:tab1');
tabView.selectTabById('form:myTabView:tab2');
```

it was not working if it already had the `#` in front.

primefaces/src/main/frontend/packages/components/src/tabview/tabview.widget.js
```diff
@@ -856,4 +856,31 @@ PrimeFaces.widget.TabView = class TabView extends PrimeFaces.widget.DeferredWidg
...
+     * @param {boolean} [silent=false] Controls whether events are triggered. When `true`, the
+     * `onTabChange` callback and AJAX events are not fired.
+     * @return {boolean} `true` if the tab was found and selected, `false` otherwise.
+     */
+    selectTabById(tabId, silent) {
+        if (!tabId) {
+            return false;
+        }
+
+        const tabPanel = PrimeFaces.expressions.SearchExpressionFacade.resolveComponentsAsSelector(this.panelContainer, tabId);
```
tandraschko: better this.jq?

primefaces/src/main/frontend/packages/core/src/core/core.expressions.ts
```diff
@@ -59,8 +60,7 @@ export class SearchExpressionFacade {
                 // PFS
                 else if (expression.indexOf("@(") == 0) {
                     //converts pfs to jq selector e.g. @(div.mystyle :input) to div.mystyle :input
-                    elements = elements.add(
-                            $(expression.substring(2, expression.length - 1)));
+                    elements = elements.add($(expression.substring(2, expression.length - 1)));
```
tandraschko: please dont change code styles in such PRs

melloware:
reverted that formatting change and updated to `this.jq`

PR #14527, opened 2025-12-17
Fix #14526: ColorPicker remove dead constants

### LINKED ISSUE #14526: ColorPicker: labels not used?
public static final String CLEAR_LABEL = "primefaces.colorpicker.CLEAR";
    public static final String CLOSE_LABEL = "primefaces.colorpicker.CLOSE";

@melloware maybe because your CDK refactoring? otherwise we should remove it and clean docs

melloware:
Fix #14526: ColorPicker remove dead constants


tandraschko:
did you check the docs? just saw in the file search

melloware:
Found it in docs and searched everywhere else and I don't see any more instances.

PR #14530, opened 2025-12-18
Fix #14055: headers update after column dropping, duplicate columntoggler state

### LINKED ISSUE #14055: DataTable column reorder breaks filtering and sorting
### Describe the bug

When the column ordering has been changed on a DataTable, filtering and sorting no longer function.
The same defect was already reported within https://github.com/primefaces/primefaces/issues/8159.

### Reproducer

Reproducible in the Primefaces Showcase DataTable Dynamic Columns page:
https://showcase.primefaces.org/ui/data/datatable/columns.xhtml

1. Go to https://showcase.primefaces.org/ui/data/datatable/columns.xhtml
2. Try filtering by "activity", filtering works.
3. Clear the "activity" filtering.
4. Move the "activity" column to be the first column, and click "Update" to refresh the table.
5. Try filtering or sorting any column
6. Note that neither filtering nor sorting function properly.

### Expected behavior

Filter and sort should work after column ordering has changed.

### PrimeFaces edition

Community

### PrimeFaces version

15.0.6

### Theme

_No response_

### JSF implementation

All

### JSF version

-

### Java version

-

### Browser(s)

_No response_

vonnai:
1. headers should be updated after column order is changed, otherwise sorting does not work well
2. columntoggler state should not be duplicated with the same ID. There must be the one.
Fix #14055

### DIFF
--- a/primefaces/src/main/frontend/packages/components/src/columntoggler/columntoggler.widget.js
+++ b/primefaces/src/main/frontend/packages/components/src/columntoggler/columntoggler.widget.js
@@ -98,7 +98,7 @@ PrimeFaces.widget.ColumnToggler = class ColumnToggler extends PrimeFaces.widget.
         this.itemContainer = this.panel.children('ul');
 
         var stateHolderId = this.tableId + "_columnTogglerState";
-        this.togglerStateHolder = $(stateHolderId);
+        this.togglerStateHolder = $(PrimeFaces.escapeClientId(stateHolderId));
         if (this.togglerStateHolder.length === 0) {
             this.togglerStateHolder = $('<input type="hidden" id="' + stateHolderId + '" name="' + stateHolderId + '" autocomplete="off"></input>');
             this.table.append(this.togglerStateHolder);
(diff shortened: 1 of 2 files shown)


primefaces/src/main/frontend/packages/components/src/columntoggler/columntoggler.widget.js
```diff
@@ -98,8 +98,11 @@ PrimeFaces.widget.ColumnToggler = class ColumnToggler extends PrimeFaces.widget.
         this.itemContainer = this.panel.children('ul');
 
         var stateHolderId = this.tableId + "_columnTogglerState";
-        this.togglerStateHolder = $('<input type="hidden" id="' + stateHolderId + '" name="' + stateHolderId + '" autocomplete="off"></input>');
-        this.table.append(this.togglerStateHolder);
+        this.togglerStateHolder = $(stateHolderId);
```
melloware:
```suggestion
        this.togglerStateHolder = $(PrimeFaces.escapeClientId(stateHolderId));
```

melloware (changes requested):
One comment shouldn't that be escaped so it s `#` ?

melloware (approved):

PR #14536, opened 2025-12-21
CDK: StyleAware

melloware:
Toolbar, Tooltip, Wizard refactor


tandraschko:
thanks!

i just wonder: should every component just have a style/styleclass? or is there one without?
we could just add it to PrimeComponent then

melloware:
I can't imaging there are too many without it but I supposed things like p:poll don't have it?

tandraschko:
yep, true
poll is one of them

PR #14542, opened 2025-12-23
CDK: Barcode

melloware:


tandraschko:
Looks good
Please double check description and Default value by yourself

PR #14544, opened 2025-12-23
CDK: BockUI

melloware:


tandraschko:
Looks good
We should Just acoid descriptions like "Default is 0..", we have DefaultValue for this
And we should write AJAX all upper case in description?

Please double check description and Default value by yourself

PR #14545, opened 2025-12-23
CDK: Menus

melloware:
This will do all the menus like BreadCrumb, ContextMenu etc


tandraschko:
Same Here
Please avoid description with: Default is ...

Please double check Default values and description

melloware:
Will do I have it pulling from Taglib.xml I will clean up the descriptions.

PR #14546, opened 2025-12-24
CDK: Fix Message

melloware:
Integrations tests for CSV were failing

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/message/MessageBase.java
+++ b/primefaces/src/main/java/org/primefaces/component/message/MessageBase.java
@@ -63,7 +63,7 @@ public abstract class MessageBase extends UIMessage implements UINotification, W
     @Override
     public abstract boolean isShowSummary();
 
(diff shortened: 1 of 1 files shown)


NicolaIsotta:
Hi @melloware , seems like this PR changes the default value of the attribute `showDetail` from 15.X. (where it was false)
Should this be mentioned in the migration guide?

cc @tandraschko

melloware:
hmmm this was weird because the integration tests were failing from 15.x to 16.x  because all the details were not being displayed.  But the XHTML was the same. Weird right?

melloware:
@NicolaIsotta can you submit a PR flipping it back and we can see all the IT's failing again?

NicolaIsotta:
I think I found the reason. In 15.x showSummary was false and showDetail was true. So showDetail defaulting to true is ok, showSummary must be inverted instead.

NicolaIsotta:
created PR #15039

PR #14547, opened 2025-12-24
CDK: Cache

melloware:


tandraschko:
I think you can use implicitDefaultValue instead of the text

melloware:
Yes!   Done.

PR #14553, opened 2025-12-30
Fix #14515: TreeNode uses wrong var in styleClass

### LINKED ISSUE #14515: Horizontal Tree treeNode references wrong var value
[horiz-tree.zip](https://github.com/user-attachments/files/24166273/horiz-tree.zip)

### Describe the bug

In PF 15 there seems to be wron value of var in p:treeNode. Inside of p:treeNode var references to correct value but when var is used in p:treeNode it seems to be previous value. 

### Reproducer

There is simple test project which shows problem. Name of style class is taken inside of node but style class is attached to treeNode.

### Expected behavior

Expected behaviour is that name of class and class should be same.

### PrimeFaces edition

Community

### PrimeFaces version

15.0.10

### Theme

_No response_

### JSF implementation

MyFaces

### JSF version

4.0.2

### Java version

JDK 21

### Browser(s)

all

melloware:
Fix #14515: TreeNode uses wrong var in styleClass


melloware:
Had to move this code down below where the new VAR is set.

PR #14558, opened 2026-01-02
CDK: Drag and Drop

melloware:


tandraschko:
Looks good
Please double check descriptions + Default values and feel free to merge

melloware:
I will check the jquery docs I think I can add some
Implicit default values here too.

PR #14570, opened 2026-01-07
CDK: Dialog

melloware:


primefaces/src/main/java/org/primefaces/component/dialog/DialogBase.java
```diff
@@ -82,270 +66,99 @@ public String getFamily() {
...
-    public void setClosable(boolean closable) {
-        getStateHelper().put(PropertyKeys.closable, closable);
-    }
+    @Property(description = "Height of the dialog.")
+    public abstract String getHeight();
 
-    public String getOnShow() {
-        return (String) getStateHelper().eval(PropertyKeys.onShow, null);
-    }
+    @Property(defaultValue = "-2147483648", description = "Minimum width in pixels.")
```
tandraschko: did you try
defaultValue = "Integer.MIN_VALUE"?

melloware: let me try it i was afriad it wouldn't work

melloware: that does work and I fixed two other places that i was using `-2147483648`

Issue #13326, opened 2025-02-13
Datatable: Nested datatable selection issue


jansohn:
### Describe the bug

Selection on nested datatable always returns `null` except for last row. Seems to be related to https://github.com/primefaces/primefaces/issues/3812 but there is still an issue I think...

### Reproducer

https://github.com/jansohn/primefaces-test/tree/nested-datatable-selection

1. Clone reproducer
2. Checkout branch `nested-datatable-selection`
3. Run with mvn jetty:run
4. Open http://localhost:8080/ in browser
5. Open Songs in first row
6. Selecting any of the song rows will always return `null`
7. Open Songs in last row
8. Selecting any of the song rows will correctly return the selected song

### Expected behavior

Selecting the rows in the nested datatable should work for all parent datatable rows

### PrimeFaces edition

Community

### PrimeFaces version

14.0.11

### JSF implementation

Mojarra

### JSF version

4.0

### Java version

21

tandraschko:
TBH im not sure if anyone will work on this.
Its a propably a deep bug inside UIData and state saving.

I would be great if you can work on it by yourself and provide a PR.

melloware:
I agree with @tandraschko and these types of UI's are overly complex for users.  I would find a way to simplify with a master detail type where they click on the row in the table and you have a "detail" below with your selection table rather that trying to embed this in ROWS.

Plus its less HTML in the browser.

melloware:
added this to the Sponsoring required DataTable ticket: https://github.com/primefaces/primefaces/issues/11251

jansohn:
Quite disappointing that nested datatables are pretty much unsupported then with PrimeFaces. I don't think it is overly complex to show a second datatable inside of a row expansion or an overlay panel, especially for datatables with a bigger page size. Showing the second datatable below the first datatable might force the user to scroll quite a bit (if they even realize and see it).

I've tried to workaround the problem by saving the list for the nested datatable in a separate field when opening the overlay panel but that did not help either. The `selection` field was still `null`. I could see from debugging that it actually first set the `selection` field to `null` before even retrieving the selected object in `DataTable.getRowData (String rowKey)`.

melloware:
@jansohn its most likely the table within table with multiple `var` attribute outer and inner that is throwing off the `selection`.  Its quite complex.

melloware:
@jansohn fixed your reproducer: 

[pf-13326.zip](https://github.com/user-attachments/files/24799776/pf-13326.zip)

[...]

Issue #3535, opened 2018-03-27
Datatable: selectAllRows(), unselectAllRows() partly broken


Hamsterbau:
## 1) Environment
- PrimeFaces version: 6.2.1
- Application server + version: Wildfly 11
- Affected browsers: Firefox (perhaps others)

## 2) Steps to reproduce

1. Click on checkbox in header (calls internally selectAllRows()
2. Unselect one checkbox from one row
3. Nothing happens, as well as pagination and other ajax stuff is broken.
4. After waiting and click another time on row or checkbox it will be unselected

--> More entries means more waiting. It is really slow for lists with > 20k entries.

## 3) Expected behavior

3. Row will be unselected, and objects in selected list will be minus one.

## 4) Sample XHTML

```xhtml
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml"
      xmlns:f="http://xmlns.jcp.org/jsf/core"
      xmlns:p="http://primefaces.org/ui"
      xmlns:h="http://xmlns.jcp.org/jsf/html">

    <h:head />

    <h:body>
        <h:form id="form">
            <p:dataTable
                    id="myTable"
                    widgetVar="myWidget"
                    var="entry"
                    rowKey="#{entry.name}"
                    selection="#{selectableTable.selected}"
                    rows="5"
                    rowSelectMode="add"
                    paginator="true"
                    value="#{selectableTable.entries}">
                <p:ajax event="rowSelect" listener="#{selectableTable.rowSelectCheckbox}" update=":form:selected" />
                <p:ajax event="rowUnselect" listener="#{selectableTable.rowUnselectCheckbox}" update=":form:selected" />
                <p:ajax event="rowSelectCheckbox" listener="#{selectableTable.rowSelectCheckbox}" update=":form:selected" />
                <p:ajax event="rowUnselectCheckbox" listener="#{selectableTable.rowUnselectCheckbox}" update=":form:selected" />
                <p:ajax event="toggleSelect" listener="#{selectableTable.toggleSelected}" update=":form:selected"/>

                <p:column selectionMode="multiple" />

                <p:column sortBy="#{entry.name}">
                    <f:facet name="header">
                        <h:outputText value="Name"/>
                    </f:facet>

                    <h:outputText value="#{entry.name}"/>
                </p:column>

            </p:dataTable>

            <h:outputText id="selected" value="Selected: #{selectableTable.selected.size()}" />
        </h:form>
    </h:body>

</html>
```

[...]

PR #14577, opened 2026-01-09
LazyDataModel: using getRowKey and getWrappedData functions to provide default for getRowData

### LINKED ISSUE #8110: Datatable: Lazy with multi selection is not keeping selected rows when paginating
**Describe the defect**
After updating from Prime 7 to Prime 11 the datatable is not being able to keep the selected rows after paginate.

**Environment:**
 - PF Version: _11 RC2_
 - JSF Mojarra 2.2.12 
 - Affected browsers:  ALL_

**To Reproduce**

2. select any row on the table
3. go to second page
4. go back to the first page

**Expected behavior**
The row that was selected on the previously should be marked as selected
however that is not happening.

This used to work just fine with primefaces 7
I noticed while debugging that the get method that keeps the selected rows is not being called in prime 11
"public List<FakeDto> getSelecteds() {"

To show that I created two reproducers, one with prime 7 and another with prime 11RC2
[primefaces-test-master.zip](https://github.com/primefaces/primefaces/files/7597126/primefaces-test-master.zip)
[primefaces-test-master7.zip](https://github.com/primefaces/primefaces/files/7597127/primefaces-test-master7.zip)

### LINKED ISSUE #11556: DataTable: checkbox unselects items from other pages
### Describe the bug

From https://stackoverflow.com/q/78023603/4605161

> In a table with pagination and checkbox selection, when a row is unchecked, the items from other pages are unchecked as well

### Reproducer

> 
> This scenario can be seen in the showcase, in the last table, the checkbox table: http://www.primefaces.org:8080/showcase/ui/data/datatable/selection.xhtml
> 
> This is the sequence to reproduce the problem:
> 
> 1. Check all rows (checkbox "Code" on the table header)
> 2. Navigate through the pages, to confirm that all items are checked.
> 3. Unselect one item (no matter which page)
> 4. Navigate to the other page. All items are unselected
> 
> Is this the correct behavior? I think it can be confusing for the user, who marked all and assumes that all of them are selected, except the one that was unchecked.
> 
> Is there a way to work around this ?
> 
> I already tried setting :
> 
> ```
> selectionPageOnly="false"
> rowSelectMode="add"
> ```
> 
> I also tried to use the selection mode without checkboxes, same behavior.
> 
> Primefaces version: 13.0.5 JSF version: 2.3
> 
> Thanks in advance

### Expected behavior

Selection in other pages should be kept

### PrimeFaces edition

Community

### PrimeFaces version

13.0.5

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

11

### Browser(s)

_No response_

### LINKED ISSUE #13640: DataTable: `selectionPageOnly="false"` issues
### Describe the bug

As I was investigating #13620 I noticed some other odd behaviors with how `selectionPageOnly="false"` works. I think a lot of it comes down to the JS not knowing _all_ the selectable rows so it can't know if they've all been selected or not (`<ID>_selection` hidden input is set to `value="@all"` to handle selecting unseen rows).

An option for better handling this would be to include a second hidden input in the DOM which contain all selectable row keys, then refactor the JS selection code to reference that if `selectionPageOnly="false"`.

1. Selecting all then deselecting a row deselects all unseen rows (i.e., rows on other pages)
2. Selecting all visible rows selects global checkbox (#13620, partially fixed in 15.0.3)
3. Selecting all after filtering to only enabled rows selects even disabled rows
4. Selecting all then paging selects even disabled rows

cc. @stolp @melloware 

### Reproducer

All scenarios can be reproduced using the same PF test: 
[primefaces-test-global-select.zip](https://github.com/user-attachments/files/19496016/primefaces-test-global-select.zip)

**Selecting all then deselecting a row deselects all unseen rows (i.e., rows on other pages)**

1. Check global select checkbox
   - All rows on all pages are selected
2. Deselect one row
   - All but one row _on current page_ are selected :x:

**Selecting all visible rows selects global checkbox (#13620, partially fixed in 15.0.3)**

<= 15.0.2
1. Select each visible row
   - global checkbox is checked :x:

= 15.0.3
1. Select each row on all pages individually
   - global checkbox remains unchecked :x:

**Selecting all after filtering to only enabled rows selects even disabled rows**

1. Check "Disable selections"
   - All albums prior to 2000 are disabled
2. Filter "Name" -> "Anti"
   - Only one row shows
3. Check global select checkbox
   - Both global select checkbox and single row checkbox are checked
4. Remove filter
   - All rows, _even disabled ones_, are selected :x:
5. Uncheck global select
   - Only disabled rows are selected :x:

**Selecting all then paging selects even disabled rows**

1. Check "Disable selections"
   - All albums prior to 2000 are disabled
3. Check global select checkbox
   - Both global select checkbox and only enabled row checkboxes are checked
4. Go to table page 2
   - All rows, _even disabled ones_, are selected :x:
5. Uncheck global select
   - Only disabled rows are selected :x:

### Expected behavior

TBD

### PrimeFaces edition

Community

### PrimeFaces version

15.0.2/15.0.3

### Theme

n/a

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

21

### Browser(s)

Chrome

### LINKED ISSUE #14538: LazyDataModel: getRowData does not use getRowKey function
### Describe the bug

When developing custom LazyDataModels, it's important to understand that merely overriding `getRowKey` is not enough and raising UnsupportedOperationException. Both `getRowData` and `getRowKey` utilize a converter for data retrieval. Instead of throwing an exception if this converter is not provided, why not using the `data` member to get the information directly? An implementation could look like the following:
```java
      if (rowKeyConverter != null) {
          return rowKeyConverter.getAsObject(context, UIComponent.getCurrentComponent(context), rowKey);
      }
      // Reusing getRowKey to retrieve rowData
      return Optional.ofNullable(getWrappedData()).orElse(new ArrayList<>()).stream()
        .filter(item -> rowKey.equals(getRowKey(item)))
        .findFirst()
        .orElse(null);
  }
```

### Reproducer

```java
public class DataModelDynamic<T> extends LazyDataModel<T> {

  private final transient Function<T, String> rowKeyFn;

  /**
   * Construct a DataModelDynamic based on a loading function.
   *
   * @param rowKeyFn Function to transform the model data type to a key value.
   */
  public DataModelDynamic(Function<T, String> rowKeyFn) {
    if (rowKeyFn == null) {
      throw new RuntimeException("Datamodel is missing a function for extracting rowKeys.");
    }
    this.rowKeyFn = rowKeyFn;
  }

  @Override
  public List<T> load(
      int first, int pageSize, Map<String, SortMeta> sortBy, Map<String, FilterMeta> filterBy) {
    return List.of(); // loading data here
  }

  @Override
    // According to JavaDoc of Primefaces in load-function:
    // if loading and counting is done in one step, this should return 0 here
    return 0;
  }

  @Override
  public String getRowKey(T obj) {
    return rowKeyFn.apply(obj);
  }
```

When selecting a row on that table, which has an selectEvent-Handling leads to UnsupportedOperationException.

### Expected behavior

Overriding `getRowKey` should be enough so that the data of a row could be loaded in custom LazyDataModels.

### PrimeFaces edition

Community

### PrimeFaces version

15.0.10

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0

### Java version

17

### Browser(s)

_No response_

AndreasIgelCC:
PullRequest for the change discussed in #14538

Fix #13640
Fix https://github.com/primefaces/primefaces/issues/8110
Fix https://github.com/primefaces/primefaces/issues/11556


melloware:
looks like some IT failures

```
DataTable006Test.lazySelectAllRows:302->assertSelections:385 expected: <2,3> but was: <>
Error:    DataTable006Test.lazySelectionMultipleCheckboxesWithPagination:128 expected: <true> but was: <false>
Error:    DataTable006Test.lazySelectionWithFiltering:327->assertSelections:385 expected: <1,3> but was: <>
Error:    DataTable036Test.multiViewStateSelection:56->testImplementationMultiViewStateSelection:79 expected: <true> but was: <false>
```

AndreasIgelCC:
Yes, there appears to be a change in the selection behavior. I’ll take a closer look to understand what exactly is happening and which feature might be affected by this change.
Thank you, @melloware, for starting the integration tests and for providing the hint that helped identify where to investigate potential side effects!

AndreasIgelCC:
I had a deeper look into the problem and my change results in failing integration tests because the way Selection is been handled. I am working on a solution now.

### Root Cause Analysis
1. JSF Lifecycle Timing Mismatch in postbacks between:
    - **Decode Phase** (APPLY_REQUEST_VALUES): Selection decoding happens here
    - **Model Loading** (PROCESS_VALIDATIONS/UPDATE_MODEL): LazyDataModel loads data here
2. Technical Details:
    - SelectionFeature.decodeMultipleSelection() calls table.getRowData(rowKey) during decode
    - For lazy tables, getRowData() fails because wrappedData is null (model not loaded yet)
    - This causes selection restoration to fail silently
    - **Critical**: User selections are lost on filtering/paging/sorting interactions

###  Solution Approach: DeferredSelection Pattern
Postpone rowKey-to-object translation until LazyDataModel is loaded while preserving the JSF lifecycle flow:
- Store rowKeys during decode phase (no data access required)
- Convert to data objects when LazyDataModel is loaded
- Loading triggert in existing loadLazyDataIfRequired()

melloware:
@AndreasIgelCC great debugging!  I would say try your solution we have the Integration Test Coverage to prove whether its working or not!

AndreasIgelCC:
@melloware While working on this fix, I noticed several issues related to selection, especially when combined with lazy loading. As can be seen in #3535 and #6730, this topic has already received a lot of thoughtful discussion.
At this point, aligning the behavior with the expectations described in #3535 and in this comment on #6730 (https://github.com/primefaces/primefaces/issues/6730#issuecomment-787836305) is significantly easier than preserving the current behavior for lazy tables.
Could you please have a look at the updated test expectations and let me know if they align with your expectations? I’d also appreciate your thoughts on whether this behavior change makes sense overall. Of course, feedback on the other changes is very welcome as well, and I’d be glad to know whether this general approach of changing the behavior is acceptable.

melloware:
Running tests now will review in the morning.

AndreasIgelCC:
The failure is related to Csv003Test.datePastInvalid. Do you have any idea why this test is failing now?
As far as I can tell, it doesn’t seem to be related to Datatables.
I updated the branch, and I was wondering whether this might be caused by a change from master that hasn’t been integrated yet. However, even with that update, the test is still passing locally for mojarra‑4.0 / Firefox.

Update: Integration tests now running through, so updating branch solved it.

melloware:
@AndreasIgelCC can you check i wonder if this also fixes the following:

https://github.com/primefaces/primefaces/issues/13640

https://github.com/primefaces/primefaces/issues/8110

https://github.com/primefaces/primefaces/issues/11556

AndreasIgelCC:
> @AndreasIgelCC can you check? I wonder if this also fixes the following:

Of course.

> #13640

This issue actually combines multiple bug reports (ideally, each bug should be tracked in a separate ticket).  
Maybe @FlipWarthog could split this into individual issues and I am adding integration tests in this PR to verify each one?

Most of the reported problems should already be fixed by this change—or may have been resolved by other recent fixes.

However, the following behavior will **not** change and will still not work as described:

- Selecting all visible rows automatically selects the global checkbox

To support this, the implementation would need to always load *all* datasets from the lazy data model, which would introduce significant overhead—especially for database-backed scenarios. For that reason, I wouldn’t consider this a desired or supported feature.

> #8110  
> #11556

Both of these issues should be resolved by the current change.  
That said, we should extend the test coverage to explicitly include these scenarios to ensure the behavior is properly validated and protected going forward.

melloware:
Excellent i linked those tickets to this PR

melloware:
@AndreasIgelCC can you also check this one: https://github.com/primefaces/primefaces/issues/13326  i am wondering if your selection refactor has also fixed this.

AndreasIgelCC:
> @AndreasIgelCC can you also check this one: #13326 i am wondering if your selection refactor has also fixed this.

For this we really need to have an integration test. Nested tables I did not test in context of this MR.

AndreasIgelCC:
> Excellent i linked those tickets to this PR

@melloware Do you think anything needs to be added (for example additional tests or extending tests), or is this fine as it is?

melloware:
let me do some testing with your PR.

melloware:
I have restested all the reproducers in the above tickets and they are all fixed!!!  Awesome job!

melloware:
@tandraschko this PR is great and i manually tested all of the bugs above and they are all resolved with this PR and our bad integration tests are also fixed.

FlipWarthog:
@melloware Including everything in #13640?

@AndreasIgelCC Agreed in general that each issue should have its own ticket, but we'd created #13640 as a tracking ticket without immediate intention to fix; if all of them are fixed by this we can probably leave it as is, but if only some are I can definitely split things out to be more accurate.

melloware:
@FlipWarthog yep I just tested all your scenarios and they all pass.

AndreasIgel:
I was not sure with the topic of "Selecting all visible rows automatically selects the global checkbox", so I tested it on my own. For selectionPageOnly=false, this would be a bug from my perspective. But it is limited to selectionPageOnly=true and so it is  fine for me too!


@melloware: Thank you for linking the other bugs and retesting all of them!

melloware:
yeah that one may be an issue but to me its a minor UI glitch that is not a major issues. THe issues YOU fixed are the important ones

melloware (approved):

AndreasIgel:
> yeah that one may be an issue but to me its a minor UI glitch that is not a major issues.

There is no issue on it, not even a glitch. This is good exactly the way it did before.

> THe issues YOU fixed are the important ones

I worked on the solution with the help of AI-assisted coding tools, so the result wasn’t achieved by me alone.

tandraschko:
i will review the next weeks
currently i have some other tasks on my backlog and this is quite a core change.

AndreasIgel:
This is a relatively large change, so I’d be grateful for a careful review. As this is my first contribution to the framework, your experience with its edges and corner cases would be especially valuable. Please feel free to take whatever time you need.

AndreasIgelCC:
After merging in the master branch, the integration-tests on my local machine stop working. Try to understand what breaks them and how to solve that.

tandraschko:
propably related to CDK changes
we have like ~20 failing tests in master

AndreasIgelCC:
> propably related to CDK changes we have like ~20 failing tests in master

Thank you for that hint, I ran the test on master too and the results are identical to this feature branch:

Metric | Feature Branch | Master Branch
-- | -- | --
Tests Run | 840 | 839
Failures | 34 | 34
Errors | 196 | 196
Failed Test Classes | 91 | 91

Run-Command: `mvn verify -Pintegration-tests,parallel-execution,mojarra-4.0,firefox,headless`

How to proceed, should I wait for fixing the master branch?

tandraschko:
yes

tandraschko:
we are ready now, can you rebase and we can rerun tests

melloware:
I have updated the branch and tests running now

AndreasIgelCC:
> we are ready now, can you rebase and we can rerun tests

Sorry, I had been in a full day meeting today. Thank you @melloware  for updating the branch! Looks like the branch of the fork has not been in sync, I updated it.

AndreasIgelCC:
Hi @tandraschko, the PR is ready for your review. Let me know if anything else is needed.

tandraschko:
Its still missing the @Property change?

melloware:
@tandraschko which change I just rebased his PR from main?

tandraschko:
Im talking about my review point here

AndreasIgelCC:
> Its still missing the @Property change?

@tandraschko I didn't receive a review comment notification here - it may not have been published yet. However, I proactively reviewed the changes and noticed that while I updated the documentation in datatable.md, I hadn't updated the corresponding @Property annotation description for selectionPageOnly. I've now fixed this and pushed the update. Please let me know if there are any other concerns.

melloware (approved):

AndreasIgelCC:
@melloware: we now have an integration test failing for one environment (Java 25 and mojarra-4.1), this should have nothing todo with the changes here on datatable or am I wrong on this?

> AccordionPanel002Test.test:49 expected: <0> but was: <1>

melloware:
I will re run sometimes it's just a flickering result.

melloware:
@tandraschko all passing now if you want to review I can squash

tandraschko:
if this PR fixes those 3 issues - are those 3 issues already tested by your changes of the Test class?

melloware:
i think the test he fix covers all of them except for https://github.com/primefaces/primefaces/issues/11556 i think

melloware:
THat being said i tested this with those other tickets myself manually with their reproducers.

tandraschko:
i would like to have a test here
DataTable is by far the most complex component, we should cover it

AndreasIgelCC:
> i would like to have a test here DataTable is by far the most complex component, we should cover it

I completely agree that DataTable's complexity requires comprehensive test coverage. 

A few clarifying questions @tandraschko:
- Should I add integration tests for all related issues (#13640, #8110, #11556) in this PR?
- Or would you prefer separate PRs for each bug fix with dedicated tests?
- Are there any other aspects of this change that need to be addressed?
 
Happy to proceed with whichever approach you think is best for the project.

tandraschko:
1) yes pls
2) please put it inside this one
3) will review soon and maybe i have some comments

melloware:
@AndreasIgel that would be great and you can use the reproducers in the tickets to see what was failing.

AndreasIgelCC:
I've added integration tests for bugs #8110, #11556, and #13640, @melloware. Please review.

Regarding bug #8110: The existing ProgrammingLanguageLazyDataModel doesn't override getRowData(String rowKey) anymore, so it uses the new default implementation from this PR. The lazy selection tests (Orders 11-12) already validate this fix correctly. Nevertheless I added two tests, which test the behaviour described in the bug.

melloware (approved):

tandraschko:
i wonder if there is a way to implement "Select All" without loading ALL entities

for big tables, this means that thousands of entities are loaded and mapped. thats not good. 
how was it before?

tandraschko:
Propably we only need to load all, when there is a associated `selection` ValueExpression. For the DataTable itself, it would be enough to remember that all/* was selected.

melloware:
@tandraschko yeah i don't think the selectAll ever worked properly which is described in https://github.com/primefaces/primefaces/issues/13640  so i think its never worked properly.

tandraschko:
yeah
but with this PR, it becomes a bit messy feature
it will load always all objects via lazyDataModel - doesnt matter if the selection is mapped via EL or not.

AndreasIgelCC:
> yeah but with this PR, it becomes a bit messy feature it will load always all objects via lazyDataModel - doesnt matter if the selection is mapped via EL or not.

Thank you for raising this, @tandraschko! You're right that this is an important consideration.

Is it common to use selectAll with LazyDataModel when selection isn't mapped via EL? If so, what would be the expected behavior?

I see two options:
1. Add a warning in documentation that selectAll with lazy loading may trigger a full data load
2. Skip the load-all behavior when no EL mapping is detected (though this would require more changes to existing behavior)

I'm happy to adjust based on your preference. What approach would you recommend?

AndreasIgelCC:
> i wonder if there is a way to implement "Select All" without loading ALL entities
> 
> for big tables, this means that thousands of entities are loaded and mapped. thats not good. how was it before?

I could write some tests to validate that, if you want this.

tandraschko:
we should definitely try to do 2) 
if its not easy, we can still consider 1)

AndreasIgelCC:
> we should definitely try to do 2) if its not easy, we can still consider 1)

Ok, I am on that. But first trying to create tests to validate the behavior.

tandraschko:
any news here?

AndreasIgelCC:
> any news here?

Not yet, sorry 🙂 Right now it’s still unclear to me how the ID translation works and whether it can be postponed. I’m planning to dig deeper into it this or next week.

tandraschko:
I remove this unrelated advertising comment, sorry.

melloware:
@AndreasIgelCC any update?

AndreasIgelCC:
Sorry for the delay... I am on it.

melloware:
thanks i would love to get this landed!

AndreasIgelCC:
I’ve updated the docs and added assertions to the tests.

Next, I plan to add a test without `ValueExpression` (probably duplicating `DataTable006Test` into `DataTable049Test`).

Two quick questions where I’d appreciate your input, @melloware:

1. **Load expectations in `DataTable006Test`**  
   Does the way I integrated them look good, or would you suggest a different approach?

2. **Side effect visible in `DataTable006Test#lazySelectAllRows`**  
   Since the *Select All* checkbox stays active, all data gets loaded again on every page navigation and submit.  
   This feels technically correct (data might have changed), but could result in repeated data load. The only way to avoid it would be to uncheck the box after processing the selection — which would be a breaking change.

   What do you think—keep it as is or change the behavior?

melloware:
@AndreasIgelCC 

For (1) you are doing it totally fine!

As for (2) I would like @tandraschko to weigh in.  I know he will be against the multiple loads.

AndreasIgelCC:
> Propably we only need to load all, when there is a associated selection ValueExpression. For the DataTable itself, it would be enough to remember that all/* was selected.

@tandraschko I implemented this (decision based on the presence of a `ValueExpression`), but it’s quite tricky to properly test.

Short explanation:
* To have the select-all checkbox available, you need to use `selectionBox="true"` on a column. However, without a `ValueExpression` for the selection, the checkbox is never populated with the correct values.
* Because of that, I currently don’t see a good DataTable XHTML example that supports select-all without a `ValueExpression`.

Would you be able to provide an example XHTML (maybe based on `dataTable006.xhtml`) that I could use as a basis for tests?

Regarding @melloware’s feedback:
> As for (2) I would like @tandraschko to weigh in. I know he will be against the multiple loads.

Do you think we should change the behavior so that the select-all checkbox is automatically unchecked after processing the selection, or should we keep the current behavior despite the repeated loads?

AndreasIgelCC:
Hi @tandraschko,

is there any update on this PR and the next steps towards merging?  
Especially regarding the direction for the select-all behavior in lazy mode.

Thanks!

tandraschko:
i think we should merge for now
thanks for your effort!

AndreasIgelCC:
Thank you @tandraschko

PR #14586, opened 2026-01-15
Fix #14584: Paginator format integers for Locale

### LINKED ISSUE #14584: Datatable Paginator: Support thousand separators in currentPageReportTemplate numbers
### Description

Currently, the currentPageReportTemplate in PrimeFaces paginators renders numeric values (_{startRecord}_, _{endRecord}_, _{totalRecords}_, _{currentPage}_, _{totalPages}_) using plain integers without any formatting.

When working with large datasets (tens or hundreds of thousands of records), this significantly affects readability. For example:

Image

Would be much easier to read as:

Image

This could also be applied to page numbers.

Provide a built-in way to format numbers in _currentPageReportTemplate_, for example:

- Automatically applying locale-aware number formatting
- Or introducing an optional attribute, such as:
    - _currentPageReportFormat="number"_
    - or _paginatorNumberFormat="true"_
- Or allowing a _NumberFormat_ / locale hook to be used by the paginator

This would greatly improve usability and accessibility for applications that handle large volumes of data.

### Additional context

_No response_

melloware:
Fix #14584: Paginator format integers for Locale

### DIFF
--- a/primefaces/src/main/frontend/packages/components/src/paginator/paginator.widget.js
+++ b/primefaces/src/main/frontend/packages/components/src/paginator/paginator.widget.js
@@ -159,7 +159,7 @@ PrimeFaces.widget.Paginator = class Paginator extends PrimeFaces.widget.BaseWidg
         PrimeFaces.skinSelect(this.jtpSelect);
         this.jtpSelect.on('change', function(e) {
             if(!$(this).hasClass("ui-state-disabled")){
-                $this.setPage(parseInt($(this).val()));
+                $this.setPage($this.parseLocalizedInt($(this).val()));
             }
         });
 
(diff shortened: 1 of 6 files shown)


tandraschko:
so you just format a int to a correct string (1 -> 1, 1000 -> 1.000), right?
i wonder that nothing else is required on client side? or parsing the string back to int?

i didnt check the code, but it seems really only a few changes for such a complex component.

melloware:
you were right the UI was already doing `parseInt` on any number it needed to send to the backend and I thought ParseInt would handle it but i had to add this to strip out the locale "," or "." to make it an Int always

```js
function parseLocalizedInt(value) {
  return parseInt(value.replace(/[.,]/g, ''), 10);
}

parseLocalizedInt("1,000"); // 1000
parseLocalizedInt("1.000"); // 1000
parseLocalizedInt("1000"); // 1000
```

tandraschko:
if its well tested -> +1

PR #14612, opened 2026-01-23
CDK: MenuItem

melloware:


primefaces/src/main/java/org/primefaces/component/menuitem/UIMenuItemBase.java
```diff
@@ -89,332 +59,156 @@ public String getFamily() {
...
-
-    @Override
-    public String getUpdate() {
-        return (String) getStateHelper().eval(PropertyKeys.update, null);
-    }
-
-    public void setUpdate(String update) {
-        getStateHelper().put(PropertyKeys.update, update);
-    }
+    @Property(description = "Component(s) to update with ajax.")
```
tandraschko: i think those AjaxSource related properties are already defined in AjaxSource.java, we can just remove them here

melloware: OK let me try that

melloware: OK removed all the AjaxSource duplicates

PR #14621, opened 2026-01-26
Fix #14620: Menu store state correctly in local storage

### LINKED ISSUE #14620: Menu: toggle state not saved in localStorage
The **p:menu** component is not working properly. According to the documentation when **p:menu** has **toggleable** attribute set to TRUE the menu toggleable state should be stored per view/page, as it does with **p:panelMenu**.

I can see that **p:menu** is storing in the **local storage** of the browser which menus are **toggled** and **untoggled**, but when refreshing the page the status is not being applied to the component, and all menus are always **untoggled** regardless of what is stored in the **local storage**.

The example on the showcase page is not working either.

[p:menu](https://showcase.primefaces.org/ui/menu/menu.xhtml)

melloware:
Fix #14620: Menu store state correctly in local storage

@tandraschko this has been wrong for a long time as the backend Submenu contains an `expanded` value which is reused in different menus. This defaults to `false` but the client side menu was assuming everything was always expanded=true and needed to be collapsed in `restoreState`.

I have modifed the widget to respect the proper state of the header and set the localStorage properly and on restore state properly called either collapse or expand on the nodes.

### DIFF
--- a/primefaces/src/main/frontend/packages/components/src/menu/menu.plainmenu.widget.ts
+++ b/primefaces/src/main/frontend/packages/components/src/menu/menu.plainmenu.widget.ts
@@ -37,6 +37,11 @@ export class PlainMenu<Cfg extends PlainMenuCfg = PlainMenuCfg> extends Menu<Cfg
     menuitemLinks: JQuery = $();
 
     /**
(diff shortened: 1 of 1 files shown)


tandraschko:
code looks good
+1 when tested

PR #14645, opened 2026-02-01
Fix #14644: Drag and Drop doc updates

### LINKED ISSUE #14644: PF 15.0.2: droppable doesn't trigger listener
### Describe the bug

The following XHTML fragment:

        <h:form>
         <p:graphicImage id="doc" name="images/dvi.png"/>
         <p:draggable for="doc"/>
         <p:fieldset id="deletedDocs" legend="Recycle Bin" style="margin-top:20px">
           <p:outputPanel id="zone" styleClass="slot">
             <h:outputText value="Drop documents into the recycle bin to delete them"
                           style="font-size:20px;"/>
           </p:outputPanel>
        </p:fieldset>
        <p:droppable for="zone">
          <p:ajax listener="#{dragDrop.onDocumentDrop}"/>
        </p:droppable>
      </h:form>

is supposed to trigger the `onDocumentDrop()` method when the draggable is dropped into the droppable. However, the debug logging statement in this method isn't executed, showing that the listener isn't triggered.

[pf.zip](https://github.com/user-attachments/files/24991682/pf.zip)

### Reproducer

To reproduce the issue:

1. `$ unzip pf.zip`
2. `$ cd pf`
3. `$ mvn package jetty:run`
4. `$ open http://localhost:8080/test.xhtml`
5. Drag the small image in the upper-left corner onto the area labeled `Recycle Bin`
6. Observe that the debug logging statement in the `onDocumentDrop()` method isn't executed.

### Expected behavior

I expect that the listener be called when the draggabale is dropped onto the droppable.

### PrimeFaces edition

Community

### PrimeFaces version

15.0.2

### Theme

saga

### JSF implementation

Mojarra

### JSF version

14

### Java version

21

### Browser(s)

Firefox

melloware:
Fix #14644: Drag and Drop doc updates

### DIFF
--- a/docs/15_0_0/components/dragdrop.md
+++ b/docs/15_0_0/components/dragdrop.md
@@ -211,7 +211,7 @@ Following example shows how to enable draggable images to be dropped on droppabl
 
 ```xhtml
 <p:graphicImage id="messi" value="barca/messi_thumb.jpg" />
-<p:draggable for="messi" />
+<p:draggable for="messi"/>
 <p:outputPanel id="zone" styleClass="slot" />
 <p:droppable for="zone">
     <p:ajax listener="#{ddController.onDrop}" />
--- a/docs/15_0_0/components/dragdrop.md
+++ b/docs/15_0_0/components/dragdrop.md
@@ -180,7 +180,7 @@ specified with the for attribute.
 
 ```xhtml
 <p:outputPanel id="slot" styleClass="slot" />
-<p:droppable for="slot" />
+<p:droppable for="slot" scope="slot-scope"/>
 ```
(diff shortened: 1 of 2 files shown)


tandraschko:
Should it be a required attr?

melloware:
I will double check I thought it worked without it but I always declare scope whenever I use DnD.

melloware:
Actually it does work without it the reproducer had multiple issues and I fixed them all the scope is just for clarity.

tandraschko:
if its not required, i would not add for every example in the doc

melloware:
OK updated

PR #14671, opened 2026-02-09
Fix "undefined TypeError" triggered for some error handler cases

### LINKED ISSUE #14670: Ajax handleError() fails with undefined TypeError
### Describe the bug

When some Ajax errors are triggered, the PrimeFaces handleError fails with an "undefined TypeError", undefined has no method find().

### Reproducer

I am providing a PR that fixes the issue. It would be more effort for me to set up a reproducer than to fix the bug.

I built a local PrimeFaces JAR and tested it with my system and it solved the problem.

### Expected behavior

It should run the handleError function completely without failing with this error.

### PrimeFaces edition

Community

### PrimeFaces version

15.0.12

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3.21

### Java version

21

### Browser(s)

_No response_

danielthegray:
If `errorName` is falsy, `exceptionHandlers` is never initialized and causes a failure later.

This is fixed in the master branch, during the Typescript migration, this error was found and fixed, but it is still present in the pure JS version (and therefore also present on the latest 15.x release).

Fix #14670 

### Testing
I build a local PrimeFaces JAR and integrated it into my project and the error was gone.

### DIFF
--- a/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
+++ b/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js
@@ -380,9 +380,10 @@ if (!PrimeFaces.ajax) {
              * @param {string} errorMessage The error message.
              */
             handleError: function(errorName, errorMessage) {
+                var exceptionHandlers = [];
                 if (errorName) {
                     // try to invoke specific AjaxExceptionHandler
-                    var exceptionHandlers = PrimeFaces.getWidgetsByType(PrimeFaces.widget.AjaxExceptionHandler);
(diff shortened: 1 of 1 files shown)


tandraschko:
Is master also affected?

melloware:
@tandraschko no it looks like master had this fix from the TypeScript refactor


```ts
let exceptionHandlers: AjaxExceptionHandler[] = [];
        if (errorName) {
            // try to invoke specific AjaxExceptionHandler
            exceptionHandlers = core.getWidgetsByType(AjaxExceptionHandler);
            for (var exceptionHandler of exceptionHandlers) {
                if (exceptionHandler.handles(errorName)) {
                    exceptionHandler.handle(errorName, errorMessage);
                    return;
                }
            }
        }
```

PR #14680, opened 2026-02-10
Fix primefaces#14669: Touchable tag overrides onclick/mousedown event

### LINKED ISSUE #14669: Paginator: Touchable tag overrides onclick/mousedown event of selectCheckboxMenu
### Describe the bug

The touchable tag messes with the onclick/mousedown event of SelectCheckboxMenus when used in custom filters in Datatables. 
The problem only occurs in chromium based browsers (tested in google-chrome:144.0.7559.132 and chromium:144.0.7559.96) and was not reproducible in Firefox. 
The bug might even depend on specific versions of chrome because the bug was not reliably reproducible on different machines, but I am able to reproduce it on the current latest. 

The bug is pretty niche since it only occurs on chromium browsers and occurs with custom filters utilizing the SelectCheckboxMenu element and can be avoided by disabling the touchable tag.
I unfortunately wasn't able to find the exact cause of this behavior but I might check again at a later time.

### Reproducer

[reproducer.zip](https://github.com/user-attachments/files/25184765/reproducer.zip)

### Expected behavior

SelectCheckboxMenu is clickable in a custom datatable filter

### PrimeFaces edition

None

### PrimeFaces version

16.0.0-SNAPSHOT

### Theme

_No response_

### JSF implementation

All

### JSF version

4.0.3

### Java version

11

### Browser(s)

google-chrome:144.0.7559.132, chromium:144.0.7559.96

joshhuyn:
Fix #14669

### DIFF
--- a/primefaces/src/main/frontend/packages/components/src/forms/forms.selectcheckboxmenu.widget.js
+++ b/primefaces/src/main/frontend/packages/components/src/forms/forms.selectcheckboxmenu.widget.js
@@ -375,7 +375,7 @@ PrimeFaces.widget.SelectCheckboxMenu = class SelectCheckboxMenu extends PrimeFac
                 $this.jq.removeClass('ui-state-hover');
                 $this.triggers.removeClass('ui-state-hover');
             }
-        }).on('mousedown.selectCheckboxMenu', function(e) {
+        }).on('click.selectCheckboxMenu', function(e) {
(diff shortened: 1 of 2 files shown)


melloware (approved):
These changes look good to me.  You are right it was weird it was doing both `mousedown` and `click`

PR #14687, opened 2026-02-13
CDK: Required fields

melloware:


primefaces/src/main/java/org/primefaces/component/api/UITree.java
```diff
@@ -487,9 +426,9 @@ public void processDecodes(FacesContext context) {
         pushComponentToEL(context, this);
 
         preDecode(context);
-        Map<String, SavedState> saved = (Map<String, SavedState>) getStateHelper().get(PropertyKeys.saved);
+        Map<String, SavedState> saved = (Map<String, SavedState>) getStateHelper().get("saved");
```
melloware: @tandraschko this is the part i am not sure about the new generated code is storing an ENUM value but this is being used inside itself?

melloware: OK so this definitely broke the TreeTable integration tests which is good.

primefaces/src/main/java/org/primefaces/component/tree/TreeBase.java
```diff
@@ -49,7 +49,7 @@
     @FacesBehaviorEvent(name = "contextMenu", event = NodeSelectEvent.class, description = "Fires when context menu is invoked on a node."),
     @FacesBehaviorEvent(name = "filter", event = AjaxBehaviorEvent.class, description = "Fires when data is filtered.")
 })
-public abstract class TreeBase extends UITree implements Widget, RTLAware, StyleAware {
+public abstract class TreeBase extends UITreeImpl implements Widget, RTLAware, StyleAware {
```
tandraschko: does it work when you just use UITree?

i think we should not use Impl classes for real base API classes

Issue #13833, opened 2025-05-22
Themes : deprecation of @import as of Dart Sass 1.8


philpil:
### Describe the bug

https://sass-lang.com/documentation/at-rules/import/
"Heads up!
As of Dart Sass 1.80.0, the @import rule is [deprecated](https://sass-lang.com/documentation/breaking-changes/import) and will be removed from the language in Dart Sass 3.0.0. Prefer the [@use rule](https://sass-lang.com/documentation/at-rules/use) instead."

current theme-base components.scss uses @import 
https://github.com/primefaces/primefaces/blob/master/primefaces-themes/theme-base/_components.scss

### Expected behavior

sass to be updated, and maybe your builds need to update their dart-sass version to see latest warning 

### PrimeFaces edition

None

### PrimeFaces version

15.0

### Theme

All I guess

### JSF implementation

All

### JSF version

2.2

### Java version

21

melloware:
@philpil  yes we are aware. This is a big refactor from how PrimeTek has structured it so in all my projects and templates I locked it to 1.79.6.

```xml
 <plugin>
                <groupId>io.github.cleydyr</groupId>
                <artifactId>dart-sass-maven-plugin</artifactId>
                <version>1.5.0</version>
                <executions>
                    <execution>
                        <id>generate-css-using-sass</id>
                        <phase>generate-resources</phase>
                        <goals>
                            <goal>compile-sass</goal>
                        </goals>
                    </execution>
                </executions>
                <configuration>
                    <inputFolder>${project.basedir}/src/main/webapp/resources/</inputFolder>
                    <outputFolder>${project.basedir}/src/main/webapp/resources/</outputFolder>
                    <update>true</update>
                    <noSourceMap>true</noSourceMap>
                    <style>COMPRESSED</style>
                    <quiet>false</quiet>
                    <version>1.79.6</version>
                </configuration>
</plugin>
```

melloware:
This will eventually have to be tackled as we can't stay on Dart SASS 1.79 forever

PR #14691, opened 2026-02-17
Switching to a central definition of dart-sass version

AndreasIgelCC:
Currently, the version of dart‑sass is defined only in the primefaces-themes module, while primefaces-showcase uses no explicit version. This makes builds not necessarily reproducible across environments. In addition, some environments require offline availability, which is only possible when the dart‑sass version is pinned consistently.

This pull request introduces a centralized definition of the dart‑sass version and ensures that both primefaces-themes and primefaces-showcase use the same version. As part of this improvement, the dependency is set to the correct dart‑sass version (1.79.6, https://github.com/sass/dart-sass/releases) according to #13833.

I hope this change helps make the build process more stable, predictable, and contributor‑friendly. Looking forward to your feedback!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -76,7 +76,7 @@
         <jacoco.version>0.8.14</jacoco.version>
         <tomcat.version>10.1.48</tomcat.version>
         <slf4j.version>2.0.17</slf4j.version>
-        <dart-sass.version>1.79.6</dart-sass.version>
+        <dart-sass.version>1.97.3</dart-sass.version>
 
         <sonar.skip>true</sonar.skip>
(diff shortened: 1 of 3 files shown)


melloware:
I thought I was on 1.79.1 as it's the last one that does not throw warnings and errors

There is an open ticket about it.

melloware:
Yep #13833

AndreasIgelCC:
> I thought I was on 1.79.1 as it's the last one that does not throw warnings and errors
> 
> There is an open ticket about it.

Thank you for your reply, but then it is even more important that this is defined on a central place. I switched the version to 1.79.6, which is named in #13833

melloware (approved):

melloware:
Thanks this is a good change

PR #14694, opened 2026-02-17
CDK: Taglib generation

melloware:
- [x] Verified both legacy primefaces namespace and new URN namespace are both generated correctly.
- [x] Deleted /src/main taglib.xml
- [x] Verified VDLDOC still works after moving this file to `/target/classes` so pointed VDLDOC plugin to it
- [x] Sorted tags and properties alphabetically in the taglib.xml to make things easier to see


primefaces-cdk/primefaces-cdk-impl/src/main/java/org/primefaces/cdk/impl/taglib/TaglibMojo.java
```diff
@@ -307,8 +312,7 @@ private TagHandlerInfo processTagHandlerClass(Class<?> tagHandlerClass) {
     private Document generateTaglibXml(List<FunctionInfo> functionInfos, List<ComponentInfo> componentInfos, List<BehaviorInfo> behaviorInfos,
                                        List<ValidatorInfo> validatorInfos, List<TagHandlerInfo> tagHandlerInfos) {
         Document document = DocumentHelper.createDocument();
-        Element faceletTaglib = document.addElement("facelet-taglib")
-                .addAttribute("xmlns", "https://jakarta.ee/xml/ns/jakartaee")
+        Element faceletTaglib = document.addElement("facelet-taglib", "https://jakarta.ee/xml/ns/jakartaee")
```
melloware, line 315: Fixed this namespace because VDLDOC complained it was missing

primefaces/pom.xml
```diff
@@ -321,8 +328,8 @@
                         </goals>
                         <phase>process-classes</phase>
                         <configuration>
-                            <uri>primefaces-test</uri>
-                            <shortName>pf-test</shortName>
+                            <uri>http://primefaces.org/ui</uri>
```
melloware, line 331: I am generating the legacy namespace as later step we create the URN version of the file.  Someday we need to remove this legacy version maybe in 17 or 18?

melloware:
OK IT's failing need to find out why

PR #14717, opened 2026-02-28
Fix #14706: Panel set cursor:pointer on toggleHeader

### LINKED ISSUE #14706: Panel: ToggleableHeader property should indicate interactivity via cursor
### Describe the bug

While the property `toggleable=true` renders a control element (+) (-) to toggle the panel...

Image

... the `toggleableHeader=true` property simply implements a click listener on the header.

Image

Image

Here the user has no haptic feedback, that they can click on the panel header.

### Reproducer

```
<p:panel header="Toggleable Panel" toggleable="true" toggleableHeader="true">
    <div>Lore</div>
    <div>Ipsum</div>
</p:panel>
```

### Expected behavior

A `pointer` cursor would be appropriate.

### PrimeFaces edition

Community

### PrimeFaces version

15.0.13

### Theme

any

### JSF implementation

Mojarra

### JSF version

4.0.8

### Java version

24

### Browser(s)

any

melloware:
Fix #14706: Panel set cursor:pointer on toggleHeader

### DIFF
--- a/primefaces-showcase/src/main/webapp/ui/panel/panel.xhtml
+++ b/primefaces-showcase/src/main/webapp/ui/panel/panel.xhtml
@@ -34,7 +34,7 @@
             </p:panel>
 
             <h5>Toggleable</h5>
-            <p:panel header="Header" toggleable="true" closable="true" toggleSpeed="500"
+            <p:panel header="Header" toggleable="true" toggleableHeader="true" closable="true" toggleSpeed="500"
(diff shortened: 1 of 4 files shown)


tandraschko:
I wonder If we should better add a styleable css class

melloware:
You are right.  I will update it.

melloware:
OK this is a much better fix and I cleaned Panel up to use `getStyleClassBuilder()`

PR #14732, opened 2026-03-04
Fix #14719: SelectOneMenu add panelAutoSize prop to always align panel with input

### LINKED ISSUE #6877: SelectOneMenu: Align width of overlay panel with input
**Description**
Overlay panel of p:selectOneMenu is changing size (width) according to size of values (width could be even wider than the screen size). There is an option 'autoWidth', but it works in opposite way - it extends width of input by size of values.

It would be nice, if there will be an options to fix overlay width with input width. 
Solution may be like this somehow (if new attribute will be named "alignPanelWidthWithInput"):
```
alignPanelWidth: function() {
        //align panel and container
        if(!this.panelWidthAdjusted) {
            var jqWidth = this.jq.outerWidth();
            if(this.cfg.alignPanelWidthWithInput || this.panel.outerWidth() < jqWidth) {
                this.panel.width(jqWidth);
            }
            else {
                this.panel.width(this.panel.width());
            }

            this.panelWidthAdjusted = true;
        }
    }
```
 + add to css .ui-selectonemenu-list-item:
```
text-overflow: ellipsis;
overflow: hidden;
```

Thanks

### LINKED ISSUE #14719: SelectOneMenu: Panel not aligned with size of select element
### Describe the bug

The bug is also visible on the primefaces showcase page: 

https://showcase.primefaces.org/ui/input/oneMenu.xhtml

If you click in a selectOneMenu input field to render a dropdown and then change the width of the browser window. Either shrink or widen it. Then the dropdown is not aligned with the size of the select element in the smaller or bigger browser window. Even if you close and open the dropdown again. The dropdown still has the same width when it was first clicked.

I would assume that the dropdown is aligned with the width of the select element also in the smaller or bigger window.

 

### Reproducer

1. Go to https://showcase.primefaces.org/ui/input/oneMenu.xhtml
2. Click on a select element.
3. Shrink or Widen the browser window.
4. See that the width of the dropdown doesnt match the width of the select element. Even as you close and open the dropdown again.

### Expected behavior

I would like to have the dropdown always to be aligned with the size of the select elment.
If this is not possible maybe correct the size of the dropdown to align the size of the select element if you open the dropdown again on the smaller or bigger browser window.

### PrimeFaces edition

None

### PrimeFaces version

15.0.13

### Theme

myTheme

### JSF implementation

Mojarra

### JSF version

4.0.9

### Java version

Temurin-21.0.10+7

### Browser(s)

Chrome

melloware:
Fix #14719: SelectOneMenu add panelAlignWidth prop to always align panel with input
Fix #6877

### DIFF
--- a/docs/16_0_0/gettingstarted/whatsnew.md
+++ b/docs/16_0_0/gettingstarted/whatsnew.md
@@ -26,6 +26,8 @@ Look into [migration guide](https://primefaces.github.io/primefaces/16_0_0/#/../
     * Added new `dragDrop` attribute to be able to disable drag and drop capabilities.
 * SelectOneButton
     * Added new `layout="custom"` to allow complete control over rendering
+* SelectOneMenu
+    * Added new `panelAutoSize="true"` option to always align the overlay panel's width with the select menu.
 * SelectManyButton
     * Added new `layout="custom"` to allow complete control over rendering
 * SplitButton
--- a/primefaces-showcase/src/main/webapp/ui/input/oneMenu.xhtml
+++ b/primefaces-showcase/src/main/webapp/ui/input/oneMenu.xhtml
@@ -64,7 +64,7 @@
                     <div class="field col-12 md:col-4">
                         <p:outputLabel for="@next" value="Advanced"/>
                         <p:selectOneMenu id="advanced" value="#{selectOneMenuView.country}" converter="#{countryConverter}"
-                                        panelStyle="width:180px" var="c"
+                                        panelAutoSize="true" var="c"
                                         filter="true" filterMatchMode="startsWith" filterNormalize="true">
 
                             <f:selectItems value="#{selectOneMenuView.countries}" var="country"
--- a/primefaces/src/main/frontend/packages/components/src/forms/forms.selectonemenu.widget.js
+++ b/primefaces/src/main/frontend/packages/components/src/forms/forms.selectonemenu.widget.js
@@ -85,6 +85,7 @@
  * which is replaced with the value of the currently selected item.
  * @prop {boolean} cfg.syncTooltip Updates the title of the component with the description of the selected item.
  * @prop {boolean} cfg.renderPanelContentOnClient Renders panel content on client.
+ * @prop {boolean} cfg.panelAutoSize Automatically adjusts the width of the overlay panel to match the width of the select menu.
  */
 PrimeFaces.widget.SelectOneMenu = class SelectOneMenu extends PrimeFaces.widget.DeferredWidget {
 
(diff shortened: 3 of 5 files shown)


tandraschko:
another name idea:
panelAutoSize=false

melloware:
Not a bad idea i thought of that too because we have a regular `autoSize` prop as well for the input this would make them align.  I will update

melloware:
Renamed `panelAutoSize`

tandraschko:
a bit unsure
panelAutoSize=true is the current default, where the panel get resized based on the content
and the user want
panelAutoSize=false?

melloware:
The current code only resizes it on initial page load

If you shrink or increase your browser it's the same size it originally was. Both of these tickets want it where the input and panel are always the same size.  

It's been there for 10+ years there is a panelWidthasjuated variable the ensures it's only called on page load

tandraschko:
hmmm
maybe we should make it the default for now and remove later, if we dont get any negative feedback?
no idea why it was done like this?

melloware:
yeah i was also considering call it `panelWidthSync` basically saying keep it in sync.  I don't think we want it to be the default because of long label support see this screenshot

image


that we specifically support when people WANT it to be longer than the input to match the panel items

tandraschko:
will check soon

tandraschko:
hmm
im not happy with the naming + description

current behavior:
the initial click calculates the width of the panel (so its actually already autosize)

the new behavior:
every click re-calculates the width of the panel

right?

melloware:
if you enable this property yes it recalculates every  time is it displayed.   Any better name for it?  `panelSyncWithInput`?

tandraschko:
but the logic is the same right?
initial width calculation = same logic as you would now call be every panel open click?

melloware:
correct before it was only calculating the width on widget creation with this flag `this.panelWidthAdjusted`  I still haven't figured out why?

This new flag just gives you the option to have it re-caculated every time its opened not juts widget creation time.

tandraschko:
IMO we can just remove the boolean and always recalculate

look
if you click the panel, it loads content (if lazy), then calls jq.outerWidth() on it and set panelWidth

it only happens if you active click on it
i did some tests on my machine and alignPanelWidth took always like 0-2ms. So even it would take 30ms, a user wont notice it. it only hapens when he click it and the panel is opening.

tandraschko:
feel free to measure that performance of the method in your app with many content

melloware:
Itotally agree.  I am not noticing any performance issue even with larger lists.  Let me make that change

tandraschko:
another case i could image:
if you click multiple times, different content is loaded, different panel size

mk-htw:
@melloware Will this feature be released in Primefaces 16? I couldn´t find the commit in the master branch. It is useful for our application at work. The problem is the calculated width of the panel at creation time. If you zoom in after the first click, the panel width won´t resize like 1Kamikazzze1 described it in this [comment](https://github.com/primefaces/primefaces/issues/14719#issuecomment-3989098362)

melloware:
@mk-htw yes this is fixed in 16.0.0 MAIN but the fix was much simpler than a new property the final fix is right here: https://github.com/primefaces/primefaces/pull/14792/changes

mk-htw:
Thanks for the quick answer @melloware 
I used your fix. I’m seeing the same behavior in my case: the panel width remains unchanged. I tried opening the dropdown at 100% zoom, and the panel is aligned with the input of the selectonemenu. If you change the zoom to 400%(for example) the panel width remains the same while the input gets narrower. As a result, the panel is not aligned with the input anymore. I assume the panel width will not be recalculated because the else-branch(l.289ff of your fix) will be processed.

melloware:
@mk-htw  the panel only resizes on DROPDOWN of the the panel not on browser zoom resizing.

Issue #14763, opened 2026-03-13
Tree Filter: Shows all nodes on no-match input


bajce:
### Describe the bug

On showcase page https://showcase.primefaces.org/ui/data/tree/filter.xhtml, the tree filter fails for non-matching text. 
When you type text that exists (like "Events"), it filters right.
But for text with no match (like "xyz"), all tree nodes are visible. This happens in both Basic and Custom tree examples.

<img width="383" height="620" alt="Image" src="https://github.com/user-attachments/assets/ce39af22-e8cb-4975-b765-77b890b8969e" />

### Reproducer

1. Go to https://showcase.primefaces.org/ui/data/tree/filter.xhtml
2. Type "Events" in filter box, as result tree filters works.
3. Clear and type "xyz" (no match), as result all nodes are still visible.

### Expected behavior

Tree should show no nodes if filter does not match.

### PrimeFaces edition

None

### PrimeFaces version

15.0.14-snapshot

### JSF implementation

All

### JSF version

4..0

### Java version

17

FlipWarthog:
For future reference, this is the line that was causing trouble but we can't change it directly or it breaks non-filtering situations: https://github.com/primefaces/primefaces/blob/7d043ecaa69b6322494bc8f7fc72520fbc76ac13/primefaces/src/main/java/org/primefaces/component/tree/TreeRenderer.java#L690

So adding a 'fake' filter that'll never match is the cleanest solution.

PR #14766, opened 2026-03-13
Fix: #14763 15.0.14 Tree Filter: Shows all nodes on no-match input

melloware:
Fix: #14763 15.0.14 Tree Filter: Shows all nodes on no-match input


primefaces/src/main/java/org/primefaces/component/tree/TreeRenderer.java
```diff
@@ -263,6 +263,10 @@ else if (tree.isFilterRequest(context)) {
             tree.getFilteredRowKeys().clear();
             if (LangUtils.isNotBlank(filteredValue)) {
                 encodeFilteredNodes(context, tree, tree.getValue(), filteredValue, filterLocale);
+                if (component.getFilteredRowKeys().isEmpty()) {
```
FlipWarthog, line 266: var name needs updating here too for PF15
```suggestion
                if (tree.getFilteredRowKeys().isEmpty()) {
```

PR #14781, opened 2026-03-18
MoveScriptsToBottom optimizations: cpu & memory

pizzi80:
This version avoid:
-  multiple `String.replace` call on the whole merged skip (memory 6x copy)
-  multiple `String.equals` call
-  mulitple `String` concatenations
- random `';'` at the end of the line due to space inside inline scripts

This version has:
 - final on effective final fields
 - early skip if there are no inline scripts
 - inline trim + check the needs of adding ';'
 - `Constants.EMPTY_STRING` instead of ""

Notes:
A page can contains a lot of inline scripts, that can also be large...
every single call to `String.replace` create a copy of the String,
also the final string prepend + append create a copy of the merged script,
also the final `String.format` internaly use the old StringBuffer plus create a copy of the content...

### DIFF
--- a/primefaces/src/main/java/org/primefaces/application/resource/MoveScriptsToBottomResponseWriter.java
+++ b/primefaces/src/main/java/org/primefaces/application/resource/MoveScriptsToBottomResponseWriter.java
@@ -25,6 +25,7 @@ package org.primefaces.application.resource;
 
 import org.primefaces.renderkit.RendererUtils;
 import org.primefaces.util.AgentUtils;
+import org.primefaces.util.Constants;
 import org.primefaces.util.LangUtils;
 
 import java.io.IOException;
@@ -34,6 +35,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.UUID;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import javax.faces.FacesException;
 import javax.faces.component.UIComponent;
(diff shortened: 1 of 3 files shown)


tandraschko:
please provide a test (to not break something) and which also prooves some numbers before vs after

primefaces/src/main/java/org/primefaces/application/resource/MoveScriptsToBottomResponseWriter.java
```diff
@@ -217,8 +234,7 @@ else if (BODY_TAG.equalsIgnoreCase(name) || (HTML_TAG.equalsIgnoreCase(name) &&
             for (Entry<String, List<Map<String, String>>> entry : state.getIncludes().entrySet()) {
 
                 List<Map<String, String>> includes = entry.getValue();
-                for (int i = 0; i < includes.size(); i++) {
-                    Map<String, String> attributes = includes.get(i);
+                for (Map<String, String> attributes : includes) {
```
melloware: Revert this as the project specifically uses for loops not iterators. Its a project standard.

melloware (changes requested):

primefaces/src/main/java/org/primefaces/application/resource/MoveScriptsToBottomResponseWriter.java
pizzi80: fixed

pizzi80:
> please provide a test (to not break something) and which also prooves some numbers before vs after

I saw that there is a test for `MoveScriptsToBottomResponseWriter`

https://github.com/primefaces/primefaces/blob/15.X/primefaces/src/test/java/org/primefaces/application/resource/MoveScriptsToBottomResponseWriterTest.java

I've also tested with my dev environment ... going in production later or tomorrow

tandraschko:
i see!
then please provide some numbers before and after

pizzi80:
> i see! then please provide some numbers before and after

Thanks for pushing me to create a test; 
I discovered that the Pattern/Matcher state machine 
is actually slower than expected. 

It negates all the benefits of avoiding multiple String.replace calls.

working on a mix of the proposed solution and the original one ...

pizzi80:
I've written a benchmark that simulate `mergeAndMinimizeInlineScripts` 
on a `List` of script taken from the home page of a real webapp


```
Benchmark                        Mode  Cnt   Score   Error  Units
InlineScriptsBenchmark.benchNew  avgt   20   5,144 ± 0,057  us/op
InlineScriptsBenchmark.benchOld  avgt   20  15,068 ± 0,101  us/op
```

melloware:
Nice!

tandraschko:
looks good for me, thanks
can you also create a PR for master?

melloware:
I can pull it forward if you want @pizzi80

melloware (approved):

tandraschko:
+1 @melloware

PR #14783, opened 2026-03-18
LangUtils: optimizations + minor warnings fix

pizzi80:
just a couple of optimizations and warning fix:

the `newLinkedHashSet` method now properly 
init the LinkedHashSet to avoid remapping 

the method `calculateHashMapCapacity`
is ported from Java 19 and optimized

### DIFF
--- a/primefaces/src/main/java/org/primefaces/util/LangUtils.java
+++ b/primefaces/src/main/java/org/primefaces/util/LangUtils.java
@@ -304,12 +304,6 @@ public class LangUtils {
 
     @SafeVarargs
     public static <E> Set<E> newLinkedHashSet(E... elements) {
-        if (elements.length == 0) {
-            return Collections.emptySet();
-        }
-        else if (elements.length == 1) {
-            return Collections.singleton(elements[0]);
-        }
         Set<E> set = new LinkedHashSet<>(calculateHashMapCapacity(elements.length));
         Collections.addAll(set, elements);
         return set;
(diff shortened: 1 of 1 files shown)


tandraschko:
-1 for the newLinkedHashSet changes
currently its modifieable

pizzi80:
> -1 for the newLinkedHashSet changes currently its modifieable

:top: ... all that unmodifiable methods ... and I got fooled


fixed

tandraschko:
and we should only apply this to master IMO
15.x is actually in bugfix mode, this changes are micro micro optimzations

pizzi80:
I know but ... micro by micro ... inside the core Utils methods ... called a lot of times ...

tandraschko:
not really actually ;D
unmodifiableList is just for static code
and the difference of calculateHashMapCapacity is not measurable in a application

if you switch from mojarra to myfaces or follow our performance guide, there are VERY big differences compared to such a change.

pizzi80:
can't understand why integration test keeps failing... 

on my pc mvn install has no errors

the 15.0.14-snapshot work as expected

melloware:
If only one fails its usually just flaky. I can re-run it right now and I bet it works.  Let me try.

melloware:
See now it passed!

melloware (approved):

pizzi80:
oh ok ... thanks ;)

tandraschko:
+1
please port to master

PR #14792, opened 2026-03-23
Fix #14719: SelectOne always recalculate panel width

### LINKED ISSUE #14719: SelectOneMenu: Panel not aligned with size of select element

melloware:
Fix #14719: SelectOne always recalculate panel width

@tandraschko simpler to open a new PR

### DIFF
--- a/primefaces/src/main/frontend/packages/components/src/forms/forms.selectonemenu.widget.js
+++ b/primefaces/src/main/frontend/packages/components/src/forms/forms.selectonemenu.widget.js
@@ -45,7 +45,6 @@
  * @prop {number} optGroupsSize The number of option groups.
  * @prop {JQuery} panel The DOM element for the overlay panel with the available selectable options.
  * @prop {JQuery} panelId ID of the DOM element for the overlay panel with the available selectable options.
- * @prop {number} panelWidthAdjusted The adjusted width of the overlay panel.
  * @prop {JQuery} preShowValue The DOM element for the selected option that is shown before the overlay panel is brought.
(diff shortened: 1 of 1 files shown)


tandraschko:
+1

lets only do it in master

melloware:
Agreed I marked this for 16.0.0 only

PR #14794, opened 2026-03-23
NativeFileUploadDecoder.getUploadDirectory optimization

pizzi80:
This version avoid the usage of Collections.list(...) 
which internally iterate over the passed Enumeration 
and create a copy returning an ArrayList.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/fileupload/NativeFileUploadDecoder.java
+++ b/primefaces/src/main/java/org/primefaces/component/fileupload/NativeFileUploadDecoder.java
@@ -29,7 +29,6 @@ import org.primefaces.util.FileUploadUtils;
 import org.primefaces.util.LangUtils;
 
 import java.io.IOException;
-import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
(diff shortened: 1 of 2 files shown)


tandraschko:
the PR itself looks fine but please target only master for such PRs
this is more like a refactorting, then a real performance improvement.

15.x is actually in maintanence mode, where we should only port bugfixes - we have just become a bit too careless to merge everything.

pizzi80:
ok, but questionable...  

I can create the same fix without introducing 
a new method in `LangUtils` if this is the problem 

in theory that method should use the Enumeration<T> 
offered by the servlet api (no comment btw...)

at the end that method is only traversing the request attributes 
to find the first elem that is a MultipartConfigElement 

using a Stream in this situations in not recommended
because at runtime it put the GC under pressure for no reason 

it also put the jit c2 compiler under pressure 
to rewrite an optimized version

tandraschko:
its not the code, its the type of change

we are working towards PrimeFaces 16x
if PF 16 is released, 15.x public releases will end

i dont want to change code that works, to gain 1/10000 performance
that are optimisations that should go in master

pizzi80:
ok for versioning 

generally speaking do not understimate the performance gain

PrimeFaces is a webapp framework not a desktop ui framework

internal core methods have a tremendous multiplier

also is only a single piece of a big stack + dependencies...

tandraschko:
dont worry, i know :)
if you look in the myfaces history, we usually optimize each micro detail

we even look at object instances creates, not just plain performance
therefore also the index-loop over arraylists, just to safe as much as possible GC

PR #14795, opened 2026-03-24
15.x NativeFileUploadDecoder optimization

pizzi80:
this version do not use the Java Stream api at all
and do not add a method to LangUtils


melloware (approved):

tandraschko:
as discussed, better just to such refactorings to master only

PR #14686, opened 2026-02-13
Attempt to fix #14685

FlipWarthog:
Attempt to fix #14685

@melloware testing a theory pushing this after ~6pm EST.

PR #14800, opened 2026-03-28
bugfix: ajax error event is handled instead of being ignored

### LINKED ISSUE #14782: Ajax-disabling CommandButton - inconsistent state when Ajax errors occur
### Describe the bug

When either client-side connection errors occur, or non-200 xml response is received from Ajax requests,
"Ajax Disabling and Status Icon" buttons will remain in a "disabled" / "in-transit" inconsistent state forever, and the only
way to get out of that state is to reload the page.

There is a workaround by using `onerror=` field with the button API, but this requires adding error elements to all Command buttons.
These types of errors should be handled by all Command buttons globally, correctly, by default.

Follow-up to #14773
Hint: This issue may not be constrained to `p:commandButton` and may be effecting other components.

### Reproducer

This is easily reproducible via she Showcase by pressing "Ajax Disabling and Status Icon" buttons in airplane mode with WiFi disabled

### Expected behavior

When there is a client error, buttons do not stay in an inconsistent state

### PrimeFaces edition

Community

### PrimeFaces version

Latest

### Theme

Any

### JSF implementation

Mojarra

### JSF version

Latest

### Java version

Latest

### Browser(s)

Any

lprimak:
Respond properly to `pfAjaxError`
fixes #14782

### DIFF
--- a/primefaces/src/main/frontend/packages/core/src/core/core.ts
+++ b/primefaces/src/main/frontend/packages/core/src/core/core.ts
@@ -1125,7 +1125,7 @@ export class Core {
                 }
                 button.prepend(loadIcon);
             }
-        }).on('pfAjaxComplete' + namespace, function(e, xhr, settings, args) {
+        }).on('pfAjaxComplete pfAjaxError' + namespace, function(e, xhr, settings, args) {
             if (isXhrSource.call(this, widget, settings)) {
                 widget.ajaxCount--;
                 if (widget.ajaxCount > 0 || !args || args.redirect) {
--- a/primefaces/src/main/frontend/packages/core/src/core/core.ts
+++ b/primefaces/src/main/frontend/packages/core/src/core/core.ts
@@ -1125,7 +1125,7 @@ export class Core {
                 }
                 button.prepend(loadIcon);
             }
-        }).on('pfAjaxComplete pfAjaxError' + namespace, function(e, xhr, settings, args) {
+        }).on('pfAjaxComplete' + namespace + ' pfAjaxError' + namespace, function(e, xhr, settings, args) {
             if (isXhrSource.call(this, widget, settings)) {
                 widget.ajaxCount--;
                 if (widget.ajaxCount > 0 || !args || args.redirect) {


lprimak:
I am not sure how to build PF, since it requires npm, and I don't want that on my machine :)
So I can't really test if it does what it's supposed to do.

Only one test failure, not sure how to track that down b/c I am unfamiliar with integration test methodolgy
```
Error:  Failures: 
Error:    Csv003Test.datePastInvalid:61->assertMessage:201 expected: <Time (@Past): must be a past date.> but was: <>
[INFO] 
```

Can you help me with the test? And can I get a PF JAR that I can test with?

melloware:
@FlipWarthog this looks like another date related bug running near UTC midnight?

melloware:
@lprimak I don't think it's related to your code I think it's the time of day this test ran.

FlipWarthog:
> @FlipWarthog this looks like another date related bug running near UTC midnight?

Yep so it looks like this wasn't all of it? 

https://github.com/primefaces/primefaces/pull/14686

melloware:
Confirmed re-ran them today and they are passing.  So definitely another timing bug

melloware (approved):

primefaces/src/main/frontend/packages/core/src/core/core.ts
```diff
@@ -1125,7 +1125,7 @@ export class Core {
                 }
                 button.prepend(loadIcon);
             }
-        }).on('pfAjaxComplete' + namespace, function(e, xhr, settings, args) {
+        }).on('pfAjaxComplete pfAjaxError' + namespace, function(e, xhr, settings, args) {
```
melloware:
```suggestion
        }).on('pfAjaxComplete' + namespace + ' pfAjaxError' + namespace, function(e, xhr, settings, args) {
```

melloware (changes requested):

tandraschko:
Can you please explain the fix?
@melloware did you retest?

lprimak:
To be clear: I did not test yet and don't know if it actually works. I do have an idea how to test it and I will get back to you with a result. 

Basically it uses error callback globally to do the same thing that success does currently, which is to reenable the button.

melloware:
@tandraschko he is right in theory the error only happens in jquery in network type failures so it's hard to test. 

However I thought I remember OnComplete is supposed to always be called even in success or error conditions?  So maybe the bug is higher up than this?

lprimak:
I had a chance to test this now. It works, but only partially.
I tested the 404/NotFound case, and that works correctly.

However, network failures do not work, and I am not sure why, and probably need some help / advice.
It looks like `pfAjaxComplete` is being trigged, but without a namespace from a network error.
However, it doesn't wind up in the [handler](https://github.com/primefaces/primefaces/blob/157cb0dde2d6cbd03943ab62dbb6bd51789bee1d/primefaces/src/main/frontend/packages/core/src/core/core.ts#L1129). I don't know why.

My theory is that its "namespace" vs no-namespace (global) callbacks. I do not understand how that stuff works, so I will need some help deciphering this. Global callbacks get called (with XHR parameter) when network error occurs, but when 404 occurs, "local" with-namespace callbacks get called.

lprimak:
I've converted this PR to draft since it's not fully working yet

tandraschko:
actually it should be quite easy to test:
start showcase
open view
shutdown showcase
click button

melloware:
Also try removing the namespaces.  They are not necessary here

lprimak:
> Also try removing the namespaces. They are not necessary here

I just tried that. Unfortunately no effect, the callback is not getting called.

@tandraschko Yes, this is pretty much what I am doing. The issue is that I don't want to have
the node ecosystem on my machine, so I can't locally build PrimeFaces :(
However, I just changed the resulting core.js in-place and re-jar the PF JAR, that's how I am testing it.

Unfortunately I am stuck as of now.

tandraschko:
the NPM stuff lives inside a subdirectory of PrimeFaces then, not sure if its a big issue.

lprimak:
@tandraschko have you heard of the axios hack?  It's just the latest issue with npm ecosystem. I am not a node/user and don't want to take on extra risk of this stuff since I don't know how to mitigate these attacks.

tandraschko:
yeah

nevertheless, i dont have time to work on this

lprimak:
We can let it "live on the downlow" for a bit maybe an idea will come up at some point

melloware:
I just tested this and its working for me.

I used the validation example: https://showcase.primefaces.org/ui/ajax/validation.xhtml

1. start showcase
2. open view
3. shutdown showcase
4. click button

After about 5 seconds the AJAX request times out and in the showcase it hits the following block:

```js
$(document).on('pfAjaxError', function(e, xhr, settings, error){
        // this is very likely a connection error
        ajax.Utils.handleError("", "AJAX failure");
    });
```

and sends me to `/error.xhtml`!  If I comment that line out `ajax.Utils.handleError("", "AJAX failure");` and do the same test I now see its hits my button handler and re-enables the button

```js
}).on('pfAjaxComplete' + namespace + ' pfAjaxError' + namespace, function(e, xhr, settings, args) {
            if (isXhrSource.call(this, widget, settings)) {
                widget.ajaxCount--;
                if (widget.ajaxCount > 0 || !args || args.redirect) {
                    return;
                }

                $this.queueTask(
                    () => $this.buttonEndAjaxDisabled(widget, button),
                    Math.max(ajax.minLoadAnimation + (widget.ajaxStart ?? 0) - Date.now(), 0)
                );
                widget.ajaxStart = null;
            }
        });
```

So @lprimak i can't reproduce your issue at all?

lprimak:
Can you attach the PF-Jakarta Jar you are using so I can test?

melloware:
@lprimak i am using the showcase with 16.0.0-SNAPSHOT i simply ran `mvn clean install` on PF Master.  Then on the showcase run `mvn -Ddev`

lprimak:
I am still on 15.0.13 so lemme see what I can find out.

melloware:
ok you should be able to just checkout the 15.X branch here and do the same thing. `mvn clean install` and run the showcase with `mvn -Ddev`

lprimak:
I already tested with the online showcase with buttons and that fails. This test was done a few weeks ago

lprimak:
Just re-tested with online showcase on the iPhone and it behaves better than before. I think this is due to @tandraschko latest changes. I need to dig into this more.

lprimak:
~Already fixed in 15.0.14~

tandraschko:
any news here?

melloware:
i think this PR can be closed as I don't think this adds anything that wasn't already being handled with `pfAjaxComplete`

lprimak:
I have plans for this PR still. They are just unfolding slowly. Please don't do anything and stay tuned.

tandraschko:
please just recreate the PR then :)
i would like to cleanup a bit and avoid such PRs with no current activity

PR #14821, opened 2026-04-08
Fix type of rowKey in DataTableRenderer

jbrower2:
`DataTable.getRowKey()` returns a `String`, so this variable can safely be a `String` rather than `Object`.


melloware (approved):
@tandraschko  this seems reasonable to me?  This is just a spot we missed along the way?

PR #14897, opened 2026-05-10
FileUpload: add new "header" facet

### LINKED ISSUE #14909: FileUpload: new facet(s) to display custom content next to the buttons
### Description

A new facet (or two, one for left and one for right) to display custom content next to the advanced FileUpload buttons

image

### Additional context

_No response_

NicolaIsotta:
I added a new "button" facet which lets the developer render customen content next to the buttons of a p:fileUpload.
Preview:
image
I am not sure about the facet name, open to suggestions...
CC @melloware @tandraschko 

fixes #14909

### DIFF
--- a/docs/16_0_0/components/fileupload.md
+++ b/docs/16_0_0/components/fileupload.md
@@ -252,6 +252,16 @@ The empty-facet is rendered when no files are added yet. It's rendered in `div` 
 </p:fileUpload>
 ```
 
+### Header Facet
+
+Use the `header` facet to render custom content in the header, next to the buttons.
+
+```xhtml
+<p:fileUpload advanced="true" ...>
+    <f:facet name="header">Only image files are allowed</f:facet>
+</p:fileUpload>
+```
+
 ## Auto Upload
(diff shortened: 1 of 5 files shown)


melloware:
@NicolaIsotta open a matching issue ticket too please.  And maybe you can do like we do for MenuBar and have `start` and `end` facets one that renders before the buttons and one that renders after the buttons?

melloware (changes requested):
see comments

tandraschko:
Need to check it

tandraschko:
we could propably call it "header" facet?

NicolaIsotta:
I thought about header too. Makes sense since the button bar it's actually the header of the component.
Should we go with one facet or double option (start/end)?

melloware:
I'll defer to @tandraschko i am OK with just `header`

tandraschko:
i would just go with header for now

people can style it like they want
maybe even the choose button should be right inestead of left buts thats another topic

NicolaIsotta:
name changed

melloware (approved):
see comments

PR #14903, opened 2026-05-12
Fix #14890: Abort AJAX on page teardown

### LINKED ISSUE #14890: Core JS: handleError interferes with periodic Poller
### Describe the bug

PrimeFaces 14.0.14 introduced a new call to PrimeFaces.ajax.Utils.handleError for AJAX errors:

```
core.ajax.js:
    $(document).on('pfAjaxError', function(e, xhr, settings, error){
        // this is very likely a connection error
        PrimeFaces.ajax.Utils.handleError("", "AJAX failure");
    });
```

It gets also called when a periodic poller is unable to complete its last cycle, because the user leaves the page.

```
template.xhtml:
<p:poll interval="3" timeout="60000" widgetVar="poll" global="false"
    partialSubmit="true" listener="#{viewBean.checkForUpdates}" onerror="window.location.reload()"/>
```

To reproduce the AJAX error it helps to delay the response from the view bean:

```
ViewBean.java
public void checkForUpdates() throws Exception {
    Thread.sleep(100000);
}
```

JavaScript output (without error-page in web.xml):
```
No ajaxExceptionHandler or error page found for '' defined! [core.js.xhtml:25:9]
AJAX failure [core.js.xhtml:25:9]
Request return with error:error. [core.js.xhtml:25:9]
```

First question: Is this an intended behaviour? It is obviously NOT a connection error, when the last poller cycle cannot be completed because the user simply changes the page / clicks a button.

Second question: How can I define an appropriate error handling? I have a catch-all error-page in web.xml. And I don't hava an AjaxExceptionHandler. So the user gets redirected to the catch-all error.xhtml and is confused, because in his opinion nothing bad has happened (which is true).

Third question: How do I define an AjaxExceptionHandler? The function handleError("", "AJAX failure") is called with an empty string as first argument. With an empty parameter errorName I cannot define a specific AjaxExceptionHandler. How do I define a global AjaxExceptionHandler instead?

### Reproducer

_No response_

### Expected behavior

When a poller cycle cannont be completed, the PrimeFaces.ajax.Utils.handleError should not be called. Because it didn't get called before PrimeFaces version 15.0.14 an such a design change is not suitable to be introduced in a minor version update.

### PrimeFaces edition

None

### PrimeFaces version

15.0.14

### Theme

_No response_

### JSF implementation

MyFaces

### JSF version

4.1.3

### Java version

21

### Browser(s)

_No response_

melloware:
Fix #14890: Abort AJAX on page teardown

### DIFF
--- a/primefaces/src/main/frontend/packages/core/src/core/core.ajax.ts
+++ b/primefaces/src/main/frontend/packages/core/src/core/core.ajax.ts
@@ -9,6 +9,12 @@ import { AjaxExceptionHandler } from "../ajaxexceptionhandler/ajaxexceptionhandl
 const unloadEvent = ("onpagehide" in window) ? "pagehide" : "unload";
 
 /**
+ * Set when the browser is tearing down or navigating away so in-flight XHR failures are not treated as application errors.
+ * Teardown does not consistently report jQuery {@code textStatus} of {@code "abort"} or a non-zero HTTP status.
+ */
+let pageUnloading = false;
+
+/**
  * Retrieves the text content of a node.
  * @param node Node from which to get its text.
  * @returns The text content of the node. 
@@ -931,6 +937,13 @@ export class AjaxRequest {
                     cfg.promise.reject({ jqXHR: xhr, textStatus: status, errorThrown: errorThrown });
                 }
 
+                // Navigation / tab close cancels requests; do not surface as real errors.
+                // Rely on pageUnloading, not jqXHR.status === 0 alone (0 is also used for CORS/network failures).
+                if (pageUnloading || status === "abort") {
+                    core.debug('Page unloaded or aborted.');
+                    return;
+                }
+
                 const location = xhr.getResponseHeader("Location");
                 if (xhr.status === 401 && location) {
                     core.debug('Unauthorized status received. Redirecting to ' + location);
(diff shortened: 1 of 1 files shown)


lprimak (approved):

tandraschko:
Need to check it

melloware:
Just a note i got here because i was debugging and noticed our "abort" calls were happing after this failure so i knew it wasn't our `abortAll` code causing this. 

### Cause:

When the browser navigates away because of:

```html
<a href="test.xhtml">Change page now!</a>
```

the browser cancels all in-flight XHR requests automatically.

That cancellation is *not* the same as calling:

```javascript
xhr.abort()
```

from your own JavaScript code.

Browsers handle page-unload cancellation differently:

* Firefox tends to surface it as a normal AJAX error/fail event
* Chrome often suppresses the AJAX callbacks entirely during navigation teardown

So during navigation:

* the document is unloading
* the JS event loop is shutting down
* pending callbacks may never execute
* logging can appear out of order or not at all

That explains why:

* Firefox gives you the `.fail()`
* Chrome appears to silently cancel

You can verify this by preventing navigation temporarily:

```html
<a href="test.xhtml" id="nav">Change page now!</a>
```

```javascript id="fz4lpn"
$('#nav').on('click', function (e) {
    e.preventDefault();

    xhr.abort();

    setTimeout(() => {
        window.location.href = this.href;
    }, 500);
});
```

You’ll probably see Chrome now consistently firing the abort-related handlers because the page is still alive long enough for the callbacks to execute.

Another useful diagnostic:

```javascript id="jpxjmo"
window.addEventListener('beforeunload', () => {
    console.log('Page unloading');
});
```

You’ll likely notice:

* unload happens before the AJAX callbacks finish in Chrome
* Firefox lets more callbacks complete during teardown

This behavior is fundamentally browser lifecycle timing, not jQuery itself.

PR #14911, opened 2026-05-16
Fix #14910: Datatable filter width CSS fix

### LINKED ISSUE #14910: DatePicker: styling broken / cut off when used as DataTable filter
See master / 16.0.0 Showcase:

Image

melloware:
Fix #14910: Datatable filter width CSS fix

image


melloware:
the `autoWidth="true"` on selectOne forces a `min-width` on the element that can't be overridden by CSS without `!important` so I updated the showcase example.  The rest now handle giving `width: 100%` to custom filters

PR #14914, opened 2026-05-18
BlockUI: don't hide on redirects

### LINKED ISSUE #14913: BlockUI: don't hide on redirect
### Description

Currently, BlockUI hides itself as soon as the ajax completes. If the ajax starts a redirect, this means the while the next page is loading, the current page/form is interactable.

I made some tests and I think don't hiding the blockUI on redirects is a UI/UX improvements.
I can make a GIF/video to show the differences if needed.

### Additional context

I have a fix/PR ready. I understand not everyone may like/want the new behavior, so we could maybe control it with an attribute?

NicolaIsotta:
fixes #14913 

I made this the new behavior as I see no downsides, but I can make it toggleable with an attribute if needed.


melloware (approved):

tandraschko:
A small comment would be great

NicolaIsotta:
Added a clarifying comment

PR #14921, opened 2026-05-20
Fix #14918: 15.0.16 DatePicker convert end date properly to 23:59:59 and not the next day

### LINKED ISSUE #14918: DatePicker: Range selection doesn't work with showTime="true"
### Describe the bug

**Description:**
When using the DatePicker to filter a table column containing date and time values (LocalDateTime), the selected date is unexpectedly increased by one day.

**Possible root cause:**
The issue seems to be caused in [DatePickerRenderer#getConvertedValue](https://github.com/primefaces/primefaces/blob/b8788199d651e9025f77fc410cf1640d4507a002/primefaces/src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java#L357), where the end date of a range is adjusted to the end of the day (line 390).

**Suggested fix:**
Please check whether this condition should use LocalDate instead of LocalDateTime. From my understanding, the end-of-day adjustment makes sense for date-only values, but not for LocalDateTime, because this already contains a concrete time component.

[Line 386:](https://github.com/primefaces/primefaces/blob/b8788199d651e9025f77fc410cf1640d4507a002/primefaces/src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java#L386)
if (end instanceof LocalDate || isDate) {
    // adjust date-only range end to the end of the selected day
}

### Reproducer

[pf-14918-1.zip](https://github.com/user-attachments/files/28144617/pf-14918-1.zip)

Input in column : 22.05.2026 **00:00** - 22.05.2026 **01:00**

Expected: 1 Entry (today @ start of day)

Result: 3 Entries (today @ start of day, today @ start of noon, today @ end of day)

Image

### Expected behavior

DataTable should filter entries within a date-time range

### PrimeFaces edition

None

### PrimeFaces version

15.0.14

### Theme

Saga

### JSF implementation

Mojarra

### JSF version

4.0

### Java version

17

### Browser(s)

All

melloware:
Fix #14918: 15.0.16 DatePicker convert end date properly to 23:59:59 and not the next day

The bug was it was assuming the date was midnight 00:00:00 already.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java
@@ -38,7 +38,6 @@ import java.text.DateFormatSymbols;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
-import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


ltlBeBoy:
To get the end of day, the correct expression would be:

```Java
endDate = endDate.toLocalDate()
                 .plusDays(1)
                 .atStartOfDay()
                 .minusNanos(1);
```

melloware:
@ltlBeBoy great suggestion I ALMOST went that way. Let me me update

melloware:
OK updated

PR #14922, opened 2026-05-20
Fix #14918 DatePicker adds one day when filtering LocalDateTime columns

### LINKED ISSUE #14918: DatePicker: Range selection doesn't work with showTime="true"

HWM-Marcin:
The end-of-day adjustment should only be applied to `LocalDate` values. A `LocalDateTime` already contains a concrete time component and should therefore not be adjusted to the end of the day again.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java
@@ -387,11 +387,11 @@ public class DatePickerRenderer extends BaseCalendarRenderer {
                     // #8351 adjust end date to end of day
                     Object end = range.get(1);
                     boolean isDate = end instanceof Date;
-                    if (end instanceof LocalDateTime || isDate) {
+                    if (end instanceof LocalDate || isDate) {
                         LocalDateTime endDate = isDate
                                 ? CalendarUtils.convertDate2LocalDateTime((Date) end)
                                 : (LocalDateTime) end;
-                        endDate = endDate.plusDays(1).minus(1, ChronoUnit.NANOS);
+                        endDate = endDate.toLocalDate().atTime(23, 59, 59);
                         range.set(1, isDate ? CalendarUtils.convertLocalDateTime2Date(endDate) : endDate);
                     }
                 }


primefaces/src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java
```diff
@@ -387,11 +387,11 @@ public Object getConvertedValue(FacesContext context, UIComponent component, Obj
                     // #8351 adjust end date to end of day
                     Object end = range.get(1);
                     boolean isDate = end instanceof Date;
-                    if (end instanceof LocalDateTime || isDate) {
+                    if (end instanceof LocalDate || isDate) {
                         LocalDateTime endDate = isDate
                                 ? CalendarUtils.convertDate2LocalDateTime((Date) end)
                                 : (LocalDateTime) end;
```
melloware, line 393: you can't do this because then the cast to (LocalDateTime) makes no sense here?

primefaces/src/main/java/org/primefaces/component/datepicker/DatePickerRenderer.java
```diff
@@ -387,11 +387,11 @@ public Object getConvertedValue(FacesContext context, UIComponent component, Obj
...
-                        endDate = endDate.plusDays(1).minus(1, ChronoUnit.NANOS);
+                        endDate = endDate.toLocalDate().atTime(23, 59, 59);
```
melloware, line 394: end date is already an end date when its a local date so no need to do toLocalDate() here.

I don't think this solution is correct your original Ticket mentions "LocalDateTime" ?

melloware (changes requested):

melloware:
cc @ltlBeBoy

PR #14923, opened 2026-05-21
Fix #14917: Datepicker - respect defaultTime when navigating months

### LINKED ISSUE #14917: DatePicker: Default time lost after navigating months
### Describe the bug

Say you have a date picker with a default time. The default time is showing correctly when opening the date picker for the first time. However, when navigating to the next month, the time resets to 0:00.

### Reproducer

The behaviour is visible in the [showcase](https://showcase.primefaces.org/ui/input/datepicker/datePickerJava8.xhtml?jfwid=4909c).

1. Select DatePicker -> DateTime -> Time (with defaults)
2. Datepicker shows with correct default time
3. Navigate to next month
4. Time gets reset to 0:00

### Expected behavior

The time should remain at the configured default value.

### PrimeFaces edition

None

### PrimeFaces version

15.0.16-SNAPSHOT

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0.15

### Java version

17

### Browser(s)

ALL

fcjm:
Fixes #14917. When navigating between months, the time was reset because the viewDate was truncated. The truncation was used to check whether the new date was still within the defined bounds (minDate, maxDate).

My proposal is to avoid truncating viewDate and instead introduce an extra variable, testDate, which is used for the bounds check.

### DIFF
--- a/primefaces/src/main/frontend/packages/datepicker/datepicker/src/jquery.ui.prime.datepicker.cjs
+++ b/primefaces/src/main/frontend/packages/datepicker/datepicker/src/jquery.ui.prime.datepicker.cjs
@@ -2500,12 +2500,10 @@ $.widget("prime.datePicker", {
                 newViewDate.setMonth(newViewDate.getMonth() - 1, 1);
             }
 
-            // previous (check first day of month at 00:00:00)
-            newViewDate = this.truncateDate(newViewDate);
-
             // #5967 check if month can be navigated to by checking last day in month
             var testDate = new Date(newViewDate.getTime()),
                 minDate = this.options.minDate;
+            testDate = this.truncateDate(testDate);
             testDate.setMonth(testDate.getMonth() + 1);
             testDate.setHours(-1);
             if (this.options.showMinMaxRange && minDate && minDate > testDate) {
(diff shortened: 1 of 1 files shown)


primefaces/src/main/frontend/packages/datepicker/datepicker/src/jquery.ui.prime.datepicker.cjs
```diff
@@ -2500,12 +2500,10 @@ $.widget("prime.datePicker", {
                 newViewDate.setMonth(newViewDate.getMonth() - 1, 1);
             }
 
-            // previous (check first day of month at 00:00:00)
-            newViewDate = this.truncateDate(newViewDate);
-
             // #5967 check if month can be navigated to by checking last day in month
             var testDate = new Date(newViewDate.getTime()),
                 minDate = this.options.minDate;
+            testDate = this.truncateDate(testDate);
```
melloware, line 2507:
```suggestion
             // previous (check first day of month at 00:00:00)
            testDate = this.truncateDate(testDate);
```

primefaces/src/main/frontend/packages/datepicker/datepicker/src/jquery.ui.prime.datepicker.cjs
```diff
@@ -2563,12 +2561,11 @@ $.widget("prime.datePicker", {
...
 
-            // next (check last day of month)
-            newViewDate = this.truncateDate(newViewDate);
-
             // #5967 check if month can be navigated to by checking first day next month
-            var maxDate = this.options.maxDate;
-            if (this.options.showMinMaxRange && maxDate && maxDate < newViewDate) {
+            var testDate = new Date(newViewDate.getTime()),
+                maxDate = this.options.maxDate;
+            testDate = this.truncateDate(testDate);
```
melloware, line 2569:
```suggestion
            // next (check last day of month)
            testDate = this.truncateDate(testDate);
```

melloware (changes requested):

melloware:
I like this idea!

melloware (approved):

melloware:
I will backport this to 15.0.16

PR #14943, opened 2026-05-30
add CLAUDE.md

christophs78:
Not sure whether you like or hate it ....

Nowdays agents like Claude Code or Codex are hard to avoid and getting state-of-the art with all according chances and risks. This would add some basic (auto-generated) guidance for Claude Code which may be extended/improved over time.
To be agent-vendor-neutral it may be an option to add a symlink from AGENTS.md to CLAUDE.md or otherwise. (https://github.com/DependencyTrack/dependency-track does such a thing.)

Feel free to say no to this if you don´t like it.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,118 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Overview
+
+PrimeFaces is a JSF/Faces component library for Java web applications. It is a multi-module Maven project combining Java backend components and TypeScript/JavaScript frontend code. Current version is `16.0.0-SNAPSHOT` (`org.primefaces:primefaces-parent`).
+
+## Module Structure
+
+| Module | Purpose |
+|---|---|
+| `primefaces` | Core library — Java Faces components + TypeScript/CSS frontend |
+| `primefaces-themes` | SASS theme compilation (Nova, Saga, etc.) via dart-sass |
+| `primefaces-cdk` | Component Development Kit — annotation processor and Maven plugin for code generation |
+| `primefaces-showcase` | Demo WAR application |
+| `primefaces-selenium` | Reusable Selenium test framework (base classes for component testing) |
+| `primefaces-integration-tests` | Selenium-based browser integration tests |
+| `primefaces-cli` | Command-line utilities |
+| `primefaces-coverage` | Code coverage aggregation (JaCoCo) |
+
+## Build Commands
+
+```bash
+# Standard build (runs tests)
+mvn clean install
+
+# Fast build — skips tests, javadoc, checkstyle
+mvn clean install -Pquick
+
+# Parallel + minified production build
+mvn clean install -T1C -Pminify
+
+# Build only the core library module
+mvn clean install -f primefaces/pom.xml
+```
+
+## Testing
+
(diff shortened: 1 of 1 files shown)


melloware (approved):

melloware:
here is what BalusC has for a PRimeFaces skill for Claude  too: https://github.com/omnifaces/claude-faces-expert/blob/main/.claude/faces/topics/primefaces.md

melloware:
@christophs78 here is a great article about what makes a good AGENTS.MD or CLAUDE.MD for open source projects

https://www.the-main-thread.com/p/coding-agent-operating-manual

he even gives an example one!  Thoughts?

jxmai:
I have general concerns about this. I know AI is the hottest topic nowadays and new terms appear every day, but if we look at the file it only describes the project using hardcoded text. Information such as library versions and structure will become outdated as we update the codebase.  

This brings us another question: who will keep the file updated as the codebase changes if Claude.md remains the canonical reference for the project structure? 

 The likely answer is no, over time the file will become inaccurate because the information is so specific.

christophs78:
Thank´s for your replies! Should have some time through this extended weekend off work.

@melloware There are a lot of good points in Markus Eisele´s AGENTS.md-guide. As of now i´m willing to pick most (all?) of it.

The IMO second argument for providing AGENTS.md / CLAUDE.md is to give the coding agent a good starting point without the need to rediscover the project over and over again. So i´ll may try to work on this too. Part of my current interesst is to extend our integration-test-coverage, which will require some manual planing / steering too.

@jxmai: Keeping AGENTS.md / CLAUDE.md is from my first experiences during the last two months not that hard. Your coding agent can do this for you (we just need to tell him) and you review / improve afterwards. Just doing this on my day job an the way from Spring Boot 3.x to 4.x.

jxmai:
Sounds great. If the main goal is to increase test coverage and overall quality, that would be a great addition. If the goal is to maintain and create components expecting AI to do the work for us, we need to be extremely cautious about that and should have at least a guideline for doing so.

I am not talking about theory, but we did experiments about that in extensions.

This is an actual example in primefaces extensions: https://github.com/primefaces-extensions/primefaces-extensions/issues/2269

What the AI model did was demonstrate strong reasoning. Everything was fine except it hallucinated in the final step, upgrading to a non‑existent version that we had to fix manually.

another one: https://github.com/primefaces-extensions/primefaces-extensions/pull/2286

The AI model initially provided useful information about the problem, but the solution could have caused harm and its tone changed depending on our input.

The technology will continue to evolve, and a problem that existed yesterday may be solved tomorrow, but we must remain vigilant.

christophs78:
IMO coding agents as of now need strong steering by an expert.

In my organization we tell our devs ...

- your realy need to know what you want (it does not help to be "somewhere" fast; you must know where you want to be "fast"(er) than in an traditional manual coding process
- you need to review the generated solution / code (does it fullfill your functional requirements; doesn´t it break some other existing things; have at least a rough look at the code and concentrate on critical parts)
- you are responsible for your commits (you commit code with your name so you take responsibilty. bad code is still bad code. does not matter whether you wrote it on your own, coppied parts from stackoverflow or generated it with coding agents)

IMO Claude Code i use primary (Codex only a bit) got better in activly asking the dev when it´s not sure. And we have plan mode to do it in a more structured way when necessary.

Issue #11440, opened 2024-02-14
DataTable: StickyHeader zIndex issue


melloware:
### Describe the bug

IF you update a table behind a dialog using the Sticky Header you get this behavior.

![image](https://github.com/primefaces/primefaces/assets/4399574/4fd6c3ad-b2b4-47ca-8c7d-99407c517358)

### Expected behavior

Sticky Header does not appear in front of the dialog.

### PrimeFaces edition

Community

### PrimeFaces version

13.0.5

### Theme

ALL

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

11

PR #14976, opened 2026-06-15
Fix #14970: Use nextStickyZIndex in DataTable resize handler

### LINKED ISSUE #14970: DataTable: Sticky resize handler assigns wrong `z-index`
### Describe the bug

The resize handler registered by datatables assigns `PrimeFaces.nextZindex()` to stickes, which can cause them to show up in front of a visible overlay (such as a modal dialogue). From what I understand so far, the resize handler should proably assign `PrimeFaces.utils.nextStickyZindex()` instead (I'm unsure of why it needs to assign a new z-index at all)?

https://github.com/primefaces/primefaces/blob/28f06579d1eaba0c01664ad1301881b27efbd508/primefaces/src/main/frontend/packages/components/src/datatable/datatable.widget.js#L5394-L5399

This should be easily reproducible by opening a dialogue on a page with sticky datatable headers and then resizing (or zooming) the browser window. The z-index of stickies is incremented beyond the z-index of the dialog mask, causing them to show up on top of the dialogue.

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

15.0.16

### Theme

_No response_

### JSF implementation

All

### JSF version

4.1

### Java version

Java 21

### Browser(s)

_No response_

iwesp:
Use the function for the next sticky z-index in DataTable's resize handler.

Potentially fixes #14970, though I'm still unsure why the z-index has to be set at all.

### DIFF
--- a/primefaces/src/main/frontend/packages/components/src/datatable/datatable.widget.js
+++ b/primefaces/src/main/frontend/packages/components/src/datatable/datatable.widget.js
@@ -5392,7 +5392,7 @@ PrimeFaces.widget.DataTable = class DataTable extends PrimeFaces.widget.Deferred
 
                 $this.stickyContainer.hide();
                 $this.resizeTimeout = PrimeFaces.queueTask(function() {
-                    $this.stickyContainer.css({ left: orginTableContent.offset().left + 'px', 'z-index': PrimeFaces.nextZindex() });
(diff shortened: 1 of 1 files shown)


melloware:
@iwesp as mentioned it was to fix this customer issue one of my customers had: https://github.com/primefaces/primefaces/issues/11440

melloware (approved):
I have confirmed this fix against both test cases and confirm this fixes the issue and doesn't re-break the other issue

PR #15011, opened 2026-06-30
fix #15009: Improved outputLabel for with SelectOneMenu

### LINKED ISSUE #15009: p:outputLabel renders an invalid for attribute when targeting a p:selectOneMenu
### Describe the bug

When a `p:outputLabel` references a `p:selectOneMenu` via its `for` attribute, the rendered `<label for>` points at an element that is not a valid label target. This is flagged by automated accessibility audits (axe / WAVE / Lighthouse), which require a <label>'s for to reference a labelable element (button, input, meter, output, progress, select, textarea).

Two distinct cases:
- Non-editable menu → for="…_label", which is a <span>. A <span> is not a labelable element → hard violation. (The attribute is also functionally dead: native <label for> can't focus a span; click-to-focus already runs through a JS bridge keyed on aria-labelledby.)
- Editable menu → for="…_focus", which is a visually hidden <input> (clip: rect(0,0,0,0)). The audit flags the label→hidden-field association, and clicking the label moves focus to the invisible input instead of the visible field the user types in (…_editableInput).

### Reproducer

```
<p:outputLabel for="@next" value="SelectOneMenu"/>
<p:selectOneMenu id="som" value="#{bean.value}">
    <f:selectItems value="#{bean.items}"/>
</p:selectOneMenu>

```
Inspect the rendered label: `<label … for="form:som_label">` where `form:som_label` is a `<span>`

### Expected behavior

The rendered HTML must satisfy the audit while preserving the screen-reader association, and `for` must never be orphaned (must always resolve to a real element in the DOM):
- Editable: `for` references the visible, labelable `_editableInput`.
- Non-editable: `for` references the hidden native `_input` <select> (a real labelable element). It is aria-hidden + tabindex="-1", but the widget already listens for `focus` on it and forwards focus to the visible interactive span, so native label click-to-focus still works. `aria-labelledby` on the visible span (unchanged) keeps the screen-reader association.

jungm:
fix #15009

`p:outputLabel for="..."` targeting a `p:selectOneMenu` rendered a `for` that pointed at a non-labelable element (`<span>`) or was left orphaned/hidden, failing axe/WAVE/Lighthouse audits.

`for` must always resolve to a real, labelable element:
- Editable menu → `for` references the visible `_editableInput`.
- Non-editable menu → `for` references the hidden native `_input` <select> (labelable, `aria-hidden` + `tabindex="-1"`). The widget already forwards `focus` on that select to the visible interactive span, so native label click-to-focus keeps working. `aria-labelledby` on the span (unchanged) preserves the screen-reader association.

Also added `InputHolder#getLabelClientId()` (defaults to `getInputClientId()`) so components can point `OutputLabel`'s `for` at a different, labelable element than their input client id.

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/outputlabel/OutputLabelRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/outputlabel/OutputLabelRenderer.java
@@ -72,6 +72,7 @@ public class OutputLabelRenderer extends CoreRenderer<OutputLabel> {
                 .add(component.getStyleClass());
 
         EditableValueHolderState forState = null;
+        String forClientId = null;
 
         final String indicateRequired = component.getIndicateRequired();
         boolean isAuto = "auto".equals(indicateRequired) || "autoSkipDisabled".equals(indicateRequired);
@@ -88,6 +89,8 @@ public class OutputLabelRenderer extends CoreRenderer<OutputLabel> {
             else {
                 callback.invokeContextCallback(context, forComponent);
             }
+
+            forClientId = callback.getForClientId();
         }
 
         boolean withRequiredIndicator = "true".equals(indicateRequired)
@@ -103,8 +106,8 @@ public class OutputLabelRenderer extends CoreRenderer<OutputLabel> {
         renderDomEvents(context, component, HTML.LABEL_EVENTS);
         renderRTLDirection(context, component);
 
-        if (!isValueBlank(_for) && forState != null && forState.getClientId() != null) {
-            writer.writeAttribute("for", forState.getClientId(), "for");
+        if (!isValueBlank(_for) && forClientId != null) {
+            writer.writeAttribute("for", forClientId, "for");
         }
 
         if (value != null) {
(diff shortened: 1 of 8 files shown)


primefaces/src/main/java/org/primefaces/component/outputlabel/OutputLabelRenderer.java
```diff
@@ -211,7 +211,7 @@ public ContextCallbackFor(String clientId, String indicateRequired, boolean isAu
         public void invokeContextCallback(FacesContext context, UIComponent target) {
             if (target instanceof InputHolder) {
                 InputHolder inputHolder = ((InputHolder) target);
-                state.setClientId(inputHolder.getInputClientId());
+                state.setClientId(inputHolder.getLabelClientId());
```
tandraschko: not sure if its good enough
state is a EditableValueHolder which is the server-side representation (UIInput)
you now overwrite its clientId with the label-clientId

jungm: the state object is being created in `ContextCallbackFor` and used only in `OutputLabelRenderer` so not really that big of an issue imo, but moved it to a seperate variable to make it more obvious what it is doing

tandraschko:
@melloware can you please also review and merge?

melloware:
Will do!

melloware (approved):

PR #15020, opened 2026-07-06
JPALazyDataModel: Adding Tests and removal of getRow implementation

### LINKED ISSUE #15013: JPALazyDataModel Integration Tests
## Current state
[JPALazyDataModel](cci:2://file:///Users/andreas.igel/Documents/entwicklung/repositories/primefaces-fork-ai-cc/primefaces/src/main/java/org/primefaces/model/JPALazyDataModel.java:70:0-526:1) currently has limited automated test coverage. The improvements done in #14577 might have effects on it too and maybe could even provide optimizations.
  
## Goal
Improving testcoverage and optimization of JPALazyDataModel:
 
## Open question
Using a real JPA-Context provider (DBUnit + H2 + Hibernate/Eclipse-Link) in test-scope or using JUnit and mocking all database connections

- [x] New unit test: JPALazyDataModelTest.java
- [x] New integration test with tracking for calls to DataTable: DataTable050.java
- [x] Validating if improvements of #14577 could be used in JPALazyDataModel too

### Additional context

_No response_

AndreasIgel:
Improving JPALazyDataModel:
1. Adding Unit test with mocking JPA and asserts on JPA-Calls
2. Adding Integration Tests with h2database, hibernate and jpa
3. Removing implementation of getRowData from JPALazyDataModel because all functionality could be solved via wrappedData and it saves a database-operation on selects

Fix #15013


tandraschko:
Lets target 17.0 for it
16.0 RC1 is around the corner

AndreasIgel:
> Lets target 17.0 for it
> 
> 16.0 RC1 is around the corner

How to do so, just remaining the PR open until Release of v16.0 is done?

And do you know why only some integration tests are succeeding? Is it something with the environment so that JPA could not be initialized? But why then not in all integration test runs?

tandraschko:
I dont think our integration tests have any JPA, AFAIR its a Tomcat.

Yeah just leave it open and work further on it

tandraschko:
any progress?

AndreasIgelCC:
Sorry for the delayed response. I assumed this could wait a bit since it's planned for a future major release. I'll review it shortly. The current implementation causes some integration test failures, and I'd like to understand and address those before moving forward.

melloware:
looks like merge conflicts?

AndreasIgelCC:
> looks like merge conflicts?

Merge conflict has been solved before. But there have been a testoptimization, which has not been tested good enough. Now it should go through the tests, but still having a deep look on the failure regarding license format... found and fixed.

AndreasIgel:
@melloware MR is ready, feel free to contact me for any questions!

primefaces/src/main/java/org/primefaces/model/JPALazyDataModel.java
```diff
@@ -320,29 +320,6 @@ protected Expression resolveFieldExpression(CriteriaBuilder cb, CriteriaQuery<?>
         return join == null ? root.get(fieldName) : join.get(fieldName);
     }
 
-    @Override
```
tandraschko, line 323: Im not sure about removing it completely.
I know that most of the times its propably inside getWrappedData().
So i propably suggest the following:
always call super.getRowData first. If its null, leave the current code.

Also i suggest optimize the parent getRowData method to something like this:

```
List<T> wrappeData = getWrappedData();
if (wrappedData != null && .size() > 0) {
            for (T item : wrappedData) {
                if (rowKey.equals(getRowKey(item))) {
                    return item;
                }
            }
}
return null;
```

AndreasIgelCC: Unless I'm missing something, the changes from #14577 should guarantee that getRowData is only called for elements on the current page during the framework's selection flow — cross-page selection objects are already preserved via getSelectionMapFromValueExpression, which reads them directly from the bean.

However, getRowData(String) is public API on SelectableDataModel, so a developer could call it directly outside the selection flow — e.g., in a custom listener or helper. In that case, the base implementation searching only getWrappedData() would return null for rows not on the current page.

So I'm fine with reintegrating it as a safety net. I'll re-add the JPA query as a fallback after super.getRowData() returns null, as you suggested. I would also switch getSingleResult() to getResultList() to avoid NoResultException when the entity doesn't exist.

AndreasIgelCC: For the second part, the for-loop does avoid some intermediate object allocations from the stream pipeline, so it's a slight improvement. In practice, the getRowKey comparison is the dominant cost either way, so the difference is marginal. I've gone ahead and made the change — happy to adjust if you'd prefer a different approach.

primefaces/src/main/java/org/primefaces/model/JPALazyDataModel.java
```diff
@@ -336,9 +337,13 @@ public T getRowData(String rowKey) {
         cq.select(root).where(criteriaBuilder.equal(root.get(rowKeyField), convertedRowKey));
 
         TypedQuery<T> query = em.createQuery(cq);
-        T result = query.getSingleResult();
+        List<T> results = query.getResultList();
+        if (results.isEmpty()) {
+            return null;
+        }
+        T result = results.get(0);
```
tandraschko: im not sure if the singleResult vs getResultList is correct
rowKey must be unique, and the old singleResult throwed a exception if not

AndreasIgel: But with getSingleResult it would throw an exception on having no findings. At the end, this has been the behavior before. Just take a decision please.

tandraschko: better leave the old behavior for now with singleResult
we can later change it if ever required

AndreasIgel: Done

PR #15027, opened 2026-07-07
SelectOneButton: rework accessibility to native-radio model (align with SelectOneRadio)

### LINKED ISSUE #15026: SelectOneButton: not accessible - orphaned outputLabel, unlabeled options, no native radio semantics
### Describe the bug

`p:selectOneButton` fails several accessibility checks (Firefox Accessibility Inspector, WAVE), especially when combined with `p:outputLabel`:

1. Invalid label target. `p:outputLabel for="selectOneButton"` renders `<label for="..."`> pointing at the component's root `<div>` (or `<span>` in custom layout). Per the HTML spec, for may only reference a labelable element — the label is dead: no click association, no accessible name contribution. `SelectOneButton` does not implement `InputHolder`, so `OutputLabelRenderer` falls back to the raw client id and never wires `aria-labelledby`.
2. Radio group has no accessible name. The button group container carries no role in default layout and is never named by the label. The label attribute of the component is not rendered at all.
3. Options are not properly labeled. The visible option text is a plain <span>; the hidden native radios rely on `aria-label` duplication. Firefox's text-label audit (FORM_NO_VISIBLE_NAME) flags the options because no label relation exists to visible text.
4. Double announcement. Screen readers see both the hidden native radios (ui-helper-hidden-accessible) and the styled option elements.
5. No standard keyboard behavior. Every option div has tabindex="0" and only Enter/Space handling — the WAI-ARIA radio group pattern (single tab stop, arrow-key navigation) is not implemented.

### Reproducer

_No response_

### Expected behavior

Same accessibility model as `p:selectOneRadio`: native radio inputs carry focus, keyboard interaction and semantics; each option text is a real `<label for>`; the group is a named `radiogroup`; `p:outputLabel` resolves to the first radio input via `InputHolder`

### PrimeFaces edition

Community

### PrimeFaces version

16.0.0-SNAPSHOT

### Theme

_No response_

### JSF implementation

All

### JSF version

any

### Java version

any

### Browser(s)

_No response_

jungm:
Fix #15026

### DIFF
--- a/docs/16_0_0/components/selectonebutton.md
+++ b/docs/16_0_0/components/selectonebutton.md
@@ -5,6 +5,34 @@ SelectOneButton is an input component to do a single select.
 ## Getting started with SelectOneButton
 SelectOneButton usage is same as selectOneRadio component, buttons just replace the radios.
 
+## Accessibility
+
+The options are backed by native radio inputs: the group is rendered as a `role="radiogroup"` element, each visible
+button text is a `label` element bound to its hidden radio input, and keyboard interaction (Tab into the group, arrow
+keys to move the selection, Space to select) is native browser behavior.
+
+Use `p:outputLabel` to give the radio group an accessible name; it renders a valid `for` attribute pointing at the
+first radio input and additionally names the group via `aria-labelledby`:
+
+```xhtml
+<p:outputLabel for="paymentMethod" value="Payment method"/>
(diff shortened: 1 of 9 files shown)


tandraschko:
@melloware can you review and retest?

melloware:
Yes I will put it on my list today.

melloware (approved):

PR #15029, opened 2026-07-07
fix #15028: SelectBooleanButtonBase implement InputHolder

### LINKED ISSUE #15028: SelectBooleanButton: outputLabel "for" produces orphaned label (points to wrapper div instead of input)
### Describe the bug

When a p:outputLabel references a p:selectBooleanButton via for, the rendered <label for="..."> points at the component's outer wrapper `<div>` instead of the actual checkbox input (clientId + "_input"). A `<div>` is not a labelable element, so accessibility checkers (e.g. WAVE) flag an "Orphaned form label".

Root cause: SelectBooleanButton does not implement org.primefaces.component.api.InputHolder. OutputLabelRenderer only resolves the input element's client id (via getLabelClientId()) and wires up aria-labelledby for InputHolder components; for all other targets it falls back to the plain component client id — the wrapper div.

Consequences:
1. Orphaned label — for targets a `<div>,` no programmatic label association (WCAG 1.3.1 / 3.3.2).
2. Wrong accessible name — the hidden checkbox only carries aria-label set to the current on/off label (e.g. "Yes"/"No"), so screen readers announce "Yes, checkbox, checked" and never the field's actual label. aria-labelledby is never rendered because OutputLabelRenderer#setAriaLabelledBy is only called for InputHolder components.
3. Label click is dead — clicking the label does not toggle the button, since for doesn't reference a form control.

p:selectBooleanCheckbox handles all of this correctly because it implements InputHolder and returns clientId + "_input" from getInputClientId(). p:selectBooleanButton renders the same hidden-checkbox structure but is missing the interface.

### Reproducer

```xhtml
<h:form id="form">
    <p:outputLabel id="lbl" for="btn" value="Newsletter"/>
    <p:selectBooleanButton id="btn" value="#{bean.newsletter}"
                           onLabel="Yes" offLabel="No"/>
</h:form>
```

Rendered output:
```xhtml
<label id="form:lbl" for="form:btn">...</label>   <!-- for points at the div below -->
<div id="form:btn" class="ui-selectbooleanbutton ...">
        <input id="form:btn_input" type="checkbox" aria-label="No"/>
    </div>
    ...
</div>
```

### Expected behavior

```
<label id="form:lbl" for="form:btn_input">...</label>
<input id="form:btn_input" type="checkbox" aria-label="No" aria-labelledby="form:lbl"/>
```

### PrimeFaces edition

None

### PrimeFaces version

16.0.0-SNAPSHOT

### Theme

_No response_

### JSF implementation

All

### JSF version

any

### Java version

any

### Browser(s)

_No response_

jungm:
fix #15028

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/selectbooleanbutton/SelectBooleanButton.java
+++ b/primefaces/src/main/java/org/primefaces/component/selectbooleanbutton/SelectBooleanButton.java
@@ -43,6 +43,16 @@ public class SelectBooleanButton extends SelectBooleanButtonBaseImpl {
 
     public static final String STYLE_CLASS = "ui-selectbooleanbutton ui-widget";
 
+    @Override
+    public String getInputClientId() {
+        return this.getClientId(getFacesContext()) + "_input";
+    }
+
+    @Override
+    public String getValidatableInputClientId() {
+        return this.getInputClientId();
+    }
+
     public String resolveStyleClass(boolean checked, boolean disabled) {
         String icon = checked ? getOnIcon() : getOffIcon();
         String styleClass = null;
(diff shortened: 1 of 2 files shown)


tandraschko:
+1

melloware (approved):

PR #15039, opened 2026-07-12
Message: set the default value of the showSummary attribute to false

NicolaIsotta:
As it was in PF 15 -> https://primefaces.github.io/primefaces/15_0_0/#/components/messages?id=attributes
https://github.com/primefaces/primefaces/blob/eb0f453dac5f0d35e7d452f0e277668647b4a6c0/primefaces/src/main/resources/META-INF/primefaces.taglib.xml#L18107-L18114

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/message/MessageBase.java
+++ b/primefaces/src/main/java/org/primefaces/component/message/MessageBase.java
@@ -59,7 +59,7 @@ public abstract class MessageBase extends UIMessage implements UINotification, W
     @Property(description = "Defines a list of keys and clientIds, which should NOT be rendered by this component. Separated by space or comma.")
     public abstract String getForIgnores();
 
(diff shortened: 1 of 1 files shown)


melloware:
ok lets see what happens!

melloware:
Looks like my change was not needed thanks for the fix!

tandraschko:
It could be i set this as default to align all components?

melloware:
this would be a change from 15.X so I should not have touched this.

tandraschko:
What do you mean?
Did you migrate the messages to cdk?

NicolaIsotta:
I checked, the 15.x default values are the same of the standard jsf "h" components
- h:message -> showDetail true and showSummary false
- h:messages -> showSummary true and showDetail false

melloware:
Yes when I migrated Message I flipped this flag I should not have

PR #15044, opened 2026-07-14
showcase: hide right sidebar on small screens

NicolaIsotta:
like other prime libraries

Before:
primefaces-showcase-zvbjj ondigitalocean app_ui_input_autoComplete_jfwid=63bc1(iPhone 14 Pro Max)

After:
localhost_8080_showcase_ui_input_autoComplete xhtml_jfwid=61e19(iPhone 14 Pro Max)


tandraschko:
Hiding is not good, its contains important use cases like Datatable

PR #15052, opened 2026-07-17
Fix #15051: DatePicker mask="true" generates wrong Inputmask format

### LINKED ISSUE #15051: DatePicker: mask="true" generates wrong Inputmask format
### Describe the bug

When using <p:datePicker> with mask="true" and a pattern containing both months and minutes (e.g. pattern="dd/MM/yyyy HH:mm:ss"), the masked input updates the wrong segments. For example, incrementing the seconds spinner in the time picker writes the digit into the minutes slot of the masked input, while the seconds slot remains as placeholder characters (__).

This works correctly in 15.0.17.

Root cause:
- https://github.com/RobinHerbots/Inputmask/issues/2834
- https://github.com/primefaces/primefaces/blob/v16.0.0-RC1/primefaces/src/main/frontend/package.json

**Workaround**: pass the mask explicitly using the Java pattern instead of mask="true", e.g. mask="dd/MM/yyyy HH:mm:ss", which bypasses convertPattern() and matches the new Inputmask notation directly.

Image

### Reproducer

```xhtml
<h:form>
    <p:datePicker id="date" pattern="dd/MM/yyyy HH:mm:ss" mask="true"
                  value="#{testView.date}" showIcon="true"/>
</h:form>
```

Steps:
1. Open the date picker and select any date.
2. Use the seconds spinner (up arrow) to increment seconds.
3. Observe the masked input: the digit appears in the minutes segment (e.g. 05/06/2026 00:01:__) instead of the seconds segment.

### Expected behavior

With mask="true" and pattern="dd/MM/yyyy HH:mm:ss", the generated Inputmask inputFormat should map each mask segment to the correct date/time part.

### PrimeFaces edition

Community

### PrimeFaces version

16.0.0-RC1

### Theme

Custom

### JSF implementation

Mojarra

### JSF version

4.1.7

### Java version

25

### Browser(s)

_No response_

wsaca:
Inputmask 5.0.10-beta.32+ (RobinHerbots/Inputmask#2834) switched to standard Unicode notation (MM=month, mm=minutes), so the M-case swapping in UICalendar#convertPattern() produced an inverted mask.

Fix #15051

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datepicker/DatePicker017.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datepicker/DatePicker017.java
@@ -1,0 +1,48 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2026 PrimeTek Informatics
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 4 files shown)


melloware (approved):

melloware:
Excellent!

melloware:
Hmm ITs failed?

wsaca:
I tested it locally using firefox, and it seems datePicker.clear(); is not working here...

melloware (approved):

PR #15056, opened 2026-07-20
optimize memory consumption for dataExporter[type=PDF]

bvfalcon:
In current implementation component dataExporter with type=PDF and batchSize=<any number> stores data from lazyDataMode in memory (structure PdfPTable). After all data is read this PdfPTable is used for creating table in PDF. This is big (and unnecessary) memory overhead. Current memory consumption could be visualized with diagram:

dataexporter-pdf

Red line -- memory consumption in current implementation
Blue line -- memory consumption with adjustments of this PR. 

As a key to optimization is a feature [LargeElement#setComplete(boolean)](https://github.com/LibrePDF/OpenPDF/blob/e5e0de83ef4369d4bfd2ecdf1b01aed639a0d574/openpdf-core/src/main/java/org/openpdf/text/pdf/PdfPTable.java#L1691) of OpenPDF used, demonstrated in details with [this unit-test](https://github.com/LibrePDF/OpenPDF/blob/e5e0de83ef4369d4bfd2ecdf1b01aed639a0d574/openpdf-core/src/test/java/org/openpdf/text/pdf/TablePdfTest.java#L31).

Adjustments of PR:
1) solve the problem OOM, described by issue #10656, for exporting to PDF 
2) make possible export process interruption from other threads

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
+++ b/primefaces/src/main/java/org/primefaces/component/datatable/export/DataTableExporter.java
@@ -34,10 +34,14 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 import jakarta.faces.context.FacesContext;
 
 public abstract class DataTableExporter<P, O extends ExporterOptions> extends TableExporter<DataTable, P, O> {
+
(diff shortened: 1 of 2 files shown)


melloware (changes requested):
Build is failing?

melloware (approved):
LGTM

tandraschko:
Will review next week

PR #15063, opened 2026-07-23
fix #15057: colspan is counted only from active checkboxex ignoring static columns

### LINKED ISSUE #15057: ColumnToggler: colspan is counted only from active checkboxex ignoring static columns
### Describe the bug

The render() method in columntoggler.widget.js intentionally excludes columns marked with .ui-static-column:

```this.columns = this.thead.find('> tr > th:not(.ui-static-column)');```

As a result, static columns are not represented in the ColumnToggler overlay, which is expected behavior.

However, calculateColspan() determines the colspan using only the active checkboxes in the ColumnToggler overlay:
```
calculateColspan() {
    return this.itemContainer
        .find('> .ui-columntoggler-item > .ui-chkbox > .ui-chkbox-box.ui-state-active')
        .length;
}
```
Because static columns are not included in the overlay, they are also omitted from the colspan calculation.

### Reproducer

data table with no records found row -> change columns using column toggler

data table with expanded row -> change columns using column toggler

### Expected behavior

calculateColspan() should include both:

- visible toggleable columns, and
- static columns (.ui-static-column).

In other words, the calculated colspan should account for all visible columns in the table, including those that cannot be toggled.

### Suggested fix

render() should continue to exclude .ui-static-column from the ColumnToggler overlay, as static columns are not user-configurable.

However, calculateColspan() should include all visible static columns when calculating the final colspan. Static columns that are hidden (e.g. via .ui-helper-hidden) should not be counted.

For example:
```
calculateColspan() {
    const visibleColumns = this.itemContainer
        .find('> .ui-columntoggler-item > .ui-chkbox > .ui-chkbox-box.ui-state-active')
        .length;

    const visibleStaticColumns = this.thead
        .find('> tr > th.ui-static-column:not(.ui-helper-hidden)')
        .length;

    return visibleColumns + visibleStaticColumns;
}
```
This ensures that the calculated colspan matches the actual number of visible table columns, preventing incorrect rendering of empty rows, expanded rows, and other rows that span the full table width.

### PrimeFaces edition

Community

### PrimeFaces version

16.0.0-RC1, 15

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4

### Java version

21

### Browser(s)

_No response_

vonnai:
calculateColspan() should include both:

- visible toggleable columns, and
- static columns (.ui-static-column).

In other words, the calculated colspan should account for all visible columns in the table, including those that cannot be toggled.

### DIFF
--- a/primefaces/src/main/frontend/packages/components/src/columntoggler/columntoggler.widget.js
+++ b/primefaces/src/main/frontend/packages/components/src/columntoggler/columntoggler.widget.js
@@ -681,7 +681,15 @@ PrimeFaces.widget.ColumnToggler = class ColumnToggler extends PrimeFaces.widget.
      * @return {number} The calculated `colspan` for the rows.
      */
     calculateColspan() {
-        return this.itemContainer.find('> .ui-columntoggler-item > .ui-chkbox > .ui-chkbox-box.ui-state-active').length;
+	const visibleColumns = this.itemContainer
+            .find('> .ui-columntoggler-item > .ui-chkbox > .ui-chkbox-box.ui-state-active')
+            .length;
+
+        const visibleStaticColumns = this.thead
+            .find('> tr > th.ui-static-column:not(.ui-helper-hidden)')
+            .length;
+
+        return visibleColumns + visibleStaticColumns;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


tandraschko:
indentations are not correct

vonnai:
fixed.

melloware (approved):

melloware:
i will backport this to 15.0.18

PR #15066, opened 2026-07-24
MyFaces 4.0.4, 4.1.4 SNAPSHOTS

melloware:
Just testing latest MyFaces snapshots


tandraschko:
votes ended, release will follow soon

i think we can close for now until we check for the upcoming 5.0 releases

PR #15069, opened 2026-07-24
fix #15058: paginator jump-to-page gets a stable id and bounded numeric input

### LINKED ISSUE #14944: Paginator: page html input use number type attribute
### Description

For easier accessibility, especially on mobile devices keyboard it would be nice if paginator page select input field would have following attributes:

`type="number" min="1" max="10" step="1"`

Maybe `size` is als a good idea calculated based on the string length of 'max' value. Then it would not default to a large input field if only 9 pages are present. But in my experience `size` is tricky depending on browser or if spinner buttons are rendered like in chrome.

Paginator page select input html should be like:

`<input class="ui-paginator-jtp-input ui-inputfield ui-inputtext ui-widget ui-state-default ui-state-filled" value="1" type="number" min="1" max="10" step="1" aria-label="Zum Eingabefeld springen">`

### Additional context

_No response_

### LINKED ISSUE #15058: Paginator "jump to page" input has no id -> loses focus after AJAX, and is not a bounded number input
### Describe the bug

JumpToPageInputRenderer renders only a bare input (verified against 15.0.17):

    writer.startElement("input", null);
    writer.writeAttribute("class", UIPageableData.PAGINATOR_JTP_INPUT_CLASS, null);
    writer.writeAttribute("value", pageable.getPage() + 1, null);
    writer.endElement("input");

This causes two problems:

1. Focus/scroll loss after every jump (the main defect).
   PrimeFaces' AJAX focus restore only re-focuses elements that carry an id.
   Because this input has none, focus is lost after typing a page number and
   triggering the jump. When the paginator sits at the bottom of a long table,
   the scroll position is lost as well. Keyboard and screen-reader users cannot
   jump through several pages in a row without re-locating the field each time.

2. Not a real numeric control.
   The element has no type and no min/max, so it renders as a plain text input
   with no browser-side clamping to the valid page range, even though that range
   (1..pageCount) is available at render time.

Note: the sibling JumpToPageDropdownRenderer has the same id omission -- its ``<select>`` is rendered with only a class and value and no id either -- so this is a generic deficiency of the jump-to-page paginator elements, not specific to the input variant.

### Reproducer

Steps (against primefaces-test on the latest 15.X branch, i.e. 15.0.x-SNAPSHOT, or the PrimeFaces showcase DataTable paginator with a paginatorTemplate containing ``{JumpToPageInput}``):

1. Add a ``<p:dataTable>`` with ``paginator="true"`` and
   ``paginatorTemplate="{JumpToPageInput} {PageLinks}"``, backed by a lazy model with many pages.
2. Click into the jump-to-page input, type a page number, and confirm the jump.
3. Observe: the input no longer has focus after the AJAX update. Inspecting the DOM shows the ``<input class="ui-paginator-jtp-input">`` has no id, no type, and no min/max.

### Expected behavior

JumpToPageInputRenderer should render an input that:

- carries a stable id (e.g. clientId + separator + "jumpToPage") so PrimeFaces' existing focus-restore re-focuses it after the jump
- is ``type="number"`` with ``min="1"`` and ``max="{pageCount}"``, so the browser enforces the valid range.

```
writer.startElement("input", null);
writer.writeAttribute("id", pageable.getClientId(context)
            + UINamingContainer.getSeparatorChar(context) + "jumpToPage", null);
writer.writeAttribute("class", UIPageableData.PAGINATOR_JTP_INPUT_CLASS, null);
writer.writeAttribute("type", "number", null);
writer.writeAttribute("min", 1, null);
writer.writeAttribute("max", Math.max(pageable.getPageCount(), 1), null);
writer.writeAttribute("value", pageable.getPage() + 1, null);
writer.endElement("input");
```
The element must keep the ui-paginator-jtp-input class and its position as a
direct child of .ui-paginator-navigator, since the Paginator widget binds its
change/clamp/value-sync handlers on that class.

(The same stable-id fix would ideally also be applied to JumpToPageDropdownRenderer, which currently has no id on its ``<select>`` either.)

### PrimeFaces edition

Community

### PrimeFaces version

15.0.17

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.1

### Java version

21

### Browser(s)

_No response_

sunriseO:
## Fixes
  Closes #15058, and implements the numeric-input request from #14944.

  ## Problem
  The paginator's jump-to-page controls (`{JumpToPageInput}` / `{JumpToPageDropdown}`) had two defects:

  1. **Focus lost after AJAX.** The elements were rendered without an `id`.
     PrimeFaces' AJAX focus restoration (`PrimeFaces.ajax.Response.handleReFocus`)
     re-focuses the previously active element **by id**, so after every page jump
     focus was lost — a real problem for keyboard and screen-reader users on long tables.

  2. **No numeric validation.** The input was a plain text field with no `type="number"`,
     `min`, or `max`, so the browser could not validate the entered page against the
     valid `1..pageCount` range, and mobile devices did not show a numeric keypad.

  ## Changes
  - **Stable ids** on both the jump-to-page `<input>` and `<select>` (`id` + `name`),
    so focus survives AJAX updates.
  - The input now renders as `type="number"` with `min="1"` and `max="{pageCount}"`
    for browser-level range validation and a numeric mobile keyboard.
  - Ids stay **unique** when `paginatorPosition="both"`: the `top`/`bottom` position is
    appended only in that case (`…:jumpToPage_top` / `…:jumpToPage_bottom`), mirroring the
    existing `…_paginator_top/bottom` container ids. The rendered position is forwarded to the
    element renderers via a `FacesContext` attribute (no change to the public
    `PaginatorElementRenderer` interface); the id-building logic is shared as a
    `default` method on that interface.
  - **Input width now fits the live page count** (relates to the width discussion in #14944):
    the renderer emits the page-count digit count as a `--p-jtp-digits` custom property and
    `paginator.css` derives the width from it (`calc(var(--p-jtp-digits, 3) * 1ch + 2.5rem)`).
    The `width` is declared in the class rule, so themes/apps can still override it with plain CSS.
  - **Showcase:** split the DataTable paginator demo into dedicated `paginatorTemplate`
    examples — page links, jump-to-page by input, and jump-to-page by dropdown — alongside
    the existing basic example.

  ## Notes
  - `size` was intentionally **not** used to constrain width: it is ignored on `type="number"`
    inputs, so the width is controlled via CSS instead.

  ## Testing
  - `mvn -pl primefaces install` builds cleanly.
  - Verified via the updated showcase paginator page: with `paginatorPosition="both"`, two
    jump-to-page inputs render with distinct ids (`…jumpToPage_top` / `…jumpToPage_bottom`),
    both `type="number"` with correct `min`/`max`, and focus is retained in the same input
    after an AJAX page jump.

### DIFF
--- a/primefaces-showcase/src/main/webapp/ui/data/datatable/paginator.xhtml
+++ b/primefaces-showcase/src/main/webapp/ui/data/datatable/paginator.xhtml
@@ -10,16 +10,19 @@
     </ui:define>
 
     <ui:define name="description">
-        DataTable provides a highly customizable ajax paginator.
+        DataTable provides a highly customizable ajax paginator. The paginatorTemplate selects which elements are
+        rendered, e.g. page links, a jump-to-page input or a jump-to-page dropdown.
     </ui:define>
 
     <ui:param name="documentationLink" value="/components/datatable"/>
     <ui:param name="widgetLink" value="DataTable-1"/>
 
     <ui:define name="implementation">
-        <div class="card">
-            <h:form>
-                <p:dataTable var="customer" value="#{dtPaginatorView.customers}" rows="10"
+
+        <h:form id="form">
+            <div class="card">
+                <h5 style="margin-top:0">Basic</h5>
+                <p:dataTable id="basic" var="customer" value="#{dtPaginatorView.customers}" rows="10"
                              paginator="true" paginatorPosition="both"
                              paginatorTemplate="{CurrentPageReport} {FirstPageLink} {PreviousPageLink} {PageLinks} {NextPageLink} {LastPageLink} {RowsPerPageDropdown}"
                              currentPageReportTemplate="{startRecord}-{endRecord} of {totalRecords} records"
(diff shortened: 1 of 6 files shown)


melloware:
This is great thank you @sunriseO !

melloware (approved):

tandraschko:
+1
@melloware can you port to master?

melloware:
Yes I will take care of this tomorrow.

PR #15071, opened 2026-07-26
Fix #15035: clean up dynamic AutoComplete panels on refresh

### LINKED ISSUE #15035: AutoComplete: dynamic panel is orphaned (visible, unclosable) when the component is ajax-updated while the panel is open
### Describe the bug

An <p:autoComplete dynamic="true"> whose suggestion panel is open at the moment the component is ajax-updated leaves the panel permanently visible: after the update, the panel node is still in the DOM under <body>, but the re-initialized widget no longer references it, so hide() is a silent no-op and neither ESCAPE, blur, nor clicking outside closes it. The panel floats over the page until the user happens to open the same autocomplete again (the next dynamic load appends a fresh panel and appendDynamicOverlay sweeps the old node).

This is easy to hit in real applications with <p:ajax event="change" update="<the component itself>"/> (or any listener that updates a region containing the autocomplete): selecting an item starts the panel hide transition and fires the ajax update concurrently; if the update response is applied before the hide completes (slow client, CPU-throttled machine), the panel is orphaned.

Root cause analysis

Three pieces interact:

1. With dynamic="true", the panel is not part of the component's rendered markup — AutoCompleteRenderer only encodes the panel when !isDynamic() or for a _dynamicload request. The widget builds the panel client-side on first query and appends it to   cfg.appendTo (default @(body)).
2. On widget refresh, the refresh listener registered by PrimeFaces.utils.registerDynamicOverlay calls PrimeFaces.utils.cleanupDynamicOverlay, which only removes appended nodes when more than one element with the panel id exists:
  cleanupDynamicOverlay: function(widget, overlay, overlayId, appendTo) {
      if (widget.cfg.appendTo && $("[id='" + overlayId + "']").length > 1) { ... }
  }
3. That guard is correct for non-dynamic components (the refreshed markup contains a new panel, giving 2 nodes). For a dynamic component the refreshed markup contains no panel, the count stays at 1, and the old body-appended panel is skipped.
4. init() then does this.panel = this.jq.children(this.panelId), which is an empty set (the leftover panel is under <body>, not a child of the component). From here on, hide()'s guard this.panel.is(':visible') is false on the empty set, so the visible panel can never be closed by the widget; the ESCAPE keydown handler routes through hide() and is equally dead.  

### Reproducer

<h:form id="form">
  <p:autoComplete id="ac" widgetVar="acWV" value="#{testBean.value}"
				  completeMethod="#{testBean.complete}"
				  dynamic="true" queryMode="server" dropdown="true">
	  <p:ajax event="change" update="ac"/>
  </p:autoComplete>
</h:form>

Deterministic reproduction (no timing luck needed):

1. Click the dropdown button and wait for the suggestion panel to appear.
2. While the panel is open, run in the browser console (simulates the change-event update landing before the panel has closed):
PrimeFaces.ajax.Request.handle({source: 'form:ac', process: 'form:ac', update: 'form:ac', formId: 'form'});
3. Observe after the response:
document.querySelectorAll("[id='form:ac_panel']").length  // 1 - still in the DOM, visible
PF('acWV').panel.length                                   // 0 - widget owns no panel
PF('acWV').hide()                                         // no-op, panel stays visible
3. ESCAPE and clicking elsewhere also do nothing. Clicking the dropdown button again "self-heals" (new dynamic load replaces the node).

The interactive variant: select a suggestion item on a CPU-throttled machine (DevTools performance throttling helps) so the change-update response arrives before the hide transition finishes — same end state.

### Expected behavior

After the component is refreshed, the previously appended dynamic panel is either removed or re-adopted by the widget; no visible panel remains that the widget cannot close.

Possible fix directions: have cleanupDynamicOverlay also remove a single leftover node when the refreshed widget is dynamic (the > 1 guard doesn't apply — the fresh markup can never contain the panel), or have AutoComplete.refresh remove/adopt the body-appended panel when cfg.dynamic is set and this.panel resolves empty.

### PrimeFaces edition

Community

### PrimeFaces version

15.0.13

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

Mojarra (WildFly), Jakarta Faces 4.x

### Java version

21

### Browser(s)

 Chrome 147 (not browser-specific)

mvanhorn:
Keep the correction in the AutoComplete lifecycle so other consumers of the shared dynamic-overlay utility retain their existing duplicate-overlay semantics. With `dynamic="true"`, AutoComplete renders its suggestion panel only during the first dynamic query and moves that panel to the configured `appendTo` container, normally `body`.

Happy path: load and display the dynamic suggestion panel, invoke the AJAX self-update while it remains open, and verify the old panel ID is absent from the DOM, the refreshed widget owns no panel until queried again, and no JavaScript errors occur; Edge cases: refresh the dynamic AutoComplete before its panel has ever loaded, then query it; also query again after an open-panel refresh and verify exactly one newly loaded panel appears with working suggestions and hide behavior.

Fixes #15035

AI was used for assistance.

### DIFF
--- a/.osc-metadata/sync.json
+++ b/.osc-metadata/sync.json
@@ -1,5 +1,0 @@
-{
-  "fork_synced_at": "2026-07-26T10:22:54.686307+00:00",
-  "commits_behind_before_sync": 0,
-  "action_taken": "noop"
-}
--- a/.osc-metadata/sync.json
+++ b/.osc-metadata/sync.json
@@ -1,0 +1,5 @@
+{
+  "fork_synced_at": "2026-07-26T10:22:54.686307+00:00",
+  "commits_behind_before_sync": 0,
+  "action_taken": "noop"
+}
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/autocomplete/AutoComplete007.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/autocomplete/AutoComplete007.java
@@ -1,0 +1,52 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2026 PrimeFaces
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.integrationtests.autocomplete;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+
+import jakarta.faces.view.ViewScoped;
+import jakarta.inject.Named;
+
+import lombok.Data;
+
+@Named
+@ViewScoped
+@Data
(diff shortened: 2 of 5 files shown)


melloware:
@mvanhorn are you a real person?  I looked at your GitHub profile and could not tell?

mvanhorn:
Yes -- Matt Van Horn. Everything's on my profile at github.com/mvanhorn (bio, projects, contribution history), and I'm [@mvanhorn on X](https://x.com/mvanhorn) and [linkedin.com/in/mattvanhorn](https://linkedin.com/in/mattvanhorn) if that's easier to verify. Previously co-founded June (acquired by Weber) and the company that became Lyft.

Straight answer on what's probably behind the question: I use AI tooling when I work on contributions. I pick the issues, review the output before it goes out, and test locally -- this one has an integration test (`AutoComplete007Test`) because I wanted to show the panel actually gets cleaned up rather than just claim it. If that's not welcome in PrimeFaces, say so and I'll respect it. Otherwise I'm happy to walk through any part of this one.

melloware:
@mvanhorn no this PR is fantastic. I am always just puzzled when someone out of the blue with your profile makes a PrimeFaces fix!  It is very welcome indeed. 

I was more puzzled as to why and when i looked at your profile i wasn't sure if it was some AI manufactured profile!

Thank you very much for the PR!

mvanhorn:
Appreciate the merge and your steer to keep the fix inside the AutoComplete lifecycle. Thanks for making sure other overlay consumers kept their semantics.

PR #15081, opened 2026-07-28
fix #15080: appendMessage() must only fill its own severity box

### LINKED ISSUE #15080: Messages: appendMessage() leaks new message into every already-rendered severity box
### Describe the bug

`PrimeFaces.widget.Messages.appendMessage()` (client-side API, called e.g. via `PF('messages').appendMessage({severity, summary})`) leaks the newly appended message into every severity box that already exists in the container, instead of only the box matching the message's own severity.

Root cause, in the minified widget code (`components.js`):

    appendMessage(t){
      let i=this.getAriaLabel("close"),
          s=this.jq.children("div.ui-messages-"+t.severity);
      s.length===0 && (s=this.jq.append('<div class="ui-messages-'+t.severity+'">...<ul></ul></div>'));
      s.find("ul").append('<li>...</li>');
    }

`jQuery.append()` returns the **caller**, not the newly created element. So when a severity box does not exist yet, `s` is reassigned to `this.jq` — the whole messages container — instead of the new div. `s.find("ul")` then matches every `<ul>` already inside the container (all previously created severity boxes), so the new `<li>` is appended to all of them, not just the new box. Once a box for a given severity already exists, the bug doesn't trigger (the `if` branch is skipped, `s` stays correctly scoped) — so only the *first* message of each new severity leaks.

### Reproducer

https://github.com/Random-user420/primefaces-test/commit/e2fc2ae4f2d54999c1a269d295cba41287669925

### Reproduction steps
Call `appendMessage` for 4 messages, one severity change at a time:

    var messages = PF('messages');
    messages.clearMessages();
    messages.appendMessage({severity: 'error', summary: 'error message'});
    messages.appendMessage({severity: 'warn',  summary: 'warn message 1'});
    messages.appendMessage({severity: 'warn',  summary: 'warn message 2'});
    messages.appendMessage({severity: 'info',  summary: 'info message'});

### Expected behavior
- error box: 1 message ("error message")
- warn box: 2 messages ("warn message 1", "warn message 2")
- info box: 1 message ("info message")

### Actual behavior
- error box: 3 messages ("error message", "warn message 1", "info message")
- warn box: 2 messages ("warn message 1", "info message")
- info box: 1 message ("info message")

### PrimeFaces edition

None

### PrimeFaces version

16.0.0-RC1

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0

### Java version

25

### Browser(s)

any

Mika-techs:
fix #15080: appendMessage() must only fill its own severity box

Messages.appendMessage() created a missing severity box with this.jq.append(html). jQuery's append() returns the container it was called on, so the local variable ended up referencing the whole messages container instead of the new box. The following find('ul').append(...) therefore matched the list of every severity box already present and added the message to all of them.

As a result the first client-side message of each new severity also showed up in every previously created severity box, e.g. appending error, warn, warn, info in that order left the error box with an extra copy of the first warn message. Later messages of the same severity were unaffected because the box already existed and the branch was skipped.

Build the box with $(html).appendTo(this.jq) instead, which returns the new element, and add an integration test that appends messages across four severities and asserts each box only contains its own.

### DIFF
--- a/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/messages/Messages002Test.java
+++ b/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/messages/Messages002Test.java
@@ -27,12 +27,19 @@ import org.primefaces.selenium.AbstractPrimePage;
 import org.primefaces.selenium.AbstractPrimePageTest;
 import org.primefaces.selenium.PrimeSelenium;
 import org.primefaces.selenium.component.Messages;
+import org.primefaces.selenium.component.model.Msg;
+import org.primefaces.selenium.component.model.Severity;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
 
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.support.FindBy;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
 public class Messages002Test extends AbstractPrimePageTest {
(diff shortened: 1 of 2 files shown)


melloware:
Running the build now.  I will backport this to 15.0.18 after everything passes

Issue #5296, opened 2019-11-03
Datatable: row reorder doesn't work with headerRow/summaryRow (IndexOutOfBoundsException)


joergmattes:
## 1) Environment
- PrimeFaces version: 7.0.8
- Does it work on the newest released PrimeFaces version? Version? No
- Does it work on the newest sources in GitHub? Don't think so
- Application server + version: Tomcat 9.0.27
- Affected browsers: All

## 2) Expected behavior
When I have a datatable with row reorder and headerRow, I expect the reorder to work correctly

## 3) Actual behavior
When reordering a row e.g. to the last row, PF throws an IndexOutOfBoundsException.

## 4) Steps to reproduce
Take the showcase SummaryRow example, add draggableRows="true" and then drag a row **towards the end**.

## 5) Sample XHTML
```
<p:dataTable draggableRows="true" sortBy"xx" ...>
    <headerRow>...</headerRow>
    ...
</p:dataTable>
```

## 6) My Analysis
The problem is that datatable.js bases the toIndex on all rows including the additional header/summary rows (`ui.item.index()`)...
```
    makeRowsDraggable: function() {
            ...
            update: function(event, ui) {
                var fromIndex = ui.item.data('ri'),
                toIndex = $this.paginator ? $this.paginator.getFirst() + ui.item.index(): ui.item.index();
            ...
```
... whereas DraggableRowsFeature.java bases the toIndex only on the actual data, not counting the header/summary rows. Therefore the toIndex is too high when a row is dragged towards the end of the table and Java throws an IndexOutOfBoundsException: `list.subList(fromIndex, toIndex + 1)`

Unfortunately, I haven't found a good patch. At the moment, I fix it in DraggableRowsFeature.java with a hack. Maybe it's better to fix it on the JS side, though.

melloware:
Reproducer: 
[pf-5296.zip](https://github.com/primefaces/primefaces/files/4451393/pf-5296.zip)

melloware:
@joergmattes can you post your hack of DraggableRowsFeature please?

joergmattes:
Sure. It's on the JSF-side, and specific for my code though. It basically calculates the index by taking into account the row groups until that index.

DraggableRowsFeature:
```
   public void decode(FacesContext context, DataTable table) {
...
            if (value instanceof List) {
                List list = (List) value;
                // hack start
                toIndex = getCorrectToIndex(table, toIndex);
                // hack end
...
         }
    }
[...]

PR #15088, opened 2026-07-29
fix #15087: DataTable rowReorder must ignore the dragged row's own expansion row

### LINKED ISSUE #15087: DataTable: rowReorder sends a toIndex one position too far up when the dragged row is expanded
### Describe the bug

When a `p:dataTable` has `draggableRows="true"` together with a `p:rowToggler` / `p:rowExpansion`, dragging an **expanded** row reports the wrong drop position. The `rowReorder` event arrives with a `toIndex` that is one position too far up, and the row is moved one slot beyond where it was dropped. Dragging the same row while it is collapsed works correctly.

Example: a table with 4 rows, expand the last one (index 3) and drop it between index 1 and index 2. Expected `toIndex=2`, actual `toIndex=1`.

Cause is in `DataTable.makeRowsDraggable()` (`datatable.widget.js`, `update` callback of the jQuery UI sortable). The drop index is taken from the DOM index of the dragged row and then decremented for every `ui-rowgroup-header` / `ui-expanded-row-content` row that is walked over (added in #6589 for #5296 / #6557):

```js
if (isDirectionUp) {
    for (let i = 0; i <= toIndex; i++) {
        fromNode = fromNode.next('tr');
        if (fromNode.hasClass('ui-rowgroup-header') || fromNode.hasClass('ui-expanded-row-content')) {
            toIndex--;
        }
    }
} else { ... }
```
Walking forward from the dragged row immediately hits the dragged row's own expansion row, so toIndex is decremented once too often.

Counting the data rows that precede the dragged row instead is direction independent and never sees the dragged row's own expansion row:

```js
var itemIndex = ui.item.prevAll('tr').not('.ui-rowgroup-header').not('.ui-expanded-row-content').length;
```

I have a PR ready in a few minutes with this fix plus an integration test that fails on master and passes with the fix.

### Reproducer

https://github.com/Mika-techs/primefaces-test/tree/datatable-rowreorder-expanded-row-wrong-toindex

The view is a single list-backed `p:dataTable` with `draggableRows="true"`, a `p:rowToggler`, a `p:rowExpansion` and a `rowReorder` listener that prints the received indices.

1. `mvn clean jetty:run` and open http://localhost:8080/test.xhtml
2. Expand the last row ("The Dark Side of the Moon", index 3) using the row toggler
3. Drag that row one position up, dropping it between "Back in Black" and "The Bodyguard"
4. The message shows `fromIndex=3, toIndex=1` and the row is placed at index 1 instead of index 2
5. Collapse the row again and repeat the same drag — the message now correctly shows `fromIndex=3, toIndex=2`

The same happens with nested/expanded rows in a real application and with row group headers.

### Expected behavior

The reported `toIndex` (and the resulting model order) should be the position the row was dropped at, regardless of whether the dragged row is expanded. An expanded row should behave exactly like the same row when collapsed.

### PrimeFaces edition

Community

### PrimeFaces version

16.0.0-RC1

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0.20

### Java version

25

### Browser(s)

Google Chrome 135.0.7049.95, Linux

Mika-techs:
Fixes #15087

### Problem

`DataTable.makeRowsDraggable()` derives the drop index from the DOM index of the dragged row and then decrements it for every `ui-rowgroup-header` / `ui-expanded-row-content` row it walks over (added in #6589 for #5296 / #6557). Walking forward from the dragged row also hits the dragged row's **own** expansion row, so dropping an expanded row reports a `toIndex` one position too far up and the row is moved one slot beyond the drop target. Collapsed rows are unaffected.

### Fix

Count the data rows that precede the dragged row instead:

```js
var itemIndex = ui.item.prevAll('tr').not('.ui-rowgroup-header').not('.ui-expanded-row-content').length;
```
Direction independent, never sees the dragged row's own expansion row, and keeps the #5296 / #6557 behavior. The Math.max(toIndex, 0) clamp is dropped because a .length cannot be negative. Rows of other kinds (e.g. ui-datatable-summaryrow) are still counted as data rows — unchanged behavior, not addressed here.

Test

Second commit adds the DataTable050 triad: the same drag is performed once with the row collapsed (control) and once expanded, asserting the reported fromIndex/toIndex and the resulting model order.

Verified both ways with mvn verify -f primefaces-integration-tests/pom.xml -Pintegration-tests,headless,chrome,mojarra-4.0 -Dit.test=DataTable050Test:

- with the fix: Tests run: 2, Failures: 0
- with the fix reverted: reorderExpandedRow fails with expected: <2> but was: <1>, reorderCollapsedRow still passes

mvn clean install (unit tests included) is green. The test commit is separate so it can be dropped if the manual jQuery UI drag stepping is not wanted — the suite has no drag helper and the DataTable selenium wrapper exposes no reorder API.

Not run: the full IT suite, MyFaces, CSP and non-Chrome browsers.

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datatable/DataTable050.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datatable/DataTable050.java
@@ -1,0 +1,71 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2026 PrimeFaces
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.integrationtests.datatable;
+
+import org.primefaces.event.ReorderEvent;
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import jakarta.annotation.PostConstruct;
+import jakarta.faces.view.ViewScoped;
+import jakarta.inject.Inject;
+import jakarta.inject.Named;
+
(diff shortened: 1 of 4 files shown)


melloware:
@Mika-techs  looks like all tests are passing and thanks for the new IT.

melloware (approved):

PR #15090, opened 2026-07-29
Ci hardening

Totara-thib:
Hi, drive-by CI hardening in two commits, one logical change each.

**Commit 1 pins every action to its commit sha**, versions kept as comments. A tag like `@v6` is a movable pointer: whoever controls the action, or anyone who compromises it, can re-point it and your next run executes their code with the job's token. The exposed spots here are `deploy-showcase.yml` (ghcr push and DigitalOcean deploy) and the publish workflows with their registry and signing secrets. Same pattern as the tj-actions/changed-files incident (CVE-2025-30066). Dependabot keeps opening its bump PRs exactly as before, it understands sha pins with a version comment.

**Commit 2 adds `permissions:` blocks to the four workflows that had none.** Without one, the GITHUB_TOKEN inherits the repository default scope. Each scope was derived from what the workflow actually does with the token, not guessed:

- `build.yml`: `contents: read` (Sonar only reads PR metadata)
- `publish-snapshot.yml`: `contents: read` (deploys through Sonatype and GPG secrets, the GitHub token is unused)
- `publish-vdldoc.yml`: `contents: write` (the last step commits the regenerated docs back)
- `deploy-showcase.yml`: `contents: read` + `packages: write` (the showcase image push to ghcr.io authenticates with GITHUB_TOKEN)

The issue automation, nightly, stale and release workflows already declare scoped permissions, so they are untouched.

One heads up: if the org uses an Actions allowlist in settings, patterns written against tags (like `owner/action@v1`) stop matching once refs are shas and workflows refuse to start. Entries need to be `owner/action@*` in that case.

All shas were resolved from the upstream repos and cross checked against their release tags. No workflow logic changes.

Found with the Plumber CLI (https://github.com/getplumber/plumber), verified on master. I will also open a PR which adds it to CI so none of this quietly drifts back, that one is a bonus, this PR stands on its own.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,9 +24,9 @@ jobs:
       matrix:
         java: [11, 25]
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6.1.0
       - name: Set up JDK ${{ matrix.java }}
-        uses: actions/setup-java@v5
+        uses: actions/setup-java@03ad4de0992f5dab5e18fcb136590ce7c4a0ac95 # v5.6.0
         with:
           distribution: 'temurin'
           java-version: ${{ matrix.java }}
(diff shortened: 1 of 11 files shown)


melloware:
@Totara-thib  doesn't this make GitHub Actions Dependabot upgrade not work?

Totara-thib:
> @Totara-thib doesn't this make GitHub Actions Dependabot upgrade not work?

No Dependabot keeps opening its bump PRs exactly as before, it understands sha pins with a version comment.

melloware:
awesome this seems great!

Totara-thib:
> awesome this seems great!

Thanks! Happy to help and contribute 🙌​

PR #15097, opened 2026-07-31
Showcase: fix highlighting of menu items using exact path match

NicolaIsotta:
Before:
image
After:
image


melloware:
Thank you @NicolaIsotta will wait for @tandraschko to review

tandraschko:
are you sure this works for components with subviews like DataTable and different sub-usecases?

NicolaIsotta:
From my test, yes:
image

PR #15098, opened 2026-08-02
fix #15036: inputs in overlay TieredMenu facets lose their value on AJAX

### LINKED ISSUE #15036: TieredMenu: add appendTo / inputs in facet 'end' not processed
### Describe the bug

Having a _p:tieredMenu_ with custom input components inside the facet 'end' (guess 'start' aswell) will not process the components when they are activated using ajax.

Maybe this is related to the overlay of menue components like _p:menue_ being appended to the html body and not being in the form anymore. 

### Reproducer

**Bean**

```
TestBean
{
	String value;
}
```

**Markup**

```
		<h:outputText id="lblValue" style="display: block;"
			value="value: #{testBean.value eq null ? 'null' : testBean.value}"/>
			
			
		<h:form id="form1">
			<h4>Form 1</h4>
			<p:inputText id="value" value="#{testBean.value}" style="width: 8em;">
				<p:ajax process="@this" update="@this, :lblValue"/>
			</p:inputText>
		</h:form>
	
		<h:form id="formMenu">
		
			<h4>Form Menu</h4>
		
			<p:link id="linkMenu" value="User"/>
		
			<p:tieredMenu overlay="true" my="left top" at="left bottom" trigger="linkMenu">

				<p:menuitem value="Item 1" action="#{testBean.setValue('Item1')}"
					process="@this" update="value2, :lblValue"/>
				
				
				<f:facet name="end">
				
					<p:inputText id="value2" value="#{testBean.value}" style="width: 8em;">
						<p:ajax process="@form" update="value2, :lblValue"/>
					</p:inputText>
					
				</f:facet>
				
			</p:tieredMenu>
		
		
		
		</h:form>
```

The input component _value2_ will not submit the value in the ajax http request.
In case its replaced with a _selectOneButton_ i tested it will submit _null_ instead.

Interestingly the click on `p:menuitem` works.

*Workaround*

Using a normal `p:overlay` instead without menu functionality though. This is not appended to body automatically.

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

15.0.15

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0

### Java version

17

### Browser(s)

_No response_

melloware:
Fix #15036

Overlay-mode menus (TieredMenu, Menubar, MegaMenu, PlainMenu, SlideMenu) relocate their whole markup to the document body via jQuery appendTo(), detaching any custom component placed in a start/end/options facet from its enclosing <form>. Client-side AJAX resolves the form to submit, and the @form search expression, purely via closest('form'), so a detached facet input's value silently dropped out of the request.

Stamp a form="..." attribute on facet form controls before the overlay is moved, and prefer the native element.form association (which honors that attribute regardless of DOM position) over closest('form') when resolving the form.

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/tieredmenu/TieredMenu001.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/tieredmenu/TieredMenu001.java
@@ -1,0 +1,42 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2026 PrimeFaces
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.integrationtests.tieredmenu;
+
+import java.io.Serializable;
+
+import jakarta.faces.view.ViewScoped;
+import jakarta.inject.Named;
+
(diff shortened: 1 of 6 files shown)


tandraschko:
not sure if we should fix it like this

just look at e.g. dialog

if it has appendTo=@body, we move it to the body and the parent form is lost; so people need to add a form inside the dialog.

this would behave completely different then

melloware:
Agreed I didn't love this solution myself. I thought the menus were already doing append to body?

tandraschko:
sure, the default might just be appendTo=@body

so either making appendTo=@form
or

```
			<p:tieredMenu overlay="true" my="left top" at="left bottom" trigger="linkMenu">

				<p:menuitem value="Item 1" action="#{testBean.setValue('Item1')}"
					process="@this" update="value2, :lblValue"/>
				
				
				<f:facet name="end">
				       <h:form>
					<p:inputText id="value2" value="#{testBean.value}" style="width: 8em;">
						<p:ajax process="@form" update="value2, :lblValue"/>
					</p:inputText>
				       </h:form>					
				</f:facet>
				
			</p:tieredMenu>
```

should fix it

melloware:
awesome let me update my Integration Test to use `@form` and remove the code!

melloware:
Ahhh I think part of the problem is TierMenu is missing the appendTo. I will investigate this.

tandraschko:
and @body is default? 
still adding a form to the facet should work.

melloware:
hmmm but we can't have a form inside a form?

so this won't work as the outer elements need a form to submit.  I think we would have to add the appendToForm on the TieredMenu?

tandraschko:
no, same for dialog

body -> form -> dialog (with appendTo=body) -> form

works fine on client side, just the server think its nested forms

tandraschko:
follow up: https://github.com/primefaces/primefaces/issues/15144

tandraschko:
whats our plan with this PR?

melloware:
I can close it it was proving the bug was legit.

PR #15119, opened 2026-08-16
ITs: stabilize FileUpload000Test and FileUpload002Test

christophs78:
FileUpload000Test.basicSingleUploadTwice:84->AbstractFileUploadTest.assertUploadedFiles:61 » NoSuchElement no such element: Unable to locate element: {"method":"css selector","selector":"#form\:uploadedfiles"}

 FileUpload002Test.basicMultipleUploadMultipleFilesTwice:125->AbstractFileUploadTest.wait4File:102 » StaleElementReference stale element reference: stale element not found

### DIFF
--- a/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/fileupload/FileUpload002Test.java
+++ b/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/fileupload/FileUpload002Test.java
@@ -24,6 +24,7 @@
 package org.primefaces.integrationtests.fileupload;
 
 import org.primefaces.selenium.AbstractPrimePage;
+import org.primefaces.selenium.PrimeSelenium;
 import org.primefaces.selenium.component.CommandButton;
 import org.primefaces.selenium.component.DataTable;
 import org.primefaces.selenium.component.FileUpload;
@@ -35,6 +36,7 @@ import org.json.JSONObject;
 import org.junit.jupiter.api.Order;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
+import org.openqa.selenium.By;
 import org.openqa.selenium.support.FindBy;
 
 import static org.junit.jupiter.api.Assertions.*;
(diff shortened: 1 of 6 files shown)


primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/fileupload/FileUpload002Test.java
```diff
@@ -107,11 +110,11 @@ void basicMultipleUploadMultipleFilesTwice(Page page) throws Exception {
         assertTrue(fileUpload.getFilename().startsWith(file1.getName()), fileUpload.getFilename());
         assertTrue(fileUpload.getFilename().matches(".*\\+\\s*1$"));
         page.button.click();
-        wait4File(page.uploadedFiles, 1, file1.getName());
+        wait4File(getUploadedFiles(), 1, file1.getName());
```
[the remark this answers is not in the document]
christophs78: Won´t change this (for now) because it would affect a significant amount of FileUpload-Tests.

primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/fileupload/FileUpload002Test.java
```diff
@@ -168,18 +170,22 @@ void basicMultipleUploadSizeLimit(Page page) {
...
         assertEquals("Invalid file size.",
                 page.messages.getMessage(0).getSummary());
         assertNoJavascriptErrors();
         // Primefaces sends "empty" request if mode=simple skinSimple=true
-        assertUploadedFiles(page.uploadedFiles);
+        assertUploadedFiles(getUploadedFiles());
         assertConfiguration(fileUpload);
     }
 
+    private static DataTable getUploadedFiles() {
```
tandraschko: Doesnt it work with @FindBy?

christophs78: The test is flaky. We sometimes get
`FileUpload002Test.basicMultipleUploadMultipleFilesTwice:125->AbstractFileUploadTest.wait4File:102 » StaleElementReference stale element reference: stale element not found`

Not totally sure, but some hope, this may improve stability.

tandraschko: Actually our FindBy should only inject a proxy and should not make a difference AFAIR
Do we have this hack already used in other cases?

christophs78: We do similar things at e.g.
https://github.com/primefaces/primefaces/blob/v16.0.0-RC2/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datatable/DataTable001Test.java#L308

How does the proxy work? Does he resolve dynamic for each access or e.g. only the first time?

Maybe we simply need a (better) wait condition. Hm. I tried to improve this already as part of this PR for non-ajax-usecases. Hm. Or should be look into `Guard#http` and recheck this implementation?

tandraschko: AFAIR it resolves the element again on each access
would you like to document it in the readme of the selenium project? and maybe remove all those manual createFragment as workaround.

christophs78: From my side this PR´s ready2merge.
Further work comes in seperate PR´s.
`PrimeSelenium.createFragment` is currently also used for IT´s which have more than one XHTML they test. (DataTable001Test is such a case.) Will take some more time to sort this out or maybe a few good prompts for the coding agent.

melloware (approved):

PR #15120, opened 2026-08-17
#9254: Fix TreeTable rendering issue after clearFilters()

### LINKED ISSUE #9254: Treetable: clearFilters() API make table breaks table when filtering is enabled

jxmai:
Resolve: #9254 instead of workaround

clearFilters() leaves stale filterBy entries with null values, so the `isEmpty()` check was skipped and filtering ran against a null root. The empty filtered value then shadowed the model on the next render.

After the resolution:

Filter:

image

Clear Filter:

image

Switch Node:

image

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/treetable/feature/FilterFeature.java
+++ b/primefaces/src/main/java/org/primefaces/component/treetable/feature/FilterFeature.java
@@ -97,9 +97,16 @@ public class FilterFeature implements TreeTableFeature {
 
     public void filter(FacesContext context, TreeTable tt, TreeNode root) {
         Map<String, FilterMeta> filterBy = tt.getFilterByAsMap();
-        // After clearFilters(), stale filterBy entries remain with null values.
-        // isEmpty() alone can't detect this, so we also check for active filters;
-        // without this, we'd filter against a null root and shadow the model value.
+        // After clearFilters(), stale filterBy entries remain with null values,
+        // so isEmpty() alone can't detect this case. Without the check below,
+        // filter() would run against the null root set by decode() and leave
+        // the table empty. Also restore the model value via the value expression
+        // so subsequent renders don't see the nulled local value.
+        //
+        // decode() calls setValue(null) to reset filter state before re-applying
+        // any filter values from the request. When there are no active filters,
+        // the original early return never restored the value, leaving that null
+        // in place to shadow the value expression on the next render.
         boolean noActiveFilters = filterBy.isEmpty() || filterBy.values().stream().noneMatch(FilterMeta::isActive);
         if (noActiveFilters) {
             tt.updateFilteredValue(context, null);
--- a/primefaces/src/main/java/org/primefaces/component/treetable/feature/FilterFeature.java
+++ b/primefaces/src/main/java/org/primefaces/component/treetable/feature/FilterFeature.java
@@ -110,7 +110,7 @@ public class FilterFeature implements TreeTableFeature {
         boolean noActiveFilters = filterBy.isEmpty() || filterBy.values().stream().noneMatch(FilterMeta::isActive);
         if (noActiveFilters) {
             tt.updateFilteredValue(context, null);
-            ValueExpression ve = tt.getValueExpression("value");
+            ValueExpression ve = tt.getValueExpression(TreeTable.PropertyKeys.value.name());
             if (ve != null) {
                 TreeNode<?> originalValue = (TreeNode<?>) ve.getValue(context.getELContext());
                 tt.setValue(originalValue);
--- a/primefaces/src/main/java/org/primefaces/component/treetable/feature/FilterFeature.java
+++ b/primefaces/src/main/java/org/primefaces/component/treetable/feature/FilterFeature.java
@@ -46,6 +46,7 @@ import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import jakarta.el.ELContext;
+import jakarta.el.ValueExpression;
 import jakarta.faces.FacesException;
 import jakarta.faces.context.FacesContext;
 import jakarta.faces.event.PhaseId;
@@ -96,7 +97,17 @@ public class FilterFeature implements TreeTableFeature {
 
     public void filter(FacesContext context, TreeTable tt, TreeNode root) {
         Map<String, FilterMeta> filterBy = tt.getFilterByAsMap();
-        if (filterBy.isEmpty()) {
+        boolean noActiveFilters = filterBy.isEmpty() || filterBy.values().stream().noneMatch(FilterMeta::isActive);
+        if (noActiveFilters) {
+            tt.updateFilteredValue(context, null);
+            ValueExpression ve = tt.getValueExpression("value");
+            if (ve != null) {
+                TreeNode<?> originalValue = (TreeNode<?>) ve.getValue(context.getELContext());
+                tt.setValue(originalValue);
+                if (originalValue != null) {
+                    tt.setRowKey(originalValue, null);
+                }
+            }
             return;
         }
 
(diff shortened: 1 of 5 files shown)


primefaces/src/main/java/org/primefaces/component/treetable/feature/FilterFeature.java
```diff
@@ -96,7 +97,17 @@ public void encode(FacesContext context, TreeTableRenderer renderer, TreeTable t
 
     public void filter(FacesContext context, TreeTable tt, TreeNode root) {
         Map<String, FilterMeta> filterBy = tt.getFilterByAsMap();
-        if (filterBy.isEmpty()) {
+        boolean noActiveFilters = filterBy.isEmpty() || filterBy.values().stream().noneMatch(FilterMeta::isActive);
```
tandraschko, line 110: please add 1-2 comments here

jxmai: Comment added. https://github.com/primefaces/primefaces/pull/15120/commits/d5401d1ffcc5f80276c84f582c0b48863b988297

jxmai: GitHub is having issue right now which could affect the build BTW.

https://www.githubstatus.com/

> Update - We are experiencing high error rates around 20% for web experiences and api traffic. Archive downloads and raw repository content downloads are experiencing an approximate 50% error rate. SAML and OIDC authentication, SCIM, and Team Sync are also impacted. We are currently performing mitigations and will post updates as we progress.
Aug 17, 2026 - 15:42 UTC

primefaces/src/main/java/org/primefaces/component/treetable/feature/FilterFeature.java
```diff
@@ -96,7 +97,20 @@ public void encode(FacesContext context, TreeTableRenderer renderer, TreeTable t
...
     public void filter(FacesContext context, TreeTable tt, TreeNode root) {
         Map<String, FilterMeta> filterBy = tt.getFilterByAsMap();
-        if (filterBy.isEmpty()) {
+        // After clearFilters(), stale filterBy entries remain with null values.
+        // isEmpty() alone can't detect this, so we also check for active filters;
+        // without this, we'd filter against a null root and shadow the model value.
+        boolean noActiveFilters = filterBy.isEmpty() || filterBy.values().stream().noneMatch(FilterMeta::isActive);
+        if (noActiveFilters) {
+            tt.updateFilteredValue(context, null);
+            ValueExpression ve = tt.getValueExpression("value");
```
tandraschko: Cant you use propetykeys here instead of the string?

primefaces/src/main/java/org/primefaces/component/treetable/feature/FilterFeature.java
tandraschko: The whole #setValue call is new but this comment doesnt explain this?

jxmai: I will add additional details to the comment to clarify what happened here, since this defect resulted from a chain of events. That way, we can stay aligned. It took some time to understand and untangle what exactly occurred.

1. Filtered applied: It clears any local value and set the active filters, and it stamps the local value with the filtered value.

https://github.com/primefaces/primefaces/blob/ee5846a5eb81532da0927ddaa361a245d51219ae/primefaces/src/main/java/org/primefaces/component/treetable/feature/FilterFeature.java#L55-L65

2. clearFilters(): JS calls filter() AJAX, which then invoke FilterFeature.decode again, and it sets the value to null. But filterBy map still contains stale entries, just with filterValue = null

That's the crucial part: Original early return check: `if (filterBy.isEmpty()) { return; }` actually return False since the map contains entry, just inactive.

https://github.com/primefaces/primefaces/blob/ee5846a5eb81532da0927ddaa361a245d51219ae/primefaces/src/main/java/org/primefaces/component/treetable/feature/FilterFeature.java#L97-L101

3. Filter then proceed with null root: 

`collectFilteredRowKeys(context, tt, root=null, ...)`, this part produce empty or null tree

https://github.com/primefaces/primefaces/blob/ee5846a5eb81532da0927ddaa361a245d51219ae/primefaces/src/main/java/org/primefaces/component/treetable/feature/FilterFeature.java#L108

4. No value restoration, `if (filterBy.isEmpty()) return;` never executes as the map contains stale entries, so the model value is never restored. Local value stays null, shadowing the selectedRoot on the next render.

---

In summary, clearFilters() method cannot clear the server-side filterBy map, it only resets filter input fields, leaving stale FilterMeta entries with null values. The original filterBy.isEmpty() check fails to catch this state. decode() calls setValue(null) but filter() returns early without restoring the model value when the map still contains stale entries.

The fix adds an isActive() check to properly detect when no filters are active, and restores the model value via getValueExpression before returning.

jxmai: added more comments: https://github.com/primefaces/primefaces/pull/15120/commits/94caecdcd6476e04ef1d5b460b27c325ad366147

primefaces/src/main/java/org/primefaces/component/treetable/feature/FilterFeature.java
jxmai: Good catch, Yes, definitely: https://github.com/primefaces/primefaces/pull/15120/changes/d38321248676df5cc9404c23f47fc500268beb9b

primefaces/src/main/java/org/primefaces/component/treetable/feature/FilterFeature.java
```diff
@@ -96,7 +97,27 @@ public void encode(FacesContext context, TreeTableRenderer renderer, TreeTable t
...
+        // so subsequent renders don't see the nulled local value.
+        //
+        // decode() calls setValue(null) to reset filter state before re-applying
+        // any filter values from the request. When there are no active filters,
+        // the original early return never restored the value, leaving that null
+        // in place to shadow the value expression on the next render.
+        boolean noActiveFilters = filterBy.isEmpty() || filterBy.values().stream().noneMatch(FilterMeta::isActive);
+        if (noActiveFilters) {
+            tt.updateFilteredValue(context, null);
+            ValueExpression ve = tt.getValueExpression(TreeTable.PropertyKeys.value.name());
```
tandraschko: I think you remove .Name()
We overload the Method in PrimeComponent to allow to pass enums

jxmai: Updated: https://github.com/primefaces/primefaces/pull/15120/changes/aa72431d73935e12472b4b9c19304c8236c9b8ea

BTW, I saw some unrelated errors from integration tests from other components. Please re-trigger CI if needed.

https://github.com/primefaces/primefaces/actions/runs/32146604565/job/95741756353?pr=15120

> [ERROR] Failures: 
[ERROR]   DataExporter002Test.csv:64->AbstractDataExporterTest.assertFileEquals:94 Expected CSV files to match but they did not! ==> expected: `<true>` but was: `<false>`

tandraschko:
Looks good now
Can we make a Integration test for it?

jxmai:
> Looks good now Can we make a Integration test for it?

Sounds great. Integrate test created https://github.com/primefaces/primefaces/pull/15120/commits/9ac11df32dcc12748af2e70d9998c5bbce2481d0, and  confirm that the newly created integration test wouldn't pass without the resolution.

PR #15122, opened 2026-08-17
Fix #15114: AccordionPanel inactive state support/improve behavior

### LINKED ISSUE #15114: AccordionPanel: render panel with no tabs open is noticeably slower
**Current Behavior**

In PrimeFaces 16, activeIndex was deprecated in favor of active, which supports both integer indices and p:tab key values. To show a accordionPanel with no tabs open i used to do:

`<p:accordionPanel dynamic="true" activeIndex="-1" cache="true">`

The replacement for primefaces 16 would be:

`<p:accordionPanel dynamic="true" active="-1" cache="true">`

**Identified Issue**

In practice, setting activeIndex="-1"/active="-1" (to render the panel with no tabs open by default) is noticeably slower in PrimeFaces 16, especially with dynamic/var-driven tabs where key is bound to an EL expression.

PrimeFaces 16 evaluates every tab's key even when the active value can never match anything.

**Reproducer**

See: 

[pf-test-accordion2.zip](https://github.com/user-attachments/files/31082975/pf-test-accordion2.zip)

It s difficult to measure/prove the result.
After a lot of page refreshes with the test project, i had as fastest (but most of time higher).

15.0.17: 
[AccordionBench] render=12ms, keyEvaluations=0
[AccordionBench] / total request time: 14ms

16.0.0-RC2: fastest (but most of time higher):
[AccordionBench] render=24ms, keyEvaluations=9000
[AccordionBench] / total request time: 25ms

**Proposed solution**

Make active "-1" and "none" a properly documented, intentional opposite of "all" rather than never match a real key/index.

See doc:
```
| activeIndex    | 0       | String     | (DEPRECATED, use active instead!) Index of the active tab or a comma separated string of indexes when multiple mode is on. Alternative: all.
| active         | 0       | String     | Index or key of the active tab or a comma separated string of indexes/keys when multiple mode is on. Alternative: all.
```

### PrimeFaces edition

Community

### PrimeFaces version

16.0.0-RC2

### JSF implementation

Mojarra

### JSF version

mojarra40

### Java version

21

### Browser(s)

Chrome latest

DazzlingBob:
- Added server/client support for active="-1" and active="none" (and legacy activeIndex="-1"), skipping tab key evaluations when no tab is active.
- Fixed empty dynamic tab rendering: Prevented JS from falsely marking Tab 0 as cached/loaded after a postback when all tabs are closed.
- Index state cleanup: Ensured setIndex(-1) is called in a finally block to properly reset the index after resolving active keys in dynamic loops.

### DIFF
--- a/docs/16_0_0/components/accordionpanel.md
+++ b/docs/16_0_0/components/accordionpanel.md
@@ -172,3 +172,11 @@ the list of structural style classes;
 | .ui-accordion | Main container |
 | .ui-accordion-header | Tab header |
 | .ui-accordion-content | Tab content |
+
+## Common Mistakes
+
+### Collapsing all tabs
+
+Use `active="none"` or `active="-1"` to render the panel with all tabs collapsed. The legacy
+`activeIndex="-1"` value is also supported. Using this allows the renderer to skip
+evaluating tab keys, avoiding unnecessary evaluation when no tab is active.
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/accordionpanel/AccordionPanel006.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/accordionpanel/AccordionPanel006.java
@@ -1,0 +1,58 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2026 PrimeFaces
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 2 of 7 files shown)


docs/16_0_0/components/accordionpanel.md
```diff
@@ -171,4 +171,12 @@ the list of structural style classes;
...
 | .ui-accordion-header | Tab header |
-| .ui-accordion-content | Tab content |
\ No newline at end of file
+| .ui-accordion-content | Tab content |
+
+## Common Mistakes
+
+### Collapsing all tabs
+
+Use `active="none"` or `active="-1"` to render the panel with all tabs collapsed. The legacy
```
melloware, line 180: Please update `Accordion` property for `active` so the Taglib is automatically updated with this information

melloware (changes requested):

docs/16_0_0/components/accordionpanel.md
DazzlingBob: Updated AccordionPanelBase.java. Is that sufficient?

melloware (approved):

melloware:
yep that is great

PR #15130, opened 2026-08-19
Fix #15128: CommmandLink renders a default aria-label value even when…

### LINKED ISSUE #15128: CommandLink: generates default documentation text as aria-label
### Describe the bug

**Current Behavior**

When using p:commandLink without an explicit aria-label attribute, PrimeFaces renders a static documentation string as the aria-label value:

```
<a id="..." href="#" class="ui-commandlink ui-widget" aria-label="The aria-label attribute is used to define a string that labels the current element for accessibility.">
    Ajax Post
</a>
```
This causes accessibility issues and unintended bloat across elements like <p:dataTable> columns that rely on default link behavior.

**Expected Behavior**

CommandLink should not output documentation text inside the aria-label attribute.

**Reproducer**

See showcase PrimeFaces 16: https://primefaces-showcase-zvbjj.ondigitalocean.app/ui/button/commandLink
For example the "Ajax Post" link.

**Possible Solutions / Discussion**

If aria-label is not explicitly set:
- Fall back to the value property if present (e.g., aria-label="Ajax Post").
- Omit the aria-label attribute entirely so screen readers read the link's text content naturally.

### PrimeFaces version

16.0.0-RC2

### JSF implementation

Mojarra

### JSF version

mojarra40

### Java version

21

### Browser(s)

Chrome latest

DazzlingBob:
… the developer never set one

See #15128 

Changed:
- moved documentation text from defaultValue to description
- only write the aria-label attribute when component.getAriaLabel() is not blank

### DIFF
--- a/primefaces/src/main/java/org/primefaces/component/commandlink/CommandLinkRenderer.java
+++ b/primefaces/src/main/java/org/primefaces/component/commandlink/CommandLinkRenderer.java
@@ -102,8 +102,9 @@ public class CommandLinkRenderer extends CoreRenderer<CommandLink> {
         writer.writeAttribute("href", "#", null);
         writer.writeAttribute("class", styleClass, null);
 
-        if (!isValueBlank(component.getAriaLabel())) {
-            writer.writeAttribute(HTML.ARIA_LABEL, component.getAriaLabel(), null);
+        String ariaLabel = component.getAriaLabel();
+        if (!isValueBlank(ariaLabel)) {
+            writer.writeAttribute(HTML.ARIA_LABEL, ariaLabel, null);
         }
 
         if (!isValueBlank(form)) {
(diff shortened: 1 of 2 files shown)


primefaces/src/main/java/org/primefaces/component/commandlink/CommandLinkRenderer.java
```diff
@@ -101,7 +101,10 @@ protected void encodeMarkup(FacesContext context, CommandLink component) throws
         writer.writeAttribute("id", clientId, "id");
         writer.writeAttribute("href", "#", null);
         writer.writeAttribute("class", styleClass, null);
-        writer.writeAttribute(HTML.ARIA_LABEL, component.getAriaLabel(), null);
+
+        if (!isValueBlank(component.getAriaLabel())) {
```
tandraschko: please make `String ariaLabel = ...` and reuse

DazzlingBob: Agreed and done!

tandraschko:
thanks!

PR #15140, opened 2026-08-21
Add support for jakarta.* namespace to PrimeFaces 8

Axel-1:
Adds jakarta.* namespace support to PrimeFaces 8 by backporting the maven-shade-plugin setup from PrimeFaces 10.

Smoke-tested on JSF 3 / Tomcat 11, some components may still break.


tandraschko:
-1 from my side, its in maintenance mode

melloware:
yeah you really should be looking to upgrade your PrimeFaces if you are moving to Faces 4.0 anyway.  Staying on 8 makes no sense.

Issue #7427, opened 2021-06-15
DataTable: Numeric filters with option to select a comparator (=, !=, <, >, <=, >=)


tuerker:
It would be really great if out-of-box numeric datatable filters would be supported such that you can enter a filter value together with the dynamic option to select a comparator (=, !=, <, >, <=, >=). This would massively enhance the usefulness of the table filters.

melloware:
I think you are looking for something similar to how PrimeVue does it. See the example here: https://primefaces.org/primevue/showcase/#/datatable/filter  and click on the "Balance" column filter icon.

tuerker:
yes, functionality-wise it is exactly that.

i would just prefer a different DEFAULT layout where you have the input field followed on the right end by a SelectOneMenu for the comparator selection. this would have the advantage that you can easily overview your entire filter criteria.

tuerker:
actually, this is nearly the second variant, the Filter Row approach, of primevue. would be great if this could be available for primefaces, too. the primevue extension looks really very powerful and generic. props to the developers.

melloware:
Up to PrimeTek or Sponsored by a PRO account. I doubt any of the open source volunteers will commit to this.

eduarddrenth:
For my pf (jpa) apps I use a library that allows users to use some syntax in values.

Based on https://bitbucket.org/fryske-akademy/crudapi/src/master/jpaservices/src/main/java/org/fryske_akademy/jpa/OPERATOR.java

see https://bitbucket.org/fryske-akademy/crudapi/src/master/

Perhaps this is of use

jepsar:
We could create a default filterFunction in PFE or PF to support this. That would be easier than adding it to the filter feature.

jepsar:
.. I must say, that Vue filter demo is awesome

eduarddrenth:
I'll have a look, perhaps I'll even refactor to stick closer to PF, or develop an extension sometime

melloware:
Yep PrimeTek has updated the 3 JS libraries Datatable for that Advanced filtering its pretty slick.

eduarddrenth:
I assume these filtering features aren't passed on into the jpa data layer and executed there?

eduarddrenth:
I have the feeling I did not explain clear enough what my solution offers, it has been very valuable for years now for us.

**jpa module**
- building jpql queries, supports all possible jpql comparisons
[...]

PR #15159, opened 2026-09-01
Add match mode selection to datatable column filters

tuerker:
A while ago, I raised this issue: https://github.com/primefaces/primefaces/issues/7427

Since then, I’ve implemented a more advanced solution that supports different data types and a wide range of match modes.

I’d be happy to share the implementation. It would be great if someone from the team could review it and, if appropriate, consider merging it into master.

image


tuerker:
sorry. include the wrong branch. created a new pull request.

PR #15160, opened 2026-09-01
Add match mode selection to datatable column filters

tuerker:
A while ago, I raised this issue: https://github.com/primefaces/primefaces/issues/7427

Since then, I’ve implemented a more advanced solution that supports different data types and a wide range of match modes.

I’d be happy to share the implementation. It would be great if someone from the team could review it and, if appropriate, consider merging it into master.

image


melloware:
I have assigned Thomas to review but @christophs78 and @Rapster would also like your opinion?

tuerker:
In the meantime, I have also fixed the vertical alignment issue of table columns (compare previous snapshot above with the below):

image

tandraschko:
Will review in october, will be in holiday soon
It will target 17.0 anyway

tuerker:
> I have assigned Thomas to review but @christophs78 and @Rapster would also like your opinion?

great. from what i have seen so far, they would be very well-fitting candidates, too. many thanks in advance.

christophs78:
First of all: Awesome to see you building this advanced feature and sharing it with the community!

Just having a first look at the diff.
Is this feature always present or does it need to be activated by the developer? (eg per column and/or per datatable)
One perspective is to give the enduser as much control as possible, the other one to have a clean, reduced UI and only offer as much features as needed. IMO both perspectives have their place.

tuerker:
> First of all: Awesome to see you building this advanced feature and sharing it with the community!
> 
> Just having a first look at the diff. Is this feature always present or does it need to be activated by the developer? (eg per column and/or per datatable) One perspective is to give the enduser as much control as possible, the other one to have a clean, reduced UI and only offer as much features as needed. IMO both perspectives have their place.

The match-mode picker is opt-out, per column. The "Clear Filters" button is opt-in, per table. There is currently no table-level switch for the picker — and I think that's the fair point in your question. 

  1. Match-mode dropdown — on by default, per column

  The control point is the new filterValueType attribute on <p:column> / <p:columns>:

-  not set → the type is auto-derived from the column's actual Java type (FilterMeta.resolveColumnJavaType, reflection over the table's row type + the field path), and the column gets the curated preset for that type (numeric, text, date, time/datetime, boolean, enum, array).
-  filterValueType="none" → no dropdown at all; plain filter input fixed to filterMatchMode, exactly as before.
-  a preset keyword or an explicit list (filterValueType="equals,notEquals,gte") → precisely that set.

  So a developer opts a column out, not in.

  2. …but auto-derivation is deliberately conservative, so it's less "always on"

  FilterMeta.resolveFilterValueType returns null (no dropdown) when:

-  the column has a custom <f:facet name="filter"> or a filterFunction — those have already opted out of match-mode-driven filtering;
-  the derived type is numeric/date/time/datetime but the column has no converter — those comparators need the string converted before comparing, so the preset is withheld rather than throwing at filter time;
-  the fallback to text would silently change an explicitly configured filterMatchMode that text doesn't offer (e.g. gte).

  Net effect on an existing app: plain String columns do pick up the picker on upgrade; unconverted numeric/date columns and facet/function columns do not.

  3. Clear Filters button — off by default, per table

  DataTableBase.clearFiltersButton defaults to false; you must ask for it. That one already follows the "reduced UI" side.

I will add the option filterValueType="none" on the table to disable the whole feature in one place.

tuerker:
Added option filterValueType="none" on the table to disable the whole feature in one place while still having the option explicitly switch-on column-wise.

melloware:
@tuerker also can you look at JpaLazyDatatable i think every one of your new filters will be needed to be added there for Criteria Query. cc @FlipWarthog

cantuerker:
> @tuerker also can you look at JpaLazyDatatable i think every one of your new filters will be needed to be added there for Criteria Query. cc @FlipWarthog
@melloware thanks for the hint. will do that.

christophs78:
Can you please be transparent which Coding Agent you are using?

Known downside of such large PR´s is it´s almost impossible to review them manually. Even when skipping Showcase and IT and only looking into PF-Core-diff. Maybe you can give a short insight how your organization handles this on your day-job. 
On my day-job we have no final solution for this. Currently part of our focus is manual testing of vibe coded features and to some degree skip code-reviews.
Whichever conclusion we should come up as OSS project.

Can you please be also transparent to which degree you tried to look over e.g. the generated IT´s. 

One thing i checked are the shortcuts for filterValueType like `numeric`. This makes sense and answers one question which raised in my mind today.

tuerker:
> Can you please be transparent which Coding Agent you are using?

I used Claude Code.

tuerker:
> Known downside of such large PR´s is it´s almost impossible to review them manually. Even when skipping Showcase and IT and only looking into PF-Core-diff. Maybe you can give a short insight how your organization handles this on your day-job. On my day-job we have no final solution for this. Currently part of our focus is manual testing of vibe coded features and to some degree skip code-reviews. Whichever conclusion we should come up as OSS project.
> 
> Can you please be also transparent to which degree you tried to look over e.g. the generated IT´s.
> 
> One thing i checked are the shortcuts for filterValueType like `numeric`. This makes sense and answers one question which raised in my mind today.

That's a fair concern. This is indeed a large PR, and I fully understand that reviewing a change of this size thoroughly is time-consuming. The size is primarily driven by the scope of the feature itself. The feature touches quite a few areas — PF-Core, the different data models, filtering, integration tests, Showcase, etc.

Regarding how we handle this at my day-job: we are a very small development team of at most three people, so our process is probably not directly transferable to an OSS project. For larger feature requests, the development team first tests the implementation manually, in addition to the automated tests and code review. Once we consider it ready, the people who requested the feature also test it manually from their perspective. We don't have a specific AI/vibe-coding process in place, so this is essentially the same process we follow regardless of how the code was developed.

For transparency, I developed the feature with the help of Claude Code. There were several iterations and revisions during the development process, and I personally worked through those iterations to ensure that the implementation behaves as intended. I reviewed the implementation, ran the tests, manually exercised the relevant functionality, and iterated further based on the results and review feedback.

The ITs contain a significant amount of generated code. Of course, I reviewed them as well. But I don't want to claim that I manually scrutinized every generated test line with the same depth as the PF-Core code.

Ultimately, I think the important thing for the review is the resulting code, its behavior, and the test coverage. If there are specific areas where a deeper review would be helpful, I'm happy to provide more context. I don't think the fact that an agent was involved should change the standard by which the contribution itself is evaluated.

And thanks for checking the numeric shortcut — that's exactly the sort of API-level sanity check that I think is particularly valuable from the maintainers.

christophs78:
Thank´s for taking the time to share the process of creating this PR, sharing your experience and thoughts!

On my day job (~60 devs distributed over about 10 agile teams working on many individual software projects) we do currently similar things. Very disruptive time where almost everyone tries to figure out how to arrage with the new realtity. Finding the right mix between stability and controll on one side and much (?) increased productivity on the other side.
A year ago without coding agents a feature like this would probably been out of scope.

Did some manual testing of your PR....
I like most things you did! Some detail feedback:

- Switching the filter mode collides sometime with the sorting feature. You have to precily click on the filter-symbol. (no show-stopper, just a bit tricky)
- Depending on the screensize some datatypes have extensive filter match modes. Especially applicable for date.
image

Looked how similar features are realized in other UI-libraries:

- https://primeng.dev/table#filter-advanced
- https://js.devexpress.com/Angular/Demos/WidgetsGallery/Demo/DataGrid/Filtering/FluentBlueLight/

Maybe input for some future PR, whoever find´s time to make additional UI improvements.

The ITs provided as part of this PR look high level good. Maybe better than most we wrote manually. So i would take them as is. Thanks for reviewing them yourself.

Feedback handwritten without AI-based review of this feedback-text. ;-) Please excuse some imperfections.

primefaces/src/main/java/org/primefaces/model/MatchMode.java
```diff
@@ -26,50 +26,315 @@
...
+        switch (this) {
+            case BETWEEN:
+            case NOT_BETWEEN:
+                return "min,max";
+            case IN:
+            case NOT_IN:
+            case CONTAINS_ANY:
+            case CONTAINS_ALL:
+            case CONTAINS_NONE:
+                return "value1, value2, ...";
```
christophs78: Nice to have would be to have the placeholder-hints localized too.

melloware (changes requested):
Looks like you need to  rebase your branch from main too as you have merge conflicts

primefaces/src/main/java/org/primefaces/model/MatchMode.java
tuerker: Thanks for the hint. Will do that.

tuerker:
> Looks like you need to rebase your branch from main too as you have merge conflicts

Thanks for heads-up. Will do resolve them.

primefaces/src/main/java/org/primefaces/model/MatchMode.java
tuerker: Thanks a lot for taking the time to manually test the PR and for all the feedback and for sharing how other UI libraries approach this. Really appreciated!

Regarding the filter UI: the implementation in this PR is an extensive version of the basic PrimeVue-style filter approach, where the available match modes are exposed directly. I am aware of the more advanced variant as well, with the additional menu where multiple filter conditions can be combined using predicates such as ALL / ANY. I agree that this could be a useful additional UI option and potentially a follow-up improvement, but I intentionally kept the scope of this PR to the more basic approach.

Regarding the number of match modes, I am aware that this can become quite extensive, especially for date types. Most of the options below is not empty are essentially quick shortcuts for commonly used filter conditions. One option could therefore be to introduce a new menu type, e.g. extensive, where these additional match modes are available, while keeping the default/basic menu more compact. This could also provide a good way to address the screen-size issue without removing any of the functionality.

Thanks especially for the feedback on the ITs. Good to hear that you consider them useful enough to take them as they are.

tuerker:
## Reviewer's guide: extended DataTable filter match modes (#7427)

The branch lets the end user pick a filter **comparator** at runtime from a menu next to each column's filter input, adds ~30 new match modes behind it, and wires them through both in-memory and JPA filtering.

Against upstream master: **39 commits, 110 files, +13,758 / −230**.

### Scope at a glance

| Area | files | +added | −removed |
|---|---:|---:|---:|
| Core model — `MatchMode`, `FilterMeta` | 2 | 711 | 25 |
| Core model — filter constraints | 16 | 1124 | 1 |
| Core model — `JPALazyDataModel` | 1 | 405 | 14 |
| Component API & renderer | 8 | 670 | 36 |
| Frontend (JS/CSS) | 3 | 867 | 60 |
| Message bundles (i18n) | 27 | 1838 | 0 |
| Unit tests | 20 | 2790 | 0 |
| Integration tests | 19 | 3915 | 0 |
| Selenium framework | 2 | 186 | 1 |
| Showcase | 11 | 1235 | 93 |

**~55% of the diff is tests and demos.** The production surface is roughly 3,700 lines across 30 files.

## Review in this order

### 1. Public API and semantics — highest value per minute

**`MatchMode.java`** (+~600). Adds ~30 enum constants, each with an `operator`, `icon`, `symbol` and `requiresValue` flag; the preset lists; `parseOptions(String)`; `hasKeyword(...)`; `placeholderHintKey()`.

Worth checking: the operator strings are the wire format (they end up in a hidden input and in `data-match-mode`), so they are effectively API — `IS_NULL`'s operator is the literal string `"null"`, which forced `.attr()` over jQuery's `.data()` in the widget. And `parseOptions` is where the whole `filterValueType` contract lives.

**`FilterMeta.java`** — :warning: **breaking signature change**:

```java
// before (upstream)
public static FilterMeta of(FacesContext, String var, UIColumn, boolean normalize)
// after
public static FilterMeta of(FacesContext, UITable<?> table, UIColumn, boolean normalize)
```

Widened because match-mode resolution needs the table (for the table-level `filterValueType`). Public static method, so any external caller breaks. Also adds `resolveMatchModeOptions`, `resolveFilterValueType` and `resolveColumnJavaType` — the last reflects the column's Java type off the table's row type, which is the trickiest logic in the branch.

**The `filterValueType` contract** (documented identically on `ColumnBase`, `ColumnsBase`, `DataTableBase`): a comma list of preset keywords, the `shortcuts` token, or individual operators; `none` opts out; unset means auto-derive.

| value | modes | menu |
|---|---:|---|
| `"date"` | 10 | default |
| `"date,shortcuts"` | 28 | + all calendar shortcuts |
| `"datetime,shortcuts"` | 32 | + date and time shortcuts |
| `"date,today,thisWeek"` | 12 | hand-picked |
| `"...,none"` | 0 | opts out, wins anywhere in the list |

:warning: **The compatibility question reviewers should settle:** with `filterValueType` unset, every filterable column whose Java type can be resolved **gets a picker by default**. That changes the look and behaviour of existing applications on upgrade. Opt-out is `filterValueType="none"`, settable table-wide. Worth an explicit decision on whether default-on is right.

### 2. Filtering correctness — 14 new constraints

`primefaces/src/main/java/org/primefaces/model/filter/` — new: `AllFilterConstraint`, `ArrayContains`, `ContainsAll`, `ContainsAny`, `IsEmpty`, `IsFalse`, `IsNull`, `IsTrue`, `MatchesRegex`, `RelativeDateRange`, `RelativeMinutesOrHours`, `RelativeNDays`, plus `CollectionFilterUtils` and `DateFilterUtils`; modified: `FilterConstraints`, `InFilterConstraint`.

Focus points:

- **`RelativeMinutesOrHoursFilterConstraint`** — a bare `LocalTime` is a cyclic 24h clock, so its range wraps past midnight. The most subtle logic here.
- **`DateFilterUtils`** — week/quarter/year boundaries; week start is locale-dependent.
- **`MatchesRegexFilterConstraint`** — user-supplied regex compiled server-side; worth a look for catastrophic-backtracking exposure.
- **`CollectionFilterUtils`** — `containsAny/All/None` over `Collection` *and* array field values.

**`JPALazyDataModel.java`** (+405): the same semantics expressed as Criteria predicates. Every relative mode has to be written twice — once in-memory, once in JPA — so **divergence between the two is the main risk**. `DataTable053Test` is the guard.

### 3. Rendering and client behaviour

**`DataTableRenderer.java`** (+~350): `encodeFilterMatchModeMenu`, `encodeFilterInput`, `encodeDateFilterWidgets`, `encodeDatePickerFilterWidget`, `encodeClearFiltersButton`, `resolveFilterValueWidget`.

The **shadow DatePicker** design deserves attention: a date-ish column renders *two* hidden `DatePicker` widgets (single + range) from first paint, because a live picker cannot switch `selectionMode` afterwards and a filter round trip only re-renders the tbody. The client shows exactly one.

Note also that "is this a date column?" must ask `MatchMode.hasKeyword(...)`, never compare the attribute to `"date"` — the value is a list. Centralised in `DataTableRenderer.isDateLike(...)`.

**`datatable.widget.js`** (+~500): `bindFilterMatchModeMenu`, `toggleFilterValueInput`, `updateFilterMatchModeIconState`, `resetColumnFilter`, `clearFilters`, `bindDatePickerFilterSync`. Menus are relocated to `document.body` (so scrollable/frozen headers do not clip them) and resolved back via `aria-controls`. The menu opens on hover after a 150ms intent delay and closes 300ms after the cursor leaves both trigger and menu.

**`datatable.css`**: the match-mode icon's hit area (27×22 around a 16×16 glyph) with asymmetric padding — 8px toward the sort icon, 3px toward the active-mode badge.

**New attributes:** `clearFiltersButton` on `p:dataTable`; `filterValueType` on `p:dataTable`, `p:column`, `p:columns`.

### 4. i18n — 27 bundles

68 new keys per bundle: 65 match-mode labels, `CLEAR_FILTERS`, and 3 placeholder hints. All translated rather than left as English placeholders. `de` and `tr` were written first-hand and checked by me personally since these are languages i know well; the rest are best-effort generated by Claude Code would benefit from native review**.

Note the labels take no `MessageFormat` parameters, so apostrophes are *not* doubled — deliberately unlike the validation messages.

### 5. Tests and demos — verification, lowest risk

- **Unit**: `MatchModeTest`, `FilterMetaTest`, `FilterConstraintsTest`, `DateFilterUtilsTest`, 10 per-constraint tests, and `JPALazyDataModelMatchModeTest` (52 tests against a real in-memory H2).
- **Integration**: `DataTable052` (table-level vs column-level precedence), `DataTable053` (lazy JPA), `DataTable054` (clear actions + `p:columns`), `DataTable055` (all modes + the shadow pickers, 53 tests).
- **Selenium framework**: `HeaderCell` and `AbstractTable` gained `filterMatchMode`, `clearColumnFilter`, `getFilterMatchModeLabels`, `getColumnFilterMatchModeIcon/Value`.
- **Showcase**: `filter.xhtml` gained a reference table and two cards (presets, and the `shortcuts` opt-in), plus `lazyJpa.xhtml`.

## Two bugs the new tests caught, worth confirming as fixed

1. **`clearFilters()` no longer cleared plain filter inputs** — a picker-only reset had replaced the generic one, so columns without a picker stayed filtered. This affected *any* existing application calling `PF('t').clearFilters()`.
2. **Clearing a date column left the `DatePicker` holding its date** — the input was cleared but the widget re-displayed its own state; now reset via `setDate(null)`.

## Verification status

| Check | Result |
|---|---|
| Full reactor `mvn install` | SUCCESS (13 modules) |
| Unit tests | 705 pass |
| DataTable + TreeTable ITs | 332 pass |
| Checkstyle / license | 0 violations / 0 missing |

The full IT suite also shows 9 failures in `InputText`, `InputTextArea`, `Slider`, `Password`, `Calendar` and `Csv`. These **reproduce identically on upstream master** with none of this branch applied (verified in a clean worktree) — shifted characters dropped during keystroke delivery in the local Chrome setup — and are unrelated to this PR.

## Decisions for maintainers

1. **Picker on by default** for auto-derivable columns — accept, or flip to opt-in?
2. **`FilterMeta.of()` signature change** — acceptable break, or keep a deprecated `String var` overload?
3. **`"datetime,shortcuts"` bundles both date and time shortcuts** rather than needing a separate token.
4. Reviewer suggestion still open: a further "extensive" menu tier combining conditions with ALL/ANY predicates — deliberately out of scope here.
5. One **upstream** fix rides along in `primefaces-showcase/pom.xml` (pinning `jackson-annotations` to 2.22, which the `chartjs-java-model` 3.0.2 bump requires; without it the showcase landing page fails to render). Arguably belongs in its own PR.

primefaces/src/main/frontend/packages/components/src/datatable/datatable.widget.js
```diff
@@ -4484,6 +5023,55 @@ PrimeFaces.widget.DataTable = class DataTable extends PrimeFaces.widget.Deferred
...
+     * Resets a single column's filter match-mode picker back to the column's own configured default: clears
+     * the value input (and the shadow single/range date pickers - see
+     * DataTableRenderer#encodeDateFilterWidgets() - via their own `change` handler in bindDatePickerFilterSync(),
+     * so their internal state, not just their displayed text, stays in sync), then re-selects the default mode.
+     * Does not itself trigger a filter() round trip - callers decide when (clearFilters() resets every column
+     * first; the per-column "Clear" menu item resets just this one, right before filtering).
+     * @private
+     * @param {JQuery} icon the `.ui-column-filter-mode-icon` trigger button for the column to reset
+     */
+    resetColumnFilter(icon) {
```
FlipWarthog, line 5084: This function and its docs are inserted between an existing function (`clearFilters`) and its docs, breaking the association.

primefaces/src/main/java/org/primefaces/model/MatchMode.java
FlipWarthog: @tuerker  I think this last comment was intended for the main thread, fyi.

primefaces/src/main/frontend/packages/components/src/datatable/datatable.widget.js
tuerker: Thanks for the hint. Must have been happened while resolving conflicts after pull from master. Will fix this asap.

PR #15162, opened 2026-09-02
Fix #15161 Tree: reset rowKey after render

### LINKED ISSUE #15161: Tree: rowKey is not reset after render, leaking the iteration variable and row prefixed client ids
## Describe the bug

`TreeRenderer` leaves the tree standing on the last node it rendered. `encodeTreeNode` calls `component.setRowKey(root, rowKey)` for every node, and the render branch of `encodeEnd` never resets it. The node expand branch of the same method does reset it. `TreeTableRenderer` resets it after encoding its rows, and `UIData.encodeEnd` resets its row index to -1. So the tree is the only one which does not.

This has two effects for the rest of the request.

First, the iteration variable leaks. `setRowKey` puts `var` (and `nodeVar`) in the request map. It removes them only when the row key is set to null. So everything which renders after the tree resolves `#{var}` to the data of the last node, instead of to nothing.

Second, everything inside the tree gets a client id with the row key in it. `UITree.getContainerClientId` appends the row key the tree stands on. Faces saves the partial state at the end of RENDER_RESPONSE. It visits the tree with `SKIP_ITERATION`, and `UITree.visitTree` resets the row key only when it iterates. So it visits the children while the tree still stands on the last node, and their state is saved as `form:tree:<lastRowKey>:child`. On the postback the tree stands on no row. There the same child is `form:tree:child`, so nothing is restored into it. Input values still work, because `UITree` keeps its own row state (`PropertyKeys.saved`: value, submitted value, `localValueSet` and `valid`) under the tree's own client id. Other state does not survive: a converter or validator added programmatically, `rendered` or `styleClass` set at runtime, or the state of a nested third party component.

Mojarra 4.1.15 reports the second effect in `ProjectStage.Development`. That is how this was found: https://github.com/eclipse-ee4j/mojarra/issues/5984

## Reproducer

```xml
<h:form id="myForm">
    <p:tree value="#{treeBean.itemTree}" var="object">
        <p:treeNode id="tn" />
    </p:tree>
    <h:outputText id="after" value="after=[#{object}]" />
    <h:commandButton id="submit" value="submit">
        <f:ajax render="@form" />
    </h:commandButton>
</h:form>
```

```java
@Named
@RequestScoped
public class TreeBean {

    private TreeNode<?> itemTree;

    public TreeNode<?> getItemTree() {
        if (itemTree == null) {
            itemTree = new DefaultTreeNode<>();
            new DefaultTreeNode<>("one", itemTree);
            new DefaultTreeNode<>("two", itemTree);
            new DefaultTreeNode<>("three", itemTree);
        }
        return itemTree;
    }
}
```

The first render already shows the leaked variable. The page renders `after=[three]` where `after=[]` is expected.

Clicking the button then logs this on Mojarra 4.1.15 in `ProjectStage.Development`:

```
WARNING The view rebuilt for this postback does not hold 1 of the components the response was rendered from, so their state is restored into nothing: [myForm:j_idt5:2:tn].
```

The `:2:` is the row key of the last rendered node. With two root children it is `:1:`, with three it is `:2:`.

## Expected behavior

After the tree has rendered, it stands on no row. `#{object}` resolves to nothing, and every component inside the tree reports its client id without a row key, so its state is saved under the id which the next request rebuilds it with.

## Proposed fix

Reset the row key at the end of `encodeEnd`. That covers the render branch and the filter branch, and it changes nothing for the node expand branch, which resets it already:

```java
    @Override
    public void encodeEnd(FacesContext context, Tree component) throws IOException {
        ...
        else {
            encodeMarkup(context, component);
            encodeScript(context, component);
        }

        // A descendant's client id carries the row key the tree is left on, so it must not stay on a node.
        component.setRowKey(root, null);
    }
```

With this applied, `Tree0*Test` and `TreeTable0*Test` run green on `master` (32 tests, on Mojarra 4.1.15). On 15.0.17 the rendered markup of the reproducer is the same with and without it, apart from the leak itself.

## Proposed test

A unit test which pins the contract. It is red without the fix (`expected: <null> but was: <1>`) and green with it:

```java
package org.primefaces.component.tree;

class TreeRendererTest {

    /**
     * The row the tree stands on ends up in the client id of everything below it, and it puts the node it holds in
     * the request map under the tree's var. So a render which iterated the tree leaves it standing on no row. The
     * state of a child is saved under the client id it reports at the end of the render, and whatever renders after
     * the tree resolves the var of its own scope.
     */
    @Test
    void encodeEndLeavesTheTreeStandingOnNoRow() throws Exception {
        FacesContext context = new FacesContextMock(new CollectingResponseWriter());

        TreeNode<String> root = new DefaultTreeNode<>();
        new DefaultTreeNode<>("one", root);
        new DefaultTreeNode<>("two", root);

        UITreeNode treeNode = new UITreeNode();
        treeNode.setType("default");

        Tree tree = new Tree();
        tree.setId("tree");
        tree.setVar("node");
        tree.setValue(root);
        tree.getChildren().add(treeNode);

        new TreeRenderer().encodeEnd(context, tree);

        assertNull(tree.getRowKey());
        assertEquals(tree.getClientId(context), tree.getContainerClientId(context));
        assertFalse(context.getExternalContext().getRequestMap().containsKey("node"));
    }
}
```

It needs three maps on `ExternalContextMock`. They return `null` today, while `applicationMap` next to them is already a plain `HashMap`:

```java
    private Map<String, Object> requestMap = new HashMap<String, Object>();
    private Map<String, String> requestParameterMap = new HashMap<String, String>();
    private Map<String, String> requestHeaderMap = new HashMap<String, String>();
```

`UITree.setRowKey` writes the var into `getRequestMap`. `Tree.isNodeExpandRequest` reads `getRequestParameterMap`. `AgentUtils` reads `getRequestHeaderMap` while the state holder inputs are written. With these three returning their map, `TreeRenderer.encodeEnd` runs end to end against the mocks, so the other renderers can be tested the same way.

An integration test like `Tree008` could cover the leaked variable in a rendered page as well. It would assert that an `h:outputText` after the tree renders empty, and not the data of the last node.

## PrimeFaces edition

Community

## PrimeFaces version

15.0.17, and `master` (16.0.0-SNAPSHOT) has the same code

## Theme

Saga

## JSF implementation

All

## JSF version

4.1

## Java version

21

## Browser(s)

Not browser specific

🤖 Generated with Claude Opus 5

melloware:
Fix #15161 Tree: reset rowKey after render- #15162

TreeRenderer left the tree standing on the last rendered node. That leaked the iteration variable into whatever renders after the tree, and gave every descendant a row prefixed client id at the end of RENDER_RESPONSE, so their partial state was saved under an id the postback never rebuilds.

Reset the row key at the end of encodeEnd, like TreeTableRenderer and UIData.encodeEnd already do.

### DIFF
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/tree/Tree008.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/tree/Tree008.java
@@ -1,0 +1,52 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2026 PrimeFaces
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.integrationtests.tree;
+
+import org.primefaces.model.TreeNode;
+
+import java.io.Serializable;
+
+import jakarta.annotation.PostConstruct;
+import jakarta.faces.view.ViewScoped;
+import jakarta.inject.Inject;
+import jakarta.inject.Named;
+
+import lombok.Data;
+
+@Named
+@ViewScoped
+@Data
+public class Tree008 implements Serializable {
+
+    @Inject
+    private TreeNodeService treeNodeService;
+
+    private TreeNode<String> root;
+
+    @PostConstruct
+    public void init() {
+        root = treeNodeService.createNodes();
+    }
+
+}
--- a/primefaces-integration-tests/src/main/webapp/tree/tree008.xhtml
+++ b/primefaces-integration-tests/src/main/webapp/tree/tree008.xhtml
@@ -1,0 +1,28 @@
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:h="jakarta.faces.html"
+    xmlns:f="jakarta.faces.core" xmlns:p="primefaces">
+
+<f:view contentType="text/html;charset=UTF-8" encoding="UTF-8">
+    <h:head>
+
+    </h:head>
+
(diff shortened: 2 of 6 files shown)


melloware:
@BalusC please review for your expectations

BalusC (approved):

BalusC:
lgtm

PR #15200, opened 2026-09-05
Fix #15163 #15164 #15165 #15166 #15167 reset iteration state after render

### LINKED ISSUE #15163: TreeTable: rowKey is not reset after the expand, collapse, cell edit, row edit and page ajax requests
## Describe the bug

Split off from #15161, as requested. Same shape as the `Tree` defect: a component sets its row index or row key, renders, and returns without resetting it. The `var` then stays in the request map, so whatever renders after it resolves `#{var}` to the last row. Where the component folds the row into the client id it hands its children, their state is saved under a client id which the next request does not rebuild.

`TreeTableRenderer.encodeTbody` resets the row key at line 470, so the full render is fine. So are the filter feature, which resets at `feature/FilterFeature.java:116`, and the sort feature, which renders through `encodeTbody`.

These five ajax features never reach that reset:

- `feature/ExpandFeature.java:43` sets the row key, renders the node children at line 51, and returns
- `feature/CollapseFeature.java:43` sets the row key and returns
- `feature/CellEditFeature.java:65` sets the row key and returns
- `feature/RowEditFeature.java:52` sets the row key and returns
- `feature/PageFeature.java:40` renders through `encodeNodeChildren` directly, which never reaches `encodeTbody`

`encodeNode` and `encodeNodeChildren` (`TreeTableRenderer.java:477` and `:827`) do not reset either.

`TreeTable` inherits `getContainerClientId` from `UITree`, which appends the row key while the tree table stands on a row.

## Reproducer

I have no page level reproducer for this one, and I want to be straight about that. These features render one fragment and nothing else renders after them in the same response, so there is no second component to observe the leaked var. What I did was read the source: the five entry points above set the row key and none of them resets it, while the two which are safe both do.

The visible consequence is on the state which Faces saves at the end of the ajax response. I could not make Mojarra report a mismatch for the expand case in a quick test, so treat that part as unproven.

## Expected behavior

After each of these ajax requests the tree table stands on no row, the same as after a full render.

## Proposed fix

`encodeEnd` is the single place where these features are dispatched, so one line there covers all five, and it is the same fix as the one proposed for `Tree` in #15161:

```java
    @Override
    public void encodeEnd(FacesContext context, TreeTable component) throws IOException {
        if (component.shouldEncodeFeature(context)) {
            for (TreeTableFeature feature : TreeTableFeatures.all()) {
                if (feature.shouldEncode(context, component)) {
                    feature.encode(context, this, component);
                }
            }
        }
        else {
            render(context, component);
        }

        // A descendant's client id carries the row key the tree table is left on, so it must not stay on a row.
        component.setRowKey(component.getValue(), null);
    }
```

## Proposed test

The assertion is one line: after the feature has encoded, `treeTable.getRowKey()` is null.

I tried to write this as a unit test in the `FacesContextMock` style, calling `new ExpandFeature().encode(context, new TreeTableRenderer(), treeTable)` with the `_expand` request parameter set. It needs more wiring than the mocks give: `Column.getColumnKey()` fails for a column which is not attached to a real view. Your integration suite is the better home for it, next to the existing `TreeTable` tests, where the assertion can be that a component rendered after the tree table shows no leaked var.

## PrimeFaces edition

Community

## PrimeFaces version

15.0.17, and `master` (16.0.0-SNAPSHOT) has the same code

## Theme

Saga

## JSF implementation

All

## JSF version

4.1

## Java version

21

## Browser(s)

Not browser specific

🤖 Generated with Claude Opus 5

### LINKED ISSUE #15164: DataTable: rowIndex is not reset after the scroll, cell edit, row edit and add row ajax requests
## Describe the bug

Split off from #15161, as requested. Same shape as the `Tree` defect: a component sets its row index or row key, renders, and returns without resetting it. The `var` then stays in the request map, so whatever renders after it resolves `#{var}` to the last row. Where the component folds the row into the client id it hands its children, their state is saved under a client id which the next request does not rebuild.

`DataTableRenderer.encodeTbody` resets the row index at lines 1116-1119, so the full render, paging, filtering and sorting are fine. So is the row expand feature, which resets at `feature/RowExpandFeature.java:69`.

These four ajax features are partial updates which never reach `encodeTbody`:

- `feature/ScrollFeature.java:66-73`, the live scroll and virtual scroll "load more" request, which loops `table.setRowIndex(i)` and returns
- `feature/CellEditFeature.java:63`
- `feature/RowEditFeature.java:52`
- `feature/AddRowFeature.java:42`

`DataTable` inherits `getContainerClientId` from `UIDataPatch` (`api/patch/UIDataPatch.java:477`), which appends the row index while the table stands on a row.

## Reproducer

As with the tree table, I have no page level reproducer, and I want to be straight about that. These features render one fragment and nothing else renders after them in the same response. What I did was read the source: these four set the row index and never reset it, while the five which are safe all end at `encodeTbody` or reset themselves.

## Expected behavior

After each of these ajax requests the table stands on no row, the same as after a full render.

## Proposed fix

`encodeEnd` is the single place where these features are dispatched, so one line there covers all four:

```java
    @Override
    public void encodeEnd(FacesContext context, DataTable component) throws IOException {
        ...
        // A descendant's client id carries the row the table is left on, so it must not stay on a row.
        component.setRowIndex(-1);

        context.getApplication().publishEvent(context, PostRenderEvent.class, component);
    }
```

## Proposed test

The assertion is one line: after the feature has encoded, `table.getRowIndex()` is -1.

A unit test in the `FacesContextMock` style does not come together for `DataTable`, because a render reaches `UITable.initFilterBy`, which needs a `SearchExpressionContextFactory` from `FactoryFinder` that the mocks do not register. Your integration suite is the better home, next to the existing `DataTable` tests.

## PrimeFaces edition

Community

## PrimeFaces version

15.0.17, and `master` (16.0.0-SNAPSHOT) has the same code

## Theme

Saga

## JSF implementation

All

## JSF version

4.1

## Java version

21

## Browser(s)

Not browser specific

🤖 Generated with Claude Opus 5

### LINKED ISSUE #15165: SubTable: rowIndex is never reset after render
## Describe the bug

Split off from #15161, as requested. Same shape as the `Tree` defect: a component sets its row index or row key, renders, and returns without resetting it. The `var` then stays in the request map, so whatever renders after it resolves `#{var}` to the last row. Where the component folds the row into the client id it hands its children, their state is saved under a client id which the next request does not rebuild.

`SubTableRenderer.encodeRow` sets the row index at line 106. The class never calls `setRowIndex(-1)`, anywhere. So the sub table stands on its last row after every render, not only after an ajax request.

`SubTableBase` extends `jakarta.faces.component.UIData` directly (`subtable/SubTableBase.java:34`), and not `UIDataPatch` as the other data components do. So it keeps the standard `UIData` behavior of folding the row index into `getClientId()` rather than into `getContainerClientId()`.

## Reproducer

```xml
<p:dataTable id="teams" var="team" value="#{bean.teams}">
    <p:subTable id="stats" var="stat" value="#{team.stats}">
        <p:column><h:outputText value="#{stat}" /></p:column>
    </p:subTable>
</p:dataTable>
<h:outputText id="afterSub" value="afterSub=[#{stat}]" />
```

With two teams of two rows each, the page renders `afterSub=[beta-2]`, the last row of the last sub table. `afterSub=[]` is expected. Checked on Tomcat with PrimeFaces 15.0.17 and Mojarra 4.1.15.

The same run also makes Mojarra report the polluted state keys on the next postback, in `ProjectStage.Development`:

```
WARNING The view rebuilt for this postback does not hold 3 of the components the response was rendered from, so their state is restored into nothing: [f:teams:stats:1, f:teams:stats:1:j_idt5, f:teams:stats:1:j_idt6].
```

## Expected behavior

After rendering, the sub table stands on no row, so `#{stat}` resolves to nothing and the state of its children is saved under the client id the next request rebuilds them with.

## Proposed fix

```java
    @Override
    public void encodeEnd(FacesContext context, SubTable component) throws IOException {
        int rowCount = component.getRowCount();

        encodeHeader(context, component);

        for (int i = 0; i < rowCount; i++) {
            encodeRow(context, component, i);
        }

        encodeFooter(context, component);

        // A descendant's client id carries the row the sub table is left on, so it must not stay on a row.
        component.setRowIndex(-1);
    }
```

## Proposed test

Red without the fix (`expected: <-1> but was: <1>`) and green with it:

```java
package org.primefaces.component.subtable;

class SubTableRendererTest {

    /**
     * The row the sub table stands on ends up in its own client id and in the client id of everything below it, and
     * it puts the row it holds in the request map under the sub table's var. So a render which iterated the sub
     * table leaves it standing on no row.
     */
    @Test
    void encodeEndLeavesTheSubTableStandingOnNoRow() throws Exception {
        FacesContext context = new FacesContextMock(new CollectingResponseWriter());

        SubTable subTable = new SubTable();
        subTable.setId("sub");
        subTable.setVar("item");
        subTable.setValue(Arrays.asList("one", "two"));

        new SubTableRenderer().encodeEnd(context, subTable);

        assertEquals(-1, subTable.getRowIndex());
        assertFalse(context.getExternalContext().getRequestMap().containsKey("item"));
    }
}
```

This needs the three maps on `ExternalContextMock` which #15161 proposes (`getRequestMap`, `getRequestParameterMap` and `getRequestHeaderMap` return `null` today).

## PrimeFaces edition

Community

## PrimeFaces version

15.0.17, and `master` (16.0.0-SNAPSHOT) has the same code

## Theme

Saga

## JSF implementation

All

## JSF version

4.1

## Java version

21

## Browser(s)

Not browser specific

🤖 Generated with Claude Opus 5

### LINKED ISSUE #15166: Columns: DynamicColumn.cleanModel() is never called, so the column model stays on the last rendered column
## Describe the bug

Split off from #15161, as requested. Same shape as the `Tree` defect: a component sets its row index or row key, renders, and returns without resetting it. The `var` then stays in the request map, so whatever renders after it resolves `#{var}` to the last row. Where the component folds the row into the client id it hands its children, their state is saved under a client id which the next request does not rebuild.

`DynamicColumn.applyModel()` sets the row index of the `Columns` component (`api/DynamicColumn.java:63-65`). `DynamicColumn.cleanModel()` resets it (`api/DynamicColumn.java:75-77`).

`cleanModel()` has no call site at all in `primefaces/src/main/java`. Every caller calls only `applyModel()`:

- `datatable/DataTableRenderer.java:423`, `:992`, `:1019`, `:1243`, `:1395`, `:1425`
- `treetable/TreeTableRenderer.java:385`, `:417`, `:530`, `:902`, `:934`
- `api/UITable.java:212`

So after a render the `Columns` component stands on whichever dynamic column was rendered last. `Columns` extends `PrimeUIData`, so it appends the index in `getContainerClientId`.

## Reproducer

```xml
<p:dataTable id="dyn" var="row" value="#{bean.rows}">
    <p:columns id="cols" var="col" value="#{bean.columns}">
        <h:outputText value="#{row}-#{col}" />
    </p:columns>
</p:dataTable>
<h:outputText id="afterCols" value="afterCols=[#{col}]" />
```

With two columns the page renders `afterCols=[col2]`, the last column. `afterCols=[]` is expected. Checked on Tomcat with PrimeFaces 15.0.17 and Mojarra 4.1.15.

## Expected behavior

After rendering, the `Columns` component stands on no column.

## Proposed fix

Call `cleanModel()` where the render of the dynamic columns ends, or drop the method if it is meant to be dead. Since `applyModel()` is called from eleven places, a `try`/`finally` around each use, or one reset at the end of `DataTableRenderer.encodeEnd` and `TreeTableRenderer.encodeEnd`, are both defensible. I would put it in the two `encodeEnd` methods, next to the row index reset the other two issues propose, so that all of it is in one place.

## Proposed test

The assertion is one line: after the table has encoded, `columns.getRowIndex()` is -1.

A unit test in the `FacesContextMock` style does not come together here, because rendering a `DataTable` reaches `UITable.initFilterBy`, which needs a `SearchExpressionContextFactory` from `FactoryFinder` that the mocks do not register. Your integration suite is the better home: render the markup above and assert that the component after the table shows an empty var.

## PrimeFaces edition

Community

## PrimeFaces version

15.0.17, and `master` (16.0.0-SNAPSHOT) has the same code

## Theme

Saga

## JSF implementation

All

## JSF version

4.1

## Java version

21

## Browser(s)

Not browser specific

🤖 Generated with Claude Opus 5

### LINKED ISSUE #15167: Timeline: var and varGroup are never removed from the request map
## Describe the bug

Split off from #15161, as requested. Same shape as the `Tree` defect: a component sets its row index or row key, renders, and returns without resetting it. The `var` then stays in the request map, so whatever renders after it resolves `#{var}` to the last row. Where the component folds the row into the client id it hands its children, their state is saved under a client id which the next request does not rebuild.

`TimelineRenderer` puts the iteration variable in the request map for every event it renders (`timeline/TimelineRenderer.java:485`) and for every group (`:328`). Neither is ever removed. So after the timeline has rendered, `#{var}` resolves to the last event and `#{varGroup}` to the last group, for the rest of the request.

`TimelineBase` extends `UIComponentBase` and has no positional client id, so this one leaks only the variables. The client ids of its children are not affected.

## Reproducer

```xml
<p:timeline id="timeline" value="#{bean.model}" var="event" />
<h:outputText id="afterTimeline" value="afterTimeline=[#{event}]" />
```

`afterTimeline` renders the data of the last event instead of nothing. The unit test below is the same thing without a page.

## Expected behavior

After rendering, neither variable is in the request map. Better still, the value which was there before is restored, so a timeline nested in an outer iteration over the same name does not destroy the outer variable.

## Proposed fix

```java
    @Override
    public void encodeEnd(FacesContext context, Timeline component) throws IOException {
        Map<String, Object> requestMap = context.getExternalContext().getRequestMap();
        Object originalVar = requestMap.get(component.getVar());
        Object originalVarGroup = requestMap.get(component.getVarGroup());

        try {
            encodeMarkup(context, component);
            encodeScript(context, component);
        }
        finally {
            // The var holds the event which is being encoded, so it must not outlive the encoding.
            restoreVar(requestMap, component.getVar(), originalVar);
            restoreVar(requestMap, component.getVarGroup(), originalVarGroup);
        }
    }

    private static void restoreVar(Map<String, Object> requestMap, String name, Object value) {
        if (LangUtils.isBlank(name)) {
            return;
        }

        if (value != null) {
            requestMap.put(name, value);
        }
        else {
            requestMap.remove(name);
        }
    }
```

Restoring rather than removing also holds when rendering throws, which is the shape `UITabPanel.process` already uses (`api/UITabPanel.java:796-799`).

## Proposed test

Red without the fix (`expected: <false> but was: <true>`) and green with it:

```java
package org.primefaces.component.timeline;

class TimelineRendererTest {

    /**
     * The timeline puts the event it renders in the request map under its var. So a render which iterated the events
     * leaves no var behind, and whatever renders after the timeline resolves the var of its own scope.
     */
    @Test
    void encodeEndLeavesNoVarBehind() throws Exception {
        FacesContext context = new FacesContextMock(new CollectingResponseWriter());
        UIViewRoot viewRoot = new UIViewRoot();
        viewRoot.setLocale(Locale.ENGLISH);
        context.setViewRoot(viewRoot);

        TimelineModel<Object, Object> model = new TimelineModel<>();
        model.add(TimelineEvent.builder().data("one").startDate(LocalDateTime.now()).build());
        model.add(TimelineEvent.builder().data("two").startDate(LocalDateTime.now()).build());

        Timeline timeline = new Timeline();
        timeline.setId("timeline");
        timeline.setVar("event");
        timeline.setValue(model);

        new TimelineRenderer().encodeEnd(context, timeline);

        assertFalse(context.getExternalContext().getRequestMap().containsKey("event"));
    }
}
```

This needs the three maps on `ExternalContextMock` which #15161 proposes.

## PrimeFaces edition

Community

## PrimeFaces version

15.0.17, and `master` (16.0.0-SNAPSHOT) has the same code

## Theme

Saga

## JSF implementation

All

## JSF version

4.1

## Java version

21

## Browser(s)

Not browser specific

🤖 Generated with Claude Opus 5

melloware:
Fix #15164 
Fix #15165 
Fix #15166 
Fix #15167
Fix #15163

Five components set a row index, a row key or an iteration variable while rendering and returned without resetting it. The var then stayed in the request map, so whatever rendered after them resolved it to the last row, and where the component folds the row into the client id it hands its children, their state was saved under a client id which the next request does not rebuild.

Rather than fixing each renderer, give the components one place to clean up after themselves, as suggested in #15167:

- RenderCleanupAware is implemented by a component whose renderer leaves iteration state behind.
- PrimeRendererWrapper calls it right after encodeEnd, also when encodeEnd threw. PrimeRenderKit hands out every CoreRenderer inside that wrapper, and PrimeRenderKitFactory installs the render kit. Renderers of the Faces implementation and of other libraries are handed out untouched.

The implementations:

- PrimeUIData resets the row index, which covers the DataTable scroll, cell edit, row edit and add row ajax requests (#15164), none of which reach encodeTbody.
- SubTable resets the row index (#15165). It extends UIData directly, so it folds the row into its own client id as well.
- DataTable and TreeTable also reset every p:columns below them, since DynamicColumn#cleanModel had no call site at all (#15166).
- Timeline removes var and varGroup from the request map (#15167).
- UITree resets the row key, which moves the fix merged for #15161 off TreeRenderer and onto the interface, and covers TreeTable too.

Badge and OverlayPanel cast the result of RenderKit#getRenderer to their own renderer type, so they now go through ComponentUtils#getUnwrappedRenderer.

### DIFF
--- a/docs/migrationguide/17_0_0.md
+++ b/docs/migrationguide/17_0_0.md
@@ -1,0 +1,28 @@
+# Migration guide 16.0.0 -> 17.0.0
+
+## Renderers are wrapped
+
+PrimeFaces now installs a `RenderKitFactory` (`org.primefaces.renderkit.PrimeRenderKitFactory`). Its render kit hands
+out every PrimeFaces renderer, meaning every `org.primefaces.renderkit.CoreRenderer`, inside a
+`org.primefaces.renderkit.PrimeRendererWrapper`. The wrapper resets the iteration state of components which implement
+`org.primefaces.component.api.IterationCleanupAware` after `decode` and after `encodeEnd`, so that a row index, a row
+key or an iteration variable never outlives the phase which set it.
+
+Renderers of the Faces implementation and of other component libraries are handed out untouched.
+
+This only matters if you cast the result of `RenderKit#getRenderer` to a PrimeFaces renderer type:
+
+```java
+// throws ClassCastException as of 17.0.0
+DataTableRenderer renderer = (DataTableRenderer) context.getRenderKit()
+        .getRenderer(DataTable.COMPONENT_FAMILY, DataTable.DEFAULT_RENDERER);
+```
+
+`PrimeRendererWrapper` implements `jakarta.faces.FacesWrapper`, so unwrap it:
+
+```java
+DataTableRenderer renderer = ComponentUtils.getUnwrappedRenderer(context,
+        DataTable.COMPONENT_FAMILY, DataTable.DEFAULT_RENDERER);
+```
+
+`ComponentUtils#getUnwrappedRenderer` already existed and works on both versions.
--- a/docs/migrationguide/migrationguide.md
+++ b/docs/migrationguide/migrationguide.md
@@ -1,5 +1,6 @@
 # Migration Guide
 
+* [16.0.0 -> 17.0.0](../migrationguide/17_0_0.md)
 * [15.0.0 -> 16.0.0](../migrationguide/16_0_0.md)
 * [14.0.0 -> 15.0.0](../migrationguide/15_0_0.md)
 * [13.0.0 -> 14.0.0](../migrationguide/14_0_0.md)
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datatable/DataTable052.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datatable/DataTable052.java
@@ -1,0 +1,73 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2026 PrimeFaces
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.integrationtests.datatable;
+
+import java.io.Serial;
+import java.io.Serializable;
+import java.util.Arrays;
+import java.util.List;
+
+import jakarta.annotation.PostConstruct;
+import jakarta.faces.view.ViewScoped;
+import jakarta.inject.Inject;
+import jakarta.inject.Named;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Named
+@ViewScoped
+@Data
+public class DataTable052 implements Serializable {
+
+    @Serial private static final long serialVersionUID = 1L;
+
+    private List<ProgrammingLanguage> progLanguages;
+    private List<Team> teams;
+
+    @Inject
+    private ProgrammingLanguageService service;
+
+    @PostConstruct
+    public void init() {
+        progLanguages = service.getLangs();
+        teams = Arrays.asList(
+                new Team("alpha", Arrays.asList("alpha-1", "alpha-2")),
+                new Team("beta", Arrays.asList("beta-1", "beta-2")));
+    }
+
+    public List<String> getColumns() {
+        return Arrays.asList("id", "name");
+    }
+
+    @Data
+    @AllArgsConstructor
+    public static class Team implements Serializable {
+
+        @Serial private static final long serialVersionUID = 1L;
+
+        private String name;
+        private List<String> stats;
+    }
+}
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/timeline/Timeline004.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/timeline/Timeline004.java
@@ -1,0 +1,58 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2026 PrimeFaces
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.integrationtests.timeline;
+
+import org.primefaces.model.timeline.TimelineEvent;
+import org.primefaces.model.timeline.TimelineGroup;
+import org.primefaces.model.timeline.TimelineModel;
+
+import java.io.Serial;
+import java.io.Serializable;
+import java.time.LocalDate;
+
+import jakarta.annotation.PostConstruct;
+import jakarta.faces.view.ViewScoped;
+import jakarta.inject.Named;
+
+import lombok.Data;
+
+@Named
+@ViewScoped
+@Data
+public class Timeline004 implements Serializable {
+
+    @Serial private static final long serialVersionUID = 1L;
+
+    private TimelineModel<String, String> model;
+
+    @PostConstruct
+    public void init() {
+        model = new TimelineModel<>();
+
+        model.addGroup(new TimelineGroup<>("releases", "Releases"));
+        model.add(TimelineEvent.<String>builder().data("PrimeFaces 14").group("releases").startDate(LocalDate.of(2024, 4, 1)).build());
+        model.add(TimelineEvent.<String>builder().data("PrimeFaces 15").group("releases").startDate(LocalDate.of(2025, 4, 1)).build());
+        model.add(TimelineEvent.<String>builder().data("PrimeFaces 16").group("releases").startDate(LocalDate.of(2026, 4, 1)).build());
+    }
+}
--- a/primefaces-integration-tests/src/main/webapp/datatable/dataTable052.xhtml
+++ b/primefaces-integration-tests/src/main/webapp/datatable/dataTable052.xhtml
@@ -1,0 +1,52 @@
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml"
+      xmlns:h="jakarta.faces.html"
+      xmlns:f="jakarta.faces.core"
+      xmlns:p="primefaces">
+
+<f:view contentType="text/html;charset=UTF-8" encoding="UTF-8">
+    <h:head>
+
+    </h:head>
+
+    <h:body>
+
+        <h:form id="form">
+
+            <p:dataTable id="dynamic" var="lang" value="#{dataTable052.progLanguages}">
+                <p:columns id="cols" var="col" value="#{dataTable052.columns}" headerText="#{col}">
+                    <h:outputText value="#{lang[col]}"/>
+                </p:columns>
+            </p:dataTable>
+
+            <!-- p:columns must not leak its var, so this renders empty and not the last column -->
+            <h:outputText id="afterCols" value="afterCols=[#{col}]"/>
+
+            <p:dataTable id="grouped" var="team" value="#{dataTable052.teams}">
+                <p:columnGroup type="header">
+                    <p:row>
+                        <p:column headerText="Stats"/>
+                    </p:row>
+                </p:columnGroup>
+
+                <p:subTable id="stats" var="stat" value="#{team.stats}">
+                    <f:facet name="header">
+                        <h:outputText value="#{team.name}"/>
+                    </f:facet>
+                    <p:column>
+                        <h:outputText value="#{stat}"/>
+                    </p:column>
+                </p:subTable>
+            </p:dataTable>
+
+            <!-- p:subTable must not leak its var, so this renders empty and not the last row -->
+            <h:outputText id="afterSub" value="afterSub=[#{stat}]"/>
+
+            <p:commandButton id="buttonUpdate" value="Update" update="@form"/>
+
+        </h:form>
+
+    </h:body>
+</f:view>
+
+</html>
--- a/primefaces-integration-tests/src/main/webapp/timeline/timeline004.xhtml
+++ b/primefaces-integration-tests/src/main/webapp/timeline/timeline004.xhtml
@@ -1,0 +1,34 @@
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml"
+      xmlns:h="jakarta.faces.html"
+      xmlns:f="jakarta.faces.core"
+      xmlns:p="primefaces">
+
+<f:view contentType="text/html;charset=UTF-8" encoding="UTF-8">
+    <h:head>
+
+    </h:head>
+
+    <h:body>
+
+        <h:form id="form">
+
+            <p:timeline id="timeline" value="#{timeline004.model}" height="250px" var="event" varGroup="group">
+                <f:facet name="group">
+                    <h:outputText value="#{group}"/>
+                </f:facet>
+                <h:outputText value="#{event}"/>
+            </p:timeline>
+
+            <!-- the timeline must not leak its vars, so these render empty and not the last event and group -->
+            <h:outputText id="afterEvent" value="afterEvent=[#{event}]"/>
+            <h:outputText id="afterGroup" value="afterGroup=[#{group}]"/>
+
+            <p:commandButton id="buttonUpdate" value="Update" update="@form"/>
+
+        </h:form>
+
+    </h:body>
+</f:view>
+
+</html>
--- a/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datatable/DataTable052Test.java
+++ b/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/datatable/DataTable052Test.java
@@ -1,0 +1,105 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009-2026 PrimeFaces
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.primefaces.integrationtests.datatable;
+
+import org.primefaces.selenium.AbstractPrimePage;
+import org.primefaces.selenium.PrimeSelenium;
+import org.primefaces.selenium.component.CommandButton;
+import org.primefaces.selenium.component.DataTable;
+
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Order;
+import org.junit.jupiter.api.Test;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.FindBy;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * GitHub #15165 and #15166: a component which iterates must not leave its iteration variable in the request map, so
+ * that whatever renders after it resolves the variable of its own scope instead of the last row it rendered.
+ */
+class DataTable052Test extends AbstractDataTableTest {
+
+    @Test
+    @Order(1)
+    @DisplayName("DataTable: GitHub #15166 p:columns does not leak its var into what renders after the table")
+    void dynamicColumnsVarDoesNotLeakAfterRender(Page page) {
+        // Arrange
+        assertTrue(page.dynamic.isDisplayed());
+
+        // Assert
+        assertEquals("afterCols=[]", page.afterCols.getText());
+
+        // Act
+        PrimeSelenium.guardAjax(page.buttonUpdate).click();
+
+        // Assert
+        assertEquals("afterCols=[]", page.afterCols.getText());
+
+        assertNoJavascriptErrors();
+    }
+
+    @Test
+    @Order(2)
+    @DisplayName("DataTable: GitHub #15165 p:subTable does not leak its var into what renders after the table")
+    void subTableVarDoesNotLeakAfterRender(Page page) {
+        // Arrange
+        assertTrue(page.grouped.isDisplayed());
+
+        // Assert
+        assertEquals("afterSub=[]", page.afterSub.getText());
+
+        // Act
+        PrimeSelenium.guardAjax(page.buttonUpdate).click();
+
+        // Assert
+        assertEquals("afterSub=[]", page.afterSub.getText());
+
+        assertNoJavascriptErrors();
+    }
+
+    public static class Page extends AbstractPrimePage {
+        @FindBy(id = "form:dynamic")
+        DataTable dynamic;
+
+        @FindBy(id = "form:afterCols")
+        WebElement afterCols;
+
+        @FindBy(id = "form:grouped")
+        DataTable grouped;
+
+        @FindBy(id = "form:afterSub")
+        WebElement afterSub;
+
+        @FindBy(id = "form:buttonUpdate")
+        CommandButton buttonUpdate;
+
+        @Override
+        public String getLocation() {
+            return "datatable/dataTable052.xhtml";
+        }
+    }
+}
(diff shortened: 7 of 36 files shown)


melloware:
@tandraschko @BalusC this might be overkill...

BalusC:
# Review

Head 87e9908e1f. Fixes claimed: #15163, #15164, #15165, #15166, #15167.

## Approach

A `RenderKitFactory` → `RenderKit` → `Renderer` wrapper chain. `PrimeRenderKitFactory` installs `PrimeRenderKit`, which hands out every `CoreRenderer` inside a `PrimeRendererWrapper`. That wrapper calls `IterationCleanupAware#cleanupIterationState` in a `finally` after both `decode` and `encodeEnd`. Renderers of the Faces implementation and of other libraries are handed out untouched.

The approach is sound. `UIData.processDecodes` and `UITree.processDecodes:517-525` both call `decode` last, so the javadoc's claim that nothing which still needs the row runs after the cleanup holds. Event broadcasting is safe too: `UITree.queueEvent` wraps the event with the row key and `broadcast` re-applies it (`UITree.java:463-478`), and `UIData` does the same with the row index.

CI was still pending on all 10 checks at the time of this review.

## Major

### M1. The showcase Documentation tab reports `PrimeRendererWrapper` as the renderer class of every component

`primefaces-showcase/src/main/java/org/primefaces/showcase/util/taglib/TagInfo.java:86` reads `renderKit.getRenderer(componentFamily, tag.getRendererType())` and then `renderer.getClass().getName()`. `PrimeRenderKit` wraps every `CoreRenderer`, so the tab shows `org.primefaces.renderkit.PrimeRendererWrapper` instead of e.g. `org.primefaces.component.datatable.DataTableRenderer`. The PR found and fixed this pattern in `Badge` and `OverlayPanel` but missed the third caller. Outside the showcase the impact is limited to code which casts a PrimeFaces renderer, since only `CoreRenderer` instances are wrapped. That is exactly what PrimeFaces Extensions does.

**Fix:** use `ComponentUtils.getUnwrappedRenderer(context, componentFamily, tag.getRendererType())` in `TagInfo`, and note the wrapper in the migration guide.

### M2. An exception from the cleanup replaces the original exception

`PrimeRendererWrapper.decode` and `PrimeRendererWrapper.encodeEnd` both call the cleanup in a bare `finally`. Example: a `p:dataTable` whose `value` EL throws mid-render. The `FacesException` enters the `finally`, `DataTableBase.cleanupIterationState` → `resetDynamicColumns` → `forEachColumn` reaches a `ui:repeat` child and calls `child.visitTree(...)` (`ColumnAware.java:144-160`), which evaluates the repeat's `value` EL again and throws again. The user then sees the second exception and the root cause is gone, exactly when it is needed most.

**Fix:** in the shared private `cleanupIterationState(context, component)` helper, keep the in-flight throwable, run the cleanup in its own `try`/`catch (Throwable)`, and `addSuppressed` it onto the original. Rethrow the cleanup failure only when nothing was in flight. One change covers both call sites.

### M3. The timeline `var`/`varGroup` leak survives on the `TimelineUpdater` path

`DefaultTimelineUpdater.processCrudOperations` runs in `beforePhase(RENDER_RESPONSE)` and calls `encodeGroup`/`encodeEvent` on the unwrapped renderer, which puts `varGroup` and `var` into the request map (`TimelineRenderer.java:328` and `:485`). The point of `TimelineUpdater` is to patch the widget over JS without re-rendering the component, so no `encodeEnd` runs and `cleanupIterationState` never fires. Both vars then stay in the request map for the whole render of that ajax response. That is the defect #15167 describes.

**Fix:** call `timeline.cleanupIterationState(context)` in a `finally` at the end of the `invokeOnComponent` callback in `processCrudOperations`.

### M4. The timeline cleanup removes the var instead of restoring the outer value

`TimelineBase.cleanupIterationState` calls `requestMap.remove(getVar())` whenever `var` is not blank, also when the renderer never set it (it only sets when `data != null`). With `<p:dataTable var="e">...<p:timeline var="e">` the timeline deletes the table's per-row `e`. The new decode hook makes this wider: the cleanup now also runs from `decode`, so the same thing can happen during the outer table's `processDecodes`, not only at render time. #15167 asks for restore explicitly: "Better still, the value which was there before is restored, so a timeline nested in an outer iteration over the same name does not destroy the outer variable."

**Fix:** let `encodeGroup`/`encodeEvent` save the previous value and let `cleanupIterationState` put it back (or remove when there was none). `ComponentUtils.executeInRequestScope` already has this idiom. Restoring is better than only removing when the renderer actually set the key, because that still destroys a same-named outer var.

### M5. The decode half is only tested for `Tree`

The only new decode test is `TreeRendererTest.decodeLeavesTheTreeStandingOnNoRow`. Nothing covers `DraggableRowsFeature.decode`, which sets the row index and returns without resetting it, and which is the DataTable decode leak that motivated the hook. Nothing covers TreeTable's five ajax feature paths either, which is what the retitled PR now claims for #15163 — and the reporter of that issue said clearly they had no page level reproducer for those.

**Fix:** add a unit test which decodes a `p:dataTable` with `_fromIndex`/`_toIndex` rowReorder parameters and asserts `getRowIndex() == -1` afterwards, plus one TreeTable decode test which asserts the row key is null after an expand request. Unit tests are better than integration tests here: these paths render one fragment and nothing renders after them, which is why the leak is not visible in a page.

### M6. No test covers the `TimelineUpdater` path

Related to M3. `TreeTableTest.cleanupIterationStateLeavesTheTreeTableStandingOnNoRowAndNoColumn` closed the TreeTable gap, but `DefaultTimelineUpdater` has none.

**Fix:** add a unit test which asserts the request map holds neither `var` nor `varGroup` after `processCrudOperations`, together with the M3 fix.

## Minor

### m7. `resetDynamicColumns` visits every `ui:repeat` for nothing

`ColumnAware.java:57` passes `unwrapDynamicColumns=false`, so its callback only matches `Columns`, while the `ui:repeat` branch of `forEachColumn` only tests `Column`. That `visitTree` can never produce a hit, but it runs on every DataTable and TreeTable render and evaluates the repeat's `value` again. It is also the call which carries the risk in M2.

**Fix:** skip `resetDynamicColumns` when the table has no `Columns` descendant, cached as a boolean next to the `columns` list. The alternative, a separate scan which does not go into `ui:repeat`, duplicates traversal logic, so the short circuit is better.

## Nit

### n8. Two `ConcurrentHashMap` lookups per renderer lookup on a hot path

`PrimeRenderKit.getRenderer` is called from `UIComponentBase` for `decode`, `getRendersChildren`, `encodeBegin`, `encodeChildren`, `encodeEnd` and `convertClientId`, so it runs several times per component per request.

**Fix:** use one map keyed by `family + '\0' + rendererType`. The nested map buys nothing, because `addRenderer` already invalidates per entry.

### n9. The replacement comment describes the change instead of a constraint

`TreeRenderer.java:289` is a comment at the end of `encodeEnd` with no code under it: `// the tree is left standing on the node which was encoded last, UITree#cleanupIterationState resets it`. It goes stale as soon as the mechanism is renamed, which already happened once inside this PR, and the contract is asserted by `TreeRendererTest#encodeEndLeavesTheTreeStandingOnNoRow`.

**Fix:** delete the comment.

## Checked and fine

- `UITree.cleanupIterationState`'s `setRowKey(null, null, null)` is equivalent to the removed `setRowKey(root, null)`, because the null-rowKey branch never reads `root`. The `rowKey != null` guard is safe: `encodeFilteredNodes` removes its own var at `TreeRenderer.java:314`, and `TreeTableRenderer` never writes the request map directly.
- The `getRowIndex() != -1` guard in `resetDynamicColumns` also catches the stateless path, because `setRowModel` writes `PropertyKeys.rowIndex` into the same StateHelper that `getRowIndex()` reads.
- `RenderKitWrapper` delegates `getComponentFamilies`, `getRendererTypes` and `getClientBehaviorRenderer*`, so the wrapper is complete. `<factory>` is an unordered choice in the 4.0 schema, so the new `<render-kit-factory>` placement is valid.
- `PrimeRenderKitFactory` caches per render kit id, so the renderer wrapper cache inside `PrimeRenderKit` stays alive.
- `ExternalContextMock.getRequestMap` returns a stable field and this PR does not touch it, so the new `assertFalse(containsKey(...))` assertions are not vacuous.
- The row reorder path is fine. `PrimeUIData.setRowIndexWithoutRowStatePreserved` does `saveDescendantState()` → `setRowModel` → `restoreDescendantState()`, so `DraggableRowsFeature`'s `setRowIndex(fromIndex)` restores that row's descendant state into the components, and the cleanup's `setRowIndex(-1)` saves the same values back under the same client id key. The list rotation changes the row data, not the save keys, so the round trip is a no-op. The same reasoning applies to the checkbox branch of TreeTable's `SelectionFeature.decode`.

🤖 Generated with Claude Opus 5

melloware:
Thank you for the review let me implement the changes!

melloware:
All nine items are addressed in a4a731e808, as a follow-up commit so the review round stays separate.

## As proposed

**M1** `TagInfo.java:86` was the third caster. It goes through `ComponentUtils.getUnwrappedRenderer` now. `docs/migrationguide/17_0_0.md` did not exist yet, so it is new and linked from the index; it documents the wrapper and the unwrap idiom.

**M2** The shared helper takes the in-flight exception, runs the cleanup in its own `try`/`catch`, and attaches a cleanup failure with `addSuppressed`. A cleanup failure only propagates when the phase itself completed. One deliberate narrowing: it catches `RuntimeException` rather than `Throwable`. `cleanupIterationState` declares nothing checked, and an `Error` out of a cleanup should win rather than be filed under a lesser exception.

**M3** `processCrudOperations` cleans up in a `finally` inside the `invokeOnComponent` callback.

**M4** `TimelineBase#setIterationVar` records what the request map held and `cleanupIterationState` puts it back, only for what the encoding actually replaced. You were right that the decode hook widened this: the new `decodeLeavesTheVarOfAnOuterIterationAlone` fails against the old code, so the hook had introduced a way to clobber an outer table's `var` during its `processDecodes`.

**M6** Added `DefaultTimelineUpdaterTest`. It needed the mock render kit to hand out a real renderer, so `RenderKitMock` honours `addRenderer` now (falling back to `RendererMock`) and `FacesContextMock` keeps one instance instead of building a new one per call.

**n9** Deleted.

## Three done differently

**M5** `ExpandFeature#shouldDecode` returns `false` on TreeTable, so expand dispatches from `encodeEnd`, not `decode`. The test drives `encodeEnd` with `_encodeFeature` and `_expand`. It needed `setFilterByAsMap(new HashMap<>())` to keep `initFilterBy` away from the `SearchExpressionContextFactory` the mocks lack; mocking that properly also needs `SearchExpressionHandlerMock#resolveComponent`, which throws `UnsupportedOperationException`, so pre-seeding was the cheaper route. The DataTable `rowReorder` decode test is as you described.

**m7** Dropped `forEachColumn` instead of adding a cached boolean. The flag needs new mutable state per component plus an invalidation point, and it leaves the cleanup calling a method which evaluates EL (`applyStatelessModel`, the `ui:repeat` `visitTree`), which is the M2 hazard itself. A cleanup running after a failed phase should evaluate nothing, so `resetDynamicColumns` walks children, column group rows and nested `ColumnAware` itself. About fifteen lines, no EL, no allocation.

**n8** Keyed the cache on the renderer instead of `family + '\0' + rendererType`. A concat key allocates a String per lookup, which costs more on that path than the second map hit. The renderer is an application scoped singleton with identity equals, so it is one lookup and no allocation, and it drops the `addRenderer` override, since a replaced renderer is simply a different key.

## Verification

504 unit tests green. Every new assertion was confirmed red against the unfixed code, in four passes:

- TreeTable expand: `expected: <null> but was: <0_0>`, the #15163 leak.
- DataTable rowReorder decode: `expected: <-1> but was: <0>`.
- `DefaultTimelineUpdaterTest`: `expected: <false> but was: <true>`, and `was: <added>` for the clobbered outer var.
- M4 restore: `expected: <the row of the outer table> but was: <null>` on encode, on decode and through the updater.
- M2: with the bare `finally`, `expected: <java.io.IOException> but was: <java.lang.IllegalStateException>`, the original exception replaced exactly as you described.

Full integration suite on Mojarra 4.1, 964 tests: the same 2 failures and 11 errors as an untouched baseline build on this machine (`DataExporter001/002/003Test` download timeouts, `Csv003Test` message bundle mismatches). I verified those by stashing the branch, rebuilding and rerunning those four classes. No new failures. Selection paths also spot checked on MyFaces 4.1. Checkstyle and license pass on all modules.

🤖 Generated with Claude Opus 5

tandraschko:
Will review mid october

BalusC:
# Re-review

Head a4a731e808. All ten CI jobs pass individually, and `build.yml` has no `continue-on-error`, so that is a real green.

## Round 1 findings — all fixed

| ID | Fix |
|---|---|
| M1 showcase renderer class | `TagInfo` now uses `ComponentUtils.getUnwrappedRenderer` |
| M2 exception masking | `PrimeRendererWrapper` keeps the in-flight exception, attaches a cleanup failure as suppressed, and rethrows the cleanup failure only when the phase itself completed. Three tests cover it |
| M3 `TimelineUpdater` leak | `finally { timeline.cleanupIterationState(context); }` in `processCrudOperations`, correctly scoped to the try-with-resources block |
| M4 remove vs restore | The new `TimelineBase.setIterationVar` records the previous value once per name, and `cleanupIterationState` puts it back, or removes it when there was none, and only for what was really replaced |
| M5 decode untested | `DataTableTest.decodeLeavesTheTableStandingOnNoRow` drives the real `DraggableRowsFeature` through the wrapper, and `TreeTableTest.encodeEndOfTheExpandFeatureLeavesTheTreeTableStandingOnNoRow` covers the five ajax paths |
| M6 `TimelineUpdater` untested | `DefaultTimelineUpdaterTest`, two tests, including the outer iteration restore |
| m7 pointless `ui:repeat` visit | `resetDynamicColumns` now walks the children directly and evaluates no EL, with a javadoc which says why a cleanup must not evaluate anything |
| n8 double map lookup | One `computeIfAbsent` |
| n9 narrating comment | Deleted |

The migration guide entry is right: master is already 17.0.0-SNAPSHOT and `17_0_0.md` is new. `PrimeRendererWrapper` does implement `FacesWrapper` through `RendererWrapper`, so the unwrap advice is correct.

I also checked the `resetDynamicColumns` rewrite against the old `forEachColumn` traversal. The `ColumnGroup` → `Row` path and the nested `ColumnAware` path are equivalent, and dropping the `Column` and `ui:repeat` branches loses nothing, because a `p:columns` cannot sit inside a `p:column` and the old `ui:repeat` branch only ever tested for `Column`.

## New — major

### M10. `PrimeRenderKit.wrappers` is now unbounded when another render kit in the chain does not cache its renderers

The map changed from `Map<String, Map<String, Renderer>>` keyed by family and type to `Map<Renderer, Renderer>` keyed by the wrapped renderer instance, and the `addRenderer` override which evicted entries is gone (`PrimeRenderKit.java:45` and `:64`). Nothing ever removes from it, and `PrimeRenderKitFactory` caches the `PrimeRenderKit` for the life of the application.

The javadoc states the assumption, "an application scoped singleton without an equals of its own", and both Mojarra and MyFaces keep their renderers in a map and return the stored instance, so a plain stack is fine. It stops being fine when a third party `RenderKitWrapper` sits in between and returns `new SomeWrapper(super.getRenderer(...))` on every lookup, which is exactly the pattern this PR itself introduces. `UIComponentBase` calls `getRenderer` several times per component per request, so the map then grows by about the component count on every request, forever. What was lost is the bounded key space which the old keying gave for free.

**Fix:** key by `family + '\0' + rendererType` and keep the identity check: return the cached wrapper when `getWrapped() == renderer`, otherwise build a new one and overwrite. That is still one lookup, it still handles a renderer replaced through `addRenderer` without needing the override, and the key space is bounded by the number of registered renderer types.

## New — nit

### n11. An `Error` from the phase now skips the cleanup

`PrimeRendererWrapper.decode` catches `RuntimeException` and `encodeEnd` catches `IOException | RuntimeException`. Where the old bare `finally` also cleaned up after a `StackOverflowError` or an `OutOfMemoryError`, the new shape does not. On an `Error` the request is usually lost anyway, so this costs little.

**Fix:** to get the old guarantee back, catch `Throwable` into the local and rethrow it with the same instanceof ladder plus a final `throw (Error) thrown`. Leaving it as it is now is also fine, but then please say so in the javadoc, so the next reader knows it was a choice.

### n12. `TimelineBase.setIterationVar` is public without an internal usage marker

It is a new public method on a public component base class, and only `TimelineRenderer` calls it. The codebase marks this kind of method with a `NOTE: this is for internal usage only!` javadoc line, for example `ColumnAware#forEachColumn`.

**Fix:** add that line to the existing javadoc.

🤖 Generated with Claude Opus 5

melloware:
All three items are addressed in 9e122fa754, again as a follow-up commit so the round stays separate.

**M10** The cache is keyed by `family + '\0' + rendererType` again, and the cached wrapper is handed out only while `getWrapped() == renderer`. That keeps the key space bounded by the number of registered renderer types no matter what a render kit further down the chain does, and it still needs no `addRenderer` override, since a replaced renderer simply fails the identity check. You were right that this PR is itself the counterexample to the assumption I keyed on: `PrimeRenderKit` is a `RenderKitWrapper` which builds something new per lookup, so assuming nobody else does that was not defensible.

Two tests. `keepsTheCacheBoundedWhenTheDelegateHandsOutANewRendererEveryTime` drives a delegate which returns `new CoreRenderer` per call and asserts the map holds one entry after 100 lookups — `expected: <1> but was: <100>` against the old keying. `keepsTheWrappersOfTwoRendererTypesApart` pins that the new key really separates two types of one family, which the old keying got for free.

**n11** Restored the old guarantee rather than documenting the gap, because it turned out to cost nothing: `catch (Throwable thrown) { cleanup(...); throw thrown; }` compiles as is under Java 7 precise rethrow, since the inferred throw set of `super.decode` is `RuntimeException | Error` and of `super.encodeEnd` adds `IOException`. No instanceof ladder, and the method is shorter than the version it replaces. Both new tests fail against the previous commit with `expected: <1> but was: <0>` cleanups.

What I did document instead is the other half you can now read off the code: an `Error` out of the *cleanup* is still not caught, so it wins over the phase exception rather than being filed under it as suppressed. `encodeEndLetsAnErrorFromTheCleanupWin` asserts it.

**n12** Added.

## Verification

509 unit tests green, checkstyle and license pass. The three new assertions were each confirmed red against a4a731e808 by stashing only the two main sources and rerunning: `expected: <1> but was: <100>` for the cache, `expected: <1> but was: <0>` twice for the `Error` cleanup.

I dropped one test I had written alongside these — a "delegate rebuilds its renderer" case which turned out to be `wrapsTheReplacementAfterAddRenderer` with the call one level lower.

🤖 Generated with Claude Opus 5

BalusC:
No new findings, good to go. Await Thomas tho.

> this might be overkill

Yeah. That's the tradeoff of being consistent. There are basically 2 options:
1. remember to implement an interface and provide cleanup methods
2. remember to properly override decode and encodeEnd and invoke cleanup methods yourself

There are currently 4 classes needing this. If there were 2 I'd have chosen option 2 instead. But there are 4. And there is potential this grows. So yeah option 1 is in long term better to avoid same mistake in future iteration components.

PR #15206, opened 2026-09-07
Fix #14295: allow summaryRow to define its own row group

### LINKED ISSUE #14295: SummaryRow component not appearing in version 12+
### Describe the bug

I tested it in the showcase, and when I remove the headerRow from the rowGroup.xhtml page, the summaryRow also disappears.

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

None

### PrimeFaces version

15

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

4.0

### Java version

17

### Browser(s)

_No response_

melloware:
Fixes #14295

## Problem

Removing `<p:headerRow>` from a table also makes `<p:summaryRow>` disappear.

Since the `SortMeta` refactor in v10, `DataTableRenderer.encodeRows` derives the row group boundary solely from `getHighestPriorityActiveSortMeta()`:

```java
SortMeta sort = table.getHighestPriorityActiveSortMeta();
boolean encodeSummaryRow = (!summaryRows.isEmpty() && sort != null);
```

A header row is the only thing that reliably puts an *active* `SortMeta` into the map — `UITable.initSortBy()` synthesizes one from `<p:headerRow>` at `MAX_PRIORITY`. Without a header row, and unless a column happens to be actively sorted, `sort == null` and no summary rows are rendered. That is also why the reporter saw the summary rows reappear intermittently: clicking any sortable column header creates an active meta.

Before v10 the group expression came from the table's own `sortBy` attribute, independently of the header row (`7.0` `DataTableRenderer` lines 1206-1223), so `summaryRow` worked without one. There was no attribute left to supply the group.

## Fix

**`summaryRow` can define its own group**, mirroring `headerRow`:

- `SummaryRowBase` — new `field` and `groupBy` properties.
- `SummaryRow.getGroupByValueExpression(context, var)` — resolves `groupBy` (a VE) or builds one from `field`; `null` when neither is set.
- `DataTableRenderer.encodeRows` — the summary group is resolved independently of the header group; the summary row's own expression wins, otherwise it falls back to the table group exactly as before.

```xhtml
<p:dataTable var="customer" value="#{bean.customers}">
    <p:column headerText="Representative" sortBy="#{customer.representative.name}" sortOrder="asc" />
    <p:column headerText="Name" field="name" />

    <!-- no headerRow needed -->
    <p:summaryRow field="representative.name">
        ...
    </p:summaryRow>
</p:dataTable>
```

**Secondary — the group no longer depends on sort priority.** `UITable.getGroupByMeta()` prefers the header-row meta, then a `groupRow` column, and only then the highest-priority active sort. This matters in multi-sort, where a `groupRow` column and a user-sorted column are both active at `MIN_PRIORITY` and `min()` resolved by declaration order. In single-sort nothing changes, because `SortFeature.decode()` deactivates every non-header-row meta on a sort click. Resolution is now also skipped entirely when the table has neither a header nor a summary row.

`encodeSummaryRow(…, SortMeta)` is kept as a delegating overload alongside the new `ValueExpression` one, so subclassed renderers still compile.

## Docs & showcase

- Removed the dead `sortBy="#{customer.representative.name}"` from the showcase Header Row demo — since v10 `sortBy` expects `SortMeta`, so this evaluates to `null` off-row and contributes nothing, while implying it drives the grouping.
- Added a **Summary Row** card to the showcase demonstrating the header-row-less pattern.
- Documented in `datatable.md` where a summary row takes its group from, and that it groups but does not sort.

## Verification

New IT triad `dataTable052` (`@Tag("DataTable-rowgroup")`) covering summary row grouping without a header row, on first render and after an AJAX update.

The test pins the reported bug — with `field="type"` removed from the page it fails with `expected: <7> but was: <5>`, i.e. zero summary rows.

```
DataTable-rowgroup + DataTable-sort:  Tests run: 115, Failures: 0, Errors: 0
primefaces unit tests:                Tests run: 480, Failures: 0, Errors: 0, Skipped: 1
checkstyle + license (both modules):  BUILD SUCCESS
```

Note this does not restore the pre-v10 `sortBy="#{var.field}"` string form on `dataTable` — that is gone by design with the `SortMeta` refactor, and reviving it would reintroduce a second, conflicting meaning for `sortBy`.

### DIFF
--- a/docs/17_0_0/components/datatable.md
+++ b/docs/17_0_0/components/datatable.md
@@ -636,6 +636,27 @@ group. To enable this method, set groupRow to true on the grouping column.
     <p:column field="id" headerText="Id" />
 </p:dataTable>
 ```
+A `summaryRow` closes the group defined by the table, so on its own it needs to know what that group is. It picks
+up the group from a `headerRow`, a `groupRow="true"` column or the active sort; when there is none of those, define
+the group on the summary row itself with `field` (or `groupBy`). Note that the summary row only groups, it does not
+sort, so make sure the rows arrive grouped - via `headerRow`, a sorted column or a pre-sorted list.
+
+```xhtml
+<p:dataTable var="car" value="#{dtRowGroupView.cars}">
+    <p:column field="brand" headerText="Brand" sortBy="#{car.brand}" sortOrder="asc" />
+
+    <p:column field="year" headerText="Year" />
+
+    <p:summaryRow field="brand">
+        <p:column style="text-align:right">
+            <h:outputText value="Total:" />
+        </p:column>
+        <p:column>
+            <h:outputText value="#{dtRowGroupView.getTotal(car.brand)}" />
+        </p:column>
+    </p:summaryRow>
+</p:dataTable>
+```
 
 ## Scrolling
 Scrolling makes the header-footer of the table fixed and the body part scrollable. Scrolling is
--- a/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datatable/DataTable052.java
+++ b/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/datatable/DataTable052.java
@@ -21,36 +21,37 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.primefaces.component.summaryrow;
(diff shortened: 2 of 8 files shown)


christophs78:
(Opinion) I like the scope of this PR. Still good to review manually. The generated code is well structured and clean without too verbose javadoc, Claude Opus sometimes generates.

PR #15215, opened 2026-09-08
Fix #14937: consistent table styling for advanced input layouts

### LINKED ISSUE #14937: CSS table consistency including advanced input components
### Description

Hello,

when using autocomplete with multiple `<p:columns>` its basically similar to usage of a `datatable` inside an overlay for us but without styling support (see image with freya 5.0 theme)

Image

It would be consistent if input components that render a table through `<p:column>` like _SelectOneMenu_, _AutoComplete_ _SelectManyMenu_, _SelectOneListbox_ could get the styling options similar as table/treetable have for:

- gridlines
- stripes
- size (small / large)

Currently there are CSS classes for _table_ and _treetable_ in themes for:

- ui-{datatable|treetable}-{gridlines|striped|sm|lg}

They can be added via _styleClass_ attribute or using the respective attributes _showGridLines_, _stripedRows_ or _size_. 

I tried using `<p:autoComplete panelStyleClass="ui-datatable-striped">` but it does not work cause the DOM is different. 

**Options for advanced table input components**

Either adding the options to advanced table layout components.

**or generic approach**

Not sure if there is a use case for a theme to style those similar looking table components differently? Does someone style a treetable different then a datatable?

If not, the html `<table>` element of all those components could share the css for those three attributes like `<p:dataTable showGridLines="true">`. Whats against using `ui-table` generic instead of `ui-datatable, ui-treetable` and others in general?

```
<p:dataTable tableStyleClass="ui-table-gridlines"/>

<p:treeTable tableStyleClass="ui-table-gridlines"/>

<p:selectOneMenu tableStyleClass="ui-table-gridlines"/>

<p:autoComplete tableStyleClass="ui-table-gridlines"/>

<!-- would produce html -->

<table class="ui-table-gridlines">
```

**Edit**

Treetrable has no css class `ui-treetable-striped` in default or premium themes.

melloware:
Closes #14937

`AutoComplete`, `SelectOneMenu`, `SelectOneListbox` and `SelectManyMenu` all render a `<table>` when `<p:column>` children are used, but there was no way to reach the gridline, striping and sizing options that `DataTable` and `TreeTable` expose. `panelStyleClass` does not help, as the reporter found, because it lands on the overlay rather than on the table.

## API

A new `advancedStyleClass` attribute on all four components, written onto the table element of the advanced (column based) layout:

```xhtml
<p:autoComplete value="#{bean.driver}" completeMethod="#{bean.complete}"
                var="driver" itemLabel="#{driver.name}" itemValue="#{driver}"
                advancedStyleClass="ui-table-gridlines ui-table-sm">
    <p:column headerText="Name">#{driver.name}</p:column>
    <p:column headerText="Number">#{driver.number}</p:column>
</p:autoComplete>
```

```html
<table class="ui-autocomplete-items ui-autocomplete-table ui-widget-content ui-widget ui-table-gridlines ui-table-sm">
```

`theme-base` gains four generic modifier classes so the names mean the same thing on every component:

| Class | Applies |
| --- | --- |
| `ui-table-gridlines` | Displays cell borders. |
| `ui-table-striped` | Displays striped rows to visually separate content. |
| `ui-table-sm` | Renders the cells with reduced padding. |
| `ui-table-lg` | Renders the cells with increased padding. |

`DataTable` and `TreeTable` are otherwise untouched and keep their existing `ui-datatable-*` / `ui-treetable-*` classes, so nothing changes for existing markup or custom CSS.

### Why not a `size` attribute

`size` could not be reused as an attribute name: `AutoCompleteBase extends AbstractPrimeHtmlInputText` and `SelectOneListboxBase extends HtmlSelectOneListbox` both already inherit a numeric `size`. That is why sizing goes through `advancedStyleClass="ui-table-sm"` rather than three attributes mirroring DataTable.

## Implementation

One `advanced-table-modifiers($rowSelector)` mixin in `theme-base/_mixins.scss`, `@include`d into each component's own table selector. Nesting it there is deliberate: a flat top-level `.ui-table-sm td` rule would lose to the existing `.ui-autocomplete-items .ui-autocomplete-item.ui-autocomplete-row > td` cell padding, so the mixin takes the component's row selector and inherits enough specificity to win. The striped block re-declares `ui-state-hover` / `ui-state-highlight` for the same reason, the way `_datatable.scss` already does.

Striping uses `:nth-child(even)` rather than renderer assigned odd/even classes, because these panels are rebuilt on the client on every query. Only rows carrying the component's row class are striped, so AutoComplete's `moreText` row is left alone.

No new theme variables: the mixin reuses `$datatableScaleSm` / `$datatableScaleLg`, `$datatableBodyRowOddBg` and the `$datatable*CellPadding` values, so all 33 theme variable files are unchanged. The 30 regenerated `theme.css` files in this diff are dart-sass output from `mvn generate-resources`.

## Also included

- **AutoComplete advanced layout `thead`** now gets the theme rules `_selectonemenu.scss` already had. This is the unstyled header visible in the screenshot on the issue.
- **`TreeTable.stripedRows`** plus `ui-treetable-striped`, closing the gap noted in the issue's Edit. TreeTable had `showGridlines` and `size` but no striping at all.

## Testing

10 new unit tests across 5 renderer test classes, using the existing `CollectingResponseWriter` / `FacesContextMock` pattern. Each asserts the exact emitted `class` attribute both with and without the attribute set, so a regression in class ordering or a stray space fails.

- `mvn clean install -f primefaces/pom.xml` — 0 Checkstyle violations on both rulesets, 490 tests / 0 failures, license check 1128 OK.
- `mvn clean install -f primefaces-themes/pom.xml` — all theme entrypoints compile; a clean rebuild produces no further CSS drift.
- Compiled selectors spot-checked in saga-blue, material-indigo-light, bootstrap4-blue-light and aura-light-emerald.

## Note for premium themes

Premium themes live outside this repo, so `ui-table-*` will not resolve in Freya (the theme in the report) or the others until the same mixin is ported. That is true of any naming choice here, but it means this PR alone does not fully close the issue for premium users.


tandraschko:
same here, i will review next month
please dont merge big features for now

melloware:
Nope and I marked it for 17.0 and I will switch it to DRAFT

PR #15219, opened 2026-09-09
Fix #15218: TabView key-based active tab state

### LINKED ISSUE #15218: TabView: key-based state instead of index-based state
### Description

p:accordionPanel got a key-based state in #13890 / #13901: active accepts the key of a p:tab alongside plain indices. p:tabView still resolves its active tab purely by index via activeIndex, so it has exactly the problem that was described for the accordion.

If the number of rendered tabs changes — via rendered on p:tab, or via value/var on the p:tabView — the index no longer points at the tab the user was on. With multiViewState="true" the stored index is restored onto a differently shaped tab list, and a different tab ends up selected.

Example:
```xhtml
<p:selectBooleanCheckbox value="#{testView.renderFirstTab}" label="render first tab">
    <p:ajax process="@this" update="myTabView"/>
</p:selectBooleanCheckbox>

<p:tabView id="myTabView" activeIndex="0" multiViewState="true">
    <p:tab title="First" rendered="#{testView.renderFirstTab}">
        should stay collapsed
    </p:tab>
    <p:tab title="Second">
        should stay active
    </p:tab>
    <p:tab title="Third">
        should stay collapsed
    </p:tab>
</p:tabView>
```
As soon as the first tab is rendered, activeIndex="0" moves from "Second" to "First". The same happens for a dynamic number of tabs, where the backing list shrinks or grows between requests.

The same API as for p:accordionPanel, so both container components behave consistently:
```xhtml
<p:tabView id="myTabView" active="second" multiViewState="true">
    <p:tab key="first" title="First" rendered="#{testView.renderFirstTab}">
        should stay collapsed
    </p:tab>
    <p:tab key="second" title="Second">
        should stay active
    </p:tab>
    <p:tab key="third" title="Third">
        should stay collapsed
    </p:tab>
</p:tabView>
```
- new active attribute on p:tabView, accepting the key of a tab or a 0-based index (p:tab key already exists, it was added with the accordion change)
- activeIndex stays supported and is only deprecated, so existing applications keep working
- the multi view state stores the key when the active tab defines one, and its index otherwise

### Additional context

- Follow-up of #13890, implemented for p:accordionPanel in #13901.
- Backwards compatible: activeIndex keeps working and is used as a fallback whenever active is not set. Tabs without a key keep being addressed by index, so mixed usage works too.
- One client-side change is unavoidable: the hidden state input is renamed from <clientId>_activeIndex to <clientId>_active (same name the accordion uses), because it now carries a key rather than an index. This affects anyone reading that input directly and needs a migration guide note.
- I have a working implementation (component, renderer, widget, docs, integration tests for static and dynamic tabs) and will open a PR against master once this is triaged.

PR is ready soon.

Mika-techs:
Adapts the key-based state of p:accordionPanel (#13890 / #13901) for p:tabView.

Fixes #15218

What changed

Java
- TabViewBase: new active attribute (String, accepts the key of a p:tab or a 0-based index). activeIndex (int) stays and is only marked deprecated.
- TabView: resolveActiveIndex() maps a key back to the current index and falls back to the deprecated activeIndex when active is not set; resolveActive() returns the key of the active tab, or its index when the tab defines no key. forEachTab(), processUpdates(), restoreMultiViewState() and resetMultiViewState() use them.
- TabViewState: int activeIndex → String active.
- TabViewRenderer: decodes _active, renders the state holder as _active, and writes data-key on the tab header and content elements.

Client side (tabview.widget.js)
- State holder is _active; new resolveIndex(keyOrIndex) resolves the stored value to the tab index, so cfg.selected stays a numeric index and the public widget API is unchanged.
- select() stores the tab key when the tab has one, its index otherwise; partialSubmitFilter for multi view state updated.

Docs
- New "Active Tab" section in docs/17_0_0/components/tabview.md.
- Migration notes in docs/migrationguide/16_0_0.md, next to the existing AccordionPanel entry.

Tests
- TabView006 — static tabs with keys, a preceding tab toggled via rendered.
- TabView007 — dynamic tabs via value/var, growing list.
- Both assert the tab addressed by key is selected initially and stays selected after the index shifts.

Backwards compatibility

activeIndex keeps working and is used whenever active is not set, including its value expression in processUpdates. Tabs without a key are still addressed by index, so mixed usage works.

One deliberate client-contract change: the hidden state input is renamed from <clientId>_activeIndex to <clientId>_active (the name p:accordionPanel already uses), since it now carries a key rather than an index. Documented in the migration guide.

Verification

- mvn install -pl primefaces (full build including the frontend bundle) — OK
- mvn test -pl primefaces — OK
- Checkstyle and license check — 0 violations
- mvn verify -f primefaces-integration-tests/pom.xml -Pintegration-tests,headless,chrome,mojarra-4.0 -Dit.test='TabView00*Test' — 9/9 pass
- Same for -Dit.test='AccordionPanel*Test' — 6/6 pass (no regression on the accordion counterpart)

### DIFF
--- a/docs/17_0_0/components/tabview.md
+++ b/docs/17_0_0/components/tabview.md
@@ -63,6 +63,24 @@ When the tabs to display are not static, use the built-in iteration feature simi
 </p:tabView>
 ```
 
+## Active Tab
+The active tab is defined with the _active_ attribute, which accepts either the 0-based index of the tab or the
+_key_ of a tab. Using keys is recommended whenever tabs are rendered conditionally or created dynamically, since
+indexes shift as soon as tabs are added or removed, whereas a key always refers to the same tab.
+
+```xhtml
+<p:tabView active="second">
+    <p:tab key="first" title="Tab One" rendered="#{bean.showFirstTab}">
+        <h:outputText value="Lorem" />
+    </p:tab>
+    <p:tab key="second" title="Tab Two">
+        <h:outputText value="Ipsum" />
+    </p:tab>
+</p:tabView>
+```
+
+The _activeIndex_ attribute is deprecated in favor of _active_ and will be removed in a future version.
+
 ## Orientations
 Four different orientations are available; _top(default)_ , _left_ , _right_ and _bottom_.
 
--- a/docs/migrationguide/16_0_0.md
+++ b/docs/migrationguide/16_0_0.md
@@ -199,6 +199,10 @@ To the following, so that any long dialog content is now scrollable by default:
 ### AccordionPanel
 * The attribute `activeIndex` has been renamed to `active`, it now supports keys defined with they `key` attribute in `p:tab` alongside indices to smoother handle dynamically rendered tabs. `activeIndex` is now deprecated and wil be removed in a future version.
 
+### TabView
+* The attribute `activeIndex` has been renamed to `active`, it now supports keys defined with the `key` attribute in `p:tab` alongside indices to smoother handle dynamically rendered tabs. `activeIndex` is now deprecated and will be removed in a future version.
+* The hidden state input has been renamed from `<clientId>_activeIndex` to `<clientId>_active`, in line with `p:accordionPanel`. It now holds the key of the active tab, if the tab defines one, and its index otherwise.
+
 ### SelectOneButton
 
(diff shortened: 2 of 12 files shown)


melloware:
@Mika-techs thank you. @tandraschko  said he will review in Mid October.

Mika-techs:
Yeah right he is on vacation.
Thank you both <3

PR #15226, opened 2026-09-11
Fix Chart Showcase failure caused by Jackson version mismatch

cguerrero1205:
## Description

Updates the Jackson BOM used by the PrimeFaces Showcase from `2.21.4` to `2.22.1`.

The current Jackson version causes the Chart.js Showcase examples to fail during initialization with the following exception:

```text
java.lang.NoClassDefFoundError: com/fasterxml/jackson/annotation/JsonApplyView
```

The exception occurs while `software.xdev.chartjs` is creating the chart model and Jackson attempts to use `JsonApplyView`, which is not available with the currently resolved Jackson dependencies.

Updating the Jackson BOM to `2.22.1` aligns the Jackson dependencies and resolves the runtime error.

### Showcase

Go to https://showcase.primefaces.org/ui/chart/bar

image

### Local

1. Build and deploy the PrimeFaces Showcase.
2. Open a Chart.js Showcase example.
3. The `ChartView` bean fails during initialization.
4. The server log contains:

```text
Caused by: java.lang.NoClassDefFoundError:
com/fasterxml/jackson/annotation/JsonApplyView
```

The failure can be reproduced when deploying the Showcase on WildFly.

## Changes

* Update `com.fasterxml.jackson:jackson-bom` from `2.21.4` to `2.22.1` in `primefaces-showcase/pom.xml`.

## Result

After updating the Jackson BOM to `2.22.1`, the Chart.js Showcase initializes successfully and the charts are rendered without the `JsonApplyView` `NoClassDefFoundError`.

image

### DIFF
--- a/primefaces-showcase/pom.xml
+++ b/primefaces-showcase/pom.xml
@@ -40,7 +40,7 @@
             <dependency>
                  <groupId>com.fasterxml.jackson</groupId>
                  <artifactId>jackson-bom</artifactId>
-                 <version>2.21.4</version>
+                 <version>2.22.1</version>
                  <scope>import</scope>
                  <type>pom</type>
             </dependency>   


melloware (approved):

tandraschko:
@melloware what about 16.x? Please note that every fix in master needs propably to ported Back to 16.x the next time

melloware:
i will backport the showcase at PF is on 17.0.0-SNAPSHOT.

Issue #14779, opened 2026-03-18
Error Handling: connection errors must trigger p:ajaxExceptionHandler / error-pages


tandraschko:
it seems currently onerror attributes, and pfAjaxError, and ajaxStatus is triggered.

we have `PrimeFaces.ajax.Utils.handleError` which takes care of p:ajaxExceptionHandler integration and error-page redirect

it seems its called for valid XML responses, as well for JSF AJAX errors, but not for PF/JQuery connection errors.

Follow up of #14773

tandraschko:
this would also mean, that if you define a error-page in web.xml, a connection error will automatically redirect to the error page

if you have a ajaxExceptionHandler, its used as first prio

lprimak:
FYI: `PrimeFaces.ajax.Utils.handleError` is not called for 404, 401, 403, 500, 504 errors etc.

tandraschko:
AjaxStatus005Test is failing now
the ITs defines a error-page in web.xml, which now leads to a redirect. its a actually a small breaking change.

what should we do?
IMO if a error-page defined (or p:ajaxExceptionHandler), it has more priority to correctly handle that error. ajaxStatus is just a "info".
should we disable the test?

melloware:
yeah lets disable the test for now

melloware:
@tandraschko we use this in AjaxStatus widget will this never be reached now?

```js
// Handle AJAX error
        .on('pfAjaxError' + namespace, (e, xhr, settings, error) => {
            this.trigger('error', [xhr, settings, error]);
        })
```

melloware:
i wonder if there is some way to tell a Connection Error

lprimak:
Looking for "connection error" is not enough. Any response besides 200 / XML payload also needs to be handled

tandraschko:
All errors are handled.

tandraschko:
And it will be reached but redirected afterwards propably

lprimak:
Ok, This looks better. Will need to test it, but definitely better than current situation
Will see how it effects #14782 
Thanks @tandraschko

tandraschko:
if you have a p:ajaxExceptionHandler, it will be called now (where you can also do your refresh current page)
if there is no p:ajaxExceptionHandler, we try to lookup the error-page and redirect

if those 2 are not present, it wont affect #14782

weissreto:
[...]

PR #15227, opened 2026-09-11
fix #15225: AJAX connection errors must reach p:ajaxExceptionHandler

### LINKED ISSUE #15225: Error Handling: Connection Error in AJAX requests
### Describe the bug

If a connection error (404, 504, etc ) occurs during an AJAX request the error is ignored and the AJAX status indicator is shown forever. The user has no cloud what is going on. 
This is a followup to https://github.com/primefaces/primefaces/issues/14779. This issue fix that in case of a connection error the method PrimeFaces.ajax.Utils.handleError is called. But because the method is called with an empty string for the parameter errorName, no AjaxExceptionHandler is found (not even the global one).
I suggest instead of an empty error name to provide something like Ajax.ConnectionError so that one can register a special error handler for these kind of errors.

### Reproducer

1. Create a Primefaces Page with a Button and a global p:ajaxExceptionHandler
2. Bind the button to an action listener method that sets a 504 status code on the response
3. When clicking on the button the AJAX indicator shows up but never disappears. The ajaxExceptionHandler is not triggered.

### Expected behavior

ajaxExceptionHandler should be called.

### PrimeFaces edition

Community

### PrimeFaces version

15.0.17

### Theme

_No response_

### JSF implementation

MyFaces

### JSF version

4.1

### Java version

25

### Browser(s)

all

melloware:
Fixes #15225 (follow-up of #14779, reported by @weissreto).

### Problem

#14779 made connection errors call `PrimeFaces.ajax.Utils.handleError`, but it passed an **empty** error name, so the fix never took effect:

1. `handleError` only resolved the `AjaxExceptionHandler` widgets *inside* the `if (errorName)` branch. With an empty name the list stayed empty, so the subsequent global-handler lookup (`exceptionHandlers.find(w => w.isGlobal())`) searched an empty array and the code always fell through to `No ajaxExceptionHandler or error page found for '' defined!`.
2. An empty error name also cannot be targeted by a specific `p:ajaxExceptionHandler type="..."`.

Net effect: on a 404/504/network failure the AJAX status indicator stayed visible forever and no handler ran.

### Fix

- Connection errors are now reported with a dedicated error name **`Ajax.ConnectionError`**, exposed as `PrimeFaces.ajax.CONNECTION_ERROR`, as suggested in the ticket. A handler can be registered for exactly this case:
  ```xhtml
  <p:ajaxExceptionHandler type="Ajax.ConnectionError" onexception="alert('Server not reachable: ' + errorMessage);" />
  ```
- `handleError` now resolves the handler widgets unconditionally, so the **global** `p:ajaxExceptionHandler` is found again.
- The error message carries the HTTP status, e.g. `AJAX failure with HTTP status 504`.

The error-page fallback is unchanged: `Ajax.ConnectionError` is not mapped in `web.xml`, so it still falls back to `java.lang.Throwable` and then to the default error page.

### Tests

New integration-test triads that reproduce the ticket (bean sets HTTP 504 + `responseComplete()`):

- `AjaxExceptionHandler001Test` — a `p:ajaxExceptionHandler type="Ajax.ConnectionError"` is invoked
- `AjaxExceptionHandler002Test` — a **global** `p:ajaxExceptionHandler` is invoked

Verified both fail on `master` and pass with this change:

```
# without the fix
[ERROR] Tests run: 2, Failures: 0, Errors: 2, Skipped: 0

# with the fix
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0
```

Checkstyle (`mvn validate`) clean on both main and test sources.

Docs: `docs/17_0_0/core/errorhandling.md` documents the new error name.

### DIFF
--- a/docs/17_0_0/core/errorhandling.md
+++ b/docs/17_0_0/core/errorhandling.md
@@ -40,10 +40,18 @@ For server-reported errors, the response contains:
 </partial-response>
 ```
 
+Connection errors have no server-reported error name, so they are reported with the dedicated error name
+`Ajax.ConnectionError` (available as `PrimeFaces.ajax.CONNECTION_ERROR`). A `p:ajaxExceptionHandler` can therefore be
+registered for exactly this kind of error:
+
+```xhtml
+<p:ajaxExceptionHandler type="Ajax.ConnectionError" onexception="alert('The server is not reachable: ' + errorMessage);" />
+```
+
 In both cases, PrimeFaces processes the error through the following chain:
 
 1. **Callbacks & Events** — `onerror` callbacks on `p:ajax` are triggered, as well as global listeners like `p:ajaxStatus` and the jQuery `pfAjaxError` event.
-2. **`p:ajaxExceptionHandler`** — If a specific or global `p:ajaxExceptionHandler` is defined, it is invoked with the error details.
+2. **`p:ajaxExceptionHandler`** — If a `p:ajaxExceptionHandler` for the error name is defined, it is invoked with the error details. Otherwise a global `p:ajaxExceptionHandler` (one without a `type`) is invoked.
 3. **Error Page Redirect** — If no `p:ajaxExceptionHandler` is found, PrimeFaces attempts to redirect to the matching `error-page` configured in `web.xml` / `web-fragment.xml`.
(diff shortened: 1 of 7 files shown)


melloware:
@lprimak this is a follow up from the previous fix

lprimak:
LGTM. Thank you.

PR #15239, opened 2026-09-13
fix #12885: FilterMeta must not turn an empty collection filter value into null

### LINKED ISSUE #12885: FilterMeta.resetToNullIfEmpty breaks selection of null values for collection-based filters
### Issue Background
For UISelectMany components, for example, SelectCheckboxMenu, value that is an empty collection (or array) does not mean the same as value == null: in case of an empty collection (or array) nothing is selected, **in case of null an item with value null will be selected**!

Null is a perfectly valid value, same as any other:
```
    <p:selectCheckboxMenu label="Filter" onchange="PF('list').filter();" value="#{clientController.filteredStatuses}"
        converter="selectedEntity">
        <f:selectItem itemLabel="No Status" itemValue="#{null}" />
        <f:selectItems value="#{clientController.clientStatuses}" var="status" itemLabel="#{status.name}" itemValue="#{status}" />
    </p:selectCheckboxMenu>
```
This works perfectly fine in general and results in a collection (array) of selected values (one of which may be null - corresponding to "No Status"item) or empty collection (array) if nothing is selected.

### Problem
Commit https://github.com/Rapster/primefaces/commit/ab0186c4ebb74dd3962bc107eb2f557c9b76f966 introduced `FilterMeta.resetToNullIfEmpty` method that sets FilterMeta.filterValue to null in case filterValue was an empty collection (or array). This immediately broke the behavior of any UISelectMany component used as a facet - because FilterMeta.filterValue suddenly became null (even though backing bean set it to an empty collection), the component will be eventually called by UITable with setValue(null) **which will cause an item with value == null (i.e. "No Status" in the example) to be marked as selected**. Exactly the same SelectCheckboxMenu control works perfectly fine outside of DataTable filter facet, as expected.

### How to Reproduce
Simply add <p:selectCheckboxMenu> filter facet to <p:dataTable> which contains <f:selectItem/> (or any other select item with value=null). Observe that this item will be selected by default. It **will not** be selected in case the same <p:selectCheckboxMenu> is moved outside of filter facet.

### Reproducer

_No response_

### Expected behavior

FIlterMeta should not corrupt the filter value set by the backing bean! If it is set to an empty collection it should be an empty collection, it shouldn't suddenly become null! I'm not sure what is the purpose of `FilterMeta.resetToNullIfEmpty` method, it looks like this method can simply be removed and then the behavior will be as expected?

### PrimeFaces edition

Community

### PrimeFaces version

14.0.5

### Theme

_No response_

### JSF implementation

Mojarra

### JSF version

2.3

### Java version

11

### Browser(s)

_No response_

melloware:
Fixes #12885

Valid report. @melloware — this is exactly your suggestion from the thread (*"leaving the value in place rather than overwriting with NULL"*), plus the one extra change that keeps the integration tests green.

## Root cause

`resetToNullIfEmpty` nulls empty collections/arrays so that `isActive()` — which was simply `filterValue != null` — reports the filter as inactive. That much is intentional.

The problem is that the same value is also written **back into the filter facet** on every render:

```java
// DataTableRenderer#encodeFilter
Object filterValue = component.getFilterValue(column);
column.setFilterValueToValueHolder(context, filterValue);   // -> EditableValueHolder#setValue(null)
```

That sets the facet's *local* value, shadowing its own `value="#{bean.filteredStatuses}"` binding. And `SelectRenderer#isSelected` has:

```java
if (itemValue == null && valueArray == null) {
    return true;
}
```

`SelectManyRenderer#getValues` returns `null` for a `null` component value, so an `<f:selectItem itemValue="#{null}"/>` renders as **selected**.

The full chain: bean sets `[]` → `FilterMeta` stores `null` → facet `setValue(null)` → the null item renders checked → **on submit that item is written back, so the bean's `[]` becomes `[null]`**. The reporter's "FilterMeta should not corrupt the filter value" is literally true — it is data corruption, not just a rendering glitch.

## Why the obvious fix doesn't work

The issue suggests `resetToNullIfEmpty` "can simply be removed". I tried exactly that (keep values, leave `isActive()` as `filterValue != null`) and ran the `DataTable-filter` suite — **16 failures**, which is the *"Definitely breaks a bunch of Integration Tests"* @melloware hit:

```
DataTable002Test.lazyFilter:168                     expected: <10> but was: <0>
DataTable026DatesTest.filterNotBetweenInvalid:187   expected: <8>  but was: <0>
DataTable002Test.lazyRowSelectWithFilterApplied     IndexOutOfBounds: Index 3 out of bounds for length 0
DataTable039Test.lazyFilterSelectCheckboxMenu       Timeout while waiting for AJAX complete!
```

An empty collection becomes an *active* filter, so `in` / `between` constraints match nothing and every row is filtered away.

## Fix

Emptiness becomes a property of the value instead of something encoded by destroying it:

* new `FilterMeta.isEmpty(Object)` — `null`, blank String, empty collection / `Iterable` / array
* `isActive()` → `!isEmpty(filterValue)`, so filtering, `getActiveFilterMeta()` and every lazy model see exactly what they saw before
* `resetToNullIfEmpty` now only normalizes **blank Strings**, which `DataTableRenderer#getFilterValue` relies on for its default-filter fallback chain; empty collections and arrays keep their identity and reach the facet intact

⚠️ `resetToNullIfEmpty` is `public static`, so this narrows its documented behaviour — called out explicitly since it is part of the published surface. The javadoc now explains why, and `isEmpty` is exposed for anyone who wants the old predicate.

## Tests

New triad `datatable/dataTable053.xhtml` + `DataTable053` + `DataTable053Test` — a `p:selectCheckboxMenu` filter facet with an `itemValue="#{null}"` item bound to an empty `List`, which is the reproducer from the issue:

| | on `master` | expected |
|---|---|---|
| no item preselected | `1` checkbox checked | `0` |
| still none after an AJAX update | `1` checkbox checked | `0` |
| backing bean keeps its empty list | `[null]` | `[]` |
| selecting a value still filters | passes | passes (regression guard) |

The first three **fail** on `master` with exactly that output and **pass** with the fix.

Regression: all 259 `DataTable*Test` / `TreeTable*Test` integration tests and the 483 `primefaces` unit tests pass on Mojarra 4.0; the filter-heavy classes also pass on MyFaces 4.0 + CSP. Checkstyle and license check clean.


melloware:
@Rapster can you review this one please

Rapster (approved):

PR #15251, opened 2026-09-15
fix #15131: Menu toggleable must keep the server side expanded state

### LINKED ISSUE #15131: Menu: toggleable ignores the server side expanded state since 15.0.13
### Describe the bug
Since 15.0.13, `<p:menu toggleable="true"> `ignores the expanded state coming from the server — every submenu is expanded on load, whether it comes from `DefaultSubMenu.expanded` or from the expanded attribute of `<p:submenu>.`

The renderer output is correct; `PlainMenu.restoreState() `expands every header that is not in the stored collapsed list, and with nothing stored that list is empty.

Regression from **#14620** (milestone 15.0.13, commit `e3edf9ed8`). 15.0.12 is the last good version

Reproducer: the Programmatic example of the showcase — `MenuView` builds `Options` with `.expanded(true)` and `Navigations` with `.expanded(false)`, yet both render expanded. https://showcase.primefaces.org/ui/menu/menu.xhtml (clear the `*-plainmenu` keys from `localStorage` first, a stored state hides the problem)

### How it affects us
Our left sidebar is built from a programmatic `MenuModel`: ~30 submenus, ~180 items, and the controller marks only the section containing the current page as expanded. Since 15.0.13 all 30 sections open on every page load and the menu becomes an unusable long scroll.

### Workaround
We override `restoreState() ` back to the pre-15.0.13 behaviour — collapse what is stored, never auto-expand:

```
PrimeFaces.widget.PlainMenu.prototype.restoreState = function () {
    var saved = localStorage.getItem(this.stateKey);
    if (!saved) {
        return; // keep the collapsedIds bindToggleable() derived from the markup
    }
    var widget = this;
    saved.split(',').forEach(function (id) {
        if (!id) { return; }
        if (widget.collapsedIds.indexOf(id) === -1) { widget.collapsedIds.push(id); }
        widget.collapseSubmenu($(PrimeFaces.escapeClientId(id)), false);
    });
};
```
It must run before the widget is constructed, so we call it from an inline script right before the `<p:menu>`.

### Feature request
Could `p:menu` get an attribute for the initial state, e.g. `initialState="server | expanded | collapsed"` ? Today only `toggleable` and `statefulGlobal` exist, so there is no supported way to say "start from what the server rendered".

### PrimeFaces edition
Community

### PrimeFaces version
15.0.17

### Theme
Any, not theme related

### JSF implementation
Mojarra (org.glassfish:jakarta.faces)

### JSF version
2.3.21

### Java version
18

### Application server
Jetty 10.0.24

### Browser(s)
chrome latest builds

### Reproducer

_No response_

### Expected behavior

_No response_

### PrimeFaces edition

Community

### PrimeFaces version

15.0.17

### Theme

Vela

### JSF implementation

Mojarra

### JSF version

2.3.21

### Java version

18

### Browser(s)

151.0.7922.138 (Official Build) (arm64)

melloware:
Fixes #15131

## Problem

Since 15.0.13 a `<p:menu toggleable="true">` renders **every** submenu expanded on load, ignoring `expanded` from `<p:submenu>` or `DefaultSubMenu.expanded`.

The renderer is correct — it writes `ui-icon-triangle-1-e` and `display:none` for a collapsed submenu. The widget then throws that away.

`PlainMenu.bindToggleable()` seeds `collapsedIds` from the rendered markup (that's the server state). `restoreState()` ran right after and unconditionally overwrote it:

```ts
if (collapsedIdsAsString) {
    this.collapsedIds = collapsedIdsAsString.split(',');
} else {
    this.collapsedIds = [];   // <-- discards the server state
}
```

With nothing in localStorage the list became empty, and the loop underneath expanded every header — including the brief expand animation reported in the issue.

Regression from e3edf9ed8 (#14620).

## Fix

Only override `collapsedIds` when a state was actually stored on the client.

An empty stored value is a *valid* state meaning "nothing is collapsed", so the check is `!== null` rather than truthiness. Without that, expanding the last collapsed submenu would store `""`, be read back as "no state", and the server state would snap it shut again on the next load.

This is the same semantic `PanelMenu.restoreState()` already implements, so the two menus now agree:

> the server-side state is the initial state; if this is changed on the client, the client side is the new default — @tandraschko

No new attribute is needed.

## Tests

New `Menu002` triad (`menu002.xhtml` / `Menu002.java` / `Menu002Test.java`) covering:

1. declarative `<p:submenu expanded="true|false">` survives load
2. programmatic `DefaultSubMenu.builder().expanded(...)` survives load — the reporter's actual scenario
3. a client-stored **expand** wins over the server state (pins the empty-string case)
4. a client-stored **collapse** wins over the server state

Each asserts the icon class, `aria-expanded`, and actual child visibility.

### Verification

```
mvn verify -f primefaces-integration-tests/pom.xml -Pintegration-tests,headless,chrome,mojarra-4.0 -Dit.test=Menu001Test,Menu002Test
Tests run: 8, Failures: 0, Errors: 0, Skipped: 0 -- BUILD SUCCESS
```

Confirmed the tests actually pin the bug — reverting the widget change makes tests 1, 2 and 4 fail, and using truthiness instead of `!== null` makes test 3 fail. The wider menu suite (MegaMenu 001/001b/002/003 + Menu001) is green.

## Note for the 15.x branch

@melloware the same one-line change applies cleanly to `15.X` if you want a 15.0.18 patch — happy to open that PR too.


melloware:
Companion backports: 16.X → #15252, 15.X → #15253 (for a 15.0.19 patch).

